PR #304, opened 2014-08-25
Handle multiple flash messages on same key

hugo:


lustrousgorilla:
Was this feature subsequently removed? I stumbled across this issue when trying to figure out how to add multiple flash messages per key, but in the current version of Phoenix I don't see `get_all` or its equivalent as a function in the `Phoenix.Controller` module?

alizain:
@chrismccord would you be willing to accept it back?

I read through #554, and I'm at the point where I _do_ need it.

PJUllrich:
I ran into this issue as well handle it by storing a list of flash messages per key. Here an example:

```elixir
# In Phoenix
        previous_messages = get_flash(conn, key) || []

        put_flash(
          conn,
          key,
          [message | previous_messages]
        )

# In Phoenix LiveView
        import Phoenix.LiveView.Utils, only: [get_flash: 1, get_flash: 2]
        
        messages = get_flash(socket) |> get_flash("#{type}") || []

        put_flash(
          socket,
          type,
          [message | messages]
        )
```

Then, in my template, I iterate over the messages per key:

```elixir
# In *.eex templates
  <%= for type <- [:success, :info, :warning, :error] do %>
      <%= for msg <- get_flash(@conn, type) || [] do %>
          <p class="alert-<%= type %>"><%= msg %></p>
      <% end %>
  <% end %>

# In *.leex templates
  <%= for type <- [:success, :info, :warning, :error] do %>
      <%= for msg <- live_flash(@flash, type) || [] do %>
         <p class="alert-<%= type %>"><%= msg %></p>
      <% end %>
  <% end %>
```

Issue #297, opened 2014-08-21
Provide way to filter and validate/cast params


chrismccord:

josevalim:
Ecto changesets will solve precisely this. We can close this. Otherwise, you can just do `Dict.take(...)`.

PR #319, opened 2014-09-06
[WIP] permit_params plug

gjaldon:
@chrismccord were you thinking of something like this for https://github.com/phoenixframework/phoenix/issues/297? It's a simple plug that takes only whitelisted params. 

If this is what you had in mind, below are some of the other stuff I was thinking of implementing:
- Raising an error if `:permit_params` isn't plugged so that devs will make sure to explicitly whitelist attributes
- Add a flag to toggle requiring of the permit_params plug.
- Store unfiltered params somewhere in conn

What do you think?

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -86,9 +86,6 @@ defmodule Phoenix.Controller do
         render_view conn, @subview_module, @layout_module, template, assigns
       end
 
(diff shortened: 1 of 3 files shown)


josevalim:
Closing this one. Ecto will solve this with changesets and right now we are pushing the concern to the data layer (as the data layer knows the expected types anyway).

Issue #320, opened 2014-09-07
Resource Protocol


scrogson:
Would like to get some more discussion/ideas around a `Resource` protocol to be used for "models". This will come in handy for improving route helpers: `Route.user_path(@user)` and many other things.

I have begun to work on some form helpers that would really work best with a protocol like this. Would be great to get some ideas rolling on what this protocol could/should be.

```elixir
defprotocol Phoenix.Resource do
  def param_key(key)
  def model_name(name)
  ...
end

defimpl Phoenix.Resource, for: Any do
  def param_key(_key), do: "id"
end
```

josevalim:
We have added FormData and Param protocol. We will add more if we need them.

PR #321, opened 2014-09-07
Form Helper

scrogson:
This is a WIP spike, please feel free to comment or suggest changes. An end goal for
me to is have something like
[SimpleForm](https://github.com/plataformatec/simple_form/) so hopefully
@josevalim will have some thoughts.

This will also rely heavily on the `Resource` protocol discussed in #320.

### DIFF
--- a/lib/phoenix/html/form.ex
+++ b/lib/phoenix/html/form.ex
@@ -1,4 +1,4 @@
-defmodule Phoenix.View.Helpers.FormBuilder do
+defmodule Phoenix.HTML.Form do
 
   defstruct [
     resource: nil,
@@ -6,13 +6,13 @@ defmodule Phoenix.View.Helpers.FormBuilder do
     opts: []
   ]
 
-  import Phoenix.View.Helpers.TagHelper
+  import Phoenix.HTML.Tag
 
-  alias Phoenix.View.Helpers.FormBuilder
+  alias Phoenix.HTML.Form
 
   def form_for(resource, opts \\ [], func) do
     opts = Dict.put_new(opts, :method, :post)
-    builder = %FormBuilder{
+    builder = %Form{
       resource: resource,
       input_prefix: input_prefix(resource),
       opts: opts}
--- a/lib/phoenix/html/form.ex
+++ b/lib/phoenix/html/form.ex
@@ -25,7 +25,9 @@ defmodule Phoenix.HTML.Form do
 
   def text_field(builder, name, opts \\ []) do
     defaults = [type: "text", value: input_value(builder, name, opts)]
-    attrs = Keyword.put_new(defaults, :name, input_name(builder, name))
+    attrs = defaults
+      |> Keyword.put_new(:name, input_name(builder, name))
+      |> Keyword.put_new(:id, dom_id(builder, name))
     input_tag(:text, attrs)
   end
 
@@ -36,6 +38,10 @@ defmodule Phoenix.HTML.Form do
     |> Module.split
     |> List.last
     |> Phoenix.Naming.underscore
+  end
+
+  defp dom_id(builder, name) do
+    builder.input_prefix <> "_#{name}"
   end
 
   defp input_name(builder, name) do
--- a/lib/phoenix/html/form.ex
+++ b/lib/phoenix/html/form.ex
@@ -31,7 +31,16 @@ defmodule Phoenix.HTML.Form do
     input_tag(:text, attrs)
   end
 
-  def label(name)
+  def label(builder, name) do
+    label_text = name
+      |> to_string()
+      |> String.capitalize()
+      |> String.replace("_", " ")
+    label(builder, name, label_text)
+  end
+  def label(builder, name, label_text) do
+    content_tag(:label, label_text, for: dom_id(builder, name))
+  end
 
   defp input_prefix(resource) do
     resource.__struct__
--- a/lib/phoenix/html/form.ex
+++ b/lib/phoenix/html/form.ex
@@ -25,7 +25,7 @@ defmodule Phoenix.HTML.Form do
 
   def text_field(builder, name, opts \\ []) do
     defaults = [type: "text", value: input_value(builder, name, opts)]
-    attrs = Dict.merge([name: input_name(builder, name)], defaults)
+    attrs = Keyword.put_new(defaults, :name, input_name(builder, name))
     input_tag(:text, attrs)
   end
 
--- a/lib/phoenix/html/form.ex
+++ b/lib/phoenix/html/form.ex
@@ -41,7 +41,7 @@ defmodule Phoenix.HTML.Form do
   end
 
   defp input_value(builder, name, opts) do
-    opts[:value] || Map.get(builder.resource, name)
+    Dict.get(opts, :value, Map.get(builder.resource, name))
   end
 end
 
--- a/lib/phoenix/html/form.ex
+++ b/lib/phoenix/html/form.ex
@@ -26,8 +26,10 @@ defmodule Phoenix.HTML.Form do
   def text_field(builder, name, opts \\ []) do
     defaults = [type: "text", value: input_value(builder, name, opts)]
     attrs = Dict.merge([name: input_name(builder, name)], defaults)
-    tag(:input, attrs)
+    input_tag(:text, attrs)
   end
+
+  def label(name)
 
   defp input_prefix(resource) do
     resource.__struct__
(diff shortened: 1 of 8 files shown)


lib/phoenix/html/tag.ex
```diff
@@ -0,0 +1,97 @@
+defmodule Phoenix.HTML.Tag do
+  @moduledoc ~S"""
+  Helpers related to producing HTML tags within templates.
+  """
+
+  alias Phoenix.HTML.Safe
+
+  @data_attrs [:method, :remote, :confirm]
```
rafaelfranca: In fact we chose to moving away from handling an special set of data attributes on Rails. We just tell people to use the data attribute directly as `tag(:a, "My link", { data: { confirm: 'Something } })` instead of `tag(:a, "My link", confirm: 'Something })`

lib/phoenix/html/tag.ex
```diff
@@ -0,0 +1,85 @@
...
+
+  ## Examples
+      iex> Phoenix.HTML.Tag.tag(:br)
+      "<br>"
+      iex> Phoenix.HTML.Tag.tag(:input, type: "text", name: "user_id")
+      "<input name="user_id" type="text">"
+  """
+  def tag(name),                              do: tag(name, [], true)
+  def tag(name, attrs) when is_list(attrs),   do: tag(name, attrs, true)
+  def tag(name, open)  when is_boolean(open), do: tag(name, [], open)
```
josevalim: Should we care about open? I think the HTML5 spec does not care about this at all anymore.

lib/phoenix/html/tag.ex
```diff
@@ -0,0 +1,85 @@
...
+
+  alias Phoenix.HTML.Safe
+
+  @data_attrs [:method, :remote, :confirm]
+
+  @doc ~S"""
+  Creates an HTML tag with the given name and options.
+
+  ## Examples
+      iex> Phoenix.HTML.Tag.tag(:br)
```
josevalim: You can just have `tag(:br)` in the examples. I think that's how most people will use it anyway as the module will be automatically imported in views. When doing doctests, you just need to do `doctest Phoenix.HTML.Tag, import: true`.

lib/phoenix/html/tag.ex
scrogson: I agree.

lib/phoenix/html/tag.ex
```diff
@@ -0,0 +1,77 @@
...
+    do: acc |> Enum.sort |> tag_attrs
+  defp build_attrs(tag, [{k,v}|t], acc) when is_list(v),
+    do: build_attrs(tag, t, nested_attrs(k, v, acc))
+  defp build_attrs(tag, [{k,v}|t], acc) when v == true,
+    do: build_attrs(tag, t, [{to_string(k),k}|acc])
+  defp build_attrs(tag, [{k,v}|t], acc),
+    do: build_attrs(tag, t, [{to_string(k),v}|acc])
+
+  defp dasherize(value) when is_atom(value), do: dasherize(Atom.to_string(value))
+  defp dasherize(value), do: String.replace(value, "_", "-")
```
josevalim: This doesn't seem to be tested?

lib/phoenix/html/tag.ex
```diff
@@ -0,0 +1,77 @@
...
+
+  @doc """
+  Creates an HTML input field with the given type and options.
+
+  ## Examples
+      iex> Phoenix.HTML.Tag.input_tag(:text, name: "name")
+      "<input name="name" type="text" />"
+  """
+  def input_tag(type, opts \\ [], open \\ false) do
+    tag(:input, Dict.merge([type: type], opts), open)
```
josevalim: `Keyword.put_new(opts, :type, type)` will be much more performatic. :)

lib/phoenix/html/tag.ex
```diff
@@ -0,0 +1,77 @@
...
+  """
+  def tag(name),                              do: tag(name, [], true)
+  def tag(name, attrs) when is_list(attrs),   do: tag(name, attrs, true)
+  def tag(name, open)  when is_boolean(open), do: tag(name, [], open)
+  def tag(name, attrs, open) do
+    "<#{name}#{build_attrs(name, attrs)}#{if open, do: ">", else: " />"}"
+  end
+
+  @doc ~S"""
+  Creates an HTML tag with given name, content, and attributes.
```
josevalim: Doctests. :)

lib/phoenix/html/form.ex
```diff
@@ -0,0 +1,49 @@
...
+    form_tag(opts, do: func.(builder))
+  end
+
+  def form_tag(opts \\ [], [do: block]) do
+    content_tag(:form, opts, do: block)
+  end
+
+  def text_field(builder, name, opts \\ []) do
+    defaults = [type: "text", value: input_value(builder, name, opts)]
+    attrs = Dict.merge([name: input_name(builder, name)], defaults)
```
josevalim: `Keyword.put_new/3` here too!

lib/phoenix/html/tag.ex
```diff
@@ -0,0 +1,77 @@
...
+      iex> Phoenix.HTML.Tag.tag(:br)
+      "<br>"
+      iex> Phoenix.HTML.Tag.tag(:input, type: "text", name: "user_id")
+      "<input name="user_id" type="text">"
+  """
+  def tag(name),                              do: tag(name, [], true)
+  def tag(name, attrs) when is_list(attrs),   do: tag(name, attrs, true)
+  def tag(name, open)  when is_boolean(open), do: tag(name, [], open)
+  def tag(name, attrs, open) do
+    "<#{name}#{build_attrs(name, attrs)}#{if open, do: ">", else: " />"}"
```
josevalim: Shouldn't this be marked as safe?

lib/phoenix/html/tag.ex
scrogson: https://github.com/phoenixframework/phoenix/pull/321/files#diff-c9185479b222ae1eafe498e496463480R13

josevalim: Those are dasherized because of this line: https://github.com/phoenixframework/phoenix/pull/321/files#diff-d7672005258a4ef2ca08bdb00f51e881R55 no?

scrogson: I think it's covered pretty well. This was something we had talked about months ago and was covered.

josevalim: We have a test for using it inside data:

https://github.com/phoenixframework/phoenix/pull/321/files#diff-c9185479b222ae1eafe498e496463480R11

But not outside. If `tag(:input, my_name: ~s("<3"))` is meant to use `my-name` attribute, then it is not covered as the test fails. If you think it is still covered, please point to the proper assertion.

lib/phoenix/html/tag.ex
```diff
@@ -0,0 +1,73 @@
...
+      iex> tag(:input, type: "text", name: "user_id")
+      "<input name=\"user_id\" type=\"text\">"
+  """
+  def tag(name), do: tag(name, [])
+  def tag(name, attrs) when is_list(attrs) do
+    "<#{name}#{build_attrs(name, attrs)}>"
+  end
+
+  @doc """
+  Creates an HTML tag with given name, content, and attributes.
```
josevalim: We need examples for this one.

lib/phoenix/html/form.ex
```diff
@@ -0,0 +1,49 @@
...
+    content_tag(:form, opts, do: block)
+  end
+
+  def text_field(builder, name, opts \\ []) do
+    defaults = [type: "text", value: input_value(builder, name, opts)]
+    attrs = Keyword.put_new(defaults, :name, input_name(builder, name))
+    input_tag(:text, attrs)
+  end
+
+  def label(name)
```
josevalim: This seems to be just hanging?

scrogson: Yeah, needs to be fully implemented and tested.

lib/phoenix/html/tag.ex
scrogson: Ok, this should suffice: https://github.com/phoenixframework/phoenix/commit/f02ac64540171ecfb195508ef695fb5f83cd563e

josevalim:
I have merged the tag functionality. There is a lot of work required on the form one still, so I have left it out. Thanks @scrogson!

schnittchen:
Will the result of content_tag be escaped by Phoenix.HTML.Engine?

josevalim:
I have fixed on another commit. Everything should be marked as safe as a return.

schnittchen:
Wohooo! thank you @josevalim and @scrogson !

PR #507, opened 2014-12-05
Controller Generator

jeregrine:
Simple generator for single controllers.

`mix phoenix.gen.controller controller_name [actions ... optional]`

generates

```
web/controllers/controller_name_controller.ex
web/views/controller_name_view.ex
web/templates/controller_name/
test/controllers/controller_name_test.ex
```

The generator will try to be smart, and output the correct resource to std out. Here are the cases I am going to try and cover:

- If the user enters no actions generate it as a resource in the controller and print to stdout the resource route to copy/paste into router.ex
- If the user enters only valid resource actions I will generate a resource, the actions, and stdout the resource with the "only: [actions, user, choice]" parameter set.
- If the user enters non-standard-resource actions I'll create them in the controller, and output a number of `get '/controller_name/action', ControllerNameController :action` with instructions to edit the `get` to the proper verb and copy into the router. 

Potential issues/Questions:
- I may need them to pass in the application name.

@josevalim @chrismccord lemme know what ya'll think. I think this is maybe step one for scaffolding, once we get ecto we can improve it even more. 

TODO
- [ ] Accept nested paths. EX: `MyApp.API.V1.User` => `api_v1_user` `views/api/v1/user_view.ex` etc


jeregrine:
I am going to close this and direct everyone to https://github.com/nickgartmann/ashes

Issue #478, opened 2014-11-27
Discussions around Phoenix socket/channels abstractions


josevalim:
Here are my thoughts and somewhat a proposal after our discussion yesterday.
1. Transport - responsible for receiving and sending events to a remote client. The transport is responsible for serializing and deserializing data. It is a framework abstraction. The user typically doesn't care about this unless he wants to write a custom ones or forbid some transports (like long pooling).
2. Publisher - responsible for publishing events. At this point, we don't really care about what are those events. It could be a GenServer, a GenEvent, it does not matter. It is a framework abstraction.
3. Topic - the publisher only receive and send events in specific topics. So everytime you subscribe, you need to choose which particular topic do you want to listen to. The topic is just an identifier (there is no abstraction behind a topic) and is made of two parts: "topic:subtopic".
4. Channel - responsible for handling the incoming and outgoing events to a remote client over a transport. "Channel" is potentially a confusing name because it is not a generic channel commonly named in CS (for example, as in Go) but a Channel is rather a connection between the publisher and a transport, filtering and choosing what comes and goes. A channel is a user abstraction, a module with a behaviour.

In other words, Publisher and Transport should exist and be complete on their own. A channel is just convenience functions for connecting those. For example, one should never subscribe to a channel. There is no need to call Phoenix.Channel functions except inside a channel itself.
### Flow

A remote client connects to the server using a transport. Once connected, the remote client can send and receive events to/from a given topic which will be filtered and managed by a channel:

```
channel "/foo/bar", ChannelModule, on: "messages:*", via: [:long_pooling, :websockets]
```

The `:on` option should be either a "topic:subtopic" or "topic:*", the latter meaning the client can join in any subtopic.

The main difference here compared to Phoenix today is that Phoenix uses the word "channel" to identify two things: the channel name and the module. The proposal here is to decouple those too. We simply have topics (and subtopics) and a channel can be associated to any topic:subtopic pair. With this proposal, a channel is only a module where we put code (similar to how a controller is only a module).

PR #570, opened 2014-12-31
Refactor Channel Layer

chrismccord:
I believe I have addressed everything discussed in #478. @josevalim please take a thorough look once you have a chance because it ended up being less rework than I thought. The Publisher (PubSub) API remained exactly the same and is implemented per your description. Notable changes:
1. `socket` API in the router. Socket endpoints are now configured via the `socket` macro and house the channel definitions. Channels can also now be scoped based on transport, i.e.:

``` elixir
  defmodule MyApp.Router do
    use Phoenix.Router

    socket "/ws", MyApp do
      channel "topic1:*",  MyChannel
      channel "baretopic", MyChannel
      channel "wsonly:*",  MyChannel, via: [Phoenix.Transports.WebSocket]
      channel "lponly:*",  MyChannel, via: [Phoenix.Transports.LongPoller]
    end
    ...

```
1. The "topic" abstraction is now _only a string identifier_. We used to conflate the channel name and topic names. Now `channel "foo:*"` refers to the topic "foo" and any pattern following "foo:". By convention, we use "topic:subtopic", but the splat "*" does not require a colon.
2. Likewise, in the channel callbacks, you now match on the topic directly:

``` elixir
# router
channel "rooms:*", RoomChannel

# channel
def join("rooms:lobby", message, socket) do ...
def join("rooms:" <> room_id, message, socket) do ...
```
1. All channel callbacks now accept `socket` as the last argument to more closely follow a GenServer API where the state is held in the last arg and the message is matched int he first.
2. Transports now use two flavors of messages, `{:socket_reply, %Message{...}}` and `{:socket_broadcast, %Message{...})` to differentiate replies from broadcasts. This was required to support `outgoing` callbacks.
3. `event` has been renamed to `incoming` and now channel callbacks can filter/intercept/modify broadcasts with an `outgoing` callback for per-socket customization, i.e.:

``` elixir
def incoming("new:msg", %{"from_id" => from_id, "body" => body}, socket) do
  # broadcast incoming new message to all subscribers
  broadcast socket, "new:msg", %{body: body, from_id: from_id}
end

# each socket's outgoing callback will be triggered for per-socket
# customization of broadcast. Sockets can also not reply to filter broadcasts
# from being relayed.
def outgoing("new:msg", msg, socket) do
  user = socket.assigns[:user]

  reply socket, "new:msg", Dict.merge(msg,
    is_editable: can_edit?(user, msg.from_id),
    is_friend: is_friend?(user, msg.from_id)
  )
end

# by default just forward reply
def outgoing(event, msg, socket) do
  reply socket, event, msg 
end
```
### Points of discussion for the 0.8.0 release:
- Should incoming/outgoing be renamed to a more GenServer'esque API? i.e. `handle_incoming`, `handle_outgoing`? I prefer the current naming, but `handle_` could be more welcoming to newcomers who would find it more familiar to GenServer style.
- Should we embed the `node()` my default in the message to allow per-node filtering? I opted not to since end-users can attach their `node()` themselves in the broadcasted message if they need per-node handling. Subscribers can pick up the message the compare their node().
- Others?
### Points of discussion for 0.9.0+:

As discussed on IRC, we need to tackle handling missed messages with a configurable backend. Today, the LongPoller requires a cookie session, and clustered nodes. The PubSub also requires clustering. We would like to move towards a configurable store with `last_seen_id` tracking of messages, for replaying messages between client drops, as well as a solution to non-clustered pubsub servers.

Let me know how it looks! Thanks

### DIFF
--- a/lib/phoenix/socket.ex
+++ b/lib/phoenix/socket.ex
@@ -1,9 +1,19 @@
 defmodule Phoenix.Socket do
-  alias Phoenix.Socket
 
   @moduledoc """
   Holds state for multiplexed socket connections and Channel authorization
+
+  ## Socket Fields
+
+  * `pid` - The Pid of the socket's transport process
+  * `topic` - The string topic, ie `"rooms:123"`
+  * `router` - The router module where this socket originated
+  * `authorized` - The boolean authorization status, default `false`
+  * `assigns` - The map of socket assigns, default: `%{}`
+
   """
+
+  alias Phoenix.Socket
 
   @derive [Access]
   defstruct pid: nil,
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -97,6 +97,10 @@ defmodule Phoenix.Channel do
         end
       end
 
+   By default, unhandled outgoing events are forwarded to each client as a reply,
+   but you'll need to define the catch-all clause yourself once you define an
+   `outgoing/3` clause.
+
   """
 
   use Behaviour
--- a/lib/phoenix/channel/transport.ex
+++ b/lib/phoenix/channel/transport.ex
@@ -19,8 +19,12 @@ defmodule Phoenix.Channel.Transport do
       remote clients, then deserialing and fowarding message through
       `Phoenix.Transport.dispatch/2`. Finish by keeping state of returned
       HashDict of `%Phoenix.Socket{}`s.
-    * Handle receiving outgoing `%Phoenix.Socket.Message{}`s as Elixir process
-      messages, then encoding and fowarding to connected remote client.
+    * Handle receiving outgoing `{:socket_reply, %Phoenix.Socket.Message{}}` as
+      Elixir process messages, then encoding and fowarding to remote client.
+    * Handle receiving outgoing `{:socket_broadcast, %Phoenix.Socket.Message{}}` as
+      Elixir process messages, then forwarding message through
+      `Phoenix.Transport.dispatch_broadcast/3`. Finish by keeping state of returned
+      HashDict of `%Phoenix.Socket{}`s.
     * Handle receiving arbitrary Elixir messages and fowarding through
       `Phoenix.Transport.dispatch_info/2`. Finish by keeping state of returned
       HashDict of `%Phoenix.Socket{}`s.
--- a/lib/phoenix/socket/message.ex
+++ b/lib/phoenix/socket/message.ex
@@ -1,4 +1,16 @@
 defmodule Phoenix.Socket.Message do
+
+  @moduledoc """
+  Defines a `Phoenix.Socket` Message dispatched over channels.
+
+  The Message format requires the following keys:
+
+    * topic - The String topic  or topic:subtopic pair namespace, ie "messages", "messages:123"
+    * event - The String event name, ie "join"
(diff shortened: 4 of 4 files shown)


jeregrine:
:+1:

lib/phoenix/controller.ex
```diff
@@ -704,4 +704,16 @@ defmodule Phoenix.Controller do
...
+
+  @doc """
+  Invokes the controller plug and calls its action
+
+  ## Examples
+
+      iex> Controller.call_action(conn, MyApp.MyController, :index)
+
+  """
+  def call_action(conn, controller, action) do
```
josevalim: I don't like this because it hides they are just plugs. :( Maybe a convenience we could add in the future but I think it is an important message to get across.

lib/phoenix/router/helpers.ex
```diff
@@ -129,4 +131,52 @@ defmodule Phoenix.Router.Helpers do
...
+        {:error, socket, :bad_transport_match}
+      end
+    end
+  end
+
+  defp to_topic_match(topic_pattern) do
+    case String.split(topic_pattern, "*") do
+      [prefix, ""] -> quote do: <<unquote(prefix) <> _rest>>
+      [bare_topic] -> bare_topic
+      _            -> raise "channels using splat patterns must end with *"
```
josevalim: `raise ArgumentError, "..."` ?

lib/phoenix/socket.ex
```diff
@@ -16,34 +15,34 @@ defmodule Phoenix.Socket do
 
 
   @doc """
-  Sets current channel of multiplexed socket connection
+  Sets current topic of multiplexed socket connection
   """
-  def set_current_channel(socket, channel, topic) do
-    %Socket{socket | channel: channel, topic: topic}
+  def set_current_topic(socket, topic) do
```
josevalim: `put_current_topic` to keep Elixir terminology?

josevalim:
It looks great! :shipit:

We just need overall improvements to the channel/socket documentation and specification:

* We need docs for Phoenix.Channel, explaining the callbacks and what not, today it has no moduledocs.
* We should also make leave, incoming and outgoing part of the Phoenix.Channel behaviour specification (today we only have join)
* We should also discuss if we should ship a default implementation of incoming too
* We should explicitly document what triggers the outgoing callback. If I broadcast something, does this message go throughout outgoing? What if I use reply?
* The documentation for Phoenix.Socket and Phoenix.Socket.Message is lacking too. What are the fields? Which of those fields are public? What do they do? Take a look at `Plug.Conn` for examples if required. Note: I would recommend removing the connection altogether from the socket struct. Maybe we can pass it as argument (or part of it) on join, but we should definitely not keep it around.

lib/phoenix/controller.ex
chrismccord: I agree. We just had a bit of duplication, but it's hardly any work to invoke ourselves, so I'll remove.

lib/phoenix/socket.ex
chrismccord: :+1:

guilleiguaran:
As user of the current implementation of channels I agree totally with the changes done in this pull request :+1:

Issue #571, opened 2015-01-03
Wrong endpoint module used when having more than one project namespace


ephe-meral:
On 0.7.2, I have a project with name Some.Project, generated and slightly modified from the 0.7.2 template project. The endpoint module thus resides in lib/some/project/endpoint.ex at Some.Project.Endpoint.

I also chose to name the app :some_project in the mixfile:

```Elixir
defmodule Some.Project.Mixfile do
  use Mix.Project

  def project do
    [app: :some_project,
     version: "0.0.1",
     elixir: "~> 1.0",
#...
```

When executing `mix phoenix.start` i get a `undefined function: :"SomeProject.Endpoint".start/0` error. 

As a workaround, i renamed the app in the mixfile to `:'some.Project'` which now forces phoenix to resolve the correct module.

Is this the expected behaviour?

josevalim:
Btw, if you migrate to master, this issue will be gone altogether, the only wrong usage will be in `mix phoenix.routes` which you can fix by passing the router name explicitly!

ephe-meral:
@josevalim for the same reason as described, a similar issue occurs in https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/template.ex#L240 - it does not resolve the right templates path since it assumes through my `Some.Project.BarView` that they are located in `web/templates/project/bar` when they actually are in `web/templates/bar`

chrismccord:
> Is this the expected behaviour?

This is expected behavior today. We could do like José said and have a :phoenix_namespace config, but you could also just manually set your root template with in your `Some.Project.View` file.

josevalim:
@chrismccord I think we should not rely on `phoenix_namespace` for views but simply inflect based on the current module. I will push a fix for it soon.

ephe-meral:
Similar issue when looking up the view from the controller side due to this macro: https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/controller.ex#L627

PR #572, opened 2015-01-03
Addresses #571 - Resolves Mix.Phoenix.base module base name by application

ephe-meral:
Addresses #571 - Resolves Mix.Phoenix.base module base name by the module given in the application configuration, instead of simply camel casing the atom given in the mixfile under project `app:`

### DIFF
--- a/lib/mix/phoenix.ex
+++ b/lib/mix/phoenix.ex
@@ -12,11 +12,10 @@ defmodule Mix.Phoenix do
   def base do
     app = Mix.Project.config |> Keyword.fetch!(:app)
 
-    Mix.Config.read!("config/config.exs")
-    |> Keyword.fetch!(app)
-    |> List.keyfind(:phoenix_namespace, 0)
-    |> elem(1)
-    |> module_to_base_name
+    case Application.get_env(app, :phoenix_namespace, app) do
+      ^app -> app |> to_string |> Phoenix.Naming.camelize
+      mod  -> mod |> module_to_base_name
+    end
   end
 
   defp module_to_base_name(mod) do
--- a/lib/mix/phoenix.ex
+++ b/lib/mix/phoenix.ex
@@ -14,12 +14,8 @@ defmodule Mix.Phoenix do
 
     case Application.get_env(app, :phoenix_namespace, app) do
       ^app -> app |> to_string |> Phoenix.Naming.camelize
-      mod  -> mod |> module_to_base_name
+      mod  -> mod |> inspect
     end
-  end
-
-  defp module_to_base_name(mod) do
-    mod |> to_string |> String.replace("Elixir.", "")
   end
 
   @doc """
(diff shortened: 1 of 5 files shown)


josevalim:
Unfortunately this assumes the application was already compiled. I would rather have the configuration inside the `config.exs` file. Maybe something like `:phoenix_namespace` inside the application configuration that could be configured. The reason I am proposing this is because in master we only use `Mix.Phoenix.base` to inflect the default router name and I guess the only other use case is for generators, so we should not assume the app file was already compiled. Thanks for raising the issue!

ephe-meral:
Ah thanks for the hint!

ephe-meral:
Since I found this similar bug in the views/templates, as described in #571, I added the config option as you suggested... Now I'll try to resolve the templates directory bug.

lib/mix/phoenix.ex
```diff
@@ -3,13 +3,24 @@ defmodule Mix.Phoenix do
...
+
   """
   def base do
-    Mix.Project.config
-    |> Keyword.fetch!(:app)
-    |> to_string
-    |> Phoenix.Naming.camelize
+    app = Mix.Project.config |> Keyword.fetch!(:app)
+
+    Mix.Config.read!("config/config.exs")
```
josevalim: Oh, we don't need to read the config. When we got read, we could assume the project was already configured. So all we need to do is invoke `Application.get_env(app, :phoenix_namespace)`. :)

josevalim: However, if no phoenix namespace is set, we need to use the same value as before, so we do need to keep the previous code! We will also need tests. :)

test/mix/phoenix_test.exs
```diff
@@ -3,6 +3,8 @@ defmodule Mix.PhoenixTest do
 
   test "base/0 returns the module base based on the Mix application" do
     assert Mix.Phoenix.base == "Phoenix"
+    Application.put_env(:phoenix, :phoenix_namespace, Phoenix.Sample.App)
```
josevalim: We need to revert this though just on the chance it can affect other tests. :) You can just use delete_env.

lib/mix/phoenix.ex
```diff
@@ -3,13 +3,23 @@ defmodule Mix.Phoenix do
...
-    |> Phoenix.Naming.camelize
+    app = Mix.Project.config |> Keyword.fetch!(:app)
+
+    case Application.get_env(app, :phoenix_namespace, app) do
+      ^app -> app |> to_string |> Phoenix.Naming.camelize
+      mod  -> mod |> module_to_base_name
+    end
+  end
+
+  defp module_to_base_name(mod) do
```
josevalim: This whole function is the same as `inspect(...)`! So you can just call it :)

ephe-meral:
@josevalim Done, i guess :smile:

josevalim:
Thank you!

Issue #581, opened 2015-01-09
[optional] lib/ code reloader


yrashk:
It'd be great if it was possible to enable optional code reloading in lib/ — the reasoning is that if I know what I am doing, I will restart the app should I have processes running there that don't handle code_change well. In my practice, however, I tend to put a lot of "basic" modules there that have nothing to do with processes and therefore perfectly upgradeable; however, currently, this requires me to restart the server each time I change one of those files.

Thoughts?

josevalim:
I have [pushed a commit to Elixir master and v1.0](https://github.com/elixir-lang/elixir/commit/2472bb79abd94ff3c0298a6421a5c38b06ec9e10) that allows Mix to correctly compile subdirectories.

With this, you should be able to change Phoenix to work with directories inside lib too. All you need to do is to change your mix.exs file:

https://github.com/phoenixframework/phoenix/blob/3c980e78e7b589a62937182987cb7015556ba6ce/priv/template/mix.exs#L8

And the code reloader:

https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/code_reloader/server.ex#L59

In fact, we could change the code reloader to pass the directories that are reloadable as argument, which we could read from the endpoint itself. Basically, I am proposing we could change [this line](https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/code_reloader.ex#L47) to something like this:

    opts[:reloader].(conn.private.phoenix_endpoint.config(:reloadable_paths))

And reloadable_paths is the new configuration that stores all reloadable paths. Feel free to explore this, a pull request would be very welcome.

chrismccord:
Sounds good to me

chrismccord:
This is on master now and will be shipping with 0.8.0. Thanks!

PR #582, opened 2015-01-10
Reloadable paths config for Code Reloader

gjaldon:
This PR addresses https://github.com/phoenixframework/phoenix/issues/581

Adds `:reloadable_paths` Endpoint config that is set to `["--elixirc-paths", "web"]` by default in dev.exs.

### DIFF
--- a/priv/template/config/dev.exs
+++ b/priv/template/config/dev.exs
@@ -3,7 +3,8 @@ use Mix.Config
 config :<%= application_name %>, <%= application_module %>.Endpoint,
   http: [port: System.get_env("PORT") || 4000],
   debug_errors: true,
-  cache_static_lookup: false
+  cache_static_lookup: false,
+  reloadable_paths: ["--elixirc-paths", "web"]
 
 # Enables code reloading for development
 config :phoenix, :code_reloader, true
--- a/lib/phoenix/code_reloader.ex
+++ b/lib/phoenix/code_reloader.ex
@@ -13,10 +13,14 @@ defmodule Phoenix.CodeReloader do
   ## Server delegation
 
   @doc """
-  Reloads code within the `web` directory.
+  Reloads code within directories specified in the `:reloadable_paths` config.
+  This is configured in your application environment like:
+
+      config :your_app, YourApp.Endpoint,
+        reloadable_paths: ["--elixirc-paths", "web"]
   """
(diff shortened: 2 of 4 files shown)


chrismccord:
Thanks Gabriel!

josevalim:
Awesome! I think though the paths option should only be ["web"] and we can
use flat_map to convert it to --elixirc-paths. If we want to keep this as
is (which I think is suboptimal) we need to rename the config to
reload_options or something else.

Also, we should add to the documentation the given paths should be a subset
of the elixirc_paths specified in mix.exs!


-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Lead Developer

chrismccord:
I also just realized we need to move the configuration to the `config :phoenix, :code_reloader`, and make it a kwlist:

```
config :phoenix, :code_reloader,
  enabled: true,
  reloadable_paths: ["web"]
```

@josevalim Can you expand on the need for flat_map? Not sure I follow. Also, will we be able to do away with this line?
https://github.com/phoenixframework/phoenix/blob/master/priv/template/mix.exs#L8

josevalim:
We should not put it in the phoenix's code_reloader  configuration because the reloadable_paths are configured per endpoint and not globally. Some endpoint may want to put stuff in "whatever" and it should not affect others.

In fact, the code reloader enabling/disabling is still global but we could even change that and move the code reloader inside the endpoint supervision tree. I just don't know if someone would like to enable/disable code compilation per endpoint.

priv/template/config/dev.exs
```diff
@@ -3,7 +3,8 @@ use Mix.Config
 config :<%= application_name %>, <%= application_module %>.Endpoint,
   http: [port: System.get_env("PORT") || 4000],
   debug_errors: true,
-  cache_static_lookup: false
+  cache_static_lookup: false,
+  reloadable_paths: ["--elixirc-paths", "web"]
```
josevalim: There is no need to add this line to  `dev.exs`. Most people won't configure this, so it could be hidden imo.

josevalim:
@chrismccord the flat_map is just how we can convert ["web"] into ["--elixirc-paths", "web"]. We cannot do away with that line though because that tells Mix everything to be compiled, without the code reloader. In fact, the reloadable_paths must be a subset of the paths listed in mix.exs

PR #584, opened 2015-01-10
Hide reloadable_paths option and add more to doc

gjaldon:
I applied the changes based on your inputs here https://github.com/phoenixframework/phoenix/pull/582.

Need clarification on one thing though, when `reloadable_paths: ["web"]` should `Phoenix.CodeReloader.reload!/1` return `["--elixirc-paths", "web"]`? I also don't get why use flat_map to get  `["--elixirc-paths", "web"]`? Can't I use `["--elixirc-paths"|["web"]]` instead?

### DIFF
--- a/lib/phoenix/code_reloader.ex
+++ b/lib/phoenix/code_reloader.ex
@@ -13,11 +13,14 @@ defmodule Phoenix.CodeReloader do
   ## Server delegation
 
   @doc """
-  Reloads code within directories specified in the `:reloadable_paths` config.
+  Reloads code within the paths specified in the `:reloadable_paths` config.
   This is configured in your application environment like:
 
       config :your_app, YourApp.Endpoint,
-        reloadable_paths: ["--elixirc-paths", "web"]
+        reloadable_paths: ["web"]
+
+  Keep in mind that the paths passed to `:reloadable_paths` must be a subset
+  of the paths specified in the `:elixirc_paths` option of `project/0` in mix.exs.
   """
   @spec reload!([binary]) :: :ok | :noop | {:error, binary()}
   defdelegate reload!(paths), to: Phoenix.CodeReloader.Server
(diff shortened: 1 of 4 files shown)


chrismccord:
In practice, this appears to recompile everything in "lib" regardless of the `reloadable_paths`. Can you double check the behavior?

josevalim:
The current code won't work because nowhere we convert ["web"] into
["--elixirc-paths", "web"]

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Lead Developer

lib/phoenix/code_reloader.ex
```diff
@@ -13,11 +13,14 @@ defmodule Phoenix.CodeReloader do
...
   @doc """
-  Reloads code within directories specified in the `:reloadable_paths` config.
+  Reloads code within the paths specified in the `:reloadable_paths` config.
   This is configured in your application environment like:
 
       config :your_app, YourApp.Endpoint,
-        reloadable_paths: ["--elixirc-paths", "web"]
+        reloadable_paths: ["web"]
+
+  Keep in mind that the paths passed to :reloadable_paths must be a subset
```
josevalim: `:reloadable_paths`

lib/phoenix/code_reloader.ex
```diff
@@ -13,11 +13,14 @@ defmodule Phoenix.CodeReloader do
...
+  of the paths specified in the :elixirc_paths option of `project/0` in mix.exs.
```
josevalim: `:elixirc_paths`. :)

lib/phoenix/endpoint/adapter.ex
```diff
@@ -42,6 +42,7 @@ defmodule Phoenix.Endpoint.Adapter do
      cache_static_lookup: false,
      http: false,
      https: false,
+     reloadable_paths: ["web"],
```
josevalim: Beautiful!

chrismccord:
Merged in 9114fccb468c02f5417a9a9415dd6ec9b8201eb4

<3 <3 <3

PR #589, opened 2015-01-15
Make websocket timeout configurable

ephe-meral:
Cowboy websockets can be initialized with a maximum timeout time. Until now, this has been left to be the default setting, meaning no timeout at all, since cowboy websockets assume `:infinity` as default.

This PR adds an optional endpoint transport configuration setting `websocket_timeout` that specifies the timeout to be used in milliseconds.

Doesn't break backwards-compatibility.

~cheers, Johanna.

### DIFF
--- a/lib/phoenix/endpoint/adapter.ex
+++ b/lib/phoenix/endpoint/adapter.ex
@@ -35,7 +35,8 @@ defmodule Phoenix.Endpoint.Adapter do
      # Transports
      transports: [
        longpoller_window_ms: 10_000,
-       websocket_serializer: Phoenix.Transports.JSONSerializer
+       websocket_serializer: Phoenix.Transports.JSONSerializer,
+       websocket_timeout: :infinity
      ],
 
      # Runtime config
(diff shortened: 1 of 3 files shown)


josevalim:
Beautiful.

josevalim:
@chrismccord a related discussion: do you think we can/should move this configuration to socket/2 in the router now? There is nothing wrong with it in the endpoint though.

josevalim:
@chrismccord alternatively, we could rename the configs to websockets_transport: [...], longpolling_transport: [...] instead of a single transports.

chrismccord:
@ephe-meral Thanks! Per irc discussion, we're going to place this on the `socket` macro. I should have this merged in with the discussed changes this evening.

chrismccord:
So there's an issue with the `:timeout` option on the `socket` macro. Since we allow multiple transports with the `:via` option, a single timeout value doesn't make sense. The only alternative I can think is supporting:

```elixir
socket "/ws", via: [{Phoenix.Transports.WebSocket, [timeout: 30_000]}, ...] do
  channel "rooms:*", RoomChannel
end
```

Keep in mind that the `:via` option can also be used per `channel`. The other option is to keep the timeout in the transport config as this PR does. Preference?

ephe-meral:
Not sure if anybody needs this granularity, but if you think it necessary then I guess this is the perfect place...

josevalim:
@chrismccord I thought we would do:

    socket "/ws", websockets: [timeout: 30_000] do

If we pass all options to the underlying transport, it will be fine as it can look only for the options it cares. I think we could merge this regardless though and open up another discussion for those options.

chrismccord:
<3 <3 <3

PR #593, opened 2015-01-17
Replace : in topic names so that the client does not need to enter it

mgwidmann:
As shown in the test below, joining a "topic:*" topic fails in `v0.8.0` and `master`. Websockets are effectively broken.

Given a Router like:
```elixir
defmodule MyWebSockets.Router do
  use Phoenix.Router

  pipeline :browser do
    plug :accepts, ~w(html)
  end

  scope "/", ElixirRailsWebsockets do
    pipe_through :browser

    get "/interface", PageController, :index

  end

  socket "/ws", MyWebSockets do
    channel "topic:*", MyChannel
  end

end
```

Symptoms: After joining a topic, the phoenix logger only outputs:

```
[info] GET /ws
[debug] Processing by Phoenix.Transports.WebSocket.upgrade/2
  Parameters: %{}
  Pipelines: []
```

Nothing else is logged and the join never completes -- it hangs indefinitely. The "catch all" version of `match_channel` is hit and `:bad_transport_match` is returned. Error handling could be improved here.

The "catch all" is hit because the `match_channel` definition has the topic as `<<"topic:" <> _rest >> when instead the client sent in `<<"topic" <> _rest>>`. Removing the `:` fixes websockets.

### DIFF
--- a/test/phoenix/channel_test.exs
+++ b/test/phoenix/channel_test.exs
@@ -312,8 +312,8 @@ defmodule Phoenix.Channel.ChannelTest do
     message = %Message{topic: "topic1",
                        event: "join",
                        payload: fn socket -> {:ok, socket} end}
-    {:error, _sockets, :bad_transport_match} = Transport.dispatch(message, sockets, self, Router, WebSocket)
-    refute_received {:join, "topic1"}
+    {:ok, sockets} = Transport.dispatch(message, sockets, self, Router, WebSocket)
+    assert_received {:join, "topic1"}
 
     message = %Message{topic: "topic1:somesubtopic",
                        event: "some:event",
@@ -414,8 +414,7 @@ defmodule Phoenix.Channel.ChannelTest do
                        payload: fn socket -> {:ok, socket} end}
     assert {:ok, _sockets} = Transport.dispatch(message, sockets, self, Router, LongPoller)
     assert_received {:join, "topic2-override:somesubtopic"}
-    assert {:error, _sockets, :bad_transport_match} =
-      Transport.dispatch(message, sockets, self, Router, WebSocket)
-    refute_received {:join, "topic2-override:somesubtopic"}
+    assert {:ok, _sockets} = Transport.dispatch(message, sockets, self, Router, WebSocket)
+    assert_received {:join, "topic2-override:somesubtopic"}
   end
 end
--- a/test/phoenix/router/helpers_test.exs
+++ b/test/phoenix/router/helpers_test.exs
@@ -228,4 +228,37 @@ defmodule Phoenix.Router.HelpersTest do
     assert Helpers.socket_url(conn, :upgrade) == url
     assert Helpers.socket_url(__MODULE__, :upgrade) == url
   end
+
+  defp extract_defchannels(channels, pos) do
+    {:__block__, _, block} = Phoenix.Router.Helpers.defchannels(channels)
+    Enum.at(block, pos) |> Macro.to_string()
+  end
+
+  test "defchannels does not generate generic match_channel when there are no channels" do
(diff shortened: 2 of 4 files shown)


test/phoenix/router/helpers_test.exs
```diff
@@ -228,4 +228,37 @@ defmodule Phoenix.Router.HelpersTest do
...
+      def(match_channel(socket, :incoming, <<\"topic\" <> _rest>>, event, msg_payload, transport) when transport in [Phoenix.Transports.WebSocket, Phoenix.Transports.LongPoller]) do
+        apply(Module, :handle_in, [event, msg_payload, socket])
+      end
+      def(match_channel(socket, :outgoing, <<\"topic\" <> _rest>>, event, msg_payload, _transport)) do
+        apply(Module, :handle_out, [event, msg_payload, socket])
+      end
+      def(match_channel(socket, :info, <<\"topic\" <> _rest>>, _event, msg_payload, transport) when transport in [Phoenix.Transports.WebSocket, Phoenix.Transports.LongPoller]) do
+        apply(Module, :handle_info, [msg_payload, socket])
+      end
+    )]
```
mgwidmann: All the `<<\"topic\" <> _rest>>` here without the change in `helpers.ex` are actually printed as `<<\"topic:\" <> _rest>>`  (and then this test fails).

test/phoenix/channel_test.exs
```diff
@@ -421,8 +421,7 @@ defmodule Phoenix.ChannelTest do
                        payload: fn socket -> {:ok, socket} end}
     assert {:ok, _sockets} = Transport.dispatch(message, sockets, self, Router, LongPoller)
     assert_received {:join, "topic2-override:somesubtopic"}
-    assert {:error, :bad_transport_match, _sockets} =
-      Transport.dispatch(message, sockets, self, Router, WebSocket)
-    refute_received {:join, "topic2-override:somesubtopic"}
+    assert {:ok, _sockets} = Transport.dispatch(message, sockets, self, Router, WebSocket)
+    assert_received {:join, "topic2-override:somesubtopic"}
```
mgwidmann: Reason why this (and the other above) were failing was because of the `:` match problem when calling `match_channel`. This test logically should do the same as `LongPoller`, which is a successful join.

chrismccord:
> As shown in the test below, joining a "topic:*" topic fails in v0.8.0 and master. Websockets are effectively broken.

It sounds like you sent up the bare topic "topic", which doesn't match the pattern "topic:". If you want the client to be able to send up "topic", then you need to define it as:

```elixir
channel "topic", MyChannel
```

> Nothing else is logged and the join never completes -- it hangs indefinitely. The "catch all" version of match_channel is hit and :bad_transport_match is returned. Error handling could be improved here.

It doesn't hang, the GET request gets upgraded to a WS connection. We are working on different error handling for the `:bad_transport_match` case, but no issue here.


So you need to either change your client code to use `socket.join("topic:subtopic")` or update your router to use `channel "topic" ...`

mgwidmann:
So having a channel of `channel "topic:*"` means you cannot do `socket.join("topic", ...`? It just seems the `*` indicates either nothing or something, making the front end need to join `socket.join("topic:", ...` instead (which I didn't figure out until looking at the code). Having a `"topic:*"` does not also make it a bare topic?

sorentwo:
The pattern as defined is `topic:` plus anything, just like any glob or regular expression. You would get the same matching behavior from any router, be it client side or server side.

mgwidmann:
Ok, I guess if the `:bad_transport_match` wasn't so silent it wouldn't be so hard to realize I had made a mismatch... 

Thanks!

chrismccord:
@mgwidmann master now logs unmatched topics. Cheers!

PR #594, opened 2015-01-19
PubSub Adapters, PG2, Redis

chrismccord:
This is a working, starting point for PubSub adapters with redis support. I extracted our pg2 setup into a `PubSub.Adapter` behavior and added a `PubSub.RedisAdapter`. I initially tracked redis subscriber state with a HashDict of HashSet approach, but after I started writing a bunch of pid monitoring code, I realized *local* pg2 works perfectly in conjunction with the redis adapter. So I'm using redis to broadcast multi-node an local pg2 to track subscribers.

Notes:

  * I could use some input on the supervision strategy for `Phoenix.PubSub.RedisSupervisor`. I set a giant `max_restarts` for now, otherwise, `redis-server` going down nukes the entire app. // @fishcakez
  * Is the `PubSub.Adapter` behavior well specified? We started with just global pg2, but the API worked just as well for redis. Is it flexible enough for other generalized pubsub adapters?
  * Should we rip out the redis adapter into a `phoenix-pubsub-redis` project now, or wait?

`redis-server` must be running for specs to pass, but all green locally.

It still needs some work, but all feedback appreciated. Thanks!

TODO:
  * [x] Add custom pg2'eque subscriber processes for local subscriptions. Move to single global pg2 group for broadcasting b/w nodes. Basically, we want to make our PG2Adapter look more like RedisAdapter.
  * [x] New local subscriber processes can be responsible for GC of their topic
  * [x] Remove pg2 from RedisAdapter entirely with new local subscriber tracking process
  * [x] Make pubsub API process based. `Phoenix.PubSub.broadcast(:my_process_name, "foo", "bar")`
  * [x] Remove configuration based pubsub module
  * [x] Pool redis clients for broadcasting to redis (single connection remains for phx subscriber)
  * [x] Decide on Channel.broadcast API from outside channels
  * [x] Add MFA style returns from PubSub adapters to offload work back onto the `PubSub` client API


chrismccord:
Also, this moves the Pubsub Adapter into the user's supervision tree

mix.exs
```diff
@@ -23,7 +23,8 @@ defmodule Phoenix.Mixfile do
      env: [code_reloader: false,
            template_engines: [],
            format_encoders: [],
-           pubsub: [garbage_collect_after_ms: 60_000..300_000],
+           pubsub: [adapter: Phoenix.PubSub.PG2Adapter,
```
josevalim: Should all be configurations in the supervision tree when starting. One option is to start the adapter directly, instead of a proxy thing.

pma:
Here's my 2¢:

*"I could use some input on the supervision strategy for Phoenix.PubSub.RedisSupervisor. I set a giant max_restarts for now, otherwise, redis-server going down nukes the entire app."*

I've had better results handling external connection failures directly in the process that holds the connection, instead of letting the error "escape" to the supervisor. In my setup, using the first approach the Supervisor would exhaust all restarts before haproxy had a chance to redirect all new reconnects to a live server node. For Riak connections I'm just using the auto-reconnect feature of the Erlang client. For RabbitMQ, the Erlang client doesn't yet have auto-reconnect built-in, so I'm handling the closed connection error and entering a reconnect, wait loop.

*Is the PubSub.Adapter behavior well specified? We started with just global pg2, but the API worked just as well for redis. Is it flexible enough for other generalized pubsub adapters?
Should we rip out the redis adapter into a phoenix-pubsub-redis project now, or wait?*

Would like to have a try at implementing a phoenix-pubsub-rabbitmq, time permitting. Personally I wouldn't use Redis, so I'd prefer if the Phoenix core didn't depend on the redis client.

chrismccord:
Thanks for the feedback @pma. @josevalim and I discussed the supervision, and we're going to try to let the client handle it in process like you said. That way we can handle expo backoff and reconnects without spamming redis-server from the supervisor. It will also let us be lazy about connections.



> Would like to have a try at implementing a phoenix-pubsub-rabbitmq, time permitting. Personally I wouldn't use Redis, so I'd prefer if the Phoenix core didn't depend on the redis client.

It will only be an optional dep, and we'll extract it entire after the pubsub api settles down. I would love to see a rabbitmq adapter. If you have the time, it would help us ensure we aren't missing anything in the behavior :)

chrismccord:
Progress update on this. Still far from complete and test, but I broke the redis server into two workers and added pooling.

## Pooling
1. A singleton RedisServer is started that subscribes to all "phx:*" messages from redis.
2. When an elixir client broadcasts, it uses a pooled `RedisBroadcaster` to publish the message to redis.
3. When a "phx:*" message is received from redis, the `RedisServer` casts out to a polled `RedisBroadcaster` to do the deserialization and sending to pg2 subscribers.

This design reduces the bottleneck of the single "phx:*" subscriber by offloading as much as possible to the pool where we can.

## Error handling / connections
The connections and error handling have been reworked. The `RedisServer` starts an `:establish_conn` message loop every 5s until either it can successfully start `:eredis_sub` or it exceeds its max attempts, and yields to it's supervisor by crashing. Once an initial redis connection is establish `:eredis` handles reconnection for us, we simply listen for the connection change messages and update our state, but we're doing worry about the connection as soon as we grab a good one.

The same goes for the `RedisBroadcaster`, except I'm lazily starting its connection whenever it is asked to broadcast. We can probably take the same approach as the connect loop, but the idea is you can broadcast anyway unless the `RedisServer` is online.

pma:
A first iteration of the RabbitMQAdapter: https://github.com/pma/phoenix/blob/cm-pubsub-adapters/lib/phoenix/pubsub/rabbitmq_server.ex

This was based on the RedisAdapter. RabbitMQ is being used just to broadcast a message between nodes, so there is one Exchange overall and one Queue for each node.

With the routing mechanism available in RabbitMQ I think it's possible to implement this without using PG2. We could use one Exchange for each topic and any process that subscribes to a certain topic would consume from it's own Queue (bound to the respective Exchange). This would probably mean:

* No need for the GarbageCollector: Exchanges and Queues can be created with the auto_delete: true option, so they're automatically deleted when all consumers disconnect or the last Bind is removed (respectively).

* Not easy to keep track of the list of subscribers to a given topic: With pg2 this can be returned with :pg2.get_members/1. But if they become consumers of a Queue, I don't think there would be an easy way to return the pids. But this list only returns subscribers inside a given node, right? How is it being used?

Other questions in my mind:

* We're assuming non-persistent subscriptions? Meaning if a subscriber temporarily goes away, the events emitted won't be queued for delivery when it reconnects (we use Rabbit as a routing layer, not a persistent queue).

* Why do we need broadcast_from? I noticed the msg_binary we send is always {from_pid, payload}. Will `from_pid` have any meaning when decoded by a different node not connected to the first using Distributed Erlang? Publishers and Subscribers could always be different processes, so broadcast_from wouldn't be needed. If not, we could just have the application set its payload to {self(), my_actual_payload} and use regular broadcast. Then on the receiving end run the from_pid != self() comparison..

lib/phoenix/pubsub/redis_broadcaster.ex
```diff
@@ -0,0 +1,77 @@
...
+  end
+  def handle_call({:publish_to_redis, topic, msg}, _from, state) do
+    handle_publish(topic, msg, state)
+  end
+
+  @doc """
+  Stops the server
+  """
+  def handle_call(:stop, _from, pid) do
+    case :eredis_client.stop(pid) do
```
josevalim: This will make you connect just to stop the connection. It feels this clause should be moved up and also account for already disconnected cases (where stop can likely raise). Also, it is likely this code won't work at all because pid is a map. :D

lib/phoenix/pubsub/redis_broadcaster.ex
```diff
@@ -0,0 +1,77 @@
...
+      :ok -> {:stop, :normal, :ok, pid}
+      err -> {:stop, err, {:error, err}, pid}
+    end
+  end
+
+  @doc """
+  Decodes binary message and fowards to pg2 subscribers of topic
+  """
+  def handle_cast({:forward_to_subscribers, topic, binary_msg}, state) do
```
josevalim: Just a question: in our case here, those subscribers are local, right?

lib/phoenix/pubsub/adapter.ex
```diff
@@ -0,0 +1,71 @@
...
+
+  @doc """
+  Unsubscribes pid from the topic
+  """
+  defcallback unsubscribe(pid :: Pid, topic :: String.t) :: :ok | {:error, reason :: term}
+
+  @doc """
+  Returns list of subscriber pids for given topic
+  """
+  defcallback subscribers(topic :: String.t) :: list | {:error, reason :: term}
```
josevalim: I have just noticed that this function returns all global members to PG2 and all local members for Redis. This means there is something inherently wrong with this function and retrieving all global subscribers will be a very hard task for Redis. I think we should remove this from the API.

lib/phoenix/pubsub/adapter.ex
```diff
@@ -0,0 +1,71 @@
...
+
+  @doc """
+  Broadcasts a message on the topic, excluding sender from receiving broadcast
+  """
+  defcallback broadcast_from(from_pid :: Pid, String.t, message :: Map.t) :: :ok | {:error, reason :: term}
+
+  @doc """
+  Returns lists of strings of all topics that exists on PubSub adapter
+  """
+  defcallback list() :: list
```
josevalim: This can also be hard to implement in Redis. We may need to remove this one too.

lib/phoenix/pubsub/adapter.ex
```diff
@@ -0,0 +1,71 @@
...
+
+  @doc """
+  Creates a PubSub group for given topic to hold subscriptions
+  """
+  defcallback create(topic :: String.t) :: :ok | {:error, reason :: term}
+
+  @doc """
+  Checks if PubSub group exists for given topic
+  """
+  defcallback exists?(topic :: String.t) :: true | false | {:error, reason :: term}
```
josevalim: This one can be tricky too. My recommendation would be to slim down this behaviour module as much as possible. :D

josevalim:
Finally, I just want to say: fantastic work @chrismccord and @pma!

josevalim:
> Not easy to keep track of the list of subscribers to a given topic: With pg2 this can be returned with :pg2.get_members/1. But if they become consumers of a Queue, I don't think there would be an easy way to return the pids. But this list only returns subscribers inside a given node, right? How is it being used?

Yes. I think we should remove the subscriber function from the API. In general, I think we should make the behaviour really small and expand on that later.

> We're assuming non-persistent subscriptions? Meaning if a subscriber temporarily goes away, the events emitted won't be queued for delivery when it reconnects (we use Rabbit as a routing layer, not a persistent queue).

Right now, that seems to be the case. There are no acks nor delivery guarantees.

> Why do we need broadcast_from? I noticed the msg_binary we send is always {from_pid, payload}. Will from_pid have any meaning when decoded by a different node not connected to the first using Distributed Erlang? Publishers and Subscribers could always be different processes, so broadcast_from wouldn't be needed. If not, we could just have the application set its payload to {self(), my_actual_payload} and use regular broadcast. Then on the receiving end run the from_pid != self() comparison..

The broadcast_from is to send a message to everyone but yourself. In this case, the `from_pid` is not of the publisher but of the subscriber sending the message. I am not sure how this can be implemented with Redis and RabbitMQ. Maybe the best option is to store the pid who sent the message and ignore the message on the subscriber process.

lib/phoenix/pubsub/redis_broadcaster.ex
chrismccord: Yep. Just local subscribers. Each node is running the server/pool locally to listen/fwd messages

lib/phoenix/pubsub/redis_broadcaster.ex
chrismccord: derp. will fix!

lib/phoenix/pubsub/adapter.ex
chrismccord: Agreed. We need it for the pg2 GC, but it does need to be part of the behavior. It would be nice to figure this out, since today on PG2Adapter, you can send/collect from all members which is nice, but I agree slimming down and extending later is the way to go.

chrismccord:
> Will from_pid have any meaning when decoded by a different node not connected to the first using Distributed Erlang?

Yes, but it won't match any pids on the other remote node, which is fine.

> We're assuming non-persistent subscriptions? Meaning if a subscriber temporarily goes away, the events emitted won't be queued for delivery when it reconnects (we use Rabbit as a routing layer, not a persistent queue).

Correct. But my hope is we make the behavior in a way that allows a third party to make a durable pubsub adapter.

> No need for the GarbageCollector: Exchanges and Queues can be created with the auto_delete: true option, so they're automatically deleted when all consumers disconnect or the last Bind is removed (respectively).

Nice. If you can get away from the GC, then by all means do it!

> Not easy to keep track of the list of subscribers to a given topic: With pg2 this can be returned with :pg2.get_members/1. But if they become consumers of a Queue, I don't think there would be an easy way to return the pids. But this list only returns subscribers inside a given node, right? How is it being used?

Like @josevalim suggested, I think it's ok to kill `subscribers/1` from the behavior for now.

> If not, we could just have the application set its payload to {self(), my_actual_payload} and use regular broadcast.

This could work, but then the receiving process needs to drop their own messages. I think `broadcast_from` makes sense to include the API.


Thanks for the taking the time to spike out an adapter @pma. It's already paying dividends by trimming down our API :)

pma:
> Yes, but it won't match any pids on the other remote node, which is fine.

I'm must be missing something, but it seems it could potentially match a pid on another VM node, resulting in a subscriber discarding a message he didn't actually publish.

Ran this on two separate side-by-side iex sessions, and the binary representation of both pids match:

*Session 1*
```
$ iex
Erlang/OTP 17 [erts-6.3] [source] [64-bit] [smp:4:4] [async-threads:10] [hipe] [kernel-poll:false]

Interactive Elixir (1.0.2) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> pid = spawn fn -> receive do end end
#PID<0.56.0>
iex(2)> bin1 = :erlang.term_to_binary(pid)
<<131, 103, 100, 0, 13, 110, 111, 110, 111, 100, 101, 64, 110, 111, 104, 111, 115, 116, 0, 0, 0, 56, 0, 0, 0, 0, 0>>
```

*Session 2*
```
$ iex
Erlang/OTP 17 [erts-6.3] [source] [64-bit] [smp:4:4] [async-threads:10] [hipe] [kernel-poll:false]

Interactive Elixir (1.0.2) - press Ctrl+C to exit (type h() ENTER for help)
iex(1)> pid = spawn fn -> receive do end end
#PID<0.56.0>
iex(2)> bin2 = :erlang.term_to_binary(pid)  
<<131, 103, 100, 0, 13, 110, 111, 110, 111, 100, 101, 64, 110, 111, 104, 111, 115, 116, 0, 0, 0, 56, 0, 0, 0, 0, 0>>
```

> Correct. But my hope is we make the behavior in a way that allows a third party to make a durable pubsub adapter.

This would require always mapping a subscriber to a Queue in a predictable way, so the Queue can be created as durable and keeps on receiving messages even if the consumer is not connected.

chrismccord:
> I'm must be missing something, but it seems it could potentially match a pid on another VM node, resulting in a subscriber discarding a message he didn't actually publish.

Hmm. The node() is encoded with the pid, but for non-clustered nodes `:nonode@nohost` might be an issue for the same pid.

chrismccord:
@pma I pushed a change that creates a ref on each RedisServer that we publish with the message. If the remote node doesn't receive an encoded ref matching it's own ref, it skips the `from_pid` check. This should resolve the potential issue.

chrismccord:
Pretty big update on this after @josevalim's feedback:

We now have a `Phoenix.PubSub.Local` server that mostly replaces what we were using `pg2` for and serves as plumbing for other pubsub adapters to use. With the `Local` server, the pubsub adapters have to implement very little to hook into phoenix's system. See the PG2Adapter and RedisServer for examples. For our PG2Adapter, we still use the Local server to track subscriptions, and only register a *single* pg2 group for communication across the cluster. This now removes the global PubSub Server and completely removes the need for our GarbageCollector. The PubSub API is now entirely process based, which will require channels to specify a `:pubsub_server` option. We can inflect it by default, but for now this branch requires it explicitly, i.e. `use Phoenix.Channel, pubsub_server: :my_app_pub`.

Also, @josevalim, @fishcakez, and I talked about using Supervisor.start_child for the Local servers, but I chose to just start_link them inside `init/1` since we don't need a registered name for the Local server. All thoughts/feedback appreciated. Thanks!

chrismccord:
@fishcakez, I'll be moving to `Supervisor.start_child` as @josevalim just enlightened me as to why you recommended it in the first place :)

pma:
I have a rough implementation of the Phoenix.PubSub.RabbitMQAdapter without using PG2 to broadcast inside the node. Instead it registers the subscriber pid directly as a consumer and RabbitMQ will route the message directly to the process. I haven't looked yet at the new Local module to see how this would fit with the new changes.

```iex
iex(1)> Phoenix.PubSub.RabbitMQ.Supervisor.start_link []  
{:ok, #PID<0.142.0>}
iex(2)>  pid = spawn fn -> receive do msg -> IO.inspect(msg) end end
#PID<0.233.0>
iex(3)> Phoenix.PubSub.RabbitMQAdapter.subscribe pid, "my_topic"   
:ok
iex(4)>  Phoenix.PubSub.RabbitMQAdapter.broadcast "my_topic", "HELLO"
:ok
"HELLO" 
```

A topic is mapped directly to a fanout Exchange. When adding a new subscriber, it opens a new Channel, and creates a new Queue bound to that Exchange. Both the Exchange and Queue are created with `auto_delete: true`, so when the subscriber terminates all resources are released. 

The above example shows that when the broadcast message is consumed, the subscriber terminates and consequently the consumer ends. Then the Queue is deleted because it has no more consumers, the Exchange is deleted because there are no longer Queues bound to it and the Channel is closed.

I didn't see a use for either `create` or `delete` in the adapter, since subscribe will create a topic if it does not exist and it will be removed when there are no subscribers left. 

The Server has to keep a HashDict to map a subscriber pid to a {channel, consumer_tag}. Otherwise it wouldn't be possible to unsubscribe.

lib/phoenix/channel.ex
```diff
@@ -124,17 +124,34 @@ defmodule Phoenix.Channel do
                                                                      {:leave, Socket.t} |
                                                                      {:error, reason :: term, Socket.t}
 
-  defmacro __using__(_options) do
+  defmacro __using__(options \\ []) do
     quote do
       @behaviour unquote(__MODULE__)
-      import unquote(__MODULE__)
+      @pubsub_server unquote(options[:pubsub_server])
+      import unquote(__MODULE__), except: [broadcast: 4,
```
josevalim: I would use `:only` because there is less chance of accidentally importing something we don't want.

lib/phoenix/channel.ex
```diff
@@ -124,17 +124,34 @@ defmodule Phoenix.Channel do
                                                                      {:leave, Socket.t} |
                                                                      {:error, reason :: term, Socket.t}
 
-  defmacro __using__(_options) do
+  defmacro __using__(options \\ []) do
     quote do
       @behaviour unquote(__MODULE__)
-      import unquote(__MODULE__)
+      @pubsub_server unquote(options[:pubsub_server])
```
josevalim: We are going to generate a default value here, right?

josevalim: Also, you should unquote the options and access the value, otherwise you are assuming the options are a keyword list at compilation time (usually true but not always :)).

lib/phoenix/channel.ex
```diff
@@ -150,12 +167,12 @@ defmodule Phoenix.Channel do
...
-  def broadcast(topic, event, message) when is_binary(topic) do
-    broadcast_from :global, topic, event, message
+  def broadcast(server, topic, event, message) when is_binary(topic) do
+    broadcast_from server, :none, topic, event, message
   end
 
-  def broadcast(socket = %Socket{}, event, message) do
-    broadcast_from :global, socket.topic, event, message
+  def broadcast(server, socket = %Socket{}, event, message) do
+    broadcast_from server, :none, socket.topic, event, message
```
josevalim: `:none` is MUCH better than `:global`.

lib/phoenix/pubsub.ex
```diff
@@ -21,150 +20,39 @@ defmodule Phoenix.PubSub do
...
-  Returns a List of all Phoenix PubSubs from :pg2
-  """
-  def list do
-    :pg2.which_groups |> Enum.filter(&match?({@pg_prefix, _}, &1))
-  end
-
-  defp call(message), do: GenServer.call(Server.leader_pid, message)
-
-  defp group(topic_name), do: {@pg_prefix, topic_name}
+  def list(server), do: call(server, :list)
```
josevalim: I am still unsure if we should keep this. It seems to be a very expensive operation to perform globally. If it is meant to be local, I would use `local_topics` or something like that.

lib/phoenix/pubsub.ex
```diff
@@ -21,150 +20,39 @@ defmodule Phoenix.PubSub do
...
-      :ok
-      iex> PubSub.delete("activetopic")
-      {:error, :active}
-
+  Returns lists of subscriber pids of members of pg2 group for topic
   """
-  def delete(topic_name) do
-    call {:delete, group(topic_name)}
-  end
+  def subscribers(server, topic),
```
josevalim: I am still unsure if we should keep this. It seems to be a very expensive operation to perform globally. If it is meant to be local, I would use local_subscribers or something like that.

lib/phoenix/pubsub/local.ex
```diff
@@ -0,0 +1,212 @@
...
+
+  @doc """
+  Subscribes the pid to the topic
+
+    * `pid` - The subscriber Pid
+    * `topic` - The string topic, ie "users:123"
+
+  ## Examples
+
+      iex> PubSub.Local.subscribe(:local_server, self, "foo")
```
josevalim: I would use `Phoenix.PubSub.Local` in the examples, no aliasing.

josevalim: Or assume they are imported!

lib/phoenix/pubsub/local.ex
```diff
@@ -0,0 +1,212 @@
...
+  end
+
+  @doc false
+  def subscription(local_server, pid) do
+    GenServer.call(local_server, {:subscription, pid})
+  end
+
+  @doc false
+  def init(_) do
+    Process.flag(:trap_exit, true)
```
josevalim: We have discussed this locally but I will repeat it here.

We should likely `Process.flag(:trap_exit, true)` indeed. However, we need to add a test because the only reason we are adding this is because we want the subscriber to link to this process (effectively crashing the subscriber when the publisher dies). My proposal is to add a `:link` option to `subscribe` which automatically links both processes, providing a complete use case for this feature.

lib/phoenix/pubsub/local.ex
```diff
@@ -0,0 +1,212 @@
...
+  for an example integration.
+  """
+
+  @doc """
+  Starts the server
+
+    * `server_name` - The name to registered the server under
+
+  """
+  def start_link(server_name) do
```
josevalim: As we discussed, we should always expect a name, this will make it much easier to put it inside a supervision tree.

lib/phoenix/pubsub/adapter.ex
```diff
@@ -0,0 +1,36 @@
+defmodule Phoenix.PubSub.Adapter do
```
josevalim: This is gone right? We have moved to a message protocol so this is not used for anything.

lib/phoenix/pubsub/local.ex
```diff
@@ -0,0 +1,212 @@
...
+    {:ok, %{topics: HashDict.new, pids: HashDict.new}}
+  end
+
+  def handle_call(:list, _from, state) do
+    {:reply, Dict.keys(state.topics), state}
+  end
+
+  def handle_call({:subscription, pid}, _from, state) do
+    case HashDict.fetch(state.pids, pid) do
+      {:ok, {_ref, topics}} -> {:reply, topics, state}
```
josevalim: I would go with with `{:ok, topics} | :error` as return type. Maybe this is a great opportunity to start adding some typespecs to the public interface.

lib/phoenix/pubsub/local.ex
```diff
@@ -0,0 +1,212 @@
...
+      :error      -> state
+      {:ok, topic_pids} ->
+        case HashDict.fetch(state.pids, pid) do
+          :error      -> state
+          {:ok, {ref, subd_topics}} ->
+            topic_pids  = HashSet.delete(topic_pids, pid)
+            subd_topics = HashSet.delete(subd_topics, topic)
+
+            cond do
+              Enum.any?(topic_pids) && Enum.any?(subd_topics) ->
```
josevalim: I would have store those into `has_topics?` and `has_subscriptions?` vars to avoid traversing the HashSet multiple times.

josevalim: Actually, it seems if you handle both cases separately, you can get rid of the duplication:

```elixir
topics = 
  if Enum.any?(topic_pids) do
    HashDict.put(state.topics, topic, topic_pids)
  else
    HashDict.delete(state.topics, topic)
  end

pids =
  if Enum.any?(subd_topics) do
    HashDict.put(state.pids, pid, {ref, subd_topics})
  else
    HashDict.delete(state.pids, pid)
  end
```

lib/phoenix/pubsub/local.ex
```diff
@@ -0,0 +1,212 @@
...
+  end
+
+  defp drop_subscriber(state, pid, ref) do
+    case HashDict.get(state.pids, pid) do
+      {^ref, topics}    ->
+        Enum.reduce(topics, state, fn topic, state ->
+          drop_subscription(state, pid, topic)
+        end)
+
+      _ref_pid_mismatch -> state # handle chance that DOWN pid doesn't match ref
```
josevalim: Yeah, just ignoring down like here is fine.

lib/phoenix/pubsub/pg2_adapter.ex
```diff
@@ -0,0 +1,58 @@
+defmodule Phoenix.PubSub.PG2Adapter do
```
josevalim: This module is beautiful. :heart:

lib/phoenix/pubsub/pg2_adapter.ex
```diff
@@ -0,0 +1,58 @@
...
+  end
+
+  def init(_opts) do
+    {:ok, local_pid} = Phoenix.PubSub.Local.start_link()
+    :ok = :pg2.create(@private_pg2_group)
+    :ok = :pg2.join(@private_pg2_group, self)
+    {:ok, %{local_pid: local_pid}}
+  end
+
+  def handle_call({:subscribe, pid, topic}, _from, state) do
```
josevalim: Maybe one idea to avoid blocking this too much is to support "redirect" in `Phoenix.PubSub`. Basically, if you return `{:redirect, pid_or_name}`, we will do the same call from `Phoenix.PubSub` but to the given pid and topic.

test/phoenix/pubsub/local_test.exs
```diff
@@ -0,0 +1,90 @@
...
+    :ok = PubSub.Local.subscribe(:phxpub, pid, "topic3")
+    :ok = PubSub.Local.unsubscribe(:phxpub, self, "topic3")
+    assert PubSub.Local.subscribers(:phxpub, "topic3") |> Enum.to_list == [pid]
+
+    :ok = PubSub.Local.unsubscribe(:phxpub, pid, "notexists")
+    assert PubSub.Local.subscribers(:phxpub, "notexists") |> Enum.to_list == []
+  end
+
+  test "pid is removed when DOWN and topic dropped if last subscriber" do
+    pid = spawn fn -> :timer.sleep(10000) end
```
josevalim: `:timer.sleep(:inifnity)`?

lib/phoenix/pubsub/pg2_adapter.ex
josevalim: Let's just call it `PG2`? As users will be starting it directly, I think we can skip the `Adapter` bit.

lib/phoenix/pubsub/redis_server.ex
```diff
@@ -0,0 +1,164 @@
...
+
+  @doc """
+  Initializes the server.
+
+  An initial connection establishment loop is entered. Once `:eredis_sub`
+  is started successfully, it handles reconnections automatically, so we
+  pass off reconnection handling once we find an initial connection.
+  """
+  def init(opts) do
+    {:ok, local_pid} = Phoenix.PubSub.Local.start_link()
```
josevalim: We also need to move this one to the supervisor. :)

lib/phoenix/pubsub/redis_server.ex
```diff
@@ -0,0 +1,164 @@
...
+    opts = Dict.merge(@defaults, opts)
+    opts = Dict.merge(opts, host: String.to_char_list(to_string(opts[:host])),
+                            password: String.to_char_list(to_string(opts[:password])))
+
+    Process.flag(:trap_exit, true)
+    send(self, :establish_conn)
+
+    {:ok, struct(RedisServer, opts: opts,
```
josevalim: Why using `struct/2` here?

lib/phoenix/pubsub/redis_server.ex
```diff
@@ -0,0 +1,164 @@
...
+      result = :poolboy.transaction :phx_redis_pool, fn worker_pid ->
+        GenServer.call(worker_pid, {:publish_to_redis, "phx:#{topic}",
+                                   {1, state.node_ref, from_pid, msg}})
+      end
+      {:reply, result, state}
+    end
+  end
+
+  def handle_info({:pmessage, "phx:*", "phx:" <> topic, binary_msg, _client_pid}, state) do
+    :poolboy.transaction :phx_redis_pool, fn worker_pid ->
```
josevalim: So the pool here is not helping with anything because we are only using it from this server. We need to improve the design of the Redis adapter generally. Please ping me later so we can talk about it.

lib/phoenix/pubsub.ex
chrismccord: I was planning on keeping it and `@doc false` it. I use it for testing/debugging/perf measuring

lib/phoenix/pubsub.ex
chrismccord: We don't need it. I was just using for debugging. `@doc false` ?

lib/phoenix/pubsub/adapter.ex
chrismccord: Yes this is gone. Now purely message protocol (which needs doc'd)

lib/phoenix/pubsub/local.ex
chrismccord: link option :+1:

lib/phoenix/pubsub/local.ex
chrismccord: This is better. I wasn't thrilled with the cond approach.

lib/phoenix/pubsub/pg2_adapter.ex
chrismccord: :+1:

lib/phoenix/pubsub/pg2_adapter.ex
chrismccord: `{:redirect, pid_or_name}` is a great idea!

lib/phoenix/pubsub/redis_server.ex
chrismccord: We don't need to. I'll move to maps

lib/phoenix/pubsub.ex
josevalim: We can't `@doc false` because the adapters need to implement it. I can think of two options:

1. Rename them to `local_*`

2. Once we make the "remote" and "local" process named, you will be able to access the local process directly from tests, so we can skip this completely and go to the local directly

I think I am more inclined towards 2.

lib/phoenix/pubsub/redis_server.ex
josevalim: I meant the `struct/2` call instead of `%RedisServer{...}`. :D But getting rid of the struct also works (:+1: from me).

lib/phoenix/channel.ex
chrismccord: Yeah I was planning on inflecting it from the app name. Like `:my_app` -> `:my_app_pub`. It would also match the name that we generate/pass in the generated supervision tree.

josevalim: Why not inflect it like everything else, i.e.: `MyApp.PubSub`?

lib/phoenix/pubsub.ex
chrismccord: Option 2 definitely. I didn't think about how easy it would be with the named process

pma:
@chrismccord @josevalim 
If the Local module will manage local subscriptions, should it be able to notify the adapter once the last subscriber leaves a given topic (so the adapter can do some cleanup)? For the Redis adapter this would be calling UNSUBSCRIBE or PUNSUBSCRIBE. For RabbitMQ it could be deleting a Bind.

Note: My earlier comment refers to an experiment using direct deliveries to the subscriber pid, using just the capabilities of the RabbitMQ client. This would mean having one queue per subscriber, instead of one queue per node on the broker.

chrismccord:
> If the Local module will manage local subscriptions, should it be able to notify the adapter once the last subscriber leaves a given topic (so the adapter can do some cleanup)?

I'm not sure. The idea is you use the pubsub adapter as just a tiny bridge across nodes, and let the Local do the rest of the work. Under this design, you only need a single topic on the adapter, so there's nothing to cleanup really. More elaborate setups that want to manage all the topics on the adapter (which is totally doable), likely would have no use for `Local`, so I don't think we need notifications?

josevalim:
I started writing a reply saying I agreed with @pma and arrived to the conclusion I agree with @chrismccord. :P

@pma your setup would be prone for race conditions. Take this:

1. PID1 subscribes to topic "foo"
2. Adapter starts monitoring to topic "foo"
3. PID1 dies
4. PID2 subscribes to topic "foo"
5. Adapter receives the notification PID1 died and topic should be removed
6. topic "foo" is removed and PID2 does not receive any message

And the solution for this would be for you track the topics and pids yourself which means you are reimplementing Local and have no need for it anymore.

lib/phoenix/pubsub/pg2.ex
```diff
@@ -0,0 +1,58 @@
...
+
+  def init(_opts) do
+    {:ok, local_pid} = Phoenix.PubSub.Local.start_link()
+    :ok = :pg2.create(@private_pg2_group)
+    :ok = :pg2.join(@private_pg2_group, self)
+    {:ok, %{local_pid: local_pid}}
+  end
+
+  def handle_call({:subscribe, pid, topic}, _from, state) do
+    {:reply, GenServer.call(state.local_pid, {:subscribe, pid, topic}), state}
```
josevalim: Btw, instead of `{:redirect, pid}`, we should do `{:client, module, fun, args}`. This is because we mostly want to execute this in the client instead of blocking the server. Moving to the MFA approach will allow us to be more generic, allowing @pma to also optimize his code even though he is not using local. It would also allow us to optimize the broadcast and connection pool usage in the Redis one.

lib/phoenix/pubsub/redis_server.ex
```diff
@@ -0,0 +1,164 @@
...
+  of `@max_connect_attemps` is tried, at which point the server terminates with
+  an `:exceeded_max_conn_attempts`.
+  """
+  def handle_info(:establish_conn, %RedisServer{status: :connected} = state) do
+    {:noreply, state}
+  end
+  def handle_info(:establish_conn, %RedisServer{reconnect_attemps: count} = state)
+    when count >= @max_connect_attemps do
+
+    {:stop, :exceeded_max_conn_attempts, state}
```
josevalim: Let me start with some background since I am copying some folks here. This module is responsible to connect to Redis pubsub and receive notifications. Once we receive those notifications from Redis, we send them to the local subscribers. Currently we try to connect to Redis 3 times right after start, with 5 seconds interval, and then crash if we can't connect.

After thinking about this, I don't think we should crash. We should just continue trying to connect every X seconds. The reason is this: do you want the whole website to fail if we can't connect to Redis? Because that's what will happen after enough time. I would rather have that part of the website disabled.

If we do this change, we need to decide what is going to happen when someone tries to subscribe locally and there is no Redis connection. We have two options then:

1. We don't allow subscribers to connect in the first place. We change the subscribe function to return either `:ok` or `{:error, _}` and we will return an error if Redis is not available. The channel/transport/whatever should handle this case and not allow new connections.

2. We allow subscribers to connect but, because Redis is down, they won't be able to send any messages. We need to decide if being unable to send messages is a hard crash of the channel (and therefore the client) or if we should just return an `{:error, _}` tuple that will be handled in the channel (and hopefully notified to the client).

I am more inclined towards 2. That's because we don't need subscriptions to work to use channels. I may have a channel that only uses replies and never broadcasts.

The issue though is that currently developers are not handling the output of broadcast in their channels. Furthermore, handling the output for every broadcast will quickly be painful. Maybe we should just exit with something meaningful (as call and cast usually do) and allow the transport to handle those specially, often notifying the client. We could even provide a callback in JS that would be invoked for exactly the cases where you can't broadcast messages.

Btw, here is a great article on the topic: http://ferd.ca/it-s-about-the-guarantees.html

/cc @ericmj @fishcakez

chrismccord:
Ok. More progress on this:

  * Both PG2 and Redis adapters are now supervisors and both start the Local server in their supervision tree
  * All registered process names and pubsub adapter topics are now namespaced under the `:name` option provided to the supervisor. This will allow starting any number of pubsub adapters without conflicts
  * The poolboy Redis pool has been removed for now until we finalize offloading work to the PubSub client api (more on that below)


### Offloading work to the client PubSub API
@josevalim and I have discussed offloading work from the pubsub servers by offloading the call's back to the PubSub API so the work is performed in the caller's process. We can do this by return an MFA style respond to certain pubsub calls and the client can transparently execute the required work. I believe I've addressed most feedback so far. I'll be improving docs and starting to benchmark against our old PubSub setup to see how we're doing. Thanks!

chrismccord:
Good news on the preliminary benchmark front. Using @seejay's channel benchmark app, this branch is is 10% faster than `0.8` with the new PG2/Local adapter, with Redis/Local clocking in at 15% slower than the new PG2 adapter.  These initial numbers are really promising because we should have a lot of room for improvement from here with the planned client offloading and potential pooling for redis.

lib/phoenix/naming.ex
```diff
@@ -53,6 +53,25 @@ defmodule Phoenix.Naming do
...
+  ## Examples
+
+      iex> Phoenix.Naming.module_to_pub_server(MyApp.MyChannel)
+      MyApp.PubSub
+
+      iex> Phoenix.Naming.module_to_pub_server(MyApp.Admin.MyChannel)
+      MyApp.PubSub
+
+  """
+  def module_to_pub_server(mod) do
```
josevalim: It would be nice to make this generic passing the PubSub as argument because there are other places we could use it, like here: https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/controller.ex#L729-L732 :)

lib/phoenix/pubsub/pg2.ex
```diff
@@ -0,0 +1,39 @@
...
+        supervisor(Phoenix.PubSub.PG2, [[name: MyApp.PubSub]]),
+      ]
+
+  ## Options
+
+    * `name` - The required name to register the PubSub processes, ie: `MyApp.PubSub`
+
+  """
+
+  def start_link(opts) do
```
josevalim: Let's make the name an argument (not an option) since it is now always required anyway? :)

lib/phoenix/pubsub/pg2.ex
```diff
@@ -0,0 +1,39 @@
...
+    supervisor_name = Module.concat(__MODULE__, Keyword.fetch!(opts, :name))
```
josevalim: I think we should use the same convention we use for local here. I.e.:

    Module.concat(Keyword.fetch!(opts, :name), Supervisor)

lib/phoenix/pubsub/pg2.ex
```diff
@@ -0,0 +1,39 @@
...
+
+  @doc false
+  def init(opts) do
+    server_name = Keyword.fetch!(opts, :name)
+    local_name  = Module.concat(server_name, Local)
+    server_opts = Keyword.put(opts, :local_name, local_name)
+
+    children = [
+      worker(Phoenix.PubSub.Local, [local_name], restart: :transient),
+      worker(Phoenix.PubSub.PG2Server, [server_opts], restart: :transient),
```
josevalim: Let's pass server_name and local_name as arguments too since they are required now?

josevalim: Why is it transient? For testing? Maybe for tests you can rather kill the supervisor?

lib/phoenix/pubsub/pg2_server.ex
```diff
@@ -0,0 +1,66 @@
...
+    {:reply, GenServer.call(state.local_name, {:subscribe, pid, topic}), state}
+  end
+
+  def handle_call({:unsubscribe, pid, topic}, _from, state) do
+    {:reply, GenServer.call(state.local_name, {:unsubscribe, pid, topic}), state}
+  end
+
+  def handle_call({:broadcast, from_pid, topic, msg}, _from, state) do
+    case :pg2.get_members(state.namespace) do
+      {:error, {:no_such_group, _}} -> {:stop, :no_global_group, state}
```
josevalim: You can ignore this clause and just do pids when is_list(pids) below. I.e. just let it crash as we assume no_such_group will never be returned. Unless it can be returned on partitions or something, then we maybe should handle it.

lib/phoenix/pubsub/pg2_server.ex
```diff
@@ -0,0 +1,66 @@
...
+    Process.flag(:trap_exit, true)
+
+    :ok = :pg2.create(pg2_namespace)
+    :ok = :pg2.join(pg2_namespace, self)
+
+    {:ok, %{local_name: local_name, namespace: pg2_namespace}}
+  end
+
+  def handle_call({:subscribe, pid, topic, link}, _from, state) do
+    if link, do: Process.link(pid)
```
josevalim: I think the link should be in the local one. After all, the local one is the responsible for tracking local stuff. If we do so, we can remove the trap exit from this process.

lib/phoenix/pubsub/redis.ex
```diff
@@ -0,0 +1,49 @@
+defmodule Phoenix.PubSub.Redis do
```
josevalim: If we change the PG2 one, we will need to change it here!

lib/phoenix/pubsub/redis_server.ex
```diff
@@ -0,0 +1,176 @@
...
+            eredis_sub_pid: nil,
+            eredis_pid: nil,
+            status: :disconnected,
+            node_ref: :erlang.make_ref,
+            reconnect_attemps: 0,
+            opts: opts}}
+  end
+
+  def handle_call({:subscribe, pid, topic, link}, _from, state) do
+    if link, do: Process.link(pid)
```
josevalim: Same here: I think the link should be in the local one.

lib/phoenix/pubsub/redis_server.ex
```diff
@@ -0,0 +1,176 @@
...
+    {:reply, GenServer.call(state.local_name, {:subscribe, pid, topic}), state}
```
josevalim: Remember to use the local API (instead of a direct GenServer call) once we move to the client offloading!

lib/phoenix/pubsub/redis_server.ex
```diff
@@ -0,0 +1,176 @@
...
+
+    Process.flag(:trap_exit, true)
+    send(self, :establish_conn)
+
+    {:ok, %{local_name: local_name,
+            namespace: redis_namespace(server_name),
+            eredis_sub_pid: nil,
+            eredis_pid: nil,
+            status: :disconnected,
+            node_ref: :erlang.make_ref,
```
josevalim: Are refs guaranteed to be unique across nodes? It may even be the same if it goes through the same initialization process?

lib/phoenix/pubsub/redis_server.ex
```diff
@@ -0,0 +1,176 @@
...
+
+  On init, an initial conection to redis is attempted when starting `:eredis_sub`.
+  If failed, the connection is tried again in `@reconnect_after_ms` until a max
+  of `@max_connect_attemps` is tried, at which point the server terminates with
+  an `:exceeded_max_conn_attempts`.
+  """
+  def handle_info(:establish_conn, %{reconnect_attemps: count} = state)
+    when count >= @max_connect_attemps do
+
+    {:stop, :exceeded_max_conn_attempts, state}
```
josevalim: We still need to discuss if we are getting rid of this before merging as per our previous discussion about failure handling!

lib/phoenix/pubsub/redis_server.ex
```diff
@@ -0,0 +1,176 @@
...
+    {:noreply, %{state | status: :disconnected}}
+  end
+
+  def terminate(_reason, %{status: :disconnected}) do
+    :ok
+  end
+  def terminate(_reason, state) do
+    case :eredis_client.stop(state.eredis_sub_pid) do
+      :ok ->
+        case :eredis_client.stop(state.eredis_pid) do
```
josevalim: Maybe those cases should not be nested. Always shut both down regardless of the outcome? If you are worried about them being started, you may need to `catch :exit, {:noproc, _}`.

lib/phoenix/pubsub/redis_server.ex
```diff
@@ -0,0 +1,176 @@
...
+
+  def handle_call({:unsubscribe, pid, topic}, _from, state) do
+    {:reply, GenServer.call(state.local_name, {:unsubscribe, pid, topic}), state}
+  end
+
+  def handle_call({:subscribers, topic}, _from, state) do
+    {:reply, GenServer.call(state.local_name, {:subscribers, topic}), state}
+  end
+
+  def handle_call({:broadcast, from_pid, topic, msg}, _from, state) do
```
josevalim: It seems we no longer need the pool indeed, given the performance is good. If we are going to use the pool, we would need to spawn tasks here which could mess up ordering. Unless we add last_message_id then we could re-order them. So :+1 for no pool.

Question though: does the redis client (not the pubsub) automatically reconnects too?

josevalim: Also what happens if we can't establish a connection to redis and a broadcast message arrives? It would probably fail with `noproc` bringing this process down?

mix.lock
```diff
@@ -1,10 +1,12 @@
 %{"cowboy": {:hex, :cowboy, "1.0.0"},
   "cowlib": {:hex, :cowlib, "1.0.1"},
   "earmark": {:hex, :earmark, "0.1.10"},
+  "eredis": {:git, "git://github.com/wooga/eredis.git", "3c9dfc5cf4e2cd79276b138692d63d3b1d298a52", []},
   "ex_doc": {:hex, :ex_doc, "0.7.0"},
   "inch_ex": {:hex, :inch_ex, "0.2.3"},
   "linguist": {:hex, :linguist, "0.1.4"},
   "plug": {:hex, :plug, "0.10.0"},
   "poison": {:hex, :poison, "1.3.0"},
+  "poolboy": {:hex, :poolboy, "1.4.2"},
```
josevalim: We can remove this one. :D

josevalim:
Great work! Even better news to know performance is better even without client offloading!

pma:
@josevalim @chrismccord Will you postpone or abandon connection pooling for Redis? The RabbitMQ adapter would still benefit from a connection pool, at least with the approach of opening one channel per subscriber, for a couple of reasons:

1) There is a hard limit of 65,535 channels per connection (a connection maps to a TCP connection and a channel is a way to multiplex multiple message flows in the same connection). 
We could get away with just one connection, one channel, and then have multiple consumers sharing the same channel. But this would mean one process handling all message dispatch (there's one dispatch process per channel). And broadcasts would keep the same dispatch process busy.
With a connection pool we would have channels more or less distributed among all connections.

2) With RabbitMQ we can easily have HA (I know about sentinel for Redis, but didn't find it easy to configure). We can connect to any node in the cluster to publish or consume messages, so a possible setup is to have HAProxy in the client node doing TCP load balancing to the RabbitMQ cluster nodes. In case one node failure, only a subset of subscribers would get disconnected. Also, the recovery will be faster because when they resubscribe they will use pull an active connection from the pool.

josevalim:
@pma you can definitely add pooling to the RabbitMQ one. The way things are currently structured, you just need you start your RabbitMQ-based supervision tree and have the pool init. The pool is not part of the "adapter interface".

lib/phoenix/pubsub/redis_server.ex
chrismccord: No, I'm sure they are not guaranteed to be unique across nodes, this just makes it much less likely. Any thoughts on other solutions? I suppose we could write a node_ref counter to a redis key, and INCR that on init to ensure a unique ref?

fishcakez: Refs include node information just like pids. I forget the precise details but a reference is approximately a 62bit counter with node name and a couple of bits that get flipped based on the node creation. Beam is very predictable, so just like iex always produces the same return value for `self()`, `make_ref()` will probably return the same value. Therefore, just like pids, references are only really of use inside the same cluster during the lifetime of the node that create them.

chrismccord: Thanks James, then we can't rely on it for non-clustered environment, especially since the node()'s will all be `:nonode@nohost`

chrismccord: I just pushed a change that generates the `node_ref` using redis' `INCR` command to keep a node unique node counter. Whenever a redis connection is established (or re-established), I bump the counter and grab the new node ref. This should fix our issues and ensure `node_ref`'s are always unique on the non-clustered nodes.

lib/phoenix/pubsub/redis_server.ex
chrismccord: done!

lib/phoenix/pubsub/pg2_server.ex
chrismccord: :+1: on let it crash. done.

lib/phoenix/naming.ex
chrismccord: Great idea, done in https://github.com/phoenixframework/phoenix/commit/7c485d910b6dbf230079fd0775504890a74d0431

lib/phoenix/pubsub/redis_server.ex
chrismccord: It should bring the process down. I'll keep this in mind for `broadcast` vs `broadcast!`

lib/phoenix/pubsub/redis_server.ex
```diff
@@ -0,0 +1,172 @@
...
+
+    {:noreply, %{state | eredis_sub_pid: eredis_sub_pid,
+                         eredis_pid: eredis_pid,
+                         status: :connected,
+                         node_ref: make_node_ref(eredis_pid, state.namespace),
+                         reconnect_attemps: 0}}
+  end
+
+  defp make_node_ref(eredis_pid, namespace) do
+    {:ok, count} = :eredis.q(eredis_pid, ["INCR", "#{namespace}:node_counter"])
```
josevalim: Beautiful.

lib/phoenix/pubsub/redis_server.ex
```diff
@@ -0,0 +1,172 @@
...
+    end
+  end
+
+  @doc """
+  Decodes binary message and fowards to local subscribers of topic
+  """
+  def handle_info({:message, _redis_topic, bin_msg, _cli_pid}, state) do
+    {_vsn, remote_node_ref, from_pid, topic, msg} = :erlang.binary_to_term(bin_msg)
+
+    if remote_node_ref == state.node_ref do
```
josevalim: Another alternative is to push this to the transport. But we would need IDs in the messages. :S

josevalim: s/transport/channel

chrismccord:
I just pushed client offloading and it boosts redis performance by 11%! The PG2 offloading only saw marginal speed boosts and I left out offloading the pg2 broadcast. Offloading the broadcast only bumped pg2 by 1-2%, and introduced a lot of complexity. I've also pushed `broadcast!` functions. @josevalim What else do we need for 0.9.0?

lib/phoenix/pubsub/local.ex
```diff
@@ -0,0 +1,219 @@
...
+
+  ## Examples
+
+      iex> subscribe(:local_server, self, "foo")
+      :ok
+
+  """
+  def subscribe(local_server, pid, topic, opts \\ [])
+  def subscribe(local_server, pid, topic, link: link),
+    do: GenServer.call(local_server, {:subscribe, pid, topic, link && true})
```
josevalim: Is `&& true` an attempt to cast it as a boolean?

lib/phoenix/pubsub/pg2_server.ex
```diff
@@ -0,0 +1,61 @@
...
+    {:reply, response, state}
+  end
+
+  def handle_call({:unsubscribe, pid, topic}, _from, state) do
+    response = {:perform, {GenServer, :call, [state.local_name, {:unsubscribe, pid, topic}]}}
+    {:reply, response, state}
+  end
+
+  def handle_call({:broadcast, from_pid, topic, msg}, _from, state) do
+    case :pg2.get_members(state.namespace) do
```
josevalim: Let it crash, let it crash, let it crash, oh let it crash!
There could be a trouble, let it craaaaaaaash, craaaash.

:D

lib/phoenix/pubsub.ex
josevalim: We can get rid of this now and call it directly in the local one. The plus side, we reduce the API required in the adapter! :D

lib/phoenix/pubsub.ex
```diff
@@ -21,150 +47,88 @@ defmodule Phoenix.PubSub do
...
   end
 
-  @doc """
-  Check if PubSub is active. To be active it must be created and have subscribers
-  """
-  def active?(topic_name) do
-    call {:active?, group(topic_name)}
+  @doc false
+  # Returns list of all topics under local server, for debug and perf tuning
+  def list(server_name) do
```
josevalim: The tests and tools can now just call Local directly!

lib/phoenix/pubsub/pg2_server.ex
```diff
@@ -0,0 +1,61 @@
...
+    pg2_namespace = pg2_namespace(server_name)
+
+    :ok = :pg2.create(pg2_namespace)
+    :ok = :pg2.join(pg2_namespace, self)
+
+    {:ok, %{local_name: local_name, namespace: pg2_namespace}}
+  end
+
+  def handle_call({:subscribe, pid, topic, link}, _from, state) do
+    response = {:perform, {GenServer, :call, [state.local_name, {:subscribe, pid, topic, link}]}}
```
josevalim: Let's call the public Local API instead of doing inner GenServer calls?

josevalim: It will make it easier in the future if we refactor the Local to, let's say, a GenFSM.

josevalim:
This is beautiful. I think we need to discuss failure handling on Redis and how to take it from the channel to the client. It definitely isn't a blocker for 0.9.0. :shipit:

lib/phoenix/pubsub/pg2_server.ex
chrismccord: haha. I actually needed to match on `:no_such_group` so we can return `{:error, reason}` to accommodate the pubsub adapter contract. Otherwise, I'd let it crash in horrible ways.

lib/phoenix/pubsub/local.ex
chrismccord: yes

chrismccord:
Redis pooling is now back. I think this is just about ready to :ship:

pma:
@chrismccord @josevalim 
I'd like to reuse the same test cases for the RabbitMQ adapter, but [tests depending on Local](https://github.com/phoenixframework/phoenix/blob/cm-pubsub-adapters/test/phoenix/pubsub/pubsub_test.exs#L41) makes it impossible. The RabbitMQ adapter can't use Local because it needs to be able to map {topic, subscriber_pid} to a {channel, consumer_tag} for unsubscribes to work.

One option would be to make specific test cases for the RabbitMQ adapter, but we should be using the same tests to guarantee a consistent behavior across adapters. It would be preferable to have a way to kill the adapter that worked without the assumption that the adapter uses Local.

Another issue is how to implement `make_node_ref`. There is no way to generate guaranteed unique IDs using RabbitMQ (other than hacks like creating a dummy queue with a server assigned name). But `:crypto.strong_rand_bytes(16)`  would generate statistically unique IDs. This implementation detail can be left to the adapter, but I was thinking that the encode/decode of the message payload that travels across nodes should be common. This would make it easier, for example, to later implement some message tracing facility that would work with all adapters.

lib/phoenix/pubsub/redis_conn.ex
```diff
@@ -0,0 +1,39 @@
...
+
+  def handle_info({:EXIT, pid, _}, {pid, opts}) do
+    {:noreply, {:disconnected, opts}}
+  end
+
+  def handle_info(_, state) do
+    {:noreply, state}
+  end
+
+  def terminate({:disconnected, _}), do: :ok
```
josevalim: My gist was wrong (sorry). Terminate receives the reason as first argument.

josevalim:
@pma I would say we rather need to clean up the tests. We can provide the tests in a module and to run the tests you need to implement a callback module with how to get specific information, like the number of subscribers and list of topics. If you use Local, it should be just a call to Local, otherwise, you provide your own. That's the direction we have gone with Ecto to test multiple adapters and we will need this once we extract the Redis adapter out.

> Another issue is how to implement make_node_ref. There is no way to generate guaranteed unique IDs using RabbitMQ (other than hacks like creating a dummy queue with a server assigned name). But :crypto.strong_rand_bytes(16) would generate statistically unique IDs.

I think this is a great idea! Much simpler and will solve the problem nicely! :+1:

chrismccord:
@josevalim is going to hate me, but I got the LongPoller.Server working over pubsub for non-distributed setups. I believe keeping the server is the way to go, and we can cheat over pubsub to get the same benefits. Once we move to `last_seen_id`, we can look at alternatives, but this will allow shipping 0.9.0 with LongPolling support for Heroku et al. I've committed it on a separate branch just in case it's a terrible idea. Feedback appreciated:

https://github.com/phoenixframework/phoenix/tree/cm-pubsub-longpoll

PS: In that branch, I moved the `pubsub_server` onto the application router, and now I'm convinced it's the way go. @josevalim I'll catch on IRC sometime and we can discuss more. Thanks!
// @ericmj

lib/phoenix/pubsub/redis_server.ex
```diff
@@ -0,0 +1,166 @@
...
+
+    {:noreply, %{state | eredis_sub_pid: eredis_sub_pid,
+                         status: :connected,
+                         node_ref: make_node_ref(state)}}
+  end
+
+  defp make_node_ref(state) do
+    :poolboy.transaction state.pool_name, fn eredis_conn ->
+      {:ok, eredis_pid} = GenServer.call(eredis_conn, :eredis)
+      {:ok, count} = :eredis.q(eredis_pid, ["INCR", "#{state.namespace}:node_counter"])
```
josevalim: What do you think of using crypto random bytes as @pma suggested?

chrismccord:
@josevalim I've made most of the discussed changes on the LP over pubsub branch https://github.com/phoenixframework/phoenix/tree/cm-pubsub-longpoll . I've had some thoughts on using router for pubsub, and I think the endpoint is better. Here's my proposal for broadcasting outside of the context of a socket:

```elixir
# config
  ...
  pubsub_server: MyApp.PubSub,
  ...

# within controller
Phoenix.Channel.broadcast(conn, "rooms:123", "new:msg", %{...})
Phoenix.Channel.broadcast(MyApp.Endpoint, "rooms:123", "new:msg", %{...})

# within channel
broadcast(socket, "rooms:123", "new:msg", %{...})
broadcast(MyApp.Endpoint, "rooms:123", "new:msg", %{...})
```

A little more verbose, but this is very similar to our URL helpers, and the Endpoint the natural place to scope pubsub. Thoughts?

josevalim:
If we are going to have the pubsub in the endpoint, we could explore other interesting things:

1. Automatically start the pubsub as part of the endpoint supervision tree
2. Have the pubsub functions in the endpoint itself

chrismccord:
@josevalim The pubsub adapters are now started in the Endpoint's supervision tree and the LP token is salted. I think this is ready to go for 0.9

lib/phoenix/endpoint.ex
```diff
@@ -139,6 +162,30 @@ defmodule Phoenix.Endpoint do
     end
   end
 
+  defp pubsub() do
```
josevalim: Beautiful.

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -6,7 +6,7 @@ defmodule Phoenix.Endpoint.Supervisor do
   use Supervisor
 
   def start_link(otp_app, endpoint) do
-    if endpoint.config(:server) do
+    if endpoint.config(:server) || endpoint.config(:pubsub)[:adapter] do
```
josevalim: I have renamed this module to `server.ex` in master exactly predicting the addition of the pubsub. So I would move the pubsub out of here and add it directly to the endpoint supervision tree (which we would change to be one for one too).

lib/phoenix/pubsub.ex
```diff
@@ -1,170 +1,140 @@
...
+    * `opts` - The optional list of options. Supported options
+               only include `:link` to link the subscriber to
+               the pubsub adapter
   """
-  def exists?(topic_name) do
-    call {:exists?, group(topic_name)}
-  end
+  def subscribe(server, pid, topic, opts \\ [])
+  def subscribe(server, pid, topic, link: link),
+    do: call(server, {:subscribe, pid, topic, link && true})
```
josevalim: I think we should pass the options to the server. We won't need to change the API in case we add other options in the future and we won't need to have the link casting logic in multiple places.

lib/phoenix/pubsub/local.ex
josevalim: Btw, this won't be cast as a boolean if link is `nil`. Another reason for just rolling with options. :)

lib/phoenix/pubsub/pg2.ex
```diff
@@ -0,0 +1,34 @@
...
+    Supervisor.start_link(__MODULE__, name, name: supervisor_name)
+  end
+
+  @doc false
+  def init(server_name) do
+    local_name = Module.concat(server_name, Local)
+
+    children = [
+      worker(Phoenix.PubSub.Local, [local_name]),
+      worker(Phoenix.PubSub.PG2Server, [[name: server_name, local_name: local_name]]),
```
josevalim: Since those are required, maybe pass them as arguments?

lib/phoenix/pubsub/pg2.ex
```diff
@@ -0,0 +1,34 @@
...
+    ]
+    supervise children, strategy: :one_for_all
```
josevalim: `:one_for_all` or `:rest_for_all`? If PG2Server crashes, should we crash the local? I would guess we don't because the local won't be reachable anyway?

lib/phoenix/pubsub/redis.ex
```diff
@@ -0,0 +1,75 @@
...
+
+  @doc false
+  def init([server_name, opts]) do
+    opts = Keyword.merge(@defaults, opts)
+    opts = Keyword.merge(opts, host: String.to_char_list(to_string(opts[:host])),
+                               password: String.to_char_list(to_string(opts[:password])))
+
+    pool_name   = Module.concat(server_name, Pool)
+    local_name  = Module.concat(server_name, Local)
+    server_opts = Keyword.merge(opts, name: server_name,
```
josevalim: Same here: since those are required, maybe pass them as arguments?

lib/phoenix/router/resource.ex
```diff
@@ -35,13 +35,15 @@ defmodule Phoenix.Router.Resource do
...
     singular = Keyword.get(options, :singular)
+    private  = Keyword.get(options, :private, %{})
     actions  = extract_actions(options, singular)
 
-    collection  = [path: path, as: as]
+    route       = [as: as, private: private]
+    collection  = [path: path, as: as, private: private]
     member_path = if singular, do: path, else: Path.join(path, ":#{name}_#{param}")
-    member      = [path: member_path, as: as, alias: alias]
+    member      = [path: member_path, as: as, alias: alias, private: private]
```
josevalim: This is a commit I pushed to master and somehow ended up here. We need to be careful when merging!

josevalim:
@chrismccord I am done reviewing this. I have also removed all old comments so the only comments left in the "FIles changed" view are the ones we need to act on!

lib/phoenix/transports/long_poller.ex
```diff
@@ -79,75 +72,119 @@ defmodule Phoenix.Transports.LongPoller do
...
-  defp session_key(conn), do: "#{router_module(conn)}_longpoll_pid"
-
-  @doc """
-  Sets the active listener process. Called by polling `conn`
-  """
-  def set_active_listener(server_pid, listener_pid) do
-    GenServer.call(server_pid, {:set_active_listener, listener_pid})
+  def ack(conn, priv_topic, messages) do
+    ref = :erlang.make_ref()
+    :ok = broadcast_from(conn, priv_topic, {:ack, messages, ref})
```
josevalim: Let's ack just the length. Otherwise we will end-up copying all messages once again!

lib/phoenix/transports/long_poller.ex
```diff
@@ -79,75 +72,119 @@ defmodule Phoenix.Transports.LongPoller do
...
+  end
+
+  defp sign(conn, priv_topic) do
+    salt = derive_salt(conn, to_string(pubsub_server(conn)), [])
+    Plug.Crypto.MessageVerifier.sign(priv_topic, conn.secret_key_base <> salt)
+  end
+
+  defp verify(conn, priv_topic, sig) do
+    salt = derive_salt(conn, to_string(pubsub_server(conn)), [])
+    case Plug.Crypto.MessageVerifier.verify(sig, conn.secret_key_base <> salt) do
```
josevalim: The secret should be only the salt, including the `conn.secret_key_base` may even be unsafe.

lib/phoenix/transports/long_poller.ex
```diff
@@ -79,75 +72,119 @@ defmodule Phoenix.Transports.LongPoller do
...
+    Phoenix.PubSub.subscribe(pubsub_server(conn), self, priv_topic, link: true)
+  end
+
+  defp broadcast_from(conn, priv_topic, msg) do
+    Phoenix.PubSub.broadcast_from(pubsub_server(conn), self, priv_topic, msg)
+  end
+
+  defp sign(conn, priv_topic) do
+    salt = derive_salt(conn, to_string(pubsub_server(conn)), [])
+    Plug.Crypto.MessageVerifier.sign(priv_topic, conn.secret_key_base <> salt)
```
josevalim: The secret should be only the salt, including the `conn.secret_key_base` may even be unsafe.

lib/phoenix/transports/long_poller.ex
```diff
@@ -79,75 +72,119 @@ defmodule Phoenix.Transports.LongPoller do
...
+      _ -> :error
+    end
+  end
+
+  defp derive_salt(%Plug.Conn{secret_key_base: base}, _key, _key_opts)
+    when base == nil or byte_size(base) < 64 do
+
+    raise "conn.secret_key_base must be at least 64 bytes for longpoll token verification"
+  end
+  defp derive_salt(conn, key, key_opts) do
```
josevalim: `key_opts` are always an empty list, so we should just remove it.

lib/phoenix/pubsub/redis_server.ex
chrismccord: Since it's easy to get a guaranteed unique value here, might as well?

lib/phoenix/transports/long_poller.ex
chrismccord: I made the message buffer a list of {ref, %Message{}}'s and now I only ack the refs. I don't think we can guarantee ordering so just acking the length won't work. acking the refs avoids most of the copying.

chrismccord:
@josevalim All feedback addressed. I think we're ready

lib/phoenix/pubsub/redis_server.ex
josevalim: Just one final question then: what happens when the incrementer overflows? There is likely an int limit in there?

lib/phoenix/transports/long_poller.ex
josevalim: The messages have no ordering but once they arrive in the long poller
server they are ordered in its buffer. Think about this. The server got
three messages: a, b and c. They are stored like this:

    [c, b, a]

Now the server got two more:

    [e, d, c, b, a]

At this moment, you dispatch them to the long polling connection. However,
they are still in the buffer, because there were not acked. Later on, you
got two more (and no ack yet):

    [g, f, e, d, c, b, a]

Now you got an ack. The ack contains five, the number of messages that you
can drop from the tail:

    [g, f]

It doesn't matter the order here because we are just acking the in-memory
messages. This will hold true as long as messages are sent one by one (i.e.
if pid1 hasn't ack, you cannot send messages to pid2).



lib/phoenix/pubsub/redis_server.ex
patrickdet: it's a 64-bit signed integer so it should take a long time to overflow

josevalim: Great!



patrickdet:
Congrats guys! This is a great feature

PR #595, opened 2015-01-22
Using originally wrapped connection, not error.conn

jeregrine:
I was throwing a 404 error and it was ignoring my json format. Talking to Chris I moved plug accepts to endpoint and changed this to get it to work.

@josevalim your input on this would be greatly appreciated.

### DIFF
--- a/lib/phoenix/endpoint/error_handler.ex
+++ b/lib/phoenix/endpoint/error_handler.ex
@@ -43,7 +43,7 @@ defmodule Phoenix.Endpoint.ErrorHandler do
       fun.()
     rescue
       e in [Phoenix.Router.NoRouteError] ->
-        maybe_render(e.conn, :error, e, System.stacktrace, opts)
+        maybe_render(conn, :error, e, System.stacktrace, opts)
 
     catch
       kind, reason ->


josevalim:
We shouldn't move accepts to the endpoint because it may change per
pipeline although I can see how exceptions in the app will use the wrong
accepts then. :(



-- 


chrismccord:
@josevalim We should be able to also wrap the controller dispatch in ErrorHandler, right? That way you get the accepts from the pipeline. If you need more granular wrapping, you can still override call/2 in the controller and use super. Thoughts?

jeregrine:
@chrismccord @josevalim I was being an idiot. MY API was expecting json but was getting html because the debug errors was capturing the exception and showing html. Which normally only did on 500 not 404 :(

BUT 

To get the error response to be json the accepts need to be in the endpoint so the point stands.

chrismccord:
@josevalim Are we going to revise any error wrapping for 0.9.0?

josevalim:
It is on my todo but I am not sure if I will fix it on my time. My advice is to just go ahead with the release, if I make it on time, good, if not, it will be in the next version.

josevalim:
Fixed in master.

PR #598, opened 2015-01-23
Added `Router.resource` for defining a singular resource

joshnuss:
Adds the ability to define a set of routes that wont require an index action or an ID parameter in the path.

Example:

```elixir
# in web.router.ex
resource "/session", SessionsController
```

Will generate the following routes:

    * `GET /session` => `:show`
    * `GET /session/new` => `:new`
    * `POST /session` => `:create`
    * `GET /session/edit` => `:edit`
    * `PATCH /session` => `:update`
    * `PUT /session` => `:update`
    * `DELETE /session` => `:delete`

### DIFF
--- a/test/phoenix/router/resource_test.exs
+++ b/test/phoenix/router/resource_test.exs
@@ -1,0 +1,121 @@
+defmodule Phoenix.Router.ResourceTest do
+  use ExUnit.Case, async: true
+  use RouterHelper
+
+  defmodule AccountController do
+    use Phoenix.Controller
+    plug :action
+    def show(conn, _params), do: text(conn, "show account")
+    def new(conn, _params), do: text(conn, "new account")
+    def edit(conn, _params), do: text(conn, "edit account")
+    def create(conn, _params), do: text(conn, "create account")
+    def update(conn, _params), do: text(conn, "update account")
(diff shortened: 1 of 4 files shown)


test/phoenix/router/resource_test.exs
```diff
@@ -0,0 +1,215 @@
...
+    use Phoenix.Router
+
+    resource "/account", AccountController, alias: Api do
+      resources "/comments", CommentController do
+        get "/special", CommentController, :special
+      end
+      resource "/session", SessionController, except: [:delete]
+    end
+
+    resource "/session", Api.SessionController, only: [:show]
```
josevalim: We can skip those routes and the ones below and their respective tests IMO. Just keeping it simple. :)

josevalim:
:+1: This looks great. I think we can simplify the tests a bit and we need tests for the named helpers!

joshnuss:
Groovy, will make those changes.

Do think we should have a seperate file for resource_test.exs and resources_test.exs? or should they be merged?

josevalim:
Keeping them separated is fine imo.

joshnuss:
Simplified the tests for resource (removed duplicate tests for `resources` stuff) and added tests for helpers too.

josevalim:

PR #605, opened 2015-02-01
Allow project creation omitting project directory

mdesantis:
Hello,

I believe I successfully implemented this feature, but I have some problems writing a successful test; after some refactorings it should pass, but it fails with this error:

```
> mix test test/mix/tasks/phoenix/new_test.exs
Compiled lib/mix/tasks/phoenix.new.ex
Generated phoenix.app
.

  1) test two arguments (both name and path arguments) (Mix.Tasks.Phoenix.NewTest)
     test/mix/tasks/phoenix/new_test.exs:48
     Expected .gitignore to exist, but does not
     stacktrace:
       test/mix/tasks/phoenix/new_test.exs:63: anonymous fn/0 in Mix.Tasks.Phoenix.NewTest.creates_files_and_directories/1
       (elixir) lib/file.ex:1061: File.cd!/2
       test/mix/tasks/phoenix/new_test.exs:57

.

Finished in 1.0 seconds (0.2s on load, 0.8s on tests)
3 tests, 1 failures

Randomized with seed 787097
```

### DIFF
--- a/lib/mix/tasks/phoenix.new.ex
+++ b/lib/mix/tasks/phoenix.new.ex
@@ -32,9 +32,9 @@ defmodule Mix.Tasks.Phoenix.New do
 
 
   def run(app, module, path, dev) do
-    pubsub_server      = module
-                         |> Module.concat(nil)
-                         |> Naming.base_concat(PubSub)
+    pubsub_server = module
+                    |> Module.concat(nil)
+                    |> Naming.base_concat(PubSub)
     binding = [application_name: app,
                application_module: module,
                phoenix_dep: phoenix_dep(dev),
--- a/lib/mix/tasks/phoenix.new.ex
+++ b/lib/mix/tasks/phoenix.new.ex
@@ -5,43 +5,46 @@ defmodule Mix.Tasks.Phoenix.New do
   @shortdoc "Creates Phoenix application"
 
   @moduledoc """
-  Creates a new Phoenix application
+  Creates a new Phoenix project.
+  It expects the path of the project as argument.
 
-      mix phoenix.new app_name path
+      mix phoenix.new PATH [--module MODULE] [--app APP]
+
+  A project at the given PATH  will be created. The
+  application name and module name will be retrieved
+  from the path, unless `--module` or `--app` is given.
 
   """
 
-  def run(args) do
-    {opts, args, _} = OptionParser.parse(args, strict: [dev: :boolean])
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks, if we are going down this road, i think we should make it the same as Elixir. I.e. the first argument is the path and both --app and --module can be configured:

https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/new.ex#L26-L30

Generation prelude code:

https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/new.ex#L50-L67

mdesantis:
I can adapt mix new task code for phoenix new; but I'm a bit undecided about what to port into phoenix new task; for example, `check_application_name!` and `check_mod_name!` should be ported?

mdesantis:
@josevalim I made some progresses adapting mix new task, but I have some concurrency problems I guess: the tests I wrote pass if ran by themselves, but they fail if ran together.

And I didn't add a test for `--module` switch, I'm thinking about how to add it.

josevalim:
>  for example, check_application_name! and check_mod_name! should be ported?

I think it would be a good idea to provide this sort of feedback here too!

mdesantis:
About code duplication, since it is needed to port some code from mix new
to mix Phoenix new: is it ok or mix should define some of these methods as
helpers and Phoenix should use them?

Il giorno lun 2 feb 2015 11:21 José Valim <notifications@github.com> ha
scritto:

josevalim:
It is ok to duplicate the elixir code.

evanfarrar:
Hey @mdesantis I took a look at this. The functionality works as described. The merge conflict is not very bad. Your branch fails on my laptop and travis, though. I believe this is due to the fact that there was only one test before and that test had a hidden externality (compiling produces files the first time, but does not need to produce them the second time when nothing has changed).

I'd suggest it is OK not to test that compilation occurs (or at least not to test it twice). But perhaps some committer can weigh in.

evanfarrar:
I fixed the tests in #628 and preserved you as author

mdesantis:
@evanfarrar thank you for your edits, I think they are :+1: . I have to finish merging of mix new features yet; I expect to do it within this weekend (sorry for the delay :/ ).

mdesantis:
@josevalim the pull request should be ok now; can you check it?

josevalim:

PR #612, opened 2015-02-06
Read prod secret_key_base from env instead of using generated value

jsanders:
Checking in the production value for `secret_key_base` (and its previous incarnations) has been a gotcha for rails apps that would be best to avoid. Not attached to the implementation, but I think something like this would be nice.

(Apologies if this has been discussed before, but I couldn't find anything in issues or on the mailing list.)

### DIFF
--- a/priv/template/config/prod.exs
+++ b/priv/template/config/prod.exs
@@ -3,7 +3,7 @@ use Mix.Config
 config :<%= application_name %>, <%= application_module %>.Endpoint,
   url: [host: "example.com"],
   http: [port: System.get_env("PORT")],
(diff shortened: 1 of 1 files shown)


josevalim:
This does not work with exrm out of the box so we are avoiding this until there is a solution from their side.

jsanders:
Ah! Makes sense.

jsanders:
If anyone else is interested in this, check out bitwalker/exrm#75. Seems like there isn't a perfect solution for this yet, but that you can accomplish something similar using `conform`.

jsanders:
Ah sorry, much better link for the current status of this functionality: bitwalker/exrm#90.

PR #613, opened 2015-02-06
Dynamic port : obtain the port number at runtime 

raw1z:
The intent of this pull request is to allow obtaining the port number at runtime. Two options are proposed :

1. In the config file, set the value of the port to an anonymous function responsible for providing the port value. Note that this won't work when releasing with exrm.

2. In the config file, set the value of the port to {:dynamic, var_name} where var_name is the name of the system environment variable which contains the port number. This option is fully compatible with exrm releases.

### DIFF
--- a/lib/phoenix/endpoint/server.ex
+++ b/lib/phoenix/endpoint/server.ex
@@ -38,13 +38,13 @@ defmodule Phoenix.Endpoint.Server do
       |> Keyword.put_new(:otp_app, otp_app)
       |> Keyword.put_new(:port, port)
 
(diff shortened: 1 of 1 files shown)


raw1z:
Sorry didn't notice that the file was renamed from supervisor.ex to server.ex

josevalim:
Thanks! I would like though to wait a bit on this. exrm is working on a way of allowing environment configs. If not, we can support `{:system, ...}` similar to what you have proposed.

chrismccord:
We've merged this feature in #622 with your same approach, so good thinking :)

PR #621, opened 2015-02-13
Headers in sockets

mgwidmann:
### I'd like to start out by saying how great I think Phoenix is coming along! Great job! :+1: :tada: 

I know this is a painful topic, but I think it needs to be fleshed out if Phoenix is to succeed in the long run. The client-side HMAC token system cannot be the only method of authentication. I think it is an excellent feature though! There are currently, and will be more to come, use cases in which this system fails. First of all, not running the application over SSL immediately makes one time use token a vulnerability. Anyone who sniffs the token and makes the connection to the application first would gain credential access. Secondly, putting a security token on the client side accessible by javascript is a big time security violation by itself. If somehow you do get some kind injection issue (XSS or otherwise), now user tokens are available. This is the reason why session cookies are marked HTTP only and cannot be accessed by javascript. The web is already built this way for a reason.

In order to use the token system without a security hole, Phoenix must guarantee SSL and prevention of XSS or other injection issues (and even then I'm still uneasy about the architecture). Those are two tall orders to make, so heres what I propose. **Allow access to the session and headers in websocket connections!**

#### You may not like what I've done, thats ok if you prefer to reimplement it. But PLEASE consider adding this feature. Phoenix and its awesome web socket system are pretty much useless to me (and likely many others) without it.

> In case you are wondering what my use case is... 
> I have a Phoenix app here: https://github.com/mgwidmann/elixir-rails-websockets which will proxy to a locally configured (still TODO in the phoenix app) rails app so that I may lighten the web socket load on my rails server behind a reverse proxy like nginx. The Phoenix proxy needs to send over the session cookie to the rails app so that it appears transparently that the request came from the client (the rails session cookie most importantly needs to be sent, but other headers could be useful as well). The whole purpose of this is that I have a built rails application (which currently is better suited being built in rails than Phoenix IMO) which needs web socket support on a very limited set of hardware (which I feel MRI cannot handle since I won't be able to have more than a handful of workers). I didn't want to use JRuby or Rubinius as I have already built the app in MRI and it is too risky at this point to change. I will be deploying to production soon, either with my forked version of Phoenix or master, whichever supports this feature.

### DIFF
--- a/lib/phoenix/channel/transport.ex
+++ b/lib/phoenix/channel/transport.ex
@@ -63,12 +63,13 @@ defmodule Phoenix.Channel.Transport do
 
   The returned `HashDict` of `%Phoenix.Socket{}`s must be held by the adapter
   """
-  def dispatch(msg = %Message{}, sockets, adapter_pid, router, pubsub_server, transport) do
+  def dispatch(msg = %Message{}, conn, sockets, adapter_pid, router, pubsub_server, transport) do
     socket = %Socket{pid: adapter_pid,
                      router: router,
                      pubsub_server: pubsub_server,
                      topic: msg.topic,
-                     transport: transport}
+                     transport: transport,
+                     conn: conn}
 
     sockets
     |> HashDict.get(msg.topic, socket)
--- a/lib/phoenix/socket.ex
+++ b/lib/phoenix/socket.ex
@@ -28,6 +28,7 @@ defmodule Phoenix.Socket do
             authorized: false,
             transport: nil,
             pubsub_server: nil,
+            conn: nil,
             assigns: %{}
 
 
--- a/lib/phoenix/transports/long_poller.ex
+++ b/lib/phoenix/transports/long_poller.ex
@@ -77,7 +77,7 @@ defmodule Phoenix.Transports.LongPoller do
       |> Kernel.<>(Base.encode64(:crypto.strong_rand_bytes(16)))
       |> Kernel.<>(:os.timestamp() |> Tuple.to_list |> Enum.join(""))
 
-    child = [router, timeout_window_ms(conn), priv_topic, pubsub_server(conn)]
+    child = [router, timeout_window_ms(conn), priv_topic, pubsub_server(conn), conn]
     {:ok, server_pid} = Supervisor.start_child(LongPoller.Supervisor, child)
 
     {conn, priv_topic, sign(conn, priv_topic), server_pid}
--- a/lib/phoenix/transports/long_poller/server.ex
+++ b/lib/phoenix/transports/long_poller/server.ex
@@ -32,12 +32,12 @@ defmodule Phoenix.Transports.LongPoller.Server do
   If the server receives no message within `window_ms`, it terminates and
   clients are responsible for opening a new session.
   """
-  def start_link(router, window_ms, priv_topic, pubsub_server) do
-    GenServer.start_link(__MODULE__, [router, window_ms, priv_topic, pubsub_server])
+  def start_link(router, window_ms, priv_topic, pubsub_server, conn) do
+    GenServer.start_link(__MODULE__, [router, window_ms, priv_topic, pubsub_server, conn])
   end
 
   @doc false
-  def init([router, window_ms, priv_topic, pubsub_server]) do
+  def init([router, window_ms, priv_topic, pubsub_server, conn]) do
     Process.flag(:trap_exit, true)
 
     state = %{buffer: [],
@@ -46,7 +46,8 @@ defmodule Phoenix.Transports.LongPoller.Server do
               window_ms: window_ms * 2,
               pubsub_server: pubsub_server,
               priv_topic: priv_topic,
-              client_ref: nil}
+              client_ref: nil,
+              conn: conn}
 
     :ok = Phoenix.PubSub.subscribe(state.pubsub_server, self, state.priv_topic, link: true)
     {:ok, state, state.window_ms}
@@ -62,7 +63,7 @@ defmodule Phoenix.Transports.LongPoller.Server do
   """
   def handle_info({:dispatch, message, ref}, state) do
     message
-    |> Transport.dispatch(state.sockets, self, state.router, state.pubsub_server, LongPoller)
+    |> Transport.dispatch(state.conn, state.sockets, self, state.router, state.pubsub_server, LongPoller)
     |> case do
       {:ok, sockets} ->
         :ok = broadcast_from(state, {:ok, :dispatch, ref})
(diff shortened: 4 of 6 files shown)


josevalim:
Allowing session access in websockets is usually a no-go because websockets allow cross origin requests. Basically, using session authentication with websockets means a malicious website will be able to connect to your websocket with the user session and emit commands as the attacked user.

> In order to use the token system without a security hole, Phoenix must guarantee SSL and prevention of XSS or other injection issues (and even then I'm still uneasy about the architecture). Those are two tall orders to make

Those are requirements today for any website. If your website has any kind of authentication and it does not provide both SSL and XSS protection, it will be vulnerable, from top to bottom. This is not optional.

mgwidmann:
In order for a cross origin request to access session data as you say would require the attacker to gain the session cookie which while not impossible is a generally accepted practice across the internet. Several other web socket systems allow access to the session, including faye in rails. 

All I'm stating is that Phoenix requires the user to implement their own security scheme, never a good plan. Users need to figure out a way of passing authentication information down to the client from a controller and back up to a web socket and keep it protected. Short answer, they will fail. Stick to the well established practices the internet has already created. If you plan on creating a new scheme as an experiment, make it optional and don't force it on others, if it is truly better others will realize it and begin using it.

josevalim:
> In order for a cross origin request to access session data as you say would require the attacker to gain the session cookie.

That's not true for websockets. Websockets will include session/cookie data by default without needing the attacker to hijack it in any way:

http://www.christian-schneider.net/CrossSiteWebSocketHijacking.html

josevalim:
I agree we need to improve the current mechanisms, at least by making the tokens easy to generate, as well as maybe provide easy Origin validation, but accessing the cookie/session headers definitely isn't the way to go here.

chrismccord:
> All I'm stating is that Phoenix requires the user to implement their own security scheme, never a good plan.

Agreed. We are planning to provide token generation/verification for you. In addition to the issues I've already load out about passing the conn in the socket on the ML, the other big issue is we abstract the transport layer. Most folks will just use the Router over http as their transport layer (which will have a %Plug.Conn{}), but what about a redis transport? A rabbitmq transport? All these don't have a `conn`, session, or headers, and all of these are possible today. You could build a Ruby channels client that publishes over redis and receives messages over redis instead of http. I understand the frustration since having this makes your life easier for your use-case, but can you simply encrypt the session on the rails end, and pass that up to the phoenix channel and decrypt with a shared key?

Rails side:
```ruby
def show
  @encrypted_session = Encryptor.encrypt(session, ENV["SOME_SHARED_KEY"])
  ...
end
```

```erb
<script>
   socket.join("foobar", {session:  "<%= @encrypted_session %>"}, function(chan){ ...
</script>
```

Phoenix Side:

```elixir
def join("foobar", %{"session" => encrypted_session}, socket) do
  case Encryptor.decrypt(encrypted_session, System.get_env("SOME_SHARED_KEY")) do
    {:ok, session} -> {:ok, assign(socket, :session, session)}
    _ -> :ignore
  end
end
```


Can this work for your use-case?

mgwidmann:
I'm not sure how comfortable I feel about security tokens in javascript, even if they are only accessible to the phoenix.js socket. It just feels like theres too much room for error, and I doubt I'm alone in this.

According to the article that @josevalim posted, Phoenix should be checking the origin to match itself and closing the connection otherwise. That would alleviate this problem, would it not? It just seems absolutely ridiculous to me that the **server depends on the client** to give it information the server  already has. I feel this is the real issue of the current design.

@chrismccord Your point about a different transport is a legitimate one, though in those use cases I wouldn't expect the `conn` to be available. In those cases though, we are not authenticating a user, but a machine potentially acting on behalf of a user. As such, the token system in place is perfectly acceptable since it is reasonable to assume the machine properly authenticated the user. This is not the use case when accepting data from a client machine.

josevalim:
> It just seems absolutely ridiculous to me that the server depends on the client to give it information the server already has.

That's not true. The server does not have this information. The client *always* need to pass it as part of the session, headers, an explicit token, etc. The only benefit of using a session is that it is managed for you on both client and server but with a huge downside that it is unsafe for websockets unless you add extra checks (and failing on those checks, you are vulnerable). We can, however, manage the current tokens automatically for you too and in a safer way. The code just isn't here yet.

So all in all, the only concern is that the current flow is not yet streamlined. We have already agreed with this and we are working on it. So *please wait* until we propose a solution. @chrismccord and I already gave plenty reasons for not relying on the session (security-wise) or on the connection (abstraction-wise).

Thank you.

josevalim:
> @chrismccord Your point about a different transport is a legitimate one, though in those use cases I wouldn't expect the conn to be available. In those cases though, we are not authenticating a user, but a machine potentially acting on behalf of a user. As such, the token system in place is perfectly acceptable since it is reasonable to assume the machine properly authenticated the user. This is not the use case when accepting data from a client machine.

There are other transports besides a RabbitMQ or Redis. For example, we could have a custom TCP protocol for handling Android, iPhone and generally mobile requests. There is no connection and there is still a user behind.

josevalim:
I have realized I haven't proposed any solution to your particular use case.

I would go with a simple Token Authentication mechanism (as in API Token authentication). On the Rails side, generate a token for the given user and store it in the database with the user id. On the Phoenix.Socket pass both the user_id and the token. On the Phoenix side, fetch the user_id and check if the tokens match.

This approach should be simple and safe as any other API Token authentication mechanism and should work cross applications.

mgwidmann:
@josevalim While that may work, it introduces an additional dependency (the database, where it wasn't necessary prior) and I was hoping to make it a bit more generic than that (if this works I plan to use it on more applications). Additionally, it puts additional work on the rails app (where there was no additional work needed prior). After all, its just a proxy, why would it need a database connection? It should just echo whatever is sent to it...

I still see no reason why the headers, at the least, can't be accessible if they exist. If the connection technology doesn't allow headers, then I wouldn't expect them, but if it does then I do. Additional use cases such as geo location in web sockets become problematic. This is not the only use case, there will be tons more later on. **As much information as is available should be accessible, period**.

josevalim:
If one is directly proxying the other, you can just include the token in the header when proxying. Then they are sync. I originally thought you would have nginx or apache proxying them.

mgwidmann:
I think your original though was correct.

It works like this:
WS connections: client -> nginx -> phoenix -> rails
HTTP connections: client -> nginx -> rails

josevalim:
So you can just forward all WS requests to Rails including the token validation and let Rails decide. You don't need the database, just generate and enforce the secret on Rails side. Search the message verifier here for examples: http://guides.rubyonrails.org/4_1_release_notes.html

mgwidmann:
> That's not true. The server does not have this information. The client always need to pass it as part of the session, headers, an explicit token, etc.

Yes the browser always needs to pass it, but as I proved by my PR, the server does have this information because the browser does always pass it.

I'm not sure I agree, but its your decision. I don't believe hiding the information as `master` does currently is going to be appealing to new users. I want to pitch using Phoenix for some other things at my workplace, but with this issue I don't feel they'll go for it.

I'm going to keep my fork, sorry.

josevalim:
Feel free to roll your own, just remember sessions are not safe without extra checks. We do plan to provide a streamlined workflow but this PR is not the direction we want to go. Hopefully you can drop your fork once we get there.

mgwidmann:
Will do, thanks.

PR #622, opened 2015-02-13
Set port to PORT env var if it isn't set in config

guilleiguaran:
This set port to PORT env var if it isn't set in config


josevalim:
Thank you! @guilleiguaran I think we should go with a specific value like `{:system, "PORT"}`. We would also need to change the endpoint/adapter.ex code to read from the environment too.

guilleiguaran:
@josevalim so, If I understand fine the user should then define in the config:

```elixir
http: [port: {:system, "PORT"}]
```

Sounds good for me :+1:

guilleiguaran:
@josevalim updated and now tests are green

chrismccord:
Very nice. Thank you!

josevalim:

Issue #624, opened 2015-02-15
Do not run redis tests if redis is not running


josevalim:
We can tag redis test case with `@moduletag :redis` and invoke `ExUnit.start exclude: [:redis]` if redis is not running.

PR #625, opened 2015-02-15
Do not run redis tests if redis is not running

guilleiguaran:
For #624

### DIFF
--- a/test/phoenix/pubsub/pubsub_test.exs
+++ b/test/phoenix/pubsub/pubsub_test.exs
@@ -4,8 +4,8 @@ defmodule Phoenix.PubSub.PubSubTest do
   alias Phoenix.PubSub
 
   @adapters [
-    {Phoenix.PubSub.Redis, :redis_pub},
-    {Phoenix.PubSub.PG2,   :pg2_pub}
+    {Phoenix.PubSub.Redis, :redis_pub, :redis},
+    {Phoenix.PubSub.PG2,   :pg2_pub, :pg2}
   ]
 
   def spawn_pid do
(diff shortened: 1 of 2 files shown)


guilleiguaran:
Currently I'm getting this output when running the tests:

```
Excluding tags: [:redis]

...............................................................................................................................................................................................................................................................................
13:29:15.628 [error] unable to establish redis connection. Attempting to reconnect...
.............

Finished in 9.8 seconds (4.9s on load, 4.9s on tests)
292 tests, 0 failures, 8 skipped
```

Not sure where it is attemping to reconnect to redis thought.

test/phoenix/pubsub/pubsub_test.exs
```diff
@@ -25,6 +25,7 @@ defmodule Phoenix.PubSub.PubSubTest do
       :ok
     end
 
+    @tag tag
```
josevalim: I think you can set @moduletag at the top, just once, and it will work too. :D

test/test_helper.exs
```diff
@@ -21,4 +21,13 @@ defmodule Phoenix.ErrorView do
   end
 end
 
-ExUnit.start()
+:erlang.process_flag(:trap_exit, true)
```
josevalim: Do we need to trap exists? Can we simply start and then stop? :)

guilleiguaran: Not sure, when I try to connect using `eredis:start_link` or `eredis_client:start_link` without trapping exits I got:

```
** (EXIT from #PID<0.147.0>) {:connection_error, {:connection_error, :econnrefused}}
```

test/phoenix/pubsub/pubsub_test.exs
guilleiguaran: Updated :)

josevalim:

josevalim:
Sorry, I gave bad advice, @moduletag does not work in this case. :P

josevalim:
I have fixed it locally.

PR #627, opened 2015-02-18
Live-reload over channels via asset change detection

chrismccord:
relevent diff for discussion here: https://github.com/phoenixframework/phoenix/commit/7bbe9c349a9b35ce7cc81730b79b1e82868367bf

@josevalim and I have been exploring different asset build tools, and my latest experiments have been around webpack. It has worked well, but the "hot reload" of assets feature only works with assets served via a separate node process. This makes node a hard dependency if we want out-of-the-box live-reload (and it's the same for other node build tools). This made me sad, so I decided to spike out a home-grown phoenix solution. ~100 LOC, making use of channels, gives us simple page-reload when assets change and this could be used regardless of the users build tool choice. In the case of webpack, we'd simply use their regular build command and `watch: ["priv/static/js/app.js"]` in our mix config to have the best of both worlds. Thoughts? I think given the small footprint, this is a huge boost to productivity and we can do it without any node dependency. If you use this branch, you can `mix phoenix.new` and see how it works for yourself by editing app.css or app.js and seeing the page reload. Feedback appreciated as always. Cheers!


josevalim:
I think this is the way to go if we decide to go with webpack. Is it worth making a webpack branch? Also, if webpack always emits a single .js file, how is debugging in development?

lud:
Thanks to the source maps, debuging and errors shows the line & filename of the original source file.

josevalim:
@lud thanks. Does it work even for the css stuff embedded in the javascript file?!

lud:
Yes it seems so though I did not try.

chrismccord:
> I think this is the way to go if we decide to go with webpack. Is it worth making a webpack branch? Also, if webpack always emits a single .js file, how is debugging in development?

Sorry this branch is a little confusing. Just to be clear, we don't need a webpack branch. I spiked out this feature after a fleeting idea on this `cm-webpack` branch, but it's generic and will work for any build tool, and will also work standalone like a new generated app on static files. If we decide to use this, I will break out the last commit on its own.

sorentwo:
Quick tertiary answer regarding the inline CSS source maps. If you specify sourceMaps in the sass-loader they will be generated, but they aren't inlined.

josevalim:
I am closing this one (web pack related).

PR #631, opened 2015-02-20
Scrubbing Params

hahuang65:
As discussed with @josevalim, we want a plug that will scrub incoming parameters and do two things:
1) Check if a specified parameter is present (can be empty), if not, raise Phoenix.MissingParamError
2) Change all the empty "" values to nil.

We also discussed the possibility of a require_param plug, which would do something similar, but instead of allowing an empty value on the specified parameter, it would raise.

require_param would be done at a later date if at all. @josevalim mentioned he wanted to explore something other things first.


lib/phoenix/controller.ex
```diff
@@ -462,6 +462,37 @@ defmodule Phoenix.Controller do
...
+    unless Map.get(conn.params, required_key) do
+      raise Phoenix.MissingParamError, key: required_key
+    end
+
+    params = Enum.reduce(conn.params, %{}, fn({k, v}, acc) ->
+      case v do
+        "" -> Map.put(acc, k, nil)
+        _ -> Map.put(acc, k, v)
+      end
+    end)
```
chrismccord: This won't work for nested params. What's the plan for that?

hahuang65: We can probably easily make this recursive to handle nested params. I think the hard part is finding a good way to represent a nested expression when calling the plug.

lib/phoenix/controller.ex
```diff
@@ -462,6 +462,37 @@ defmodule Phoenix.Controller do
...
+    `raise Phoenix.MissingParamError, key: required_key`
+
+    ## Examples
+
+        iex> scrub_params(conn, "user")
+
+  """
+  @spec scrub_params(Plug.Conn.t, [String.t]) :: Plug.Conn.t
+  def scrub_params(conn, required_key) do
+    unless Map.get(conn.params, required_key) do
```
josevalim: We should traverse the parameter returned from the key and leave everything else intact.

lib/phoenix/controller.ex
josevalim: @hahuang65 Chris meant if you have a map inside the map, something like `%{"user" => %{"account" => ...}}`. We need to handle those cases too. The easiest way is to call `do_scrub_param(param)` that checks if it is a list or a map and act accordingly.

lib/phoenix/exceptions.ex
```diff
@@ -13,3 +13,24 @@ defmodule Phoenix.NotAcceptableError do
...
+    * Changes all empty parameters to nils ("" -> nil).
+
+  If you are seeing this error, you should handle the error and surface it
+  to the end user that there is a parameter missing from the request.
+  """
+
+  defexception [:message, plug_status: 400]
+
+  def exception([key: value]) do
+    msg = "Expected key for #{inspect value} to be present."
```
josevalim: Error messages typically start in lowercase and have no dots. :)

lib/phoenix/controller.ex
hahuang65: I misunderstood. I thought he was talking about the case for checking user[account] in this case. To make sure it's present as opposed to just user.

mix.lock
```diff
@@ -5,7 +5,7 @@
   "ex_doc": {:hex, :ex_doc, "0.7.1"},
   "inch_ex": {:hex, :inch_ex, "0.2.3"},
   "linguist": {:hex, :linguist, "0.1.4"},
-  "plug": {:hex, :plug, "0.10.0"},
+  "plug": {:git, "git://github.com/elixir-lang/plug.git", "39c8f5f827ef8bdd41a650fe9b5748f334504d98", []},
```
hahuang65: Ummm oops. This was caused by merging master. Should I take it out?

josevalim: It is fine. I will handle it when merging.

josevalim:
merged manually, thank you

jeregrine:
:+1: nice work! 

Pretty sure about this but I want to confirm, if you have a url param `/user/:id` do those params get into conn.params? I know it gets into the controller action params.

josevalim:
It gets into params but you may not want to scrub those. The focus of scrub is on form related stuff. :heart:

jeregrine:
Makes sense!

PR #633, opened 2015-02-22
Add asset handling via brunch with live-reload

chrismccord:
This PR contains asset handling with [brunch](http://brunch.io/). @josevalim and I have experimented with just about all the node build tools, and brunch has been the simplest to set up and integrated with phoenix. With this PR, new projects get js bundling witih es6 support, and Sass css compilation. We also provide an automatic live-reload feature that refreshes the browser with any assets changes, as well as any phoenix templates.

Screencast of how it works for the lazy: 	https://www.youtube.com/watch?v=CgSjuCRHKvg&feature=youtu.be

A few notes:

1) Node is *not* a hard dependency
2) Nothing is hardcoded for brunch. We've integrated it in a way that will allow you to bring any asset build tool you like and have it plug right in.
3) It Just Works^tm

Now, a few questions for feedback:

1) We are currently running `npm install` automatically when running `mix phoenix.new ...`. Is this preferable? Should we require the manual step for the end user? Should we prompt y/n?

2) We are polling the file system for changes, with these default paths to watch in the endpoint config:

```elixir
  assets: [
    watchers: [{Path.expand("node_modules/brunch/bin/brunch"), ["watch"]}],
    live_reload: ["priv/static/app.js",
                  "priv/static/app.css",
                  Path.wildcard("web/templates/**/*")]
  ]
```
Polling all template ctimes is going to be expensive for large projects, so a fs-event alternative would be preferable, but I haven't yet looked into third party deps for this. Any decent alternatives?

Thoughts?


josevalim:

lib/mix/tasks/phoenix.new.ex
```diff
@@ -33,6 +34,12 @@ defmodule Mix.Tasks.Phoenix.New do
 
     copy_from template_dir, path, application_name, &EEx.eval_file(&1, binding)
     copy_from static_dir, Path.join(path, "priv/static"), application_name, &File.read!(&1)
```
josevalim: We should not copy anything to `priv/static` anymore and we should rather .gitignore it.

lib/mix/tasks/phoenix.new.ex
```diff
@@ -33,6 +34,12 @@ defmodule Mix.Tasks.Phoenix.New do
...
+
+    # TODO decide to auto npm install or not
+    if npm_path && Mix.env == :dev do
+      IO.puts "Installing brunch.io dependencies..."
+      System.cmd("npm", ["install", "--prefix", path])
```
josevalim: Should we output all the command to stdout in case there are failures using :into?

lib/mix/tasks/phoenix.new.ex
```diff
@@ -33,6 +34,12 @@ defmodule Mix.Tasks.Phoenix.New do
 
     copy_from template_dir, path, application_name, &EEx.eval_file(&1, binding)
     copy_from static_dir, Path.join(path, "priv/static"), application_name, &File.read!(&1)
+
+    # TODO decide to auto npm install or not
+    if npm_path && Mix.env == :dev do
```
josevalim: I am OK with installing it. I think though that, 

1. Should we provide an option to disable npm install?

2. Should we provide an option to disable brunch?

3. If NPM is not available, I think we should emit a message to the user: "You don't have npm installed. This means we were not able to install all packages... yada yada... install npm and run npm instal inside the app"

lib/phoenix/channel/control_channel.ex
```diff
@@ -0,0 +1,16 @@
...
+  @moduledoc """
+  Phoenix's control channel for framework specific messages
+  """
+
+  def join("phoenix", _msg, socket) do
+    {:ok, socket}
+  end
+
+  def handle_out("assets:change", _message, socket) do
+    reply socket, "assets:change", %{}
```
josevalim: Should we prefix this with `phx`?

lib/phoenix/code_reloader.ex
```diff
@@ -187,4 +187,29 @@ defmodule Phoenix.CodeReloader do
 
   defp method(%Plug.Conn{method: method}), do:
     method
+
+  defp before_send_inject_reloader(conn) do
```
josevalim: It is OK for this to be here for now but we should seriously consider this to be another plug. :)

lib/phoenix/endpoint/adapter.ex
```diff
@@ -9,7 +9,10 @@ defmodule Phoenix.Endpoint.Adapter do
   """
   def start_link(otp_app, mod) do
     import Supervisor.Spec
-    pub_conf = config(otp_app, mod)[:pubsub]
+    conf       = config(otp_app, mod)
+    pub_conf   = conf[:pubsub]
+    asset_conf = conf[:assets]
```
josevalim: It is weird to call this assets because we don't call them assets anywhere (remember the brunch discussion?). Maybe it should be static?

lib/mix/tasks/phoenix.new.ex
chrismccord: We still need to copy the "built" app.js boilerplate because it bundles phoenix.js, which live-reload needs. Alternatively, we can have the injected scripts point to a different priv/static/js/phoenix.js file, but then this breaks our vendor idioms, so I chose to build the bundle into phoenix's priv/static and still copy over.

lib/phoenix/endpoint/adapter.ex
```diff
@@ -17,14 +20,33 @@ defmodule Phoenix.Endpoint.Adapter do
         [supervisor(adapter, [mod.__pubsub_server__(), pub_conf[:options] || []])]
     end
 
-    children = pubsub_children ++ [
+    asset_children =
+      Enum.reduce(asset_conf[:watchers], asset_children, fn {cmd, args}, acc ->
+        acc ++ [worker(Task, [fn ->
```
josevalim: Let's create a Phoenix.Endpoint.Watcher instead of inlining the logic here.

lib/phoenix/endpoint/adapter.ex
```diff
@@ -60,7 +82,13 @@ defmodule Phoenix.Endpoint.Adapter do
      secret_key_base: nil,
      server: Application.get_env(:phoenix, :serve_endpoints, false),
      url: [host: "localhost"],
-     pubsub: []]
+     pubsub: [],
+
+     # Assets
+     assets: [
+       watchers: [],
```
josevalim: I don't think watchers are specific to assets. It happens that our default watcher is an asset related thing but they are completely distinct. To me it would be another config on root.

priv/template/lib/application_name/endpoint.ex
```diff
@@ -4,7 +4,8 @@ defmodule <%= application_module %>.Endpoint do
   # Serve at "/" the given assets from "priv/static" directory
   plug Plug.Static,
     at: "/", from: :<%= application_name %>,
-    only: ~w(css images js favicon.ico robots.txt)
+    only: ~w(css images js favicon.ico robots.txt
+             app.js app.js.map app.css app.css.map)
```
josevalim: Ah, this would be a reason to not compile stuff to root. Let's change the brunch file?

priv/template/mix.exs
```diff
@@ -22,7 +22,7 @@ defmodule <%= application_module %>.Mixfile do
   #
   # Type `mix help deps` for examples and options
   defp deps do
-    [<%= phoenix_dep %>,
+    [{:phoenix, github: "phoenixframework/phoenix", branch: "cm-brunch"},
```
josevalim: Remember to roll this change back. :D

priv/template/web/static/js/app.js
```diff
@@ -0,0 +1,5 @@
+import {Socket} from "./phoenix"
+
+// let socket = new Socket("/ws")
```
josevalim: Cool example!

lib/mix/tasks/phoenix.new.ex
chrismccord: > Should we provide an option to disable npm install?

Probably. Otherwise, if you're just building an API and happen to have node installed, you need to wait for `npm install` to finish, then rm -rf node_modules.

> Should we provide an option to disable brunch?

I don't think we need it, because you can just set, `watchers: []` in your `dev.exs` to disable brunch from running. You will have to delete brunch-config and package.json though manually. wdyt?

> If NPM is not available, I think we should emit a message to the user: "You don't have npm installed. This means we were not able to install all packages... yada yada... install npm and run npm instal inside the app"

+1

priv/template/brunch-config.js
```diff
@@ -0,0 +1,31 @@
+exports.config = {
+  // See http://brunch.io/#documentation for docs.
```
josevalim: We should move this line up and add a couple comments about what is brunch and so on since we are making this decision for the user.

priv/template/web/static/js/phoenix.js
```diff
@@ -0,0 +1,555 @@
+"use strict";
```
josevalim: I thought we should rather import from phoenix in deps? I don't see why we would embed the phoenix version directly here (even more the non ES6 ones).

lib/mix/tasks/phoenix.new.ex
chrismccord: yes, def

lib/phoenix/channel/control_channel.ex
chrismccord: We are using the topic "phoenix", so it's all under the "phoenix" namespace

priv/static/app.js
```diff
@@ -0,0 +1,668 @@
+(function(/*! Brunch !*/) {
```
josevalim: This file should likely be priv/static/phoenix.js. We need to change the brunch-config in Phoenix itself.

lib/mix/tasks/phoenix.new.ex
josevalim: Well, you do want to gitignore priv static when using a build tool, so we need to lookup for alternatives. We will also need to move the phoenix logo to web/static/assets.

PR #641, opened 2015-02-27
Polish the docs in Phoenix.Endpoint

whatyouhide:
Mainly markup, some very slight rephrasings here and there. Let me know if I have to change something or if I can do anything else!


jeregrine:
:+1:

PR #643, opened 2015-02-28
Improve content-type negotiation

oneeman:
Do not fall back to html due to `*/*` if it has a `q` value other than 1

This bug was pointed out in the IRC channel by @tmjoen.  jQuery includes `*/*; q=0.01` on its headers, which was causing Phoenix to fall back to html in spite of the q value.  This PR adds a failing test and a fix.

For reference, see:

http://bugs.jquery.com/ticket/12250 
and
https://github.com/jquery/jquery/blob/master/src/ajax.js#L594-L601

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -636,10 +636,24 @@ defmodule Phoenix.Controller do
   # But before we check for */* because if one exists and we serve html,
   # we unfortunately need to assume it is a browser sending us a request.
   defp handle_header_accept(conn, [header|_], accepted) do
-    if header =~ "*/*" and "html" in accepted do
+    types = String.split(header, ",")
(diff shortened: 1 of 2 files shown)


josevalim:
I am :-1:. Browsers send `*/*` with `q` < 1 and we should still default to HTML there. Basically, we cannot trust any accept header with `*/*` in it because browsers have historically messed this up. The fix is to tell jQuery to send exactly the content-type you expect.

josevalim:
http://www.newmediacampaigns.com/blog/browser-rest-http-accept-headers

oneeman:
Oh, dear!  Thanks for chiming in, I wasn't aware and didn't research it enough.

josevalim:
No problem, thanks for the PR!

chrismccord:
As a follow up, you can simply add the `format` param to your jquery request, `?format=json`

Issue #583, opened 2015-01-10
Phoenix does not generate proper files inside umbrella


josevalim:
Inside an umbrella, the mix.exs file has to be slightly different. Here is the source in mix:

https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/new.ex#L243-L244

And here is the code that verifies if we are inside an umbrella:

https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/new.ex#L175-L186

PR #649, opened 2015-03-03
Generate proper files under umbrella

mdesantis:
This should fix issue #583.

The test passes when I run it alone, but when ran toghether with the other tests sometimes fails. Is it some cache issue?

### DIFF
--- a/lib/mix/tasks/phoenix.new.ex
+++ b/lib/mix/tasks/phoenix.new.ex
@@ -56,7 +56,8 @@ defmodule Mix.Tasks.Phoenix.New do
                pubsub_server: pubsub_server,
                secret_key_base: random_string(64),
                encryption_salt: random_string(8),
-               signing_salt: random_string(8)]
+               signing_salt: random_string(8),
+               in_umbrella: in_umbrella?(path)]
 
     copy_from template_dir, path, app, &EEx.eval_file(&1, binding)
     copy_from static_dir, Path.join(path, "priv/static"), app, &File.read!(&1)
@@ -106,6 +107,19 @@ defmodule Mix.Tasks.Phoenix.New do
   defp check_mod_name!(name) do
     unless name =~ ~r/^[A-Z]\w*(\.[A-Z]\w*)*$/ do
       Mix.raise "Module name must be a valid Elixir alias (for example: Foo.Bar), got: #{inspect name}"
+    end
+  end
+
+  defp in_umbrella?(app_path) do
+    apps = Path.expand(Path.join [app_path, ".."])
+
+    try do
+      Mix.Project.in_project(:umbrella_check, Path.expand(Path.join [apps, ".."]), fn _ ->
+        path = Mix.Project.config[:apps_path]
+        path && Path.expand(path) == apps
+      end)
+    catch
+      _, _ -> false
     end
   end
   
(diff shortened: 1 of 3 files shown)


test/fixtures/umbrella/mix.exs
```diff
@@ -0,0 +1,12 @@
+defmodule Umbrella.Mixfile do
```
josevalim: I think it is easier to just create this in the test using `File.write!` instead of duplicating the whole in_fixture machine. Wdyt?

josevalim:
I have added one tiny comment, thanks!

mdesantis:
Yessss ;) should I answer here or there? I don't know if you receive notifications there... anyway, I copied Mix code because it implements the same feature, and so paradoxically it was easier for me using that code which is already working than writing slightly different code; do you think that code is too much overengineered for this context?

josevalim:
>  do you think that code is too much overengineered for this context?

Precisely. :) There we need that code for multiple tests, in here, it is just one test, so it is not worth it. :)

josevalim:
(I receive notifications for both).

mdesantis:
I agree it is overengineered for this context too. I was thinking though that keeping the code aligned with the Mix code could be an advantage, because they implement the same functionality, so for example the refactoring or the bug fixing of a piece of code can be ported to the other part and viceversa; what do you think about it?

josevalim:
I really don't think we need to worry about it. We are still keeping the test, we are just generating the parent mix.exs file directly without of using the whole fixtures machinery. We can add complexity later if we need it. :)

mdesantis:
Seems legit :+1:

mdesantis:
Ok, check it now :) there's still that issue of interfering with other tests; for example:

```
> mix test test/mix/tasks/phoenix.new_test.exs --seed 11659
..

  1) test new inside umbrella (Mix.Tasks.Phoenix.NewTest)
     test/mix/tasks/phoenix.new_test.exs:115
     Assertion with =~ failed
     code: file =~ "deps_path: \"../../deps\""
     lhs:  "defmodule PhotoBlog.Mixfile do\n  use Mix.Project\n\n  def project do\n    [app: :photo_blog,\n     version: \"0.0.1\",\n\n     elixir: \"~> 1.0\",\n     elixirc_paths: [\"lib\", \"web\"],\n     compilers: [:phoenix] ++ Mix.compilers,\n     deps: deps]\n  end\n\n  # Configuration for the OTP application\n  #\n  # Type `mix help compile.app` for more information\n  def application do\n    [mod: {PhotoBlog, []},\n     applications: [:phoenix, :cowboy, :logger]]\n  end\n\n  # Specifies your project dependencies\n  #\n  # Type `mix help deps` for examples and options\n  defp deps do\n    [{:phoenix, github: \"phoenixframework/phoenix\"},\n     {:cowboy, \"~> 1.0\"}]\n  end\nend\n"
     rhs:  "deps_path: \"../../deps\""
     stacktrace:
       test/mix/tasks/phoenix.new_test.exs:123: anonymous fn/1 in Mix.Tasks.Phoenix.NewTest.test new inside umbrella/1
       (elixir) lib/file.ex:1061: File.cd!/2
       test/mix/tasks/phoenix.new_test.exs:116

..

Finished in 0.6 seconds (0.2s on load, 0.4s on tests)
5 tests, 1 failures

Randomized with seed 11659
```

josevalim:
Thank you! I have fixed the test issues on master!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

mdesantis:
Wonderful! \o/

PR #654, opened 2015-03-03
Ignore the /rel directory in generated projects

sorentwo:
Using exrm/relx is the preferred and suggested method for doing releases. Compiled output is stored in `/rel`, and there isn't any reason to check it in.

### DIFF
--- a/priv/template/.gitignore
+++ b/priv/template/.gitignore
@@ -1,5 +1,6 @@
 /_build
 /deps
-/rel
+rel/*
+!rel/relx.config
 erl_crash.dump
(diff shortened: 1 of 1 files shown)


josevalim:
It would vote to leave it out and unless Elixir or Phoenix ships with
releases. :)

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Lead Developer

sorentwo:
That makes sense. I'll try documenting it instead for the phoenix deployment guide.

PR #655, opened 2015-03-05
Brunch integration with phoenix.new options

chrismccord:


chrismccord:
So the file handling for the brunch/no brunch cases is kind of a mess and I could use some feedback on making it cleaner. There's overlap on what files need to go where `priv/static` or `web/static` and either case requires copying or removing directories from the template. The new options are the following:

    # Do not include any brunch related files (package.json, brunch-config.js, web/static/)
    mix phoenix.new /path/to --skip-brunch
    # include brunch, but not automatically run `npm install` in project
    mix phoenix.new /path/to --skip-npm

chrismccord:
I also just realized, we need to conditionally generate the template/config/dev.exs to exclude `watches: []`

lib/mix/tasks/phoenix.new.ex
```diff
@@ -42,30 +51,73 @@ defmodule Mix.Tasks.Phoenix.New do
         mod = opts[:module] || Naming.camelize(app)
         check_mod_name!(mod)
 
-        run(app, mod, path, opts[:dev])
+        run(app, mod, path, opts[:dev], opts[:skip_brunch], opts[:skip_npm])
```
josevalim: At this point we should just pass all `opts`? Don't worry though, I can refactor it when working on the model generator.

lib/phoenix/endpoint/adapter.ex
```diff
@@ -4,26 +4,45 @@ defmodule Phoenix.Endpoint.Adapter do
...
+    pub_conf   = conf[:pubsub]
 
     pubsub_children = case pub_conf[:adapter] do
       nil     -> []
       adapter -> [supervisor(adapter, [mod.__pubsub_server__(), pub_conf])]
     end
 
-    children = pubsub_children ++ [
+    watcher_children =
+      Enum.reduce(conf[:watchers], [], fn {cmd, args}, acc ->
```
josevalim: Isn't this a map?

josevalim: I mean, `Enum.map/2`.

lib/phoenix/endpoint/adapter.ex
```diff
@@ -4,26 +4,45 @@ defmodule Phoenix.Endpoint.Adapter do
...
+      paths -> [worker(ChangeDetector, [paths, {__MODULE__, :assets_change, [mod]}])]
+    end
+
+    children = watcher_children ++ live_reload_children ++ pubsub_children ++ [
       worker(Phoenix.Config, [otp_app, mod, defaults(otp_app, mod)]),
-      supervisor(Phoenix.Endpoint.Server, [otp_app, mod])
+      supervisor(Phoenix.Endpoint.Server, [otp_app, mod]),
     ]
 
     Supervisor.start_link(children, strategy: :rest_for_one, name: mod)
```
josevalim: It needs to definitely be `:one_for_one`.

priv/static/phoenix.js.map
```diff
@@ -0,0 +1 @@
...\n        case 0:\n        case 500:\n          this.onerror()\n          this.closeAndRetry()\n          break\n        default: throw(`unhandled poll status ${status}`)\n      }\n    })\n  }\n\n  send(body){\n    Ajax.request(\"POST\", this.endpointURL(), \"application/json\", body, this.onerror.bind(this, \"timeout\"), (status, resp) => {\n      if(status !== 200){ this.onerror(status) }\n    })\n  }\n\n  close(code, reason){\n    this.readyState = this.states.closed\n    this.onclose()\n  }\n}\n\n\nexport var Ajax = {\n\n  states: {complete: 4},\n\n  request: function(method, endPoint, accept, body, ontimeout, callback){\n    let req = XMLHttpRequest ? new XMLHttpRequest() : // IE7+, Firefox, Chrome, Opera, Safari\n                               new ActiveXObject(\"Microsoft.XMLHTTP\") // IE6, IE5\n    req.open(method, endPoint, true)\n    req.setRequestHeader(\"Content-type\", accept)\n    req.onerror = function(){ callback && callback(500, null) }\n    req.onreadystatechange = () => {\n      if(req.readyState === this.states.complete && callback){ callback(req.status, req.responseText) }\n    }\n    if(ontimeout){ req.ontimeout = ontimeout }\n\n    req.send(body)\n  }\n}\n"]}
```
josevalim: Do we need this .map file? Shouldn't the compiler phoenix.js already be minified? I.e. compiled in production mode?

priv/template/config/dev.exs
```diff
@@ -3,7 +3,12 @@ use Mix.Config
 config :<%= application_name %>, <%= application_module %>.Endpoint,
   http: [port: System.get_env("PORT") || 4000],
   debug_errors: true,
-  cache_static_lookup: false
+  cache_static_lookup: false,
+  watchers: [{Path.expand("node_modules/brunch/bin/brunch"), ["watch"]}],
+  live_reload: ["priv/static/app.js",
+                "priv/static/app.css",
+                Path.wildcard("web/templates/**/*")]
```
josevalim: I thought we were not going to ship with template watching by default?

priv/template/lib/application_name/endpoint.ex
```diff
@@ -4,7 +4,8 @@ defmodule <%= application_module %>.Endpoint do
   # Serve at "/" the given assets from "priv/static" directory
   plug Plug.Static,
     at: "/", from: :<%= application_name %>,
-    only: ~w(css images js favicon.ico robots.txt)
+    only: ~w(css images js favicon.ico robots.txt
```
josevalim: I though we agreed to compile everything into `css` and `js` so we don't need the extra line below?

lib/mix/tasks/phoenix.new.ex
chrismccord: :+1: agree

lib/phoenix/endpoint/adapter.ex
chrismccord: Yes. Initially I was joining the live reload + watchers as one list, so this can just be a map now

josevalim:
I think the generator could be cleaner if you remove all static files out from `priv/template` and keep them in `priv/static`. Then create two maps of how you want them to be copied:

```elixir
@brunch %{
  "phoenix.js" => "vendor/phoenix.js",
  "phoenix.png" => "assets/images/phoenix.png"
  ...
}

@static %{
  "phoenix.js" => "..."
}
```

Now you run copy_from or something similar passing the map:

```elixir
# With brunch
copy_from "priv/static", "web/static", @brunch

# Without
copy_from "priv/static", "priv/static", @static
```

Otherwise today we are copying files and then removing them and it ends up being very confusing. Also, because there is no map in the current implementation and the code is scattared around a couple places, it is very hard for me to look at the code and understand how the target directory is going to look like, be it for brunch, or for static.

priv/template/lib/application_name/endpoint.ex
chrismccord: Ah yes. Updated!

chrismccord:
@josevalim Thanks. That sounds like a really nice solution. I'll make the changes this evening

priv/static/phoenix.js.map
chrismccord: We don't need the map, and yes should be min'd

priv/template/config/dev.exs
chrismccord: I think we should include template watching by default, but add a comment along the lines of:

```
  # Watch assets and templates for live-reload changes
  # Note: larger projects will use higher CPU in dev as the number of files grows
```

priv/static/phoenix.js.map
josevalim: :shipit:

chrismccord:
@josevalim I made the discussed changes and it's waaaay nicer. Thanks! Give it a look and I think we're ready!

scrogson:
Looks really good @chrismccord! :+1:

PR #662, opened 2015-03-06
Fix syntax error in config/dev.exs

scrogson:
Fixes a failure on master when generating a new application.


josevalim:
Thanks. I will fix the broken build next.

PR #664, opened 2015-03-07
handle the case when the response has no content type

briksoftware:
if the response has no content type, get_resp_header returns an empty list which hd does not like. This PR handles this situation.


josevalim:

PR #686, opened 2015-03-10
More precise type spec for Form struct

asonge:
This trivial change should make it slightly easier to make a proper form struct at a glance.


josevalim:

PR #688, opened 2015-03-10
Add `button/2` to HTML link helpers

napcs:
Add support for `button` link helper to create navigation buttons to
destructive actions. This provides an alternative to link/2 which
can have accessibility issues as it requires JavaScript. Also prevents
web spidering software from following links to potentially destructive
actions.

### DIFF
--- a/lib/phoenix/html/link.ex
+++ b/lib/phoenix/html/link.ex
@@ -46,4 +46,40 @@ defmodule Phoenix.HTML.Link do
       end
     end
   end
+
+  @doc """
+  Generates a button that uses a regular HTML form to submit to the given URL.
+  Useful to ensure that links that change data are not triggered by
+  search engines and other spidering software.
+
+  ## Examples
+
+      <%= button("hello", to: "/world") %>
+
+  generates
+
+      <form action="/world" class="button" method="post">
+        <input name="_csrf_token" value=""><input type="submit" value="hello">
+      </form>
+
+  ## Options
+
+    * `:to` - the page to link to. This option is required
+
+    * `:method` - the method to use with the link. Defaults to :post.
+
+  """
+  def button(text, opts) do
+    {to, opts} = Keyword.pop(opts, :to)
+    {method, opts} = Keyword.pop(opts, :method, :post)
+    {htmlClass, opts} = Keyword.pop(opts, :class, :button)
+
+    unless to do
+      raise ArgumentError, "option :to is required in button/2"
+    end
+
+    form_tag(to, [method: method, class: htmlClass, enforce_utf8: false] ++ opts) do
+      tag(:input, [type: "submit", value: text])
+    end
+  end
 end
--- a/lib/phoenix/html/link.ex
+++ b/lib/phoenix/html/link.ex
@@ -85,13 +85,13 @@ defmodule Phoenix.HTML.Link do
   """
   def button(text, opts) do
     {to, opts} = Keyword.pop(opts, :to)
-    {htmlClass, opts} = Keyword.pop(opts, :class, :button)
+    {html_class, opts} = Keyword.pop(opts, :class, :button)
 
     unless to do
       raise ArgumentError, "option :to is required in button/2"
     end
 
-    form_tag(to, [class: htmlClass, enforce_utf8: false] ++ opts) do
+    form_tag(to, [class: html_class, enforce_utf8: false] ++ opts) do
       tag(:input, [type: "submit", value: text])
     end
   end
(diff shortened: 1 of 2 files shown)


paulcsmith:
This will be super helpful! Thanks @napcs

lib/phoenix/html/link.ex
```diff
@@ -46,4 +46,40 @@ defmodule Phoenix.HTML.Link do
...
+
+    * `:to` - the page to link to. This option is required
+
+    * `:method` - the method to use with the link. Defaults to :post.
+
+  """
+  def button(text, opts) do
+    {to, opts} = Keyword.pop(opts, :to)
+    {method, opts} = Keyword.pop(opts, :method, :post)
+    {htmlClass, opts} = Keyword.pop(opts, :class, :button)
```
josevalim: I think we can get rid of this line altogether?

lib/phoenix/html/link.ex
```diff
@@ -46,4 +46,40 @@ defmodule Phoenix.HTML.Link do
       end
     end
   end
+
+  @doc """
+  Generates a button that uses a regular HTML form to submit to the given URL.
+  Useful to ensure that links that change data are not triggered by
```
josevalim: Enter a new line here please so they are considered different paragraphs. ExDoc uses the first paragraph as a summary. :)

lib/phoenix/html/link.ex
```diff
@@ -46,4 +46,40 @@ defmodule Phoenix.HTML.Link do
...
+  ## Options
+
+    * `:to` - the page to link to. This option is required
+
+    * `:method` - the method to use with the link. Defaults to :post.
+
+  """
+  def button(text, opts) do
+    {to, opts} = Keyword.pop(opts, :to)
+    {method, opts} = Keyword.pop(opts, :method, :post)
```
josevalim: I think you can get rid of this line too.

lib/phoenix/html/link.ex
```diff
@@ -46,4 +46,40 @@ defmodule Phoenix.HTML.Link do
...
+      <form action="/world" class="button" method="post">
+        <input name="_csrf_token" value=""><input type="submit" value="hello">
+      </form>
+
+  ## Options
+
+    * `:to` - the page to link to. This option is required
+
+    * `:method` - the method to use with the link. Defaults to :post.
+
```
josevalim: Please add a note that all other options are forwarded to the underlying form and link to `Phoenix.HTML.Form.form_tag/2`. :)

test/phoenix/html/link_test.exs
```diff
@@ -24,4 +24,39 @@ defmodule Phoenix.HTML.LinkTest do
                    ~s[<a href="#" onclick="this.parentNode.submit(); return false;">hello</a>] <>
                    ~s[</form>]}
   end
+
+  test "button with post (default)" do
+
```
josevalim: No new lines after `test` and `before` (please check the other tests in the file). :)

josevalim:
I have added some comments, thanks!

lib/phoenix/html/link.ex
napcs: I'd like to ensure it has a default class though of "button" since it's a form generated by a button. Is there a better way?

josevalim: I see! You can use `Keyword.put_new(opts, :class, "button")` then.

Btw, for the hidden form generated in `link/2`, we use the class "linkmethod". Is it good enough? Maybe it should be just "link" then?

lib/phoenix/html/link.ex
napcs: Done :) See updated commit.

lib/phoenix/html/link.ex
```diff
@@ -46,4 +46,53 @@ defmodule Phoenix.HTML.Link do
...
+    * `:class` - the CSS class for the form. Defaults to "button".
+
+  All other options are forwarded to the underlying `<form>` tag.
+  See `Phoenix.HTML.Tag.form_tag/2` for more information on the
+  options above.
+  """
+  def button(text, opts) do
+    {to, opts} = Keyword.pop(opts, :to)
+    {htmlClass, opts} = Keyword.pop(opts, :class, :button)
```
scrogson: `html_class` rather than `htmlClass`

lib/phoenix/html/link.ex
```diff
@@ -46,4 +46,53 @@ defmodule Phoenix.HTML.Link do
...
+
+    unless to do
+      raise ArgumentError, "option :to is required in button/2"
+    end
+
+    form_tag(to, [class: htmlClass, enforce_utf8: false] ++ opts) do
```
scrogson: same here

PR #689, opened 2015-03-13
Do not connect automatically on `new Phoenix.Socket(..)`

scrogson:
```javascript
let socket = new Socket("/ws")
socket.connect()
...
```

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -3,6 +3,9 @@
 ## v0.11.0-dev
 
 * Backwards incompatible changes
+  * [phoenix.js] The `Socket` instance no long connects automatically. You must
+    explicitly call `connect()`
+  * [phoenix.js] `close()` has been renamed to `disconnect()`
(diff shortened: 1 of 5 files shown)


josevalim:
:+1:

chrismccord:
@scrogson lets rename `close` to `disconnect` so we have parity with `connect()`. Also please doc the changelog

scrogson:
So right now we are shipping with a minified version of `phoenix.js`. I think this may not be the route we want to take as it can cause issues in development and it's probably best to leave minification up to the user and their build tool? Would love to hear other thoughts.

I realize that this will be much better once we can extract `phoenix.js` into it's own bower/npm package. Would be nice to provide the original ES6 file, ES5 unminified and ES5 minified.

ericmj:
:+1: on shipping with non-minified phoenix.js.

scrogson:
Rebased on `master`

chrismccord:
<3 <3 <3

PR #700, opened 2015-03-13
Bugfix: hours, minutes days out of order in forms

ijt:
The fix is to put them in a list instead of a map.

### DIFF
--- a/lib/phoenix/html/form.ex
+++ b/lib/phoenix/html/form.ex
@@ -88,6 +88,7 @@ defmodule Phoenix.HTML.Form do
   alias Phoenix.HTML.Form
   import Phoenix.HTML
   import Phoenix.HTML.Tag
+  import Phoenix.HTML.Form.Datetime
 
   @doc """
   Defines the Phoenix.HTML.Form struct.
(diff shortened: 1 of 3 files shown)


josevalim:
@ijt good catch. Can we please add a simple test case too? We should just check two selects are not out of order. Which markup/select options were you getting before this change?

ijt:
Sure, but it will take me some time since I have never written a test for Phoenix before. Without this change, the options are in whatever jumbled order they occur in the map.

chrismccord:
Check out this test to get a good idea of what you'll need:
https://github.com/phoenixframework/phoenix/blob/master/test/phoenix/html/form_test.exs

ijt:
Okay, I ended up moving the datetime stuff out to a separate module called Phoenix.HTML.Form.Datetime so I could expose the hours, days etc. publicly for testing without making them part of the Form interface.

josevalim:
The reason we should have a test through the form is because maps are ordered in the current OTP version. So we should always have 0, 1, 2, ...

However this is not going to be true in future releases, that's why the change is welcome. No worries though, I will investigate here.

josevalim:
Merged locally, thank you!

PR #704, opened 2015-03-15
Channel processes with handle_info

chrismccord:
Thanks for @ericmj for pushing on me on this. It was definitely the way to go :)

This PR makes each channel subscription run it in own process, supporting handle_info and and all the OTP goodness we've come to expect. This also has the huge benefit of making the transport mulitplexing async, so channels won't block eachother when handling messages. I could use a thorough code review on this one, especially around failure handling. Right now, I trap exits in the Channel.Server so we can invoke the `leave/2` callback. Is this preferable? Sidenote: one of these days I need to clean up the channel tests as there's a lot of duplication and it was a painpoint on this refactoring.

example:

```elixir
  def join("rooms:lobby", message, socket) do
    reply socket, "join", %{status: "connected"}
    broadcast! socket, "user:entered", %{user: message["user"]}
    :timer.send_interval(5000, :ping)
    {:ok, socket}
  end

  def handle_info(:ping, socket) do
    reply socket, "new:msg", %{user: "SYSTEM", body: "ping"}
  end
```

TODO per @josevalim's feedback:

- [x] connection crashes, all channel crashes
- [x] if the channel crashes, connection cleans up
- [x] connection links and monitors each channel but the connection trap exits
- [x] Remove `:put_socket` and `:delete_socket` and use monitor instead
- [ ] Phoenix.PubSub should wrap messages ie `{:pubsub, topic, msg}`. Remove `{:socket_broadcast` wrapping.
- [x] Add client hooks for onClose/onError per channel


chrismccord:
Also, a crash in one multiplexed channel process will crash the adapter, and everyone else. This is intentional since I think other failure recovery becomes more difficult on the client. I'm also working to remove the `socket_reply` messaging duplication for the WebSocket adapter.

lib/phoenix/channel/server.ex
```diff
@@ -0,0 +1,115 @@
...
+
+  def init([socket, auth_payload]) do
+    Process.flag(:trap_exit, true)
+
+    case socket.channel.join(socket.topic, auth_payload, socket) do
+      {:ok, socket} ->
+        {:ok, socket}
+          socket = put_in(socket, [:pid], self)
+          Process.link(socket.adapter_pid)
+          send(socket.adapter_pid, {:put_socket, socket.topic, self})
```
chrismccord: The Transport.dispatch returns `{:ok, socket_pid}`, but I leaned towards keeping it process based in case we want to pool/offload the joins later.

josevalim:
> Right now, I trap exits in the Channel.Server so we can invoke the leave/2 callback. Is this preferable?

No. Don't trop exists. It is the same with GenServer. terminate is not guaranteed to be invoked unless you need to for some reason and trap exists. Unless we document exactly this, it should be fine. We may consider renaming leave to terminate now exactly because of those semantics (although I would still keep join).

chrismccord:
> No. Don't trop exists. It is the same with GenServer. terminate is not guaranteed to be invoked unless you need to for some reason and trap exists

The problem is, on <= 0.10, we say that `leave/2` is invoked when the socket closes. If we don't trap now, we've broken that (which can be okay). I think pushing more towards GenServer conventions is better since we are running inside one now. I'll remove and doc. s/leave/terminate might be the way to go

lib/phoenix/channel/server.ex
```diff
@@ -0,0 +1,115 @@
...
+          PubSub.subscribe(socket.pubsub_server, socket.pid, socket.topic, link: true)
+
+          {:ok, socket}
+
+      :ignore -> :ignore
+
+      result ->
+        Logger.error fn -> """
```
josevalim: The GenServer will already log this I think? We need to check if the error message won't be logged twice.

lib/phoenix/channel/server.ex
```diff
@@ -0,0 +1,115 @@
...
+  end
+
+  def init([socket, auth_payload]) do
+    Process.flag(:trap_exit, true)
+
+    case socket.channel.join(socket.topic, auth_payload, socket) do
+      {:ok, socket} ->
+        {:ok, socket}
+          socket = put_in(socket, [:pid], self)
+          Process.link(socket.adapter_pid)
```
josevalim: You don't need this link because the adapter already calls start_link. Unless the adapter_pid can be different of the started_pid. Is that the case?

lib/phoenix/channel/server.ex
josevalim: The transport layer should automatically add the starter process pid to the list (as it knows the topic and has the pid as a result of start_link).

lib/phoenix/channel/server.ex
```diff
@@ -0,0 +1,115 @@
...
+      Expected callback to return `{:ok, socket} | {:error, reason, socket} || {:leave, socket}`,
+      got #{inspect result}
+    """
+  end
+
+  defp leave_and_stop(reason, socket) do
+    {:ok, socket} = socket.channel.leave(reason, socket)
+
+    PubSub.unsubscribe(socket.pubsub_server, socket.pid, socket.topic)
+    send socket.adapter_pid, {:delete_socket, socket.topic}
```
josevalim: This is not reliable. The transport process should monitor each channel and automatically clean up once it receives a DOWN message.

chrismccord:
Okay, updates on this:

Most of @josevalim's feedback has been addressed, and clients can now bind to individual channel errors and close events, and rejoin (on a per channel basis). So one channel can now crash without killing the entire transport connection. Here's what it looks like on the client:

```javascript
    socket.join("rooms:lobby", {}, chan => {

      chan.onError( e => console.log("something went wrong", e) )
      chan.onClose( e => console.log("chan close", e) )   // e will be "error" if closing due to crash
      ...
```

Question 1) We currently `chan.rejoin()` onError automatically, basically we call this on channel init:

```javascript
this.onError( e => this.rejoin() )
``` 

This gives the user no way to prevent auto rejoin (same as on 0.10). This also lines up with the behavior of auto rejoinAll() when the transport connection dies. Should we make this configurable. I can't think of a case where you wouldn't want to rejoin automatically? All thoughts appreciated on the Elixir + js slides.

josevalim:
Do we even need `socket.pid` now? Maybe we should get rid of it? Also, should we rename `socket.adapter_pid` to `socket.transport_pid`?

chrismccord:
> Do we even need socket.pid now? Maybe we should get rid of it? Also, should we rename socket.adapter_pid to socket.transport_pid?

:+1:

lib/phoenix/channel/server.ex
chrismccord: Yeah we can kill the Logger.error

chrismccord:
Thanks for the feedback everyone and thanks for @ephe-meral for some initial legwork in this area. This is a really nice feature

ephe-meral:
Thanks for pushing this forward @chrismccord! That will make the integration of channels a whole lot easier :)

PR #705, opened 2015-03-16
Fix wording when `--no-brunch` is passed to phoenix.new

scrogson:
Today the wording is confusing as it says brunch was set up, when in fact it was not.


chrismccord:
<3 <3 <3

PR #711, opened 2015-03-19
Allow host to be configured for live_reload

scrogson:
Live reload doesn't work when using [pow](http://pow.cx) in development. Pow
doesn't work with websockets, so it just throws errors in the console.

I propose adding a `live_reload_host` option that can be set to
`ws://localhost:4000` in order to work around this issue.

This is just a quick hack, so I'm open to suggestions.

If we are cool with this, I'll add documentation.

### DIFF
--- a/lib/phoenix/router/live_reloader.ex
+++ b/lib/phoenix/router/live_reloader.ex
@@ -25,7 +25,7 @@ defmodule Phoenix.Router.LiveReload do
     register_before_send conn, fn conn ->
       if conn |> get_resp_header("content-type") |> html_content_type? do
         [page | rest] = String.split(to_string(conn.resp_body), "</body>")
-        body = page <> reload_assets_tag() <> Enum.join(["</body>" | rest], "")
+        body = page <> reload_assets_tag(conn) <> Enum.join(["</body>" | rest], "")
 
         put_in conn.resp_body, body
       else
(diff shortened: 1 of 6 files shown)


josevalim:
@scrogson +1! We do need something like this but we shouldn't call it host if it is going to include the port and scheme. Btw, what is your `url: [host: ...]` configuration in development in this case?

scrogson:
@josevalim so I had `url: [host: "issues.dev"]`. But I think that only effects route `*_url` helpers?

josevalim:
Yes, I was just being curious. :)

scrogson:
@josevalim do you have any suggestions as to the best approach for this? `phoenix.js` will automatically use `window.location` unless `ws://...` is explicitly specified.

josevalim:
Maybe call it live_reload_url and just specify the whole thing (we default to "/phoenix"). It would also be nice to nest the configurations so we don't have a bunch of stray `live_reload_*` keys but rather `live_reload: [url: "...", dirs: "..."]`.

Note the dirs configuration is likely going away anyway once we move to fs (which I think is slated to the next release). So feel free to break the current configuration as we will need to change it anyway.

scrogson:
Sounds good. I like that.

lib/phoenix/router/live_reloader.ex
```diff
@@ -36,12 +36,14 @@ defmodule Phoenix.Router.LiveReload do
...
-  defp reload_assets_tag() do
+  defp reload_assets_tag(conn) do
+    config = conn.private.phoenix_endpoint.config(:live_reload)
+    url = config[:url] || ""
     """
     <script>
       #{@phoenix_js}
       var phx = require("phoenix")
-      var socket = new phx.Socket("/phoenix")
+      var socket = new phx.Socket("#{url}/phoenix")
```
chrismccord: Path.join should be used here, else trailing/non-trailing url's will be problematic

chrismccord:
After we settle on these changes, remember to doc the changelog

josevalim:
:shipit:

chrismccord:
<3 <3 <3

PR #714, opened 2015-03-19
Friendly plugs - raise AlreadySentError when calling layout functions on sent conn

paulcsmith:
It's easy to think that the layout and view is set when calling `plug :action` in a controller so people (myself included) put their plugs below `plug :action` and then wonder why `put_layout` and friends don't work as expected. In reality the view and layout are set when you call `use Phoenix.Controller` so user's plugs should go above `plug :action` if you're working with the view or layout (and in most other cases too).

This PR makes it so that when you call `put_layout`, `put_new_layout`, `put_view`, `put_new_view` and `put_layout_formats` on a conn that was already sent, you get a friendly error instead of wondering why the wrong view/layout was rendered.

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -272,21 +272,25 @@ defmodule Phoenix.Controller do
   Raises `Plug.Conn.AlreadySentError` if the conn was already sent.
   """
   @spec put_layout(Plug.Conn.t, {atom, binary} | binary | false) :: Plug.Conn.t
-  def put_layout(conn, layout)
-
-  def put_layout(%Plug.Conn{state: state}, _layout) when not state in @unsent do
-    raise Plug.Conn.AlreadySentError
+  def put_layout(%Plug.Conn{state: state} = conn, layout) do
+    if state in @unsent do
+      _put_layout(conn, layout)
+    else
+      raise Plug.Conn.AlreadySentError
+    end
   end
 
-  def put_layout(conn, false) do
+  def _put_layout(conn, layout)
+
+  def _put_layout(conn, false) do
     put_private(conn, :phoenix_layout, false)
   end
 
-  def put_layout(conn, {mod, layout}) when is_atom(mod) do
+  def _put_layout(conn, {mod, layout}) when is_atom(mod) do
     put_private(conn, :phoenix_layout, {mod, layout})
   end
 
-  def put_layout(conn, layout) when is_binary(layout) or is_atom(layout) do
+  def _put_layout(conn, layout) when is_binary(layout) or is_atom(layout) do
     update_in conn.private, fn private ->
       case Map.get(private, :phoenix_layout, false) do
         {mod, _} -> Map.put(private, :phoenix_layout, {mod, layout})
@@ -294,6 +298,8 @@ defmodule Phoenix.Controller do
       end
     end
   end
+
+
 
   @doc """
   Stores the layout for rendering if one was not stored yet.
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -3,6 +3,8 @@ defmodule Phoenix.Controller do
 
   require Logger
 
+  @unsent [:unset, :set]
+
   @moduledoc """
   Controllers are used to group common functionality in the same
   (pluggable) module.
@@ -93,6 +95,15 @@ defmodule Phoenix.Controller do
       plug :put_new_layout, {Phoenix.Controller.__layout__(__MODULE__), :application}
       plug :put_new_view, Phoenix.Controller.__view__(__MODULE__)
     end
+  end
+
+  defmodule AlreadySentError do
+    defexception message: "the response was already sent. Ensure your plug " <>
+                          "is above `plug: action`"
+
+    @moduledoc """
+    Error raised when trying to modify already sent response
+    """
   end
 
   @doc """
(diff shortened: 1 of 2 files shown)


paulcsmith:
Am I missing any other functions that should raise under the same conditions?

lib/phoenix/controller.ex
```diff
@@ -3,6 +3,8 @@ defmodule Phoenix.Controller do
 
   require Logger
 
+  @unsent [:unset, :set]
```
scrogson: Should these be `:unsent` and `:sent` ?

scrogson: Nevermind, I see in Plug: 

```elixir
@type state :: :unset | :set | :file | :chunked | :sent
```

lib/phoenix/controller.ex
```diff
@@ -206,16 +217,29 @@ defmodule Phoenix.Controller do
 
   @doc """
   Stores the view for rendering.
+
+  Raises `Phoenix.ControllerAlreadySentError` if the conn was already sent.
   """
   @spec put_view(Plug.Conn.t, atom) :: Plug.Conn.t
+  def put_view(%Plug.Conn{state: state}, _module) when not state in @unsent do
```
lancehalvorsen: What do you think about changing things around so that the guard would read `when state in @sent` instead? My brain gets hung up on the double negative.

lib/phoenix/controller.ex
```diff
@@ -255,10 +279,15 @@ defmodule Phoenix.Controller do
       iex> layout(conn)
       {AppView, :print}
 
+  Raises `Phoenix.ControllerAlreadySentError` if the conn was already sent.
   """
   @spec put_layout(Plug.Conn.t, {atom, binary} | binary | false) :: Plug.Conn.t
   def put_layout(conn, layout)
 
+  def put_layout(%Plug.Conn{state: state}, _layout) when not state in @unsent do
```
lancehalvorsen: See my previous question about the double negative.

lib/phoenix/controller.ex
```diff
@@ -278,8 +307,14 @@ defmodule Phoenix.Controller do
 
   @doc """
   Stores the layout for rendering if one was not stored yet.
+
+  Raises `Phoenix.ControllerAlreadySentError` if the conn was already sent.
   """
   @spec put_new_layout(Plug.Conn.t, {atom, binary} | false) :: Plug.Conn.t
+  def put_new_layout(%Plug.Conn{state: state}, _layout) when not state in @unsent do
```
lancehalvorsen: See my previous questions about the double negative.

lib/phoenix/controller.ex
scrogson: Yes, I agree.

scrogson:
```elixir
def put_view(%Plug.Conn{state: :sent}, _module), do: raise AlreadySentError
```

Then there is no need for the `@unsent` module attribute (unless one of the other states is relevant)

scrogson: Since the final states in Plug are either `:sent` or `:chunked`, should change it to:

```elixir
@sent [:chunked, :sent]
```

And change the guards as @lancehalvorsen suggested above.

josevalim: Don't change to `@sent` because we may introduce new sent states in Plug. Furthermore, `:file` is also considered as sent.

Checking for `@unsent` as today is the best way to go. We can remove the double negative by reordering the clauses.

lib/phoenix/controller.ex
```diff
@@ -95,6 +97,15 @@ defmodule Phoenix.Controller do
     end
   end
 
+  defmodule AlreadySentError do
```
josevalim: Oh, we should just use the plug exception. No need to define a new one. :)

paulcsmith: Good point!

lib/phoenix/controller.ex
paulcsmith: I'll reorder the clauses :)

paulcsmith:
I'll work on changing the guards tomorrow or sometime next week

chrismccord:
The biggest thing to realize is that `plug :action` invokes the controller action function, and nothing else, and it's most often the last plug in your controller. This is important to understand because plugging downstream of :action is going to be rare, because as you've found, the request is already sent.

paulcsmith:
Yeah that makes sense. It's just easy to forget or accidentally put it below and then wonder why the view isn't rendered correctly. This will at least give people a clue as to where to look (I hope) :)

lib/phoenix/controller.ex
```diff
@@ -255,19 +269,28 @@ defmodule Phoenix.Controller do
...
   """
   @spec put_layout(Plug.Conn.t, {atom, binary} | binary | false) :: Plug.Conn.t
-  def put_layout(conn, layout)
+  def put_layout(%Plug.Conn{state: state} = conn, layout) do
+    if state in @unsent do
+      _put_layout(conn, layout)
+    else
+      raise Plug.Conn.AlreadySentError
+    end
+  end
```
paulcsmith: Is there a better way to do this? I'm still pretty new to Elixir so I'd love some feedback

lib/phoenix/controller.ex
```diff
@@ -255,19 +269,28 @@ defmodule Phoenix.Controller do
...
+  Raises `Plug.Conn.AlreadySentError` if the conn was already sent.
   """
   @spec put_layout(Plug.Conn.t, {atom, binary} | binary | false) :: Plug.Conn.t
-  def put_layout(conn, layout)
+  def put_layout(%Plug.Conn{state: state} = conn, layout) do
+    if state in @unsent do
+      _put_layout(conn, layout)
+    else
+      raise Plug.Conn.AlreadySentError
+    end
```
paulcsmith: Is this conditional ok or is there a way to use guards without having the double negative, e.g. `when not state in @unsent`

lib/phoenix/controller.ex
josevalim: You can do this as:

```elixir
def put_layout(%Plug.Conn{state: state} = conn, layout) when state in @unsent do
  _put_layout(...)
end

def put_layout(_, _) do
  raise Plug.Conn.AlreadySentError
end
```

But in my opinion it does not offer anything compared to your solution.

paulcsmith:
I think all comments were addressed. Is this good to merge?

josevalim:

PR #716, opened 2015-03-21
Improve generated form styling.

scrogson:
Since we ship with bootstrap by default, let's use the markup and classes
to make things look a little nicer by default.

Alt text

### DIFF
--- a/priv/static/app.css
+++ b/priv/static/app.css
@@ -29,8 +29,7 @@ body, form, ul {
 }
 
 /* Flash messages */
-p.info { color: green; }
-p.error { color: red; }
+.alert:empty { display: none; }
 
 /* Model forms */
 form.link, form.button {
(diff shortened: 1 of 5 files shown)


josevalim:
:+1:

josevalim:
Can you please look into these things too and if we can change them due to bootstrap?

* The index table: https://github.com/phoenixframework/phoenix/blob/master/priv/templates/resource/index.html.eex#L3
* The flash messages in the layout: https://github.com/phoenixframework/phoenix/blob/master/priv/templates/new/web/templates/layout/application.html.eex#L23-L24
* And maybe some of those styles are no longer necessary: https://github.com/phoenixframework/phoenix/blob/master/priv/static/app.css#L31-L58

scrogson:
@josevalim yes, will do.

scrogson:
All set.

josevalim:

PR #717, opened 2015-03-21
Ignore prod.secret.exs

scrogson:
Since it's probably not a good idea for prod.secret.exs to be in version
control, it's best we add it to .gitignore by default.

### DIFF
--- a/priv/static/bare/.gitignore
+++ b/priv/static/bare/.gitignore
@@ -5,3 +5,5 @@
 
 # Generate on crash by the VM
 erl_crash.dump
+
+/config/prod.secret.exs
--- a/priv/static/brunch/.gitignore
+++ b/priv/static/brunch/.gitignore
@@ -14,3 +14,5 @@ erl_crash.dump
 # comment this depending on your deployment strategy.
 /priv/static/css
 /priv/static/js
+
+/config/prod.secret.exs


josevalim:
I am :-1: because then your first deploy doesn't work. We fixed a similar issues with assets lately. We could leave it commented that you would likely want to remove it from version control as we do with assets.

scrogson:
Hmmm...I'm not sure I understand what you mean by the first deploy not working? I think I missed the other discussions around the assets.

`/priv/static/{js,css}` are included in `.gitignore` when generating a new app with `brunch`.

Let me know what I'm missing.

josevalim:
According to @chrismccord, the plan is to not ignore those (i.e. leave them commented). Anyway, whatever we decide here, needs to be synchronized with everything else.

Something else we need to consider is that the assets we will be generating in development will differ from the ones we want to use in production so an explicit step for deploying to production will be required anyway.

maveonair:
Ruby on Rails "solves" this situation by suggesting to use a environment variable and this is actual an idea that we can use here two:

```elixir
use Mix.Config

# In this file, we keep production configuration that
# you likely want to automate and keep it away from
# your version control system.
config :hello_phoenix, HelloPhoenix.Endpoint,
  secret_key_base: System.get_env("SECRET_KEY_BASE")
```

maveonair:
PR: https://github.com/phoenixframework/phoenix/pull/757

josevalim:
You and @limhoff-r7 were right. I have pushed this and added some comments to master. Sorry for beng stubborn and thank you!

PR #723, opened 2015-03-23
Add fs for fast and efficient live reloading

chrismccord:
This is a working starting point. This PR replaces the ChangeDetector worker and polling mechanism for `:fs` file system events. The result is much faster and efficient live reload detection.

@josevalim a couple questions:

1. Is the list of regex `:patterns` in `priv/templates/new/config/dev.exs` what you had in mind?
2. When we add md5 checking, I'm assuming you're okay with multiple clients doing the md5 lookups in their own channel? ie, no shared md5 lookup?
3. `:fs` doesn't include a way to ignore certain dirs, so I'm concerned it might be expensive if someone is constantly writing to a log file and we are receiving those fs events.


josevalim:
This looks really good! Patterns look great too! Should we add a pattern
for views in there too? Also, should we remove the md5 when a file is
deleted? It is fine to keep one md5 Hash Dict per channel.

I don't think we need to worry about performance in case of a log file. I
will also ping the fs folks about a hex release so we can extract this into
phoenix_live_reload.

-- 


chrismccord:
> Should we add a pattern for views in there too?

Definitely


> will also ping the fs folks about a hex release so we can extract this into phoenix_live_reload.

Awesome. I was hoping we could get fs on hex sooner rather than later

> Also, should we remove the md5 when a file is deleted? It is fine to keep one md5 Hash Dict per channel.

Yup. We can detect the event type and should be able to remove.

chrismccord:
I have extracted the live-reloader to a `phoenix_live_reload` project. Everything works and we can release `0.11` as soon as we get `fs` on hex. @josevalim extraction changes on this branch make the phoenix.new test fail around trying to compile a new project - the phoenix_live_reload dep is seemingly not included, but I'm not sure what the issue is. Can you take a look when you get a chance?

josevalim:
> @josevalim extraction changes on this branch make the phoenix.new test fail around trying to compile a new project 

We have two options. Depend on phoenix_live_reload or introduce a fake Phoenix.LiveReload plug that does nothing for that test. Which one do you prefer?

priv/templates/new/lib/application_name/endpoint.ex
```diff
@@ -10,6 +10,7 @@ defmodule <%= application_module %>.Endpoint do
   # :code_reloader configuration of your endpoint.
   if code_reloading? do
     plug Phoenix.CodeReloader
+    plug Phoenix.LiveReloader
```
josevalim: Maybe this one should be first? We don't need to code reload code if we are hitting the LiveReloader router.

chrismccord: :+1:

chrismccord:
> We have two options. Depend on phoenix_live_reload or introduce a fake Phoenix.LiveReload plug that does nothing for that test. Which one do you prefer?

+1 to fake Phoenix.LiveReload plug, since the goal of extraction is to not depend on this stuff.

PR #725, opened 2015-03-25
Move bootstrap css to own file when using brunch

mikker:
The huge blob of one-line css made my mind and my vim stumble so I'm proposing moving it to its' own file when using brunch. Benefits:

* Get the huge one-line blob out of the way
* Encourage modularizing / splitting up (s)css into several files


test/mix/tasks/phoenix.new_test.exs
```diff
@@ -109,6 +109,7 @@ defmodule Mix.Tasks.Phoenix.NewTest do
       assert_file "photo_blog/priv/static/images/phoenix.png"
       assert_file "photo_blog/web/static/vendor/phoenix.js"
       assert_file "photo_blog/web/static/js/app.js"
+      assert_file "photo_blog/web/static/css/_bootstrap.scss"
```
chrismccord: Should the prefix "_" be included in the filename?

josevalim:
I am :-1:. I don't think the plan is to ship with bootstrap in Phoenix. That is just our default layout. If you would really like to use bootstrap, then you should bring in the current bootstrap version and whatever makes sense to your project.

josevalim:
In any case, let's wait for others feedback, thanks for the PR!

chrismccord:
> I am :-1:. I don't think the plan is to ship with bootstrap in Phoenix. That is just our default layout. If you would really like to use bootstrap, then you should bring in the current bootstrap version and whatever makes sense to your project.

These are my thoughts also. Bootstrap is only included as an out-of-the-box nice looking default, with the idea that for real applications, you would select-all -> delete the app.scss and start with your css of choice. That said, the compiled glob also makes emacs crawl, which isn't fun. Promoting splitting css is a Good Thing, but I'm not sure we want to brand bootstrap in this way.

mikker:
I'm totally with you on pulling everything out when you start out – still, some amount of default styles add a lot to the _magic_ experience of generators.

How about defining a small set of styles - possibly with bootstrap as a starting point - and calling them `phoenix.scss`. Then we could import that at the top of `app.scss` and let the user decide when he wants to move away from the base styles (if ever).

test/mix/tasks/phoenix.new_test.exs
mikker: It's there to make sass ignore the file. So it should unless we want it to be built.

josevalim:
@mikker :+1: for everything you said. It also mimics what we have been doing with javascript. Just wait a bit longer with this change though, those files are being drastically changed in a couple minutes. :) I will ping back here.

chrismccord:
> I'm totally with you on pulling everything out when you start out – still, some amount of default styles add a lot to the magic experience of generators.
> 
> How about defining a small set of styles - possibly with a subset of bootstrap as a starting point - in a file _phoenix.scss. Then we could import that at the top of app.scss and let the user decide when he wants to move away from the base styles (if ever).

I think this is a great idea that sidesteps my concerns of branding "bootstrap" anywhere. Can you make it happen? I don't think we need to worry about trimming things down.

mikker:
I will be glad to begin setting it up when @josevalim pings :+1:

josevalim:
Ok, all done! :) You will see the files moved location a little bit, they are now in the installer directory.

mikker:
So this is how I see it done with the least amount of effort.

If you like Bootstrap we should just keep it. I never really did though. I'd love to give something more minimal and possibly tailored to Phoenix a shot if you'd like me to?

mikker:
But as we've discussed the styles are only there as a starting point/placeholder anyway so perhaps the effort is better invested on something else... I'm just eager to find places where I can actually help with my limited Elixir knowledge :new_moon_with_face:

josevalim:
@mikker thanks. I would move everything except [this rule](https://github.com/mikker/phoenix/commit/c2115cc4a03e196d0457e141b3f9095b68117dcf#diff-a699efa946e6b225747d89d242a40f68R28) to the phoenix.scss file. But then I agree with you the current patch is not adding much. If it is a matter of deleting a file, deleting `_phoenix.scss` or changing `app.scss` is not a big different. This change also duplicates brunch and non-brunch .css files which can  cause some pain in the future.

chrismccord:
> If you like Bootstrap we should just keep it. I never really did though. I'd love to give something more minimal and possibly tailored to Phoenix a shot if you'd like me to?

We shouldn't worry about this at this stage. So stock bootstrap is fine :) Watch the other issues for "level:starter" labels and feel free to chime in if one looks like fun

mikker:
I did as you suggested @josevalim.

I agree though that the duplication of the css is definitely going to give someone a headache at some point so it's up to you if you think this pr is worth it. Either way is fine by me.

I will browse the _starter_ issues and see if there's anything I am confident enough to take on :+1:

josevalim:
Thanks @mikker! I would leave things as is, unless someone provides better reasons to move to another approach. I will leave the decision up to @chrismccord though.

chrismccord:
I think we can wait on this for now. The slow editor issue is a pain, but it's a super simple fix to break things out yourself. Thanks for taking the time & I look forward to your future PRs :)

PR #731, opened 2015-03-28
Password input tag

robottokauf3:
I'm refactoring a project to use v0.10 and I noticed that there wasn't a password input helper in the form builder.


chrismccord:
This looks good! I'm tagging with `needs:review` since we require two code reviews for html related PR's. I'll let @josevalim verify and we should be good. Thanks!

josevalim:

PR #734, opened 2015-03-30
Add _input suffix to textarea/3 form input type

mutablestate:
Update `textarea/3` to `textarea_input/3` for improved naming consistency with other form input types.


lexmag:
I personally think that current naming is correct since it is actually `textarea` and not `input` like other defs.

mutablestate:
@lexmag thanks for your `input`. Looking at the HTML spec more closely i'm inclined to agree it should stay as is even though my template looks inconsistent ;-)

PR #736, opened 2015-03-30
Ignore Plug's "patch" version requirement

onkel-dirtus:


chrismccord:
This shouldn't be a problem since the patch-level is treated the same way, ie:
```elixir
iex(2)> Version.match?("0.11.1", "~> 0.11.0")
true
iex(3)> Version.match?("0.11.2", "~> 0.11.0")
true
iex(4)> Version.match?("0.11.1", "~> 0.11")
true
```

onkel-dirtus:
Beautiful

PR #737, opened 2015-03-31
Sync channel messaging with new push API

chrismccord:
We've overhauled the channel API to allow "synchronous" messaging, and I really love the results. By synchronous, I mean being able to reply to an incoming event directly, while ensuring messaging ordering for the same incoming events. This not only lets you do proper request/response messaging where necessary, but it also fixes issues we have in our <= 0.10 apis where joins were not synchronous and messages could be dropped if you fired them before you were fully joined. With these changes, we have a few high-level concepts which make up channels:

1. The client and server push messages down the socket to communicate
2. The server can reply directly to a pushed message
3. The server can broadcast events to be pushed to all subscribers

The flows looks like this:
- client `push("ev1")` -> server `handle_in("ev1")` -> `server push("ev2")` -> client `on("ev2")`
- client `push("ev1")` -> server `handle_in("ev1")` -> `server broadcast("ev2")` -> N subscribers `handle_out("ev2")` -> N subscribers `push("ev2") -> N clients `on("ev2")`
- client `push("ev1")` -> server `handle_in("ev")` -> server `{:reply, :ok, ...}` -> client `receive("ok", ...)`

Now let's see some cli/server code:
```javascript
    socket.join("rooms:lobby", {})
      .after(5000, () => console.log("We're having trouble connecting...") )
      .receive("ignore", () => console.log("auth error") )
      .receive("ok", chan => {

        // can now bind to channel crash/close events since channels are own processes
        chan.onClose( () => console.log("The channel disconnected") )
        chan.onError( () => console.log("The channel crashed!") )

        $input.onEnter( e => {
          // push without response
          chan.push("new_msg", {body: e.text, user: currentUser}) 
        })
        
        chan.on("status_change", ({status}) => $status.html(status) )
        
        chan.on("new_msg", msg => $messages.append(msg) )
        
        // push with `receive`'d response, and optional `after` hooks
        $createNotice.onClick( e => {
          chan.push("create_notice", e.data)
              .receive("ok", notice =>  console.log("notice created", notice) )
              .receive("error", reasons =>  console.log("creation failed", reasons) )
              .after(5000, () => console.log("network interruption") )
        })
    })
```

```elixir
defmodule Chat.RoomChannel do
  use Phoenix.Channel

  def join("rooms:lobby", message, socket) do
    send(self, {:after_join, message})

    {:ok, socket}
  end
  def join("rooms:" <> _private_subtopic, _message, _socket) do
    :ignore
  end

  def handle_info({:after_join, msg}, socket) do
    broadcast! socket, "user_entered", %{user: msg["user"]}
    push socket, "status_change", %{status: "waiting for users"}
    {:noreply, socket}
  end

  def handle_in("create_notice", attrs, socket) do
    changeset = Notice.changeset(%Notice{}, attrs)

    if changeset.valid? do
      Repo.insert(changeset)
      {:reply, {:ok, changeset}, socket}
    else
      {:reply, {:error, changeset.errors}, socket}
    end
  end
  
  def handle_in("new_msg", msg, socket) do
    broadcast! socket, "new_msg", %{user: msg["user"], body: msg["body"]}
    {:noreply, socket}
  end
  
  # this is forward by the default `handle_out`, but show here for clarity
  def handle_out("new_msg", msg, socket) do
    push socket, "new_msg, msg
    {:noreply, socket}
  end
end
```

Note that `{:reply, {:ok, resp}, socket}` on the server, triggers `.receive("ok", resp => {})` on the client. The "status" of the reply can be anything, ie `{:reply, {:queued, resp}, socket}` on the server, triggers `.receive("queued", resp => { })` on the client.
Also note that client joining, push, and receiving replies all have the same semantics and API now, which is quite nice.

All feedback appreciated before I merged this upstream. Thanks!

### DIFF
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -177,13 +177,16 @@ defmodule Phoenix.Channel do
       :ok
 
   """
-  def broadcast(%Socket{} = socket, event, msg) do
-    Phoenix.Channel.broadcast(socket.pubsub_server, socket, event, msg)
+  def broadcast(%Socket{joined: true} = socket, event, msg) do
+    broadcast(socket.pubsub_server, socket, event, msg)
+  end
+  def broadcast(%Socket{joined: _}, _event, _msg) do
+    raise_not_joined()
   end
   def broadcast(server, topic, event, message) when is_binary(topic) do
     broadcast_from server, :none, topic, event, message
   end
-  def broadcast(server, socket = %Socket{}, event, message) do
+  def broadcast(server, %Socket{} = socket, event, message) do
     broadcast_from server, :none, socket.topic, event, message
     {:ok, socket}
   end
@@ -192,8 +195,11 @@ defmodule Phoenix.Channel do
   Same as `Phoenix.Channel.broadcast/4`, but
   raises `Phoenix.PubSub.BroadcastError` if broadcast fails.
   """
-  def broadcast!(%Socket{} = socket, event, msg) do
-    Phoenix.Channel.broadcast!(socket.pubsub_server, socket, event, msg)
+  def broadcast!(%Socket{joined: true} = socket, event, msg) do
+    broadcast!(socket.pubsub_server, socket, event, msg)
+  end
+  def broadcast!(%Socket{joined: _}, _event, _msg) do
+    raise_not_joined()
   end
   def broadcast!(server, topic, event, message) when is_binary(topic) do
     broadcast_from! server, :none, topic, event, message
@@ -215,11 +221,14 @@ defmodule Phoenix.Channel do
 
   """
   def broadcast_from(%Socket{} = socket, event, msg) do
-    Phoenix.Channel.broadcast_from(socket.pubsub_server, socket, event, msg)
+    broadcast_from(socket.pubsub_server, socket, event, msg)
   end
-  def broadcast_from(pubsub_server, socket = %Socket{}, event, message) do
+  def broadcast_from(pubsub_server, %Socket{joined: true} = socket, event, message) do
     broadcast_from(pubsub_server, self, socket.topic, event, message)
     {:ok, socket}
+  end
+  def broadcast_from(_pubsub_server, %Socket{joined: _}, _event, _message) do
+    raise_not_joined()
   end
   def broadcast_from(pubsub_server, from, topic, event, message) when is_map(message) do
     PubSub.broadcast_from pubsub_server, from, topic, {:socket_broadcast, %Message{
@@ -235,11 +244,14 @@ defmodule Phoenix.Channel do
   raises `Phoenix.PubSub.BroadcastError` if broadcast fails.
   """
   def broadcast_from!(%Socket{} = socket, event, msg) do
-    Phoenix.Channel.broadcast_from!(socket.pubsub_server, socket, event, msg)
+    broadcast_from!(socket.pubsub_server, socket, event, msg)
   end
-  def broadcast_from!(pubsub_server, socket = %Socket{}, event, message) do
+  def broadcast_from!(pubsub_server, %Socket{joined: true} = socket, event, message) do
     broadcast_from!(pubsub_server, self, socket.topic, event, message)
     {:ok, socket}
+  end
+  def broadcast_from!(_pubsub_server, %Socket{joined: _}, _event, _message) do
+    raise_not_joined()
   end
   def broadcast_from!(pubsub_server, from, topic, event, message) when is_map(message) do
     PubSub.broadcast_from! pubsub_server, from, topic, {:socket_broadcast, %Message{
@@ -253,7 +265,7 @@ defmodule Phoenix.Channel do
   @doc """
   Sends Dict, JSON serializable message to socket.
   """
-  def push(socket, event, message) when is_map(message) do
+  def push(%Socket{joined: true} = socket, event, message) when is_map(message) do
     send socket.transport_pid, {:socket_push, %Message{
       topic: socket.topic,
       event: event,
@@ -261,7 +273,27 @@ defmodule Phoenix.Channel do
     }}
     {:ok, socket}
   end
-  def push(_, _, _), do: raise_invalid_message
+  def push(_socket, _event, message) when is_map(message) do
+    raise_not_joined()
+  end
+  def push(_, _, _), do: raise_invalid_message()
 
   defp raise_invalid_message, do: raise "Message argument must be a map"
+  defp raise_not_joined do
+    raise """
+    `push` and `broadcast` can only be called after the socket has finished joining.
+    To push a message on join, send to self and handle in handle_info/2, ie:
+
+        def join(topic, auth_msg, socket) do
+          ...
+          send(self, :after_join)
+          {:ok, socket}
+        end
+
+        def handle_info(:after_join, socket) do
+          push socket, "feed", %{list: feed_items(socket)}
+        end
+
+    """
+  end
 end
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -134,22 +134,23 @@ defmodule Phoenix.Channel do
   alias Phoenix.Socket.Message
 
   defcallback join(topic :: binary, auth_msg :: map, Socket.t) :: {:ok, Socket.t} |
-                                                                  :ignore |
-                                                                  {:error, reason :: term, Socket.t}
+                                                                  :ignore
 
-  defcallback leave(msg :: map, Socket.t) :: :ok | {:error, reason :: term}
+  defcallback terminate(msg :: map, Socket.t) :: :ok | {:error, reason :: term}
 
-  defcallback handle_in(event :: String.t, msg :: map, Socket.t) :: {:ok, Socket.t} |
-                                                                    {:leave, Socket.t} |
-                                                                    {:noreply, Socket.t} |
+  defcallback handle_in(event :: String.t, msg :: map, Socket.t) :: {:noreply, Socket.t} |
                                                                     {:reply, {status :: atom, response :: map}, Socket.t} |
                                                                     {:reply, status :: atom, Socket.t} |
-                                                                    {:error, reason :: term, Socket.t}
+                                                                    {:stop, reason :: term, Socket.t} |
+                                                                    {:stop, reason :: term, reply :: {status :: atom, response :: map}, Socket.t} |
+                                                                    {:stop, reason :: term, reply :: status :: atom, Socket.t}
 
   defcallback handle_out(event :: String.t, msg :: map, Socket.t) :: {:ok, Socket.t} |
-                                                                     {:leave, Socket.t} |
                                                                      {:noreply, Socket.t} |
-                                                                     {:error, reason :: term, Socket.t}
+                                                                     {:error, reason :: term, Socket.t} |
+                                                                     {:stop, reason :: term, Socket.t}
+
+
 
   defmacro __using__(_) do
     quote do
@@ -158,7 +159,7 @@ defmodule Phoenix.Channel do
       import unquote(__MODULE__)
       import Phoenix.Socket
 
-      def leave(message, socket), do: :ok
+      def terminate(reason, socket), do: :ok
 
       def handle_in(_event, _message, socket), do: {:noreply, socket}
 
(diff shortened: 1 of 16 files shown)


lib/phoenix/channel.ex
```diff
@@ -137,14 +137,18 @@ defmodule Phoenix.Channel do
                                                                   :ignore |
                                                                   {:error, reason :: term, Socket.t}
 
-  defcallback leave(msg :: map, Socket.t) :: {:ok, Socket.t}
+  defcallback leave(msg :: map, Socket.t) :: :ok | {:error, reason :: term}
```
josevalim: What was our conclusion regarding `leave`? I think we were discussing to rename it to terminate, weren't we? I don't remember the rationale anymore.

lib/phoenix/channel.ex
```diff
@@ -137,14 +137,18 @@ defmodule Phoenix.Channel do
...
 
   defcallback handle_in(event :: String.t, msg :: map, Socket.t) :: {:ok, Socket.t} |
```
josevalim: Are we keeping `:ok | :leave | :error` for backwards compatibility here for now? Also, we may want to consider supporting `{:leave, reason, reply, state}` and `{:leave, reason, state}` as in GenServer (maybe we should just rename it to :stop and go with terminate as mentioned above).

lib/phoenix/channel.ex
```diff
@@ -261,7 +278,28 @@ defmodule Phoenix.Channel do
...
-  def push(_, _, _), do: raise_invalid_message
+  def push(_socket, _event, message) when is_map(message) do
+    raise_not_joined()
+  end
+  def push(_, _, _), do: raise_invalid_message()
 
   defp raise_invalid_message, do: raise "Message argument must be a map"
+  defp raise_not_joined do
+    raise """
+    `push` and `broadcast` can only be called after the socket has finished joining.
```
josevalim: Fantastic error message! :heart_eyes:

lib/phoenix/channel.ex
josevalim: Ah, I remember. Leave is not always called... only if we trap exits. So I would definitely go with terminate as we have the same definition in there too.

lib/phoenix/channel/server.ex
```diff
@@ -79,23 +88,55 @@ defmodule Phoenix.Channel.Server do
...
+        {:leave, socket}
+
+    got #{inspect result}
     """
   end
 
   defp leave_and_stop(reason, socket) do
-    {:ok, socket} = socket.channel.leave(reason, socket)
-
     PubSub.unsubscribe(socket.pubsub_server, self, socket.topic)
```
josevalim: Remember we can't rely on this. Can we simply remove this line? Won't it automatically unsubscribe after it dies?

lib/phoenix/channel/transport.ex
```diff
@@ -70,39 +85,53 @@ defmodule Phoenix.Channel.Transport do
...
+  """
+  def chan_close_message(topic) do
+    %Message{topic: topic, event: "phx_chan_close", payload: %{}}
+  end
+
+  @doc """
+  Returns the `%Phoenix.Message{}` for a channel error event
+  """
+  def chan_error_message(topic) do
+    %Message{topic: topic, event: "phx_chan_error", payload: %{}}
```
josevalim: Sorry, minor nitpicking. You have used `phx_join` up and `phx_chan_close` error. Should we unify them?

lib/phoenix/channel.ex
josevalim: Btw, we don't support `:ok`, so we should probably remove it. Also, I don't think we should support `:error`... but we can talk about it later. :)

josevalim:
Beautiful. I love the join example!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

priv/static/phoenix.js
```diff
@@ -369,12 +509,13 @@ var Socket = exports.Socket = (function () {
...
       chan.rejoin();
     }
+    return chan;
   };
 
   Socket.prototype.leave = function leave(topic) {
     var message = arguments[1] === undefined ? {} : arguments[1];
 
-    this.push({ topic: topic, event: "leave", payload: message });
+    this.push({ topic: topic, event: "phx_leave", payload: message });
```
jeregrine: should we have some constants for this? Phoenix.LEAVE for example.

web/static/js/phoenix.js
```diff
@@ -1,33 +1,143 @@
...
-    this.socket   = socket
-    this.bindings = null
+    this.topic      = topic
+    this.message    = message
+    this.callback   = callback
+    this.socket     = socket
+    this.bindings   = []
+    this.afterHooks = []
+    this.recHooks   = {}
+    this.joinPush   = new Push(this, "phx_join", this.message)
```
jeregrine: Oops should have commented here.

lib/phoenix/channel/server.ex
chrismccord: yes, you're right. It's linked/trapped by the PubSub.Local server so this isn't needed

web/static/js/phoenix.js
jeregrine: Should we move this into a constant? All of these strings.

chrismccord: :+1:

stoodder: +2

lib/phoenix/channel/transport.ex
chrismccord: Nitpicking welcome. agree

web/static/js/phoenix.js
```diff
@@ -1,33 +1,143 @@
...
+        if(mergePush.recHooks.hasOwnProperty(status)){
+          this.receive(status, mergePush.recHooks[status])
+        }
+      }
+    }
+  }
+
+  send(){
+    var ref      = this.chan.socket.makeRef()
+    var refEvent = this.chan.replyEventName(ref)
```
stoodder: would make these both `let`

stoodder:
Dig the usage of ES6 here. You might consider adding semicolons to make this more reliable for minifiers or for people who concatenate all of their vendor javascript code in to a single file (points at self).  Not really strongly opinionated on this point though.

chrismccord:
> Dig the usage of ES6 here. You might consider adding semicolons to make this more reliable for minifiers or for people who concatenate all of their vendor javascript code in to a single file (points at self).

Thanks! Babel inserts semicolons in the compiled priv/static/phoenix.js, so not an issue for minifiers.

stoodder:
Perfect! Then no worries!

paulcsmith:
This is outstanding work. The examples are very clear :+1: 

* If returning `{:noreply, socket}` in `handle_in` is the most common scenario, is it a good idea for `broadcast!` (and maybe `push`?) to return `{:no_reply, socket}` so that developers don't need to type that every time? Or is `{:ok, socket}` still most common? 
* It seems that `broadcast!` is preferred over `broadcast`. The docs do not explain why that is. You could infer that it's because it throws an error if there is a failure, but it's not clear. It also seems odd (coming from Ruby) that the most used version is the `!` version. I'm sure there are good reasons for this, but it does seem counter-intuitive to me.

chrismccord:
> If returning {:noreply, socket} in handle_in is the most common scenario, is it a good idea for broadcast! (and maybe push?) to return {:no_reply, socket} so that developers don't need to type that every time? Or is {:ok, socket} still most common?

I considered this because before I had `broadcast/3` and `reply/3` both return `{:ok, socket}`, but now with the true `{:reply` vs `{:noreply,` semantics, I think explicit is better. What are other's thoughts?


> It seems that broadcast! is preferred over broadcast. The docs do not explain why that is.

I will bump the docs to better explain why. Your guess is correct. If the pubsub server is down, you likely want to crash, but there are cases where you may only want to optimistically broadcast but messages being dropped aren't mission critical.

paulcsmith:
> I considered this because before I had broadcast/3 and reply/3 both return {:ok, socket}, but now with the true {:reply vs {:noreply, semantics, I think explicit is better. What are other's thoughts?

Knowing this, I think it makes sense to return `:noreply` as a default and overriding only when you mean to reply. I would expect `push` and `broadcast` not to reply, but maybe that's just me :)

BlakeWilliams:
I can see it going both ways, but I agree with @paulcsmith here since there is the explicit reply and noreply I wouldn't expect `push` and `broadcast` to reply for you.

I think there might be value in making two methods, `push_and_reply` and `broadcast_and_reply` to make the distinction more clear.

josevalim:
:+1: to being explicit. The implicit return would make your code suddenly
fail by moving the lines around. For example you decide to log at the last
line and won't realize until later because of the implicit return.

How would push_and_reply help though? Wouldn't you need to pass the
contents you need to reply to anyway? If so, the tuple is definitely better.

Should we also change broadcast and push to just return :ok?


-- 


BlakeWilliams:
That's a good point, unless you pass both it wouldn't make much sense. In that case I'd lean towards being explicit each time.

chrismccord:
> The implicit return would make your code suddenly fail by moving the lines around.

I think this argument seals it. Explicit > implicit here.

> Should we also change broadcast and push to just return :ok?

:+1:

paulcsmith:
> :+1: to being explicit. The implicit return would make your code suddenly
fail by moving the lines around. For example you decide to log at the last
line and won't realize until later because of the implicit return.

I don't agree with this reasoning. In the example your provided the same thing would happen if you used an explicit return. 
```elixir
{:noreply, socket}
IO.puts "Oops"
```

You could say it's more obvious, but once you know `broadcast` returns `{:noreply, socket}` then it's just as obvious, and you would *have* to know broadcast returns {:noreply, socket} to build a channel right?

Anecdotally, I haven't run into problems with Phoenix as it is now. broadcast and reply return {:ok, socket} and I've never had issues with it because the guides state that that's what it does. Maybe others have had issues with it?

josevalim:
We do have the same failure with the tuple but it is much more unlikely to
happen because it is explicit and it will remind you.

Generally speaking, moving code around will cause breakage, the goal is
exactly to make those explicit so it will happen less often!

-- 


paulcsmith:
Oops, sorry I deleted my comment. I saw that Chris already made a decision so I deleted it.

You have a great point. Less breakage is good :)

josevalim:
That's ok! I am stuck on mobile, so I get to see everything :P

-- 


test/phoenix/channel_test.exs
```diff
@@ -157,7 +158,7 @@ defmodule Phoenix.ChannelTest do
   end
 
   test "#broadcast_from and #broadcast_from! raises error when msg isn't a Map" do
-    socket = Socket.put_topic(new_socket, "top:subtop")
+    socket = new_socket |> put_in([:joined], true) |> Socket.put_topic("top:subtop")
```
paulcsmith: `new_socket |> put_in([:joined], true)` is used fairly often. Maybe extract a function called `new_joined_socket`?

chrismccord: Yeah. This whole test file needs cleaned up :)

web/static/js/phoenix.js
```diff
@@ -1,33 +1,143 @@
...
       callback(reason)
-      this.trigger("chan:close", "error")
+      this.trigger("phx_chan_close", "error")
     })
   }
 
-  reset(){ this.bindings = [] }
+  reset(){
+    this.bindings = []
+    let newJoinPush = new Push(this, "phx_join", this.message, this.joinPush)
```
BlakeWilliams: `this.joinPush =` here?

web/static/js/phoenix.js
```diff
@@ -40,7 +150,14 @@ export class Channel {
                  .map( bind => bind.callback(msg) )
   }
 
-  push(event, payload){ this.socket.push({topic: this.topic, event: event, payload: payload}) }
+  push(event, payload){
+    let pushEvent = new Push(this, event, payload)
```
BlakeWilliams: `const` since this never changes?

web/static/js/phoenix.js
```diff
@@ -192,6 +311,14 @@ export class Socket {
     }
   }
 
+  // Return the next message ref, accounting for overflows
+  makeRef(){
+    let newRef = this.ref + 1
+    if(newRef === this.ref){ this.ref = 0 } else { this.ref = newRef }
```
BlakeWilliams: Expand this for clarity?

scrogson:
Beautiful work @chrismccord! Can't wait to use it.

PR #742, opened 2015-04-02
[WIP] Accept view in controller render

gjaldon:
refs: https://github.com/phoenixframework/phoenix/issues/720

This still breaks some tests. Will get back to this later and fix it up. At first glance, it was breaking tests that shouldn't break when running it locally. `action_name/1` would fail since it couldn't find `conn.private.phoenix_action` when running "render_errors_test.exs".

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -424,7 +424,16 @@ defmodule Phoenix.Controller do
   By default, Controllers render templates in a view with a similar name to the
   controller. For example, `MyApp.UserController` will render templates inside
   the `MyApp.UserView`. This information can be changed any time by using the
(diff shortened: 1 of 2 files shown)


chrismccord:
I think @scrogson was also working on this yesterday. Be sure to ping the issues before starting work. I'll let him chime in since maybe he hasn't started  yet. Thanks!

scrogson:
Yep, I started work on #720 yesterday. I'm almost finished with it. Will turn the issue into a PR shortly.

chrismccord:
@gjaldon Thanks for the work. Just make sure to ping first in the future :)

gjaldon:
@chrismccord @scrogson sorry about not coordinating first! will make sure to ping first in the future :)

PR #720, opened 2015-04-02
Allow render in controllers to accept the view

scrogson:
We should support `render(conn, View, "template")` and `render(conn, View, "template", assigns)`.

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -505,8 +505,7 @@ defmodule Phoenix.Controller do
     |> render(template, assigns)
   end
 
-  defp do_render(conn, template, format, assigns)
-    when is_binary(template) and is_binary(format) do
+  defp do_render(conn, template, format, assigns) do
     assigns = to_map(assigns)
     content_type = Plug.MIME.type(format)
     conn = prepare_assigns(conn, assigns, format)
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -423,10 +423,14 @@ defmodule Phoenix.Controller do
 
   By default, Controllers render templates in a view with a similar name to the
   controller. For example, `MyApp.UserController` will render templates inside
-  the `MyApp.UserView`. This information can be changed any time by using the
-  `put_view/2` function:
+  the `MyApp.UserView`. This information can be changed any time by using
+  `render/3`, `render/4` or the `put_view/2` function:
 
-      def show(conn) do
+      def show(conn, _params) do
+        render(conn, MyApp.SpecialView, :show, message: "Hello")
+      end
+
+      def show(conn, _params) do
         conn
         |> put_view(MyApp.SpecialView)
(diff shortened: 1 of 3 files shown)


lib/phoenix/controller.ex
```diff
@@ -468,26 +472,41 @@ defmodule Phoenix.Controller do
...
+  @spec render(Plug.Conn.t, atom, atom | binary, Dict.t) :: Plug.Conn.t
+  def render(conn, view, template, assigns)
+    when is_atom(view) and is_binary(template) or is_atom(template) do
+    conn
+    |> put_view(view)
+    |> render(template, assigns)
+  end
+
+  defp do_render(conn, template, format, assigns)
+    when is_binary(template) and is_binary(format) do
```
chrismccord: There is only one `do_render` clause so this guard is unnecessary

scrogson: Good point. Fixed.

lib/phoenix/controller.ex
```diff
@@ -468,26 +472,40 @@ defmodule Phoenix.Controller do
   `layout_formats/2` and `put_layout_formats/2` can be used to configure
   which formats support/require layout rendering (defaults to "html" only).
   """
-  @spec render(Plug.Conn.t, binary | atom, Dict.t) :: Plug.Conn.t
-  def render(conn, template, assigns) when is_atom(template) do
+  @spec render(Plug.Conn.t, binary | atom, atom | binary | Dict.t) :: Plug.Conn.t
```
josevalim: Wouldn't it be better to define multiple clauses?

```elixir
@spec render(Plug.Conn.t, binary | atom, Dict.t) :: Plug.Conn.t
@spec render(Plug.Conn.t, module, binary | atom) :: Plug.Conn.t
```

?

lib/phoenix/controller.ex
```diff
@@ -468,26 +472,40 @@ defmodule Phoenix.Controller do
...
   end
 
-  def render(conn, template, format, assigns) do
+  def render(conn, view, template)
+    when is_atom(view) and is_binary(template) or is_atom(template) do
+    render(conn, view, template, [])
+  end
+
+  @spec render(Plug.Conn.t, atom, atom | binary, Dict.t) :: Plug.Conn.t
+  def render(conn, view, template, assigns)
```
josevalim: We should rather make this call the canonical one and have the other variants calling here so we avoid the put_view indirection!

josevalim:
Once we are done, we just need to squash and I think we are good to go!

lib/phoenix/controller.ex
scrogson: I wasn't aware that was valid.

lib/phoenix/controller.ex
josevalim: Please ignore this comment.

scrogson:
Squashed and merged manually cd1fc7c

PR #747, opened 2015-04-07
Provide Phoenix.ConnTest and ship with a sample in new apps

### LINKED ISSUE #172: RFC: Controller Testing DSL
It'd be great to have an easy DSL to test controllers.

I was thinking it might be neat to write tests like:

```elixir
defmodule YourApp.Tests.Controller.Pages do
  use Phoenix.Test.Controller

  test "get :index should succeed" do
    conn = get :index
    assert conn.status == 200
  end

  test "get / should succeed" do
    conn = get "/"
    assert conn.status == 200
  end

  # Testing with cookies and session data
  test "get / with a builder should succeed" do
    conn = get "/", fn(c) ->
      c |> put_req_cookie("some_cookie", "a value")
        |> fetch_session
        |> put_session("some_session_key", "another value")
    end
    assert conn.status == 200
  end
end
```

`Phoenix.Test.Controller` would automatically determine that it's meant to test the `Pages` controller. The first test would call the `index` action directly, while the second would go through the router.

The various http method functions in the snippet above would have method signatures as follows, attempting to be a light wrapper around `Plug.Adapters.Test.Conn`:
```
get(action_or_url, params, opts, builder)
post(action_or_url, body, opts, builder)
etc...
```
where `builder` is a function for any advanced modifications to the `conn` before the request is sent off. `params`, `body`, `opts` and `builder` would be optional, hopefully using pattern matching to determine the user's intent.

josevalim:
Closes #172.

Although this is not ready for merge (we need tests for `Phoenix.ConnTest`) the API is there. Feedback is welcome!


scrogson:
Beautiful! :+1:

lib/phoenix/conn_test.ex
```diff
@@ -1,4 +1,97 @@
...
+      |> get("/")
+
+  The endpoint being tested is accessed via the `@endpoint` module
+  attribute.
+
+  ## Controller testing
+
+  The functions in this module can also be used for controller
+  testing. While not the default way of testing in Phoenix
+  applications, it may be handful in some situations.
```
jeregrine: Could you link here to default way :)

jeregrine:
Wonderful!!

chrismccord:
:shipit:

PR #757, opened 2015-04-09
Read secret key base value from the environment for production

maveonair:
Instead of suggesting to avoid check in the ```prod.secret.exs``` in source control, we rather should suggest to read the secret key base value from the environment. 

See discussion: https://github.com/phoenixframework/phoenix/pull/717

### DIFF
--- a/installer/lib/phoenix_new.ex
+++ b/installer/lib/phoenix_new.ex
@@ -140,7 +140,6 @@ defmodule Mix.Tasks.Phoenix.New do
                phoenix_dep: phoenix_dep(dev),
                pubsub_server: pubsub_server,
                secret_key_base: random_string(64),
-               prod_secret_key_base: random_string(64),
                encryption_salt: random_string(8),
                signing_salt: random_string(8),
                in_umbrella: in_umbrella?(path),
--- a/installer/templates/new/config/prod.secret.exs
+++ b/installer/templates/new/config/prod.secret.exs
@@ -1,7 +1,6 @@
 use Mix.Config
 
-# In this file, we keep production configuration that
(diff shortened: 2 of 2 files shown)


josevalim:
We can't really do this because this breaks releases. And that is not the only secret we have, we also have database credentials if usernames and passwords... :(

maveonair:
What do you think of rather mention this idea in the "Advanced deployment" of the documentation? With this approach we don't break releases but we show an approach that could might solve this issue for a few?

josevalim:
Isn't advanced deployment about releases? Anyway, I am not sure if we should advocate something that is not going to work with releases or something that we don't do by default.

josevalim:
Also, using environment variables is fine if you are deploying to heroku, because that is about the only option. However, if you are using Chef, I would prefer to use the file. So maybe we need multiple deployment guides, or explore the different approaches to configuration (and ship with a reasonable default).

jeregrine:
:+1: for multiple guides. No one solution is going to fit everything.

maveonair:
Yeah this was the idea. I mean in the end, everybody can use the preferred solution for that problem but it would be at least a good idea to mention the different solutions. And the reason why I came up with the deployment documentation is that it is relevant for your production deployment or rather "running your production system".

josevalim:
Good. Can we move this discussion to the guides then? I think Phoenix should not promote environment variables for now...

chrismccord:
Agree. This would be easier to address in Phoenix if we had the proposed guides in place first. https://github.com/lancehalvorsen/phoenix-guides Thanks!

maveonair:
Awesome! Thank you very much :)

Issue #735, opened 2015-03-30
Digest, gzip and generate manifest files for assets


josevalim:
We need to do this before Plug 1.0 as we need to figure out how digest files will be served by Plug.Static.

laurocaetano:
I'm working on this one!

I'll open a W.I.P Pull Request as soon as I get something working.

laurocaetano:
I have some questions:

1 - How the links will work? Example: A CSS that points to an image.
```css 
.class { background-image: url('image.png') }
```
How can we make it point to the digested file?

2 - How it will be integrated with third party assets? One [example](https://github.com/rails/sprockets/blob/master/UPGRADING.md#preference-for-asset-manifest-and-links) is how links works on sprockets.

I'm completely new to phoenix codebase, so I might be saying something not related with this issue at all. Ideas?

josevalim:
That's a very good point we will need to discuss them eventually. I think we can leave them to a later stage though since our static system will still serve image.png with conditional cache.

laurocaetano:
That's great, thanks!

laurocaetano:
Hi there!

I've been working on this issue the past few days and I have a draft for the implementation: https://gist.github.com/laurocaetano/ee876a71fc0c32afa6dd What do you think?

How is this going to work on Phoenix? Is it going to be a new mix task or something?

laurocaetano:
I think this can be closed now, right?

PR #758, opened 2015-04-09
Introduce the digest task

laurocaetano:
This task is responsible for digesting and compressing the assets files as described at #735.

I used the [discourse's](https://github.com/discourse/discourse) assets to make a benchmark since it has ~ 991 assets files. The results were pretty nice.

```
phoenix (introduce-digest-task) $ time mix phoenix.digest /Users/laurocaetano/Desktop/discourse/app/assets -o /tmp/assets
Check your digested files at '/tmp/assets'

real	0m2.446s
user	0m1.926s
sys	0m1.290s
```

This task will read the input/output path and proceed as following:
  * Filter for files (excluding directories)
  * Get the file digest
  * Compress the file
  * Save the files to disk
  * Generate the manifest

It also takes account the relative path when saving the files to the output path, see:

```
   |-fonts
   |-images
   |---favicons
   |-javascripts
   |---admin
   |-----components
   |-----controllers
   |-------modals
   |-----helpers
   |-----models
   |-----routes
   |-----templates
   |-------components
   |-------logs
   |-------modal
   |-------reports
   |-------site-settings
   |-----views
   |-------logs
   |-------modals
   |---defer
   |---discourse
#....
```
and the `manifest.json` will look like:

```json
   {
     "./phoenix.png" : "./phoenix-5bd99a0d17dd41bc9d9bf6840abcc089.png",
     "./phoenix.js" : "./phoenix-4d4e132be213566829e0f28686fc04a0.js"
   }
```

Thoughts? 

Thanks @josevalim for helping :heart:

### DIFF
--- a/lib/mix/tasks/phoenix.digest.ex
+++ b/lib/mix/tasks/phoenix.digest.ex
@@ -1,0 +1,46 @@
+defmodule Mix.Tasks.Phoenix.Digest do
+  use Mix.Task
+  @default_input_path "priv/static"
+
+  @shortdoc "Digests and compress static files."
+
+  @moduledoc """
+  Digests and compress static files.
+
+      mix phoenix.digest priv/static -o public/assets
+
+  The first argument is the path where the static files are located. The
+  `-o` option indicates the path that will be used to save the digested and
+  compressed files.
+
+  If no path is given, it will use `priv/static` as the input and output path.
+
+  The output folder will contain:
+
+    * the original file
+    * a compressed file with gzip
+    * a file containing the original file name and its digest
+    * a compressed file containing the file name and its digest
+    * a manifest file
+
+  Example of generated files:
+
+    * application.js.erb
+    * application.js.erb.gz
+    * application.js-eb0a5b9302e8d32828d8a73f137cc8f0.erb
+    * application.js-eb0a5b9302e8d32828d8a73f137cc8f0.erb.gz
+    * manifest.json
+  """
+
+  @doc false
+  def run([input|args]) do
+    {args, _, _} = OptionParser.parse(args, aliases: [o: :output])
+    input_path = input || @default_input_path
+    output_path = args[:output] || input_path
+
+    case Phoenix.Digester.compile(input_path, output_path) do
+      :ok -> Mix.shell.info [:green, "Check your digested files at '#{output_path}'."]
+      {:error, :invalid_path} -> Mix.raise "The input path '#{input_path}' does not exist."
+    end
+  end
+end
--- a/lib/phoenix/digester.ex
+++ b/lib/phoenix/digester.ex
@@ -1,0 +1,97 @@
+defmodule Phoenix.Digester do
+  @digested_file_regex ~r/(-[a-fA-F\d]{32})/
+
+  @moduledoc """
+  Digests and compress static files.
+
+  For each file under the given input path, Phoenix will generate a digest
+  and also compress in `.gz` format. The filename and its digest will be
+  used to generate the manifest file. It also avoid duplications checking
+  for already digested files.
+  """
+
+  @doc """
+  Digests and compress the static files and save them in the given output path.
+
+    * `input_path` - The path where the assets are located
+    * `output_path` - The path where the compiled/compressed files will be saved
+  """
+  @spec compile(String.t, String.t) :: :ok | {:error, :invalid_path}
+  def compile(input_path, output_path) do
+    if File.exists?(input_path) do
+      unless File.exists?(output_path), do: File.mkdir_p!(output_path)
+
+      input_path
+      |> filter_files
+      |> do_compile(output_path)
+      |> generate_manifest(output_path)
+      :ok
+    else
+      {:error, :invalid_path}
+    end
+  end
+
+  defp filter_files(input_path) do
+    input_path
+    |> Path.join("**")
+    |> Path.wildcard
+    |> Enum.filter(&(!File.dir?(&1) && !compiled_file?(&1)))
+    |> Enum.map(&(map_file(&1, input_path)))
+  end
+
+  defp do_compile(files, output_path) do
+    Enum.map(files, fn (file) ->
+      file
+      |> digest
+      |> compress
+      |> write_to_disk(output_path)
+    end)
+  end
+
+  defp generate_manifest(files, output_path) do
+    entries = Enum.reduce(files, %{}, fn (file, acc) ->
+      Map.put(acc, Path.join(file.relative_path, file.filename),
+        Path.join(file.relative_path, file.digested_filename))
+    end)
+
+    manifest_content = Poison.Encoder.encode(entries, [])
+    File.write!(Path.join(output_path, "manifest.json"), manifest_content)
+  end
+
+  defp compiled_file?(file_path) do
+    Regex.match?(@digested_file_regex, Path.basename(file_path)) ||
+      Path.extname(file_path) == ".gz"
+  end
+
+  defp map_file(file_path, input_path) do
+    %{absolute_path: file_path,
+      relative_path: Path.relative_to(file_path, input_path) |> Path.dirname,
+      filename: Path.basename(file_path),
+      content: File.read!(file_path)}
+  end
+
+  defp compress(file) do
+    Map.put(file, :compressed_content, :zlib.gzip(file.content))
+  end
+
+  defp digest(file) do
+    name = Path.rootname(file.filename)
+    extension = Path.extname(file.filename)
+    digest = Base.encode16(:erlang.md5(file.content), case: :lower)
+    Map.put(file, :digested_filename, "#{name}-#{digest}#{extension}")
+  end
+
+  defp write_to_disk(file, output_path) do
+    File.mkdir_p!(Path.join(output_path, file.relative_path))
+    path = Path.join(output_path, file.relative_path)
+
+    # compressed files
+    File.write!(Path.join(path, file.digested_filename <> ".gz"), file.compressed_content)
+    File.write!(Path.join(path, file.filename <> ".gz"), file.compressed_content)
+    # uncompressed files
+    File.write!(Path.join(path, file.digested_filename), file.content)
+    File.write!(Path.join(path, file.filename), file.content)
+
+    file
+  end
+end
--- a/test/mix/tasks/phoenix.digest_test.exs
+++ b/test/mix/tasks/phoenix.digest_test.exs
@@ -1,0 +1,27 @@
+defmodule Mix.Tasks.Phoenix.DigestTest do
+  use ExUnit.Case, async: true
+
+  test "fails when the given paths are invalid" do
+    assert_raise Mix.Error, "The input path 'invalid_path' does not exist.", fn ->
+      Mix.Tasks.Phoenix.Digest.run(["invalid_path"])
+    end
+  end
+
+  test "digests and compress files" do
+    output_path = Path.join("tmp", "digested")
+    input_path = "priv/static"
+
+    File.mkdir_p!(output_path)
+
+    Mix.Tasks.Phoenix.Digest.run([input_path, "-o", output_path])
+    assert_received {:mix_shell, :info, ["Check your digested files at 'tmp/digested'."]}
+  end
+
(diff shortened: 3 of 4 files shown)


lib/mix/tasks/phoenix.digest.ex
```diff
@@ -0,0 +1,47 @@
+defmodule Mix.Tasks.Phoenix.Digest do
+  use Mix.Task
+  @default_input_path "priv/static"
+
+  @shortdoc "Digests and compress assets files"
```
josevalim: static files

lib/mix/tasks/phoenix.digest.ex
```diff
@@ -0,0 +1,47 @@
...
+
+  @moduledoc """
+  Digests and compress assets files.
```
josevalim: static files :)

lib/mix/tasks/phoenix.digest.ex
```diff
@@ -0,0 +1,47 @@
...
+
+    mix phoenix.digest priv/static -o public/assets
```
josevalim: Four spaces for code (as in markdown!)

lib/mix/tasks/phoenix.digest.ex
```diff
@@ -0,0 +1,47 @@
...
+
+  The first argument is the path where the original assets are located. The
+  `-o` option indicates the path that will be used to save the digested and
+  compressed files.
+
+  If no path is given, it will use "priv/static" as the input and output path.
+
+  The output folder will contain:
```
josevalim: Great docs!

lib/mix/tasks/phoenix.digest.ex
```diff
@@ -0,0 +1,47 @@
...
+
+    * the original file
+    * a compressed file with gzip
+    * a file containing the original file name and its digest
+    * a compressed file containing the file name and its digest
+    * a manifest file
+
+  Example of generated files:
```
josevalim: New line!

lib/mix/tasks/phoenix.digest.ex
```diff
@@ -0,0 +1,47 @@
...
+    * manifest.json
+  """
+
+  @doc false
+  def run([input|args]) do
+    {args, _, _} = OptionParser.parse(args, aliases: [o: :output])
+    input_path = input || @default_input_path
+    output_path = args[:output] || input_path
+
+    Mix.Task.run "app.start"
```
josevalim: Do we need to start the app? I don't think we do!

lib/mix/tasks/phoenix.digest.ex
```diff
@@ -0,0 +1,47 @@
...
+
+    case Phoenix.Digester.compile(input_path, output_path) do
+      :ok -> Mix.shell.info [:green, "Check your digested files at '#{output_path}'."]
+      {:error, :invalid_path} -> Mix.shell.error [:red, "The given paths are invalid."]
```
josevalim: Now I am wondering... should we raise (`Mix.raise`) or just log? I would raise because if you are chaining, the next task won't run. Thoughts?

lib/phoenix/digester.ex
```diff
@@ -0,0 +1,94 @@
+defmodule Phoenix.Digester do
+  @digested_file_regex ~r/(-[a-fA-F\d]{32})/
+
+  @moduledoc """
+    Digests and compress assets files.
```
josevalim: No indentation here.

lib/phoenix/digester.ex
```diff
@@ -0,0 +1,94 @@
...
+
+  @doc """
+  Compile the assets and save them in the given output path.
+
+    * `input_path` - The path where the assets are located
+    * `output_path` - The path where the compiled/compressed files will be saved
+  """
+  @spec compile(String.t, String.t) :: :ok | {:error, :invalid_path}
+  def compile(input_path, output_path) do
+    if File.exists?(input_path) && File.exists?(output_path) do
```
josevalim: The output path does not need to exist. Just call `File.mkdir_p!(output_path)` to ensure we have created it.

lib/phoenix/digester.ex
```diff
@@ -0,0 +1,94 @@
...
+      |> do_compile(output_path)
+      |> generate_manifest(output_path)
+      :ok
+    else
+      {:error, :invalid_path}
+    end
+  end
+
+  defp filter_files(input_path) do
+    Path.wildcard(Path.join(input_path, "**"))
```
josevalim: You could go more pipeline-ish:

```elixir
input_path
|> Path.join("**")
|> Path.wildcard
|> ...
```

:)

lib/phoenix/digester.ex
```diff
@@ -0,0 +1,94 @@
...
+    |> Enum.filter(&(!File.dir?(&1) && !compiled_file?(&1)))
+    |> Enum.map(&(map_file(&1, input_path)))
+  end
+
+  defp do_compile(files, output_path) do
```
josevalim: Beautiful.

lib/phoenix/digester.ex
```diff
@@ -0,0 +1,94 @@
...
+    File.write!(Path.join(output_path, "manifest.json"), manifest_content)
+  end
+
+  defp compiled_file?(file_path) do
+    Regex.match?(@digested_file_regex, Path.basename(file_path)) ||
+      Path.extname(file_path) == ".gz"
+  end
+
+  defp map_file(file_path, input_path) do
+    %{}
```
josevalim: What about:

```elixir
%{absolute_path: file_path,
   relative_path: ...,
   ...}
```

be cleaner?

test/phoenix/digester_test.exs
```diff
@@ -0,0 +1,34 @@
+defmodule Phoenix.DigesterTest do
+  use ExUnit.Case, async: true
+
+  test "fails when the given paths are invalid" do
+    assert {:error, :invalid_path} = Phoenix.Digester.compile("../some invalid?/path", "/other_invalid ?? /path")
+  end
+
+  test "digests and compress files" do
+    output_path = Path.join("tmp", "digested")
```
josevalim: We should use a different directory from the mix task. Just in case they are both running at the same file and something weird happens.

josevalim: Also we are not testing that compiling twice won't regenerate files (that for example we won't have `.gz.gz`).

test/phoenix/digester_test.exs
```diff
@@ -0,0 +1,34 @@
...
+    input_path = "priv/static"
+    File.mkdir_p!(output_path)
+
+    assert :ok = Phoenix.Digester.compile(input_path, output_path)
+
+    input_files = assets_files(input_path)
```
josevalim: I don't like tests like this because we are not really testing anything. If `priv/static` becomes empty (no assets) the test would still pass. I would explicitly pick one file from `priv/static` (or even create a dummy input directory in tmp) and explicitly assert on it.

lib/mix/tasks/phoenix.digest.ex
laurocaetano: Oh we don't need to :sweat_smile:. I will remove, thanks :)

lib/mix/tasks/phoenix.digest.ex
laurocaetano: I didn't think in this case and it totally makes sense. :+1: for raising

test/phoenix/digester_test.exs
laurocaetano: :+1:

test/phoenix/digester_test.exs
laurocaetano: You are right, :+1: for picking one file from `priv/static`.

laurocaetano:
@josevalim updated with your suggestion. What do you think?

Thanks again for your review :)

test/phoenix/digester_test.exs
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  test "doesn't duplicate files when digesting and compressing twice" do
+    path = Path.join("tmp", "static_files")
+    input_file = Path.join(path, "file.js")
+    File.mkdir_p!(path)
+    File.write!(input_file, "console.log('test');")
+
+    assert :ok = Phoenix.Digester.compile(path, path)
+    assert :ok = Phoenix.Digester.compile(path, path)
```
laurocaetano: Its digesting twice for the same path. It is now checking if it's duplicating the `.gz` and the digest.

josevalim: Should we add a comment there?

test/phoenix/digester_test.exs
```diff
@@ -0,0 +1,50 @@
...
+  end
+
+  test "digests and compress files" do
+    output_path = Path.join("tmp", "compiled")
+    input_path = "priv/static"
+    File.mkdir(output_path)
+
+    assert :ok = Phoenix.Digester.compile(input_path, output_path)
+
+    [input_file, _]= assets_files(input_path)
```
josevalim: Let's pick one file explicitly? The test will be clearer if we have this:

```elixir
assert Enum.member?(output_files, "images/phoenix.png")
assert Enum.member?(output_files, "images/phoenix.png.gz")
```

and so on.

test/phoenix/digester_test.exs
```diff
@@ -0,0 +1,50 @@
...
+  use ExUnit.Case, async: true
+
+  test "fails when the given paths are invalid" do
+    assert {:error, :invalid_path} = Phoenix.Digester.compile("../some invalid?/path", "/other_invalid ?? /path")
+  end
+
+  test "digests and compress files" do
+    output_path = Path.join("tmp", "compiled")
+    input_path = "priv/static"
+    File.mkdir(output_path)
```
josevalim: Shouldn't the compile task create the directory for us?

lib/mix/tasks/phoenix.digest.ex
```diff
@@ -0,0 +1,46 @@
...
+
+  @doc false
+  def run([input|args]) do
+    {args, _, _} = OptionParser.parse(args, aliases: [o: :output])
+    input_path = input || @default_input_path
+    output_path = args[:output] || input_path
+
+    case Phoenix.Digester.compile(input_path, output_path) do
+      :ok -> Mix.shell.info [:green, "Check your digested files at '#{output_path}'."]
+      {:error, :invalid_path} -> Mix.raise "The given paths are invalid."
```
josevalim: Now we can say "The input path '#{input_path}' does not exist."

test/mix/tasks/phoenix.digest_test.exs
```diff
@@ -0,0 +1,34 @@
+Code.require_file "../../../installer/test/mix_helper.exs", __DIR__
+
+defmodule Mix.Tasks.Phoenix.DigestTest do
+  use ExUnit.Case
+
+  setup do
+    Mix.Task.clear
```
josevalim: We don't need this because you call the tasks directory and not via `Mix.Task.run/1`. This test case can even be async (async: true).

test/phoenix/digester_test.exs
josevalim: Ignore me. It is fine as is as it is a new test case.

test/phoenix/digester_test.exs
```diff
@@ -0,0 +1,50 @@
...
+    File.mkdir_p!(path)
+    File.write!(input_file, "console.log('test');")
+
+    assert :ok = Phoenix.Digester.compile(path, path)
+    assert :ok = Phoenix.Digester.compile(path, path)
+
+    output_files = assets_files(path)
+
+    duplicated_digested_file_regex = ~r/(-[a-fA-F\d]{32}.[\w|\d]*.[-[a-fA-F\d]{32})/
+    assert Enum.any?(output_files, fn (f) ->
```
josevalim: It should be `Enum.all?` :)

josevalim: Or just be explicit and check explicitly the duplicated files do not exist. For example, no `file.js.gz.gz`.

laurocaetano:
@josevalim thanks again for reviewing and sorry for those newbies mistakes, now I think we're good to go :)

josevalim:

Issue #759, opened 2015-04-10
Decimal ecto fields specified in mix phoenix.gen.html create number fields in the HTML


lsimoneau:
Example:

```
mix phoenix.gen.html Product products name:string price:decimal quantity:integer
```

At least in Chrome, this makes it impossible to submit the form with a decimal input.

![screen shot 2015-04-10 at 7 31 24 pm](https://cloud.githubusercontent.com/assets/31768/7085251/41b92480-dfb8-11e4-9b80-848c7e06082c.png)

It's easy enough to change the template back to a text field, but it seems like it would be best to only output a number field for integer columns.

josevalim:
> It's easy enough to change the template back to a text field, but it seems like it would be best to only output a number field for integer columns.

Yes, defnitely, please send a PR?

josevalim:
Another option is this: https://blog.isotoma.com/2012/03/html5-input-typenumber-and-decimalsfloats-in-chrome/

Just add step: "any" to float and decimals.

lsimoneau:
Yep, ok, I'll have a look at making a PR for this (using `step: "any"`, that seems like the best solution).

lsimoneau:
Ok, I've had a look into the code, and I'd like to know what others think before diving in. Looking at the `inputs/1` function [here](https://github.com/phoenixframework/phoenix/blob/1ebea0bd1651a03914c6bd455c4be69bf1033108/lib/mix/tasks/phoenix.gen.resource.ex#L69), a two-element tuple is passed along in the bindings to the template, in the form `{ :name_of_field, :form_helper_function }`. In order to add `step: "all"` to decimal and float column types, I'd need to either:

1. Update `inputs/1` to return a 3-element tuple, with the last element being a map of options to be passed to the helper function. This is complicated and requires more changes to more files, but feels like it provides some greater flexibility for handling other similar cases where different input types might potentially want different form attributes.

2. Create a new helper function like `decimal_input` that generates a number input with `step: "any"`. This feels like a hack to me, but is certainly much narrower in scope.

Any thoughts?

lsimoneau:
Alternately I could just make decimals and floats output a text field. That feels kind of like giving up, but at least avoids the bad experience of generating a form that doesn't work at all.

josevalim:
@lsimoneau If you do a 3 tuple, something like this would be the best:

[...]

PR #761, opened 2015-04-10
Add "step: any" to generated number inputs for decimal and float fields

lsimoneau:
Addresses #759. Implemented as per the discussion there.


josevalim:

PR #762, opened 2015-04-11
Adds defcallback and default impl for channel handle_info/2

ephe-meral:
This is to avoid having to provide default implementations of `handle_info` in the channels.

I also found that `handle_out` defcallback still had the `{:ok, Socket.t}` result listed as a valid result, which afaik it is not anymore...

Should not break backwards compatibility.
~ cheers, Johanna


josevalim:
We should add tests that the default channels work out of the box. If the default channel is broken... it will cause a lot of confusion.

PR #765, opened 2015-04-12
Print better error when status is formatted incorrectly

BlakeWilliams:
When the status isn't a two element tuple or an atom show an error that
displays the two proper formats of `reply`.

The wording could probably be better, I'd love feedback on that. I'm also not sure where/how exactly to add a test for this and would also love some pointers for doing that.

### DIFF
--- a/lib/phoenix/channel/server.ex
+++ b/lib/phoenix/channel/server.ex
@@ -95,6 +95,16 @@ defmodule Phoenix.Channel.Server do
                                 response: %{}}
     {:noreply, socket}
   end
+  defp handle_result({:reply, status, _socket}, :handle_in) do
+    raise """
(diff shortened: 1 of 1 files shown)


josevalim:
@chrismccord are we testing those raises currently or there is no way right now?

josevalim:
@BlakeWilliams and thanks for fixing this! <3

chrismccord:
@josevalim We can test that a bad return at least causes a crash. I will merge this locally and tweak the error msg. Thanks @BlakeWilliams

josevalim:
@chrismccord please merge this. :)

chrismccord:
Give me 15 min

chrismccord:
<3 <3 <3

PR #769, opened 2015-04-15
Fix the case of Location in redirected_to/2

parkerl:
@josevalim love the changes you made to `redirected_to`! However our shiny new method was not working in the tests I am generating for that other PR. It looks like `Location` needs to be capitalized as that is how it is set in a real response? This change makes `redirected_to` work beautifully for me.

### DIFF
--- a/lib/phoenix/conn_test.ex
+++ b/lib/phoenix/conn_test.ex
@@ -276,7 +276,7 @@ defmodule Phoenix.ConnTest do
   end
 
   def redirected_to(%Conn{status: status} = conn, status) do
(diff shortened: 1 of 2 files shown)


parkerl:
By the way there are 3 tests failing in this PR but it look like they are not my doing.

josevalim:
They should be lowercased. It is a bug in Phoenix. I will fix in a sec.

josevalim:
Fixed in master. :)

parkerl:
Nice! Thanks.

PR #770, opened 2015-04-15
Generate test files for gen html

parkerl:
@josevalim how does this look? I think there might be a couple more test cases around not redirecting when the params for new or update are invalid?

### DIFF
--- a/lib/mix/tasks/phoenix.gen.html.ex
+++ b/lib/mix/tasks/phoenix.gen.html.ex
@@ -98,13 +98,13 @@ defmodule Mix.Tasks.Phoenix.Gen.Html do
           [k, "float"]    -> {k, "120.5"}
           [k, "decimal"]  -> {k, "120.5"}
           [k, "boolean"]  -> {k, true}
-          [k, "text"]     -> {k, "a binary"}
-          [k, "date"]     -> {k, "a binary"}
-          [k, "time"]     -> {k, "a binary"}
-          [k, "datetime"]  -> {k, [year: 2014, month: 12, day: 1, hour: 12, min: 1]}
+          [k, "text"]     -> {k, "some content"}
+          [k, "date"]     -> {k, "2010-04-17"}
+          [k, "time"]     -> {k, "14:00:00"}
+          [k, "datetime"]  -> {k, "2010-04-17 14:00:00"}
           [k, "uuid"]     -> {k, "7488a646-e31f-11e4-aace-600308960662"}
-          [k, _]          -> {k, "a binary"}
-          [k]             -> {k, "a binary"}
+          [k, _]          -> {k, "some content"}
+          [k]             -> {k, "some content"}
         end
       {to_atom(k), v}
     end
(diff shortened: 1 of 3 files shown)


lib/mix/tasks/phoenix.gen.html.ex
```diff
@@ -88,6 +89,29 @@ defmodule Mix.Tasks.Phoenix.Gen.Html do
...
+    sample_params = Enum.map attrs, fn attr ->
+      {k, v} =
+        case String.split(attr, ":", parts: 3) do
+          [k, _, _]       -> {k, []}
+          [k, "integer"]  -> {k, 42}
+          [k, "float"]    -> {k, "120.5"}
+          [k, "decimal"]  -> {k, "120.5"}
+          [k, "boolean"]  -> {k, true}
+          [k, "text"]     -> {k, "a binary"}
+          [k, "date"]     -> {k, "a binary"}
```
josevalim: "2010-04-07"

lib/mix/tasks/phoenix.gen.html.ex
```diff
@@ -88,6 +89,29 @@ defmodule Mix.Tasks.Phoenix.Gen.Html do
...
+          [k, "time"]     -> {k, "a binary"}
```
josevalim: "14:00:00"

lib/mix/tasks/phoenix.gen.html.ex
```diff
@@ -88,6 +89,29 @@ defmodule Mix.Tasks.Phoenix.Gen.Html do
...
+          [k, "datetime"]  -> {k, [year: 2014, month: 12, day: 1, hour: 12, min: 1]}
```
josevalim: "2010-04-17 14:00:00"

josevalim:
@parkerl This looks great! I have added some comments related to the sample params. Also, maybe it would be best if we use "content" or "some content" instead of "a binary"?

parkerl:
What!?! You didn't like my extremely creative sample data? ;-)

Was wondering if we need to generate create and update tests that fail validation and then don't redirect?

josevalim:
Your dates are fine. I was just suggesting a simplified format.

josevalim:
> Was wondering if we need to generate create and update tests that fail validation and then don't redirect?

I think that is a great idea. I have merged this so I can push the model stuff too. Please send another pull request! I think we can make an @invalid_params that by default is just `[user: []]`.

Thank you so much, this is great!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

priv/templates/resource/controller_test.exs
```diff
@@ -0,0 +1,38 @@
+defmodule <%= module %>ControllerTest do
+  use <%= base %>.ConnCase
+
+  @valid_params [<%= singular %>: <%= sample_params %>]
+
+  test "GET /<%= plural %>" do
+    conn = get conn(), <%= singular %>_path(conn, :index)
+    assert conn.resp_body =~ "Listing <%= plural %>"
```
josevalim: This can now be `html_response(conn, 200) =~ "Listing <%= plural %>"`. Don't worry, I have changed it already but you will want to use the `html_response` stuff for the failure scenarios in create/update. :)

PR #784, opened 2015-04-19
Add missing space in prod config

koleksiuk:
This PR adds missing space in prod config template. :)

### DIFF
--- a/installer/templates/new/config/prod.exs
+++ b/installer/templates/new/config/prod.exs
@@ -21,7 +21,7 @@ config :<%= application_name %>, <%= application_module %>.Endpoint,
 # To get SSL working, you will need to add the `https` key
 # to the previous section:
 #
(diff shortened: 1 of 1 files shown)


installer/templates/new/config/prod.exs
```diff
@@ -21,7 +21,7 @@ config :<%= application_name %>, <%= application_module %>.Endpoint,
 # To get SSL working, you will need to add the `https` key
 # to the previous section:
 #
-#  config:<%= application_name %>, <%= application_module %>.Endpoint,
```
josevalim: The space should be on the left side of `:`. :)

koleksiuk: You're right, I don't know how it happened. :) Fixed

chrismccord:
Thanks!

PR #787, opened 2015-04-20
Declare missing applications in mix.exs

sevos:
postgrex and phoenix_ecto should be also listed in
the application configuration in order to make these
libs' modules available in within an OTP release.

Reference: https://github.com/bitwalker/exrm#common-issues

I need small help with testing the changed generator. The approach described in docs (with `--dev` param) seems to use my global phoenix installation. (I tried from the git checkout directory and from installer/ directory)

### DIFF
--- a/installer/templates/new/mix.exs
+++ b/installer/templates/new/mix.exs
@@ -19,7 +19,8 @@ defmodule <%= application_module %>.Mixfile do
   # Type `mix help compile.app` for more information
   def application do
     [mod: {<%= application_module %>, []},
-     applications: [:phoenix, :cowboy, :logger<%= if ecto do %>, :ecto<% end %>]]
+     applications: [:phoenix, :cowboy, :logger<%= if ecto do %>,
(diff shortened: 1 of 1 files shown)


josevalim:
Great catch!

@sevos there is no need to make them included applications. I think the correct patch would be to replace `, :ecto` in the current code by `, :postgrex, :phoenix_ecto`.

scrogson:
We should probably add an option to specify the adapter. Right now we're assuming everyone will use Postgres.

josevalim:
@scrogson an adapter option would be great if someone wants to tackle it.

scrogson:
@josevalim sure. I'll create an issue and start it.

sevos:
@josevalim I've tested it on my pet-app and it works! I still don't see a way to test the generator itself (to see that it creates a project correctly). The change is simple though, so maybe it is not necessary ;)

scrogson:
@sevos thanks! I'll merge and create a test when adding the `--adapter` option.

:+1:

PR #788, opened 2015-04-20
Add --adapter option to phoenix.new

scrogson:
We should have an option to specify which adapter should be used with `phoenix_ecto`. Right now we assume everyone is using Postgres.

```
mix phoenix.new my_app --adapter mysql
```

Should it default to `pg` ?

### DIFF
--- a/installer/lib/phoenix_new.ex
+++ b/installer/lib/phoenix_new.ex
@@ -89,8 +89,8 @@ defmodule Mix.Tasks.Phoenix.New do
     * `--module` - the name of the base module in
       the generated skeleton
 
-    * `--db-adapter` - specify the database adapter for ecto.
-      Values can be `mysql`, `tds`, or `pg`. Defaults to `pg`
+    * `--database` - specify the database adapter for ecto.
+      Values can be `mysql` or `mssql`. Defaults to `postgres`
 
     * `--no-brunch` - do not generate brunch files
       for static asset building
@@ -139,7 +139,7 @@ defmodule Mix.Tasks.Phoenix.New do
     ecto = Keyword.get(opts, :ecto, true)
     brunch = Keyword.get(opts, :brunch, true)
 
-    {adapter_app, adapter_module} = set_ecto_adapter(opts[:db_adapter])
+    {adapter_app, adapter_module} = set_ecto_adapter(opts[:database])
     pubsub_server = set_pubsub_server(mod)
 
     binding = [application_name: app,
@@ -329,7 +329,7 @@ defmodule Mix.Tasks.Phoenix.New do
     end
   end
 
-  defp set_ecto_adapter("tds"),   do: {:tds_ecto, Tds.Ecto}
+  defp set_ecto_adapter("mssql"), do: {:tds_ecto, Tds.Ecto}
   defp set_ecto_adapter("mysql"), do: {:mariaex, Ecto.Adapters.MySQL}
   defp set_ecto_adapter(_),       do: {:postgrex, Ecto.Adapters.Postgres}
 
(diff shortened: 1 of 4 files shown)


installer/lib/phoenix_new.ex
```diff
@@ -89,6 +89,10 @@ defmodule Mix.Tasks.Phoenix.New do
     * `--module` - the name of the base module in
       the generated skeleton
 
+    * `--adapter` - specify the database adapter
```
josevalim: We should support the `tds_ecto` one. Please check ecto's readme.

installer/lib/phoenix_new.ex
```diff
@@ -184,7 +188,7 @@ defmodule Mix.Tasks.Phoenix.New do
 
       # Configure your database
       config :#{binding[:application_name]}, #{binding[:application_module]}.Repo,
-        adapter: Ecto.Adapters.Postgres,
+        adapter: #{binding[:adapter_module]},
```
josevalim: You want to `inspect binding[:adapter_module]` otherwise it will be `Elixir.Ecto.Adapters.Postgres`.

installer/lib/phoenix_new.ex
andrzejsliwa: +1

Database                | Ecto Adapter           | Dependency
:---------------------- | :--------------------- | :-------------------
PostgreSQL              | Ecto.Adapters.Postgres | [postgrex][postgrex]
MySQL                   | Ecto.Adapters.MySQL    | [mariaex][mariaex]
MSSQL                   | Tds.Ecto               | [tds_ecto][tds_ecto]

scrogson: Cool :+1:

installer/lib/phoenix_new.ex
scrogson: Ah, right. Derp.

test/mix/tasks/phoenix.new_test.exs
```diff
@@ -75,6 +75,22 @@ defmodule Mix.Tasks.Phoenix.NewTest do
     end
   end
 
+  test "bootstraps generated project with adapter option" do
```
josevalim: This isn't running a new application (nor it should) so it should not be here. There are tests in the installer directory and just asserting on the generated content there should be enough.

scrogson: Cool. Thanks.

chrismccord:
We need to name it something other than `--adapter`. It's ambiguous because phoenix has Transport adapters, PubSub adapters, and Ecto Database adapters. How about `--ecto adapter` ? `--db-adapter` ?

scrogson:
@chrismccord that's really good point. `--db-adapter` is my preference.

optikfluffel:
Being able to configure the PubSub adapter for the Channels when creating a new Project would be cool as well.

chrismccord:
I don't think we need to worry about that. They are one-liners or client specific

andrzejsliwa:
@chrismccord @scrogson what about just ---database=name like in rails, then simple name in it like "mysql", "mssql", "postgres" (with mapping to correct library/driver) there is no sense to expose word "adapter" (pattern -> implementation detail) on outside.

josevalim:
:+1: for `--database`

chrismccord:
:+1

scrogson:
Updated. Thanks for the feedback folks. Good stuff!

installer/lib/phoenix_new.ex
```diff
@@ -326,6 +329,17 @@ defmodule Mix.Tasks.Phoenix.New do
     end
   end
 
+  defp set_ecto_adapter("mssql"), do: {:tds_ecto, Tds.Ecto}
+  defp set_ecto_adapter("mysql"), do: {:mariaex, Ecto.Adapters.MySQL}
+  defp set_ecto_adapter(_),       do: {:postgrex, Ecto.Adapters.Postgres}
```
josevalim: We should do `opts[:database] || "pg"` upstream and here check for "pg". If anything else is given, we should fail with `Mix.raise`.

josevalim:
Beautiful. :shipit:

Issue #789, opened 2015-04-20
Update brunch dependencies


josevalim:
We have two git dependencies and they can be updated to proper packages. We should require the latest brunch and sass-brunch though and do a local test just to be sure it works™.

https://github.com/phoenixframework/phoenix/blob/master/installer/templates/static/brunch/package.json#L5

scrogson:
Closed via https://github.com/phoenixframework/phoenix/pull/790

PR #790, opened 2015-04-21
Update brunch dependencies

ptrckbrwn:
This should solve https://github.com/phoenixframework/phoenix/issues/789. I generated a new phoenix project from the changes and made sure that brunch was working fine.

I was pretty conservative on the range of versions for brunch and brunch-sass; let me know if that needs to be changed :smile:.

Changes in `priv/static/phoenix.js` was due to an `npm install`.

First PR here, so be gentle :sweat_smile:.

### DIFF
--- a/installer/templates/static/brunch/package.json
+++ b/installer/templates/static/brunch/package.json
@@ -2,12 +2,12 @@
   "repository": {
   },
   "dependencies": {
-    "brunch": ">= 1.8 < 1.9",
+    "brunch": "^1.8.1",
     "babel-brunch": "^4.0.0",
     "clean-css-brunch": ">= 1.0 < 1.8",
     "css-brunch": ">= 1.0 < 1.8",
     "javascript-brunch": ">= 1.0 < 1.8",
-    "sass-brunch": ">= 1.8 < 1.9",
+    "sass-brunch": "^1.8.10",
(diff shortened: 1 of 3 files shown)


josevalim:
Awesome, thank you!

Does npm support something like ~> 1.18.6 and it will use only 1.18.6
ahead? If so, I would prefer that because we must use the latest of brunch
and sass-brunch.


-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Lead Developer

optikfluffel:
The `~` or `^` would be your friends here depending on how restricted you want the version handling to be. There's a [short summary at Stackoverflow](http://stackoverflow.com/a/22345808/980206) that explains the main difference.
Long story short:

* `~1.2.3` will match anything up to `1.2.n` without upgrading to `1.3.0` and above
* `^1.2.3` will match anything up to `1.n.n` without upgrading to `2.0.0` and above

For anyone interested in details I'd recommend [the blogpost 'Semver: Tilde and Caret' by Tim Oxley](https://nodesource.com/blog/semver-tilde-and-caret).

ptrckbrwn:
@josevalim @optikfluffel thank you for the feedback! I'll update accordingly.

(I think we'll want to use the `^` in this case)

PR #791, opened 2015-04-21
HTML5 inputs

ricn:
I added the remaining HTML5 input types.

### DIFF
--- a/lib/phoenix/html/form.ex
+++ b/lib/phoenix/html/form.ex
@@ -269,6 +269,15 @@ defmodule Phoenix.HTML.Form do
     generic_input(:color, form, field, opts)
   end
 
+  @doc """
+  Generates a range input.
+
+  See `text_input/3` for example and docs.
+  """
+  def range_input(form, field, opts \\ []) do
+    generic_input(:range, form, field, opts)
+  end
+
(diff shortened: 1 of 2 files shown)


lib/phoenix/html/form.ex
```diff
@@ -260,6 +260,69 @@ defmodule Phoenix.HTML.Form do
...
+    generic_input(:time, form, field, opts)
+  end
+
+  @doc """
+  Generates a datetime input.
+
+  See `text_input/3` for example and docs.
+  """
+  def datetime_input(form, field, opts \\ []) do
+    generic_input(String.to_atom("datetime-local"), form, field, opts)
```
josevalim: `:"datetime-local"` :)

josevalim:
Thank you. I will ready carefully about those and cherry pick the relevant commits in case we won't merge all.

josevalim:
I have added the range_input only. All others are still not fully supported by major browsers and require special recommendations before using it according to http://www.wufoo.com/html5/ so I don't think we should promote them yet. Thank you!

ricn:
Aha ok. I just added all of them because Rails supports them. I was looking here: http://guides.rubyonrails.org/form_helpers.html (chapter 1.4)

Issue #795, opened 2015-04-23
channels do not terminate when the transport exits normally


chrismccord:

chrismccord:
@josevalim need to pair up on this at some point

josevalim:
Closing in favor of #815.

PR #796, opened 2015-04-23
Kill channels on WS transport terminate

novaugust:
This is just proof of concept to demo a fix for #795.  Sadly, it doesn't matter what we return in the `ws_terminate` method as cowboy's just going to kill us no matter what and call it a day. So, we need to send some sort of notification to the transport's open channels to notify them they should exit. I've hacked that by just having the VM do it for me ;)


novaugust:
Closing against 3a60e42f7598d7e1e7a8dd87d85f349559aaa72c :)

PR #799, opened 2015-04-24
Test against Elixir v1.0.4

sgerrand:
:information_desk_person: This change updates Phoenix to run tests against version 1.0.4 of Elixir. 

Might also be worth running the test suite against a matrix of Erlang and Elixir versions, but I'll work on that in a follow up changeset.


chrismccord:
Thanks! As discussed, I would love to have the matrix versions when you have a moment to put it together.

PR #802, opened 2015-04-24
Clarify description of generators

### LINKED ISSUE #801: Improve description of mix phoenix.gen.html
From IRC: 

> edmz: "Generates HTML files for a resource"
> edmz: sounds like it *only* generates html files, which in my mind sounded like "views/templates"

Maybe this should be something like "Generates controller, and HTML files for a resource"?

paulcsmith:
closes #801 

Does this make sense? Should it be more specific about what it generates?

### DIFF
--- a/lib/mix/tasks/phoenix.gen.html.ex
+++ b/lib/mix/tasks/phoenix.gen.html.ex
@@ -1,7 +1,7 @@
 defmodule Mix.Tasks.Phoenix.Gen.Html do
   use Mix.Task
 
-  @shortdoc "Generates HTML files for a resource"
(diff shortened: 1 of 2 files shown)


lib/mix/tasks/phoenix.gen.html.ex
```diff
@@ -1,7 +1,7 @@
 defmodule Mix.Tasks.Phoenix.Gen.Html do
   use Mix.Task
 
-  @shortdoc "Generates HTML files for a resource"
```
paulcsmith: Or maybe, "Generates all files necessary for an HTML resource"?

scrogson:
Let's go with @edmz's changes in #803

PR #803, opened 2015-04-24
Make it more clear that the generators dont only generate views

edmz:
I started a new project and I was sure (in my mind) that there was a generator for controllers or resources.

Running `mix -h` showed:

```text
  ...
  mix new                # Create a new Elixir project
  mix phoenix.gen.html   # Generates HTML files for a resource
  mix phoenix.gen.model  # Generates an Ecto model
  ...
```

The description of the `phoenix.gen.html` made me believe that it generated only HTML files, which in my mind sounded like views/templates.

I spent sometime searching for a missing resource generator. I know that this info is in the upgrade guide, but I wasn't upgrading, so I didnt read it. Same thing could happen to a new user.

I think the description could be a little better. This PR is just one suggestion.

(I know the answer lies in checking `mix help phoenix.gen.html` but reading the current description I was sure that `phoenix.gen.html` was *not* for generating controller, model, etc code)

((feel free to ignore me as I feel rantish/stupid after realising the answer was always under my nose))

### DIFF
--- a/lib/mix/tasks/phoenix.gen.html.ex
+++ b/lib/mix/tasks/phoenix.gen.html.ex
@@ -1,7 +1,7 @@
 defmodule Mix.Tasks.Phoenix.Gen.Html do
   use Mix.Task
 
-  @shortdoc "Generates HTML files for a resource"
+  @shortdoc "Generates controller, model, views for an HTML-based resource"
 
   @moduledoc """
   Generates a Phoenix resource.
--- a/lib/mix/tasks/phoenix.gen.html.ex
+++ b/lib/mix/tasks/phoenix.gen.html.ex
@@ -1,7 +1,7 @@
 defmodule Mix.Tasks.Phoenix.Gen.Html do
   use Mix.Task
 
-  @shortdoc "Generates controller, model, views for an HTML-based resource"
+  @shortdoc "Generates controller, model and views for an HTML-based resource"
 
   @moduledoc """
   Generates a Phoenix resource.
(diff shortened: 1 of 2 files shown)


paulcsmith:
I also created a PR #802 but I like @edmz better. LMK if I should close mine

lib/mix/tasks/phoenix.gen.json.ex
```diff
@@ -1,7 +1,7 @@
 defmodule Mix.Tasks.Phoenix.Gen.Json do
   use Mix.Task
-
-  @shortdoc "Generates JSON files for a resource"
+  
+  @shortdoc "Generates controller, model for an JSON-based resource"
```
scrogson: Maybe this should be "controller and model..."

lib/mix/tasks/phoenix.gen.html.ex
```diff
@@ -1,7 +1,7 @@
 defmodule Mix.Tasks.Phoenix.Gen.Html do
   use Mix.Task
 
-  @shortdoc "Generates HTML files for a resource"
+  @shortdoc "Generates controller, model, views for an HTML-based resource"
```
scrogson: "controller, model, and views..."

edmz: I did it on purpose because I tried to not have a long sentence, as most of
the @shordocs of other tasks are quite short.

But you are right.

edmz:
I added missing 'and's.

chrismccord:
thanks!

Issue #741, opened 2015-04-01
Add a channel generator


josevalim:

gjaldon:
Hi @josevalim! I'd like to take this on if no one else has worked on it yet.

josevalim:
please go ahead :)

gjaldon:
Is this channel generator going to generate Channel files for a resource similar to `phoenix.gen.json` and `phoenix.gen.html`? Where instead of a controller we generate a channel which handles incoming CRUD events?

josevalim:
The channel will generate just the channel (no model for it).

PR #816, opened 2015-04-26
Channel generator

gjaldon:
refs: https://github.com/phoenixframework/phoenix/issues/741

This channel generator generates the a channel and channel_test file. The channel generated is very basic. It defines two `join/3` functions that pattern-match on either "topic:lobby" or "topic:" <> subtopic_id and a `handle_in/3` for every event arg. The `handle_in/3` just broadcasts the event with the attrs.

The channel_test here makes use of the helper in the PR here https://github.com/phoenixframework/phoenix/pull/652. Will make changes to it when necessary.

Let me know if there are any changes that need done. Thanks in advance! :)

### DIFF
--- a/installer/templates/new/test/support/channel_case.ex
+++ b/installer/templates/new/test/support/channel_case.ex
@@ -1,0 +1,46 @@
+defmodule <%= application_module %>.ChannelCase do
+  @moduledoc """
+  This module defines the test case to be used by
+  channel tests.
+
+  Such tests rely on `Phoenix.ChannelTest` and also
+  imports other functionality to make it easier
+  to build and query models.
+
+  Finally, if the test case interacts with the database,
+  it cannot be async. For this reason, every test runs
+  inside a transaction which is reset at the beginning
+  of the test unless the test case is marked as async.
+  """
+
+  use ExUnit.CaseTemplate
+
+  using do
+    quote do
+      # Import conveniences for testing with channels
+      use Phoenix.ChannelTest
+<%= if ecto do %>
+      # Alias the data repository and import query/model functions
+      alias <%= application_module %>.Repo
+      import Ecto.Model
+      import Ecto.Query, only: [from: 2]
+<% end %>
+
+      # The default endpoint for testing
+      @endpoint <%= application_module %>.Endpoint
+    end
+  end
+
+  setup_all do
+    @endpoint.start_link()
+    :ok
+  end
+
+  setup tags do
+<%= if ecto do %>    unless tags[:async] do
+      Ecto.Adapters.SQL.restart_test_transaction(<%= application_module %>.Repo, [])
+    end
+<% end %>
+    :ok
+  end
+end
--- a/lib/mix/tasks/phoenix.gen.channel.ex
+++ b/lib/mix/tasks/phoenix.gen.channel.ex
@@ -1,0 +1,54 @@
+defmodule Mix.Tasks.Phoenix.Gen.Channel do
+  use Mix.Task
+
+  @shortdoc "Generates a Phoenix channel"
+
+  @moduledoc """
+  Generates a Phoenix channel.
+
+      mix phoenix.gen.channel Room rooms new_msg
+
+  The first argument is the module name for the channel.
+  The second argument is the plural used as the topic.
+  The rest of the other arguments will be treated as events.
+
+  """
+  def run([singular, plural|events] = args) do
+    if String.contains?(plural, ":"), do: raise_with_help
+    Mix.Task.run "phoenix.gen.model", args
+
+    binding = Mix.Phoenix.inflect(singular)
+    path    = binding[:path]
+
+    binding = binding ++ [plural: plural, events: events]
+
+    Mix.Phoenix.copy_from source_dir, "", binding, [
+      {:eex, "channel.ex",       "web/channels/#{path}_channel.ex"},
+      {:eex, "channel_test.exs", "test/channels/#{path}_channel_test.exs"},
+    ]
+
+    Mix.shell.info """
+
+    Add the channel to the proper scope in web/router.ex:
+
+        channel "#{plural}:*", #{binding[:scoped]}Channel
+    """
+  end
+
+  def run(_) do
+    raise_with_help
+  end
+
+  defp raise_with_help do
+    Mix.raise """
+    mix phoenix.gen.channel expects both the module name and topic name
+    followed by any number of event names:
+
+        mix phoenix.gen.channel Room rooms new_msg
+    """
+  end
+
+  defp source_dir do
+    Application.app_dir(:phoenix, "priv/templates/channel")
+  end
+end
--- a/lib/mix/tasks/phoenix.gen.channel.ex
+++ b/lib/mix/tasks/phoenix.gen.channel.ex
@@ -6,21 +6,24 @@ defmodule Mix.Tasks.Phoenix.Gen.Channel do
   @moduledoc """
   Generates a Phoenix channel.
 
-      mix phoenix.gen.channel Room rooms new_msg
+      mix phoenix.gen.channel Room rooms
 
   The first argument is the module name for the channel.
   The second argument is the plural used as the topic.
-  The rest of the other arguments will be treated as events.
+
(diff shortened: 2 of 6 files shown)


priv/templates/channel/channel.ex
```diff
@@ -0,0 +1,18 @@
+defmodule <%= module %>Channel do
+  use <%= base %>.Web, :channel
+
+  def join("<%= plural %>:lobby", _auth_message, socket) do
```
chrismccord: We should just have a single join with the _id of the module name, i.e.: `join("users:" <> user_id` ? Lobby is chat specific. Make sense?

priv/templates/channel/channel.ex
```diff
@@ -0,0 +1,18 @@
...
+    {:ok, socket}
+  end
+
+  def join("<%= plural %>:" <> _<%= singular %>_id, _auth_message, socket) do
+    {:ok, socket}
+  end
+
+  <%= for event <- events do %>
+    def handle_in("<%= event %>", attrs, socket) do
+      broadcast! socket, event, attrs
```
chrismccord: This is tricky because not all events are going to be broadcasted. I'm not sure what the best default body would be here. We have three different scenarios. Sync reply, async push, broadcast + combination of first two

priv/templates/channel/channel.ex
gjaldon: Ok will remove the join with the "lobby" subtopic and just the one with the id.

priv/templates/channel/channel.ex
gjaldon: How 'bout if we have channel generators for specific use cases? Like maybe have `mix.phoenix.channel.gen.crud User users Post` where it would generate a `UserChannel` module that joins `users` topic  and handles CRUD events like `create_post`, `delete_post`, `show_post`, `show_posts` and `update_post`. Its create_post handler will look like the sample in the docs:
```elixir
def handle_in("create_post", attrs, socket) do
  changeset = Post.changeset(%Post{}, attrs)

  if changeset.valid? do
    Repo.insert(changeset)
    {:reply, {:ok, changeset}, socket}
  else
    {:reply, {:error, changeset}, socket}
  end
end
```

Not yet sure how the generic channel-generator will create `handle_in` functions for events passed.

priv/templates/channel/channel.ex
josevalim: What about "<%= plural %>:join"?

chrismccord: We usually scope to resource id, or some 'top level' thing, i.e. "lobby", including ":join" as part of the topic doesn't make sense?

josevalim:
@chrismccord can you think about a minimal channel example you would like to generate? I don't think we should worry about ids, let's generate the simplest channel we can. In my opinion, it would be nice to have one handle_in with reply, another with broadcast, and handle_out explicitly defined. Something like:

```elixir
def join("rooms:lobby", payload, socket) do
  if authorized?(payload) do
    {:ok, socket}
  else
    {:error, %{reason: "unauthorized"}}
  end
end

# Channels can be used in a request/response fashion
# by sending replies to requests from the client
def handle_in("ping", payload, socket) do
  {:reply, {"pong", payload}, socket}
end

# It is also common to receive messages from the client and
# broadcast to everyone in the current topic (rooms:lobby).
def handle_in("shout", payload, socket) do
  broadcast socket, "shout", payload
  {:noreply, socket}
end

# This is invoked every time a notification is being broadcast
# to the client. The default implementation is just to push it
# downstream but one could filter or change the event.
def handle_out(event, payload, socket) do
  push socket, event, payload
  {:noreply, socket}
end

defp authorized?(_payload) do
  true
end
```

Perhaps it is a bit verbose but channel concepts are quite new and this gives a nice walkthrough. It would also help us push some good amount of tests out of the box too. Thoughts?

chrismccord:
I really like this! I would make just one change to join/3. @josevalim I know you think ids shouldn't be focused on, but I want to make it clear that channels can and should be used for multiple, dynamic topics, so I would do something like:

```elixir
def join("rooms:lobby", payload, socket) do
  if authorized?(payload) do
    {:ok, socket}
  end
end
def join("rooms:" <> _priv_room_id, _payload, _socket) do
    {:error, %{reason: "unauthorized"}}
end
```

> Perhaps it is a bit verbose but channel concepts are quite new and this gives a nice walkthrough. It would also help us push some good amount of tests out of the box too.

Agree completely. It's worth the verbosity for the good instruction and tests.

josevalim:
@chrismccord the issue with the suggestion above is that we would also need to change the router or generate two routes which is confusing. :(

chrismccord:
> @chrismccord the issue with the suggestion above is that we would also need to change the router or generate two routes which is confusing. :(

Two routes? the single topic "rooms:*" would work.

josevalim:
Sorry, that is what I mean. We will be teaching the generic route by default, I would like to do the opposite. :(

chrismccord:
> Sorry, that is what I meant. We will be teaching the generic route by default, I would like to do the opposite. :(

Let's go with @josevalim's approach . The channel guides can pick up where the default fixed topic leaves off

gjaldon:
Ok, copy that!

gjaldon:
@chrismccord @josevalim this should be good for review. The failing test in the CI build doesn't seem to be related with this PR.

lib/mix/tasks/phoenix.gen.channel.ex
```diff
@@ -0,0 +1,60 @@
...
+    Mix.Phoenix.copy_from source_dir, "", binding, [
+      {:eex, "channel.ex",       "web/channels/#{path}_channel.ex"},
+      {:eex, "channel_test.exs", "test/channels/#{path}_channel_test.exs"},
+    ]
+
+    Mix.shell.info """
+
+    Add the channel to the proper scope in web/router.ex:
+
+        channel "#{plural}:*", #{binding[:scoped]}Channel
```
chrismccord: I believe this should be `channel "#{plural}:lobby", #{binding[:scoped]}Channel` per @josevalim

gjaldon: Just updated this PR with this change @chrismccord @josevalim.

gjaldon:
@josevalim @chrismccord added a `Phoenix.ChannelCase` here and used that in the generated channel tests.

josevalim:
Perfect, I want to get this in ASAP. :) We are ready!

josevalim:
I am adding some comments but don't worry about them, I will tackle them right after merging it!

installer/templates/new/test/support/channel_case.ex
```diff
@@ -0,0 +1,46 @@
...
+      import Ecto.Model
+      import Ecto.Query, only: [from: 2]
+<% end %>
+
+      # The default endpoint for testing
+      @endpoint <%= application_module %>.Endpoint
+    end
+  end
+
+  setup_all do
```
josevalim: This should not be required. We need in Phoenix tests but it should be started in apps.

josevalim:

gjaldon:
Awesome :+1:

priv/templates/channel/channel_test.exs
```diff
@@ -0,0 +1,24 @@
+defmodule <%= module %>ChannelTest do
+  use <%= base %>.ChannelCase
+
+  alias <%= module %>
```
josevalim: Missing `.Channel` in here.

priv/templates/channel/channel_test.exs
```diff
@@ -0,0 +1,24 @@
...
+    {:ok, _, socket} = join(<%= scoped %>Channel, "<%= plural %>:lobby")
+
+    ref = push socket, "ping", %{"hello" => "there"}
+    assert_reply ref, :pong, %{"hello" => "there"}
+  end
+
+  test "shout broadcasts to <%= plural %>:lobby" do
+    {:ok, _, socket} = subscribe_and_join(<%= scoped %>Channel, "<%= plural %>:lobby")
+
+    push socket, "broadcast", %{"foo" => "bar"}
```
josevalim: The event here is "shout".

PR #819, opened 2015-04-26
Track explicit render/2 clauses

chrismccord:
@josevalim this fixes the issue we had with not tracking explicit render/2 definitions for the `render_existing`. This also has the nice side-effect of including all defined render/2's from both compiled templates and regular functions. This is the first legitimate use I've had for `@on_definition`, so I'm PR'ing to make sure it's sane. Very fun though!

### DIFF
--- a/installer/templates/new/web/web.ex
+++ b/installer/templates/new/web/web.ex
@@ -47,7 +47,7 @@ defmodule <%= application_module %>.Web do
       use Phoenix.View, root: "web/templates"
 
       # Import convenience functions from controllers
-      import Phoenix.Controller, only: [get_flash: 2]
+      import Phoenix.Controller, only: [get_flash: 2, view_module: 1]
 
       # Import URL helpers from the router
       import <%= application_module %>.Router.Helpers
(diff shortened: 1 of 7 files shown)


josevalim:
Unfortunately this does not solve the issue too. Someone could write:

    def render("show." <> format, assigns) when format in ~w(json xml) do
    end

I am not sure how we could fix this. The best idea I had so far is that `render_existing` could set `render_existing: true` in assigns and the template_not_found function would not raise if this key exists. The problem is that, since this key will be in assigns, how to avoid it being passed to upcoming calls to render with `assigns`? We can do it for templates... but I have no idea on how to do it for the user based clauses.

josevalim:
One better idea is to set `render_existing: {view_module, template_name, assigns}` and we would not raise it only if they match, so it would reduce the chance of us invoking it when we should not.

josevalim:
Another idea is to use `@on_definition` as you used and rewrite all functions call from  `render(foo, bar) when baz, do: bat` to `render?(foo, bar) when bar, do: true`, i.e. we keep everything, except the body, which now just returns true. And then, if there is a clause just with variables, we need to rewrite it to false. It is hackish but it may work too.

Issue #699, opened 2015-03-13
Token Authorization for Channels and API's


jeregrine:
We need to figure out a standard way for authentication/authorization in Channels and API's. Our suggestion today is "use tokens" but users are going to implement custom and thus broken solutions. This issue is to scope out our options and what we want to do. 

I've implemented a couple and also looked into what third parties are also using. I'll lay them out with pros/cons here with the Example: 
- User gets the auth_token from the session or user routes.
- User can use the auth_token for channel authentication.
- User can use the auth_token for `Authentication: Bearer auth_token` headers for api requests. Allowing them to circumvent cor's cookie restrictions.

What we need to decide on is, how to generate.
### Plug Tokens
https://gist.github.com/jeregrine/beba5dd6f83a9e07254d#file-channel_auth-ex

I'm not sure thats super idiomatic elixir as I am generating an encryptor and everything everytime but it works okay and shows the point.

*Pros*
- No extra dependencies. We already have, plug works.
- Same token generation scheme as cookies. 
- If we wanted to store more we'd have to serialize it into a string. 

*Cons* 
- Not interoperable with other token generation services (maybe rails?)
- Doesn't handle expiration of tokens, once generated the token is valid until the server secrets change. Browser cookies handled this for us. 

### JWT Tokens using [joken](https://github.com/bryanjos/joken)
https://gist.github.com/jeregrine/beba5dd6f83a9e07254d#file-joken_auth-ex

Again this file is kind of messy but I just kinda tossed it together. 

*Pros*
- Commonly used [token](http://jwt.io/), interoperable with third parties using same secret. 
- Handles simple expiration and other complex expiration methods if required.
- Can store more than just strings as its a map.

*Cons* 
- Has dependency on JSX (working on changing over https://github.com/bryanjos/joken/issues/14)
- Has dependency on Timex which he uses to generate timestamps. 
- My example also uses Timex
- Another dependency to add to apps.

### Storing generated keys in the Database

I haven't personally implemented this one but Hex.pm currently uses this https://github.com/hexpm/hex_web/blob/master/lib/hex_web/api/key.ex 

I cannot really speak to the pros and cons of this approach over just generating bearer tokens. Would be helpful if @ericmj could chime in on this approach.

Again this is an issue to get the discussion going. Let me know what your opinions are, I am exploring this as we speak for my projects and it can benefit the entire Phoenix Project as well.

jeregrine:
[...]

PR #820, opened 2015-04-26
Token Auth

jeregrine:
I tried to follow the API from #699 I need some suggestions on testing this, since configuration all goes through the endpoint. I'm assuming the best way is to just "integration" test this. 

Also have some open questions:

1) Should we include expiration? 
2) Is this enough? 

cc @chrismccord @josevalim

### DIFF
--- a/lib/phoenix/token.ex
+++ b/lib/phoenix/token.ex
@@ -58,11 +58,12 @@ defmodule Phoenix.Token do
   @doc """
   Encrypts your data into a token you can send down to clients
   """
+  @spec gen_token(Plug.Conn.t | Phoenix.Socket.t, term, List) :: String.t | no_return
   def gen_token(context, data, opts \\ []) do
     {secret, sign_secret, max_age, encoder} = get_endpoint(context) |> encryptor()
 
-    if Dict.has_key?(opts, :max_age) do
-      max_age = opts[:max_age]
+    max_age = if Dict.has_key?(opts, :max_age) do
+      opts[:max_age]
     end
 
     if max_age do
@@ -81,12 +82,13 @@ defmodule Phoenix.Token do
   @doc """
   Decrypts the token into the originaly present data.
   """
+  @spec verify_token(Plug.Conn.t | Phoenix.Socket.t, String.t) :: :error | :token_expired | term | no_return
   def verify_token(context, token) do
-    {secret, sign_secret, max_age, encoder} = get_endpoint(context) |> encryptor()
+    {secret, sign_secret, _max_age, encoder} = get_endpoint(context) |> encryptor()
     case MessageEncryptor.verify_and_decrypt(token, secret, sign_secret) do
       :error -> :error
       {:ok, message} -> 
-        %{ "data" => data, "exp" => exp} = Poison.decode!(message)
+        %{ "data" => data, "exp" => exp} = encoder.decode!(message)
         if exp < now_ms() do
           :token_expired
         else
--- a/lib/phoenix/token.ex
+++ b/lib/phoenix/token.ex
@@ -26,8 +26,8 @@ defmodule Phoenix.Token do
           user_id ->
             socket = assigns(socket, :user, Repo.get!(User, user_id))
             reply socket, "join", %{}
-          :expired -> :ignore
-          :error -> :ignore
+          {:error, :expired} -> :ignore
+          {:error, :invalid} -> :ignore
         end
       end
     end
@@ -59,21 +59,11 @@ defmodule Phoenix.Token do
   Signs your data into a token you can send down to clients
   """
   def sign_token(context, salt,  data, opts \\ []) when is_binary(data) or is_integer(data) or is_map(data) do
-    secret = get_endpoint(context) |> get_secret(salt)
-
-    max_age = if Dict.has_key?(opts, :max_age) do
-      opts[:max_age]
-    end
-
-    if max_age do
-      exp = now_ms() + max_age
-    else
-      exp = nil
-    end
+    secret = get_endpoint(context) |> get_secret(salt, opts)
 
     message = %{
       data: data,
-      exp: exp
+      signed: now_ms()
     } |> :erlang.term_to_binary()
     MessageVerifier.sign(message, secret)
   end
@@ -81,14 +71,19 @@ defmodule Phoenix.Token do
   @doc """
   Decrypts the token into the originaly present data.
   """
-  def verify_token(context, salt, token) do
-    secret = get_endpoint(context) |> get_secret(salt)
+  def verify_token(context, salt, token, opts \\ []) do
+    secret = get_endpoint(context) |> get_secret(salt, opts)
     case MessageVerifier.verify(token, secret) do
-      :error -> :error
+      :error -> {:error, :invalid}
       {:ok, message} ->
-        %{data: data, exp: exp} = :erlang.binary_to_term(message)
-        if exp && exp < now_ms() do
-          :expired
+        %{data: data, signed: signed} = :erlang.binary_to_term(message)
+
+        max_age = if Dict.has_key?(opts, :max_age) do
+          opts[:max_age]
+        end
+
+        if max_age && (signed + max_age) < now_ms() do
+          {:error, :expired}
         else
           data
         end
@@ -99,9 +94,16 @@ defmodule Phoenix.Token do
   defp get_endpoint(%Phoenix.Socket{} = socket), do: socket.endpoint
 
   # Gathers configuration and generates the key secrets and signing secrets.
-  defp get_secret(endpoint, salt) do
+  defp get_secret(endpoint, salt, opts \\ []) do
     secret_key_base = endpoint.config(:secret_key_base)
-    KeyGenerator.generate(secret_key_base, salt)
+    iterations = Keyword.get(opts, :key_iterations, 1000)
+    length = Keyword.get(opts, :key_length, 32)
+    digest = Keyword.get(opts, :key_digest, :sha256)
+    key_opts = [iterations: iterations,
+                length: length,
+                digest: digest,
+                cache: Plug.Keys]
+    KeyGenerator.generate(secret_key_base, salt, key_opts)
   end
 
   defp time_to_ms({mega, sec, micro}),
--- a/lib/phoenix/token.ex
+++ b/lib/phoenix/token.ex
@@ -14,26 +14,36 @@ defmodule Phoenix.Token do
   alias Plug.Crypto.MessageEncryptor
 
   @doc """
-    Encrypts your data into a token you can send down to clients
+  Encrypts your data into a token you can send down to clients
   """
   def gen_token(context, data) do
-    {secret, sign_secret} = get_endpoint_for(context) |> encryptor()
-    MessageEncryptor.encrypt_and_sign(data, secret, sign_secret)
+    {secret, sign_secret, max_age, encoder} = get_endpoint(context) |> encryptor()
+    message = %{ 
+      data: data,
+      exp: now_ms() + max_age
+    } |> encoder.encode!()
+    MessageEncryptor.encrypt_and_sign(message, secret, sign_secret)
   end
 
   @doc """
-    Decrypts the token into the originaly present data.
+  Decrypts the token into the originaly present data.
   """
   def verify_token(context, token) do
-    {secret, sign_secret} = get_endpoint_for(context) |> encryptor()
+    {secret, sign_secret, max_age, encoder} = get_endpoint(context) |> encryptor()
     case MessageEncryptor.verify_and_decrypt(token, secret, sign_secret) do
       :error -> :error
-      {:ok, data} -> 
+      {:ok, message} -> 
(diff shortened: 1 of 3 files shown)


test/phoenix/token_text.exs
```diff
@@ -0,0 +1,3 @@
+defmodule Phoenix.Test.Token do
```
scrogson: I think this should be `Phoenix.TokenTest`

lib/phoenix/token.ex
```diff
@@ -0,0 +1,49 @@
...
+  """
+  def verify_token(context, token) do
+    {secret, sign_secret} = get_endpoint_for(context) |> encryptor()
+    case MessageEncryptor.verify_and_decrypt(token, secret, sign_secret) do
+      :error -> :error
+      {:ok, data} -> 
+    end
+  end
+
+  defp get_endpoint_for(conn = %Plug.Conn{}), do: Phoenix.Controller.endpoint_module(conn)
```
scrogson: I think the convention is `%Plug.Conn{} = conn`

lib/phoenix/token.ex
```diff
@@ -0,0 +1,49 @@
...
+  @moduledoc """
+  Will generate, verify, and decrypt tokens for use in channel or API 
+  authentication. Typically you will want to store the basic information
+  you need to authorize a user. Typically as simple as a user id in a 
+  database. 
+
+      iex> user_id = 1
+      iex> token = encrypt(user_id, endpoint)
+      iex> user_id == decrypt(token, endpoint)
+      true
```
chrismccord: Let's get some high-level examples of usage in the moduledoc. Maybe even showing using in a `def join(...`. We also need to add examples to the `gen_token` and `verify_token` with their return values.

lib/phoenix/token.ex
```diff
@@ -0,0 +1,49 @@
...
+  """
+  alias Plug.Crypto.KeyGenerator
+  alias Plug.Crypto.MessageEncryptor
+
+  @doc """
+    Encrypts your data into a token you can send down to clients
```
chrismccord: Begin nitpick-mode: Add examples and don't indent :)

lib/phoenix/token.ex
```diff
@@ -0,0 +1,49 @@
...
+  end
+
+  @doc """
+    Decrypts the token into the originaly present data.
+  """
+  def verify_token(context, token) do
+    {secret, sign_secret} = get_endpoint_for(context) |> encryptor()
+    case MessageEncryptor.verify_and_decrypt(token, secret, sign_secret) do
+      :error -> :error
+      {:ok, data} ->
```
chrismccord: nothing is returned in this clause?

lib/phoenix/token.ex
chrismccord: I also personally am moving away from `_for`, even though my Ruby code is littered with it. `get_endpoint` is just fine

chrismccord:
@jeregrine Looks nice! Better docs and tests and we should be close:

> 1) Should we include expiration?

Yes

> 2) Is this enough?

I think so, but better moduledocs will let us know by showing a real example/scenario.

For the endpoint testing, you can define an endpoint, with configuration, inline in the test, like we do here:
https://github.com/phoenixframework/phoenix/blob/master/test/phoenix/integration/endpoint_test.exs#L12-L15

jeregrine:
@chrismccord @scrogson thanks for feedback/suggestions.

I added a expiration date and cleared out the nitpicky stuff. Will follow up with much better docs and tests tomorrow.

test/phoenix/token_text.exs
```diff
@@ -0,0 +1,52 @@
...
+    {:ok, pid} = TokenEndpoint.start_link
+    on_exit(fn ->
+      Process.unlink(pid)
+      Process.exit(pid, :shutdown)
+    end)
+  end
+
+  test "happy path for the encoder" do
+    id = 1
+    token = Token.gen_token(conn(), 1)
```
pel-daniel: Should the second parameter be `id` instead of 1 ??

chrismccord:
Your changes and docs look great! I think we should have a way to optionally set max_age/expire on the `gen_token` call. I think defaulting to the max_age config like you have now should stay, but I should be able to `get_token(socket, %{user_id: user_id}, max_age: ...)`. We could also make the `:max_age` key part of the data payload itself. Thoughts?

paulcsmith:
One other thought, is there a way to manually expire/revoke a token? If so, maybe an example would be good to include?

jeregrine:
I will add the option to set max_age in gen_token and to check if its nil
before using it it.

Paul these tokens are more akin to cookies. If the user wanted to revoke
access they would need to define their own logic around it. Usually in the
database. Ex

UserTable 1 - 1 AuthKeyTable{user_id, id, valid}

id = user.auth_key.id
gen_token(socket, id)

If the user needed ability to revoke access.

paulcsmith:
@jeregrine Gotcha! I misunderstood. Thanks for the example :+1:

chrismccord:
@jeregrine : so my only concern with this api is if people encrypt an explicit map, they might accidentally include `max_age:` as a payload key, instead of the optional third arg. One option is to only allow maps to be used as the encrypted payload, with an optional `max_age:` key in the map. Is there any drawback to this? I think what we have is fine, but it's a thought that just crossed my mind.

jeregrine:
I don't know if its a huge issue IMO. I could make max_age just be a third
optional param?

lib/phoenix/token.ex
```diff
@@ -0,0 +1,120 @@
...
+
+    defmodule MyChannel do
+      def join("my:" <> id = topic, %{token: token}, socket) do
+        case verify_token(socket, token) do
+        user_id ->
+          socket = assigns(socket, :user, Repo.get!(User, user_id))
+          reply socket, "join", %{}
+        :expired -> {:error, :auth_expired, socket}                                                                                                                                                           
+        :error -> {:error, :bad_auth, socket}
+        end
```
chrismccord: indentation needs fixed here, and the return values will just be `:ignore`

lib/phoenix/token.ex
```diff
@@ -0,0 +1,120 @@
...
+    def create(conn, params) do
+      user = User.create(params)
+      render conn, "user.json", %{token: gen_token(conn, user.id), user: user}
+    end
+
+    Then the client can use it all the way through. If you'd like to 
+    use the same token for API authorization then I suggest adding it
+    as a header to your authroized http requests
+    
+      Authorization: Bearer %{token}
```
chrismccord: Is this supposed to be `Authorization: Bearer #{token}` ?

test/phoenix/token_text.exs
```diff
@@ -0,0 +1,67 @@
...
+  end
+
+  test "given a junk token it fails" do
+    assert :error == Token.verify_token(conn(), "garbage")
+  end
+
+  test "bad expiration it fails" do
+    token = Token.gen_token(conn(), 1)
+    Stream.timer(60) |> Enum.map(fn (_) ->
+      assert :token_expired == Token.verify_token(conn(), token)
```
chrismccord: `:expired` as the return value is enough

lib/phoenix/token.ex
```diff
@@ -0,0 +1,120 @@
...
+    end
+  end
+
+  defp get_endpoint(%Plug.Conn{} = conn), do: Phoenix.Controller.endpoint_module(conn)
+  defp get_endpoint(%Phoenix.Socket{} = socket), do: socket.endpoint
+
+  # Gathers configuration and generates the key secrets and signing secrets.
+  defp encryptor(endpoint) do
+    config = endpoint.config(:token_auth)
+    secret_key_base = Dict.get(config, :secret_key_base)
```
josevalim: We should use the same secret_key_base that we use for the connection here: `endpoint.config(:secret_key_base)`

lib/phoenix/token.ex
```diff
@@ -0,0 +1,120 @@
...
+      Application.get_env(:phoenix, :format_encoders)
+      |> Keyword.get(:json, Poison)
+
+    secret = KeyGenerator.generate(secret_key_base, encryption_salt)
+    sign_secret = KeyGenerator.generate(secret_key_base, signing_salt)
+    {secret, sign_secret, max_age_in_ms, encoder}
+  end
+
+  defp time_to_ms({mega, sec, micro}),
+    do: ((((mega * 1000000) + sec) * 1000000) + micro) / 1000 |> trunc()
```
josevalim: Use `div`?

test/phoenix/token_text.exs
```diff
@@ -0,0 +1,67 @@
+defmodule Phoenix.Test.TokenTest do
```
josevalim: `Phoenix.TokenTest`

lib/phoenix/token.ex
```diff
@@ -0,0 +1,120 @@
...
+        end
+    end
+  end
+
+  defp get_endpoint(%Plug.Conn{} = conn), do: Phoenix.Controller.endpoint_module(conn)
+  defp get_endpoint(%Phoenix.Socket{} = socket), do: socket.endpoint
+
+  # Gathers configuration and generates the key secrets and signing secrets.
+  defp encryptor(endpoint) do
+    config = endpoint.config(:token_auth)
```
josevalim: Actually, since the salt must be explicit, I would get rid of this configuration. Just ask everything explicitly in `opts` and we ship with reasonable defaults: https://github.com/elixir-lang/plug/blob/master/lib/plug/session/cookie.ex#L60-L66

lib/phoenix/token.ex
jeregrine: Got this from a different area in phoenix :-|

jeregrine:
@chrismccord @josevalim is this closer?

lib/phoenix/token.ex
```diff
@@ -0,0 +1,110 @@
...
+  alias Plug.Crypto.MessageVerifier
+
+  @doc """
+  Signs your data into a token you can send down to clients
+  """
+  def sign_token(context, salt,  data, opts \\ []) when is_binary(data) or is_integer(data) or is_map(data) do
+    secret = get_endpoint(context) |> get_secret(salt)
+
+    max_age = if Dict.has_key?(opts, :max_age) do
+      opts[:max_age]
```
josevalim: So, I wouldn't use max_age when signing. Simply store the timestamp on sign. I would use the max_age when verifying. This is because if you accidentally set the max_age high, you can still change the verification and everything will be fine.

josevalim:
@jeregrine  can we please add these defaults? https://github.com/elixir-lang/plug/blob/master/lib/plug/session/cookie.ex#L60-L66

Also, let's make it return: `{:ok, token} | {:error, :invalid} | {:error, :expired}` because it is much easier to pattern match on. The current API forces us to first explicitly match on :invalid and :expired before I can match on the token.

jeregrine:
@josevalim I think I am going to pass through straight to the KeyGenerator since it has the same opts. I'll just need to document it.

josevalim:
@jeregrine does it have the cache option too? I don't think it does?

jeregrine:
@josevalim oops meant to link to this https://github.com/elixir-lang/plug/blob/master/lib/plug/crypto/key_generator.ex#L27-L30 <3

josevalim:
Yeah, we need to pass the cache optino only because it has no default.

jeregrine:
@josevalim oh. Okay will do

Akkuma:
Is there a reason the extra security from encrypt and sign was removed for only signing? If there is no encryption what advantages does this have over using a JWT implementation or even using https://github.com/bryanjos/joken

jeregrine:
Our suggestion is to 1) always use SSL and 2) this doesn't encrypt but it
does sign the token, making it impossible to tamper with or generate
outside of the context of our server.

If you need something more complex or different I suggest you do look into
Joken/JWT. I use it on other projects as well. This is just a general
implementation that will be useful 9/10 times.

lib/phoenix/token.ex
```diff
@@ -0,0 +1,128 @@
...
+  communication you verify the token, query the database and authorize
+  the user.
+
+  When using it with a socket a typical example might be:
+
+    defmodule MyChannel do
+      def join("my:" <> id = topic, %{token: token}, socket) do
+        case verify_token(socket, "user", token) do
+          user_id ->
+            socket = assigns(socket, :user, Repo.get!(User, user_id))
```
chrismccord: let's make sure we show a query/check where we authorize the user to access `id`

williamgueiros:
ping

chrismccord:
@williamgueiros We're still working on this. It will be out in the next release

SebastianSzturo:
Any updates on this?

chrismccord:
@SebastianSzturo Coming soon. We have some reviews to go through still and don't want to rush security related features.

josevalim:
Sorry for taking so long. This looks perfect. The docs are outdated because we added Phoenix.Socket but I will fix it as I was the one blocking this all this time.

jeregrine:
💗💖💜💕🔥

williamgueiros:
:+1:

Issue #806, opened 2015-04-24
Brunch: "SyntaxError: Unexpected end of input"


endel:
I've installed pixi.js via bower and it worked. But as I import my source files on `web/static/js/app.js` (using ES6 import) the following error pops out randomly: 

```
Uncaught SyntaxError: Unexpected end of input
(index):16 Uncaught ReferenceError: require is not defined(anonymous function) @ (index):16
```

**Screenshot:**

![screen shot 2015-04-24 at 4 32 20 pm](https://cloud.githubusercontent.com/assets/130494/7327011/2da23750-eaa0-11e4-85a9-0402e8525361.png)

Note that there is no syntax errors in my sources, since they work just fine using browserify.

I'm using phoenix v0.11 on OSX, and node v0.12.0.

scrogson:
I've also had weird issues between `brunch` and `bower`. Would need to know a bit more of your setup to determine the problem.

chrismccord:
Can you provide the contents of `web/static/js/app.js` ?

endel:
After removing files, and blocks of code, I've found the (scary) problem.

When you have a JavaScript block without implementation, just with comments, the "unexpected end of input" errors starts to appear.

```javascript
if (true) {
  // this comment breaks everything
}
```

It seems to be a problem with [babel-brunch](https://github.com/babel/babel-brunch), which I've created an issue here: https://github.com/babel/babel-brunch/issues/10

josevalim:
Than you for finding the source and reporting the bug! Awesome work!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

josevalim:
I am closing this but if you don't agree, let us know and we will gladly reopen.

PR #822, opened 2015-04-27
upgrade babel-brunch dependency on package.json

endel:
This patch fixes the issue previously reported here: https://github.com/phoenixframework/phoenix/issues/806


chrismccord:
Thanks!

Issue #826, opened 2015-04-28
Change default DB username/password when generating app with "--database mysql"


chvanikoff:
Currently it defaults to "postgres":"postgres" and I think it should be "root":"" for mysql. I can send a patch if it's ok.

chrismccord:
PR please

PR #828, opened 2015-04-28
different credentials in generated app for different databases

chvanikoff:
patch for #826 
besides default mysql username/password, MSSQL credentials updated as well to be different. Also updated MixHelper.assert_file/2 function to allow list of regular expressions to be matched.


josevalim:

PR #831, opened 2015-04-29
Separate bootstrap.css into it's own file

mmrobins:
Really long lines make vim slow, and having the minified bootstrap css
in the same file as the rest of my app's css triggers this issue.  If
bootstrap is going to be included by default seems like it ought to be
in it's own file anyway.


josevalim:
We have discussed this in the past. The plan is not to introduce bootstrap, but introduce some sort of default layout. It just happened that we used bootstrap for it.

We have two options here: bite the bullet and put them apart or leave them as is. In any case, I don't think we should have two stylesheet imports in the layout.

I will leave this decision to @chrismccord. Thank you!

josevalim:
Given we already use bootstrap classes all over our generated html, I think it is fair game to split the bootstrap on its own file. Let's make it official. Removing it is just a matter of removing the css file anyway.

My only restriction is that I don't want to have many stylesheet links in the layout. That's easy with brunch, not sure about the non brunch setup.

Thoughts @scrogson @chrismccord?

chrismccord:
> My only restriction is that I don't want to have many stylesheet links in the layout. That's easy with brunch, not sure about the non brunch setup.

I only want to generate a single css link in the layout regardless of brunch or no-brunch, so I'm not sure how easy it is going to be to concat. Maybe we can just concat _bootstrap.css into app.css for no-brunch? At any rate, I'm :+1: on a separate file, even if to just stop vim and emacs from choking on the file

mmrobins:
I removed the separate import.  I can look into concatenating the files in the no-brunch case, but that does just end up with the original issue where the bootstrap css as one long line will be in the app.css file.  That was why I originally put the separate import in the layout, but realize now it didn't really do what I wanted.  

I got confused that answering `Install brunch.io dependencies? [Yn] n` wasn't the same thing as passing the `--no-brunch` when running `mix phoenix.new`, so wasn't really trying the nobrunch case, just wondering why I was getting the a node.js error when starting and an app.scss when I expected app.css

```
% mix phoenix.server
[info] Running Bar.Endpoint with Cowboy on port 4000 (http)

module.js:340
    throw err;
          ^
Error: Cannot find module '/Users/matt/tmp/bar/node_modules/brunch/bin/brunch'
    at Function.Module._resolveFilename (module.js:338:15)
    at Function.Module._load (module.js:280:25)
    at Function.Module.runMain (module.js:497:10)
    at startup (node.js:119:16)
    at node.js:901:3
```

Sounds like the better solution is to extract *just* the styling that's being used from bootstrap into app.css and get rid of the rest, so I'll probably try to do that first instead of concatenation in the no-brunch case.

josevalim:
So why don't we do this. We always generate two files.

1. For the brunch case, we include only the app.css one
2. For the non-brunch case, we explicitly include both

Thoughts?

josevalim:
Ok, after some discussions on IRC, the plan is to move away from bootstrap. I am closing this and we will hopefully have a cleaner looking CSS file before 1.0. Thank you!

asymmetric:
Is there any ETA for bootstrap removal?

chrismccord:
The stock css will remain as is until we have good reason to change it

PR #778, opened 2015-05-02
Any plans to add support for foreign keys to phoenix.gen.model?

scrogson:
I'm not sure what the syntax would look like, but It'd be cool to be able to do something like

```
mix phoenix.gen.model Foo foos ...
mix phoenix.gen.model Bar bars foo_id:references(:foos)
```


josevalim:
:shipit:

PR #838, opened 2015-05-02
Fix travis after_script invocation

rrrene:
Hi Chris,

this PR fixes the invocation of the `after_script` on Travis.

Since we switch to the `installer/`dir after `mix test` in order to invoke it again, we have to return to `TRAVIS_BUILD_DIR` for the subsequent invocation of InchEx.


chrismccord:
thanks!

PR #840, opened 2015-05-03
Add default robots.txt file to phoenix.new

dabit:
Because all web apps need a robots.txt file

### DIFF
--- a/installer/lib/phoenix_new.ex
+++ b/installer/lib/phoenix_new.ex
@@ -16,6 +16,7 @@ defmodule Mix.Tasks.Phoenix.New do
     {:eex,  "new/config/test.exs",                           "config/test.exs"},
     {:eex,  "new/lib/application_name.ex",                   "lib/application_name.ex"},
     {:eex,  "new/lib/application_name/endpoint.ex",          "lib/application_name/endpoint.ex"},
+    {:text, "new/priv/static/robots.txt",                    "priv/static/robots.txt"},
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you. The issue is that this is not going to work if we are using brunch. So with brunch, this needs to be copied to web/static/assets/robots.txt. Without brunch, the location is correct.  Can you please update your commit?

josevalim:
To be more precise, we need to update here:

https://github.com/dabit/phoenix/blob/robots/installer/lib/phoenix_new.ex#L55

and here:

https://github.com/dabit/phoenix/blob/robots/installer/lib/phoenix_new.ex#L47

Putting the file at the static/robots.txt. :)

djm:
The thing is that a robots.txt isn't an asset in the traditional sense.

For example, if you were using a CDN you'd likely be serving your assets
from a domain such as `username.example-cdn.com`. Serving the robots.txt
file at the root path there would cause the directions inside to be applied
to the CDN domain and not `your-domain.com`, which is highly likely not
what you want.

As asset hosting on a separate (sub)domain is common and indeed recommended
I'm not sure changeset this covers the use cases it needs to.

josevalim:
@djm Brunch call assets everything that is copied to "priv/static" with no changes. So regardless if it is an asset or not, that's where we need to put it.

josevalim:
I was wrong, everyone was right. :) We don't ignore priv/static by default (just css and js).

PR #852, opened 2015-05-08
Pass arguments from task phoenix.server to app.start

hayesgm:
It may be desired to pass arguments from `mix phoenix.server` to "app.start".  An example of this would be `--no-deps-check` which comes up when .git directories are wiped on a PaaS (in my case, `flynn` re: [elixir issue #1430](https://github.com/elixir-lang/elixir/issues/1430)).  This pull request passes through those arguments for the Mix task.

I'd be happy to discuss the merits of this patch.

### DIFF
--- a/lib/mix/tasks/phoenix.server.ex
+++ b/lib/mix/tasks/phoenix.server.ex
@@ -5,6 +5,16 @@ defmodule Mix.Tasks.Phoenix.Server do
 
   @moduledoc """
   Starts the application by configuring all endpoints servers to run.
+
+  ## Command line options
+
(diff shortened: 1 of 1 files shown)


josevalim:
I think this is great addition. We need, however, to document that it accepts the same options as `app.start`. Can you update the `@moduledoc` please?

hayesgm:
Just added documentation for this. Decided it was best to make a note to refer to docs for "app.start" instead of putting the flags in directly since I don't want to get out of sync with `elixir` and `phoenix` for these flags.

chrismccord:
❤️❤️❤️❤️

PR #854, opened 2015-05-09
babel-brunch option key has changed from ES6to5 to babel.

stevedomin:
Backward-compatibility is maintained but we might as well be using the new name. It's also less confusing.

Here is the commit that introduced that change: https://github.com/babel/babel-brunch/commit/fa8dd643431e99d8a4fd05ad6637204264e9f497

### DIFF
--- a/brunch-config.js
+++ b/brunch-config.js
@@ -29,7 +29,7 @@ exports.config = {
 
   // Configure your plugins
   plugins: {
-    ES6to5: {
+    babel: {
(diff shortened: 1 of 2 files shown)


chrismccord:
Thanks!

josevalim:
@stevedomin don't we need to update the package.js file then to include the new babel?

josevalim:
Update package.json in master.

stevedomin:
@josevalim good point! Sorry for this oversight.

PR #855, opened 2015-05-09
Added `:empty` to controller template to prevent deprecation warnings

crismali:


crismali:
Actually looks like #844 will take care of this.

PR #856, opened 2015-05-10
Commented example of having a separate vendor.js bundle

stevedomin:
We decided to leave the default as is but to give an example of separating the two bundles (app and vendor).


chrismccord:
<3 <3 <3

PR #857, opened 2015-05-10
Made generated controller tests assert records are created/updated

crismali:


chrismccord:
Thanks! I'll let @josevalim sign off on this one since he's heading up the testing effort

josevalim:
Thanks, I am not super sure about this. For the JSON tests, since we have the id, I would fetch something by id. For the HTML tests, it is a bit harder, although I am not sure if we should check if something exists with those parameters or if we simply check the number of entries in the database changed.

Thoughts?

crismali:
I think checking that records exist by their attributes is usually better since it ensures nothing is filtering out parameters. I think unexpected parameter filtering is a more common gotcha than accidentally creating too many records. (strong params in Rails has tripped me up in this way haha)

Plus it's the only way to ensure that an update action is working properly so it incidentally adds some consistency to the tests by making the create action test similar. 

But while if I had to choose one I'd choose the attribute tests I don't see a reason not to also assert the count in create action tests. 

Also for the JSON and update tests it might be worth it to search for the record by id and all of the provided parameters.

josevalim:
Thank you, those are good points. It is worth giving this a try. Let's wait until v0.13 is out and merge this. :)

crismali:
Sounds good :)

josevalim:
:heart:

PR #863, opened 2015-05-11
Add Phoenix.ChannelTest

josevalim:
This commit introduces many other changes on channels and the channel server. Since we are adding `Phoenix.ChannelTest`, conceptually it means we are adding a new client, which is not a transport, but the test process itself. This forced us to clean up the server API and its replies.

This pull request also moved some logic to transports. We are going to clean it up next.


lib/phoenix/channel.ex
```diff
@@ -353,7 +332,22 @@ defmodule Phoenix.Channel do
...
+  ## Examples
+
+      iex> socket.assigns[:token]
+      nil
+      iex> socket = assign(socket, :token, "bar")
+      iex> socket.assigns[:token]
+      "bar"
+
+  """
+  def assign(socket = %Socket{}, key, value) do
```
josevalim: I have moved this here only so we don't need to look at two different modules for functions. Everything is in Phoenix.Channel.

lib/phoenix/channel/server.ex
```diff
@@ -181,4 +203,40 @@ defmodule Phoenix.Channel.Server do
     got #{inspect result}
     """
   end
+
+  ## Handle replies
+
+  defp handle_reply(socket, {status, payload}, :handle_in)
```
josevalim: We were not verifying replies on `:stop`, only on `:reply`, so I created this function.

lib/phoenix/channel/transport.ex
```diff
@@ -95,28 +95,42 @@ defmodule Phoenix.Channel.Transport do
                   endpoint: endpoint,
                   pubsub_server: endpoint.__pubsub_server__(),
                   topic: msg.topic,
-                  ref: msg.ref,
                   channel: channel,
                   transport: transport}
 
-        Phoenix.Channel.Server.start_link(socket, msg.payload)
+        case Phoenix.Channel.Server.join(socket, msg.payload) do
```
josevalim: We will need to change this for long polling so we return the replies on the same request.

lib/phoenix/socket.ex
```diff
@@ -13,13 +14,20 @@ defmodule Phoenix.Socket do
...
+    Raised when the socket message is invalid.
+    """
+    defexception [:message]
+  end
+
   alias Phoenix.Socket
 
   @type t :: %Socket{assigns: %{},
                      channel: atom,
+                     channel_pid: pid,
```
josevalim: broadcast_from uses the channel_pid in the socket instead of using self.

lib/phoenix/socket.ex
```diff
@@ -37,27 +46,72 @@ defmodule Phoenix.Socket do
...
 
   """
-  def assign(socket = %Socket{}, key, value) do
-    put_in socket.assigns[key], value
-  end
+
+  defstruct topic: nil, status: nil, payload: nil, ref: nil
+end
+
+defmodule Phoenix.Socket.Broadcast do
```
josevalim: Besides Phoenix.Socket.Message, I have introduced Reply and Broadcast. Since we need to make this distinction in the test client, we may need to do it for other transports too. For example, if the client does a sync message, it will work now that we have specific Reply messages. It doesn't matter for JS because everything is async.

There is also Broadcast which is what goes into PubSub. I have created a specific struct for it because it may change in the future once we add things like Connectable. So better to set them apart now.

lib/phoenix/channel.ex
chrismccord: great idea

lib/phoenix/transports/long_poller/server.ex
```diff
@@ -81,19 +83,27 @@ defmodule Phoenix.Transports.LongPoller.Server do
...
   """
-  def handle_info({:socket_push, msg}, state) do
+  def handle_info(%Message{} = msg, state) do
     publish_reply(msg, state)
   end
 
+  def handle_info(%Reply{} = reply, state) do
+    %{topic: topic, status: status, payload: payload, ref: ref} = reply
+
+    # TODO: Don't duplicate the reference. The client
```
josevalim: @chrismccord I will need your help for this one. We are encoding the reference in the payload and in the message. We don't need the payload one but we need to change the client for that.

lib/phoenix/transports/websocket.ex
```diff
@@ -90,10 +92,25 @@ defmodule Phoenix.Transports.WebSocket do
...
 
-  def ws_info({:socket_push, message}, %{serializer: serializer} = state) do
+  def ws_info(%Message{} = message, %{serializer: serializer} = state) do
     {:reply, serializer.encode!(message), state}
   end
 
+  def ws_info(%Reply{} = reply, %{serializer: serializer} = state) do
+    %{topic: topic, status: status, payload: payload, ref: ref} = reply
+
+    # TODO: Don't duplicate the reference. The client
```
josevalim: @chrismccord Same here.

lib/phoenix/transports/long_poller/server.ex
chrismccord: :+1: on not duping the `ref` field. I will fix on the client when we merge this

josevalim:
Code wise. This is ready. I just need to write the docs giving a full overview of how to test channels.

josevalim:
Ah, I also need to write tests for `Phoenix.Channel`.

Issue #853, opened 2015-05-09
Consider shipping errors_on with ModelCase


josevalim:
It will be convenient and it will teach folks how to add custom functions to their test cases.

chrismccord:
+1

gjaldon:
Hi guys. I'd love to take this on. What should `errors_on` do? Check that a specific error is found in the Ecto changeset after running a validation?

josevalim:
The code is literally:

```elixir
def errors_on(model, data) do
  model.__struct__.changeset(model, data).errors
end
```

We need to include it in the model case and guarantee it is imported. We should also include some docs with examples, for example:

      assert {:password, "is unsafe"} in errors_on(%User{}, password: "password")

danielberkompas:
+1. Since @josevalim pointed this out to me, I've been adding it to ModelCase manually on all my Phoenix projects.

gjaldon:
This issue can be closed now @josevalim @chrismccord :)

josevalim:
:heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #867, opened 2015-05-11
Add errors_on to ModelCase

gjaldon:
refs: https://github.com/phoenixframework/phoenix/issues/853

### DIFF
--- a/installer/templates/ecto/model_case.ex
+++ b/installer/templates/ecto/model_case.ex
@@ -20,7 +20,7 @@ defmodule <%= application_module %>.ModelCase do
       alias <%= application_module %>.Repo
       import Ecto.Model
       import Ecto.Query, only: [from: 2]
-      import unquote(__MODULE__)
+      import <%= application_module %>.ModelCase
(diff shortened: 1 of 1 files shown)


installer/templates/ecto/model_case.ex
```diff
@@ -17,6 +20,7 @@ defmodule <%= application_module %>.ModelCase do
       alias <%= application_module %>.Repo
       import Ecto.Model
       import Ecto.Query, only: [from: 2]
+      import unquote(__MODULE__)
```
josevalim: I think this pattern is confusing for people stared. Can't we just repeat the name of the case? :)

josevalim:
I have added one small comment, everything else is perfect :+1:

installer/templates/ecto/model_case.ex
gjaldon: Was debating that in my head for a bit. The module name would definitely be much clearer. :)

josevalim:

Issue #617, opened 2015-02-10
Support multiple assets hosts


josevalim:

josevalim:
We need to do this before Plug 1.0 as we need to figure out how digest files will be served by Plug.Static.

laurocaetano:
I'm on this one :)

laurocaetano:
Hi there! 

Just pinging here to tell that I didn't fish the first draft, but I'm working on it. I'll open a PR soon :)

PR #871, opened 2015-05-12
Add `static_url` configuration.

laurocaetano:
This option is the first step to achieve multiple assets hosts (#617).

When accessing the `static_url` function it will return the URL based on this configuration. If no configuration for `static_url` is provided, it will fallback to `url`.


josevalim:

PR #874, opened 2015-05-13
Allow channel messages to be either a map or a list

D1plo1d:
This allows channel messages to be arrays for events that are necessarily plural in their nature but are a distinct set that should be distinguished from other plural message payloads/sets.

One example is a set of changes (like a list of diffs of different files) that should be atomically applied.

### DIFF
--- a/lib/phoenix/channel/server.ex
+++ b/lib/phoenix/channel/server.ex
@@ -72,7 +72,7 @@ defmodule Phoenix.Channel.Server do
   The message is encoded as `Phoenix.Socket.Broadcast`.
   """
   def broadcast(pubsub_server, topic, event, payload)
-      when is_binary(topic) and is_binary(event) and is_map(payload) do
+      when is_binary(topic) and is_binary(event) and (is_map(payload) or is_list(payload)) do
     PubSub.broadcast pubsub_server, topic, %Broadcast{
       topic: topic,
       event: event,
@@ -88,7 +88,7 @@ defmodule Phoenix.Channel.Server do
   Raises in case of crashes.
   """
   def broadcast!(pubsub_server, topic, event, payload)
-      when is_binary(topic) and is_binary(event) and is_map(payload) do
+      when is_binary(topic) and is_binary(event) and (is_map(payload) or is_list(payload)) do
     PubSub.broadcast! pubsub_server, topic, %Broadcast{
       topic: topic,
       event: event,
@@ -104,7 +104,7 @@ defmodule Phoenix.Channel.Server do
   The message is encoded as `Phoenix.Socket.Broadcast`.
   """
   def broadcast_from(pubsub_server, from, topic, event, payload)
-      when is_binary(topic) and is_binary(event) and is_map(payload) do
+      when is_binary(topic) and is_binary(event) and (is_map(payload) or is_list(payload)) do
     PubSub.broadcast_from pubsub_server, from, topic, %Broadcast{
       topic: topic,
       event: event,
@@ -120,7 +120,7 @@ defmodule Phoenix.Channel.Server do
   Raises in case of crashes.
   """
   def broadcast_from!(pubsub_server, from, topic, event, payload)
-      when is_binary(topic) and is_binary(event) and is_map(payload) do
+      when is_binary(topic) and is_binary(event) and (is_map(payload) or is_list(payload)) do
     PubSub.broadcast_from! pubsub_server, from, topic, %Broadcast{
       topic: topic,
       event: event,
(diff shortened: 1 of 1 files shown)


chrismccord:
We discussed a bit on irc, but I'll repeat here. We've considered this, but for es6 js clients at least, it's a matter of:

```javascript
.receive("ok", ({messages}) => ..)
// vs
.receive("ok", (messages) => ..)
```
Requiring maps makes serializers less complex/extensible, and so far we've said "all payloads are maps" which simplifies our apis. I think lists could be made to work for most serializers, but these are why we have so far punted on this idea. I'll let others chime in as I'm not sold on this yet. @josevalim @ericmj @scrogson @jeregrine

scottmessinger:
I'm still learning elixir, but I find the decision to only use maps an excellent choice. In my rails app, I send deltas back and forth between client and server and they're always inside a map. With a map, I'm able to give each message a unique id and send back acks. Also, both my client and server code becomes simpler as I can always expect a map. I would think a list of payloads is better is better sent as a series of individual payloads or as part of a map. E.g. {messages: []}.

D1plo1d:
@scottmessinger Phoenix automatically adds a unique id ("ref" - it's an auto-incremented integer) to each message so that it can be tracked. So whether it's lists or maps your going to be able to reply to that message in a uniquely identifiable way.

This PR refers to an array inside the payload key of the message object so there's still plenty of room for meta data :)

scottmessinger:
@D1plo1d Thanks for the clarification! Semantically, I think maps still make sense. If I'm sending a list of messages, why not just send the messages one by one? If they need to be grouped together (e.g. a transaction or commit), it seems there's a map that's can include the list. Could the methods be changed to accept a list but send each element in the list as it's own message?

D1plo1d:
Just thought of this (totally not my original reason for creating this PR) but it might be the case that constraining the payload to a map could be similarly limiting for serializer implementors. IMO Phoenix should provide good defaults but ultimately allow complete flexibility over the structure of the messages.

@scottmessinger my use case is a websocket channel that deals exclusively in list events that act effectively like SQL transactions. So splitting them up into separate events isn't good for us because we loose that implicit transactional aspect of the single event. It is possible to send messages in a list inside a map like you mentioned and we explored that option but going that route we ended up with things like `message["payload"]["payload"]` or `message["payload"]["events"]` in our non-web clients. It wasn't elegant and we were running out of synonyms for "the place that the actual payload goes".

What it comes down to is that as a developer creating a websocket API I'd like to have:
1. good defaults
2. complete control over the way my data is structured (ie. the ability to override those defaults)

josevalim:
I am honestly ok with allowing the message to be anything, as long as it is serializable. As others said, Phoenix already wraps the payload in a message, so we are safe on the metadata side for Phoenix itself.

chrismccord:
> I am honestly ok with allowing the message to be anything, as long as it is serializable. As others said, Phoenix already wraps the payload in a message, so we are safe on the metadata side for Phoenix itself.

We still have the issue if this not working:

    {:reply, {:ok, [some: "keyword list"]}, socket}

The fact that it is already wrapped in a message struct still doesn't remove the map restrictions. We could say we support strings, lists, and maps, so long as there are no keyword lists.

josevalim:
@chrismccord we support anything that is serializable. You can still pass a keyword list inside a map. PIDs, references won't be supported too. The question is if we should check for them explicitly or just let the serializer blow up. Then we need to be careful because a failure in the serializer can bring the transport down?

josevalim:
Or you can just use any tuple and that will crash the serializer too. The only argument I can think about being strict is that it is more likely to play with other transports.

So I think I will change my instance here and adopt my "rule": let's go strict for now (only maps) and we can revisit this in the future. It is always better to be strict until we figure out how other transports/serializers work.

chrismccord:
> let's go strict for now (only maps) and we can revisit this in the future. It is always better to be strict until we figure out how other transports/serializers work.

:+1: This is where I'm at. You are right that even with map restrictions, `%{foo: [this: "will blowup"]}`, is still possible. @D1plo1d we can revisit this later, preferably when we have another common serializer other than json. Thanks for prompting the discussion and the PR!

D1plo1d:
@chrismccord Cool. I'll maintain the branch since we're already using it here at TouchBistro.

Now I'm wondering if and when I'll have enough time to implement a message pack serializer so that we can re-open this discussion \*evilly tents fingers\*

chrismccord:
> Now I'm wondering if and when I'll have enough time to implement a msg pack serializer so that we can re-open this discussion *evilly tents fingers*

You have until July 1.0!

PR #878, opened 2015-05-13
gen.html task: use PUT rather than POST on edit

voltone:


josevalim:
We automatically set it to put based on the resource:

https://github.com/phoenixframework/phoenix_ecto/blob/master/lib/phoenix_ecto/html.ex#L29-L30

So this shouldn't be needed. Are you seeing a behaviour that says otherwise?

voltone:
Ah, that would explain why I couldn't find anyone else with the same issue. I had a hunch it might be related to the fact that I'm using Ecto's Changesets, but not the Repo layer. I guess I'll have to check phoenix_ecto to see how to make sure form_for sets the right options for my changesets.

I may have some suggestions for documentation improvements once I figure this out, but I'll send a separate PR for that. Thanks :)

josevalim:
@voltone so if the resource is loaded we use put. If not, then we default to POST. it may be easier for you to just pass method in your case.

PR for the docs are always welcome too! Thank you!

PR #881, opened 2015-05-14
Add bootstrap comment re: meta order

kkirsche:
Rather than explain the issue that can occur if they are not in this order, I am referencing the original issue which this comment was added to avoid. 

twbs/bootstrap#15601


chrismccord:
Thanks for the reference. We are still in the process of deciding what we're doing wrt to bootstrap, so I'm going to wait off on this one.

kkirsche:
It doesn't change the issues with IE compatibility if that order is changed regardless of the framework you use though @chrismccord

chrismccord:
Right, but I'm not sure a comment warning users in this case is the answer. There are lots of IE edge cases that could be inline commented about, so for now I'll leave it to the end users.

kkirsche:
Fair enough. Thanks for the consideration.

PR #882, opened 2015-05-14
Encourage accessible HTML

kkirsche:
Encourage others to write HTML which is accessible. Roles aid assistive technologies in deciding what the correct way to announce content to a user is. For this reason, I think it is important to encourage developers to write accessible HTML using roles, as recommended by Bootstrap which is used here, rather than bare HTML.


chrismccord:
Thanks!

PR #883, opened 2015-05-14
[Close #847] Provide better error message for dead connections

### LINKED ISSUE #847: Provide better error messages for when the connection is not assigned in tests
http://stackoverflow.com/questions/30063819/runtimeerror-expected-connection-to-have-a-response

kkirsche:
[Close #847] Provide better error message when connection doesn't receive a response.

RE: phoenixframework/phoenix#847

### DIFF
--- a/lib/phoenix/test/conn_test.ex
+++ b/lib/phoenix/test/conn_test.ex
@@ -308,7 +308,8 @@ defmodule Phoenix.ConnTest do
   """
   @spec response(Conn.t, status :: integer | atom) :: binary | no_return
   def response(%Conn{state: :unset}, _status) do
-    raise "expected connection to have a response but no response was set/sent"
+    raise "expected a response from connection but no response was set/sent." <>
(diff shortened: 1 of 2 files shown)


chrismccord:
We need a bit more than this because we need to flag the connection, i.e. as José said:

> The idea is to store something when we create the connection and remove it when the user calls get/post/etc. Then, in response, if we see the flag is still there, we can give a really nice example of what should be done!

Let's also wait and see if @parkerl has a pending PR in the works since he volunteered first. Thanks!

kkirsche:
When you say you want to "flag the connection" what exactly do you mean?

josevalim:
On a second though, flagging the connection will cause issues though, so I will go with a solution close to this but with a more detailed error message.

kkirsche:
@josevalim what kind of more detailed? I'm not clear what this was missing other than not being the one who responded in the thread...

PR #895, opened 2015-05-16
Provide info to run migration for phoenix.gen.model

shankardevy:
When Phoenix.gen.html generates model and migration files, it gives info about running migrations. However, when Phoenix.gen.model generates the model and migration files, this info is missing. This commit makes it even.

### DIFF
--- a/lib/mix/tasks/phoenix.gen.model.ex
+++ b/lib/mix/tasks/phoenix.gen.model.ex
@@ -68,6 +68,11 @@ defmodule Mix.Tasks.Phoenix.Gen.Model do
       {:eex, "model.ex",       "web/models/#{path}.ex"},
       {:eex, "model_test.exs", "test/models/#{path}_test.exs"},
     ]
+    
+    Mix.shell.info """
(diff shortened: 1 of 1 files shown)


josevalim:
If we merge this, this information will be shown twice in the `phoenix.gen.html` and `phoenix.gen.json` generators, so we would need to pass a flag to not show it in such cases. Also, we could argue that phoenix.gen.model is more of a building block than a learning tool like html/json and that at this point folks will be familar with running migrations.

shankardevy:
We should probably have the message to run 'Migration' only on `phoenix.gen.model` (where it should actually belong) and remove it from `phoenix.gen.html` and `phoenix.gen.json`. So we don't need additional flags to avoid duplicate messages. I will update my patch for the same, if this is ok.

josevalim:
Note We need to keep the order though (migration info should come last) and
this may be hard to implement. So duplicating the message may be the
easiest way to go.

-- 


josevalim:
I am closing this as it would warn nice for phoenix.gen.html.

PR #900, opened 2015-05-19
Allow custom refs in child_specs

voltone:
This lets me run multiple listeners for the same endpoint under a single supervisor, for instance an IPv4 and an IPv6 listener. (There are a few other places where Phoenix assumes that there will only ever be one HTTP and one HTTPS listener, for instance in configuration and, by extension, the URL building functions.)

### DIFF
--- a/lib/phoenix/endpoint/cowboy_handler.ex
+++ b/lib/phoenix/endpoint/cowboy_handler.ex
@@ -24,14 +24,16 @@ defmodule Phoenix.Endpoint.CowboyHandler do
   end
 
   def child_spec(scheme, endpoint, config) do
-    # Use put_new to allow custom dispatches
-    config = Keyword.put_new(config, :dispatch, [{:_, [{:_, __MODULE__, {endpoint, []}}]}])
+    # Use put_new to allow custom dispatches and refs
+    config = config
+    |> Keyword.put_new(:dispatch, [{:_, [{:_, __MODULE__, {endpoint, []}}]}])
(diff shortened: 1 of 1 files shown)


josevalim:

chrismccord:
We need to handle only stating a single pubsub server along with this, right?

voltone:
Just to clarify my use-case, there is really only one Endpoint, and I'm starting it from the top-level supervisor as usual, but without any servers. I then start any listeners under my own supervisor. Live reloading works for those listeners, so I guess the PubSub processes under the Endpoint are working fine.

chrismccord:
How are you configuring the pubsub processes? We register the pubsub server under a local name, so only one could be running for the configure `pubsub: [name: ...]` on the endpoint.

josevalim:
@chrismccord he wants to bind the same server on multiple ip/ports. He still wants just one pubsub, one config, etc. He is just running the same cowboy instance on more ports.

voltone:
I'm not: my app's Endpoint starts as usual, with its CodeReloader and Config workers, and the PubSub supervisor and workers. I just told the Endpoint to start without the Server (server: false).

The only difference between a vanilla Phoenix app and my setup once all is running, apart from the fact that the Cowboy listeners live under a different supervisor, seems to be the absence of a Watcher process. I can live with that for now.

voltone:
("I'm not" refers to @chrismccord's "How are you configuring...")

chrismccord:
Ahhh I see

PR #902, opened 2015-05-19
Ignore priv static

gjaldon:
Closes https://github.com/phoenixframework/phoenix/issues/875

When generating a new Phoenix app with brunch, it generates all static files in `web/static` and gitignores all files in `priv/static`.


installer/templates/static/brunch/.gitignore
```diff
@@ -12,8 +12,7 @@ erl_crash.dump
 # Since we are building js and css from web/static,
```
josevalim: We need to update the comments. :)

gjaldon:
Updated the comments!

josevalim:

Issue #887, opened 2015-05-14
Allow assigns to be merged on a router


josevalim:
Similar to how we do private:

https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/router.ex#L588

chrismccord:
What's the use-case for this that private's don't include. Defaults? Just curious if a specific need came up or if this is for future convenience

josevalim:
Yeah, folks want this to be in assigns because they want to use it directly and not as part of a library (which is what privates are advised for).

chrismccord:
:+1:

gjaldon:
@josevalim I'd like to help by working on this. 

Just to clarify, we want an `:assigns` option added to the `scope` macro? Whatever map we pass to this `assigns` opt will be merged to the conn struct when the route matches?

josevalim:
Yes! It is literally the same as the private feature we have today. :)

-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Lead Developer*

PR #903, opened 2015-05-19
Add assigns

gjaldon:
Closes https://github.com/phoenixframework/phoenix/issues/887


lib/phoenix/router.ex
```diff
@@ -586,6 +586,7 @@ defmodule Phoenix.Router do
     * `:host` - a string containing the host scope, or prefix host scope, ie
                 `"foo.bar.com"`, `"foo."`
     * `:private` - a map of private data to merge into the connection when a route matches
+    * `:assigns` - a map of data to merge into the connection when a route matches
```
chrismccord: we need to clarify which takes precedence. Something like, "The default map of assigns for the connection"

josevalim:
Looks perfect to me.

chrismccord:
I'll bump the doc line. Thanks!

PR #904, opened 2015-05-21
Initial redis pubsub extraction

chrismccord:
@josevalim I dup'd the test cases. We talked before about maybe some ways we could dry up the tests b/w the core PG2 adapter and redis (and others). I'll find you on irc so you can remind me :)


josevalim:
The idea is that we expose the tests in separate files, for example `test/pubsub/func1.exs`, `test/pubsub/func2.exs` and publish those in the hex pages. Then phoenix_redis_pubsub will be able to setup the proper environment and do `Code.require_file "deps/phoenix/test/pubsub/func1.exs"`. This is what TDS Ecto does:

https://github.com/livehelpnow/tds_ecto/blob/master/test/test_helper.exs#L72-L78

josevalim:
In our case, it should be a bit easier, because all we need to do is to ask the pubsub to register itself with a given name... and we can use this name during the tests.

chrismccord:
@josevalim I extracted the testcases into a `.test/support/pubsub_setup.exs` file and now both the PG2 adapter and the `phoenix_pubsub_redis` project use it. Hows' the location and name of the test file? This should be ready to merge

josevalim:
@chrismccord this looks beautiful. I am not sure test/support is the best name though... maybe `test/shared`? Also, I would end the module name with `Test` too, as it is still a test. Maybe something like `Phoenix.PubSubTest`?

Finally, you need change hex files and include "test/shared" (or whatever we decide to call it) into the hex package.

test/test_helper.exs
```diff
@@ -14,10 +14,5 @@ end
 # Start transport levels
 Process.flag(:trap_exit, true)
```
josevalim: We can remove those trap exits too.

chrismccord:
test/shared and PubSubTest sound good to me. I'll make the changes and merge. Thanks!

chrismccord:
> Finally, you need change hex files and include "test/shared" (or whatever we decide to call it) into the hex package.

Anything specific we need to do here when we publish next?

josevalim:
Yeah, check this out: https://github.com/elixir-lang/ecto/blob/master/mix.exs#L56-L57

I will add this to Phoenix master though if you haven't yet. :)

PR #911, opened 2015-05-23
Fix static_url helper to use static_url endpoint config

follmann:
Configured `static_url` was not used to generate the url via the `static_url` helper and the `url` helper was also broken.


follmann:
The tests are definitely passing locally.

josevalim:
The failure is a timeout being triggered due to CI being slow. This patch is perfect, thank you.

follmann:
Happy to contribute!

PR #915, opened 2015-05-25
implemented jsonp support for controllers and added tests

pb82:
Adds jsonp support to controllers. While CORS is usually preferable for cross domain requests I think there is also some value in having jsonp support: works well with older browsers (IE < 10), good jQuery integration, etc.

The syntax for the jsonp feature is the same as for json:

```elixir
jsonp conn, %{foo: :bar}
```
This will produce the following output on request (GET /?callback=cb):

```javascript
/**/ typeof cb === 'function' && cb({"foo":"bar"});
```

It's possible to override the callback parameter:

```elixir
jsonp conn, %{foo: :bar}, callback: "cb"
```
that allows you to use the overridden callback name in the query string (GET /?cb=cb)

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -133,6 +133,12 @@ defmodule Phoenix.Controller do
   @spec controller_template(Plug.Conn.t) :: binary
   def controller_template(conn), do: conn.private[:phoenix_template]
 
+  @compile {:inline, get_json_encoder: 0}
+  defp get_json_encoder do
+    Application.get_env(:phoenix, :format_encoders)
+    |> Keyword.get(:json, Poison)
+  end
+
   @doc """
   Sends JSON response.
 
@@ -146,11 +152,54 @@ defmodule Phoenix.Controller do
   """
   @spec json(Plug.Conn.t, term) :: Plug.Conn.t
   def json(conn, data) do
-    encoder =
-      Application.get_env(:phoenix, :format_encoders)
-      |> Keyword.get(:json, Poison)
-
+    encoder = get_json_encoder()
+    
     send_resp(conn, conn.status || 200, "application/json", encoder.encode_to_iodata!(data))
+  end
+
+  @doc """
+  Sends a JSON response with JSONP callback support.
+
+  It uses the configured `:format_encoders` under the `:phoenix`
+  application for `:json` to pick up the encoder module.
+
+  If the query params contain a key for the callback name, the 
+  response body is wrapped in an invocation of this callback.
+
+  The default key for the callback name is `"callback"`. It can be
+  overridden with the `:callback` option and is expected to be a 
+  string.
+
+  ## Examples
+
+      iex> jsonp conn, %{id: 123}, callback: "cb"
+
+  """
+  @spec jsonp(Plug.Conn.t, term, Keyword.t) :: Plug.Conn.t
+  def jsonp(conn, data, opts \\ []) do
+    callback_param = Keyword.get(opts, :callback, "callback")
+    callback = Dict.get(conn.query_params, callback_param)
+
+    unless callback do
+      json(conn, data)
+    else
+      encoder = get_json_encoder()
+
+      # Restrict callback name to alphanumeric characters, $, _, . and 
+      # square brackets
+      callback = Regex.replace(~r/[^\[\]\w$.]/, callback, "")
+  
+      # Replace unicode characters that are valid in json but not
+      # in javascript (line separator, paragraph separator)
+      body = encoder.encode!(data)
+      body = body
+      |> String.replace(<< 0x2028 :: utf8 >>, "\\u2028")
+      |> String.replace(<< 0x2029 :: utf8 >>, "\\u2029")
+      
+      body = "/**/ typeof " <> callback <> " === 'function' && " <> callback <> "(" <> body <> ");"
+
+      send_resp(conn, conn.status || 200, "text/javascript", body)
+    end
   end
 
   @doc """
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -169,7 +169,7 @@ defmodule Phoenix.Controller do
   overridden with the `:callback` option and is expected to be a 
   string.
 
-  Only alphanumeric characters and underscore is allowed in the
+  Only alphanumeric characters and underscore are allowed in the
   callback name. Otherwise an exception is raised.
 
   ## Examples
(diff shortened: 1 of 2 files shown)


lib/phoenix/controller.ex
```diff
@@ -133,6 +133,12 @@ defmodule Phoenix.Controller do
   @spec controller_template(Plug.Conn.t) :: binary
   def controller_template(conn), do: conn.private[:phoenix_template]
 
+  @compile {:inline, get_json_encoder: 0}
```
josevalim: There is no need to inline it.

lib/phoenix/controller.ex
```diff
@@ -146,14 +152,57 @@ defmodule Phoenix.Controller do
...
+    callback = Dict.get(conn.query_params, callback_param)
+
+    unless callback do
+      json(conn, data)
+    else
+      encoder = get_json_encoder()
+
+      # Restrict callback name to alphanumeric characters, $, _, . and 
+      # square brackets
+      callback = Regex.replace(~r/[^\[\]\w$.]/, callback, "")
```
josevalim: I would rather raise if the format is not valid. We should avoid regexes too.

lib/phoenix/controller.ex
```diff
@@ -146,14 +152,57 @@ defmodule Phoenix.Controller do
...
+  
+      # Replace unicode characters that are valid in json but not
+      # in javascript (line separator, paragraph separator)
+      body = encoder.encode!(data)
```
josevalim: encode! is not part of the format_encoder contract. You should call encoder_to_iodata! and then use `IO.iodata_to_binary`.

test/phoenix/controller/controller_test.exs
```diff
@@ -132,6 +132,45 @@ defmodule Phoenix.Controller.ControllerTest do
...
+  test "jsonp/3 returns javascript when callback param is present" do
+    conn = conn(:get, "/?callback=cb") |> fetch_query_params()
+    conn = jsonp(conn, %{foo: :bar})
+    assert conn.resp_body == "/**/ typeof cb === 'function' && cb({\"foo\":\"bar\"});"
+    assert get_resp_content_type(conn) == "text/javascript"
+    refute conn.halted
+  end
+    
+  test "jsonp/3 allows to override the callback name" do
+    conn = conn(:get, "/?cb=cb") |> fetch_query_params()
```
josevalim: The callback name is the same as the one below. :)

josevalim:
Thank you. I have added some comments!

lib/phoenix/controller.ex
```diff
@@ -146,14 +152,57 @@ defmodule Phoenix.Controller do
...
+
+      iex> jsonp conn, %{id: 123}, callback: "cb"
+
+  """
+  @spec jsonp(Plug.Conn.t, term, Keyword.t) :: Plug.Conn.t
+  def jsonp(conn, data, opts \\ []) do
+    callback_param = Keyword.get(opts, :callback, "callback")
+    callback = Dict.get(conn.query_params, callback_param)
+
+    unless callback do
```
chrismccord: We should avoid unless...else because it's hard to understand. @josevalim since when is unless/else allowed? I didn't think it was valid and was glad when I thought that was the case.

doomspork: Positive case first, :+1: for this suggestion.  Would be awesome to have an Elixir styleguide or best practices document, don't believe there is one currently.

lib/phoenix/controller.ex
pb82: I thought that since `get_json_encoder` is both small and (potentially) called frequently it would make a good candidate for inlining.

lib/phoenix/controller.ex
josevalim: It was always allowed. But yes, unless/else is asking for trouble.

pb82:
Thank you for the review @josevalim and @chrismccord. I will update my pull request as soon as possible and take your comments into account.

lib/phoenix/controller.ex
pb82: I will rewrite that using `case` :smirk:

josevalim: `if` would also be fine. The trouble is just the double negation.

lib/phoenix/controller.ex
josevalim: We should not even support dots. See here: http://www.benhayak.com/2015/05/stealing-private-photo-albums-from-Google.html

I would keep it to alphanumeric and underscore. No square brackets too as `window[variable]` can be a vector too.

pb82:
I've updated the pull request: 

* removed the inlining.
* no more double negation.
* only alphanumeric and underscore allowed for callback names.
* no more regex use. I'm not sure if my solution is optimal though.

What do you think?

lib/phoenix/controller.ex
```diff
@@ -146,14 +151,66 @@ defmodule Phoenix.Controller do
...
+
+      iex> jsonp conn, %{id: 123}, callback: "cb"
+
+  """
+  @spec jsonp(Plug.Conn.t, term, Keyword.t) :: Plug.Conn.t
+  def jsonp(conn, data, opts \\ []) do
+    callback_param = Keyword.get(opts, :callback, "callback")
+    callback = Dict.get(conn.query_params, callback_param)
+
+    if callback && String.length(callback) > 0 do
```
josevalim: `String.length` is going to traverse the whole string. You want to use `byte_size`, which is `O(1)`, or just check `callback != ""`.

lib/phoenix/controller.ex
```diff
@@ -146,14 +151,66 @@ defmodule Phoenix.Controller do
...
+      encoder = get_json_encoder()
+
+      allowed_chars = [?0..?9, ?A..?Z, ?a..?z, [?_]]
```
josevalim: You could have a validate_callback! function:

```elixir
def validate_callback!(<<h, t::binary>>) when h in ?0..?9 or h in ?A..?Z or h in ?a..?z or h == ?_, do: validate_callback!(t)
def validate_callback!(<<>>), do: :ok
def validate_callback!(_), do: raise(ArgumentError, "invalid callback ......")
```

lib/phoenix/controller.ex
```diff
@@ -146,14 +151,66 @@ defmodule Phoenix.Controller do
...
+
+      body = body
+      |> String.replace(<< 0x2028 :: utf8 >>, "\\u2028")
+      |> String.replace(<< 0x2029 :: utf8 >>, "\\u2029")
+      
+      body = "/**/ typeof " <> callback <> " === 'function' && " <> callback <> "(" <> body <> ");"
+
+      send_resp(conn, conn.status || 200, "text/javascript", body)
+    else
+      json conn, data
```
josevalim: Maybe we should raise if there is no callback? What is the purpose of returning json for a jsonp call? maybe it could even be a security issue?

lib/phoenix/controller.ex
```diff
@@ -146,14 +151,66 @@ defmodule Phoenix.Controller do
...
+      |> String.to_char_list()
+      |> Enum.all?(&(&1 in allowed_chars))
+
+      callback_valid || raise ArgumentError, "the callback name contains invalid characters"
+
+      body = encoder.encode_to_iodata!(data) 
+      |> IO.iodata_to_binary()
+
+      body = body
+      |> String.replace(<< 0x2028 :: utf8 >>, "\\u2028")
```
josevalim: Nitpick (sorry!): no spaces after `<<` and before `>>`

josevalim:
I've added a couple more comments and one small nitpick. :)

lib/phoenix/controller.ex
pb82: The advantage is that it let's you define an API that works both, same domain and cross domain. When it's not cross domain you just leave away the callback and get a normal json response. Otherwise you would have to treat those two cases separately.

josevalim: Ah, I see, perfect. Thank you!

lib/phoenix/controller.ex
pb82: Ah yes, you're right :) I thought that i can't do this because callback may be `nil`. But of course i can set the default value of `Dict.get` to the empty string. Will do that!

lib/phoenix/controller.ex
```diff
@@ -146,14 +151,66 @@ defmodule Phoenix.Controller do
...
+      |> Enum.map(&Enum.to_list/1)
+      |> List.flatten()
+
+      callback_valid = callback
+      |> String.to_char_list()
+      |> Enum.all?(&(&1 in allowed_chars))
+
+      callback_valid || raise ArgumentError, "the callback name contains invalid characters"
+
+      body = encoder.encode_to_iodata!(data)
```
jeregrine: could you just call encoder.encode! here and skip the IO.iodata_to_binary?

https://github.com/devinus/poison/blob/master/lib/poison.ex#L39

jeregrine:
Very cool!

lib/phoenix/controller.ex
josevalim: I asked him to not call encode! because it is not part of our  `format_encoder` contract. We just call `encode_to_iodata!`.

pb82:
Thank you, your comments are really useful! I updated the pull request and took the freedom to add your suggested validation function (just made it private) ;)

josevalim:
This looks perfect, thank you. @chrismccord any final remarks?

chrismccord:
<3 <3 <3

PR #917, opened 2015-05-26
Sample data generated for float and decimal should be of correct type

gjaldon:
Just changed data generated in test to use proper type. `decimal` should use `Decimal` type and `float` should be a `float`.

### DIFF
--- a/lib/mix/phoenix.ex
+++ b/lib/mix/phoenix.ex
@@ -80,8 +80,8 @@ defmodule Mix.Phoenix do
       {k, {:array, _}} -> {k, []}
       {k, :belongs_to} -> {k, nil}
       {k, :integer}    -> {k, 42}
-      {k, :float}      -> {k, "120.5"}
-      {k, :decimal}    -> {k, "120.5"}
+      {k, :float}      -> {k, 120.5}
+      {k, :decimal}    -> {k, Decimal.new(43.20)}
       {k, :boolean}    -> {k, true}
       {k, :text}       -> {k, "some content"}
       {k, :date}       -> {k, %{year: 2010, month: 4, day: 17}}


lib/mix/phoenix.ex
```diff
@@ -80,8 +80,8 @@ defmodule Mix.Phoenix do
       {k, {:array, _}} -> {k, []}
       {k, :belongs_to} -> {k, nil}
       {k, :integer}    -> {k, 42}
-      {k, :float}      -> {k, "120.5"}
-      {k, :decimal}    -> {k, "120.5"}
+      {k, :float}      -> {k, 120.5}
+      {k, :decimal}    -> {k, Decimal.new(43.20)}
```
josevalim: This data is meant to be sent over the form or via changesets, which can do casting. In fact, sending `Decimal.new(43.20)` may not even work. Is there a particular error you are seeing?

gjaldon: I only see `params` passed to the `model_test` and `controller_test`s for html and json. They don't seem to affect changesets or the form.

Anyway, no error, but saw the sample data generated in the tests used the string "120.5" for decimal. Ran  `validate_number/3` for the decimal field and it would always pass with the string. Checked out the docs in `Schema` and saw that `:decimal` uses a `Decimal` module. Using the `Decimal` module makes the validation work as expected.

So thought of passing `Decimal` and a `float`. The `Decimal.new` above makes 2 tests fail though and will need to change. Should I continue to change or do we close this? :)

josevalim: > Anyway, no error, but saw the sample data generated in the tests used the string "120.5" for decimal. Ran validate_number/3 for the decimal field and it would always pass with the string.

If this is the case... then we have a bug. The value should be cast up to the point we reach validate_number. However i have just realized validate_number likely won't work with Decimal. So we need to fix it anyway.

Other than that, I don't think we need to change anything in Phoenix. :)

gjaldon: Ok closing this PR and opening an issue/PR in Ecto regarding this. :)

PR #925, opened 2015-05-30
Add for attr to label tag in html generator

gjaldon:


josevalim:
The solution is perfect but I don't like the fact we *need* it. I can see two options:

1. Move the input inside the label. I am not sure if bootstrap would be fine with this though;

2. If we need to use `for="..."`, then I would add a `label(f, input, text, opts \\ [])` to Phoenix.HTML.Form, ship v1.2 and use it here.

Thoughts?

gjaldon:
Of the two options you point out, I think two is the better one. Nesting input under label messes with bootstrap styling and is generally harder to style.

Went ahead and created the issue at https://github.com/phoenixframework/phoenix_html/issues/9. :)

Will open another PR here once 1.2 of phoenix_html is shipped.

PR #926, opened 2015-05-30
Set different assets directory

gjaldon:
I encountered a problem when installing bootstrap-sass through bower where I noticed its JS files weren't being compiled. Turns out bootstrap-sass has an `/assets` directory and brunch, by default, copies all files in any `/assets` dir, even ones in bower deps, straight to `priv/static`.

This PR fixes this problem and shows where to copy non-css and non-js assets to, which is `web/static/assets` by default.

### DIFF
--- a/installer/templates/static/brunch/brunch-config.js
+++ b/installer/templates/static/brunch/brunch-config.js
@@ -27,6 +27,13 @@ exports.config = {
     }
   },
 
+  conventions: {
+    // This option sets where we should place non-css and non-js assets in.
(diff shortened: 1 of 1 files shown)


josevalim:

chrismccord:
Thanks! This fixes the issues that @scrogson had with bootstrap-sass

josevalim:
Yeah, perfect solution. Solves problems and documents the whole thing.

gjaldon:
:smiley:

PR #928, opened 2015-05-30
Use label tag to generate labels in phoenix.gen.html

gjaldon:


lib/mix/tasks/phoenix.gen.html.ex
```diff
@@ -87,6 +87,17 @@ defmodule Mix.Tasks.Phoenix.Gen.Html do
     """
   end
 
+  defp labels(attrs) do
```
josevalim: Since this is duplicate from the inputs, i think we should change how inputs are generated. maybe make them tuples with three elements, containing the form function, the input name, and then the rest of the information.

test/mix/tasks/phoenix.gen.html_test.exs
```diff
@@ -42,6 +42,13 @@ defmodule Mix.Tasks.Phoenix.Gen.HtmlTest do
         assert file =~ ~s(<%= checkbox f, :famous, class: "form-control" %>)
         assert file =~ ~s(<%= datetime_select f, :born_at, class: "form-control" %>)
         assert file =~ ~s(<%= text_input f, :secret, class: "form-control" %>)
+        assert file =~ ~s(<%= label f, :name, "Name" %>)
+        assert file =~ ~s(<%= label f, :age, "Age" %>)
+        assert file =~ ~s(<%= label f, :height, "Height" %>)
+        assert file =~ ~s(<%= label f, :address_id, "AddressId" %>)
+        assert file =~ ~s(<%= label f, :famous, "Famous" %>)
+        assert file =~ ~s(<%= label f, :born_at, "BornAt" %>)
```
josevalim: Should we add humanize to Phoenix.Naming so we can use it here?

https://github.com/phoenixframework/phoenix_html/blob/master/lib/phoenix_html/form.ex#L138-L149

It would have to be duplicated... but i guess it is fine.

gjaldon:
@josevalim this is ready for review again :)

josevalim:

PR #929, opened 2015-05-30
Generators should be aware of reserved names

### LINKED ISSUE #921: generator unaware of reserved names
In a new phoenix project: 
```bash
$ mix phoenix.new throwaway 
#install all dependencies
$ cd throwaway    
$ mix phoenix.gen.html Repo repos
# add resources "/repos", RepoController to web/router.ex
$ mix ecto.migrate
=INFO REPORT==== 29-May-2015::10:40:22 ===
    application: logger
    exited: stopped
    type: temporary
** (Mix) Could not start application throwaway: Throwaway.start(:normal, []) returned an error: shutdown: failed to start child: Throwaway.Repo
    ** (EXIT) an exception was raised:
        ** (UndefinedFunctionError) undefined function: Throwaway.Repo.start_link/0
            (throwaway) Throwaway.Repo.start_link()
            (stdlib) supervisor.erl:314: :supervisor.do_start_child/2
            (stdlib) supervisor.erl:297: :supervisor.start_children/3
            (stdlib) supervisor.erl:263: :supervisor.init_children/2
            (stdlib) gen_server.erl:306: :gen_server.init_it/6
            (stdlib) proc_lib.erl:237: :proc_lib.init_p_do_apply/3
```

The problem being that `Repo` is a name already in use.

doomspork:
RFC @josevalim — Thanks for letting me take a stab at this, looking forward to your feedback.

This closes #921.

### DIFF
--- a/lib/mix/phoenix.ex
+++ b/lib/mix/phoenix.ex
@@ -93,6 +93,16 @@ defmodule Mix.Phoenix do
   end
 
   @doc """
+  Checks the availability of a given module name.
+  """
+  def check_module_name_availability!(name) do
+    name = Module.concat(Elixir, name)
+    if Code.ensure_loaded?(name) do
+      Mix.raise "Module name #{inspect name} is already taken, please choose another name"
+    end
+  end
+
+  @doc """
   Returns the module base name based on the configuration value.
 
       config :my_app
--- a/lib/mix/tasks/phoenix.gen.channel.ex
+++ b/lib/mix/tasks/phoenix.gen.channel.ex
@@ -25,6 +25,8 @@ defmodule Mix.Tasks.Phoenix.Gen.Channel do
 
     binding = binding ++ [plural: plural]
 
+    Mix.Phoenix.check_module_name_availability!(binding[:module] <> "Channel")
+
     Mix.Phoenix.copy_from source_dir, "", binding, [
       {:eex, "channel.ex",       "web/channels/#{path}_channel.ex"},
       {:eex, "channel_test.exs", "test/channels/#{path}_channel_test.exs"},
--- a/lib/mix/tasks/phoenix.gen.html.ex
+++ b/lib/mix/tasks/phoenix.gen.html.ex
@@ -39,6 +39,9 @@ defmodule Mix.Tasks.Phoenix.Gen.Html do
     binding = binding ++ [plural: plural, route: route,
                           inputs: inputs(attrs), params: Mix.Phoenix.params(attrs)]
 
+    Mix.Phoenix.check_module_name_availability!(binding[:module] <> "Controller")
+    Mix.Phoenix.check_module_name_availability!(binding[:module] <> "View")
+
     Mix.Phoenix.copy_from source_dir, "", binding, [
       {:eex, "controller.ex",       "web/controllers/#{path}_controller.ex"},
       {:eex, "edit.html.eex",       "web/templates/#{path}/edit.html.eex"},
--- a/lib/mix/tasks/phoenix.gen.json.ex
+++ b/lib/mix/tasks/phoenix.gen.json.ex
@@ -37,6 +37,9 @@ defmodule Mix.Tasks.Phoenix.Gen.Json do
     route   = String.split(path, "/") |> Enum.drop(-1) |> Kernel.++([plural]) |> Enum.join("/")
     binding = binding ++ [plural: plural, route: route, params: Mix.Phoenix.params(attrs)]
 
+    Mix.Phoenix.check_module_name_availability!(binding[:module] <> "Controller")
+    Mix.Phoenix.check_module_name_availability!(binding[:module] <> "View")
+
     files = [
       {:eex, "controller.ex",       "web/controllers/#{path}_controller.ex"},
       {:eex, "view.ex",             "web/views/#{path}_view.ex"},
--- a/lib/mix/tasks/phoenix.gen.model.ex
+++ b/lib/mix/tasks/phoenix.gen.model.ex
@@ -56,6 +56,8 @@ defmodule Mix.Tasks.Phoenix.Gen.Model do
     path      = binding[:path]
     migration = String.replace(path, "/", "_")
 
+    Mix.Phoenix.check_module_name_availability!(binding[:module])
+
     {assocs, attrs} = partition_attrs_and_assocs(attrs)
 
     binding = binding ++
--- a/test/mix/tasks/phoenix.gen.channel_test.exs
+++ b/test/mix/tasks/phoenix.gen.channel_test.exs
@@ -1,5 +1,8 @@
 Code.require_file "../../../installer/test/mix_helper.exs", __DIR__
 
+defmodule Phoenix.DupChannel do
+end
+
 defmodule Mix.Tasks.Phoenix.Gen.ChannelTest do
   use ExUnit.Case
   import MixHelper
@@ -74,6 +77,12 @@ defmodule Mix.Tasks.Phoenix.Gen.ChannelTest do
   test "passing extra args raises error" do
     assert_raise Mix.Error, fn ->
       Mix.Tasks.Phoenix.Gen.Channel.run ["Admin.Room", "rooms", "new_message"]
+    end
+  end
+
+  test "name can't already be defined" do
+    assert_raise Mix.Error, fn ->
+      Mix.Tasks.Phoenix.Gen.Channel.run ["Dup", "dups"]
     end
   end
 end
--- a/test/mix/tasks/phoenix.gen.model_test.exs
+++ b/test/mix/tasks/phoenix.gen.model_test.exs
@@ -1,5 +1,8 @@
 Code.require_file "../../../installer/test/mix_helper.exs", __DIR__
 
+defmodule Phoenix.Dup do
+end
+
 defmodule Mix.Tasks.Phoenix.Gen.ModelTest do
   use ExUnit.Case
   import MixHelper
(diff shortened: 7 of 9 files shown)


lib/mix/phoenix.ex
```diff
@@ -93,6 +93,16 @@ defmodule Mix.Phoenix do
   end
 
   @doc """
+  Checks the availability of a given module name.
+  """
+  def check_module_name_availability!(name) do
```
doomspork: I copied the `check_module_name_availability!` verbatim, as I saw little need to modify it.

lib/mix/tasks/phoenix.gen.channel.ex
```diff
@@ -52,6 +52,11 @@ defmodule Mix.Tasks.Phoenix.Gen.Channel do
     unless length(args) == 2 do
       raise_with_help
     end
+
+    args
```
josevalim: The generated module contains `Channel` at the end. So I wouldn't check it here but just after we call `Mix.Project.inflect`.

lib/mix/tasks/phoenix.gen.html.ex
```diff
@@ -66,7 +66,8 @@ defmodule Mix.Tasks.Phoenix.Gen.Html do
     end
   end
 
-  defp validate_args!([_, plural | _] = args) do
+  defp validate_args!([singular, plural | _] = args) do
+    Mix.Phoenix.check_module_name_availability!(singular)
```
josevalim: Same here. The generate modules end with "Controller" and "View".

lib/mix/tasks/phoenix.gen.json.ex
```diff
@@ -65,7 +65,8 @@ defmodule Mix.Tasks.Phoenix.Gen.Json do
     end
   end
 
-  defp validate_args!([_, plural | _] = args) do
+  defp validate_args!([singular, plural | _] = args) do
+    Mix.Phoenix.check_module_name_availability!(singular)
```
josevalim: Same here. The generate modules end with "Controller" and "View".

lib/mix/tasks/phoenix.gen.channel.ex
josevalim: Also, the generate module is almost always namespaced and the namespace is not considered here.

lib/mix/tasks/phoenix.gen.model.ex
```diff
@@ -70,7 +70,8 @@ defmodule Mix.Tasks.Phoenix.Gen.Model do
     ]
   end
 
-  defp validate_args!([_, plural | _] = args) do
+  defp validate_args!([singular, plural | _] = args) do
+    Mix.Phoenix.check_module_name_availability!(singular)
```
josevalim: Same here, too soon, it is not considering the namespace.

test/mix/tasks/phoenix.gen.channel_test.exs
```diff
@@ -76,4 +76,10 @@ defmodule Mix.Tasks.Phoenix.Gen.ChannelTest do
       Mix.Tasks.Phoenix.Gen.Channel.run ["Admin.Room", "rooms", "new_message"]
     end
   end
+
+  test "name can't already be defined" do
+    assert_raise Mix.Error, fn ->
+      Mix.Tasks.Phoenix.Gen.Channel.run ["String", "strings"]
```
josevalim: So this should work because `MyApp.StringChannel` would be generated.

josevalim:
Thanks @doomspork!  Unfortunately we are checking at the wrong place, I think we should check the arguments after we call inflect. :) Thoughts?

doomspork:
@josevalim thanks for the feedback!  My thinking was to exit as soon as possible for keywords but you've raise good points, I will make those changes today.

lib/mix/tasks/phoenix.gen.channel.ex
doomspork: @josevalim does the namespace need to be considered?  In issue #921, even within the `Throwaway` namespace an error is still raised:

```elixir
(UndefinedFunctionError) undefined function: Throwaway.Repo.start_link/0
```

Apologizes in advance if I'm confused :smile:, thanks for your guidance.

lib/mix/tasks/phoenix.gen.model.ex
doomspork: @josevalim does the namespace need to be considered?  In issue #921, even within the `Throwaway` namespace an error is still raised:

```elixir
(UndefinedFunctionError) undefined function: Throwaway.Repo.start_link/0
```

Apologizes in advance if I'm confused :smile:, thanks for your guidance.

lib/mix/tasks/phoenix.gen.html.ex
doomspork: Sounds like this check should only occur in the Model generator where it is not appended with `View`, `Controller`, or `Channel`.  Thoughts?

test/mix/tasks/phoenix.gen.channel_test.exs
doomspork: I'll remove the checks/tests for `Channel` since it does not appear to suffer from keyword problem..

lib/mix/tasks/phoenix.gen.html.ex
josevalim: It doesn't work even for the model generator because we need to consider the namespace. If we don't consider the namespace, we won't even detect the case in #921 as the conflict there is effectively with a namespace module.

lib/mix/tasks/phoenix.gen.channel.ex
```diff
@@ -25,6 +25,8 @@ defmodule Mix.Tasks.Phoenix.Gen.Channel do
 
     binding = binding ++ [plural: plural]
 
+    Mix.Phoenix.check_module_name_availability!(binding[:module])
```
doomspork: @josevalim is this what you had in mind?  Thank you very much for your feedback.

josevalim: Almost! You need to add the suffix here:

     Mix.Phoenix.check_module_name_availability!(binding[:module] <> "Channel")

lib/mix/tasks/phoenix.gen.html.ex
```diff
@@ -39,6 +39,8 @@ defmodule Mix.Tasks.Phoenix.Gen.Html do
     binding = binding ++ [plural: plural, route: route,
                           inputs: inputs(attrs), params: Mix.Phoenix.params(attrs)]
 
+    Mix.Phoenix.check_module_name_availability!(binding[:module])
```
josevalim: This should be called twice:

     Mix.Phoenix.check_module_name_availability!(binding[:module] <> "Controller")
     Mix.Phoenix.check_module_name_availability!(binding[:module] <> "View")

lib/mix/tasks/phoenix.gen.json.ex
```diff
@@ -37,6 +37,8 @@ defmodule Mix.Tasks.Phoenix.Gen.Json do
     route   = String.split(path, "/") |> Enum.drop(-1) |> Kernel.++([plural]) |> Enum.join("/")
     binding = binding ++ [plural: plural, route: route, params: Mix.Phoenix.params(attrs)]
 
+    Mix.Phoenix.check_module_name_availability!(binding[:module])
```
josevalim: Twice too!

lib/mix/tasks/phoenix.gen.model.ex
```diff
@@ -56,6 +56,8 @@ defmodule Mix.Tasks.Phoenix.Gen.Model do
     path      = binding[:path]
     migration = String.replace(path, "/", "_")
 
+    Mix.Phoenix.check_module_name_availability!(binding[:module])
```
josevalim: This one is correct.

test/mix/tasks/phoenix.gen.channel_test.exs
```diff
@@ -76,4 +76,10 @@ defmodule Mix.Tasks.Phoenix.Gen.ChannelTest do
       Mix.Tasks.Phoenix.Gen.Channel.run ["Admin.Room", "rooms", "new_message"]
     end
   end
+
+  test "name can't already be defined" do
+    assert_raise Mix.Error, fn ->
+      Mix.Tasks.Phoenix.Gen.Channel.run ["Channel", "channels"]
```
josevalim: If you call it with "Channel", it is going to generate `Phoenix.ChannelChannel` so this test is passing by accident because you are checking the wrong thing. :) Once you do the changes above, this test should fail again.

lib/mix/tasks/phoenix.gen.model.ex
doomspork: I got one right! :tada:

test/mix/tasks/phoenix.gen.channel_test.exs
doomspork: You are correct, this test now fails with the updated changes.  Any recommendations on how best to test this?

josevalim: You will have to define dummy modules at the top of the file. I would call them "Dup". `Phoenix.DupChannel`, `Phoenix.DupHTMLView`, `Phoenix.DupHTMLController`, etc.

doomspork: Great, that's what I was thinking but I wasn't sure if it was kosher.  Appreciate all your feedback @josevalim, dziękuję!

doomspork:
Thanks for all of your feedback @josevalim, I believe I have incorporated it all.

test/mix/tasks/phoenix.gen.model_test.exs
```diff
@@ -1,5 +1,8 @@
 Code.require_file "../../../installer/test/mix_helper.exs", __DIR__
 
+defmodule Phoenix.Dup do
```
doomspork: I created a `Dup` model for consistency sake.

doomspork:
Let me know if you have any additional feedback @josevalim, thanks again.

josevalim:
Thank you! Can you please squash your commits together? Then this is ready to go! :heart:

doomspork:
Squashed!  Thanks @josevalim, you da :bomb:.

doomspork:
@josevalim any ideas why this is failing on Travis?  It passes for me with the ` 881247` seed.

josevalim:
There are some race conditions in the test suite when we are doing integration tests, i will take a look at those later.

josevalim:

PR #930, opened 2015-05-30
Removes some duplication in the installer web template

dan-palmer:
Removes some duplication around the `def model` function in the installer web template & fixes some indentation.


josevalim:
The reason it looks the way it does is to avoid trailing whitespace on the generated file. We can clean this up on Elixir v1.1  which supports EEx trimming but we can't right now.

PR #935, opened 2015-06-02
deprecate resource in favor of using resources with singleton: true

jxs:


chrismccord:
@jxs Thanks! We need to keep `resource` and deprecate it only in this next release. Can you add it back and warn when called? We will remove it in a future release.

jxs:
ok sure :) btw how do you want me to warn, `IO.puts`? should I also state on doc that has been depcreated?
also do you want me to duplicate `resource_tests.ex` to `resources_tests.ex`, or leave it as it is where i just renamed `resource`  to `resources` on `resource_tests.ex`

lib/phoenix/router.ex
```diff
@@ -464,75 +464,50 @@ defmodule Phoenix.Router do
...
-
-  ## Options
-
-  This macro accepts the same options as `resources/4`
-
-  """
-  defmacro resource(path, controller, opts, do: nested_context) do
-    add_resource path, controller, opts, do: nested_context
+  defmacro resources(path, controller, opts, do: nested_context) do
+    if opts[:singleton] do
```
josevalim: Unfortunately this won't work because you are expecting `opts[:singleton]` to be available at compile time. So if someone calls as:

    opts = [singleton: true]
    resources "/users", UserController, opts

It wouldn't work. We need to move it to inside the functions. We should probably rename `Resource.plural` and `Resource.singular` to a single function called `Resource.build` and then choose which functions to invoke.

josevalim:
For the deprecation:

```elixir
stacktrace = __ENV__ |> Macro.Env.stacktrace |> Exception.format_stacktrace
IO.write :stderr, "[deprecate] resource/3 in Phoenix router is deprecated, please use " <>
                resources/3 with the singleton option instead.\n" <> stacktrace
```

jxs:
thanks @josevalim !
updated it,  moved the `singleton: true` check to the `Resource.build` function, and then added `singular` field to `%Resource`, to be checked then on the `add_resource` function on `router.ex` what do you think?

josevalim:

PR #937, opened 2015-06-03
Subscribers function returns a list not a hashset

larshesel:
This fixes the documentation of the Phoenix.PubSub.Local.subscribers function, so it is documented what is actually returned (list instead of hashset).

Is there a reason for converting the hashset to a list? I'd prefer to just get the hashset. Also converting the hashset to a list might be very expensive.

### DIFF
--- a/lib/phoenix/pubsub/local.ex
+++ b/lib/phoenix/pubsub/local.ex
@@ -82,7 +82,7 @@ defmodule Phoenix.PubSub.Local do
   ## Examples
 
       iex> subscribers(:local_server, "foo")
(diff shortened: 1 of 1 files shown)


josevalim:
We should just say it returns an `Enumerable.t`. But yeah, there is no need to convert it to a list.

larshesel:
Ok - I'll prepare a PR changing it to just the HashSet if you like

josevalim:
Please do!

larshesel:
Replaced this by PR #938

PR #938, opened 2015-06-03
Fix local subscribers return value

larshesel:
Return the HashSet directly instead of converting it to a list. The
conversion is not really needed and will be expensive if the number of
subscribers is large.

Updated the function doc so we say we return a `Enumerable.t` instead of the concrete HashSet type.

This replaces PR #937


larshesel:
@josevalim is it normal that that test fails?

Issue #942, opened 2015-06-04
phoenix.new creates faulty page when --no-brunch is used


pb82:
Not critical, but using the `--no-brunch` switch with `phoenix.new` results in a javascript error:

```
ReferenceError: require is not defined
<script>require("web/static/js/app")</script>
```
I'm using Phoenix 0.13.1

Steps to reproduce:

1. Create project with `mix phoenix.new testapp --no-brunch`
2. Start the server and open localhost:4000 in a browser
3. You will see the javascript error in the browser console

The reason is that without brunch, phoenix.js is not merged into app.js and thus `require` is not available. app.js is an empty file.

Proposed solution:

When not using brunch the layout template (application.html.eex) should explicitely load phoenix.js and provide a stub app.js. The question is what this stub should contain: maybe register an empty module named `app`?

Gazler:
I'll take a stab at this one - not generating the require line should prevent this issue, but still let you use app.js from `priv`

pb82:
Thanks for looking into this! Not generating the require will suppress the javascript error. But then phoenix.js is still not available without brunch. What about including the line

```html
<script src="<%= static_path(@conn, "/js/phoenix.js") %>"></script>
```

in the layout template? Then `require` is also available.

PR #944, opened 2015-06-04
fix(installer): do not attempt to use require without brunch

Gazler:
If the `--no-brunch` flag is set when creating a new project using
installer, e.g.

    mix phoenix.new /tmp/testapp --no-brunch

The line that is responsible for running the JavaScript for assets using
brunch is not included.

This prevents the JavaScript error:

 > ReferenceError: require is not defined

This closes issue #942

### DIFF
--- a/installer/templates/new/web/templates/layout/app.html.eex
+++ b/installer/templates/new/web/templates/layout/app.html.eex
@@ -27,6 +27,8 @@
 
     </div> <!-- /container -->
     <script src="<%%= static_path(@conn, "/js/app.js") %>"></script>
+    <%= if brunch do %>
     <script>require("web/static/js/app")</script>
+    <% end %>
   </body>
 </html>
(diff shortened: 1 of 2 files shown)


installer/templates/new/web/templates/layout/app.html.eex
```diff
@@ -27,6 +27,8 @@
 
     </div> <!-- /container -->
     <script src="<%%= static_path(@conn, "/js/app.js") %>"></script>
-    <script>require("web/static/js/app")</script>
+    <%= if brunch do %>
+      <script>require("web/static/js/app")</script>
```
chrismccord: Let's not indent here and let's make it so there is no extra white space with the eex `if`

Gazler:
@chrismccord That thought did cross my mind, I guess it is better for the template to look a bit weird than the generated layout. It has been updated now.

chrismccord:
<3 <3 <3

PR #948, opened 2015-06-05
Clearer error msg when module passed to route is not a Phoenix Controller

gjaldon:
I accidentally passed a non-controller module to a route and saw the error raised wasn't helpful to beginners `** (UndefinedFunctionError) undefined function: MyApp.User.init/1`. Replaced it with a clearer error msg if the module passed isn't a Plug.


josevalim:
Unfortunately this is going to affect performance on all requests. :( I don't have any better idea right now but I believe this is not the way to go.

chrismccord:
We could try to check at compile time, but it's tricky. For now I think we'll have to punt on this issue. Thanks for tanking a look

gjaldon:
Will a try-catch approach be slow too?

Issue #957, opened 2015-06-10
Mix generated controllers spits out 500 instead of 404 error when things aren't found


foxnewsnetwork:
Mix generated json (and html) controllers for common resourceful functions such as delete look like this:
```elixir
def delete(conn, %{"id" => id}) do
  camera = Repo.get(Camera, id)

  camera = Repo.delete(camera)
  render(conn, "show.json", camera: camera)
end
```
so if you hit this function with an id not in your database, `camera` is nil which causes `Repo.delete` to throw a 500 error. In my opinion, a 404 error would be more appropriate here. Therefore, I would like to make a pull request to change the template controllers living in `priv/templates/json/controller.ex` from using `Repo.get` to using `Repo.get!`... but is this a good idea?

josevalim:
Great catch. We should defintiely use `Repo.get!`. Pull request please? :)

gjaldon:
Sounds like a good idea to me :)

foxnewsnetwork:
Cool, I will pr tonight after work

PR #959, opened 2015-06-11
Switch to use Repo.get! to properly 404

foxnewsnetwork:
In reference to this issue: https://github.com/phoenixframework/phoenix/issues/957

I switched from using `get` to `get!` then I added some tests in the mix task section and everything worked, which was great. 

Next, I got a little ambitious and tried to use the modified mix tasks to generate controllers so I could make sure they would work. This step wasn't so great because I couldn't get it to work without resorting to using a weird `ln -S` trick in my `~/.mix/phoenix-0.14.0-dev` dir (lol it's very likely I just didn't install the .ez archive correctly, but please forgive, I'm new to elixir).

Next, I used the 0.14-dev phoenix to start a new `hello_phoenix` project to test. I ran `mix phoenix.gen.json User users name:string email:string` and then ran `mix test`

I was greeted by:

```console
** (CompileError) test/controllers/user_controller_test.exs:14: function user_path/2 undefined
    (stdlib) lists.erl:1336: :lists.foreach/2
    (stdlib) erl_eval.erl:657: :erl_eval.do_apply/6
```
Alarmed that such a small change in what I thought were unrelated parts could cause this error, I uninstalled phoenix-0.14.0-dev and went back to 0.13 to try again...

~~And got the same thing with user_path being undefined... so unless I'm going crazy, scaffolded tests aren't working (I will confirm this sometime tomorrow and file an issue if I'm not crazy).~~

edit: yeah, it was me, I forgot to follow instructions and edit my router.ex file

### DIFF
--- a/priv/templates/html/controller.ex
+++ b/priv/templates/html/controller.ex
@@ -30,18 +30,18 @@ defmodule <%= module %>Controller do
   end
 
   def show(conn, %{"id" => id}) do
-    <%= singular %> = Repo.get(<%= alias %>, id)
+    <%= singular %> = Repo.get!(<%= alias %>, id)
     render(conn, "show.html", <%= singular %>: <%= singular %>)
   end
 
   def edit(conn, %{"id" => id}) do
-    <%= singular %> = Repo.get(<%= alias %>, id)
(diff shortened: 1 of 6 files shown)


josevalim:

PR #963, opened 2015-06-13
Add docs for custom dispatch options

darthdeus:
As per discussion on IRC, this adds a bit of docs on how to setup custom dispatch for cowboy handlers.

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -212,6 +212,28 @@ defmodule Phoenix.Endpoint do
     * `call(conn, opts)` - invoked on every request (simply dispatches to
       the defined plug pipeline)
 
+  ## Custom dispatch options
+
+  You can provide custom dispatch options in order to use Phoenix's builtin
+  cowboy server with a custom cowboy handler (for example to handle raw WebSockets
+  [as shown in cowboy's examples](https://github.com/ninenines/cowboy/tree/1.0.x/examples)).
+
+  The options are passed to both `:http` and `:https` keys in the endpoint
+  config section. In order to perserve the default dispatch of Phoenix's router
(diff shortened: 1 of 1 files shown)


josevalim:
Thank you! While I definitely think this should be documented, my trouble with this is that it is highly dependent on the Cowboy API which may even change with Cowboy 2.0. So we need to be clear this is a last resort thing. Maybe there is somewhere more "hidden" we could document it? Or at least add a disclaimer it depends on the Cowboy API which may change in the future?

chrismccord:
I will merge and edit the comments to be less pronounced and will clear caveats. @josevalim I agree with your points, but we need it doc'd somewhere because I had to go digging/experimenting to properly answer @darthdeus need.

darthdeus:
@josevalim I've added a disclaimer to the docs :)

lib/phoenix/endpoint.ex
```diff
@@ -212,6 +212,31 @@ defmodule Phoenix.Endpoint do
...
+
+  *Disclaimer: This feature depends on the internals of cowboy 1.x API, which
+  might change in the future when cowboy 2.x is released.*
+
+  You can provide custom dispatch options in order to use Phoenix's builtin
+  cowboy server with a custom cowboy handler (for example to handle raw WebSockets
+  [as shown in cowboy's examples](https://github.com/ninenines/cowboy/tree/1.0.x/examples)).
+
+  The options are passed to both `:http` and `:https` keys in the endpoint
+  config section. In order to perserve the default dispatch of Phoenix's router
```
pel-daniel: There is a typo in perserve -> preserve

darthdeus: fixed

chrismccord:
<3 <3 <3

Issue #971, opened 2015-06-19
Improve README


josevalim:
README should link to the guides and resources.

We should also consider including deploy buttons on the README to make it really easy to get up and running. Many platforms support them, we could get started with Heroku.

gjaldon:
I'd like to take this on. What sample app should we deploy to Heroku? Just a new basic Phoenix app or should we add some basic CRUD stuff and features that make use of channels?

josevalim:
Sorry, I was not super precise. We need to improve the README of the *generated* application. Heroku supports deploy to Heroku buttons, which we need to figure out how they work too.

gjaldon:
My bad. That makes more sense now. Will have a PR ready for this over the weekend. Looked up the [Heroku deploy button](https://devcenter.heroku.com/articles/heroku-button) and it looks simple enough. 

Looking at the [app.json schema](https://devcenter.heroku.com/articles/app-json-schema#example-app-json), not sure if it supports multiple buildpacks and not yet clear to me how to dynamically generate the user's "repository". I'm roughly thinking asking for the user's github name when creating a new app and encouraging them to create a `"https://github.com/username/myapp"` repo. Or add instructions in the readme to put in their app's repo name in `app.json`. The latter one looks like the winner to me.

josevalim:
Definitely the latter. :)

-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

josevalim:
Thanks @gjaldon! After some thought, we have decided to not ship with this in favor of Phoenix guides. I will be bringing some of the improvements here over there though! Thank you very much!

gjaldon:
You're welcome! :)

PR #979, opened 2015-06-23
Add Heroku button to generated app's README

gjaldon:
Closes https://github.com/phoenixframework/phoenix/issues/971

Apart from adding files that are needed by the Heroku deploy button, I also removed `prod.secret.exs` in favor of using ENV variables inside `prod.exs`. Should we expose a `random_string` function so users have an easy way to generate their `SECRET_KEY_BASE` for prod? I'm thinking yes, but not sure which module to put it in. 

Looking forward to your feedback!

### DIFF
--- a/installer/lib/phoenix_new.ex
+++ b/installer/lib/phoenix_new.ex
@@ -12,7 +12,6 @@ defmodule Mix.Tasks.Phoenix.New do
     {:eex,  "new/config/config.exs",                         "config/config.exs"},
     {:eex,  "new/config/dev.exs",                            "config/dev.exs"},
     {:eex,  "new/config/prod.exs",                           "config/prod.exs"},
-    {:eex,  "new/config/prod.secret.exs",                    "config/prod.secret.exs"},
     {:eex,  "new/config/test.exs",                           "config/test.exs"},
     {:eex,  "new/lib/application_name.ex",                   "lib/application_name.ex"},
     {:eex,  "new/lib/application_name/endpoint.ex",          "lib/application_name/endpoint.ex"},
@@ -35,6 +34,8 @@ defmodule Mix.Tasks.Phoenix.New do
     {:eex,  "new/web/web.ex",                                "web/web.ex"},
     {:eex,  "new/mix.exs",                                   "mix.exs"},
     {:eex,  "new/README.md",                                 "README.md"},
+    {:eex,  "new/app.json",                                  "app.json"},
+    {:text, "new/.buildpacks",                               ".buildpacks"},
   ]
 
   @ecto [
@@ -153,7 +154,6 @@ defmodule Mix.Tasks.Phoenix.New do
                phoenix_dep: phoenix_dep(dev),
                pubsub_server: pubsub_server,
                secret_key_base: random_string(64),
-               prod_secret_key_base: random_string(64),
                signing_salt: random_string(8),
                in_umbrella: in_umbrella?(path),
                brunch: brunch,
@@ -222,14 +222,12 @@ defmodule Mix.Tasks.Phoenix.New do
         size: 1
       """
 
-      append_to path, "config/prod.secret.exs", """
+      append_to path, "config/prod.exs", """
 
       # Configure your database
       config :#{binding[:application_name]}, #{binding[:application_module]}.Repo,
         adapter: #{inspect binding[:adapter_module]},
-        username: #{inspect binding[:db_user]},
-        password: #{inspect binding[:db_password]},
-        database: "#{binding[:application_name]}_prod",
+        url: System.get_env("DATABASE_URL"),
         size: 20 # The amount of database connections in the pool
       """
     end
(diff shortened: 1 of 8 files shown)


josevalim:
Awesome!

The idea though is to not make Phoenix exclusive to Heroku. For example, we should still keep the database configuration and `config/prod.secret.exs` as before. That will hurt running Heroku out of the box but we could have a small set of instructions:

1. Remove `config/prod.secret.exs` from .gitignore

2. Replace the important bits by environment variables

3. Deploy!

If we could support both at the same time even better. For example, we could have `url: System.get_env(...)` by default in production as long as it does not affect the regular code.

installer/templates/new/app.json
```diff
@@ -0,0 +1,19 @@
+{
+  "name": "Phoenix",
+  "description": "Elixir Web Framework targeting full-featured, fault tolerant applications with realtime functionality",
+  "website": "http://www.phoenixframework.org/",
+  <%= if ecto do %>"scripts": {
+    "postdeploy": "mix ecto.migrate"
+  },<% end %>
+  "env": {
+    "BUILDPACK_URL": "https://github.com/ddollar/heroku-buildpack-multi",
```
josevalim: @tsloughter do we have support for multiple buildpacks in app.json too?

tsloughter: Argh, I just asked and you can't yet.

josevalim: Do you know if there is a timeline? Who can we bother? :P

gjaldon: I asked on heroku channel in freenode and a Heroku dev responded saying he had a pending PR he is waiting to be reviewed. Forgot his name though. :D So there's a pending PR but he said no definite date on when it will be shipped.

tsloughter: There is a PR. Still has to be tested more.

tsloughter: @gjaldon was it Owen? He is the one handling it.

tsloughter: May be done this week.

gjaldon: Ah yes, it was Owen. :) Thanks for the update!

gjaldon:
Was initially thinking of removing `config/prod.secret.exs` and users who don't deploy to Heroku set the DATABASE_URL and SECRET_KEY_BASE env vars in their prod env. But I guess, copying `prod.secret.exs` to their prod is more convenient? 

Anyway, will try to make it more convenient for user to deploy on Heroku or otherwise.

gjaldon:
@josevalim - this now provides defaults for secret_key and database_url through DATABASE_URL and SECRET_KEY_BASE env vars. Config for those will be overridden though if user copies prod.secret.exs to their prod server.

For heroku deploy button to work, user needs to comment out `import_config "prod.secret.exs"` in prod.exs. Let me know if there's anything else that needs done here.

josevalim:
@gjaldon Sorry for not being clear. I don't think we should have the configuration twice. I was saying we should do:

    config :foo, Foo.Repo,
      url: System.get_env("DATABASE_URL"),
      other_options

This will be fine as long as we guarantee the ones in the URL in (which I am not sure they do). :)

installer/templates/new/app.json
tsloughter: It is out: https://devcenter.heroku.com/articles/app-json-schema#buildpacks

gjaldon:
@josevalim updated the instructions and removed duplicate config. I also removed multi-buildpack and used the newly implemented buidpacks opt in app.json.

installer/lib/phoenix_new.ex
```diff
@@ -226,6 +227,7 @@ defmodule Mix.Tasks.Phoenix.New do
       # Configure your database
       config :#{binding[:application_name]}, #{binding[:application_module]}.Repo,
         adapter: #{inspect binding[:adapter_module]},
+        url: System.get_env("DATABASE_URL"),
```
josevalim: Cool! Have you checked if the URL configuration overrides the regular one?

installer/templates/new/app.json
```diff
@@ -0,0 +1,22 @@
+{
+  "name": "Phoenix",
+  "description": "Elixir Web Framework targeting full-featured, fault tolerant applications with realtime functionality",
+  "website": "http://www.phoenixframework.org/",
+  <%= if ecto do %>"scripts": {
+    "postdeploy": "mix ecto.migrate"
+  },<% end %>
+  "env": {
+    "SECRET_KEY_BASE": {
```
josevalim: Should we ask users to set this somehow?

installer/lib/phoenix_new.ex
gjaldon: I checked and found that if url is nil, the regular ones will be used. If url is present and is valid, it will override the regular one.

installer/templates/new/app.json
gjaldon: This automatically generates a random secret and adds it to Heroku as a SECRET_KEY_BASE. User doesn't need to set it. I added a `System.get_env("SECRET_KEY_BASE")` to prod.secret.exs. 

By the way, also tested this from this https://github.com/gjaldon/test. I also set it to private and added a template parameter and it worked fine too.

installer/templates/new/README.md
```diff
@@ -6,3 +6,16 @@ To start your new Phoenix application:
 2. Start Phoenix endpoint with `mix phoenix.server`
 
 Now you can visit `localhost:4000` from your browser.
+
+# Deploying your App to Heroku
+
+[![Deploy](https://www.herokucdn.com/deploy/button.png)](https://heroku.com/deploy)
+
+The deploy button above allows you to instantly deploy your applications to Heroku from your Github
```
henrik: Github -> GitHub I believe.

PR #983, opened 2015-06-24
Add forward macro to router

chrismccord:
@josevalim I could use a code review on this one

### DIFF
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -182,7 +182,7 @@ defmodule Phoenix.Router do
     quote do
       unquote(prelude())
       unquote(defs())
-      unquote(match_dispatch(__CALLER__))
+      unquote(match_dispatch())
     end
   end
 
@@ -190,6 +190,7 @@ defmodule Phoenix.Router do
     quote do
       Module.register_attribute __MODULE__, :phoenix_routes, accumulate: true
       Module.register_attribute __MODULE__, :phoenix_channels, accumulate: true
+      @phoenix_forwards %{}
 
       import Phoenix.Router
       import Plug.Conn
@@ -254,19 +255,7 @@ defmodule Phoenix.Router do
     end
   end
 
-  defp match_dispatch(env) do
-    plugs = [{:dispatch, [], true}, {:match, [], true}]
-    {conn, pipeline} = Plug.Builder.compile(env, plugs, [])
-
-    call =
-      quote do
-        unquote(conn) =
-          update_in unquote(conn).private,
-            &(&1 |> Map.put(:phoenix_router, __MODULE__)
-                 |> Map.put(:phoenix_pipelines, []))
-        unquote(pipeline)
-      end
-
+  defp match_dispatch() do
     quote location: :keep do
       @behaviour Plug
 
@@ -281,9 +270,7 @@ defmodule Phoenix.Router do
       @doc """
       Callback invoked by Plug on every request.
       """
-      def call(unquote(conn), opts) do
-        unquote(call)
-      end
+      def call(conn, opts), do: do_call(conn, opts)
 
       defp match(conn, []) do
         match(conn, conn.method, Enum.map(conn.path_info, &URI.decode/1), conn.host)
@@ -310,7 +297,24 @@ defmodule Phoenix.Router do
     Helpers.define(env, routes)
     escaped = Enum.map(routes, fn {route, _} -> Macro.escape(route) end)
 
+    plugs = [{:dispatch, [], true}, {:match, [], true}]
+    {conn, pipeline} = Plug.Builder.compile(env, plugs, [])
+
+    call =
+      quote do
+        unquote(conn) =
+          update_in unquote(conn).private,
+            &(&1 |> Map.put(:phoenix_pipelines, [])
+                 |> Map.put(:phoenix_router, __MODULE__)
+                 |> Map.put(__MODULE__, {unquote(conn).script_name, @phoenix_forwards}))
+        unquote(pipeline)
+      end
+
     quote do
+      defp do_call(unquote(conn), opts) do
+        unquote(call)
+      end
+
       @doc false
       def __routes__,  do: unquote(escaped)
 
@@ -775,12 +779,25 @@ defmodule Phoenix.Router do
       pipe_through [:browser, :admin]
 
       forward "/admin", SomeLib.AdminDashboard
-      forward "/api/:version", ApiRouter
+      forward "/api", ApiRouter
     end
 
   """
   defmacro forward(path, plug, plug_opts \\ [], router_opts \\ []) do
-    add_route(:forward, :*, path, plug, plug_opts, router_opts)
+    quote unquote: true, bind_quoted: [path: path, plug: plug] do
+      case Plug.Router.Utils.build_path_match(path) do
+        {[], path_segments} ->
+          if @phoenix_forwards[plug] do
+            raise "`#{inspect plug}` has already been forwarded to. A module can only be forwarded a single time."
+          end
+          @phoenix_forwards Map.put(@phoenix_forwards, plug, path_segments)
+        _ ->
+          raise """
+          Dynamic segment `"#{path}"` not allowed when forwarding. Use a static path instead.
+          """
+      end
+      unquote(add_route(:forward, :*, path, plug, plug_opts, router_opts))
+    end
   end
 
 end
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -213,7 +213,7 @@ defmodule Phoenix.Router do
         exprs = Route.exprs(route)
         @phoenix_routes {route, exprs}
 
-        defp match(var!(conn), unquote(route.verb), unquote(exprs.path),
+        defp match(var!(conn), unquote(exprs.verb_match), unquote(exprs.path),
                    unquote(exprs.host)) do
           unquote(exprs.dispatch)
         end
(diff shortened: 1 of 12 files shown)


chrismccord:
@josevalim just committed discussed changes. Please take a look!

lib/phoenix/router.ex
```diff
@@ -760,4 +763,41 @@ defmodule Phoenix.Router do
...
+      pipe_through [:browser, :admin]
+
+      forward "/admin", SomeLib.AdminDashboard
+      forward "/api", ApiRouter
+    end
+
+  """
+  defmacro forward(path, plug, plug_opts \\ [], router_opts \\ []) do
+    quote unquote: true, bind_quoted: [path: path, plug: plug] do
+      case Plug.Router.Utils.build_path_match(path) do
```
josevalim: Let's move this to to `Phoenix.Router.Scope` or something similar.

lib/phoenix/router/console_formatter.ex
```diff
@@ -24,17 +25,21 @@ defmodule Phoenix.Router.ConsoleFormatter do
...
     String.ljust(verb, verb_len) <> "  " <>
     String.ljust(path, path_len) <> "  " <>
-    inspect(controller) <> "." <> Atom.to_string(action) <> "/2\n"
+    "#{inspect(plug)} #{inspect(opts)}\n"
   end
 
   defp route_name(nil),  do: ""
   defp route_name(name), do: name <> "_path"
+
+  defp verb_name(:*), do: "*"
```
josevalim: We don't need this clause!

lib/phoenix/router/helpers.ex
```diff
@@ -8,6 +8,41 @@ defmodule Phoenix.Router.Helpers do
...
+      :error -> nil
+    end
+  end
+
+  defp build_conn_forward_path(conn, router, path) do
+    case Map.fetch(conn.private, conn.private[:phoenix_router]) do
+      {:ok, {_script_name, forwards}} ->
+        case Map.fetch(forwards, router) do
+          {:ok, local_script} ->
+            path_with_script(path, conn.script_name ++ local_script)
```
josevalim: `conn.script_name` here is the same as `_script_name`, so you can simply use the former.

lib/phoenix/router/helpers.ex
```diff
@@ -8,6 +8,41 @@ defmodule Phoenix.Router.Helpers do
...
+  defp build_own_forward_path(conn, router, path) do
+    case Map.fetch(conn.private, router) do
+      {:ok, {local_script, _}} ->
+        path_with_script(path, local_script)
+      :error -> nil
+    end
+  end
+
+  defp build_conn_forward_path(conn, router, path) do
+    case Map.fetch(conn.private, conn.private[:phoenix_router]) do
```
josevalim: In case `conn.private[:phoenix_router]` returns nil and there is a cruel soul that stores nil in `conn.private`, I would explicitly check if phoenix_router is actually set. You can even pattern match on it in the function head.

lib/phoenix/router/route.ex
```diff
@@ -96,16 +108,39 @@ defmodule Phoenix.Router.Route do
...
 
+  defp build_pipes(%Route{kind: :forward} = route) do
+    {_params, path_segments} = Plug.Router.Utils.build_path_match(route.path)
+
+    quote do
+      var!(conn)
+      |> Plug.Conn.put_private(:phoenix_pipelines, unquote(route.pipe_through))
+      |> Plug.Conn.put_private(:phoenix_route, fn conn ->
+        opts = unquote(route.plug).init(unquote(route.opts))
+        mount = conn.path_info -- unquote(path_segments)
```
josevalim: Why don't you do this operation inside forward?

test/phoenix/router/console_formatter_test.exs
```diff
@@ -33,31 +33,35 @@ defmodule Phoenix.Router.ConsoleFormatterTest do
...
-    image_path  DELETE  /image       Phoenix.ImageController.delete/2
+    image_path  GET     /image/edit  Phoenix.ImageController :edit
+    image_path  GET     /image/new   Phoenix.ImageController :new
+    image_path  GET     /image       Phoenix.ImageController :show
+    image_path  POST    /image       Phoenix.ImageController :create
+    image_path  PATCH   /image       Phoenix.ImageController :update
+                PUT     /image       Phoenix.ImageController :update
+    image_path  DELETE  /image       Phoenix.ImageController :delete
+    admin_path  *       /admin       Phoenix.Admin.Router []
+       f1_path  *       /f1          Phoenix.F1 []
```
josevalim: I thought those had no paths? Why are they showing up? :)

test/phoenix/router/forward_test.exs
```diff
@@ -0,0 +1,144 @@
...
+  test "forwards any request starting with forward path" do
+    conn = call(Router, :get, "admin/stats")
+    assert conn.script_name == []
+    assert conn.assigns[:fwd_script_name] == ["admin"]
+    assert conn.status == 200
+    assert conn.resp_body == "stats"
+  end
+
+  test "phoenix_main_router is set for the first router" do
+    # conn = call(Router, :get, "internal/api/v1/users")
```
josevalim: ?

lib/phoenix/router/route.ex
```diff
@@ -96,16 +108,39 @@ defmodule Phoenix.Router.Route do
...
+    end |> pipe_through(route)
+  end
+  defp pipe_through(initial, route) do
     Enum.reduce(route.pipe_through, initial, &{&1, [], [&2, []]})
   end
+
+  @doc """
+  Forwards requests to another Plug at a new path.
+  """
+  def forward(%Plug.Conn{path_info: path, script_name: script} = conn, new_path, target, opts) do
```
josevalim: Let's unit test this too?

test/phoenix/router/forward_test.exs
```diff
@@ -0,0 +1,144 @@
+defmodule Phoenix.Router.ForwardTest do
+  use ExUnit.Case, async: true
+  use RouterHelper
+
+  defmodule Controller do
```
josevalim: What I would do is to make the Controller as pass through (it doesn't do anything) and call the helpers inside the test themselves. It is going to make the test clearer because I don't need to look at the controller then at the test to know what is being generated and what I am asserting on.

chrismccord: The issue is the forward process does the script_name dance, so if we call the helpers in the test before/after the request, the conn is not in the context of the forwarded plug?

josevalim: The context is set after you call the router and is completely in the connection. As long as you have the copy of the connection as it entered the controller, you are fine.

lib/phoenix/router/helpers.ex
chrismccord: :+1:

test/phoenix/router/console_formatter_test.exs
chrismccord: Ah good catch

test/phoenix/router/forward_test.exs
chrismccord: I made the changes and the tests are indeed much nicer <3

chrismccord:
@josevalim all feedback addressed. :ship: ?

lib/phoenix/router/helpers.ex
```diff
@@ -8,6 +8,42 @@ defmodule Phoenix.Router.Helpers do
 
   @transports [Phoenix.Transports.WebSocket, Phoenix.Transports.LongPoller]
 
+  @doc false
```
josevalim:
```elixir
@doc """
Callback invoked by path generated in each helper module
"""
```

We could even call it `path` itself to mirror the generated API.

lib/phoenix/router/route.ex
```diff
@@ -43,20 +47,28 @@ defmodule Phoenix.Router.Route do
...
+    {params, segments} = case route.kind do
+      :forward -> Plug.Router.Utils.build_path_match(path <> "/*_forward_path_info")
+      :match   -> Plug.Router.Utils.build_path_match(path)
+    end
 
-    binding = Enum.map(params, fn var ->
-      {Atom.to_string(var), Macro.var(var, nil)}
-    end)
+    binding =
+      params
```
josevalim: Btw, this could be cleaner as a for-comprehension or a List.delete (it doesn't matter though).

lib/phoenix/router/route.ex
```diff
@@ -96,16 +108,59 @@ defmodule Phoenix.Router.Route do
...
+  def forward_path_segments(path, plug, phoenix_forwards) do
+    case Plug.Router.Utils.build_path_match(path) do
+      {[], path_segments} ->
+        if phoenix_forwards[plug] do
+          raise "`#{inspect plug}` has already been forwarded to. A module can only be forwarded a single time."
+        end
+        path_segments
+      _ ->
+        raise """
+        Dynamic segment `"#{path}"` not allowed when forwarding. Use a static path instead.
```
josevalim: Btw, remember this pattern introduces an extra new line at the end of the message.

lib/phoenix/router/route.ex
```diff
@@ -96,16 +108,59 @@ defmodule Phoenix.Router.Route do
...
+  Validates and returns the list of forward path segments.
+
+  Raises RuntimeError plug is already forwarded or path contains
+  a dynamic segment.
+  """
+  def forward_path_segments(path, plug, phoenix_forwards) do
+    case Plug.Router.Utils.build_path_match(path) do
+      {[], path_segments} ->
+        if phoenix_forwards[plug] do
+          raise "`#{inspect plug}` has already been forwarded to. A module can only be forwarded a single time."
```
josevalim: Should be `ArgumentError`.

lib/phoenix/router/route.ex
josevalim: Should be `ArgumentError`.

test/phoenix/router/forward_test.exs
```diff
@@ -0,0 +1,130 @@
...
+    use Phoenix.Controller
+
+    plug :assign_fwd_conn
+
+    def index(conn, _params), do: text(conn, "admin index")
+    def stats(conn, _params), do: text(conn, "stats")
+    def api_users(conn, _params), do: text(conn, "api users")
+    def api_root(conn, _params), do: text(conn, "api root")
+    defp assign_fwd_conn(conn, _), do: assign(conn, :fwd_conn, conn)
```
josevalim: Do you need to do this? Can't you just use the conn return itself? I think the private information in the `conn` return and the one assigned is the same, no?

josevalim:
I have added some minor comments.

test/phoenix/router/forward_test.exs
chrismccord: We need this to test the helpers in the context of a forwarded conn. We store the `fwd_conn` because in the controller context, we have a different path_info and script_name, so the helper will be hitting different code paths.

chrismccord:
All new comments addressed

test/phoenix/router/forward_test.exs
```diff
@@ -0,0 +1,119 @@
+defmodule Phoenix.Router.ForwardTest do
+  use ExUnit.Case, async: true
+  use RouterHelper
+
+  defmodule Controller do
+    use Phoenix.Controller
+
+    plug :assign_fwd_script
```
josevalim: You can remove this plug now. :)

josevalim:
I added one last comment. :shipit:

PR #996, opened 2015-06-29
Add documentation about passing `false` to put_layout/2

caffo:
Just clarifying that `false` can be passed to `put_layout/2` and no layout would be rendered. I wanted to add a doctest but they are currently disabled for this file.


josevalim:

PR #994, opened 2015-07-01
Associations in model generator should use "references" for belongs_to

scrogson:


chrismccord:
:heart: :heart: :heart:

josevalim:
Sorry, that's not what I meant at all. :)

We have a references(...) function in migrations that sets up a foreign
key. We should use it instead of just creating an index column.

Maybe we could rename belongs_to to references though, but this is
certainly not finished. :)


-- 


josevalim:
*instead of creating an *integer* column.

-- 


scrogson:
Ah! I see. I'll fix. Thanks for the clarification.

scrogson:
@josevalim so we've got an issue with using `references` from `Ecto.Migration`.

It expects a pluralized value for the reference, which we don't currently support in `Phoenix.Naming`.

Running `mix phoenix.gen.model Comment comments post:references`:

The migration should have `add :post_id, references(:posts)`

@chrismccord thought of maybe using `post:references:posts`. I'm not really a fan of that since it's probably just going to be difficult to remember all the details.

Thoughts?

josevalim:
Yes, I am aware this would be an issue. What helps is that model is likely defined at this point, so we could try to load it and figure things out. Otherwise we will need to ask the user to give this information (we could raise).

I will open up another issue because this one had a pull request attached to it and I can't reopen it.

PR #1004, opened 2015-07-04
Version channel transport contract

chrismccord:
TODO:
- [ ] Test failure semantics

Some notes:

- Sends vsn param on socket connect
- Default to 1.0.0 for legacy clients
- Raises on incompatible versions (resulting in socket.onError callbacks on client)

Some questions:

Is the naming what we want here? We talk about the "transport adapter contract", but I'm calling this a "protocol version". I feel like I could do better with naming.

Are the failure semantics what we want? The websocket/longpolling adapters raise on init if the client sends an incompatible version. I don't think we can do much else, but the client would only receive a `socket.onError` event, without the reason for the crash. Having the vsn doesn't buy us much now, but down the road it would let us (or a third-party transport) support legacy clients in a forward compatible way if we release v2 of a channels api/contract. All thoughts appreciated. // @josevalim @ericmj @jeregrine @scrogson

### DIFF
--- a/lib/phoenix/channel/transport.ex
+++ b/lib/phoenix/channel/transport.ex
@@ -1,10 +1,4 @@
 defmodule Phoenix.Channel.Transport do
-
-  require Logger
-  alias Phoenix.Socket
-  alias Phoenix.Socket.Message
-
-
   @moduledoc """
   Handles dispatching incoming and outgoing Channel messages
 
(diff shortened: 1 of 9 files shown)


chrismccord:
And one more question: Are we ready to mark the channels protocol as 1.0?

josevalim:
@chrismccord should we now send the version on connect? Also do we need to store the version in the socket struct? it feels unnecessary?

chrismccord:
> should we now send the version on connect? Also do we need to store the version in the socket struct? it feels unnecessary?

I think we should send the version as soon as we are ready to brand a 1.0 socket/transport contract version. I stored it in the socket struct only because later you might want to know about it in a channel for some reason for legacy clients? I can't think of great examples atm though.

josevalim:
If you are in an old version, you should not even connect though? The
version is for our own protocol, we should be the ones managing it and
telling users to upgrade.

>

-- 

PR #1005, opened 2015-07-04
update phoenix.js so it passes jshint linter

mathieul:
[Jshint](http://jshint.com) is really useful to catch errors, and it is setup by default with several frameworks like Ember (with ember-cli). In order to use Phoenix with Ember we currently need to copy the ES2015 original version to the Ember application folder. Which runs jshint on `phoenix.js` and generate warnings.

I've set two Jshint options to allow for not using semi-colons ([asi](http://jshint.com/docs/options/#asi)) and for using assignments in boolean expressions ([boss](http://jshint.com/docs/options/#boss)).

Here are the remaining warnings I've fixed:

```
utils/phoenix.js: line 446, col 17, 'key' is already defined.
utils/phoenix.js: line 540, col 20, 'status' is already defined.
utils/phoenix.js: line 543, col 14, 'status' used out of scope.
utils/phoenix.js: line 545, col 11, 'messages' used out of scope.
utils/phoenix.js: line 561, col 49, 'status' used out of scope.
utils/phoenix.js: line 591, col 23, 'ActiveXObject' is not defined.
utils/phoenix.js: line 167, col 35, 'ref' is defined but never used.
utils/phoenix.js: line 286, col 29, 'ref' is defined but never used.
utils/phoenix.js: line 286, col 20, 'payload' is defined but never used.
utils/phoenix.js: line 286, col 13, 'event' is defined but never used.
utils/phoenix.js: line 540, col 13, 'status' is defined but never used.
utils/phoenix.js: line 536, col 34, 'messages' is defined but never used.
utils/phoenix.js: line 575, col 15, 'reason' is defined but never used.
utils/phoenix.js: line 575, col 9, 'code' is defined but never used.

14 errors
```

If you find Jshint useful it could easily be add to new Phoenix applications using [jshint-brunch](https://github.com/brunch/jshint-brunch).

Thanks again for Phoenix!

### DIFF
--- a/web/static/js/phoenix.js
+++ b/web/static/js/phoenix.js
@@ -1,3 +1,5 @@
+// jshint asi: true, expr: true
+
 // Phoenix Channels JavaScript client
 //
 // ## Socket Connection
@@ -163,7 +165,7 @@ class Push {
 
   // private
 
-  matchReceive({status, response, ref}){
+  matchReceive({status, response}){
     this.recHooks.filter( h => h.status === status )
                  .forEach( h => h.callback(response) )
   }
@@ -282,7 +284,7 @@ export class Channel {
   // Overridable message hook
   //
   // Receives all events for specialized message handling
-  onMessage(event, payload, ref){}
+  onMessage(/* event, payload, ref */){}
 
   // private
 
@@ -441,8 +443,8 @@ export class Socket {
 
   chan(topic, chanParams = {}){
     let mergedParams = {}
-    for(var key in this.params){ mergedParams[key] = this.params[key] }
-    for(var key in chanParams){ mergedParams[key] = chanParams[key] }
+    for(let key in this.params){ mergedParams[key] = this.params[key] }
+    for(let key in chanParams){ mergedParams[key] = chanParams[key] }
 
     let chan = new Channel(topic, mergedParams, this)
     this.channels.push(chan)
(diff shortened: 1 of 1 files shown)


chrismccord:
So I'm not sure about this. What about other tools. jshint vs jslint vs ... ? I'm also not a fan of the way it makes you comment out or remove unused args, when they are clearly significant. Is there any way to give jshint a single line that avoids linting the file entirely?

mathieul:
Afaik jshint is a more flexible (hence not as opinionated) version of jslint. I don't know if we can setup jslint to allow for no semicolons for instance.

I wish we could signal jshint a variable as throw away, like we do with a leading underscore in Elixir, but it is currently not possible (see issue https://github.com/jshint/jshint/issues/1762).

You can tell jshint to ignore a block of code with options `ignore:start` and `ignore:end`, so it could be used to not lint a whole file. But is it worth giving up on all the benefits of the tool when it can be so helpful to catch errors?

chrismccord:
> Afaik jshint is a more flexible (hence not as opinionated) version of jslint. I don't know if we can setup jslint to allow for no semicolons for instance.

Which means this only works for ember's/some peoples' case and for noone running jslint. I think these tools are very nice if they fit your workflow, but unfortunately, it's another thing for me to maintain and make sure I have no jshint warnings.  Because it's not part of my workflow today, I'm :-1: for now. Once we extract the project into its own js project, I may revisit this. Thanks!

mathieul:
Sorry if I was not clear. What it means is that jslint is opiniated and would not allow you to write code without semicolons for instance. Jshint gives you that option by including inline options in a file, or is a resource file to be shared by the project.

I like jshint because it tells you about errors/warnings about your code like a compiler would in a compiled language. It is part of the ember-cli tool chain because like a test suite it is helping you catch bugs in your application. But it is surely not limited to Ember users.

Now I understand you don't want to include it in Phoenix, but I thought I'd clarify things out. Thanks.

PR #1007, opened 2015-07-04
mix.exs template: apps and deps on individual lines

henrik:
How do you feel about this style change?

It arguably reads cleaner since the first and last line is no longer special.

It has some additional minor benefits: cleaner diffs, lets you more easily move lines around (without accidentally introducing syntax errors thanks to the trailing comma on the last line), won't unnecessarily shift the git blame of the previous line.

It does make this file a few lines longer due to the additional line breaks – I don't mind but maybe someone else does?

### DIFF
--- a/installer/templates/new/mix.exs
+++ b/installer/templates/new/mix.exs
@@ -2,25 +2,35 @@ defmodule <%= application_module %>.Mixfile do
   use Mix.Project
 
   def project do
-    [app: :<%= application_name %>,
-     version: "0.0.1",<%= if in_umbrella do %>
-     deps_path: "../../deps",
-     lockfile: "../../mix.lock",<% end %>
-     elixir: "~> 1.0",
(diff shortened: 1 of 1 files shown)


chrismccord:
We match the convention of Elixir and mix generators so my preference is to follow Elixir. I think this is a fine convention if you prefer it. Thanks!

henrik:
@chrismccord Alright, if this is an Elixir convention I can see that you'd want to follow it. Easy enough to change if you feel there are benefits to doing it differently. Thanks for the feedback!

henrik:
@josevalim @ericmj Since Elixir/Mix are still young and plastic, how do you feel about changing this convention in Mix generators? :) I'd feel remiss not to ask since I think it has some benefits.

josevalim:
As one would expect, I actually prefer the current style.  :)

-- 


henrik:
@josevalim It was worth a shot ;) Thank you for taking the time!

PR #1016, opened 2015-07-10
Add #controller_name method to Phoenix.Controller

acconrad:
Added a `#controller_name` function into `Phoenix.Controller`.

Input: `Plug.Conn.t` connection

Output: lowercase name of the controller.

Example:

```elixir
$ Phoenix.Controller.controller_name @conn
=> "page"
```
where `@conn` was generated by the `PageController`.

**Why?**

When dealing with CSS styling, it's useful to be able to separate out styles by page. The best way to get a page is the combination of it's controller and action. So for example, I can specifically style the homepage if I have `.page.index` as the classes to the `<body>` tag of a page. It also helps for modular loading of JS as well as CSS styles. Coming from Rails, it was great to have access to `controller_name` and `action_name`, and I found it strange that Phoenix only had `action_name` but not `controller_name`, and I felt like this would be a good addition to the core functionality of `Phoenix.Controller`.

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -114,12 +114,8 @@ defmodule Phoenix.Controller do
   """
   @spec controller_name(Plug.Conn.t) :: String.t
   def controller_name(conn) do
-    conn.private.phoenix_controller
-    |> Atom.to_string
-    |> String.split(".")
-    |> List.last
-    |> String.replace("Controller", "")
-    |> String.downcase
+    Phoenix.Naming.resource_name(conn.private.phoenix_controller, "Controller")
+    |> String.replace("_", "-")
   end
 
   @doc """
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -115,7 +115,6 @@ defmodule Phoenix.Controller do
   @spec controller_name(Plug.Conn.t) :: String.t
   def controller_name(conn) do
     Phoenix.Naming.resource_name(conn.private.phoenix_controller, "Controller")
-    |> String.replace("_", "-")
   end
 
   @doc """
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -110,6 +110,19 @@ defmodule Phoenix.Controller do
   def action_name(conn), do: conn.private.phoenix_action
 
   @doc """
+  Returns the controller name as a string, raises if unavailable.
+  """
+  @spec controller_name(Plug.Conn.t) :: String.t
+  def controller_name(conn) do
+    conn.private.phoenix_controller
+    |> Atom.to_string
+    |> String.split(".")
+    |> List.last
+    |> String.replace("Controller", "")
+    |> String.downcase
+  end
+
+  @doc """
   Returns the controller module as an atom, raises if unavailable.
   """
   @spec controller_module(Plug.Conn.t) :: atom
--- a/test/phoenix/controller/controller_test.exs
+++ b/test/phoenix/controller/controller_test.exs
@@ -23,6 +23,9 @@ defmodule Phoenix.Controller.ControllerTest do
   test "controller_name/1" do
     conn = put_private(%Conn{}, :phoenix_controller, Hello)
     assert controller_name(conn) == "hello"
+
+    conn_multi = put_private(%Conn{}, :phoenix_controller, HelloWorld)
+    assert controller_name(conn_multi) == "hello-world"
   end
 
   test "controller_module/1" do
--- a/test/phoenix/controller/controller_test.exs
+++ b/test/phoenix/controller/controller_test.exs
@@ -20,6 +20,11 @@ defmodule Phoenix.Controller.ControllerTest do
     assert action_name(conn) == :show
   end
 
+  test "controller_name/1" do
+    conn = put_private(%Conn{}, :phoenix_controller, Hello)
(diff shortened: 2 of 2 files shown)


josevalim:
I think we have a function in Phoenix.Naming that does exactly what you want. It is probably resource_name with "Controller" as second argument.

lib/phoenix/controller.ex
```diff
@@ -110,6 +110,19 @@ defmodule Phoenix.Controller do
   def action_name(conn), do: conn.private.phoenix_action
 
   @doc """
+  Returns the controller name as a string, raises if unavailable.
+  """
+  @spec controller_name(Plug.Conn.t) :: atom
```
josevalim: The spec says it returns an atom but it actually returns a string.

josevalim:
Thanks, I have added some comments. It feels like this would be a good addition but I will leave it up for @chrismccord to decide.

acconrad:
Thank you @josevalim ! I updated the code based on your comments to include the correct datatype returned.

chrismccord:
Thanks @acconrad ! I think this is a welcome addition, but we need to extend it a bit first. For example, it fails for multi-world controllers, i.e.:

```elixir
iex(2)> Phoenix.Controller.controller_name(%{private: %{
  phoenix_controller: Foo.AdminBarController
}})
"adminbar"
```

We would need to dasherize or similar, but I don't want the code to get too complex. We also need to discuss if it should include all "namespaces" back to the root namespace, so MyApp.Admin.BarController would return `"admin-bar"` instead of just `"bar"`. Thoughts? // @josevalim

lib/phoenix/controller.ex
```diff
@@ -110,6 +110,19 @@ defmodule Phoenix.Controller do
   def action_name(conn), do: conn.private.phoenix_action
 
   @doc """
+  Returns the controller name as a string, raises if unavailable.
+  """
+  @spec controller_name(Plug.Conn.t) :: String.t
+  def controller_name(conn) do
+    conn.private.phoenix_controller
+    |> Atom.to_string
```
josevalim: The implementation of this function should be:

    Phoenix.Naming.resource_name(conn.private.private_controller, "Controller")

lib/phoenix/controller.ex
```diff
@@ -110,6 +110,19 @@ defmodule Phoenix.Controller do
   def action_name(conn), do: conn.private.phoenix_action
 
   @doc """
+  Returns the controller name as a string, raises if unavailable.
+  """
+  @spec controller_name(Plug.Conn.t) :: String.t
+  def controller_name(conn) do
```
josevalim: The implementation of this function should be just:

    Phoenix.Naming.resource_name(conn.private.private_controller, "Controller")

josevalim:
@chrismccord the issues you have mentioned are solved by using `Phoenix.Naming.resource_name/2`, as I have originally mentioned.

About namespacing, that is a bit trickier because it would require storing the namespace in the connection too.

acconrad:
@josevalim @chrismccord I've taken your suggestions and updated the code accordingly. For the sake of CSS conventions, I did a simply `String.replace` to "dasherize" the output, as it originally snake_cases multi-word controller names

josevalim:
@chrismccord I don't think we should dasherize. If we want to dasherize, we should move this to Phoenix.HTML as `controller_dom_id` or something of sorts.

chrismccord:
Now that I think about it, dasherize is only for a css specific use-case, and not something we should leak to this function

chrismccord:
@josevalim beats me at the internet again, by 3 seconds, :+1:

acconrad:
good points re: dasherize. perhaps as a compromise, have the `controller_name` function within `Phoenix.Controller` and then `Phoenix.HTML.controller_name` can simply call `Phoenix.Controller.controller_name |> String.replace("_", "-")`?

acconrad:
@josevalim @chrismccord I've reverted the dasherization pipeline and just left it as is. I still think this function would be useful on the backend, but I think it would be worth extending on `Phoenix.HTML` as well, but I have to ask this somewhat naive question - I just poured through the `phoenix_html` project codebase and I couldn't find any `Phoenix` module references outside of `Phoenix.HTML`, so is it even possible for me to call `Phoenix.Naming` inside of that project, and how would I reference a `Plug.Conn.t`? I noticed in `Phoenix.HTML.Form` you can pass connection data, but it's manipulated via the path, so you don't actually have to handle the connection type directly.

josevalim:
@acconrad that's a very good point. At best, we could keep controller_name here and Phoenix.HTML would only dasherize it.

acconrad:
@josevalim relating to my previous comment, how exactly could `Phoenix.HTML` dasherize that function - I'm not seeing anywhere in the code where that package references anything outside of `Phoenix.HTML`, so I'm not sure how I could call `Phoenix.Controller` in there.

acconrad:
@chrismccord just wanted to check back in here - been a few weeks and wanted to see what else I needed to do before we can merge this in

chrismccord:
> just wanted to check back in here - been a few weeks and wanted to see what else I needed to do before we can merge this in

This is still on my list to think about and review, so hang tight!

chrismccord:
@acconrad Thanks very much for this contribution, but I'm going to wait on this until after 1.0. For now, users can provide their own helper easily by using the `controller_module(conn)` and providing whatever idiom to class names that they follow. Thanks!

PR #1023, opened 2015-07-14
Add Socket handlers and move socket mounts to Endpoint

chrismccord:
The main motivations for these changes and additions is to:

1. Skip the endpoint pipeline for socket connections
2. Introduce the `Phoenix.Socket` behaviour for socket handlers to authenticate connections a single time and set up default assigns for the channel sockets. This not only allows authentication in one place, but it lets us send a `"disconnect"` event to terminate a users active socket connections and multiplexed channels.

Ready for review // @josevalim


lib/phoenix/socket.ex
```diff
@@ -1,10 +1,11 @@
 defmodule Phoenix.Socket do
-  @moduledoc """
+  @moduledoc ~S"""
```
jeregrine: whats the ~S do?

josevalim: Makes you a superhero so you can fight batman in the next movie.

PS: it is the string sigil and ~S removes quoting / interpolation.

lib/mix/tasks/phoenix.gen.channel.ex
```diff
@@ -34,11 +34,9 @@ defmodule Mix.Tasks.Phoenix.Gen.Channel do
 
     Mix.shell.info """
 
-    Add the channel to a socket scope in web/router.ex:
+    Add the channel to a socket handler, ie:
```
josevalim: Please remove the `ie`. It is confusing to most non-native speakers. :)

lib/mix/tasks/phoenix.gen.channel.ex
```diff
@@ -34,11 +34,9 @@ defmodule Mix.Tasks.Phoenix.Gen.Channel do
...
 
-        socket "/ws", #{binding[:base]} do
-          channel "#{plural}:lobby", #{binding[:scoped]}Channel
-        end
+      channel "#{plural}:lobby", #{binding[:scoped]}Channel
```
josevalim: Four spaces to please the markdown gods!

lib/phoenix/controller.ex
```diff
@@ -128,6 +128,12 @@ defmodule Phoenix.Controller do
   def endpoint_module(conn), do: conn.private.phoenix_endpoint
 
   @doc """
+  Returns the socket handler module as an atom, raises if unavailable.
+  """
+  @spec socket_handler_module(Plug.Conn.t) :: atom
+  def socket_handler_module(conn), do: conn.private.phoenix_socket_handler
```
josevalim: This will never be filled when we get here unless for long polling I assume. So let's :fire:.

lib/phoenix/endpoint.ex
```diff
@@ -388,11 +389,25 @@ defmodule Phoenix.Endpoint do
 
   @doc false
   defmacro __before_compile__(env) do
+    sockets = Module.get_attribute(env.module, :phoenix_sockets)
     plugs = Module.get_attribute(env.module, :plugs)
     {conn, body} = Plug.Builder.compile(env, plugs, [])
 
+    socket_intercepts = for {path, module} <- sockets do
```
josevalim: Can you add a TODO here so we move this to the adapter? I will handle it later.

lib/phoenix/transports/long_poller/server.ex
```diff
@@ -103,6 +108,13 @@ defmodule Phoenix.Transports.LongPoller.Server do
   end
 
   @doc """
+  Detects disconnect broadcasts and shuts down
+  """
+  def handle_info(%Broadcast{event: "disconnect"}, state) do
+    {:stop, :normal, state}
```
josevalim: `{:stop, {:shutdown, :disconnected}, state}`

`:normal` is dangerous because links won't break.

lib/phoenix/socket/router.ex
```diff
@@ -0,0 +1,30 @@
...
+  use Plug.Builder
+  alias Phoenix.Transports.WebSocket
+  alias Phoenix.Transports.LongPoller
+
+  plug Plug.Logger
+  plug :fetch_query_params
+  plug :dispatch
+
+  def dispatch(conn, _) do
+    case {conn.method, transport(conn)} do
```
jeregrine: why not just have multiple dispatch function heads?

lib/phoenix/socket/router.ex
```diff
@@ -0,0 +1,30 @@
...
+      {"GET",  WebSocket}     -> WebSocket.call(conn, [])
+      {"POST", WebSocket}     -> WebSocket.call(conn, [])
+      {"OPTIONS", LongPoller} -> LongPoller.call(conn, :options)
+      {"GET", LongPoller}     -> LongPoller.call(conn, :poll)
+      {"POST", LongPoller}    -> LongPoller.call(conn, :publish)
+      _                       -> conn |> send_resp(:bad_request, "") |> halt()
+    end
+  end
+
+  defp transport(conn) do
```
jeregrine: ditto here regarding multiple heads

lib/phoenix/socket/router.ex
```diff
@@ -0,0 +1,30 @@
...
+    case conn.query_params["transport"] do
```
josevalim: Why didn't we need this in the past and we need it now?

josevalim: Ah, I see. In the endpoint, match on the ["ws"|_] prefix. This way we can continue to do route matching for long pooling instead of relying on query parameters.

josevalim:
@chrismccord what if we generate a new Phoenix.Socket with new applications?

lib/phoenix/controller.ex
chrismccord: It will be there for long polling

lib/phoenix/endpoint.ex
chrismccord: what part is moving to the adapter?

lib/phoenix/socket/router.ex
chrismccord: You know I rewrote this about six times with and without multiple dispatch heads. I generally prefer multiple heads, so may change :)

lib/phoenix/socket/router.ex
chrismccord: I had that at first, but stomping the entire socket mount namespace seems wrong. So we could hardcode `["ws", "poll"]`, with an extra clause, but this also felt wrong. So I re-evaluated why we needed an extra path, and I think it makes more since to have a single socket mount point and the client tells us which transport it wants. It is a breaking change for long polling clients (they need new js), but old ws clients would still work, which I am okay with. Better?

chrismccord:
Yes, we'll want to generate a default `UserSocket` for sure. Let's handle that after we merge.

lib/phoenix/endpoint.ex
josevalim: The paths matching needs to be compiled as a cowboy route or something similar if we want to avoid extra processes in http2. but we will cross that bridge when it comes. Just leave a todo. :)

lib/phoenix/controller.ex
josevalim: Yeah, moving the function to the long polling module is :+1:.

lib/phoenix/socket/router.ex
chrismccord: Also, consider if we later add an http2 streaming transport, the client just sends up the new query string param vs yet-another path postfix

josevalim: Ok, you convinced me. :)

lib/phoenix/controller.ex
chrismccord: Oh sorry, I mean both Transport.WebSocket and Transport.LongPoller use it. We can remove and reference the priv in both places though?

josevalim: Why don't you pass it as argument to the transport then? :) They don't need to mimic plug APIs.

lib/phoenix/transports/long_poller/server.ex
chrismccord: :+1: Fixed. In the cowboy case, we can only return `{:shutdown, req, state}`. I assume the links are handled properly in these cases by cowboy's shutdown reason?

chrismccord:
@josevalim all feedback addressed :heart: :purple_heart: :yellow_heart:

lib/phoenix/socket.ex
```diff
@@ -15,8 +16,56 @@ defmodule Phoenix.Socket do
...
+
+  ## Socket Behaviour
+
+  Socket handlers are mounted in Endpoints and must define two callbacks:
+
+    * `connect/2` - receives the socket params and authenticates the connection.
+                    Often used to wire up default `%Phoenix.Socket{}` assigns
+                    for all channels.
+    * `id/1` - receives the socket returned by `connect/2`, and returns the
+               string id of this connection. Used for forcing a disconnect for
```
josevalim: It should be:

      * `id/1` - receives the socket returned by `connect/2`, and returns the
        string id of this connection. Used for forcing a disconnect for

Otherwise it shows all messed up in markdown.

lib/phoenix/socket/router.ex
```diff
@@ -0,0 +1,37 @@
...
+
+  plug Plug.Logger
+  plug :fetch_query_params
+  plug :transport_dispatch
+
+  def transport_dispatch(conn, _) do
+    dispatch(conn, conn.method, conn.private.phoenix_socket_transport)
+  end
+
+  defp dispatch(conn, method, @websocket) when method in ["GET", "POST"] do
```
josevalim: Beautiful.

josevalim:
@chrismccord I have added one more tiny comment.

There are two last things that can be addressed later.

1. How the transport configuration will look like? Are we keeping it in the endpoint or moving it to the socket?

2. Reminder: generate a socket by default in new apps

chrismccord:
> How the transport configuration will look like? Are we keeping it in the endpoint or moving it to the socket?

I think it is fine for the config to remain on the endpoint. I also kept the previous transport options, untouched, on the `channel` macro. This leads to more noise, but I am okay with it, for example:


    channel "rooms:*", App.RoomChannel, via: [WebSocket]
    channel "posts:*", App.PostChannel, via: [WebSocket]
    channel "inbox:*", App.InboxChannel, via: [WebSocket]

chrismccord:
@josevalim Transport configuration has been moved to the socket handlers. I like this change a lot so nice gut instinct. Let me know what you think about the how I exposed the config via `__transport__/1` and if we're ready to ship. @ericmj I had to slightly bump the origin checking, so please review if you have a chance. Thanks!

lib/phoenix/socket.ex
```diff
@@ -47,6 +117,135 @@ defmodule Phoenix.Socket do
...
+      @behaviour Phoenix.Socket
+      import unquote(__MODULE__)
+      Module.register_attribute(__MODULE__, :phoenix_channels, accumulate: true)
+      @phoenix_transports %{}
+
+      transport :websocket, Phoenix.Transports.WebSocket,
+        serializer: Phoenix.Transports.JSONSerializer,
+        timeout: :infinity
+
+      transport :longpoll, Phoenix.Transports.LongPoll,
```
josevalim: Let's rather ask users to have this explicit in the code. It is more inline with Phoenix philosophy. We could provide reasonable defaults for both them, so I just need to have: `transport :websocket, Phoenix.Transports.WebSocket`.

josevalim: To be clear, both transports will be explicit. We can generate them by default with the new user socket in new apps, long pooling will be commented.

lib/phoenix/socket.ex
```diff
@@ -47,6 +117,135 @@ defmodule Phoenix.Socket do
...
+
+  """
+  defmacro transport(name, module, config \\ []) do
+    quote bind_quoted: [name: name, module: module, config: config] do
+      merged_conf = case @phoenix_transports[name] do
+        nil -> config
+        {^module, existing_conf} -> Keyword.merge(existing_conf, config)
+        {dup_module, _} ->
+          raise ArgumentError, """
+          Duplicate transports (`#{inspect dup_module}`, `#{inspect module}`) defined for `:#{name}`".
```
josevalim: error messages usually start in lowercase.

lib/phoenix/socket.ex
```diff
@@ -47,6 +117,135 @@ defmodule Phoenix.Socket do
...
+        timeout: 10_000
+
+      # define separate transport, using websocket handler
+      transport :websocket_slow_clients, Phoenix.Transports.WebSocket,
+        timeout: 60_000
+
+  """
+  defmacro transport(name, module, config \\ []) do
+    quote bind_quoted: [name: name, module: module, config: config] do
+      merged_conf = case @phoenix_transports[name] do
```
josevalim: We should move the whole quote to a callback function that we would call passing the module name. Remember, we should avoid code in quotes.

lib/phoenix/socket.ex
```diff
@@ -47,6 +117,135 @@ defmodule Phoenix.Socket do
...
+  a plain string is provied, only that topic will match the channel handler.
+  Most use-cases will use the "topic:*" pattern to allow more versatile topic
+  scoping.
+
+  See `Phoenix.Channel` for more information
+  """
+  defmacro channel(topic_pattern, module, opts \\ []) do
+    quote do
+      @phoenix_channels {unquote_splicing([
+        topic_pattern,
```
josevalim: Why do you need splicing? Is this just a shortcut to avoid unquoting one by one?

josevalim:
I have added some comments.

lib/phoenix/socket.ex
chrismccord: :+1: If you ship with the long polling code anyway, I'm inclined to support it out of the box

lib/phoenix/socket.ex
chrismccord: Yes. Laziness :)

lib/phoenix/socket.ex
chrismccord: I had everything outside the quote with __CALLER__.module, but was getting messy. Will take another pass with a function

chrismccord: Actually, if we include the :websocket/:longpoller explicitly, then we don't need merging and I can ditch most of the code (other than raising on dups)

lib/phoenix/socket.ex
josevalim: I just think it should be commented to push people to not rely on it if they don't need to. Then it is easier for us to move it elsewhere in the future.

PR #1026, opened 2015-07-16
Prevent Push leaking receive hooks

maxnordlund:
This makes `receive` return early if it already has been received in order to allow `matchReceive` to safely clear the receive hooks array.

I use `null` instead of `delete`<sup>[1][] [2][]</sup> to not [change the shape][3] of the Push object, since that is a known [optimization killer][4]. The last link doesn't talk about `delete` specifically, but contains useful general optimization advice.

[1]: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/delete#section_5
[2]: http://perfectionkills.com/understanding-delete/
[3]: http://www.html5rocks.com/en/tutorials/speed/v8/
[4]: https://github.com/petkaantonov/bluebird/wiki/Optimization-killers

### DIFF
--- a/web/static/js/phoenix.js
+++ b/web/static/js/phoenix.js
@@ -146,6 +146,7 @@ class Push {
   receive(status, callback){
     if(this.receivedResp && this.receivedResp.status === status){
       callback(this.receivedResp.response)
+      return this
     }
 
     this.recHooks.push({status, callback})
@@ -166,6 +167,7 @@ class Push {
   matchReceive({status, response, ref}){
     this.recHooks.filter( h => h.status === status )
                  .forEach( h => h.callback(response) )
+    this.recHooks = null
   }
 
   cancelRefEvent(){ this.chan.off(this.refEvent) }


chrismccord:
Thanks! Can you better describe the bug that this fixes? I don't quite follow the from the code or description? Are you seeing buggy behavior or?

maxnordlund:
Yes, sorry for that. Communication is hard :smiley_cat: Hopefully this will make it clearer.

The memory leak comes from `receive` unconditionally pushed the callback onto `recHooks`. This can get pretty nasty if the callback is a closure, since that could be much bigger then just a function (reference) getting leaked.

The `return this` fixes that, but the `recHooks` array also needs to be cleared. It seams not to be needed after `matchReceive` as far as I can see. Using `delete` would put it in slow mode, aka a hash map, instead of something more like a C struct. So instead just set it to `null`, like the timer is a bit below.

----

The Travis failure seams a bit odd, and I'm not sure if it is relevant to this but you never know. Could you restart it to see if it was just sad or I accidentally broke something?

chrismccord:
I still don't see that there is a memory leak with the current code. This PR may be an optimization, but the entire Push object should be garbage collected at any rate, right? This also breaks reconnects. Consider this code:

```javascript
chan.join()
  .receive("ok", func1)
  .receive("error", func2)
```

We re-invoke this chain whenever the connection drops and reconnects. So we can't discard the receive hooks, because the response chain may be different on subsequent reconnects. We wouldn't need to enforce this for non "join" events, so other pushes could introduce this optimization, but if there is no real memory leak and without hard memory usage examples, I don't think we need this change. Can you verify?

PS. That you very much for working to make Phoenix better :)

chrismccord:
@maxnordlund ping?

chrismccord:
Closing for now. Please follow up if you think there is still a leak here

PR #1028, opened 2015-07-17
Generate UserSocket module in new projects

chrismccord:
@josevalim Can I get a signoff on the generate module and docs? Should we add comments with more info in UserSocket or is this good?


lib/mix/tasks/phoenix.gen.channel.ex
```diff
@@ -34,7 +34,7 @@ defmodule Mix.Tasks.Phoenix.Gen.Channel do
 
     Mix.shell.info """
 
-    Add the channel to a socket handler, for example:
+    Add the channel to your `web/channels/user_socke.ex` handler, for example:
```
Gazler: Shouldn't this be user_socket.ex?

PR #1037, opened 2015-07-21
Remove bottlenecks from Phoenix PubSub Local

josevalim:
The clients are the ones sending the messages now, except for distributed messages, which are spread by the PG2 Server.

### DIFF
--- a/lib/phoenix/pubsub/local.ex
+++ b/lib/phoenix/pubsub/local.ex
@@ -16,7 +16,7 @@ defmodule Phoenix.PubSub.Local do
 
   """
   def start_link(server_name) do
-    GenServer.start_link(__MODULE__, [], name: server_name)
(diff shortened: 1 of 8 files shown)


lib/phoenix/pubsub/local.ex
```diff
@@ -82,30 +88,35 @@ defmodule Phoenix.PubSub.Local do
...
+    catch
+      :error, :badarg -> []
+    end
   end
 
   @doc false
-  def list(local_server) do
-    GenServer.call(local_server, :list)
+  # This is an expensive and private operation. DO NOT USE IT IN PROD.
+  def list(local_server) when is_atom(local_server) do
```
jeregrine: should you make this defp then?

josevalim: We need it for tests. It is already `@doc false`, it is just a warning for those two decide to read the source and not respect the fact it is private.

PR #1046, opened 2015-07-22
Do as much serialization as possible in the channels. Closes #1042

### LINKED ISSUE #1042: Do as much serialization as possible in the channel process instead of transport

chrismccord:
@josevalim first pass and could use feedback. Do we want to merge this before starting the serialization fastlaning or do them both at once?

### DIFF
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -205,7 +205,6 @@ defmodule Phoenix.Channel do
       @before_compile unquote(__MODULE__)
       import unquote(__MODULE__)
       import Phoenix.Socket, only: [assign: 3]
-      Module.register_attribute(__MODULE__, :phoenix_handle_outs, accumulate: true)
 
       @phoenix_intercepts []
 
@@ -273,7 +272,6 @@ defmodule Phoenix.Channel do
                "An intercept for event \"#{event}\" has not yet been defined in #{env.module}.handle_out/3. " <>
                "Add \"#{event}\" to your list of intercepted events with intercept/1"
     end
-    Module.put_attribute(env.module, :phoenix_handle_outs, event)
   end
   def __on_definition__(_env, _kind, _name, _args, _guards, _body) do
   end
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -192,6 +192,10 @@ defmodule Phoenix.Channel do
               {:stop, reason :: term, Socket.t} |
               {:stop, reason :: term, reply, Socket.t}
 
+  defcallback handle_out(event :: String.t, msg :: map, Socket.t) ::
+              {:noreply, Socket.t} |
+              {:stop, reason :: term, Socket.t}
+
   defcallback handle_info(term, Socket.t) ::
               {:noreply, Socket.t} |
               {:stop, reason :: term, Socket.t}
@@ -209,7 +213,14 @@ defmodule Phoenix.Channel do
       import Phoenix.Socket, only: [assign: 3]
       Module.register_attribute(__MODULE__, :phoenix_handle_outs, accumulate: true)
 
+      @phoenix_intercepts []
+
       def handle_in(_event, _message, socket) do
+        {:noreply, socket}
+      end
+
+      def handle_out(event, message, socket) do
+        push(socket, event, message)
         {:noreply, socket}
       end
 
@@ -217,16 +228,26 @@ defmodule Phoenix.Channel do
 
       def terminate(_reason, _socket), do: :ok
 
-      defoverridable handle_info: 2, handle_in: 3, terminate: 2
+      defoverridable handle_info: 2, handle_out: 3, handle_in: 3, terminate: 2
     end
   end
 
-  defmacro __before_compile__(_) do
+  defmacro __before_compile__(env) do
+    intercepts = Module.get_attribute(env.module, :phoenix_intercepts)
+    handle_outs = Module.get_attribute(env.module, :phoenix_handle_outs)
+    for event <- handle_outs -- intercepts do
+      IO.write "[warning] #{__MODULE__}.handle_out/3 defined for \"#{event}\", but event not intercepted. " <>
+               "Add \"#{event}\" to your list of intercepted events with intercept/1"
+    end
+
     quote do
-      if @phoenix_handle_outs != [] do
-        def __fastlane__?(event) when event in @phoenix_handle_outs, do: false
-      end
-      def __fastlane__?(event), do: true
+      def __phoenix_intercepts__(), do: unquote(intercepts)
+    end
+  end
+
+  defmacro intercept(events) when is_list(events) do
+    quote do
+      @phoenix_intercepts unquote(events)
     end
   end
 
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -62,8 +62,8 @@ defmodule Phoenix.Channel do
   You can also push a message directly down the socket:
 
       # client asks for their current rank, push sent directly as a new event.
-      def handle_in("current:rank", socket) do
-        push socket, "current:rank", %{val: Game.get_rank(socket.assigns[:user])}
+      def handle_in("current_rank", socket) do
+        push socket, "current_rank", %{val: Game.get_rank(socket.assigns[:user])}
         {:noreply, socket}
       end
 
@@ -100,12 +100,16 @@ defmodule Phoenix.Channel do
         end
       end
 
-  ### Outgoing Events
+  ### Intercepting Outgoing Events
 
   When an event is broadcasted with `Phoenix.Channel.broadcast/3`, each channel
-  subscriber's `handle_out/3` callback is triggered where the event can be
-  relayed as is, or customized on a socket by socket basis to append extra
-  information, or conditionally filter the message from being delivered.
+  subscriber can choose to intercept the event and have their `handle_out/3`
+  callback triggered. This allows the event's payload can be to customized on a
+  socket by socket basis to append extra information, or conditionally filter
+  the message from being delivered. If the event is not intercepted with
+  `Phoenix.Channel.intercept/1`, then the message is pushed to directly to the client:
+
+      intercept ["new_msg", "user_joined"]
 
       def handle_in("new_msg", %{"uid" => uid, "body" => body}, socket) do
         broadcast! socket, "new_msg", %{uid: uid, body: body}
@@ -121,18 +125,15 @@ defmodule Phoenix.Channel do
         {:noreply, socket}
       end
 
-      # do not send broadcasted `"user:joined"` events if this socket's user
+      # do not send broadcasted `"user_joined"` events if this socket's user
       # is ignoring the user who joined.
-      def handle_out("user:joined", msg, socket) do
+      def handle_out("user_joined", msg, socket) do
         unless User.ignoring?(socket.assigns[:user], msg.user_id) do
-          push socket, "user:joined", msg
+          push socket, "user_joined", msg
         end
         {:noreply, socket}
       end
 
-  By default, unhandled outgoing events are forwarded to each client as a push,
-  but you'll need to define the catch-all clause yourself once you define an
-  `handle_out/3` clause.
 
   ## Broadcasting to an external topic
 
@@ -157,6 +158,7 @@ defmodule Phoenix.Channel do
         MyApp.Endpoint.broadcast! "rooms:superadmin", "new_msg", %{uid: uid, body: body}
         redirect conn, to: "/"
       end
+
 
   ## Terminate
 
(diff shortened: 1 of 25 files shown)


chrismccord:
@josevalim I just added fastlaning to this branch. I'm sure we'll need to Skype tomorrow about it. I don't like how tightly coupled it makes pubsub and channels, but it may be unavoidable. On the plus side, the fastlaning doubles throughput in my benchmarks! I also used `on_definition` to track the handle_outs, but this has known caveats. Mainly I cheated here to get something together quickly. We can discuss tomo. Let me know how it looks!

chrismccord:
Goes without saying, but still needs docs and tests

lib/phoenix/channel/server.ex
```diff
@@ -149,14 +149,12 @@ defmodule Phoenix.Channel.Server do
   @doc """
   Pushes a reply message to the given pid.
   """
-  def reply(pid, ref, topic, payload)
+  def reply(pid, ref, topic, payload, serializer)
```
josevalim: I thought we would kill this function though because it means the transport is sending itself messages?

josevalim:
So the fastlane does not have cache keys, you are probably aware of that, but right now different transports are going to use the same cache. That's why I was thinking the fastlane implementation would not invoke a function on the channel because, if we do, we need to use the channel as key, which means that we can only share the cache if they are meant for the same channel. On the other hand, this may be fine, it is unlikely different channels will be consuming the same topic.

In any case, we have two options:

1. Something along the lines of the current patch

2. Give the serializer (with the fastline callback), the transport pid, and the topics to NOT fastlane to the subscriber. This way, we never call the transport, we have all we need to make the decision on the spot

lib/phoenix/channel/server.ex
```diff
@@ -233,6 +235,29 @@ defmodule Phoenix.Channel.Server do
     socket.channel.terminate(reason, socket)
   end
 
+  def fastlane(topic, %Broadcast{event: event} = msg, cache, transport_pid, channel, serializer) do
+    if channel.__fastlane__?(event) do
+      cache_key = {channel, event, serializer}
```
josevalim: we shouldn't need the event in the cache key because the cache key does not outlast an event.

lib/phoenix/channel.ex
```diff
@@ -207,26 +203,40 @@ defmodule Phoenix.Channel do
...
+      if @phoenix_handle_outs != [] do
+        def __fastlane__?(event) when event in @phoenix_handle_outs, do: false
+      end
+      def __fastlane__?(event), do: true
     end
   end
 
+  def __on_definition__(env, :def, :handle_out, [event, _payload, _socket], _, _)
+    when is_binary(event) do
+    Module.put_attribute(env.module, :phoenix_handle_outs, event)
```
josevalim: The problem with this solution is that someone can write:

    def handle_out(event, payload, socket) do
       # Add something to all messages
    end

And we are still going to fastlane. So the proper implementation would be, as soon as we find a handle_out clause that does not a binary event name, we need to assume we can no longer fastlane anything, at all.

josevalim:
Twice the throughput is a really nice improvement. Which benchmark were you running? The same as usual?

lib/phoenix/channel/server.ex
chrismccord: yeah still on my todo list

lib/phoenix/channel/server.ex
chrismccord: ah, good cache

chrismccord:
> Twice the throughput is a really nice improvement. Which benchmark were you running? The same as usual?

I added a websocket client simulator to the PubsubStress app. I still used the controller broadcasts, so we might even see better gains if we measured solely channel messaging. I need to update @seejee's phoenix channel benchmark app and see what we get there.

seejee:
@chrismccord Sorry that it's gotten a little out of date, but I'll update it today.

Twice the throughput would be :+1: :smile_cat:

chrismccord:
> Sorry that it's gotten a little out of date, but I'll update it today.

Talk about service! We need to get it on master which has breaking changes. Find me on irc if you get to it and want help getting things rolling. Thanks!

seejee:
Sounds good. I'll get it updated to 0.14 (I love following the upgrade guides!), make sure everything works, and then find you in IRC.

lib/phoenix/channel/server.ex
lancehalvorsen: (Best typo ever?)

chrismccord:
@josevalim ready for final review when you have a moment

lib/phoenix/channel.ex
```diff
@@ -100,12 +100,16 @@ defmodule Phoenix.Channel do
...
-  information, or conditionally filter the message from being delivered.
+  subscriber can choose to intercept the event and have their `handle_out/3`
+  callback triggered. This allows the event's payload can be to customized on a
+  socket by socket basis to append extra information, or conditionally filter
+  the message from being delivered. If the event is not intercepted with
+  `Phoenix.Channel.intercept/1`, then the message is pushed to directly to the client:
+
+      intercept ["new_msg", "user_joined"]
 
       def handle_in("new_msg", %{"uid" => uid, "body" => body}, socket) do
```
josevalim: Let's take the `handle_in` out of here to avoid confusion?

lib/phoenix/channel.ex
```diff
@@ -204,29 +202,85 @@ defmodule Phoenix.Channel do
...
+
+      def terminate(_reason, _socket), do: :ok
+
+      defoverridable handle_info: 2, handle_in: 3, terminate: 2
+    end
+  end
+
+  defmacro __before_compile__(env) do
+    intercepts = Module.get_attribute(env.module, :phoenix_intercepts)
+    handle_outs = Module.get_attribute(env.module, :phoenix_handle_outs)
```
josevalim: We don't need to store the handle outs, do we? We can emit the warning directly from `on_definition`, which would even allow us to use the file and line info.

lib/phoenix/channel.ex
```diff
@@ -204,29 +202,85 @@ defmodule Phoenix.Channel do
...
 
-      defoverridable handle_info: 2, handle_out: 3, handle_in: 3, terminate: 2
+  """
+  defmacro intercept(events) when is_list(events) do
+    quote do
+      @phoenix_intercepts unquote(events)
     end
   end
 
+  def __on_definition__(env, :def, :handle_out, [event, _payload, _socket], _, _)
```
josevalim: `@doc false`.

lib/phoenix/channel/server.ex
```diff
@@ -193,7 +178,12 @@ defmodule Phoenix.Channel.Server do
   end
 
   defp join(socket, reply, parent, ref) do
-    PubSub.subscribe(socket.pubsub_server, self(), socket.topic, link: true)
+    PubSub.subscribe(socket.pubsub_server, self(), socket.topic,
+      link: true,
+      fastlane: {socket.transport_pid,
+                 socket.serializer,
+                 socket.channel.__phoenix_intercepts__()})
```
josevalim: Nitpick: Let's drop the `phoenix` from `__intercepts__` because we typically do prefix our metadata in Phoenix with `phoenix`. The only exception is the template reloading but only because it can be added to all modules.

lib/phoenix/pubsub.ex
```diff
@@ -108,10 +108,16 @@ defmodule Phoenix.PubSub do
     * `server` - The Pid registered name of the server
     * `pid` - The subscriber pid to receive pubsub messages
     * `topic` - The topic to subscribe to, ie: `"users:123"`
-    * `opts` - The optional list of options. Supported options
-      only include `:link` to link the subscriber to the pubsub
-      adapter
-
+    * `opts` - The optional list of options. Supported options:
```
josevalim: Create a subsection `## Options`.  Better than nesting. :)

lib/phoenix/pubsub/local.ex
```diff
@@ -69,13 +71,43 @@ defmodule Phoenix.PubSub.Local do
       :ok
 
   """
+  def broadcast(local_server, from, topic, %Broadcast{event: event} = msg)
```
josevalim: Beautiful.

lib/phoenix/pubsub/local.ex
```diff
@@ -69,13 +71,43 @@ defmodule Phoenix.PubSub.Local do
...
+        if event in event_intercepts do
+          send(pid, msg)
+          cache
+        else
+          case Map.fetch(cache, serializer) do
+            {:ok, encoded_msg} ->
+              send(fastlane_pid, encoded_msg)
+              cache
+            :error ->
+              encoded_msg = serializer.encode!(msg)
```
josevalim: I was planning to call the callback `fastlane!`. The reason is that encoding is a special implenentation detail for fastlaning for transports. But it is not a requirement for all pubsub subscribers. Calling something like `fastlane!` wouldn't couple to our channel+transports system (at the costing of slightly augmenting the serializer API surface).

lib/phoenix/test/channel_test.ex
```diff
@@ -127,8 +127,33 @@ defmodule Phoenix.ChannelTest do
 
   alias Phoenix.Socket
   alias Phoenix.Socket.Message
+  alias Phoenix.Socket.Broadcast
+  alias Phoenix.Socket.Reply
   alias Phoenix.Channel.Server
 
+  defmodule NoopSerializer do
+    def encode!(%Reply{} = reply) do
```
josevalim: Any reason why you are not simply returning `%Reply{}`?

lib/phoenix/channel.ex
chrismccord: This is only if the user has previously called `intercepts` though, which if you are using good style, should be first, but __before_compile__ ensures we're set to do the check. Would you still prefer directly warning?

lib/phoenix/test/channel_test.ex
```diff
@@ -127,8 +127,33 @@ defmodule Phoenix.ChannelTest do
...
+        topic: msg.topic,
+        event: msg.event,
+        payload: msg.payload
+      }
+    end
+    def encode!(%Message{} = msg) do
+      msg
+    end
+
+    def decode!(message, :text), do: message
```
josevalim: Should we pass still `:text` into `decode!`. It seems something the transport and serializer itself should figure out? Can we be possibly leaking websockets concerns into all serializers?

lib/phoenix/transports/json_serializer.ex
```diff
@@ -4,17 +4,38 @@ defmodule Phoenix.Transports.JSONSerializer do
...
   @behaviour Phoenix.Transports.Serializer
 
+  alias Phoenix.Socket.Reply
+  alias Phoenix.Socket.Message
+  alias Phoenix.Socket.Broadcast
+
   @doc """
   Encodes a `Phoenix.Socket.Message` struct to JSON string.
   """
-  def encode!(message), do: {:text, Poison.encode_to_iodata!(message)}
```
josevalim: I would rename this module to websocket_serializer. It just happens to use JSON (which we would document.

lib/phoenix/pubsub/local.ex
josevalim: Another reason for `fastlane!` is that `encode!` shouldn't really receive Broadcast messages. So moving it to another function would make it explicit.

test/shared/pubsub_test.exs
```diff
@@ -40,7 +40,7 @@ defmodule Phoenix.PubSubTest do
     pid = spawn_pid
     assert Local.subscribers(config.local, "topic4") |> Dict.size == 0
     assert PubSub.subscribe(config.test, pid, "topic4")
-    assert Local.subscribers(config.local, "topic4") |> Enum.to_list == [pid]
+    assert Local.subscribers(config.local, "topic4") |> Enum.to_list == [{pid, nil}]
```
josevalim: Should we change subscribers to return the fastlanes? Maybe it should just return PIDs?

lib/phoenix/channel.ex
josevalim: Just word it in a way it makes sense: "an intercept has not yet been defined".

chrismccord:
@josevalim All feedback addressed. :heart: :purple_heart: :yellow_heart:  :shipit: ?

lib/phoenix/channel.ex
```diff
@@ -204,29 +197,87 @@ defmodule Phoenix.Channel do
...
 
+
   defmacro __using__(_) do
     quote do
       @behaviour unquote(__MODULE__)
+      @on_definition unquote(__MODULE__)
+      @before_compile unquote(__MODULE__)
       import unquote(__MODULE__)
       import Phoenix.Socket, only: [assign: 3]
+      Module.register_attribute(__MODULE__, :phoenix_handle_outs, accumulate: true)
```
josevalim: :fire:

lib/phoenix/channel.ex
```diff
@@ -204,29 +197,87 @@ defmodule Phoenix.Channel do
...
+  @doc false
+  def __on_definition__(env, :def, :handle_out, [event, _payload, _socket], _, _)
+    when is_binary(event) do
+
+    unless event in Module.get_attribute(env.module, :phoenix_intercepts) do
+      IO.write "#{Path.relative_to(env.file, File.cwd!)}:#{env.line}: [warning] " <>
+               "An intercept for event \"#{event}\" has not yet been defined in #{env.module}.handle_out/3. " <>
+               "Add \"#{event}\" to your list of intercepted events with intercept/1"
+    end
+    Module.put_attribute(env.module, :phoenix_handle_outs, event)
```
josevalim: :fire:

josevalim:
:shipit:

PR #1060, opened 2015-07-28
Fix latin abbreviations in Phoenix.js comments

shreve:
Previously, the comments read "ie", which is short for "id est", meaning, "that is". This implies that what follows is what you're really trying to say. For example, on line 211, ie meant that the event was always going to be `"phx_join"`.
Now, the comments read "e.g.", which is short for "exempli gratia", meaning "for example". That same example now means that event is going to be a string from the list of events, like `"phx_join"`

### DIFF
--- a/lib/phoenix/channel/transport.ex
+++ b/lib/phoenix/channel/transport.ex
@@ -119,7 +119,7 @@ defmodule Phoenix.Channel.Transport do
 
     * `topic` - The String value "phoenix"
     * `event` - The String value "heartbeat"
-    * `payload` - An empty JSON message payload, ie {}
+    * `payload` - An empty JSON message payload, that is {}
 
   The server will respond to heartbeats with the same message
   """
--- a/lib/phoenix/channel/transport.ex
+++ b/lib/phoenix/channel/transport.ex
@@ -119,7 +119,7 @@ defmodule Phoenix.Channel.Transport do
 
     * `topic` - The String value "phoenix"
     * `event` - The String value "heartbeat"
-    * `payload` - An empty JSON message payload, that is {}
+    * `payload` - An empty JSON message payload (`{}`)
 
   The server will respond to heartbeats with the same message
   """
--- a/priv/static/phoenix.js
+++ b/priv/static/phoenix.js
@@ -208,8 +208,8 @@ var Push = (function () {
   // Initializes the Push
   //
   // chan - The Channel
-  // event - The event, ie `"phx_join"`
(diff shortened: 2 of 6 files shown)


priv/static/phoenix.js
```diff
@@ -1203,4 +1203,4 @@ Object.defineProperty(exports, "__esModule", {
   value: true
 });
  }});
-if(typeof(window) === 'object' && !window.Phoenix){ window.Phoenix = require('phoenix') };
\ No newline at end of file
+if(typeof(window) === 'object' && !window.Phoenix){ window.Phoenix = require('phoenix') };
```
shreve: I'm not sure why GitHub added the line ending. This was not my intention.

ericmj:
This file is generated, please do the changes in https://github.com/phoenixframework/phoenix/blob/master/web/static/js/phoenix.js instead.

josevalim:
Thank you! We should rather remove all `i.e.` and `e.g.` from comments and documentation as it is not as clear as words like "for example", "in other words", etc.

lib/phoenix/channel/transport.ex
```diff
@@ -119,7 +119,7 @@ defmodule Phoenix.Channel.Transport do
 
     * `topic` - The String value "phoenix"
     * `event` - The String value "heartbeat"
-    * `payload` - An empty JSON message payload, ie {}
+    * `payload` - An empty JSON message payload, that is {}
```
shreve: This one still feels awkward. Any suggestions to improve the wording here?

josevalim: This one is fine. We just need to put ticks around `{}`. Another approach is:

     An empty JSON message payload (`{}`)

shreve: I like it. :+1: :

priv/static/phoenix.js
josevalim: The line ending is fine. :)

priv/static/phoenix.js
```diff
@@ -208,8 +208,8 @@ var Push = (function () {
   // Initializes the Push
   //
   // chan - The Channel
-  // event - The event, ie `"phx_join"`
-  // payload - The payload, ie `{user_id: 123}`
+  // event - The event, e.g. `"phx_join"`
+  // payload - The payload, e.g. `{user_id: 123}`
```
josevalim: I think only those `e.g.` remain. :)

chrismccord:
<3 <3 <3

PR #1074, opened 2015-07-31
Only gzip files that are known to be plain text.

keichan34:
It doesn't help to gzip bitmap files like PNG or JPEG, and can sometimes be bigger than the uncompressed file.

I'm using a "gzip whitelist" by file extension here, but if there's a better way to do it, I'm all ears.

### DIFF
--- a/lib/phoenix/digester.ex
+++ b/lib/phoenix/digester.ex
@@ -1,5 +1,6 @@
 defmodule Phoenix.Digester do
   @digested_file_regex ~r/(-[a-fA-F\d]{32})/
+  @gzippable_files ~w(.js .css .txt .text .html .json)
 
   @moduledoc """
   Digests and compress static files.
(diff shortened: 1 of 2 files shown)


lib/phoenix/digester.ex
```diff
@@ -90,8 +96,11 @@ defmodule Phoenix.Digester do
     File.mkdir_p!(path)
 
     # compressed files
-    File.write!(Path.join(path, file.digested_filename <> ".gz"), file.compressed_content)
-    File.write!(Path.join(path, file.filename <> ".gz"), file.compressed_content)
+    unless is_nil(file.compressed_content) do
```
josevalim: `if file.compressed_content do`

josevalim:
The patch is perfect, I have added just a tiny comment.

keichan34:
Fixed. Thanks for the quick review!

josevalim:

PR #1076, opened 2015-07-31
Return 201 status code when creating a JSON resource

fbjork:
Changes status code from 200 to 201 when creating a JSON resource.


josevalim:

denispeplin:
@josevalim, should 201 also be applied to HTML resources?

PR #1080, opened 2015-08-02
Fix this.log in Channel.leave where it isn't defined.

jacknoble:
So `this.log` isn't defined here. It looks like it should be `this.socket.log` as it is in the onClose and onError callbacks.


chrismccord:
:heart: :purple_heart: :yellow_heart:

PR #1083, opened 2015-08-03
Move each transport to its own handler

josevalim:
This is not complete. We still need to:

1. Add logging to each handler
2. Add force_ssl to each handler
3. Document the transport behaviour


chrismccord:
> Add logging to each handler

This is in place with the `:log` option, no?

josevalim:
I have removed the Phoenix.Socket.Router which logged. I need to add it back.

lib/phoenix/endpoint/cowboy_websocket.ex
```diff
@@ -1,21 +1,46 @@
...
+  end
+
+  def resume(module, fun, args) do
     try do
       apply(module, fun, args)
     catch
       kind, [{:reason, reason}, {:mfa, _mfa}, {:stacktrace, stack} | _rest] ->
         reason = format_reason(kind, reason, stack)
-        exit({reason, {__MODULE__, :call, [conn, []]}})
+        exit({reason, {__MODULE__, :resume, [module, fun, args]}})
```
fishcakez: By using `{__MODULE__, :resume, [module, fun, args]}` we keep the `args` around, which includes the initial `req` and other things. Those arguments probably aren't that helpful to log in the error and we don't need to hold on to them. Instead we could recreate the real mfa, using `:mfa`, `:message`, `:req` and `:state` from the keyword list.

(Un/)Fortunately `:req` is a keyword list and not the actual `cowboy_req` record. However we could turn it back into the record or use the keyword to create a `Cowboy.Request` struct that could implement the inspect protocol.

Thoughts?

lib/phoenix/endpoint/cowboy_handler.ex
```diff
@@ -1,34 +1,85 @@
...
-  It is also important to specify your handlers first, otherwise Phoenix will
-  intercept the requests before they get to your handler
+  It is also important to specify your handlers first, otherwise
+  Phoenix will intercept the requests before they get to your handler.
+  """
+
+  require Logger
+
+  @doc """
+  Generates sa childspec to be used in the supervision tree.
```
fishcakez: "Generates a childspec"

josevalim:
I am merging this because there is enough work and improvements in. I still need to add the session affinity bits.

PR #1086, opened 2015-08-03
Fix path to generated favicon file.

wtcross:
This commit fixes a bug in the installer `phoenix.new` task
that was causing the `favicon.ico` file to not be found
when requested by browsers at `/favicon.ico`.

The `favicon.ico` file was generated in `web/static/assets/images`
so a `phoenix.digest` will create `priv/static/images/favicon.ico`.
The mount set up by `Plug.Static` can never find this file, since
the mount is relative to `priv/static`.

This commit changes the path of the generated `favicon.ico` to be
`web/static/assets/favicon.ico` so that it will be correctly
digested at `priv/static/favicon.ico`, where it can be found by
the mount.


josevalim:
:+1:

chrismccord:
<3 <3 <3

PR #1106, opened 2015-08-08
Add Coverage Reports using Coveralls

optikfluffel:
For this to work you'd simply have to sign in at [Coveralls](https://coveralls.io) using your Github account and then activate this repo.
Oh and I also bumped the Elixir and OTP versions in `.travis.yml`.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -13,5 +13,6 @@ script:
   - cd installer && mix test
 after_script:
   - cd $TRAVIS_BUILD_DIR
+  - MIX_ENV=test mix coveralls.travis
   - mix deps.get --only docs
   - MIX_ENV=docs mix inch.report
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,9 +1,10 @@
 language: elixir
-elixir: 1.0.4
+elixir: 1.0.5
 services:
   - redis-server
 otp_release:
-  - 17.1
+  - 17.5
+  - 18.0
 sudo: false
 before_script:
   - mix deps.get --only test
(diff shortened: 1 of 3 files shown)


chrismccord:
Thanks, but last we checked shields.io was not a good api citizen. For example, it does no caching, and loading the home page makes like dozens of calls to the hex api alone.

optikfluffel:
And if I change the badge url to the one from coveralls itself?

```markdown
[![Coverage Status](https://coveralls.io/repos/phoenixframework/phoenix/badge.svg?branch=coveralls&service=github)](https://coveralls.io/github/phoenixframework/phoenix?branch=coveralls)
````
[Coverage Status](https://coveralls.io/github/optikfluffel/phoenix?branch=coveralls)

Or don't you like coverage reporting?

chrismccord:
It's not that I don't like it, it's just another thing to go wrong, a dependency to depend on, or have to maintain in the future, so I'm trying to limit things like that at the moment. I think code coverage reports are great fwiw.

nickmerwin:
@chrismccord as the founder/maintainer of Coveralls I could help keep this integrated properly; it should be set-it-and-forget-it unless there are major changes to Elixir's coverage library.

You would also get automated coverage change status updates on PR's; I see there aren't many open right now, but in the future it may be handy.

Let me know if I can help, thanks! :+1:

ericmj:
Another issue is that coverage tooling is not that great in Elixir at this point. It misses covering some lines which leads to weird code to make sure the line is covered. Also, code run at compile time is not covered at all so large parts of phoenix would never be covered.

Issue #1105, opened 2015-08-07
Log when the origin check fails


josevalim:
We should log something with debug so users can act on it. Should we do something similar for ssl?

@chrismccord thoughts?

chrismccord:
Yes. I spent a long while last night helping someone debug an origin issue. Logging would have made it immediately apparent.

romul:
I'm sure, that such logging would be very helpful. I spent several hours trying to figure out what's going on and why I got 403. And only thanks to Chris found a reason.

josevalim:
@romul thanks for the input. I have pushed a commit that adds a very complete logging as well as fixes the proper tests to avoid logging during the test run.

romul:
@josevalim  ok, thank you :-)

asaaki:
Btw would it make sense to extend [this comment in config/prod.exs](https://github.com/phoenixframework/phoenix/blob/f693d1701fb7dc6a89f1a393aa87a9c09e81ffcb/installer/templates/new/config/prod.exs#L7-L8) regarding the origin issue?
I've noticed that a lot of people just stumble upon this. And to come up with the idea to lower the logger level in prod releases might not always the be first thing they consider.
At least we could try to avoid going this step by giving this useful information already in the config file.
What do you think?

josevalim:
Yes, I will work on that. Btw, the log level has been increased to error in the origin case. :)

asaaki:
Ah, that sounds great, too!

PR #1107, opened 2015-08-08
Log when the origin check fails

romul:
Fix for #1105


chrismccord:
This has been taken care of, but thank you very much for the PR

PR #1109, opened 2015-08-09
Update the app.css-template to the latest version of Bootstrap

Theemuts:
I noticed that the app.css generated for new projects used an outdated version of Bootstrap. As far as I've noticed, using the new version does not change the main page of newly generated projects.


chrismccord:
I'd prefer to wait until a major bootstrap release or a decision around how we're going to handle bootstrap. We've discussed different options about including a separate file, not using bootstrap, etc, so I'm going to wait on this one for now. Thanks!

PR #1113, opened 2015-08-10
Makes routing helpers use request data to determine URLs when invoked with a Plug.Conn

stevegraham:
This is a PR to discuss Phoenix routing helpers using request data to determine URLs instead of the `Endpoint`. A use case for this is multi-tenant apps.

One test is broken, the ScriptName one, because I'm not sure which should take precedence, i.e. the script or the connection?

This is my first PR to an OSS Elixir project so please feel free to comment on anything from the general idea to the implementation. I would find it most helpful.

### DIFF
--- a/lib/phoenix/router/helpers.ex
+++ b/lib/phoenix/router/helpers.ex
@@ -72,8 +72,17 @@ defmodule Phoenix.Router.Helpers do
       @doc """
       Generates the connection/endpoint base URL without any path information.
       """
-      def url(%Conn{private: private}) do
-        private.phoenix_endpoint.url
+
+      def url(%Conn{host: host, port: 80, scheme: :http}) do
+        "http://" <> host
+      end
+
+      def url(%Conn{host: host, port: 443, scheme: :https}) do
+        "https://" <> host
+      end
+
+      def url(%Conn{host: host, port: port, scheme: scheme}) do
+        "#{scheme}://" <> host <> ":#{port}"
       end
 
       def url(%Socket{endpoint: endpoint}) do
--- a/test/phoenix/router/helpers_test.exs
+++ b/test/phoenix/router/helpers_test.exs
@@ -257,6 +257,14 @@ defmodule Phoenix.Router.HelpersTest do
     conn(:get, "/") |> put_private(:phoenix_endpoint, __MODULE__)
   end
 
+  defp conn_with_options(opts \\ []) do
+    conn(:get, "/") |> :maps.merge(:maps.from_list(opts))
+  end
+
+  def ssl_conn do
+    conn_with_options(scheme: :https, port: 443)
+  end
+
   defp socket_with_endpoint do
     %Phoenix.Socket{endpoint: __MODULE__}
   end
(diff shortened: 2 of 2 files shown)


josevalim:
The trouble with this is that you are relying the host to always be correct and, if you have a proxy, it certainly won't be unless you set up a bunch of config. I really like the simplicity we only check the url config when generating URLS.

I see exactly why you need this though and I would love to make this possible. Although I don't think this is quite the way to go.

stevegraham:
@josevalim Ok, I was relying on the fields of the connection struct to be correct. What will be incorrect with a proxy setup? The request URI or Host request header? I'm confused because as I understand it proxies are supposed to rewrite the Host header to the target origin server (http://tools.ietf.org/html/draft-ietf-httpbis-p1-messaging-14#section-9.4)

```
   When an HTTP/1.1 proxy receives a request with a request-target in
   the form of an absolute-URI, the proxy MUST ignore the received Host
   header field (if any) and instead replace it with the host
   information of the request-target.  When a proxy forwards a request,
   it MUST generate the Host header field based on the received
   absolute-URI rather than the received Host.
```

josevalim:
@stevegraham sorry, I was thinking of reverse proxies, which sets things like "x-forwarded-host". Maybe you could have your own URL helpers that use the host/subdomain function. I am very wary of doing it by default though. :S

stevegraham:
@josevalim sure. it changes how the routing helpers behave, so I absolutely understand your hesitance. IMO it's how they should behave, i.e. if a helper is invoked with the state of the request, I expect it to be used to generate URLs. Anyway, I am happy to fix this up such that you're happy, otherwise feel free to close this out. Either way, thank you for discussing this with me.

josevalim:
@stevegraham I would rather try to provide a pluggable system for this. So you can generate the URL from the connection when you want. Otherwise we end-up falling in the same trap as Rails where the URL generation based on the connection has TONS of options because it needs to handle subdomains, default prefixes and what not. Maybe we could call the router whenever there is a connection and by default it would call the endpoint. If you have that then you would be able to do whatever you want... but I think our default, at least for now, is sane. Thanks!

josevalim:
@stevegraham I would rather try to provide a pluggable system for this. So you can generate the URL from the connection when you want. Otherwise we end-up falling in the same trap as Rails where the URL generation based on the connection has TONS of options because it needs to handle subdomains, default prefixes and what not.

Maybe we could call the endpoint whenever there is a connection, but this time, passing the connection too (something like endpoint.url(conn)). The default implementation would just use the pre-calculated url but if you want, you could override this function and then you would be able to do whatever you want... this way we can keep this default but you would be able to do whatever crazy thing you want.

Would this be something you would like to try out?

chrismccord:
> The default implementation would just use the pre-calculated url but if you want, you could override this function and then you would be able to do whatever you want... this way we can keep this default but you would be able to do whatever crazy thing you want.

I believe this is what we settled on in IRC. @josevalim can you confirm? If this is the direction we're going, let's get this in tomorrow. I can take it on.

josevalim:
@chrismccord yes, precisely. Or that, or we don't do anything.

josevalim:
If we are going to add this, we should bring the users that need this feature and ask if it would solve their issue.

josevalim:
Because my main concern is that this feature still wouldn't work for places you don't have a connection. So I would rather tell folks to explicitly use something like `<%= subdomain_url(endpoint, user) %>` or something of sorts.

josevalim:
We are adding a similar feature with #1156.

PR #1114, opened 2015-08-11
Providing port in URL overrides all other ports when calculating URL

gjaldon:
My Phoenix app generates a url like `https://myapp.com:80` because I set `:url` config to `[host: "myapp.com", port:80]` and https. This PR fixes that invalid URL by displaying the https url without the port when it is either `80` or `443`.

Although technically we wouldn't set https to port 80, this was the quickest and least intrusive fix. Another way is to have `url[:port]` be a flag instead. When it's `false`, it will not include port when generating the url. When it's true, it will include the port that we set in `http` and `https` config.

Which is the better way to go?

### DIFF
--- a/lib/phoenix/endpoint/adapter.ex
+++ b/lib/phoenix/endpoint/adapter.ex
@@ -168,7 +168,7 @@ defmodule Phoenix.Endpoint.Adapter do
     port   = port_to_string(url[:port] || port)
 
     case {scheme, port} do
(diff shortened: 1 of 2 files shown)


josevalim:
This is just not correct. Why don't you set the port to 443 in your url host? We use this information *only* to generate URLs.

gjaldon:
Ahh, I see. :sweat_smile: Sorry for the trouble then. :bow:

PR #1123, opened 2015-08-12
add ArgumentError if ets is not started

KazuCocoa:
How about raise an error if ets is not initialized ?

### DIFF
--- a/lib/phoenix/config.ex
+++ b/lib/phoenix/config.ex
@@ -42,7 +42,7 @@ defmodule Phoenix.Config do
   """
   @spec cache(module, term, (module -> {:cache | :nocache, term})) :: term
   def cache(module, key, fun) do
-    case :ets.lookup(module, key) do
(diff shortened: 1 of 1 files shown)


chrismccord:
The solution would be to make sure ets is always started which I thought was default?

KazuCocoa:
Yes, this PR is to check it. I read codes again, you are right.
hm.. I've been think this PR too much handling...

chrismccord:
Thanks for the attempted fix! If you have a way to replicate your original issue, please open an issues with steps to reproduce and I'll see if something needs fixed. Thanks!

KazuCocoa:
Thanks you :)

josevalim:
No, let's add this. It can happen if you call something like `MyApp.Endpoint.url` without starting the endpoint.

josevalim:
Actually, never mind. It would only happen if somehow you don't add phoenix to your supervision tree, which is the default.

Issue #1125, opened 2015-08-14
phoenix.gen.json generator handles output of fields differently from phoenix.gen.html


dmarkow:
When you run `mix phoenix.gen.html User users name:string`, the index/show templates automatically include the `name` field.

However, when you do the same using `phoenix.gen.json`, the map in the `web/views/user_view.ex` file only returns the `id` field:

```elixir
def render("user.json", %{user: user}) do
  %{id: user.id}
end
```

Not sure if this is intentional (I tried searching previous issues and couldn't find anything relevant), but wanted to check before spending time on a PR.

josevalim:
@dmarkow Yes, let's make them consistent. :)

PR #1126, opened 2015-08-14
Include generated model fields in JSON output

dmarkow:
Addresses #1125.

This will include generated fields in the default JSON output. So if you run `mix phoenix.gen.json User users name:string email:string` you'll get the following in the `web/views/user_view.ex` file:

```elixir
def render("user.json", %{user: user}) do
  %{
    id: user.id,
    name: user.name,
    email: user.email
  }
end
```

I added tests similar to how the HTML generator did it (it didn't explicitly check for all fields, just one of them). 

I opted to build the list of fields in the generator itself instead of iterating inside the template as I found it easier to get clean output without extra whitespace, without an errant comma after the last field, etc. This also gracefully handles the case where the user doesn't provide any fields (e.g. `mix phoenix.gen.json User users`).

Let me know if I should approach this differently; any feedback is welcome. Thanks!

### DIFF
--- a/priv/templates/phoenix.gen.json/view.ex
+++ b/priv/templates/phoenix.gen.json/view.ex
@@ -10,6 +10,6 @@ defmodule <%= module %>View do
   end
 
   def render("<%= singular %>.json", %{<%= singular %>: <%= singular %>}) do
-    %{id: <%= singular %>.id}
+    %{<%= json_fields %>}
   end
 end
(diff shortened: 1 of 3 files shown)


priv/templates/phoenix.gen.json/view.ex
```diff
@@ -10,6 +10,8 @@ defmodule <%= module %>View do
   end
 
   def render("<%= singular %>.json", %{<%= singular %>: <%= singular %>}) do
-    %{id: <%= singular %>.id}
+    %{
+      <%= json_fields %>
```
josevalim: Let's please keep the previous syntax:

```
%{id: user.id,
  foo: user.foo,
  ...}
```

dmarkow:
Got it, fixed the syntax. Thanks!

josevalim:

PR #1130, opened 2015-08-14
Improve channel testing helpers

paulcsmith:
I chatted with José in IRC awhile back about adding a subscribe_and_join! function. Here's the PR! I also added some other functions that I have needed when testing channels lately.

* Add subscribe_and_join!/3 for easier for more concise tests
* Add ability to assert that pushes/broadcasts were for a specific
  topic. This makes it possible to test that a push/broadcast was sent for other topics. I

### DIFF
--- a/lib/phoenix/test/channel_test.ex
+++ b/lib/phoenix/test/channel_test.ex
@@ -157,6 +157,21 @@ defmodule Phoenix.ChannelTest do
   end
 
   @doc """
+  Same as subscribe_and_join/3 but returns either the socket or throws an error
+
+  This is helpful when you are not testing authentication and just need the
+  socket.
+  """
+  defmacro subscribe_and_join!(channel, topic, payload \\ Macro.escape(%{})) do
+    quote do
+      case subscribe_and_join(@endpoint, unquote(channel), unquote(topic), unquote(payload)) do
+        {:ok, _, socket} -> socket
+        {:error, error} -> raise "Could not join channel. Got error: #{inspect(error)}"
+      end
+    end
+  end
+
+  @doc """
   Subscribes to the given topic and joins the channel
   under the given topic and payload.
 
--- a/test/phoenix/test/channel_test.exs
+++ b/test/phoenix/test/channel_test.exs
@@ -180,7 +180,7 @@ defmodule Phoenix.Test.ChannelTest do
   end
 
   test "pushes and broadcast messages" do
-    {:ok, _, socket} = subscribe_and_join(Channel, "foo:ok")
+    socket = subscribe_and_join!(Channel, "foo:ok")
     push socket, "broadcast", %{"foo" => "bar"}
     assert_broadcast "broadcast", %{"foo" => "bar"}
   end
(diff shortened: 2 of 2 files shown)


lib/phoenix/test/channel_test.ex
```diff
@@ -356,13 +394,36 @@ defmodule Phoenix.ChannelTest do
...
   end
+
+  @doc """
+  Same as assert_broadcast/2 but adds the option to specify the topic of the broadcast.
+
+      assert_broadcast "room:123", "some_event", %{"data" => _}
+
+  The default timeout is 100ms, but can be customized.
+  """
+  defmacro assert_broadcast(topic, event, payload, timeout \\ 100) do
```
paulcsmith: Oops, looks like this clause won't match. Is there a better way to handle these functions with the default timeout? There is probably something simpler than what I'm trying to do

lib/phoenix/test/channel_test.ex
```diff
@@ -157,6 +157,19 @@ defmodule Phoenix.ChannelTest do
...
 
   @doc """
+  Same as subscribe_and_join/3 but returns either the socket or throws an error
+
+  This is helpful when you are not testing authentication and just need the
+  socket.
+  """
+  defmacro subscribe_and_join!(channel, topic, payload \\ Macro.escape(%{})) do
+    quote do
+      {:ok, _, socket} = subscribe_and_join(@endpoint, unquote(channel), unquote(topic), unquote(payload))
```
josevalim: We need to use a `case` and raise a nice error message. something like: raise "could not join channel, got error: #{inspect error}" would be enough.

josevalim:
> Add ability to assert that pushes/broadcasts were for a specific topic

This is weird because we won't be subscribed to those topics so we don't receive the messages. Maybe if that is the case, it would be better for you to just use `assert_receive` yourself?

josevalim:
Thanks @paulcsmith! I am :+1: on the `subscribe_and_join!`, however, I would like to postpone the topic ones for now.

paulcsmith:
Ok that sounds good. I'll remove the others. Should I open an issue or discussion about testing a specific topic, or just leave it?

lib/phoenix/test/channel_test.ex
paulcsmith: Good idea. I'll add that

josevalim:
> Ok that sounds good. I'll remove the others. Should I open an issue or discussion about testing a specific topic

If you feel it is required, then certainly start a discussion! It would be nice if you had some use cases though, so we are not shooting in the dark. Also, in the process of getting ready to 1.0, we should start moving our discussions to the phoenix-core mailing list and start keeping the issues tracker on actionable issues. I.e. once something is agreed to be part of Phoenix, we move it here.

paulcsmith:
Ah ok, I wasn't sure if we were using the mailing list for that. I'll move the discussion there and give a use case that we've run into. Thanks!

- Sent from my phone. Please excuse typos

paulcsmith:
@josevalim I updated the PR. Is that case statement what you had in mind?

josevalim:

PR #1135, opened 2015-08-17
Rename format to _format and add get_format/1. Closes #1132

### LINKED ISSUE #1132: Consider option to opt-out of format param
We have had our first report of the `format` param clashing with a json post body. We could either namespace this param for less likely clashing, or (my preference), have an accept's plug option to opt-out /set the format param. Something like

```elixir
plug :accepts, formats: ~w(html), param: "format"
# or 
plug :accepts, formats: ~w(html), param: :none
```

We could still have the plug default to accepting a list of formats as it does today for backwards compat. Thoughts?

chrismccord:
@josevalim I could use a review here when you have a moment

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,10 @@
 * Deprecations
   * [Endpoint] `render_errors: [default_format: "html"]` is deprecated in favor of `render_errors: [accepts: ["html"]]`
 
+* Backward incompatible changes
+  * [Controller] - The "format" param for overriding the accept header has been renamed to "_format" and is no longer injected into the params when parsing the Accept headers. Use `get_format/1` to access the negotiated format.
+
+
 ## v0.16.1 (2015-8-6)
 
 * JavaScript client bug fixes
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -469,8 +469,8 @@ defmodule Phoenix.Controller do
     * `conn` - the `Plug.Conn` struct
 
     * `template` - which may be an atom or a string. If an atom, like `:index`,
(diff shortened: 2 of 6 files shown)


chrismccord:
This change also leave the user unable to pattern match the format in the controller, unless they happen to pass the query param. This is easily supported though with a simple plug:

```elixir
pipeline :browser do
  ...
  plug :accepts, ~w(html json)
  plug :put_format_param
end

...
defp put_format_param(conn, _) do
  put_in conn.params, "_format", Phoenix.Controller.get_format(conn)
end
```

Is this sufficient to have users throw into their project as needed?

CHANGELOG.md
```diff
@@ -12,6 +12,10 @@
 * Deprecations
   * [Endpoint] `render_errors: [default_format: "html"]` is deprecated in favor of `render_errors: [accepts: ["html"]]`
 
+* Backward incompatible changes
+  * [Controller] - The "format" param for overriding the accept header has been renamed to "_format" and is no longer injected into the params when parsing the Accept headers. Use `get_format/1` to access the negotiated format.
```
josevalim: `_format` because of markdown syntax.

lib/phoenix/controller.ex
```diff
@@ -605,6 +605,20 @@ defmodule Phoenix.Controller do
   end
 
   @doc """
+  Puts the format in the connection.
+
+  See `get_format/1` for retrieval
```
josevalim: Missing `.`.

josevalim:
:shipit:

josevalim:
> Is this sufficient to have users throw into their project as needed?

Yes.

Issue #1139, opened 2015-08-17
Generated controller tests use atom keys rather than string keys


neoeno:
Summary: phoenix generators generate controller tests that pass atom-keyed maps as controller params, but in production controllers will receive string-keyed maps. This can be misleading if you're new to the framework.

---

Currently `phoenix.gen.html` generates tests more or less like this:

```elixir
defmodule Project.UserControllerTest do
  use Project.ConnCase

  alias Project.User
  @valid_attrs %{name: "some content", age: 42}
  @invalid_attrs %{}

# [...]

  test "creates resource and redirects when data is valid", %{conn: conn} do
    conn = post conn, user_path(conn, :create), user: @valid_attrs
    assert redirected_to(conn) == user_path(conn, :index)
    assert Repo.get_by(User, @valid_attrs)
  end
```

As you can see, `@valid_attrs` is generated using atom keys. This works well enough, but is a little misleading as the controller will actually receive string keys most of the time (ref http://hexdocs.pm/plug/Plug.Conn.html#t:params/0).

Since generators are often used by the newly initiated (read: me) as canonical example code, it would be nice to have them reflect the reality — controllers take string keys.

I wrote a PR to change this, but it turns out that the form of the generated tests relies on `@valid_attrs` being atoms (`Repo.get_by/2` expects atom keys). Since it's named `@valid_attrs`, not `@valid_params`, this makes sense.

So if we were to pass string keys to the controller, we could generate a `@valid_params` that is the same as `@valid_attrs` but with string keys, or convert one to the other — or something else?

What do you think?

I'm happy to work on this — but it'd be good to get a steer on what everyone thinks is the best solution before I dive in.

josevalim:
I completely agree. Please send a PR indeed. :) We can convert the existing parameters.

neoeno:
For anyone who sees this: 

Discovered today that controllers always receive string keys, even if they're passed atom keys by the tests. Which makes this less of an inconvenience than I thought.

PR #1140, opened 2015-08-17
Make mix generators supply string keys for controller params

neoeno:
Ref #1139.

---

Controllers are expected to receive parameters with string keys, but the generators generate tests with atom keys. At best this means the user has to re-write them, at worst it can be misleading about what the standard is.

This commit makes `Mix.Phoenix.params` return a map with string keys, and adds a method `Mix.Phoenix.expected_attr` that returns the same thing with atom keys.

It then amends the generated tests to rename `@valid_attrs` and `@invalid_attrs` to `@*_params`, and adds `@expected_attrs`. The tests themselves are then amended to send `@*_params` to the controller, and test against `@expected_attrs`.

---

I'm guessing a bit with your favoured approach here — I hope you'll set me right if I've gone astray.

### DIFF
--- a/priv/templates/phoenix.gen.html/controller_test.exs
+++ b/priv/templates/phoenix.gen.html/controller_test.exs
@@ -2,8 +2,9 @@ defmodule <%= module %>ControllerTest do
   use <%= base %>.ConnCase
 
   alias <%= module %>
-  @valid_attrs <%= inspect params %>
-  @invalid_attrs %{}
+  @valid_params <%= inspect params %>
+  @expected_attrs <%= inspect expected_attrs %>
+  @invalid_params %{}
 
   setup do
     conn = conn()
@@ -21,13 +22,13 @@ defmodule <%= module %>ControllerTest do
   end
 
   test "creates resource and redirects when data is valid", %{conn: conn} do
-    conn = post conn, <%= singular %>_path(conn, :create), <%= singular %>: @valid_attrs
+    conn = post conn, <%= singular %>_path(conn, :create), <%= singular %>: @valid_params
     assert redirected_to(conn) == <%= singular %>_path(conn, :index)
-    assert Repo.get_by(<%= alias %>, @valid_attrs)
+    assert Repo.get_by(<%= alias %>, @expected_attrs)
   end
 
   test "does not create resource and renders errors when data is invalid", %{conn: conn} do
-    conn = post conn, <%= singular %>_path(conn, :create), <%= singular %>: @invalid_attrs
+    conn = post conn, <%= singular %>_path(conn, :create), <%= singular %>: @invalid_params
     assert html_response(conn, 200) =~ "New <%= singular %>"
   end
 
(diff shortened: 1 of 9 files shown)


priv/templates/phoenix.gen.html/controller_test.exs
```diff
@@ -2,8 +2,9 @@ defmodule <%= module %>ControllerTest do
   use <%= base %>.ConnCase
 
   alias <%= module %>
-  @valid_attrs <%= inspect params %>
-  @invalid_attrs %{}
+  @valid_params <%= inspect params %>
+  @expected_attrs <%= inspect expected_attrs %>
```
neoeno: Alternative approach suggested by @paulanthonywilson:

```elixir
@valid_params <%= inspect params %>
@expected_attrs for {k, v} <- @valid_params, do: {String.to_atom(k), v}
```

josevalim:
Here is my proposal:

1. Keep params with atom keys
2. Add a Mix.Phoenix.stringify_keys/1 and just call it in the test case directly (no need to pass it as a binding)

The reason is: converting atoms to string is more sane than otherwise. And to keep the bindings lighter. Thoughts?

chrismccord:
I have always been fine with the test passing param keys with atoms and assuming they will be converted to to strings. Unless this PR is addressing failing tests, I'm fine with keeping the behavior as is?

josevalim:
@chrismccord I would prefer to map to what is received in the controller. It is much clearer, specially if someone is getting started.

paulanthonywilson:
I like @josevalim's idea of Stringifying the keys in the test, to make it clear that the keys come through as strings in the controller without getting too confusing.

In the simple generated controllers it's not a problem, but I've had issues with adding extra parameters (eg parent ids, or calculated params).  It all worked in the tests, but not when I tried it out in dev due Ecto not being fond of mixing string and atom keys.

(TBH this wasn't actually in controller tests, but I think the principle is the same.)

neoeno:
@josevalim I just gave that a go, but when I ran `mix test` the test case didn't seem to have `Mix.Phoenix.stringify_keys/1` available to it — maybe it's only available in the template and not in the generated test case? Or have I misunderstood something?

josevalim:
Sorry. I don't want to stringify the keys in the test. I proposed to stringify the keys when generating the test file. Something like:

    @valid_params <%= inspect Mix.Phoenix.stringify_keys(params) %> 

Or something of sorts.

josevalim:
Actually, I see the issue raised by @chrismccord. Now we need to keep both stringified and non-stringified version. That's a :-1: from me. Sorry for not seeing it earlier. :(

neoeno:
Oh! I guess that's this closed then. I'm sure it'll get worked out some other way. Feel free to close the issue as well unless there's something else to be done.

Thanks for all your work!

josevalim:
Thank you for the contribution! I am sorry we did not get to merge it this time, but we do appreciate it!

PR #1143, opened 2015-08-20
Make sure json view generator joins fields with commas.

vic:
Previous to this change, the `phoenix.gen.json` generator was
creating invalid elixir code when the model included more than
one field. The generator was not placing commas between fields
for the singular method.

I got into this while creating a brand new phoenix project with
version 0.17 and after doing:

```shell
mix phoenix.gen.json Human humans name age:integer
```

I got the following error while trying to run the migrations:

```
== Compilation error on file web/views/human_view.ex ==
** (SyntaxError) web/views/human_view.ex:14: syntax error before: name
    (elixir) lib/kernel/parallel_compiler.ex:97: anonymous fn/4 in Kernel.ParallelCompiler.spawn_compilers/8
```

The problem was in the generated `web/views/human_view.ex`
missing some commas in the following function:

```elixir
  def render("human.json", %{human: human}) do
    %{id: human.id
      name: human.name
      age: human.age}
  end
```

Modified a test to expose this bug.

### DIFF
--- a/lib/mix/tasks/phoenix.gen.json.ex
+++ b/lib/mix/tasks/phoenix.gen.json.ex
@@ -72,7 +72,7 @@ defmodule Mix.Tasks.Phoenix.Gen.Json do
   defp json_fields(binding, attrs) do
     [{:id, nil}] ++ attrs
     |> Enum.map(fn {k, _} -> "#{k}: #{binding[:singular]}.#{k}" end)
-    |> Enum.join("\n      ")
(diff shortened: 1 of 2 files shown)


chrismccord:
Thanks @vic! Can you add a test-case to ensure this works for the singular field case? I believe it is broken if you only have an id field. Unlikely, but we should account for it.

vic:
Sure, will add a test for it

chrismccord:
<3 <3 <3

PR #1147, opened 2015-08-20
Added method Phoenix.version to allow getting Phoenix version in runtime without hassle.

romul:
At the moment we have no easy way to get Phoenix version in runtime, especially after release.
It's possible:
`:application.loaded_applications |> Enum.find(&( elem(&1, 0) == :phoenix )) |> elem(2)`
But it isn't obvious. 
After some investigations I figured out, that `:application.get_key(:phoenix, :vsn)` could be used here. But I think, that it should be much easier, so I propose to add `Phoenix.version` function to achieve this goal.

### DIFF
--- a/lib/phoenix.ex
+++ b/lib/phoenix.ex
@@ -40,4 +40,12 @@ defmodule Phoenix do
     # Start the supervision tree
     Phoenix.Supervisor.start_link
   end
+
+  @doc """
+  Returns the version of Phoenix
+  """
+  def version do
+    {:ok, vsn} = :application.get_key(:phoenix, :vsn)
+    vsn
+  end
 end


josevalim:
I am :-1:. I would love to add some sort of `Application.get_key/2` to Elixir and make that the official API, I don't think we need to have a function in Phoenix that does just that though.

josevalim:
Thanks for the pull request anyway, we appreciate it! :D

romul:
@josevalim Ok, I'll send PR to Elixir about `Application.get_key/2` && `Application.get_all_keys/1`

romul:
It will be also useful, but getting version will be still not obvious without learning OTP.

PR #1152, opened 2015-08-23
Add overridable event/4 for custom event handling

chrismccord:
The idea here is the overridable `action/2` in controllers is super nice for customizing and injecting arguments to actions, so this adds the same for channel events. Thoughts? Better name than event/4?

### DIFF
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -207,7 +207,11 @@ defmodule Phoenix.Channel do
       import unquote(__MODULE__)
       import Phoenix.Socket, only: [assign: 3]
 
-      def handle_in(_event, _message, socket) do
+      def event(type, event, params, socket) do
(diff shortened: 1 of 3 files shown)


chrismccord:
We could also call it action/4 and use "action" terminology instead of event for channel events. Action feels a little word to me, but it would unify controller and channel concepts, just like calling the payload "params".

chrismccord:
Lastly, I suppose we don't need to inject anything and the user could simply proxy themselves, i.e.:

```elixir
    def handle_in(name, params, socket) do
      handle_in(name, params, socket.assigns.user, socket)
    end
    def handle_out(name, params, socket) do
      handle_out(name, params, socket.assigns.user, socket)
    end
```

Which is probably the clearest approach, and since all the callbacks have the same function name, we don't need the action/4. Aaaaand I think that settles it.

josevalim:
:+1: for not changing anything.

PR #1153, opened 2015-08-23
Make sure controller test from generator uses all model fields

milafrerichs:
## What does this PR do?

This PR will fix a failing test after using the json generator.
```
mix phoenix.gen.json User users name:string
```

I got the following failing test after running the migrations and `mix test `
```
test/controllers/user_controller_test.exs:18
Assertion with == failed
code: json_response(conn, 200)["data"] == %{"id" => user.id()}
lhs:  %{"id" => 20, "name" => nil}
rhs:  %{"id" => 20}
stacktrace:
test/controllers/user_controller_test.exs:21
```
The controller test that fails contains the following lines:
```
assert json_response(conn, 200)["data"] == %{
  "id" => <%= singular %>.id
}
```
but it's missing the name.

I'm new to Phoenix and Elixir and was checking out the generators (great stuff btw) when I stumbled upon this. Thought I might submit a PR.


josevalim:

josevalim:
Beautiful, thank you!

PR #1162, opened 2015-08-28
Full Module Name in Example

cas27:
I believe the full module name should be in the example so it works out of the box without aliasing your channel in user_socket.ex


josevalim:
Yes. :)

chrismccord:
<3 <3 <#

PR #1170, opened 2015-08-31
Add location header to gen.json create action

michalmuskala:
As discussed on the mailing list

### DIFF
--- a/priv/templates/phoenix.gen.json/controller.ex
+++ b/priv/templates/phoenix.gen.json/controller.ex
@@ -50,7 +50,7 @@ defmodule <%= module %>Controller do
 
     # Here we use delete! (with a bang) because we expect
     # it to always work (and if it does not, it will raise).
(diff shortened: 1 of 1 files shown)


priv/templates/phoenix.gen.json/controller.ex
```diff
@@ -49,7 +50,7 @@ defmodule <%= module %>Controller do
 
     # Here we use delete! (with a bang) because we expect
     # it to always work (and if it does not, it will raise).
-    <%= singular %> = Repo.delete!(<%= singular %>)
+    _<%= singular %> = Repo.delete!(<%= singular %>)
```
josevalim: Maybe we just should not assign it to anything? Or it is good because it names the return value?

michalmuskala: I think it would be fine not to assign it at all.

michalmuskala:
Updated

PR #1172, opened 2015-08-31
Support multiple json mime types in json_response/2

scrogson:


lib/phoenix/test/conn_test.ex
```diff
@@ -277,10 +277,27 @@ defmodule Phoenix.ConnTest do
     end
   end
 
+  defp response_content_type?(header, :json) do
+    case parse_content_type(header) do
+      {"application", subpart} ->
+        subpart == "json" || String.ends_with?(subpart, "+json")
```
josevalim: I think we could generalize this and check the subpart or String.ends_with for all formats, without special casing JSON. Thoughts? Something like:

    format = Atom.to_string(format)
    format in Plug.MIME.extensions(part <> "/" <> subpart)
      or format == subpart or String.ends_with?(subpart, format)

Thoughts?

scrogson: Certainly could. I will switch.

josevalim:

PR #1174, opened 2015-08-31
Set hostname to localhost by default for dev and test

### LINKED ISSUE #1173: Set hostname to localhost by default
See: http://iamvery.com/2015/08/29/phoenix-database-config.html

stevedomin:
Fixes #1173 

@josevalim I assumed this was only needed in dev and test but not for prod. Let me know if that's not the case


josevalim:
Isn't the configuraiton "hostname" instead of only "host"?

stevedomin:
:( :( :( so sorry I missed that, my bad

chrismccord:
thanks!

josevalim:

PR #1177, opened 2015-09-01
Consistent quoting in `brunch-config.js`

robhurring:
Update `brunch-config.js` to use single-quotes for all strings.

*Minor nitpick, but my linter complained so I figured I'd "fix" it*


josevalim:
I would use double quotes because Elixir uses double quotes, so it is less context switch. :)

robhurring:
Fair enough. Looking at the other `.js` files they are all double-quotes as well, so if this is about consistency I had it backwards :smile:

chrismccord:
Thanks but double quotes is my pref, which probably goes against the grain form some ja tools

PR #1181, opened 2015-09-03
Strip trailing whitespace in template

henrik:
I have git hooks which warn me about trailing whitespace when I commit, and it complained about these in a freshly generated Phoenix app.

Also consistent with e.g. line 46.


josevalim:

PR #1186, opened 2015-09-06
token `max_age` is checked in ms, make docs consistent

jeffweiss:
Prior to this commit, we had a disconnect between the documented units
(seconds) and the actual units (milliseconds) that were used for
validating the maximum age of a token. This meant that the suggestion of
a 2 week maximum age as 1209600 seconds was, in reality, checked as
milliseconds resulting in a default maximum age of a little over 20
minutes. This commit updates the documentation to reflect that `max_age`
is measured in milliseconds, updates the defaults to their millisecond
equivalents, and tests that the `max_age` does, in fact, validate using
milliseconds.

### DIFF
--- a/lib/phoenix/token.ex
+++ b/lib/phoenix/token.ex
@@ -72,7 +72,7 @@ defmodule Phoenix.Token do
 
     message = %{
       data: data,
-      signed: now_ms()
+      signed: now_seconds()
     } |> :erlang.term_to_binary()
     MessageVerifier.sign(message, secret)
   end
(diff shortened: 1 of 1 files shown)


josevalim:
Good catch @jeffweiss. I would rather make the max_age truly in seconds. Miliseconds is just too low to really matter. @chrismccord?

chrismccord:
Agree

josevalim:
Just a heads up that after we merge this in, we will release 1.0.2. :)

jeffweiss:
Great. I'll fix up this commit to be seconds across the board. I'll likely remove the test then too. I felt it was okay to add a few milliseconds to test it, but adding a second or two to the test run time seems like bad practice.

jeffweiss:
updated

josevalim:

PR #1196, opened 2015-09-08
Fix a typo in socket.js

0xBBB:


josevalim:
Thank you but the code is correct. :) There is no current_user in that function, it is an assignment.

0xBBB:
Sorry.

josevalim:
No problem!

Issue #1192, opened 2015-09-07
Provide --no-html (or --api) on phoenix.new


josevalim:

stevedomin:
I'm fairly keen on tackling this one if no one is on it yet

josevalim:
Please go ahead!

railsmechanic:
I am interested in such an option, too.

jadercorrea:
That's really a nice option to have. :+1:

chrismccord:
:+1: to --api

Tica2:
:+1: to --api

railsmechanic:
:+1: to --api

josevalim:
@chrismccord so the reason I am worried about `--api` is that folks could argue it should not include channels.

evax:
What about finer grained control on what's to be included or not, e.g. `--no-channels`, `--no-html`?

lexmag:
:+1: to `--no-html`

josevalim:
@evax I am :-1: on adding more options unless necessary. The only reason for adding `--no-html` (or `--api`) is because the HTML changes are quite pervasive and touch many files. When it comes to channels, it is mostly a matter of deleting the channel folder.

railsmechanic:
IMHO one of the "main problems" with rails and even the rails-api gem is/was that you have to disable and remove a lot of components manually to get an API foundation of an "acceptable level". As Phoenix is quite new, you're able to integrate this function right from the beginning. So I think a command line switch that bypasses generation of all unused components and leads to a good API foundation without the need of using third party packages.

evax:
@josevalim I get your point. Arguably we could define a `phoenix.new.api` task with it's own set of options to keep `phoenix.new` clean.

josevalim:
@railsmechanic the main problem with Rails is that you need to go through a bunch of hops to disable behaviour in the first place since all of your middleware stack, dependencies and what not are hidden from you.

This is much less of a problem with Phoenix because the phoenix_html dependency is explicit, the imports are explicit, and the pipelines are explicit. So all we need to do is to *not generate code*, in contrast to *generate new code that disables the defaults*.

In other words, Rails gives you `x + y + z` and the only way to remove HTML is by removing `y` and `z` in your app by adding lines of code to remove that functionality. Phoenix gives you `x` and `y + z` exists only in your app, so it is straight-forward to remove them.

[...]

PR #1201, opened 2015-09-10
Add --no-html flag to Phoenix installer

stevedomin:
This allows to skip generating HTML views and including related
dependencies (phoenix_html, phoenix_live_reload).

Fixes https://github.com/phoenixframework/phoenix/issues/1192

### DIFF
--- a/installer/lib/phoenix_new.ex
+++ b/installer/lib/phoenix_new.ex
@@ -17,23 +17,17 @@ defmodule Mix.Tasks.Phoenix.New do
     {:eex,  "new/lib/application_name.ex",                   "lib/application_name.ex"},
     {:eex,  "new/lib/application_name/endpoint.ex",          "lib/application_name/endpoint.ex"},
     {:keep, "new/test/channels",                             "test/channels"},
-    {:eex,  "new/test/controllers/page_controller_test.exs", "test/controllers/page_controller_test.exs"},
+    {:keep, "new/test/controllers",                          "test/controllers"},
     {:eex,  "new/test/views/error_view_test.exs",            "test/views/error_view_test.exs"},
-    {:eex,  "new/test/views/page_view_test.exs",             "test/views/page_view_test.exs"},
-    {:eex,  "new/test/views/layout_view_test.exs",           "test/views/layout_view_test.exs"},
     {:eex,  "new/test/support/conn_case.ex",                 "test/support/conn_case.ex"},
     {:eex,  "new/test/support/channel_case.ex",              "test/support/channel_case.ex"},
     {:eex,  "new/test/test_helper.exs",                      "test/test_helper.exs"},
     {:eex,  "new/web/channels/user_socket.ex",               "web/channels/user_socket.ex"},
-    {:eex,  "new/web/controllers/page_controller.ex",        "web/controllers/page_controller.ex"},
+    {:keep, "new/web/controllers",                           "web/controllers"},
     {:keep, "new/web/models",                                "web/models"},
-    {:eex,  "new/web/templates/layout/app.html.eex",         "web/templates/layout/app.html.eex"},
-    {:eex,  "new/web/templates/page/index.html.eex",         "web/templates/page/index.html.eex"},
-    {:eex,  "new/web/views/error_view.ex",                   "web/views/error_view.ex"},
-    {:eex,  "new/web/views/layout_view.ex",                  "web/views/layout_view.ex"},
-    {:eex,  "new/web/views/page_view.ex",                    "web/views/page_view.ex"},
     {:eex,  "new/web/router.ex",                             "web/router.ex"},
     {:keep, "new/web/static/vendor",                         "web/static/vendor"},
+    {:eex,  "new/web/views/error_view.ex",                   "web/views/error_view.ex"},
     {:eex,  "new/web/web.ex",                                "web/web.ex"},
     {:eex,  "new/mix.exs",                                   "mix.exs"},
     {:eex,  "new/README.md",                                 "README.md"},
@@ -57,6 +51,17 @@ defmodule Mix.Tasks.Phoenix.New do
     {:text, "static/robots.txt",              "web/static/assets/robots.txt"},
   ]
 
+  @html [
+    {:eex,  "new/test/controllers/page_controller_test.exs", "test/controllers/page_controller_test.exs"},
+    {:eex,  "new/test/views/layout_view_test.exs",           "test/views/layout_view_test.exs"},
+    {:eex,  "new/test/views/page_view_test.exs",             "test/views/page_view_test.exs"},
+    {:eex,  "new/web/controllers/page_controller.ex",        "web/controllers/page_controller.ex"},
+    {:eex,  "new/web/templates/layout/app.html.eex",         "web/templates/layout/app.html.eex"},
+    {:eex,  "new/web/templates/page/index.html.eex",         "web/templates/page/index.html.eex"},
+    {:eex,  "new/web/views/layout_view.ex",                  "web/views/layout_view.ex"},
+    {:eex,  "new/web/views/page_view.ex",                    "web/views/page_view.ex"},
+  ]
+
   @bare [
     {:text, "static/bare/.gitignore", ".gitignore"},
     {:text, "static/app.css",         "priv/static/css/app.css"},
@@ -67,7 +72,7 @@ defmodule Mix.Tasks.Phoenix.New do
   # Embed all defined templates
   root = Path.expand("../templates", __DIR__)
 
-  for {format, source, _} <- @new ++ @ecto ++ @brunch ++ @bare do
+  for {format, source, _} <- @new ++ @ecto ++ @brunch ++ @html ++ @bare do
     unless format == :keep do
       @external_resource Path.join(root, source)
       def render(unquote(source)), do: unquote(File.read!(Path.join(root, source)))
@@ -109,6 +114,8 @@ defmodule Mix.Tasks.Phoenix.New do
     * `--no-ecto` - do not generate ecto files for
       the model layer
 
+    * `--no-html` - do not generate HTML views.
+
     * `--binary-id` - use `binary_id` as primary key type
       in ecto models
 
@@ -127,7 +134,7 @@ defmodule Mix.Tasks.Phoenix.New do
   """
   @switches [dev: :boolean, brunch: :boolean, ecto: :boolean,
              app: :string, module: :string, database: :string,
-             binary_id: :boolean]
+             binary_id: :boolean, html: :boolean]
 
   def run([version]) when version in ~w(-v --version) do
     Mix.shell.info "Phoenix v#{@version}"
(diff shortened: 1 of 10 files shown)


installer/lib/phoenix_new.ex
```diff
@@ -153,7 +164,8 @@ defmodule Mix.Tasks.Phoenix.New do
   def run(app, mod, path, opts) do
     db = Keyword.get(opts, :database, "postgres")
     ecto = Keyword.get(opts, :ecto, true)
-    brunch = Keyword.get(opts, :brunch, true)
+    html = Keyword.get(opts, :html, true)
+    brunch = Keyword.get(opts, :brunch, html && true)
```
josevalim: I am not sure if we should do this. I can have my javascript app all written with brunch and just use json. Maybe we do need an `--api` flag that is a shortcut for `--no-html` and `--no-brunch`?

installer/lib/phoenix_new.ex
```diff
@@ -184,6 +196,7 @@ defmodule Mix.Tasks.Phoenix.New do
                static_deps_prefix: static_deps_prefix,
                brunch: brunch,
                ecto: ecto,
+	       html: html,
```
josevalim: indententation

installer/templates/new/lib/application_name/endpoint.ex
```diff
@@ -1,7 +1,7 @@
 defmodule <%= application_module %>.Endpoint do
   use Phoenix.Endpoint, otp_app: :<%= application_name %>
 
-  socket "/socket", <%= application_module %>.UserSocket
+  socket "/socket", <%= application_module %>.UserSocket<%= if html do %>
```
josevalim: I would keep Plug.Static regardless. :)

installer/templates/new/lib/application_name/endpoint.ex
```diff
@@ -9,13 +9,13 @@ defmodule <%= application_module %>.Endpoint do
   # when deploying your static files in production.
   plug Plug.Static,
     at: "/", from: :<%= application_name %>, gzip: false,
-    only: ~w(css fonts images js favicon.ico robots.txt)
+    only: ~w(css fonts images js favicon.ico robots.txt)<% end %>
 
   # Code reloading can be explicitly enabled under the
   # :code_reloader configuration of your endpoint.
-  if code_reloading? do
+  if code_reloading? do<%= if html do %>
```
josevalim: Sweet.

installer/templates/new/mix.exs
```diff
@@ -33,9 +33,9 @@ defmodule <%= application_module %>.Mixfile do
   defp deps do
     [<%= phoenix_dep %>,<%= if ecto do %>
      {:phoenix_ecto, "~> 1.1"},
-     {<%= inspect adapter_app %>, ">= 0.0.0"},<% end %>
+     {<%= inspect adapter_app %>, ">= 0.0.0"},<% end %><%= if html do %>
```
josevalim: Double sweet.

josevalim:
Thank you @stevedomin! I have added some comments. I think you forgot a "Phoenix.HTML" inside `web/web.ex`. You can test your new changes inside the installer directory:

    cd installer
    mix phoenix.new demo --dev --no-html
    mix test

If the tests pass, we are good to go. :)

installer/lib/phoenix_new.ex
stevedomin: That makes sense.

Part of me like it, another part feels like we already have a lot of flags.

josevalim: Let's keep the `--api` flag out for now, we can start another discussion for it. :)

stevedomin:
@josevalim this what the view function look like in `web/web.ex` with the --no-html flag.

```
def view do
    quote do
      use Phoenix.View, root: "web/templates"

      # Import convenience functions from controllers
      import Phoenix.Controller, only: [get_csrf_token: 0, get_flash: 2, view_module: 1]

      import PhotoBlog.Router.Helpers
    end
  end
```

Do you think we should remove that as well?

```
use Phoenix.View, root: "web/templates"

# Import convenience functions from controllers
import Phoenix.Controller, only: [get_csrf_token: 0, get_flash: 2, view_module: 1]
```

josevalim:
No, let's keep them. get_flash is likely not necessary BUT the other two may still be useful, so better to not change because of one import.

stevedomin:
Ok, then I think it's ready :)

installer/lib/phoenix_new.ex
```diff
@@ -57,6 +51,19 @@ defmodule Mix.Tasks.Phoenix.New do
     {:text, "static/robots.txt",              "web/static/assets/robots.txt"},
   ]
 
+  @html [
+    {:eex,  "new/test/controllers/page_controller_test.exs", "test/controllers/page_controller_test.exs"},
+    {:eex,  "new/test/views/error_view_test.exs",            "test/views/error_view_test.exs"},
```
josevalim: We also use the error view for json. I would just generate it as is.

installer/lib/phoenix_new.ex
```diff
@@ -57,6 +51,19 @@ defmodule Mix.Tasks.Phoenix.New do
...
+    {:eex,  "new/test/views/layout_view_test.exs",           "test/views/layout_view_test.exs"},
+    {:eex,  "new/test/views/page_view_test.exs",             "test/views/page_view_test.exs"},
+    {:eex,  "new/web/controllers/page_controller.ex",        "web/controllers/page_controller.ex"},
+    {:eex,  "new/web/templates/layout/app.html.eex",         "web/templates/layout/app.html.eex"},
+    {:eex,  "new/web/templates/page/index.html.eex",         "web/templates/page/index.html.eex"},
+    {:eex,  "new/web/views/error_view.ex",                   "web/views/error_view.ex"},
```
josevalim: So we should also keep the error view tests.

josevalim:
I added a couple more comments. We also need to change the config files.

1. this should be only json:

https://github.com/phoenixframework/phoenix/blob/master/installer/templates/new/config/config.exs#L17

2. this can be skipped:

https://github.com/phoenixframework/phoenix/blob/master/installer/templates/new/config/dev.exs#L18-L25

And I think that's all!

stevedomin:
@josevalim I fixed the issues you reported

Not too happy with the error view as it will be trying to render "4|5xx.json" but as it won't find the template it's going to use 500.html. I feel like the installer could use a bit of refactoring, what do you think? I wrote down a few suggestions, I can send to phoenix-core if that's useful.

josevalim:
@stevedomin should we fix the views for json then? generate 404.json and 500.json returning maps?

> I wrote down a few suggestions, I can send to phoenix-core if that's useful.

Please do!

stevedomin:
@josevalim here are the files I changed:

[error_view_test.exs](https://github.com/stevedomin/phoenix/blob/no-html/installer/templates/new/test/views/error_view_test.exs)
[error_view.ex](https://github.com/stevedomin/phoenix/blob/no-html/installer/templates/new/web/views/error_view.ex)
[phoenix_new_test](https://github.com/stevedomin/phoenix/blob/no-html/installer/test/phoenix_new_test.exs)

I'm not sure if that's the correct way of writing the test.

installer/templates/new/test/views/error_view_test.exs
```diff
@@ -17,5 +17,18 @@ defmodule <%= application_module %>.ErrorViewTest do
   test "render any other" do
     assert render_to_string(<%= application_module %>.ErrorView, "505.html", []) ==
            "Server internal error"
+  end<% else %>test "renders 404.json" do
+    assert Poison.decode!(render_to_string(<%= application_module %>.ErrorView, "404.json", [])) ==
```
josevalim: In those cases, I would just call `render`, and not `render_to_string`, and compare the returned map with a map. :)

josevalim:
@stevedomin for JS view tests, we can call `render` instead of `render_to_string` and just directly compare maps. :)

stevedomin:
@josevalim of course, that makes total sense... This is now fixed.

josevalim:
@chrismccord this is ready to ship. Please review and merge :)

josevalim:
@stevedomin  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

optikfluffel:
You're awesome

installer/templates/new/web/views/error_view.ex
```diff
@@ -13,5 +13,17 @@ defmodule <%= application_module %>.ErrorView do
   # template is found, let's render it as 500
   def template_not_found(_template, assigns) do
     render "500.html", assigns
+  end<% else %>def render("404.json", _assigns) do
+    %{"errors" => %{"detail" => "Page not found"}}
   end
+
+  def render("500.json", _assigns) do
+    %{"errors" => %{"detail" => "Server internal error"}}
```
chrismccord: should we prefer string keys here? My personal preference is atom keys when constructing json maps by hand. Strings everywhere is more consistent, but a bit cumbersome.

chrismccord:
I have one small comment. Once addressed (or not if we're happy with it), this is ready to go!

installer/templates/new/web/views/error_view.ex
josevalim: Don't we recommend string keys for payloads in the channel layer? I would return whatever the channel layer returns.

chrismccord: > Don't we recommend string keys for payloads in the channel layer? I would return whatever the channel layer returns.

Not in our doc examples. My general rule is string keys coming in from the outside, and atoms going out

josevalim: Ok! @stevedomin sorry for one extra back and forth but can you please change them to use atom keys and fix the generated tests accordingly? Thank you!

fbjork:
Can't wait to see this work merged. Nice work @stevedomin!

chrismccord:
<3 <3 <3

stevedomin:
Thank you for merging @chrismccord, was on holidays so couldn't change the keys to atoms :)

emilsoman:
@stevedomin I generated an app with `--no-html` , now when I tried to run the app with `mix phoenix.server`, I get the following error:

```
== Compilation error on file web/views/error_view.ex ==
** (CompileError) web/views/error_view.ex:2: module MyApp.ErrorHelpers is not loaded and could not be found
    expanding macro: MyApp.Web.__using__/1
    web/views/error_view.ex:2: MyApp.ErrorView (module)
    (elixir) expanding macro: Kernel.use/2
    web/views/error_view.ex:2: MyApp.ErrorView (module)
    (elixir) lib/kernel/parallel_compiler.ex:100: anonymous fn/4 in Kernel.ParallelCompiler.spawn_compilers/8
```

If I remove the `web/views` directory, it works fine. I'm using phoenix v1.1.0. Here's the full command that I used to generate the app:

```
mix phoenix.new my_app --no-ecto --no-brunch --no-html --module MyApp
```

Here's a link to the repo containing the source: https://github.com/emilsoman/my_app . Am I doing anything wrong ?

Issue #1198, opened 2015-09-08
Provide aliases for ecto.reset and ecto.setup


josevalim:
They are convenient and will fit as an example of aliases. Here is the code:

```elixir
["ecto.setup": ["ecto.create", "ecto.migrate", "run priv/repo/seeds.exs"],
 "ecto.reset": ["ecto.drop", "ecto.setup"]]
```

Thoughts?

josevalim:
Alternatively, we can ship with those by default on Ecto, but Ecto wouldn't load the seeds file.

dukex:
I'm newbie and I don't understand the big picture but looks nice the ```ecto.setup```/```echo.reset``` in phoenix(running ```seeds.exs```)

chrismccord:
:+1: but I need to open that issue on ecto where drop/reset raises if not dev or test (with --force option). I like that this give a mix alias example.

parkerl:
:heart: this idea but I would like to see `ecto.reset` and the like in ecto proper.

Tica2:
In ecto issue was here:
https://github.com/elixir-lang/ecto/issues/936

stevedomin:
@chrismccord are you referring to this issue: https://github.com/elixir-lang/ecto/issues/934

gjaldon:
I agree with @parkerl. Would be nice to have these conveniences in Ecto.

josevalim:
If those conveniences are in ecto, they can't do things like running seed files.

PR #1202, opened 2015-09-10
Add ecto.reset and ecto.setup aliases

stevedomin:
Fixes https://github.com/phoenixframework/phoenix/issues/1198

I took my comment in mix.exs from the Mix doc but I'm wondering if we should put something else. Did you have something in mind @josevalim?

Also, I'm not sure how useful is it to test it but I added some tests just in case.


josevalim:
I like this. @chrismccord, just merge if you agree. :)

chrismccord:
Beautiful.

PR #1224, opened 2015-09-19
Allow generating --dev projects outside Phoenix source

jeffkreeftmeijer:
Right now, generating a Phoenix project with the `--dev` flag throws an error if the generated project is not inside the `installer` directory in the Phoenix source;

```
  $ mix phoenix.new /Users/jeff/opensource/phoenix_edge --dev
  Compiled lib/phoenix_new.ex
  Generated phoenix_new app
  ** (Mix) --dev project must be inside Phoenix directory
```

After poking around for a bit, I wasn’t able to find a reason for this, and I found that projects generated with the `--dev` flag would work properly anywhere if I just changed the path to Phoenix to an absolute one in the project’s `mix.exs` file.

So, I removed the code that checks if the project path is inside `phoenix/installer` and simply returned the absolute path to the Phoenix source when the `--dev` flag is used, which results in this line in the generated mix.exs file;

```
  [{:phoenix, path: "/Users/jeff/opensource/phoenix", override: true},
```

That eliminates the "--dev project must be inside Phoenix directory”-error and cleans up `installer/lib/phoenix_new.ex` quite a bit. What do you think? Did I miss anything?

### DIFF
--- a/README.md
+++ b/README.md
@@ -26,8 +26,6 @@ $ mix phoenix.new path/to/your/app --dev
 
 The command above will create a new application, using your current Phoenix checkout thanks to the `--dev` flag.
 
-Note that `path/to/your/app` must be within the directory containing the Phoenix source code. This is so that a relative path can be used for the `:phoenix` dependency.
-
 ### Building phoenix.js
 
 ```bash
--- a/installer/lib/phoenix_new.ex
+++ b/installer/lib/phoenix_new.ex
@@ -182,7 +182,6 @@ defmodule Mix.Tasks.Phoenix.New do
                application_module: mod,
                phoenix_dep: phoenix_dep(phoenix_path),
                phoenix_path: phoenix_path,
-               phoenix_static_path: phoenix_static_path(phoenix_path),
                pubsub_server: pubsub_server,
                secret_key_base: random_string(64),
                prod_secret_key_base: random_string(64),
@@ -462,8 +461,6 @@ defmodule Mix.Tasks.Phoenix.New do
   defp phoenix_dep("deps/phoenix"), do: ~s[{:phoenix, "~> 1.0.2"}]
   defp phoenix_dep(path), do: ~s[{:phoenix, path: #{inspect path}, override: true}]
 
-  defp phoenix_static_path("deps/phoenix"), do: "deps/phoenix"
-  defp phoenix_static_path(path), do: Path.join("..", path)
 
   defp phoenix_path(true), do: @phoenix
   defp phoenix_path(false), do: "deps/phoenix"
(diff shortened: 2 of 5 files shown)


josevalim:
Hey @jeffkreeftmeijer, the issue I ran with this was related to brunch. I believe brunch can't find your assets if you use absolute paths. Try scaffolding a resource in an app with --dev and see if the delete button works and see if Phoenix.Socket is there.

josevalim:
And thanks for the PR!

jeffkreeftmeijer:
Hey @josevalim, thanks for the pointers and your quick reply. Brunch worked fine for me, so I did some testing and found that the assets were generated properly on Brunch 1.8.5, but not on earlier versions.

So, depending on "^1.8.5" instead of "^1.8.1" would solve this issue. Do you want me to change that?

josevalim:
Yes, please. I am also worried about this:

https://github.com/jeffkreeftmeijer/phoenix/blob/dev-projects-outside-phoenix-directory/installer/lib/phoenix_new.ex#L466

This means the brunch file has a path to "../Users/jeff/..". It would be really weird if that is working. If you can check that too, it would be great, if not, I can check it before merging. Thank you!

jeffkreeftmeijer:
After some more testing, I found `Phoenix.Socket` was indeed missing, caused by using the absolute path to phoenix.js in the Phoenix source in `web/static/js/app.js`;

```
  import {Socket} from "/Users/jeff/opensource/phoenix/web/static/js/phoenix"
```

It works with a relative path, so I think our best bet is to reinstate `phoenix_static_path` and have that return a relative path from the project directory to the phoenix source directory. What do you think?

Also, I tried finding something like [Ruby's Pathname#relative_path_from](http://apidock.com/ruby/Pathname/relative_path_from) in Elixir, but couldn't find it. Does a function like that exist, before I reinvent the wheel?

josevalim:
We have `Path.relative_from` but it is not guaranteed to give a relative path if they are two distinct path structures. In other words, if you have it /foo/bar and /bat/bat, it won't make it ../../foo/bar because they are disjoint.

At this point, I am wondering if it is worth to continue chasing this. The `--dev` flag is a development only thing, so we can try things out, and requiring the same filepath seems like an OK restriction.

jeffkreeftmeijer:
Yes, I found `Path.relative_from`, but wondered if I was missing anything.

This seemed like a simple thing to do as my first patch, but I agree this has gotten out of hand a little, so I'll close this. Thanks a lot for your help!

wsmoak:
Thanks for exploring whether this could be changed!  https://github.com/phoenixframework/phoenix/pull/1227 proposes adding more info to the README so the next person along doesn't have to re-discover it. :)

PR #1226, opened 2015-09-19
Fixes wrong reply tuple in Phoenix.Channel moduledocs

ephe-meral:
`{:ok, socket}` is not a correct result according to Phoenix.Channel


josevalim:

PR #1227, opened 2015-09-20
Improve docs on using the --dev switch

wsmoak:
- clarify that --dev must be used in the installer directory and with a relative path to the project to be generated
- note that Mix will use an installed archive before the local source code
- link to related issue and PR for more info
- prefer 'project' to 'app' to avoid confusion with OTP Applications


josevalim:

PR #1229, opened 2015-09-24
Transform controller action function clause errors into 400's. Closes…

chrismccord:
@josevalim ready for review. How lenient or strict should we be on the action clause match? I made it so we match on actions that have %Plug.Conn{}, %{}, and any further args.

### DIFF
--- a/lib/phoenix/controller/pipeline.ex
+++ b/lib/phoenix/controller/pipeline.ex
@@ -82,6 +82,7 @@ defmodule Phoenix.Controller.Pipeline do
   which is responsible for dispatching the appropriate action
   after the plug stack (and is also overridable).
   """
+  require Logger
 
   @doc false
   defmacro __using__(_) do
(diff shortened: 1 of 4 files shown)


lib/phoenix/controller/pipeline.ex
```diff
@@ -125,11 +126,40 @@ defmodule Phoenix.Controller.Pipeline do
...
+            )
+        end
       end
     end
   end
 
+  @doc false
+  def __catch__(conn, :error = kind, :function_clause = reason, stack, controller, action) do
+    case stack do
+      [{^controller, ^action, [%Plug.Conn{}, %{} | _], location} | _] ->
```
josevalim: I would check only the first argument. :) Also, I think you could just raise a Phoenix specific error in this clause, something like:

```elixir
defmodule Phoenix.ActionClauseError do
  defexception [conn: ..., plug_status: 400, controller: nil, action: nil]

  # ...
end
```

And let the error layer take care of it. This way  you don't need to log things like file, line and stacktrace and parameters. The error layer will take care of it for us.

lib/phoenix/controller/pipeline.ex
```diff
@@ -125,11 +126,40 @@ defmodule Phoenix.Controller.Pipeline do
         _ = var!(controller)
         _ = var!(action)
 
-        unquote(body)
+        try do
+          unquote(body)
+        catch
+          kind, reason ->
+            Phoenix.Controller.Pipeline.__catch__(
```
josevalim: Because this is wrapping the whole pipeline. The chance of false positives can be big. Maybe we should override `action/2` only?

lib/phoenix/controller/pipeline.ex
jeregrine: :+1: on error.

lib/phoenix/exceptions.ex
```diff
@@ -20,3 +20,14 @@ defmodule Phoenix.MissingParamError do
...
   end
 end
+
+defmodule Phoenix.ActionClauseError do
+  defexception [message: nil, plug_status: 400, controller: nil, action: nil]
+
+  def exception(opts) do
+    msg = "bad request to #{inspect opts[:controller]}.#{opts[:action]}, " <>
+          "no matching action clause to process request"
+    %Phoenix.ActionClauseError{message: msg}
```
josevalim: Or you should pass the other options in (controller + action) or you should remove them from `defexception`.

chrismccord:
Merged in 4ad8aa2799f7b8a7b38bd8f8b745847d71ca2cd6

PR #1230, opened 2015-09-24
Fix incorrect error message on channel join pattern mismatch.

CD1212:
The pattern expected on error seems to only require `{:error, reply :: map}` whereas the outputted error also mentioned a socket was required. Hope this is correct and not something I'm missing.

### DIFF
--- a/lib/phoenix/channel/server.ex
+++ b/lib/phoenix/channel/server.ex
@@ -165,7 +165,7 @@ defmodule Phoenix.Channel.Server do
 
             {:ok, Socket.t} |
             {:ok, reply :: map, Socket.t} |
-            {:error, reply :: map, Socket.t}
+            {:error, reply :: map}
 
         got #{inspect other}
         """


chrismccord:
The socket is required. In what cases are you seeing otherwise?

CD1212:
If my channel `join` function returns
```
{:error, %{message: "No auth_token supplied"}, socket}
```

I get an exception eg.

```
** (EXIT from #PID<0.281.0>) an exception was raised:
         ** (RuntimeError) Channel join is expected to return one of:
     
         {:ok, Socket.t} |
         {:ok, reply :: map, Socket.t} |
         {:error, reply :: map, Socket.t}
     
     got {:error, %{message: "No auth_token supplied"}, %Phoenix.Socket{...}}
```

But if the `join` function just returns the below, everything seems to work fine.
```
{:error, %{message: "No auth_token supplied"}}
```

I wondered if the `raise` error message was just incorrect with what is matched. I'm very new to Elixir though, so I might not be working out what is / isn't matched correctly.

chrismccord:
Ah, sorry, you are right. Ignore me! :) Thanks!

PR #1231, opened 2015-09-25
change `sockets` to `channels` in Transport.dispatch

trarbr:
Noticed this when reading through the source for the longpoll and websocket transports.

The documentation also mentions that transports dispatch to channels: https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/socket/transport.ex#L179

### DIFF
--- a/lib/phoenix/socket/transport.ex
+++ b/lib/phoenix/socket/transport.ex
@@ -191,14 +191,14 @@ defmodule Phoenix.Socket.Transport do
       must be sent as result
 
   """
-  def dispatch(msg, sockets, socket)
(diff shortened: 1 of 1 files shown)


chrismccord:
Thanks, but can we revert this? Sockets are multiplexed, so it might have appear confusing, but each channel has its own socket. `sockets` in this context was correct

chrismccord:
Ignore me, @josevalim changed the transports to use `channels` instead of `sockets` as their state name, and I missed this change in https://github.com/phoenixframework/phoenix/commit/bbebb81faf2178b263eb8f88430b5c0260d26cd7
:+1: for consistency

PR #1233, opened 2015-09-27
fix link to official site doc

y-yagi:
I think with www is correct URL.


chrismccord:
We have a www redirect rule, so I think the naked domain is fine for now. Thanks for the PR!

y-yagi:
OK. Thank you for explanation!

PR #1235, opened 2015-09-28
Adding brackets to match rest of style

leighhalliday:
I noticed that all other method calls in this controller
use brackets. Adding brackets to match this style
in the render call of the show action.


chrismccord:
<3 <3 <3

Issue #832, opened 2015-04-30
Generate implementations of Phoenix.Param on the fly


josevalim:
Today, `Phoenix.Param` is expensive because falling back to Any is expensive as it implies a code server miss (and therefore traversing the code path).

We can optimize this if we automatically generate implementation for the struct every time we reach `Phoenix.Param`. However, we should only do this if the protocol was not consolidated and currently there is no way to check this in Elixir v1.0. However, we can implement this feature in Elixir v1.1 and therefore speed up the development environment considerably with this. We would need to be careful though to not define the same module over and over again (a central dispatch would be needed).

Adding this as an issue so we don't forget this feature in the future.

Related to #823

acconrad:
Looks like @josevalim [released 1.1.0-dev](https://github.com/elixir-lang/elixir/blob/master/CHANGELOG.md) 2 days ago, how would one go about implementing this? Branch to a 1.0.1?

josevalim:
I have not release 1.1.0 and we cannot add this whole 1.0 is supported. :)

-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

josevalim:
We no longer need this since Elixir v1.2 will always consolidate.

PR #1244, opened 2015-10-02
Remove Access protocol and use Dict.get

dukex:
With this commit all tests are green on elixir 1.1.0

### DIFF
--- a/test/phoenix/integration/long_poll_test.exs
+++ b/test/phoenix/integration/long_poll_test.exs
@@ -235,7 +235,7 @@ defmodule Phoenix.Integration.LongPollTest do
       :timer.sleep @ensure_window_timeout_ms
 
       resp = poll(:post, "/ws", session)
-      assert resp.body["status"] == 410
+      assert Dict.get(resp.body, "status") == 410
       assert_receive {:DOWN, _, :process, ^channel, {:shutdown, :inactive}}
     end
   end


josevalim:
This should not be required as the previous syntax works fine with Elixir 1.1. The `opts[key]` syntax is not deprecated.

dukex:
:ok_hand: I  made this pull request only because I ran the tests and received a error (I can't reproduced now) but it's very strange because now I see this file has many `opts[key]`

So there is any special thing to not use elixir 1.1.0 and close https://github.com/phoenixframework/phoenix/issues/832

dukex:
I found, running ```mix test -h``` and got

```
** (exit) an exception was raised:
    ** (Phoenix.Socket.InvalidMessageError) missing key "topic"
        (phoenix) lib/phoenix/socket/message.ex:31: Phoenix.Socket.Message.from_map!/1
        (phoenix) lib/phoenix/transports/long_poll.ex:171: Phoenix.Transports.LongPoll.publish/4
        (plug) lib/plug/adapters/cowboy/handler.ex:15: Plug.Adapters.Cowboy.Handler.upgrade/4
        (cowboy) src/cowboy_protocol.erl:442: :cowboy_protocol.execute/4


  1) test pubsub: shuts down after timeout (Phoenix.Integration.LongPollTest)
     test/phoenix/integration/long_poll_test.exs:227
     ** (FunctionClauseError) no function clause matching in Access.fetch/2
     stacktrace:
       (elixir) lib/access.ex:48: Access.fetch("", "status")
       (elixir) lib/access.ex:72: Access.get/3
       test/phoenix/integration/long_poll_test.exs:238
```

pma:
Hi,

I've had this problem before and figured it's related to some .beam files
in _build and/or deps previously compiled in pre 1.1.x. So I did:

1. rm -rf _build deps
2. mix deps.get
3. mix compile

And apparently this fixed the problem.

chrismccord:
Thanks for taking a look, but this is actually a race condition in our long poller tests that is triggered occasionally. So the symptom was an Access.get error, but the Dict.get would be the same issue. The problem is the request body is blank

dukex:
Thanks for explication @chrismccord

PR #1246, opened 2015-10-03
Phoenix.Controller.Pipeline docstring update

djm:
Hi,

Really small edit for clarity. I believe `or` is the better wording than `and` for this line explaining the plug guards as the action can only be one value at a time.

Cheers,
Darian


chrismccord:
Thanks!

PR #1261, opened 2015-10-09
Adds Phoenix.Template.HTML as a default js encoder

JustMikey:
I was not sure if it was worth creating another module specifically for javascript (like `Phoenix.Template.Javascript`) since it would do exactly the same so I added `Phoenix.Template.HTML` as the default encoder for js.


josevalim:

PR #1264, opened 2015-10-09
Phoenix.Digest mix task digests asset urls in stylesheets

tokafish:
This enhances the mix phoenix.digest task to replace assets found in stylesheets with their digested paths. The method is a simple regex, following the [spec](http://www.w3.org/TR/CSS2/syndata.html#value-def-uri).

A couple things to note:
* Absolute URLs found in stylesheets are not processed
* Absolute paths are assumed to be rooted in the directory being digested (with the manifest)
* Relative paths are handled
* If the asset is not found in the manifest, the stylesheet is left untouched.

Also, I thought it would be more consistent to only add this behavior to the digested version of the stylesheet. In other words, `app.css` is left unchanged, but the assets are replaced with digested versions in `app-afa653ae9d91f8a94534f8933a496bf8.css`

I feel like I could test it a bit more comprehensively if it were in a separate module, but I wanted to get overall feedback before continuing further.

See discussion [here](https://groups.google.com/forum/#!topic/phoenix-talk/5XRjFRrxlT0)


lib/phoenix/digester.ex
```diff
@@ -21,11 +25,13 @@ defmodule Phoenix.Digester do
...
       unless File.exists?(output_path), do: File.mkdir_p!(output_path)
 
-      input_path
-      |> filter_files
-      |> do_compile(output_path)
-      |> generate_manifest(output_path)
-      :ok
+      digested_files = input_path
+        |> filter_files
+        |> Enum.map(&(digest(&1)))
```
josevalim: `&digest/1`

josevalim:
@tokafish this looks great, thank you! Should we add a test that checks that the absolute one was not digested?

josevalim:
@tokafish this looks great, thank you! Should we add a test that checks that the absolute one was not digested?

chrismccord:
This is awesome! And it's build-tool agnostic which is great. We also need to make it aware of `static_url` so you can use your cdn and such.

tokafish:
@josevalim I added the additional assertions. @chrismccord I'm not sure exactly how it would interact `static_url`, let me know what you have in mind and I'll be happy to add to the commit.

chrismccord:
Perhaps the task can take an Endpoint as an option, but default to App.Endpoint? Then we can lookup the endpoint's static_url for that environment.

tokafish:
I'm still getting my head around the codebase, so I'm not sure how the
defaulting would work - but once static_url was available, it would prefix
URLs in the digested stylesheet?

tokafish:
@josevalim @chrismccord Not sure how I should proceed on this PR. Personally, I don't think using static_url is necessary. There was some discussion [here](https://groups.google.com/forum/#!topic/phoenix-talk/5XRjFRrxlT0), but it seems to me if a user wants to use a CDN, simply hosting the stylesheet from the CDN should be enough, and then the URLs within it will be interpreted properly (even if they are relative and don't specify a host).

chrismccord:
Your are right, ignore me :)

I'll review and merge soon. Thanks!

chrismccord:
This is so nice. Thank you!

Issue #1265, opened 2015-10-09
phoenix.js can't leave() if channel is not joined


chrismccord:
Easily reproduced by trying to leave on a failed join. The same would happen for leaving on an `onError` callback:

```javascript
// channel onClose callbacks are never triggered and socket membership is not cleaned up
// This is because we wait for the server to ack our leave to avoid races, but 
// we aren't joined.
chan.join().receive("error", reason => chan.leave() )
```

chrismccord:
cc @ericmj

josevalim:
@chrismccord we need to solve this along side the other channel issue and think about a new release.

chrismccord:
Closed in https://github.com/phoenixframework/phoenix/commit/58405d6a490fb32050e7ae111147931462c06761

PR #1267, opened 2015-10-10
Do not buffer channel pushes and fix leave bug whe not joined.

### LINKED ISSUE #1241: Discuss if messages should be automatically queued on the client

chrismccord:
Closes #1241 #1265 

@josevalim this solves both issues, but we need to discuss if we really want to ditch the send buffer. I can go either way, but I've heard that this has helped folks in practice. // @ericmj @ernie

This approach simply immediately invokes the "error" match if the channel is not joined. The leave bug is fixed by making sure we cleanup on both "ok" and "error". Thoughts?

### DIFF
--- a/priv/static/phoenix.js
+++ b/priv/static/phoenix.js
@@ -1,42 +1,59 @@
-(function(/*! Brunch !*/) {
+(function() {
   'use strict';
 
-  var globals = typeof window !== 'undefined' ? window : global;
+  var globals = typeof window === 'undefined' ? global : window;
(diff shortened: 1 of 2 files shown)


josevalim:
I remember we discussed a hybrid approach. We will submit again until a timeout is reached. We also had to define what would be the API for such timeouts `receive("timeout")` or `after(...)`. We can talk about this later!

ernie:
@chrismccord this is way cleaner than the hack I coded up. Not that I'm surprised. :)

Thanks!

PR #1269, opened 2015-10-11
fix comment for socket ids

tony612:


chrismccord:
Thanks, but I think the 's is okay in this case :)

tony612:
@chrismccord OK, thanks for your reply :)

PR #1270, opened 2015-10-11
Use npm script for building brunch

jarofghosts:
This updates the installer to utilize an [npm run script](https://docs.npmjs.com/cli/run-script) for running `brunch build`.

npm run scripts are run with any locally installed bins in the `PATH`, so that you can call the executables directly without having to call into `node_modules/...` or ask that someone install a dependency globally. This is preferable because the current method of reaching into `node_modules/...` is prone to breaking if the `brunch` package ever changes where it stores the bin, whereas this will always work as long as it defines an executable in its `package.json`.

### DIFF
--- a/installer/lib/phoenix_new.ex
+++ b/installer/lib/phoenix_new.ex
@@ -293,7 +293,7 @@ defmodule Mix.Tasks.Phoenix.New do
   end
 
   defp install_brunch(install?) do
-    maybe_cmd "npm install && node node_modules/brunch/bin/brunch build",
+    maybe_cmd "npm install && npm run brunch-build",
               File.exists?("brunch-config.js"), install? && System.find_executable("npm")
   end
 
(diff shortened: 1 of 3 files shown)


chrismccord:
Can you verify that this works on windows? I know we had to invoke node explicitly to cover some windows issues

chrismccord:
Verify in that it really works, not "should/supposed to work" :)

jarofghosts:
definitely! will report back

jarofghosts:
verified working on windows, given recently pushed commit

chrismccord:
<3 <3 <3

josevalim:
Should we also use this for the brunch watch inside config/dev.exs?

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

jarofghosts:
ah, yes, i missed that usage. i can open a new PR for that later today

PR #1273, opened 2015-10-13
Use npm script for brunch watch tasks

jarofghosts:
This is a follow-up to conversation that happened in #1270 . Moves the `brunch watch --stdin` call into an npm script.

### DIFF
--- a/installer/templates/new/config/dev.exs
+++ b/installer/templates/new/config/dev.exs
@@ -12,7 +12,7 @@ config :<%= application_name %>, <%= application_module %>.Endpoint,
   code_reloader: true,
   cache_static_lookup: false,
   check_origin: false,
-  watchers: <%= if brunch do %>[node: ["node_modules/brunch/bin/brunch", "watch", "--stdin"]]<% else %>[]<% end %>
+  watchers: <%= if brunch do %>[npm: ["run", "brunch-watch"]]<% else %>[]<% end %>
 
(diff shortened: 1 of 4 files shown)


josevalim:
:+1:

chrismccord:
Thanks!

chrismccord:
This change actually breaks brunch usage for someone on windows :(

I was troubleshooting a different brunch issue and had them try out the new npm script and watcher, and they get an eaacces error. We may need to roll this one back

jarofghosts:
EACCESS sounds like a permission problem, is it possible it's unrelated?

chrismccord:
Yes, but it works fine under the previous watcher configuration, so it's not unrelated. The previous watcher config that this replaced was added specifically to workaround some perm issues we had on windows

jarofghosts:
Ah, I'm sorry. I confirmed it working, but it was only on Windows 8, maybe there is some difference on different versions.

chrismccord:
yeah I believe you :) I'm not sure how prevalent the perm issue with be for windows with this change. Maybe an isolated case, but who knows

josevalim:
Don't forget to also revert #1270 then!

rstacruz:
> EACCESS sounds like a permission problem

hypothesis: sounds like the kind of error you'd run into when the binstubs (node_modules/.bin) aren't made properly. maybe `node_modules` was copied over from an osx/linux installation?

PR #1276, opened 2015-10-13
serve all static files by default

troyk:
The new app template was limiting static file serving to `only: ~w(css fonts images js favicon.ico robots.txt)`

I think if a file exists in /priv/static that it should be served for the following reasons:

  * it is common to deploy behind a proxy where the proxy will first check the uri in a static location and then forward to the dynamic backend
  * element of least surprise, user happiness:  other frameworks and static file servers I use serve the entire dir by default.  Thus I'm submitting this pull request because this caused me some delay figuring out why my `js` and `css` files were being served but not `img` (note, I use img by convention over images)

I'm not sure if this needs testing, but I did run mix test before and after all tests passed.

Thanks,

Troy

### DIFF
--- a/installer/templates/new/lib/application_name/endpoint.ex
+++ b/installer/templates/new/lib/application_name/endpoint.ex
@@ -8,8 +8,7 @@ defmodule <%= application_module %>.Endpoint do
   # You should set gzip to true if you are running phoenix.digest
   # when deploying your static files in production.
   plug Plug.Static,
-    at: "/", from: :<%= application_name %>, gzip: false,
-    only: ~w(css fonts images js favicon.ico robots.txt)
+    at: "/", from: :<%= application_name %>, gzip: false
 
   # Code reloading can be explicitly enabled under the
   # :code_reloader configuration of your endpoint.


chrismccord:
Thanks, but we use explicit dirs so we can avoid hitting disk on every request

troyk:
@chrismccord that makes sense, in fact I would like to disable the static plug all together in production.  Would you be interested in making static configurable like code reloading is?

chrismccord:
You can do that yourself with something like: 

```elixir
# config/*.exs
  ...
  serve_static: true

# endpoint
if Application.get_env(:my_app, __MODULE__)[:serve_static] do
  plug Plug.Static, ...
end
```

troyk:
@chrismccord Thanks Chris, works like a charm.  We are a week or so from deploying to prod but when we do I'll check into the perf of disabling the static plug in production if it's not needed and put a blog post out if the results are interesting enough

josevalim:
Btw, why disable it in production? Usually there is no need to put nginx nor apache in front of your Phoenix application. :) Or are you using a CDN?

troyk:
@josevalim no CDN, maybe one day.  nginx (or apache) kick butt at serving static files and terminating ssl, I can't imagine not using it.  In our setup, nginx handles assets via pagespeed module (but looking forward to falling forward to http2), caching backed by redis, ssl termination, we serve statics from several dirs (e.x.: marketing gets the write carte blanche to the mktg dir via their static site gen tool), stuff like that.

josevalim:
Sounds great, thanks for clarifying!

Issue #1286, opened 2015-10-16
URL helpers aren't properly escaping dynamic segments


foxnewsnetwork:
Say I have an `Account` model with an `id` field that is a base64 encoded string. If I run this through the URL helpers generated by the `Phoenix.Router`, it will fail if the base64 string happens to have slashes or pluses in it.

For example: If I have:

```elixir
account = %Account{id: "sPupZ6593LQ16nwMYChvCRqcATwKpqo10KAil7oGV4E="}
path = conn |> account_path(:show, account.id)
conn |> get(path, [])
# Everything is fine, 200

account = %Account{id: "k7mboJUEaF32w66hqN4LP7Y/v5+/R8029Wy7Iu4MUmo="}
path = conn |> account_path(:show, account.id)
conn |> path(path, [])
# 406 error
# The slashes aren't escaped, and the router 406s due to it the request is some weird format
```

Looking through the code:

https://github.com/phoenixframework/phoenix/blob/v1.0.3/lib/phoenix/router/helpers.ex#L175

It seems we just sort drop the segments without escaping them...

Can we get some sort of fix where instead of:
```elixir
defp segments(segments, [], _reserved) do
    segments
end
```
We have something like:
```elixir
defp segments(segments, [], _reserved) do
    encode_uri_component segments
end
```

josevalim:
If you call `URI.encode_www_form` here in those two places

1. https://github.com/phoenixframework/phoenix/blob/v1.0.3/lib/phoenix/router/helpers.ex#L173
2. https://github.com/phoenixframework/phoenix/blob/v1.0.3/lib/phoenix/router/helpers.ex#L170

We should be good to go? can you please try a patch with tests? Thank you!

foxnewsnetwork:
okay, understood

PR #1287, opened 2015-10-16
added encode_www_form to make sure the path helpers are properly escaping dynamic segments

foxnewsnetwork:
This is in reference to:

https://github.com/phoenixframework/phoenix/issues/1286

@josevalim suggested to put the `encode_www_form` inside `to_param`, however, because we already use `to_param` to properly encode querystrings, doing so resulted in a double encoding. The next best place was to place the `encode_www_form` directly inside the `expand_segments` function.

Admittedly, I am a little surprised all the tests passed despite the fact I didn't also add `encode_www_form` to L238. I can spend a little more time trying to break into corner cases, but here's the working pr for now.

### DIFF
--- a/test/phoenix/router/helpers_test.exs
+++ b/test/phoenix/router/helpers_test.exs
@@ -33,7 +33,7 @@ defmodule Phoenix.Router.HelpersTest do
 
     assert extract_defhelper(route, 1) == String.strip """
     def(hello_world_path(conn_or_endpoint, :world, bar, params)) do
-      path(conn_or_endpoint, segments(("" <> "/foo") <> "/" <> Enum.join(bar, "/"), params, ["bar"]))
+      path(conn_or_endpoint, segments(("" <> "/foo") <> "/" <> Enum.map_join(bar, "/", &(URI.encode_www_form() / 1)), params, ["bar"]))
     end
     """
   end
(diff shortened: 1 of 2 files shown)


josevalim:
Ah, good catch. to_param is called from other places and i missed that!

josevalim:
@foxnewsnetwork we need to change line 238 too. Maybe by using `Enum.map_join` and passing encode as the mapper function. That clause is called when you have `/foo/*path` segments. I hope this info helps you write another failing test. :)

foxnewsnetwork:
Got it, thank you. Will get on it.

foxnewsnetwork:
okay, as per your instructions, I replaced `Enum.join` with `Enum.map_join` and added a bunch more tests. 

I was, at first, concerned with the fact I didn't need to change the code at

https://github.com/phoenixframework/phoenix/blob/v1.0.3/lib/phoenix/router/helpers.ex#L238

to also include `URI.encode_www_form`, but after tinkering around a bit more with the code, I realized that part would only require changing if someone decides to use odd characters in the non-dynamic parts of their route in the router like so:

```elixir
get "/dog+=3ad/:id", DogController, :show
```
As this use-case seems rather silly, I opted to not alter the code at L238 nor to write tests for this.

test/phoenix/router/helpers_test.exs
```diff
@@ -152,6 +152,22 @@ defmodule Phoenix.Router.HelpersTest do
     end
   end
 
+  test "top-level named routes with complex ids" do
+    assert Helpers.post_path(__MODULE__, :show, "8==D--+") ==
```
josevalim: The choice of characters `8==D` do not really belong to a technical codebase. If you haven't noticed it, it is OK because I haven't either the first time I reviewed it. Can we pick something else? Dropping the number 8 would be fine.

foxnewsnetwork:
Acknowledged, it is done. 

On an unrelated note, I noticed the following test would fail on running the tests after restarting my computer:

```bash
1) test starts watching and writes to stdio (Phoenix.Endpoint.WatcherTest)
     test/phoenix/endpoint/watcher_test.exs:7
     No message matching {:DOWN, ^ref, :process, ^pid, :normal} after 100ms. The process mailbox is empty.
     stacktrace:
       (ex_unit) lib/ex_unit/capture_io.ex:99: ExUnit.CaptureIO.do_capture_io/3
       test/phoenix/endpoint/watcher_test.exs:8
```
I'm pretty sure no amount of strings comparing "8"s to "D"s on my part could be causing this... can it be 100ms is too short on a fresh ram computer?

josevalim:
@foxnewsnetwork it is a race condition somewhere. :(

foxnewsnetwork:
@josevalim I can't reproduce this fail test on my machine any more, regardless of restarts or whatever. I will try doing it on travis

josevalim:
Merged manually, thank you!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

foxnewsnetwork:
np (I'm sorry I occasionally have the mind of a 12 year old lol).

I have no idea what could be causing the race condition, but I've managed to reproduce it after rebuilding a couple of times on travis:

https://travis-ci.org/foxnewsnetwork/phoenix/builds/85776438#L335

I have no idea what could be causing this. But it seems to just be capture_io related, so could it be that this bug isn't even in phoenix?

josevalim:
Elixir v1.1 fixed a race condition on capture_io, so it may as well be fixed in there. :)

PR #1289, opened 2015-10-16
Support creating unique indexes with the phoenix.gen.model mix task

Theemuts:
This is what issue #1281 is aiming for, if I'm not mistaken. The unique indexes can be generated either as field:unique or field:unique:datatype, the former defaults to :string.


josevalim:
@Theemuts beautiful! Can we make it "attr:type:unique" though? We also need to update the docs. :)

lib/mix/phoenix.ex
```diff
@@ -109,18 +110,20 @@ defmodule Mix.Phoenix do
...
+        {k, :float}              -> {k, "120.5"}
+        {k, :decimal}            -> {k, "120.5"}
+        {k, :boolean}            -> {k, true}
+        {k, :map}                -> {k, %{}}
+        {k, :text}               -> {k, "some content"}
+        {k, :date}               -> {k, "2010-04-17"}
+        {k, :time}               -> {k, "14:00:00"}
+        {k, :datetime}           -> {k, "2010-04-17 14:00:00"}
+        {k, :uuid}               -> {k, "7488a646-e31f-11e4-aace-600308960662"}
+        {k, {:unique, :integer}} -> {k, 42}
```
josevalim: What if someone uses unique:float or unique:uuid? I am wondering if we should pre-process the unique out before reaching this function, similar to how we do a couple lines above with references. :)

josevalim: What if someone uses unique:float or unique:uuid? I am wondering if we should rewrite this function to this:

```
Enum.into(%{}, fn
  k, {:unique, t} -> {k, type_to_default(t)}
  k, t -> {k, type_to_default(t)}
end)

defp type_to_default(t) do
  case t do
    :integer -> 42
    :float -> "120.5"
    :decimal -> "120.5"
    ...
  end
end
```

Thoughts?

lib/mix/tasks/phoenix.gen.model.ex
```diff
@@ -87,11 +93,12 @@ defmodule Mix.Tasks.Phoenix.Gen.Model do
 
     Mix.Phoenix.check_module_name_availability!(binding[:module])
 
+    {uniques, attrs} = extract_uniques(attrs)
```
josevalim: What if we change `attrs` to return `{attrs, uniques}`? I think it will cause less changes to the code, otherwise we need to filter out the `:unique` in a bunch of different places like we do today. :)

josevalim: Ignore me again. Sorry. :)

lib/mix/tasks/phoenix.gen.model.ex
```diff
@@ -183,8 +204,9 @@ defmodule Mix.Tasks.Phoenix.Gen.Model do
 
   defp types(attrs) do
     Enum.into attrs, %{}, fn
-      {k, {c, v}} -> {k, {c, value_to_type(v)}}
-      {k, v}      -> {k, value_to_type(v)}
+      {k, {:unique, v}} -> {k, value_to_type(v)}
```
josevalim: Do we need this? We extract the uniques out of `attrs` before we call `types/1`.  :)

priv/templates/phoenix.gen.model/model.ex
```diff
@@ -20,5 +20,6 @@ defmodule <%= module %> do
   def changeset(model, params \\ :empty) do
     model
     |> cast(params, @required_fields, @optional_fields)
-  end
+<%= for {k, _} <- uniques do %>    |> unique_constraint(<%= inspect k %>)
```
josevalim: Beautiful.

josevalim:
This is great @Theemuts! I was wondering if we should make it "attr:type:unique" although I see why you went with "attr:unique:type". Thoughts?

lib/mix/phoenix.ex
Theemuts: Good point, stupid oversight from my end. I think that should work, the only change is that it might be easier to work with {t, :unique} if I change the task to work with attr:type:unique.

Theemuts:
I've added the type_to_default function, changed the command to attr:type:unique, and removed the unnecessary case.

josevalim:

Theemuts:
I was wondering if it would be better to pattern match the type_to_default function, rather than using case. Also, I've forgotten to update the documentation, I'll fix that right away.

josevalim:
Oops, I just merged. So please send another PR. I think case is cleaner though, less repetition on this case.

PR #1290, opened 2015-10-16
Update bootstrap version to latest (3.3.5)

wiserfirst:
Just noticed the bootstrap version in a newly generated phoenix app is still 3.1.1, which was released early last year. I'm not sure if there is a particular reason why this has not been updated, but since all 3.x versions of bootstrap are supposed to be compatible, updating to 3.3.5 shouldn't break anything.


chrismccord:
Have you verified that things look appropriate?

wiserfirst:
@chrismccord I built and installed the installer ez package, and generated a new test app with it. The home page looks exactly the same as before on Chrome, Firefox and Safari.

chrismccord:
<3 <3 <3

wiserfirst:
:tada:

PR #1292, opened 2015-10-16
Update HTML layout to use sectioning elements

napcs:
This commit improves accessibility in the default layout by incorporating HTML5 sectioning elements instead of `div` elements. It adds additional elements but does not change the appearance.

* Replace `div` for header with `header` element
* Add `nav` element with `role="navigation"`
* Wrap import with `main` element
* Move `role="main"` from outer `div` to new `main` element as `main` role is for the main content of the document.

I have left the `class` attributes intact so as not to break the CSS.

### DIFF
--- a/installer/templates/new/web/templates/layout/app.html.eex
+++ b/installer/templates/new/web/templates/layout/app.html.eex
@@ -12,18 +12,22 @@
   </head>
 
   <body>
-    <div class="container" role="main">
-      <div class="header">
-        <ul class="nav nav-pills pull-right">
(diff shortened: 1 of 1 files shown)


chrismccord:
a11y is very important. Have you verified that this actually maps 1:1 with bootstraps current styles? If so, I say it looks good to me

napcs:
@chrismccord Yes. It looks right to me.

See <http://skitch.bphogan.com/2015-10-16_hw1flac01fdx8ex_Hello_Phoenix.png>

chrismccord:
<3 <3 <3

PR #1293, opened 2015-10-16
Add support for unique arrays and foreign keys in the phoenix.gen.model mix task

Theemuts:
I've extended the functionality of the code I submitted earlier today to also allow unique foreign keys with "user_id:references:users:unique" and unique arrays with "nicknames:array:string:unique"

### DIFF
--- a/lib/mix/phoenix.ex
+++ b/lib/mix/phoenix.ex
@@ -94,7 +94,7 @@ defmodule Mix.Phoenix do
   def attrs(attrs) do
     Enum.map(attrs, fn attr ->
       attr
-      |> String.split(":", parts: 4)
+      |> String.split(":", parts: 3)
       |> list_to_attr()
       |> validate_attr!()
     end)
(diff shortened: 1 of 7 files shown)


josevalim:
Thank you @Theemuts! I think we need to solve this differently. What it `attrs` already removed the unique information and returned `attrs`, as before, as the list of unique names? This code is getting more and more complex because we need to change a lot of places to support the unique field. So I am thinking if we should just split and if the last bit is "unique", we take the unique information right at the moment we parse and store the unique information elsewhere.

Thoughts?

Theemuts:
I feel like I've managed to make the code a little less complex by prepending the unique-tag, but I think this isn't what you're aiming for (especially since I've now had to introduce a new function in order to strip that info for the html and json-generators)

josevalim:
Actually, it is probably best if attrs just ignore the unique information
and we have another function that receives args and parses the unique
fields out. We can even use String.ends_with?(args, ":unique") to find out
the unique ones. Thoughts?



-- 


Theemuts:
Mix.Phoenix.attrs is called from all three generators, so if that function is reverted to its old form, all three generators will first need to go over the received attrs to check which ones are unique, remove the trailing :unique's, and return a list with the unique attrs and those sanitized attrs. 

The list of unique keys is then used to check which references don't get an index and have to be added to the required_attrs, and to generate the unique constraints and the unique_indexes.

josevalim:
@Theemuts that's exactly why I am proposing to not change the return results of attrs. Let's make it return what it did before, it does not return the unique information at all. Then, the model generator, the only one really interested in the unique information, can parse the original attributes and get only the ones that are unique. Does it make sense?

Theemuts:
I think this is much more along the lines you're thinking about

josevalim:
@Theemuts I have pushed a commit to master: 5f230cf52a5f601e4a1cd35dbb6a2d8ba8082728

The idea is that all of the unique handling is in the `uniques` function, the attrs function does not really care about it. Can you please give a shot at implementing this feature on master? Theoretically, you only need to change the uniques function to also consider references and when returning the name and you should be good to go!

chrismccord:
@Theemuts ping. How are we looking on this?

chrismccord:
@Theemuts please ping this PR when you'd like to continue and I will re-open. Thanks!

PR #1296, opened 2015-10-16
Just updating the docs to render the raw function

bryansray:
Simple docs for render_existing to leverage the raw/1 function. Thanks to @chrismccord for the help.

### DIFF
--- a/lib/phoenix/view.ex
+++ b/lib/phoenix/view.ex
@@ -229,7 +229,7 @@ defmodule Phoenix.View do
   either a precompiled template, or by implementing the function directly, ie:
 
       def render("scripts.html", _assigns) do
-        "<script src=\"...\">"
+        raw("<script src=\"...\">")
       end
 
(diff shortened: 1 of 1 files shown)


chrismccord:
Let's use `~E` so we don't show an example using `raw` ?

josevalim:
@chrismccord oops, sorry. :)

chrismccord:
No worries :) I recommended @bryansray submit a pr with `raw` earlier, but I just realized ~E would be better so we don't promote the use of raw as much as possible

josevalim:
I agree. @bryansray would you be kind to send another PR that flips it to `~E`?

bryansray:
@josevalim @chrismccord certainly. I'm not entirely sure I did this correctly, but there's a new PR using the sigil. Thanks, guys.

https://github.com/phoenixframework/phoenix/pull/1301

PR #1297, opened 2015-10-17
Run `npm cache clean` before running `npm install`

juandazapata:
Sometimes, `npm` will fail to install some packages because its cache
gets in some weird state. Running `npm cache clean` ensures that the
installation is done with fresh packages instead of the cached ones.


josevalim:
Thank you @juandazapata but I would rather not do this. If installation fail, they still get the error on the next command or when they run the server. I don't think we should try to work around npm bugs. :)

juandazapata:
Fair enough! Thanks for replying @josevalim

PR #1303, opened 2015-10-17
Fix socket.ex docu

ephe-meral:
`MyApp.Endpoint.broadcast/2` is not generated by default by the endpoint, needs to have a 3rd parameter as well.


ephe-meral:
Generally speaking this sample line might be misleading...
I thought it would do the things that it says in the comment, however this will only happen if the channels define a handler for this event, no?

chrismccord:
No, it does forcefully kick the multiplexed socket, i.e.
https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/transports/websocket.ex#L140-L143

Thanks!

PR #1312, opened 2015-10-26
Moved subscribe and unsubscribe to Local + deletion optimization

gabiz:
First commit moves the subscribe and unsubscribe to local + removes write_concurrency and public flags.
Second commit implements optimized deletion using a reverse ETS to avoid contention when clients are dropped.
Separate commits allow measuring performance on each change separately.


chrismccord:
Moving the ets work back inside the server with public write_concurreny removal did not cause any difference. So this is good. The code is simpler this way without client offloading. The optimize deletion did also not appear to make a difference, but it's definitely the way to go. I will merge soon after a few more tests. Thanks!!!

josevalim:
I would like to sign this one off too before merging if you are ok with it
Chris. I will ready only by the end of the day, however.

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

chrismccord:
> would like to sign this one off too before merging if you are ok with it Chris. I will ready only by the end of the day, however.

"I will merge soon after a few more tests" was code for "I hope José signs online so I can get him to sign off :)". I'll catch up with you later :)

gabiz:
Got that :) I am surprised that deletion performance was the same. How did you test deletion performance? With a separate async GC is going to return immediately but the process queue will get backlogged.

josevalim:
If you can't notice differences though, it would be nice to prepare a new
batch of tests, for example, the ones that connect only for 15 seconds and
then closes the connection. Hopefully it will allow us to measure
improvements.

-- 


josevalim:
To be fair, we don't care about that process queue. If it takes too long to
remove the item, it will stay in the ets, dead, but it should not affect
anything else.

I am not saying we should let it queue necessarily on purpose but it is not
a constraint we have. :)


-- 


gabiz:
It is not entirely accurate that items in the ets are dead and should not affect anything else.
The entries pending deletion in the ets are returned in lookups and messages may be sent to dead processes.
In addition, the lookup time is not constant but it increases with the number of entries returned (due to the copy overhead) so having entries pending deletion increases the lookup time.
For example on my mac one lookup of 3k entries takes 1ms, 30k entries takes 10ms and 300k entries takes 100ms. These latencies increase linearly when doing concurrent lookups from separate processes at the same time.
It is possible that all this is negligible, however it may be worth considering lookups when measuring performance.

chrismccord:
> I am surprised that deletion performance was the same.

It turns out, there is a match_delete bottleneck that this PR solves..... at 1.5 million clients! :)

We still have more tests to do, so I'll have more info soon.

chrismccord:
Below is master vs this PR. You can see we hit a bottleneck after 1 million clients, and with the PR we are solid. Having said that, this PR causing floods of subscribe timeouts, which we don't get on master (with the queuing GC):

master_delete_bottleneck
master_optimized_delete

gabiz:
Another issue that I noticed is that the unsubscribe requests are handled in the context of the GC gen_server, so if the process queue gets backlogged then unsubscribe call requests may timeout. If we end up keeping the GC approach we should probably move the unsubscribe to the context of the caller.

chrismccord:
That's a very good point

gabiz:
If we are getting timeouts then the gen_server is becoming the bottleneck.
Another option then is to handle the subscriptions and unsubscriptions on the context of the caller (change the ETSs to write_concurrency) and then Local server effectively becomes the GC.

chrismccord:
We are doing that in master, where we offload ets inserts on subscribe to the caller. I think we should continue to do this, but the issue is when we get a DOWN, we don't have a caller to offload to, other than offloading to the GC.

gabiz:
Aren't the DOWN events queued in the Local process queue even before the handle_info is called?

chrismccord:
yeah they are, which there isn't much we can do there other than pool local servers, but then it's hard to say if we're optimizing for real workloads (thousands of clients dropping at exactly the same time)

josevalim:
Yes Gabi! The linear costs are accurate but it is worth mentioning a dead
process does not perform copying on message send. Anyway, this is an
improvement according to numbers and that beats any assumption. :)

More notes:

1. Can we add the features in this PR but keeping the GC process?

2. Unsubscribe is rarely called in Phoenix, so while moving it to the
client is a great idea, the focus should be down messages

In other words: let's add the new table but keep things in the client as
much as we can. Is it possible?



-- 


gabiz:
Yes, let's recap:
1) Keep same structure as in master: Local + GC.
2) Add the reverse ETS table to handle deletes.
3) Move unsubscribe to the client on the GC module.
Is this right?

josevalim:
If that works out it sounds good!



-- 


gabiz:
Two comments:
- Moved `unsubscribe
When moving the `unsubscribe` to the caller context the local_server from the gen_server state is not available anymore to access the ETS table. Passing

gabiz:
Two comments:
- Moved the `unsubscribe` to the Local server. Rationale: On the context of the GC caller there is no access to the local_server stored in the state of the gen_server so there is the need to pass this explicitly each time as an extra parameter. Since this is called in the context of the caller there is no difference if it is located in the Local or GC. In the local it requires less parameters and seems more consistent to me.
- We previously called `unsubscribe` to avoid race conditions in the test. But now we do not have a handle_call for this. So I added a dummy one.

chrismccord:
Thanks @gabiz!  We're still seeing a bad dropoff at 1.3M subscribers (but not as many timeouts). If we aren't at the limits of the ets setup, the only way to go from here that I can think of is to pool Local and GC servers and randomly choose a pool when subscribing/unsubscribing. I quickly hacked this approach before the ets duplicate_bag change, and it helped, but it may be dancing around another problem just before. The main problem is our single Local server handling DOWN of thousands of clients at once. I see no way around this than pooling.

screen shot 2015-10-26 at 11 40 32 pm

chrismccord:
If we pool locals, then we may be able to get away with not having a GC since the pool will distribute the cleanup load (maybe). Othewrise we'll need two pools. One for local, and one for GC

chrismccord:
Okay, we have 2.1M subscriber with few timeouts, via a random pool with subscriber sharding (just a spike atm): https://github.com/phoenixframework/phoenix/commit/8eb9dfa9edc842b474c36e081c2361717eda8e82

@mobileoverlord had the neat idea of trialing a sharded subscriber lookup so we could make a broadcast with millions of subscribers snappy, by Task.async'ing the broadcast per shard. We trialed this and it worked, so I combined the shard idea with a Local pool idea. So this both pools Local servers who manage their own ets tables, which serve as the shards for pubsub subscriptions. One thing that is completely broken in that PR is unsubscribing, since I choose a shard at random, so you'd have to hit all locals to unsub right now. Buth with a shard size of 250, this holds 2M connections *and the chat is totally usable*, snappy even. My mind is being blown over here. Thoughts?

screen shot 2015-10-27 at 2 01 24 am

gabiz:
Awesome!
Breaking up the huge ETS into smaller ones is certainly the way to go. 
On the other hand creating on the fly pool_size * Tasks per broadcast is not going to allow scaling up messages/sec which I think is the most important benchmark (more clients can be handled by horizontal scaling).
Maybe preallocating a dedicated pool of handlers per ETS ("shard") could address efficient broadcasts.

chrismccord:
Thanks @gabiz! Merged in https://github.com/phoenixframework/phoenix/pull/1334
<3 <3 <3

PR #1315, opened 2015-10-28
Fix scrub_params/2 typespec

lpil:
Hello!

I'm not 100% sure I'm correct here, but I think that this typespec stated that this function takes a list of params, but the implementation takes a binary.

Cheers,
Louis


chrismccord:
thanks!

lpil:
:dancer:

PR #1325, opened 2015-10-30
Allow async replies. Closes #1324

### LINKED ISSUE #1324: Add reply/3 to channel api
So you can reply out-of-band to a push by providing the ref

chrismccord:
@josevalim I'd like to get your sign-off on this one. I think it's time we can bring back `reply` in the channel API. If you remember, we had `reply`, which we renamed to `push` when we added true replies with the 3-tuple `{:reply, ..., socket}`. We've had the first use-case of needing to async reply to a push so it's time to add this feature. I think it has been plenty long since the name change that this won't cause problems. This is analogous to `GenServer.reply`.

### DIFF
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -176,6 +176,9 @@ defmodule Phoenix.Channel do
   alias Phoenix.Channel.Server
 
   @type reply :: status :: atom | {status :: atom, response :: map}
+  @type socket_ref :: {transport_pid :: Pid, serializer :: Module.t,
+                       topic :: binary, ref :: binary}
+
 
   defcallback join(topic :: binary, auth_msg :: map, Socket.t) ::
               {:ok, Socket.t} |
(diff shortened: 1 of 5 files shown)


lib/phoenix/channel.ex
```diff
@@ -338,13 +338,39 @@ defmodule Phoenix.Channel do
...
+        {:noreply, socket}
+      end
+
+      def handle_info({:work_complete, result, ref}, socket) do
+        reply socket, ref, {:ok, result}
+        {:noreply, socket}
+      end
+
+  """
+  def reply(socket, ref, {status, payload}) do
```
josevalim: My reservation is that we should not depend on the socket. The socket has the whole assigns which we don't want to be passing around. Thoughts?

chrismccord: Analogous to GenServer.reply, maybe isn't the right word, but the idea is you generally call `reply` within the context of your channel server, and only pass around your `socket.ref`, like the example I layed out. The other option is to force users to provide `socket.transport_pid`, but then we are leaking a little bit the details. This works well for Tasks for example:

      def handle_in("work", params, socket) do
        ref = socket.ref
        Task.async(fn -> reply(socket, ref, {:ok, expensive_call(params) end)
        {:noreply, socket}
      end

chrismccord:
@josevalim ideas we discussed on irc are implemented. Let me know how it looks

lib/phoenix/channel.ex
```diff
@@ -338,13 +341,55 @@ defmodule Phoenix.Channel do
     Server.push(transport_pid, topic, event, message, socket.serializer)
   end
 
+  @doc """
+  Replies asynchronously to a socket push.
+
+  Useful when you need to reply to a push that can't otherwise be handled using
```
josevalim: Should we document that, if you are using tasks, to make sure to pass only the socket_ref? To be more precise, those docs say "how" but not the "why". It would be nice to explain folks why there is a need for socket_ref and why it is a bad idea to pass the socket around.

chrismccord: Yes good idea

PR #1334, opened 2015-11-05
Shard PubSub with Local and GC Pool

chrismccord:
TODO

  - [x] Do not spawn tasks if pool size is 1


lib/phoenix/pubsub/gc.ex
```diff
@@ -14,49 +14,53 @@ defmodule Phoenix.PubSub.GC do
...
+      end
+      true = :ets.match_delete(state.pids, {pid, :_})
+    catch
+      :error, :badarg -> :badarg
+    end
+
     {:noreply, state}
   end
+
+  def handle_call(_, _from, state) do
```
josevalim: You can remove this one and let the default one.

lib/phoenix/pubsub/local.ex
```diff
@@ -129,39 +173,90 @@ defmodule Phoenix.PubSub.Local do
...
+    if opts[:link], do: Process.link(pid)
     Process.monitor(pid)
-    {:reply, :ok, state}
+    {:reply, {:ok, {state.topics, state.pids}}, state}
+  end
+
+  def handle_call({:subscription, pid}, _from, state) do
+    try do
+      {:reply, :ets.lookup_element(state.pids, pid, 2), state}
+    catch
```
josevalim: Can't we get the table in the client because it is in ETS?

lib/phoenix/pubsub/local.ex
```diff
@@ -129,39 +173,90 @@ defmodule Phoenix.PubSub.Local do
...
+  def handle_call({:subscribe, pid, _topic, opts}, _from, state) do
+    if opts[:link], do: Process.link(pid)
     Process.monitor(pid)
-    {:reply, :ok, state}
+    {:reply, {:ok, {state.topics, state.pids}}, state}
+  end
+
+  def handle_call({:subscription, pid}, _from, state) do
+    try do
+      {:reply, :ets.lookup_element(state.pids, pid, 2), state}
```
josevalim: Can't we get the subscription in the client because the table is in ETS?

lib/phoenix/pubsub/local.ex
```diff
@@ -129,39 +173,90 @@ defmodule Phoenix.PubSub.Local do
...
+    |> local_for_shard(pubsub_server)
+  end
+
+  defp local_for_shard(shard, pubsub_server) do
+    {local_server, _gc_server} = pools_for_shard(shard, pubsub_server)
+    local_server
+  end
+
+  defp pools_for_shard(shard, pubsub_server) do
+    [{^shard, {local_server, gc_server}}] = :ets.lookup(pubsub_server, shard)
```
josevalim: `[{^shard, {_, _} = servers}]`?

lib/phoenix/pubsub/local.ex
```diff
@@ -129,39 +173,90 @@ defmodule Phoenix.PubSub.Local do
...
-    |> Enum.uniq
+  def subscription(pubsub_server, pool_size, pid) when is_atom(pubsub_server) do
+    pid
+    |> pid_to_shard(pool_size)
+    |> local_for_shard(pubsub_server)
+    |> GenServer.call({:subscription, pid})
+  end
+
+  @doc false
+  def local_name(pubsub_server, shard) do
```
josevalim: Nitpick: I don't like those two functions. I think those names should be generated just in the supervisor init and adding those functions will allow us to be lazy in the future and start relying on them at runtime.

lib/phoenix/pubsub/local_supervisor.ex
```diff
@@ -0,0 +1,46 @@
...
+  end
+
+  @doc false
+  def init([server, pool_size, dispatch_rules]) do
+    # Define a dispatch table so we don't have to go through
+    # a bottleneck to get the instruction to perform.
+    ^server = :ets.new(server, [:set, :named_table, read_concurrency: true])
+    true = :ets.insert(server, {:subscribe, Phoenix.PubSub.Local, [server, pool_size]})
+    true = :ets.insert(server, {:unsubscribe, Phoenix.PubSub.Local, [server, pool_size]})
+    for rule <- dispatch_rules do
```
josevalim: I see, sweet.

lib/phoenix/pubsub/pg2.ex
```diff
@@ -15,29 +15,21 @@ defmodule Phoenix.PubSub.PG2 do
...
     children = [
-      worker(Phoenix.PubSub.Local, [local_name, gc_name]),
-      worker(Phoenix.PubSub.GC, [gc_name, local_name]),
-      worker(Phoenix.PubSub.PG2Server, [server_name, local_name]),
+      supervisor(Phoenix.PubSub.LocalSupervisor, [server, pool_size, dispatch_rules]),
+      worker(Phoenix.PubSub.PG2Server, [server]),
     ]
 
-    supervise children, strategy: :rest_for_one
+    supervise children, strategy: :one_for_one
```
josevalim: We could keep it `:rest_for_one` now if we want. It would be the best most likely.

lib/phoenix/pubsub/local.ex
chrismccord: yes, I added this to avoid races in tests because it's only exposed for test, but you are right it's not needed in the server

lib/phoenix/pubsub/local_supervisor.ex
josevalim: Actually, you don't need the loop `:ets.insert` accepts many items I think (or there is a variant that does).

josevalim:
Really neat! :shipit:

lib/phoenix/pubsub/local.ex
josevalim: Ah, if you see my earlier commits, the one that introduced ETS, I have worked around this by sending a fake request to the server. Sometimes a subscribe, just to ensure it has caught up.

lib/phoenix/pubsub/local.ex
chrismccord: I don't like them either :)... but I only added them because we reference the named processes in tests a couple places, and I was constantly failing tests when changing names/shards. We can kill these and dup the alias construction in the tests?

lib/phoenix/pubsub/local.ex
```diff
@@ -129,39 +173,90 @@ defmodule Phoenix.PubSub.Local do
...
-    local_server
+  def list(pubsub_server, shard) when is_atom(pubsub_server) do
+    shard
+    |> local_for_shard(pubsub_server)
     |> :ets.select([{{:'$1', :_}, [], [:'$1']}])
     |> Enum.uniq
   end
 
   @doc false
-  # This is an expensive and private operation. DO NOT USE IT IN PROD.
```
josevalim: Should we add this back just to be double sure? :)

lib/phoenix/pubsub/local.ex
josevalim: I see. :shipit:

chrismccord:
Thanks @josevalim, all feedback addressed!

PR #1342, opened 2015-11-08
Default to preserving view assigns

bklang:
Without this, there can be confusion if the controller assigns views to the template, but they are not propagated back to the form partial


josevalim:
I am :-1: because we want to be explicit on purpose. :) Thank you for the PR though!

chrismccord:
Thanks for the idea! We've gone back and forth on this, and ultimately settled on the side of being explicit with what assigns we pass down as @josevalim said. Thanks!

PR #1343, opened 2015-11-08
test_helper.exs: get rid of extra blank line

henrik:
I have git hooks that warn me about extra blank lines at the end of files, and it reacted to this one.

I'm not sure if moving up the "end" is the nicest way. Would it be better to use [the trim mode](https://github.com/elixir-lang/elixir/pull/3194/files) for all templates? Seems a fairly new addition so maybe not backwards compatible.


chrismccord:
Thanks @henrik, but the trim mode is the way to go, but I think we need to wait until that Elixir version is more supported. Until then, the warning from your git hook is something you'll have to fix for new projects

henrik:
Completely fair. Let's wait for that. The git hook is easily workaroundable, just wanted to clean this up for the next person.

PR #1344, opened 2015-11-08
Phoenix.Channel docs: Add a note about not using splat when broadcasting

felixbr:
While building my first phoenix application I tried to broadcast via endpoint using a splat like `rooms:*`, which doesn't work.

I'm glad the community helped me out on this one, so I want to prevent others from making the same mistake.

Feel free to rephrase/reformat as you see fit.

### DIFF
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -150,6 +150,9 @@ defmodule Phoenix.Channel do
         redirect conn, to: "/"
       end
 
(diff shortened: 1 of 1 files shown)


chrismccord:
Thanks! I think the "Topics & Callbacks" section covers the definition of a "topic" well, being only an identifier, so I'm not convinced this is necessary. It's tricky with adding docs on what you *can't* do, because the possibilities are unlimited. If this starts coming up frequently, I will revisit. Thanks for helping to make Phoenix better :)

PR #1346, opened 2015-11-09
Improve default Postgres config

### LINKED ISSUE #1345: Improve default Postgres config
I was following the [Up and Running](http://www.phoenixframework.org/docs/up-and-running) guide and came across an error when I ran `mix ecto.create`:

    ** (Mix) The database for HelloPhoenix.Repo couldn't be created, reason given: psql: FATAL:  role "postgres" does not exist

So I looked up the generated *dev* config and saw this (similar for *test*):

```elixir
# Configure your database
config :hello_phoenix, HelloPhoenix.Repo,
  adapter: Ecto.Adapters.Postgres,
  username: "postgres",
  password: "postgres",
  database: "hello_phoenix_dev",
  hostname: "localhost",
  pool_size: 10
```

I then removed the `username`, the `password`, and the `hostname` and `mix ecto.create` worked.

I'd like to propose that these arguments are removed in postgres's case since they're not needed by default. The database name is enough. Rails default config for postgres doesn't specify these settings: https://github.com/rails/rails/blob/823d3a8de0ccfe2301e51ac1e06e106035edb8e6/railties/lib/rails/generators/rails/app/templates/config/databases/postgresql.yml#L17-L31

fphilipe:
By not specifying the host, username and password, `mix ecto.create` works out of the box after generating a new application.

This also reduces some duplication by extracting shared configs between sql databases into a `sql_base_config` function.

Closes #1345

@chrismccord Went ahead and took a shot at it. Hope that's OK!

### DIFF
--- a/installer/lib/phoenix_new.ex
+++ b/installer/lib/phoenix_new.ex
@@ -395,22 +395,22 @@ defmodule Mix.Tasks.Phoenix.New do
   end
 
   defp get_ecto_adapter("mssql", app, module) do
-    {:tds_ecto, Tds.Ecto, db_config(app, module, "db_user", "db_password")}
+    {:tds_ecto, Tds.Ecto, sql_db_config(app, module, "db_user", "db_password")}
   end
   defp get_ecto_adapter("mysql", app, module) do
-    {:mariaex, Ecto.Adapters.MySQL, db_config(app, module, "root", "")}
+    {:mariaex, Ecto.Adapters.MySQL, sql_db_config(app, module, "root", "")}
   end
   defp get_ecto_adapter("postgres", app, module) do
-    {:postgrex, Ecto.Adapters.Postgres, db_config(app, module, "postgres", "postgres")}
+    {:postgrex, Ecto.Adapters.Postgres,
+      [dev:  [database: "#{app}_dev"],
+       test: [database: "#{app}_test", pool: Ecto.Adapters.SQL.Sandbox],
+       prod: [database: "#{app}_prod"]] ++ sql_base_config(module)}
   end
   defp get_ecto_adapter("sqlite", app, module) do
     {:sqlite_ecto, Sqlite.Ecto,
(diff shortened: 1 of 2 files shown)


chvanikoff:
I think it is worth to keep username/password defaults for `prod` env.

tispratik:
+1.

fphilipe:
@chvanikoff Not sure why I removed it, but I guess I saw that MongoDB also didn't specify username/pass and thought that you'll be changing that anyways once you move to production. I could add it back. But I'd say it'd be nice to have a consistent base setting across all the DBs.

chvanikoff:
In case of care about consistency, I think we should go even deeper and set username/password to `nil` by default then.This way we will keep consistency across all DB adapters. But I'm not sure if `nil` would be properly handled at Ecto side

josevalim:
We don't care about consistency here. context > consistency.

chrismccord:
@fphilipe Sorry to give you the runaround on this, but I'm going to hold for now on not including username/password in configuration. Some postgres setups I've learned require username/password, and we have a very few people reporting issues with the defaults. I also think the explicit configuration makes it easier to change when the user/pass needs to be updated. Thanks!

fphilipe:
@chrismccord No worries.

> Some postgres setups I've learned require username/password

You mean in development mode, i.e. local postgres installation?

guido4000:
I had the problem too that prefilled values in the dev config broke mix ecto.create on OSX. At least this should be mentioned in the docs.

irkanu:
+1 @guido4000 As a beginner and first time installer for OSX - this is a boundary that is not easily solved by a beginner like myself. :)

novarac23:
I had the same problem on OSX when trying to set up my first phoenix project. 

I agree that it probably should stay the way it is tough, because some systems require explicit arguments for the db, but it should maybe be noted in the docs that this might happen and what to do if it happens.

I would be happy to do my first Phoenix open source contribution by updating docs :dancer: 

What do you peeps think ?

PR #1356, opened 2015-11-17
Remove channels

alex-shapiro:
Orlo can't use Phoenix Channels or PubSub because they're opinionated and don't match UMTP-C2S. Customize Phoenix by ripping them out and replacing them with raw WebSockets.


josevalim:
@alex-shapiro btw, you can just not use those features and directly use cowboy websockets. They are very easy to turn off in a regular application.

PR #1357, opened 2015-11-17
Fix router.ex docs

pedroassumpcao:
Remove query string output from url when no query string is passed to the helper function.


josevalim:

PR #1360, opened 2015-11-18
Drop inner in favor of explicit render

chrismccord:
@josevalim I still need to bump the docs, but I'm PR'ing now because I think I'll have to convince you this is the way to go :) There is some necessary coupling b/w the inner view and layout that the `conn` approach made murky. This PR introduces two new "magic" assigns, but it does it the clearest way, with the clearest coupling. It also avoids coupling the rendering to the `%Conn{}`, which may not always be present.

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -158,7 +158,10 @@ defmodule Phoenix.Controller do
   (or nil if no template was rendered).
   """
   @spec view_template(Plug.Conn.t) :: binary | nil
-  def view_template(conn), do: conn.private[:phoenix_template]
+  def view_template(conn) do
+    IO.puts :stderr, "[deprecation] view_template/1 has been deprecated in favor of the @view_template assign"
+    conn.private[:phoenix_template]
+  end
 
   defp get_json_encoder do
     Application.get_env(:phoenix, :format_encoders)
@@ -352,6 +355,7 @@ defmodule Phoenix.Controller do
   """
   @spec view_module(Plug.Conn.t) :: atom
   def view_module(conn) do
+    IO.puts :stderr, "[deprecation] view_module/1 has been deprecated in favor of the @view_module assign"
     conn.private.phoenix_view
   end
 
--- a/installer/templates/new/web/templates/layout/app.html.eex
+++ b/installer/templates/new/web/templates/layout/app.html.eex
@@ -26,7 +26,7 @@
       <p class="alert alert-danger" role="alert"><%%= get_flash(@conn, :error) %></p>
 
       <main role="main">
-        <%%= @inner %>
+        <%%= render @view_module, @view_template, assigns %>
       </main>
 
     </div> <!-- /container -->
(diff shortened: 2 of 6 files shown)


josevalim:
@chrismccord should we introduce those assigns or should we introduce this data to private and provide functions to access it? Also, regardless of which solution we choose, we can likely get rid of this one: https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/controller.ex#L156-L161 (at least deprecate it).

The only downside of assigns is that it ends up polluting the user assigns map, it is a weird precedent to set. For example, we should likely change the places where we inspect assigns in Phoenix.HTML and in the template not found error to ignore those two special assigns.

josevalim:
Ignore me, we cannot use private as the view layer does not expect a connection. So we need to do:

1. deprecate the view_template function
2. special case those attributes everytime we print assigns (I can recall in Phoenix.HTML and TemplateNotFound error or something of sorts)

chrismccord:
> Ignore me, we cannot use private as the view layer does not expect a connection.

yeah this was the issue when I trailed the `conn` approach and relying on functions like originally planned. It coupled the `conn` to rendering

> deprecate the view_template function

I didn't even recall us adding this, so :+1: 

> special case those attributes everytime we print assigns (I can recall in Phoenix.HTML and TemplateNotFound error or something of sorts)

If we go this direction, I think we *want* to print them when we print assigns, since the user would be using them when available (unless we hide it behind a function, but I'd like to avoid it)

josevalim:
Ok, so ship it. Also, since they are now part of the public API, don't forget to add specific tests and update the docs.

lib/phoenix/controller.ex
```diff
@@ -352,6 +355,7 @@ defmodule Phoenix.Controller do
   """
   @spec view_module(Plug.Conn.t) :: atom
   def view_module(conn) do
+    IO.puts :stderr, "[deprecation] view_module/1 has been deprecated in favor of the @view_module assign"
```
josevalim: Add `"\n" <> Exception.format_stacktrace` at the end so we know the caller.

lib/phoenix/controller.ex
```diff
@@ -158,7 +158,10 @@ defmodule Phoenix.Controller do
   (or nil if no template was rendered).
   """
   @spec view_template(Plug.Conn.t) :: binary | nil
-  def view_template(conn), do: conn.private[:phoenix_template]
+  def view_template(conn) do
+    IO.puts :stderr, "[deprecation] view_template/1 has been deprecated in favor of the @view_template assign"
```
josevalim: Add "\n" <> Exception.format_stacktrace at the end so we know the caller.

josevalim:
@chrismccord, I just realized we cannot raise for @inner in Phoenix.HTML because it does not depend on Phoenix, so it would be impossible to version. Instead, what I would do, is to store something in `@inner` that says: "The @inner assign in your layout is deprecated, please do render(...) instead" so something would be printed and users can act on it.

PR #1363, opened 2015-11-20
Fixes channel process not closing when websocket client times out

jlwebster:
Thoughts? Is this the correct fix? Was this actually a bug, or should the channel process remain running if the client times out? This was creating an issue for me where channel processes where not getting cleaned up.

Thanks!

### DIFF
--- a/lib/phoenix/transports/websocket.ex
+++ b/lib/phoenix/transports/websocket.ex
@@ -146,8 +146,8 @@ defmodule Phoenix.Transports.WebSocket do
   end
 
   @doc false
-  def ws_terminate(_reason, _state) do
-    :ok
+  def ws_terminate(_reason, state) do
+    {:shutdown, state}
   end
 
   @doc false
--- a/test/phoenix/integration/websocket_test.exs
+++ b/test/phoenix/integration/websocket_test.exs
@@ -45,6 +45,10 @@ defmodule Phoenix.Integration.WebSocketTest do
     def handle_in("boom", _message, _socket) do
       raise "boom"
     end
+    
+    def handle_in("timeout", _message, _socket) do
+      :timer.sleep(1000)
+    end
 
     def handle_out("new_msg", payload, socket) do
       push socket, "new_msg", Map.put(payload, "transport", inspect(socket.transport))
@@ -160,6 +164,21 @@ defmodule Phoenix.Integration.WebSocketTest do
     WebsocketClient.close(sock)
 
     assert_receive {:DOWN, _, :process, ^channel, {:shutdown, :closed}}
+  end
+  
+  test "channels are terminated if transport times out" do
+    {:ok, sock} = WebsocketClient.start_link(self, "ws://127.0.0.1:#{@port}/ws/websocket")
+
+    WebsocketClient.join(sock, "rooms:timeout", %{})
+    assert_receive %Message{event: "phx_reply", ref: "1", payload: %{"response" => %{}, "status" => "ok"}}
+    assert_receive %Message{event: "joined"}
+    channel = Process.whereis(:"rooms:timeout")
+    assert channel
+    Process.monitor(channel)
+    WebsocketClient.send_event(sock, "rooms:timeout", "timeout", %{})
+    :timer.sleep(200)
+
+    assert_receive {:DOWN, _, :process, ^channel, :shutdown}
   end
 
   test "refuses websocket events that haven't joined" do


chrismccord:
This would have no effect because we return `:ok` after calling the handle's ws_terminate, 
https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/endpoint/cowboy_websocket.ex#L92

I will double check cowboy's docs, but the handler should be closing once the client drops.

chrismccord:
and cowboy is calling the handler's callback when the connection terminates, so this shouldn't be the cause of your issue:
https://github.com/ninenines/cowboy/blob/1.1.x/src/cowboy_websocket.erl#L761

Can you describe a little more what you were seeing with connections hanging?

jlwebster:
Ok, yeah. I just running my test without my "fix" and it still passes. I swear the test was failing. :) I had added call to `ws_close(state)` to `ws_terminate` in the websocket transport to fix my issue, but that was causing another test to fail for some reason. 

What I'm seeing in real life is, even though I unplug the device holding the web socket connection, the terminate function is never called on my channel. After some digging, I noticed that `ws_terminate` was getting called after the web socket timeout, but the channel process was staying alive. 

I'll have to do some more testing and make sure I can reproduce on the latest master.

jlwebster:
So yeah, the issue is, cowboy is calling the handlers callback when the connection terminates, but nothing is being done in Phoenix (that I can see) to cleanup the channel processes in that particular case, so they stick around, at least from what I am seeing. I'll look into it more. Thanks.

chrismccord:
> What I'm seeing in real life is, even though I unplug the device holding the web socket connection, the terminate function is never called on my channel.

The terminate function in your channel is only ever called if you are trapping exits `Process.flag(:trap_exit, true)`. Trapping exits should be used only if absolutely necessary, so let's rule that out first, then we can talk about what you're trying to do in `terminate` that we can handle better elsewhere :)

jlwebster:
I see now. That was my issue. :D I was not trapping exits, but terminate was being called at times on the channel, probably due to some of the other conditions. Now terminate is getting called as expected and duh, because the transport process is closing, of course the child channel processes would close as well. Should have realized that before. :D

In any case, what I am trying to do in terminate is track the connection status of the device: setting a connected state to true on successful join and setting it to false on channel terminate. We only have one device per topic. I would love to find a better way to tell what devices are connected, if you have any ideas. Thanks for your help!

chrismccord:
The better approach would be to have your device tracker monitor or link to the channel process, then when it receives a DOWN or EXIT, you do whatever work is required. Here's a gist that will give you a good start. Keep in mind you'll need to do more than this if you have more than 1 node. Phoenix.Presence will solve all this for you, but it's a WIP right now:

```elixir
# lib/my_app.ex

children = [
  ...
  worker(ChannelWatcher, [:rooms])
]

# web/channels/room_channel.ex

def join("rooms:", <> id, params, socket) do
  uid = socket.assigns.user_id]
  :ok = ChannelWatcher.monitor(:rooms, self(), {__MODULE__, :on_leave, [id, uid]})

  {:ok, socket}
end

def on_leave(room_id, user_id) do
  MyApp.Endpoint.broadcast("rooms:", <> room_id, "user_left", %{user_id: user_id})
end

# lib/my_app/channel_watcher.ex

defmodule ChannelWatcher do
  use GenServer

  ## Client API

  def monitor(server_name, pid, mfa) do
    GenServer.call(server_name, {:monitor, pid, mfa})
  end

  def demonitor(server_name, pid) do
    GenServer.call(server_name, {:demonitor, pid})
  end

  ## Server API

  def start_link(name) do
    GenServer.start_link(__MODULE__, [], name: name)
  end

  def init(_) do
    Process.flag(:trap_exit, true)
    {:ok, %{channels: HashDict.new()}}
  end

  def handle_call({:monitor, pid, mfa}, _from, state) do
    Process.link(pid)
    {:reply, :ok, put_channel(state, pid, mfa)
  end

  def handle_call({:demonitor, pid}, _from, state) do
    case HashDict.fetch(state.channels, pid) do
      :error       -> {:reply, :ok, state}
      {:ok,  _mfa} ->
        Process.unlink(pid)
        {:reply, :ok, drop_channel(state, pid)}
    end
  end

  def handle_info({:EXIT, pid, _reason}, state) do
    case HashDict.fetch(state.channels, pid) do
      :error -> {:noreply, state}
      {:ok, {mod, func, args}} ->
        Task.start_link(fn -> apply(mod, func, args) end)
        {:noreply, drop_channel(state, pid)}
    end
  end

  defp drop_channel(state, pid) do
    %{state | channels: HashDict.delete(state.channels, pid)}}
  end

  defp put_channel(state, pid, mfa) do
    %{state | channels: HashDict.put(channels, pid, mfa)}}
  end
end
```

jlwebster:
Whoa, this is awesome and a way better approach! Seriously, thanks a ton for you help on this and your time. I totally thought I was helping by fixing a bug. Haha.

PR #1365, opened 2015-11-20
Bump brunch to 2.0.0

stevedomin:
I'm thinking I must have missed something obvious because everything just worked.

- [x] Brunch still listen to stdin and shuts down when we close Phoenix
- [x] Generated JS works as expected
- [x] Generated config works as expected
- [x] Live reloading of assets works as expected after a change


josevalim:
Looks good to me! Thanks @stevedomin!

chrismccord:
<3 <3 <3

PR #1368, opened 2015-11-23
Fix installer's user_socket code sample

stevedomin:
By default user_id will be an integer and integers are not valid
arguments to Kernel.<>


chrismccord:
<3 <3 <3

PR #1370, opened 2015-11-23
Update Router docs to make paths consistent

joshprice:
Applied back ticks and double quotes to paths in `router.ex` docs.


chrismccord:
Since it's a string it should be backtracked, as well as quoted

joshprice:
Fixed, is that what you were after?

chrismccord:
thanks!

PR #1374, opened 2015-11-26
Upgrade babel-brunch and commit its output.

realistschuckle:
I wanted to use the new babeljs pipeline with Phoenix. I got my [pull request for babel-brunch](https://github.com/babel/babel-brunch/pull/18) merged. Now, after updating the package.son, running `mix test` and integrating it into `https://github.com/chrismccord/phoenix_chat_example`, it all seems to work. Feedback and guidance appreciated, if needed.

### DIFF
--- a/brunch-config.js
+++ b/brunch-config.js
@@ -31,8 +31,7 @@ exports.config = {
   plugins: {
     babel: {
       // Do not use ES6 compiler in vendor code
-      ignore: [/^(web\/static\/vendor)/],
-      loose: "all"
+      ignore: [/^(web\/static\/vendor)/]
     }
   }
 };
--- a/package.json
+++ b/package.json
@@ -3,7 +3,7 @@
   },
   "dependencies": {
     "brunch": "^2.0.0",
-    "babel-brunch": "^5.1.1",
+    "babel-brunch": "^6.0.0",
     "clean-css-brunch": ">= 1.0 < 1.8",
     "css-brunch": ">= 1.0 < 1.8",
     "javascript-brunch": ">= 1.0 < 1.8",
--- a/priv/static/phoenix.js
+++ b/priv/static/phoenix.js
@@ -6,19 +6,19 @@
 
   var modules = {};
   var cache = {};
-  var has = ({}).hasOwnProperty;
-
   var aliases = {};
+  var has = ({}).hasOwnProperty;
 
   var endsWith = function(str, suffix) {
     return str.indexOf(suffix, str.length - suffix.length) !== -1;
   };
 
+  var _cmp = 'components/';
   var unalias = function(alias, loaderPath) {
     var start = 0;
     if (loaderPath) {
-      if (loaderPath.indexOf('components/' === 0)) {
-        start = 'components/'.length;
+      if (loaderPath.indexOf(_cmp) === 0) {
+        start = _cmp.length;
       }
       if (loaderPath.indexOf('/', start) > 0) {
         loaderPath = loaderPath.substring(start, loaderPath.indexOf('/', start));
@@ -26,33 +26,32 @@
     }
     var result = aliases[alias + '/index.js'] || aliases[loaderPath + '/deps/' + alias + '/index.js'];
     if (result) {
-      return 'components/' + result.substring(0, result.length - '.js'.length);
+      return _cmp + result.substring(0, result.length - '.js'.length);
     }
     return alias;
   };
 
-  var expand = (function() {
-    var reg = /^\.\.?(\/|$)/;
-    return function(root, name) {
-      var results = [], parts, part;
-      parts = (reg.test(name) ? root + '/' + name : name).split('/');
-      for (var i = 0, length = parts.length; i < length; i++) {
-        part = parts[i];
-        if (part === '..') {
-          results.pop();
-        } else if (part !== '.' && part !== '') {
-          results.push(part);
-        }
+  var _reg = /^\.\.?(\/|$)/;
+  var expand = function(root, name) {
+    var results = [], part;
+    var parts = (_reg.test(name) ? root + '/' + name : name).split('/');
+    for (var i = 0, length = parts.length; i < length; i++) {
+      part = parts[i];
+      if (part === '..') {
+        results.pop();
+      } else if (part !== '.' && part !== '') {
+        results.push(part);
       }
-      return results.join('/');
-    };
-  })();
+    }
+    return results.join('/');
+  };
+
   var dirname = function(path) {
     return path.split('/').slice(0, -1).join('/');
   };
 
   var localRequire = function(path) {
-    return function(name) {
+    return function expanded(name) {
       var absolute = expand(dirname(path), name);
       return globals.require(absolute, path);
     };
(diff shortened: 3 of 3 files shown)


brunch-config.js
```diff
@@ -31,8 +31,7 @@ exports.config = {
   plugins: {
     babel: {
       // Do not use ES6 compiler in vendor code
-      ignore: [/^(web\/static\/vendor)/],
-      loose: "all"
```
ericmj: What happened with this option in babel 6?

realistschuckle: The "loose" option should now be applied on a plugin-by-plugin basis for those plugins that support it as specified in the [6.0.0 release notes](https://github.com/babel/babel/blob/4c311ec2ab86544a23df866cb74feb25a859f0df/CHANGELOG.md#600).

As an example you can see from the [setup for the test for "loose"](https://github.com/babel/babel/blob/0df276272aff0e590176534c5c5179b839571c7a/packages/babel-plugin-transform-es2015-classes/test/fixtures/loose/options.json) in the class transformer, the options is pushed down further.

If it's important that we activate that flag, then I can change this PR to include each of the 2015 transformations with loose activated for the appropriate plugins:

* `babel-plugin-transform-es2015-for-of`
* `babel-plugin-transform-es2015-template-literals`
* `babel-plugin-transform-s2015-computed-properties`
* `babel-plugin-transform-es2015-classes`

This means that we will not use the `es2015` preset but, rather, list out each of the plugins used for  transpiling to the 2015 standard with the loose option activated.

The new section of the `brunch-config` would look like this:

```javascript
  babel: {
    presets: [],
    plugins: [
      ["babel-plugin-transform-es2015-template-literals", { loose: true }],
      "babel-plugin-transform-es2015-literals",
      "babel-plugin-transform-es2015-function-name",
      "babel-plugin-transform-es2015-arrow-functions",
      "babel-plugin-transform-es2015-block-scoped-functions",
      ["babel-plugin-transform-es2015-classes", { loose: true }],
      "babel-plugin-transform-es2015-object-super",
      "babel-plugin-transform-es2015-shorthand-properties",
      ["babel-plugin-transform-es2015-computed-properties", { loose: true }],
      ["babel-plugin-transform-es2015-for-of", { loose: true }],
      "babel-plugin-transform-es2015-sticky-regex",
      "babel-plugin-transform-es2015-unicode-regex",
      "babel-plugin-check-es2015-constants",
      "babel-plugin-transform-es2015-spread",
      "babel-plugin-transform-es2015-parameters",
      "babel-plugin-transform-es2015-destructuring",
      "babel-plugin-transform-es2015-block-scoping",
      "babel-plugin-transform-es2015-typeof-symbol",
      "babel-plugin-transform-es2015-modules-commonjs",
      ["babel-plugin-transform-regenerator", { async: false, asyncGenerators: false }],
    ],
    ignore: [/web\/static\/vendor/]
  }
  
```

chrismccord: :( Is all this really necessary. Do we need the loose option at all? This list seems like a guaranteed churn area for future updates

realistschuckle: @chrismccord That's why I just stuck with the default of `babel-brunch` and the `es2015` preset.

chrismccord: @ericmj do you remember why we added loose?

ericmj: It was added for IE8 and IE9 support iirc. If we remove it we should
definitely test IE again.

On Thursday, 26 November 2015, Chris McCord <notifications@github.com>
wrote:

> In brunch-config.js
> <https://github.com/phoenixframework/phoenix/pull/1374#discussion_r45997822>
> :
>
> > @@ -31,8 +31,7 @@ exports.config = {
> >    plugins: {
> >      babel: {
> >        // Do not use ES6 compiler in vendor code
> > -      ignore: [/^(web\/static\/vendor)/],
> > -      loose: "all"
>
> @ericmj <https://github.com/ericmj> do you remember why we added loose?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/phoenixframework/phoenix/pull/1374/files#r45997822>.
>


-- 
Eric Meadows-Jönsson

realistschuckle: @chrismccord Found the [commit](https://github.com/phoenixframework/phoenix/commit/41bfcf314ae3dd6a9fb7e1231ac0db7ba1def5e1) by @ericmj on 17 February 2015. The commit message includes the following:

> Generate phoenix.js with the “Loose Mode” option on es6to5 which in conjunction with es5-shim makes the javascript work on IE8.

I can jumpstart a VM, this weekend, to test it on IE8, if that browser is still a concern for the development team.

josevalim: We could also say we don't  support IE8 and people can chase down this solution if they want it. @ericmj and others would be fine as long as he doesn't upgrade his own babel stuff.

bryansray: @josevalim that seems like a good direction to me. Possibly documenting the old solution somewhere for those who want the IE8 support?

michalmuskala: [Microsoft is soon dropping the support for old IEs](https://www.microsoft.com/en-us/WindowsForBusiness/End-of-IE-support), so it may be one of the reasons to drop it here too

chrismccord: 👍🏻

ericmj: I am against removing support for IE8-10 but if you do remove it you should also remove the `XDomainRequest` code.

josevalim: We need to know all versions that loose is meant to support. Also, is it a
binary thing? Or we update Babel or we keep IE8 support? I know we could
list plugins one by one... But that's a no go for me.

On Thursday, November 26, 2015, Eric Meadows-Jönsson <
notifications@github.com> wrote:

> In brunch-config.js
> <https://github.com/phoenixframework/phoenix/pull/1374#discussion_r46008564>
> :
>
> > @@ -31,8 +31,7 @@ exports.config = {
> >    plugins: {
> >      babel: {
> >        // Do not use ES6 compiler in vendor code
> > -      ignore: [/^(web\/static\/vendor)/],
> > -      loose: "all"
>
> I am against removing support for IE8-10 but if you do remove it you
> should also remove the XDomainRequest code.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/phoenixframework/phoenix/pull/1374/files#r46008564>.
>


-- 


ericmj: IIRC loose mode is to support IE8, XDomainRequest is to support IE8-9.

josevalim: So let's confirm that. If it is only IE8, I am all in for dropping it. We
then keep the X domain stuff to support IE9 and people can simply not
upgrade it.

On Thursday, November 26, 2015, Eric Meadows-Jönsson <
notifications@github.com> wrote:

> In brunch-config.js
> <https://github.com/phoenixframework/phoenix/pull/1374#discussion_r46008867>
> :
>
> > @@ -31,8 +31,7 @@ exports.config = {
> >    plugins: {
> >      babel: {
> >        // Do not use ES6 compiler in vendor code
> > -      ignore: [/^(web\/static\/vendor)/],
> > -      loose: "all"
>
> IIRC loose mode is to support IE8, XDomainRequest is to support IE8-9.
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/phoenixframework/phoenix/pull/1374/files#r46008867>.
>


-- 


ericmj: What are we gaining from upgrading to babel 6? If there is no clear benefit maybe we should wait with the update until there are benefits.

bryansray: http://babeljs.io/blog/2015/10/29/6.0.0/

Pretty big chunk of "benefits." I guess the question for me is, Do those benefits out weight IE 8 support? For me, the answer would be "yes." But I'm also not concerned with Internet Explorer.

chrismccord: > What are we gaining from upgrading to babel 6? If there is no clear benefit maybe we should wait with the update until there are benefits.

I am generally in agreement here, but staying behind the curve on anything in the node ecosystem is a problematic. Is it right to say that we could upgrade babel as proposed, keep xdomain, for IE9/10, and that folks that still need IE8 can include that list of plugin options? If so it seem like everyone's happy? If you're stuck supporting ie8, copying 15 lines of config is probably the easiest job you'll have that day :)

bryansray: > If you're stuck supporting ie8, copying 15 lines of config is probably the easiest job you'll have that day :)

:+1:

realistschuckle: @ericmj I don't want the burden of my changes to fall on anyone else. How can I test this so we all feel comfortable with the change? Is there a test-bed project that I can use that exercises `phoenix.js` and the excellent work that you did to get it well-tested in IE8?

chrismccord:
<3 <3 <3

PR #1375, opened 2015-11-26
Support for instrumentation in Phoenix

whatyouhide:
Let me know what you think, I'm pretty happy with how this is already coming out. :D :tada:


josevalim:
Great job @whatyouhide! I think we can improve this first iteration by doing more and less work at compile time at the same time.

In particular, the generation of clauses can be done at compile time but all the instrumentation work at runtime. Here is some pseudo code:

```elixir
defmacro instrument(event, runtime \\ [], fun) do
  compile = compile_opts(__CALLER__)
  quote do
    Phoenix.Endpoint.instrument(unquote(event), unquote(compile), unquote(runtime), unquote(fun))
  end
end

# Now for each event in any of the modules, we will generate
def instrument(:render_view, compile, runtime, fun) do
  res0 = Mod0.render_view(:start, compile, runtime)
  res1 = Mod1.render_view(:start, compile, runtime)
  res2 = Mod2.render_view(:start, compile, runtime)

  start = current_time()
  try do
    fun.()
  after
    stop = current_time()
    diff = time_diff(start, stop)
    Mod0.render_view(:stop, diff, res0)
    Mod1.render_view(:stop, diff, res0)
    Mod2.render_view(:stop, diff, res0)
  end
end

# catch all - nobody is interested in those events
def instrument(_event, _compile, _runtime, fun) do
  fun.()
end

if erlang monotonic time do
  defp current_time do
    ...
  end

  defp time_diff do
    ...
  end
else
  ...
end
```

josevalim:
Also, let's move the code to `Phoenix.Endpoint.Instrument`. It is preferable to call `Phoenix.Endpoint.Instrument.definstrument(app)` directly from `Phoenix.Endpoint` (it could be during before before_compile hook) than to `use Phoenix.Instrument`. We rely a lot on this pattern on Phoenix (check the endpoint and the router) as it makes it explicit and we don't need to chase before_compile using hooks through different files.

whatyouhide:
So the cost of wrapping the code inside a `fn` is near to zero? Because with the way you proposed, when we nobody is interested in an event, then we still call `fun.()`; with my version, if nobody is interested in an event, absolutely nothing happens as far as instrumentation goes.

I have to say this version looks cleaner though, so it may be worth it :) (and I actually remember you saying on Skype that just calling `fun.()` as almost no cost, not sure though).

josevalim:
It is very low cost, we don't need to worry about it. So yes, let's just
ask them to pass functions.

-- 


whatyouhide:
@josevalim ready for review, I implemented all the changes you mentioned. I had to rely on some public (but with `@doc false`) functions in `Phoenix.Endpoint.Instrument` in order to clean up `__using__` a little bit.

Also, do we need to enforce that `runtime` is a map? I think it could be just any term, e.g., with `:render_view` it could just be `"index.html"`. Right now, this is what I implemented but changing it back to  an enforced map would take just a couple of minutes.

Let me know what you think!

chrismccord:
This looks great @whatyouhide. Would it make any sense to map the results as a list, then map again to report the results instead of building the intermediate vars? Not sure if there are perf implications, but it might make the code a tad simpler. Either way the existing code very very nice

josevalim:
> @josevalim ready for review, I implemented all the changes you mentioned. I had to rely on some public (but with @doc false) functions in Phoenix.Endpoint.Instrument in order to clean up __using__ a little bit.

Feel free to have as many functions in `Phoenix.Endpoint.Instrument`, you can even `@doc` them as long as `@moduledoc` false. The whole purpose of that module to exist is to define named functions. :)

lib/phoenix/endpoint/instrumenter.ex
```diff
@@ -0,0 +1,138 @@
+defmodule Phoenix.Endpoint.Instrument do
```
josevalim: `@moduledoc false`

whatyouhide:
> Would it make any sense to map the results as a list, then map again to report the results instead of building the intermediate vars? Not sure if there are perf implications, but it might make the code a tad simpler.

@chrismccord I think we settled on not looping over lists but expanding all the necessary stuff at compile time. Yes mapping over the list of calls to get results and looping again on `:stop` may be cleaner, but I think this way is faster?

@josevalim right, I'll make the module "private" and also call `definstrument` on the endpoint.

lib/phoenix/endpoint/instrumenter.ex
```diff
@@ -0,0 +1,138 @@
...
+  @doc false
+  defmacro __using__(_opts) do
+    quote unquote: false do
+      # The `otp_app` variable is set by Phoenix.Endpoint.
+      # TODO: maybe it's better to make Phoenix endpoints export __otp_app__()
+      # or something similar?
+      config        = Application.get_env(var!(otp_app), __MODULE__, [])
+      instrumenters = config[:instrumentation] || []
```
lexmag: Maybe better to use `Keyword.get/3` to stop it from silently working with `instrumentation: nil`?

lib/phoenix/endpoint/instrumenter.ex
```diff
@@ -0,0 +1,138 @@
...
+  end
+
+  # Given an instrumenter module, this function returns the result variable to
+  # use for that instrumenter module (the result variable is just the result
+  # returned by the start callback of an event).
+  defp build_result_variable(inst) do
+    inst
+    |> Atom.to_string()
+    |> String.replace(".", "_")
+    |> (fn s -> "res_" <> s end).()
```
lexmag: :see_no_evil:

lib/phoenix/endpoint/instrumenter.ex
```diff
@@ -0,0 +1,138 @@
+defmodule Phoenix.Endpoint.Instrument do
+  @doc false
+  defmacro __using__(_opts) do
+    quote unquote: false do
+      # The `otp_app` variable is set by Phoenix.Endpoint.
+      # TODO: maybe it's better to make Phoenix endpoints export __otp_app__()
+      # or something similar?
+      config        = Application.get_env(var!(otp_app), __MODULE__, [])
```
josevalim: You can move all this code inside `events_to_instrumenters`. Maybe call it `app_instrumenters`.

lib/phoenix/endpoint/instrumenter.ex
```diff
@@ -0,0 +1,138 @@
...
+
+      ## Examples
+
+          instrument :render_view, [view: "index.html"], fn ->
+            render conn, "index.html"
+          end
+
+      """
+      defmacro instrument(event, runtime \\ nil, fun) do
+        compile = Macro.escape(__CALLER__)
```
josevalim: We don't need the whole __CALLER__. It may in fact be too much. Literally this: https://github.com/elixir-lang/elixir/blob/master/lib/logger/lib/logger.ex#L533-L538

(we can even copy the compile time app thing).

lib/phoenix/endpoint/instrumenter.ex
```diff
@@ -0,0 +1,138 @@
...
+            fun.()
+          after
+            var!(diff) = time_diff(start, current_time())
+            unquote(Phoenix.Endpoint.Instrument.compile_stop_callbacks(event, instrumenters))
+          end
+        end
+      end
+
+      # Catch-all clause
+      def instrument(_event, _compile, _runtime, fun) do
```
josevalim: Let's add guards here too!

lib/phoenix/endpoint/instrumenter.ex
```diff
@@ -0,0 +1,138 @@
...
+  # modules interested in that event.
+  def events_to_instrumenters(instrumenters) do
+    instrumenters                                # [Inst1, Inst2, ...]
+    |> instrumenters_and_events()                # [{Inst1, ev1}, {Inst1, ev2}, {Inst2, ev1}, ...]
+    |> Enum.group_by(fn {_inst, evt} -> evt end) # %{ev1 => [Inst1, Inst2], ev2 => ...}
+    |> Enum.map(fn {evt, insts} -> {evt, Enum.map(insts, fn {inst, _evt} -> inst end)} end)
+  end
+
+  defp instrumenters_and_events(instrumenters) do
+    for inst <- instrumenters, {event, _arity} <- inst.__info__(:functions) do
```
josevalim: Let's just copy the ones with `arity == 3`

lib/phoenix/endpoint/instrumenter.ex
```diff
@@ -0,0 +1,138 @@
...
+        unquote(inst).unquote(event)(:stop, var!(diff), unquote(build_result_variable(inst)))
+      end
+    end
+  end
+
+  # Given an instrumenter module, this function returns the result variable to
+  # use for that instrumenter module (the result variable is just the result
+  # returned by the start callback of an event).
+  defp build_result_variable(inst) do
+    inst
```
josevalim: Variables could be simpler. Just receive an integer based on the number of instrumenters and name them res_0, res_1, res_2 and so on. You can use `Enum.with_index` to tie the instrumenters to an index.

lib/phoenix/endpoint/instrumenter.ex
```diff
@@ -0,0 +1,138 @@
...
+  end
+
+  @doc false
+  # Takes a list of instrumenter modules and returns a list of `{event,
+  # instrumenters}` tuples where each tuple represents an event and all the
+  # modules interested in that event.
+  def events_to_instrumenters(instrumenters) do
+    instrumenters                                # [Inst1, Inst2, ...]
+    |> instrumenters_and_events()                # [{Inst1, ev1}, {Inst1, ev2}, {Inst2, ev1}, ...]
+    |> Enum.group_by(fn {_inst, evt} -> evt end) # %{ev1 => [Inst1, Inst2], ev2 => ...}
```
josevalim: I think the comment is wrong. it will become `%{ev1 => [{Inst1, ev}, ...]}`

whatyouhide: Whoops, yes I mixed that line and the next line up :)

lib/phoenix/endpoint/instrumenter.ex
josevalim: Let's keep the current pattern. Most of the times we don't care about the distinction between those and this code snippet is more concise.

lib/phoenix/endpoint/instrumenter.ex
josevalim: I just knew because I was bitten by something similar a couple day ago :P

whatyouhide:
@josevalim @chrismccord @lexmag this is ready for review.

I addressed all the comments @josevalim mentioned (stripped the `__CALLER__`, only took events with arity of `3`, used `Instrument.definstrument(var!(otp_app))` in the endpoint, called result variables `res0` ).

I think the code that resulted here is pretty clean, but we may be able to clean it up even further with something like:

```elixir
defmodule Phoenix.Endpoint.Instrument do
  defmacro definstrument(app) do
    quote bind_quoted: [app: app] do
      Phoenix.Endpoint.Instrument.def_instrument_macro()
      Phoenix.Endpoint.Instrument.def_instrument_function(app, __MODULE__)
    end
  end
end
```

but while cleaning up `definstrument/1`, I think this makes the rest slightly harder to follow. Your choice anyways, it shouldn't take long to make these changes :)

lib/phoenix/endpoint/instrument.ex
```diff
@@ -0,0 +1,199 @@
...
+        defp time_diff(start, stop), do: stop - start
+      else
+        defp current_time, do: :os.timestamp()
+        defp time_diff(start, stop), do: :timer.now_diff(stop, start)
+      end
+    end
+  end
+
+  @doc """
+  Reads a list of the instrumnters from the config of `otp_app` and finds all
```
Gazler: instrumenters

lib/phoenix/endpoint/instrument.ex
```diff
@@ -0,0 +1,199 @@
...
+
+    unless is_list(instrumenters) and Enum.all?(instrumenters, &is_atom/1) do
+      raise ":instrumentation must be a list of instrumenter modules"
+    end
+
+    events_to_instrumenters(instrumenters)
+  end
+
+  @doc """
+  Strips a `Macro.Env` struct, leaving only intereseting compile-time metadata.
```
Gazler: interesting not intereseting

Gazler:
@whatyouhide I just found a couple of typos in the documentation. Looks great apart from that.

lib/phoenix/endpoint.ex
```diff
@@ -226,6 +226,9 @@ defmodule Phoenix.Endpoint do
       unquote(pubsub())
       unquote(plug())
       unquote(server())
+
+      require Phoenix.Endpoint.Instrument
+      Phoenix.Endpoint.Instrument.definstrument(var!(otp_app))
```
josevalim: I was thinking more about something like this (inside `__before_compile__`):

```elixir
otp_app = Module.get_attribute(env.module, :otp_app)
instrumentation = Phoenix.Endpoint.Instrument.definstrument(otp_app, env.module)

quote do
  unquote(instrumentation)
end
```

There is no reason to use a macro in this particular case. :)

whatyouhide:
@josevalim aaah, ok I see, makes a lot of sense :) Pushed 77a6873 with the new code!
 to

lib/phoenix/endpoint/instrument.ex
```diff
@@ -0,0 +1,187 @@
...
+      end
+
+      # Catch-all clause
+      def instrument(_event, compile, _runtime, fun)
+          when is_map(compile) and is_function(fun, 0) do
+        fun.()
+      end
+
+      # Let's define the current_time/0 and time_diff/2 functions based on the
+      # existence of :erlang.monotonic_time/1.
```
josevalim: Let's leave a todo: `# TODO: Remove this once Phoenix supports only Elixir 1.2`.

lib/phoenix/endpoint/instrument.ex
```diff
@@ -0,0 +1,187 @@
...
+
+  # This is the arity that event callbacks in the instrumenter modules must
+  # have.
+  @event_callback_arity 3
+
+  def definstrument(otp_app, endpoint) do
+    app_instrumenters = app_instrumenters(otp_app, endpoint)
+
+    quote bind_quoted: [app_instrumenters: app_instrumenters] do
+      @doc """
```
josevalim: Those docs are excellent! We should also include a snippet about instrumentation covering all aspects, from configuring up to defining your own module up to instrument in a new section in `Phoenix.Endpoint`.

josevalim:
It looks perfect, we just need a bit more of docs and we are ready to go! :D

whatyouhide:
@josevalim ok, I added docs in `Phoenix.Endpoint`. Let me know what you think!

josevalim:
Ship it! :D Now we need to start making use of this!

whatyouhide:
pz96rxa

(@josevalim I can't ship it :))

josevalim:
@whatyouhide I was going to merge then I checked the build and I think you have a heisentest: https://travis-ci.org/phoenixframework/phoenix/builds/93650805#L377

@chrismccord you need to get on top of those other heisentests :)

whatyouhide:
@josevalim ok, possibly fixed the heisentest :)

chrismccord:
I thought we fixed our Heisentests with @gazler's patch?

lib/phoenix/endpoint.ex
```diff
@@ -77,6 +77,10 @@ defmodule Phoenix.Endpoint do
 
       The default format is used when none is set in the connection.
 
+    * `:instrumentation` - a list of instrumenters modules whose callbacks will
```
josevalim: Shouldn't we call it `instrumenters`? Given the description, that seems to be what makes most sense.

whatyouhide: @josevalim makes sense, pushed a fix for that.

josevalim:
@chrismccord see: https://travis-ci.org/phoenixframework/phoenix/builds/93650805#L358

PR #1377, opened 2015-11-28
Remove deprecated usage of behaviours

rodrigues:


josevalim:
We can't because it means Phoenix would no longer be compatible with Elixir v1.0. We need to wait a bit more and probably start the move by the time Elixir v1.2 is out.

rodrigues:
ah, ok 😅 closing it then

PR #1378, opened 2015-11-29
Add bypass feature to ConnTest

chrismccord:


chrismccord:
@josevalim I cheated by exposing a public `__pipeline__` function on the router, but this was the least complex option. Thoughts?

lib/phoenix/router.ex
```diff
@@ -293,8 +293,15 @@ defmodule Phoenix.Router do
       end
 
     quote do
-      defp do_call(unquote(conn), opts) do
-        unquote(call)
+      defp do_call(unquote(conn) = conn, opts) do
+        case conn.private[:phoenix_bypass] do
```
josevalim: Please pattern match this on the function head, it is going to be much, much, much faster.

lib/phoenix/router.ex
```diff
@@ -375,6 +382,7 @@ defmodule Phoenix.Router do
               Plug.Conn.WrapperError.reraise(unquote(conn), kind, reason)
           end
         end
+        def __pipeline__(conn, unquote(plug)), do: unquote(plug)(conn, [])
```
josevalim: Shouldn't we rather just make the pipeline function public? It doesn't hurt... :)

lib/phoenix/router.ex
```diff
@@ -293,8 +293,15 @@ defmodule Phoenix.Router do
       end
 
     quote do
-      defp do_call(unquote(conn), opts) do
-        unquote(call)
+      defp do_call(unquote(conn) = conn, opts) do
+        case conn.private[:phoenix_bypass] do
+          nil -> unquote(call)
+          {__MODULE__, pipeline} ->
```
josevalim: Here is what I would do: I would always store it as `{__MODULE__, pipelines}`. This way, we get to support both 0 and  many pipelines, all we need to do is to reduce over the connection:

    Plug.Conn.halt Enum.reduce pipelines, conn, &apply(__MODULE__, &1, [&2, []])

In the test helper, the default pipeline is an empty list, but we will also call `List.wrap` on the given argument so we can convert a single pipeline into a list.

lib/phoenix/test/conn_test.ex
```diff
@@ -446,4 +446,76 @@ defmodule Phoenix.ConnTest do
...
+  Or only inoke the Endpoint's plugs:
+
+    conn =
+      conn
+      |> bypass()
+      |> MyApp.RequireAuthentication.call([])
+    assert conn.halted
+  """
+  @spec bypass_pipeline(Conn.t, Module.t) :: Conn.t
+  def bypass_pipeline(conn, endpoint),
```
josevalim: We have agreed we won't call the endpoint. You should pipe into `get/2` or whatever you want to yourself because get already has the convenience for recycling, setting http verb, params and so on. Ideally, those are the only signatures we need:

```elixir
defmacro bypass_pipeline(conn, router, pipelines \\ [])
def bypass_pipeline(conn, endpoint, router, pipelines)
```

Also, which name are we going with? `bypass`? `bypass_pipeline`? `bypass_through`?

josevalim: We have agreed we won't call the endpoint. You should pipe into `get/2` or whatever you want to yourself because those helpers already has the convenience for recycling, setting http verb, params and so on.

Ideally, this is the only signature that we need:

```elixir
def bypass_pipeline(conn, router, pipelines \\ [])
```

Also, which name are we going with? `bypass`? `bypass_pipeline`? `bypass_through`?

lib/phoenix/router.ex
chrismccord: good point, :+1:

lib/phoenix/test/conn_test.ex
chrismccord: You are right! I was thinking about it too narrowly. :+1: `bypass_through` is maybe the best, but you are not always bypassing through a pipeline, you have the option to bypass the Router only, or just call the Endpoint without the router. SO `ConnTest.bypass/3` with the `bypass` macro gets my vote. I'm open to alternatives though

josevalim: If we make the pipelines argument required, then `bypass_through` works. You just pass an empty list if you don't care.

chrismccord:
@josevalim ready for review

josevalim:
Call the router function __bypass__ so it is not imported into the user
module and that's it :)


-- 


josevalim:
One of the examples is also outdated. It is using bypass instead of
bypass_through.

-- 


chrismccord:
Thanks for the feedback @josevalim !

josevalim:
Don't forget to bring it to master!

PR #1379, opened 2015-11-29
Various instrumentations

whatyouhide:
This PR adds a bunch of stuff:

- `instrument/3-4` macros to `Phoenix.Endpoint`: we need this because we often don't have the endpoint at compile time, so we can't call `endpoint.instrument`
- instrumentation for controller dispatching
- instrumentation for controller rendering

(this is what @josevalim suggested for now)

This works but I don't think it's very pretty. I can't think of a better way to do it now, but I feel we need generic instrumentation very strongly :( I would love to do

```elixir
config :phoenix, instrumenters: [...]

# and also, for my custom instrumentations
config :my_app, MyApp.Endpoint, instrumenters: [...]

# and, why not
config :postgrex, instrumenters: [...]

# and so on
```

Just to write this down somewhere where it won't get buried (instead of IRC), the problem with this is that if I change the `:instrumenter` config for the `:phoenix` app (which is a dependency of `:my_app`), then phoenix will (obviously) not be recompiled. :( I think this problems extends to all dependencies that would want to use configuration at compile time, so maybe finding a solution could benefit more things other than instrumentation.

Still, let's see how this goes!

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -618,6 +618,8 @@ defmodule Phoenix.Controller do
   end
 
   defp do_render(conn, template, format, assigns) do
+    import Phoenix.Endpoint, only: [instrument: 3, instrument: 4]
+
     assigns = to_map(assigns)
     content_type = Plug.MIME.type(format)
     conn =
@@ -627,8 +629,17 @@ defmodule Phoenix.Controller do
 
     view = Map.get(conn.private, :phoenix_view) ||
             raise "a view module was not specified, set one with put_view/2"
-    data = Phoenix.View.render_to_iodata(view, template,
-                                         Map.put(conn.assigns, :conn, conn))
+
+    data =
+      if endpoint = conn.private[:phoenix_endpoint] do
+        instrument endpoint, :phoenix_controller_render, template <> format, fn ->
+          Phoenix.View.render_to_iodata(view, template,
+                                        Map.put(conn.assigns, :conn, conn))
+        end
+      else
+        Phoenix.View.render_to_iodata(view, template, Map.put(conn.assigns, :conn, conn))
+      end
+
     send_resp(conn, conn.status || 200, content_type, data)
   end
 
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -3,6 +3,7 @@ defmodule Phoenix.Controller do
   alias Plug.Conn.AlreadySentError
 
   require Logger
+  require Phoenix.Endpoint
 
   @unsent [:unset, :set]
 
@@ -618,8 +619,6 @@ defmodule Phoenix.Controller do
   end
 
   defp do_render(conn, template, format, assigns) do
-    import Phoenix.Endpoint, only: [instrument: 3, instrument: 4]
-
     assigns = to_map(assigns)
     content_type = Plug.MIME.type(format)
     conn =
@@ -630,7 +629,7 @@ defmodule Phoenix.Controller do
     view = Map.get(conn.private, :phoenix_view) ||
             raise "a view module was not specified, set one with put_view/2"
 
-    data = instrument conn, :phoenix_controller_render, template <> format, fn ->
+    data = Phoenix.Endpoint.instrument conn, :phoenix_controller_render, template <> format, fn ->
       Phoenix.View.render_to_iodata(view, template, Map.put(conn.assigns, :conn, conn))
     end
 
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -630,15 +630,9 @@ defmodule Phoenix.Controller do
     view = Map.get(conn.private, :phoenix_view) ||
             raise "a view module was not specified, set one with put_view/2"
 
-    data =
-      if endpoint = conn.private[:phoenix_endpoint] do
-        instrument endpoint, :phoenix_controller_render, template <> format, fn ->
-          Phoenix.View.render_to_iodata(view, template,
-                                        Map.put(conn.assigns, :conn, conn))
-        end
-      else
-        Phoenix.View.render_to_iodata(view, template, Map.put(conn.assigns, :conn, conn))
-      end
+    data = instrument conn, :phoenix_controller_render, template <> format, fn ->
+      Phoenix.View.render_to_iodata(view, template, Map.put(conn.assigns, :conn, conn))
+    end
 
     send_resp(conn, conn.status || 200, content_type, data)
   end
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -629,7 +629,8 @@ defmodule Phoenix.Controller do
     view = Map.get(conn.private, :phoenix_view) ||
             raise "a view module was not specified, set one with put_view/2"
 
-    data = Phoenix.Endpoint.instrument conn, :phoenix_controller_render, template <> format, fn ->
+    runtime_data = %{template: template, format: format}
+    data = Phoenix.Endpoint.instrument conn, :phoenix_controller_render, runtime_data, fn ->
       Phoenix.View.render_to_iodata(view, template, Map.put(conn.assigns, :conn, conn))
     end
 
--- a/lib/phoenix/controller/pipeline.ex
+++ b/lib/phoenix/controller/pipeline.ex
@@ -103,11 +103,7 @@ defmodule Phoenix.Controller.Pipeline do
                  &(&1 |> Map.put(:phoenix_controller, __MODULE__)
                       |> Map.put(:phoenix_action, action))
 
-        if endpoint = conn.private[:phoenix_endpoint] do
-          Phoenix.Endpoint.instrument endpoint, :phoenix_controller_call, fn ->
-            phoenix_controller_pipeline(conn, action)
-          end
-        else
+        Phoenix.Endpoint.instrument conn, :phoenix_controller_call, fn ->
           phoenix_controller_pipeline(conn, action)
         end
       end
--- a/lib/phoenix/controller/pipeline.ex
+++ b/lib/phoenix/controller/pipeline.ex
@@ -88,6 +88,8 @@ defmodule Phoenix.Controller.Pipeline do
     quote do
       @behaviour Plug
 
+      require Phoenix.Endpoint
+
       import Phoenix.Controller.Pipeline
       Module.register_attribute(__MODULE__, :plugs, accumulate: true)
       @before_compile Phoenix.Controller.Pipeline
@@ -100,7 +102,14 @@ defmodule Phoenix.Controller.Pipeline do
         conn = update_in conn.private,
                  &(&1 |> Map.put(:phoenix_controller, __MODULE__)
                       |> Map.put(:phoenix_action, action))
-        phoenix_controller_pipeline(conn, action)
+
+        if endpoint = conn.private[:phoenix_endpoint] do
+          Phoenix.Endpoint.instrument endpoint, :phoenix_controller_call, fn ->
+            phoenix_controller_pipeline(conn, action)
+          end
+        else
+          phoenix_controller_pipeline(conn, action)
+        end
       end
 
       def action(%{private: %{phoenix_action: action}} = conn, _options) do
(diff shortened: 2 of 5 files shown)


lib/phoenix/endpoint.ex
```diff
@@ -533,6 +533,24 @@ defmodule Phoenix.Endpoint do
     end
   end
 
+  @doc false
+  # This is private because it's really just a "hack" to instrument with an
```
josevalim: We can make this public. It is alright. We document it is useful when you want to dynamically dispatch to an instrumentation endpoint.

lib/phoenix/controller/pipeline.ex
```diff
@@ -100,7 +102,14 @@ defmodule Phoenix.Controller.Pipeline do
         conn = update_in conn.private,
                  &(&1 |> Map.put(:phoenix_controller, __MODULE__)
                       |> Map.put(:phoenix_action, action))
-        phoenix_controller_pipeline(conn, action)
+
+        if endpoint = conn.private[:phoenix_endpoint] do
+          Phoenix.Endpoint.instrument endpoint, :phoenix_controller_dispatch, fn ->
```
josevalim: :phoenix_controller_dispatch -> :phoenix_controller_call?

whatyouhide:
@josevalim addressed your comments!

josevalim:
I don't think we can make multiple instrumenters in other projects work. However, we could have a single instrumenter that you call throughout your project, similar to Logger.

We could have an instruction in "mix.exs" that says to "mix deps.clean instrumenter" every time you change the configuration option, that should be enough to avoid pitfalls and it isn't something people would invoke all the time anyway:

    # If you add or remote an instrumenter, run: mix deps.clean phoenix
    config :phoenix, instrumenters: [Foo, Bar, Baz]

We need to choose if this should belong to Phoenix or a separate project. FWIW, I don't envision Ecto using it for now.

whatyouhide:
Off to a brighter future with `Phoenix.Instrument`, closing this in the meantime :)

whatyouhide:
@josevalim pushed the `try`/`catch` stuff, ready for review.

lib/phoenix/controller.ex
```diff
@@ -627,8 +629,17 @@ defmodule Phoenix.Controller do
 
     view = Map.get(conn.private, :phoenix_view) ||
             raise "a view module was not specified, set one with put_view/2"
-    data = Phoenix.View.render_to_iodata(view, template,
-                                         Map.put(conn.assigns, :conn, conn))
+
+    data =
+      if endpoint = conn.private[:phoenix_endpoint] do
+        instrument endpoint, :phoenix_controller_render, template <> format, fn ->
```
chrismccord: I would make `instrument`, `%Plug.Conn{}` aware. That way we can match on the conn, and if it has a phoenix_endpoint we do our thing, otherwise we just execute the block. 

```elixir
    data =
      instrument conn, :phoenix_controller_render, template <> format, fn ->
        Phoenix.View.render_to_iodata(view, template,
                                      Map.put(conn.assigns, :conn, conn))
      end
```

chrismccord:
I don't like that this config is global:

    # If you add or remote an instrumenter, run: mix deps.clean phoenix
    config :phoenix, instrumenters: [Foo, Bar, Baz]

I would almost prefer to skip instrumentation if we don't have the context of the `conn` or Endpoint, which is on the rarer side. It would only come up with you render without the context of either, for sending emails, or sub renders.

lib/phoenix/controller.ex
whatyouhide: @chrismccord so you mean making `Phoenix.Endpoint.instrument/3` have two clauses: one for a regular endpoint and one for a `Plug.Conn`, and decide what do to based on that?

whatyouhide:
@chrismccord we dumped the idea of a global Phoenix instrumenter and went back to what we have now, which is:

```elixir
config :my_app, MyApp.Endpoint, instrumenters: [...]
```

Personally, I'd avoid tying instrumentation to the `conn` because we would private users of instrumenting whatever they want with `MyApp.Endpoint.instrument` :\

lib/phoenix/controller.ex
chrismccord: @whatyouhide exactly. It cleans up this code and the view code a bunch. If we have a conn without and endpoint, it just execute the block.

whatyouhide: That we can do. Wait a couple minutes :)

chrismccord:
> Personally, I'd avoid tying instrumentation to the conn because we would private users of instrumenting whatever they want with MyApp.Endpoint.instrument

If the function lives on the endpoint itself, it's naturally `conn` aware anyway, since the Endpoint is a plug. So I don't see an issue with it. The catch-all clause will still behave as it did before

test/phoenix/endpoint/instrument_test.exs
```diff
@@ -95,4 +99,21 @@ defmodule Phoenix.Endpoint.InstrumentTest do
...
+        send self(), :ok
+      end
+    end
+
+    assert_receive :ok
+    # We have the correct stacktrace.
+    assert log =~ Path.relative_to_cwd(__ENV__.file)
+    # And we're sure the exception is logged twice:
+    err = "** (RuntimeError) oops"
+    assert Regex.scan(~r/#{Regex.escape(err)}/, log) == [[err], [err]]
```
lexmag: Maybe to be uber-pedantic we could use `^` (beginning of the line) and `$` (end of the line) in the regex? :bowtie:

whatyouhide: I thought about it and was lazy. I knew I couldn't escape @lexmag :smile: :heart_decoration: will do!

whatyouhide:
@chrismccord updated!

lib/phoenix/endpoint.ex
```diff
@@ -533,6 +533,35 @@ defmodule Phoenix.Endpoint do
...
+      endpoint = MyApp.Endpoint
+      Phoenix.Endpoint.instrument endpoint, :render_view, fn -> ... end
+
+  """
+  defmacro instrument(endpoint_or_conn, event, runtime \\ nil, fun) do
+    compile = Phoenix.Endpoint.Instrument.strip_caller(__CALLER__) |> Macro.escape()
+
+    quote do
+      case unquote(endpoint_or_conn) do
+        %Plug.Conn{private: %{phoenix_endpoint: endpoint}} ->
```
josevalim: This brings other questions like: should we support sockets as well? I dunno.

lib/phoenix/endpoint/instrument.ex
```diff
@@ -138,7 +140,11 @@ defmodule Phoenix.Endpoint.Instrument do
     Enum.map Enum.with_index(instrumenters), fn {inst, index} ->
       quote do
         unquote(build_result_variable(index)) =
-          unquote(inst).unquote(event)(:start, var!(compile), var!(runtime))
+          try do
+            unquote(inst).unquote(event)(:start, var!(compile), var!(runtime))
+          catch
+            kind, error -> Logger.error Exception.format(kind, error)
```
josevalim: Let's add a bit of meta text here: `"Instrumenter #{instrumenter}.#{event}/3 failed.\n" <> Exception.format`. Keep in mind the string part can be generated at compile time.

lib/phoenix/controller.ex
```diff
@@ -627,8 +629,11 @@ defmodule Phoenix.Controller do
 
     view = Map.get(conn.private, :phoenix_view) ||
             raise "a view module was not specified, set one with put_view/2"
-    data = Phoenix.View.render_to_iodata(view, template,
-                                         Map.put(conn.assigns, :conn, conn))
+
+    data = instrument conn, :phoenix_controller_render, template <> format, fn ->
```
josevalim: Let's make the runtime a map and enforce it. This will help us avoid running into exactly the mistake we did here. What if we want to add more information in the future? Because we are passing `template <> format`, it would be a breaking change. If we use a map, you can always add new keys.

lib/phoenix/controller.ex
```diff
@@ -618,6 +618,8 @@ defmodule Phoenix.Controller do
   end
 
   defp do_render(conn, template, format, assigns) do
+    import Phoenix.Endpoint, only: [instrument: 3, instrument: 4]
```
josevalim: Why are we importing both functions? Is it even worth importing if it is a single call?

lib/phoenix/controller/pipeline.ex
```diff
@@ -88,6 +88,8 @@ defmodule Phoenix.Controller.Pipeline do
     quote do
       @behaviour Plug
 
+      require Phoenix.Endpoint
```
josevalim: Please remove the require to outside the `quote` block.

lib/phoenix/endpoint.ex
josevalim: At least, because the socket uses the same structure, we can just match on `%{__struct__: struct, private: ...} when struct in [Plug.Conn, Phoenix...]`

josevalim: So i guess it is fine to add socket support.

test/phoenix/endpoint/instrument_test.exs
```diff
@@ -95,4 +99,21 @@ defmodule Phoenix.Endpoint.InstrumentTest do
     assert_receive {__MODULE__.MyInstrumenter, :common_event_stop}
     assert_receive {__MODULE__.MyOtherInstrumenter, :common_event_stop}
   end
+
```
josevalim: We need a test for the endpoint redirect macro this commit is adding. :)

lib/phoenix/endpoint.ex
chrismccord: :+1:

lib/phoenix/controller/pipeline.ex
whatyouhide: It doesn't work if I move the `require` outside of the quote :\

josevalim: Waaaaaaaaaat. It should work just fine. I will investigate it later

whatyouhide:
@josevalim I should have addressed all your comments.

lib/phoenix/endpoint/instrument.ex
```diff
@@ -72,11 +74,16 @@ defmodule Phoenix.Endpoint.Instrument do
       end
 
       # Catch-all clause
-      def instrument(_event, compile, _runtime, fun)
-          when is_map(compile) and is_function(fun, 0) do
+      def instrument(_event, compile, runtime, fun)
+          when is_map(compile) and is_map(runtime) and is_function(fun, 0) do
```
josevalim: Assert `event` is an atom. :)

lib/phoenix/endpoint/instrument.ex
```diff
@@ -72,11 +74,16 @@ defmodule Phoenix.Endpoint.Instrument do
...
         fun.()
       end
 
+      # Bad args clauses
+      def instrument(_event, _compile, runtime, _fun) do
```
josevalim: That's not necessarily true, it can be invoked because compile for some reasons is not a map, fun is not a function, etc. Just yank this clause. :)

test/phoenix/endpoint/instrument_test.exs
```diff
@@ -95,4 +99,51 @@ defmodule Phoenix.Endpoint.InstrumentTest do
...
+    assert_receive :ok
+    # We have the correct stacktrace.
+    assert log =~ Path.relative_to_cwd(__ENV__.file)
+    assert log =~ "Instrumenter #{inspect __MODULE__.MyInstrumenter}.raising_event/3 failed.\n"
+    # And we're sure the exception is logged twice:
+    err = "** (RuntimeError) oops"
+    assert Regex.scan(~r/#{Regex.escape(err)}$/m, log) == [[err], [err]]
+  end
+
+  test "runtime data must be a map" do
```
josevalim: You can yank this test too. :)

test/phoenix/endpoint/instrument_test.exs
```diff
@@ -95,4 +99,51 @@ defmodule Phoenix.Endpoint.InstrumentTest do
...
+    import Endpoint
+
+    assert_raise ArgumentError, "runtime metadata must be a map, was: :not_a_map", fn ->
+      instrument :my_event, :not_a_map, fn -> :ok end
+    end
+  end
+
+  test "Phoenix.Endpoint.instrument/4 macro proxies to the endpoint instrument function" do
```
josevalim: Beautiful. :)

jeregrine:
:tada:

PR #1392, opened 2015-12-06
Alphabetical order in phoenix.new package.json

jimbeaudoin:
Otherwise when you add a new dependency with `npm install --save`, brunch is alphabetically reordered in the list. This can cause confusion in a git commit.

### DIFF
--- a/installer/templates/static/brunch/package.json
+++ b/installer/templates/static/brunch/package.json
@@ -2,8 +2,8 @@
   "repository": {
   },
   "dependencies": {
-    "brunch": "^2.0.0",
     "babel-brunch": "^5.1.1",
+    "brunch": "^2.0.0",
     "clean-css-brunch": ">= 1.0 < 1.8",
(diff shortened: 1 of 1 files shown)


chrismccord:
Is this a node convention? The "start your package.json as alphabetical"? It seems `npm install --save` will always reorder, no?

jimbeaudoin:
Yes `--save` will always reorder. But when I start a new project, if I add a new dependency, brunch will be reordered and it will appear in my git commit and then I can have a comment from a coworker like you changed brunch? Anyway, I thought it was a good thing to have the default list already in the correct order when you create a new project.

doomspork:
+1 for this.   Since `npm install --save` reorders and we typically alphabetize our dependencies anyways, we end up with this change in each Phoenix project.

tispratik:
+1. Agree with the change.

chrismccord:
Understood. Thanks!

Issue #1396, opened 2015-12-08
improve generated path helpers for glob routes


manukall:
For routes like `get "/app*jsroute", AppController, :index` the path helper has to be called with a list as argument: `app_path(@conn, :index, ["appsomething"])`. This is not very intuitive. As @chrismccord adviced on slack, I'm creating this ticket to discuss this.

josevalim:
@manukall yes, it expects the path segments. We could allow the whole path to be given but folks need to be keep in mind we won't be able to escape it. For example, if you write: "/users/:username" and pass a username of "foo/bar", we will escape "/". We do the same for "*jsroute" but only because you are giving a list. If we support strings, it would be appended as is.

manukall:
@josevalim would it make sense to add a short note about that to the router docs?

josevalim:
Sure, better docs are always welcome!

On Wednesday, December 9, 2015, Manuel Kallenbach <notifications@github.com>
wrote:

> @josevalim <https://github.com/josevalim> would it make sense to add a
> short note about that to the router docs?
>
> —
> Reply to this email directly or view it on GitHub
> <https://github.com/phoenixframework/phoenix/issues/1396#issuecomment-163207668>
> .
>


-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

PR #1398, opened 2015-12-09
example for a path helper for glob-like patterns

manukall:
It wasn't obvious to me how to use path helpers for routes with glob-like patterns. This adds a short paragraph with an example about that to the router documentation.

see  #1396

### DIFF
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -64,6 +64,12 @@ defmodule Phoenix.Router do
       MyApp.Router.Helpers.page_url(conn_or_endpoint, :show, "hello", some: "query")
       "http://example.com/pages/hello?some=query"
 
+  If the route contains glob-like patterns, parameters for those have to be given as
+  list:
+
+      MyApp.Router.Helpers.dynamic_path(conn_or_endpoint, :show, ["dynamic", "something"])
+      "/dynamic/something"
+
   The url generated in the named url helpers is based on the configuration for
   `:url`, `:http` and `:https`.
 


Gazler:
Thanks for this.

Maybe we should document that it can be called like:

      MyApp.Router.Helpers.dynamic_path(conn_or_endpoint, :show, ["some", "route"])

lib/phoenix/router.ex
```diff
@@ -64,6 +64,12 @@ defmodule Phoenix.Router do
       MyApp.Router.Helpers.page_url(conn_or_endpoint, :show, "hello", some: "query")
       "http://example.com/pages/hello?some=query"
 
+  If the route contains glob-like patterns, parameters for those have to be given as
+  list and will not be escaped:
+
+      MyApp.Router.Helpers.dynamic_path(conn_or_endpoint, :show, ["/some/route"])
```
josevalim: Oh, those should be escaped. If they are not escaped, it is a bug. The whole point of receiving a list is so we escape them. :)

manukall:
@Gazler actually the whole example was wrong, as the segment list has to match the path from the route definition. fixed that and also included/kept the other segment.

@josevalim what i'm trying to say is that the list elements will not be escaped, so `some/route` will not turn into `some%2Froute`. i also tried some other values which didn't get escaped, though.

josevalim:
@manukall what I am saying is that if it behaves like you documented, it is a bug. :)

manukall:
@josevalim i did some more testing on this. routes like `get "/dynamic*anything", DynamicController, :show` will not be escaped, while routes like `get "/dynamic/*anything", DynamicController, :show` (with a `/` before the `*`) will be.

if that's a bug, i would like to try fixing it...

josevalim:
Please give it a try. I think both should escape. :)

josevalim:
Ok, now we can update and merge this too. :) Thank you @manukall!

manukall:
updated and rebased. ok like that?

MaartenG:
There's a typo in the dynamic_path example, where you have "dymanic" instead of "dynamic". The line below it has it correctly.

manukall:
@MaartenG i swear i read that 1000 times. fixed now. thanks :-)

PR #1401, opened 2015-12-10
feat(Phoenix.Router.Helpers): better errors on path helpers

Gazler:
Instead of simply returning an error when passing an atom to a route
helper that hasn't been defined, the error now prompts the user to input
a defined route instead.

Previously an `UndefinedFunctionError` was raised.

Now an `ArgumentError` is raised with the following message:

    No route helper clause for MyApp.Router.Helpers.foo_path/2 defined
    for action :skip.
    The following foo_path/2 actions are defined under your router:

     * :file
     * :show

The only thing that is missing that I can think of is returning the error for functions which have an `arity` higher than 3.

For example:

user_channel_post_path(conn, :show, user, channel, post, foo: "bar")

**TODO**

 * [x] Add a catch all function for each helper
 * [x] Sort the valid actions alphabetically.
 * [x] Handle an arity > 3
 * [x] Show a different error for a valid action with an invalid arity

### DIFF
--- a/lib/phoenix/router/helpers.ex
+++ b/lib/phoenix/router/helpers.ex
@@ -109,6 +109,13 @@ defmodule Phoenix.Router.Helpers do
   def define(env, routes) do
     ast = for {route, exprs} <- routes, do: defhelper(route, exprs)
 
+    catch_all =
+      routes
+      |> Enum.filter(fn {route, _exprs} ->
+        (not is_nil(route.helper) and not (route.kind == :forward)) end)
+      |> Enum.group_by(fn {route, _exprs} -> route.helper end)
+      |> Enum.map(&defhelper_catch_all/1)
+
     # It is in general bad practice to generate large chunks of code
     # inside quoted expressions. However, we can get away with this
     # here for two reasons:
@@ -123,6 +130,8 @@ defmodule Phoenix.Router.Helpers do
       Module with named helpers generated from #{inspect unquote(env.module)}.
       """
       unquote(ast)
+
+      unquote(catch_all)
 
       @doc """
       Generates the connection/endpoint base URL without any path information.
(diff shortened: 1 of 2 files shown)


lib/phoenix/router/helpers.ex
```diff
@@ -231,6 +240,36 @@ defmodule Phoenix.Router.Helpers do
...
+
+  @doc false
+  def raise_route_error(module, helper, action, valid_routes) do
+    module_without_prefix = module |> to_string() |> String.split(".", parts: 2) |> Enum.at(1)
+    valid_actions = Enum.join(["" | valid_routes], "\n * :")
+
+    message = "no function clause matching in #{module_without_prefix}.#{helper}\n\n" <>
+              "This is likely due to the use of :#{action} as the function name.\n\n" <>
+              "Only the following atoms can be used:\n" <>
+              valid_actions
```
chrismccord: How about

```
No route helper clause for Router.Helpers.foo_path defined for action :show.
The following `foo_path` actions are defined under your router:
    ...
```

lib/phoenix/router/helpers.ex
```diff
@@ -231,6 +240,36 @@ defmodule Phoenix.Router.Helpers do
...
+    raise ArgumentError,
+    message: String.strip(message)
```
chrismccord: can single line this

lib/phoenix/router/helpers.ex
Gazler: @chrismccord I prefer the wording. I think displaying the arity of the function is important though.

chrismccord: yes arity is definitely important, so tweak as desired. Along these lines, it depends how far we want to go with this, i.e. if I have defined a nested route, but not provided all the route segments, I'll still get a function clause error with this PR. Should we also account for those cases?

test/phoenix/router/helpers_test.exs
```diff
@@ -221,6 +244,51 @@ defmodule Phoenix.Router.HelpersTest do
...
+   The following user_comment_path/3 actions are defined under your router:
+
+     * :create
+     * :delete
+     * :edit
+     * :index
+     * :new
+     * :show
+     * :update
+    """ |> String.strip
```
chrismccord: Since we have all the routes, we could even print the `ConsoleFormatter` version of the routes for only this particular helper, that way the user gets something like:

```
No route helper clause for MyApp.Router.Helpers.user_post_path/3 defined
for action :show with an arity of 3.
Please ensure you are passing all dynamic segments of the path as arguments.
The following user_post_path/3 actions are defined under your router:

   user_post_path  GET     /users/:user_id/post/:id       MyApp.PostController :index
   user_post_path  GET     /users/:user_id/posts/new      MyApp.PostController :new
   user_post_path  GET     /users/:user_id/post/:id       MyApp.PostController :show
   user_post_path  POST    /users/:user_id/posts          MyApp.PostController :create
```

Gazler: This is a neat idea. I'll try it out when I get some time.

Gazler:
@chrismccord I've implemented it using the console formatter. The easiest way to get this working was passing an optional filter to the format function. Another (less clean in my opinion) option is to perform the filter before calling `format` and then pass a map with an `__routes__` key to the formatter.

Any feedback is welcome.

test/phoenix/router/helpers_test.exs
```diff
@@ -154,9 +154,32 @@ defmodule Phoenix.Router.HelpersTest do
     assert Helpers.top_path(__MODULE__, :top, id: 5) == "/posts/top?id=5"
     assert Helpers.top_path(__MODULE__, :top, %{"id" => 5}) == "/posts/top?id=5"
 
-    assert_raise UndefinedFunctionError, fn ->
+    error_message = (fn helper ->
+    """
+    No route helper clause for Phoenix.Router.HelpersTest.Router.Helpers.#{helper} defined for action :skip.
+    The following #{helper} actions are defined under your router:
+
+     * :file
```
josevalim: two spaces?

josevalim: before the `*`

lib/phoenix/router/helpers.ex
```diff
@@ -231,6 +240,59 @@ defmodule Phoenix.Router.Helpers do
...
+      end
+    end
+  end
+
+  @doc false
+  def raise_route_error({m, f, a}, router, helper, action, valid_routes) do
+    module_without_prefix = m |> to_string() |> String.split(".", parts: 2) |> Enum.at(1)
+    message = case action in valid_routes do
+      true ->
+        "No route helper clause for #{module_without_prefix}.#{f}/#{a} " <>
```
josevalim: What if we skip the module name? Most users won't care. Plus it will likely be in the stacktrace. Just keep fun + arity.

chrismccord:
@Gazler Let's rollback to the original approach of only defining catch-all clauses for the "correct" arities and writing the available actions. @josevalim had a great point on irc that I hadn't considered: by catching the wrong arities we are turning a compile time error into a runtime error, which we don't want to do. Let's rollback to before I sent you on that goose chase and then this is ready. Thanks!

Gazler:
@chrismccord @josevalim Great point about the compile time errors.

I'll need to figure out a way of determining the correct arities for each helper and defining a catch-all for it.

Gazler:
I've changed to use two spaces for the actions list and removed the module name as @josevalim suggested.

Only the correct arities have a catch-all function defined now.

lib/phoenix/router/helpers.ex
```diff
@@ -231,6 +240,55 @@ defmodule Phoenix.Router.Helpers do
...
 
+  def defhelper_catch_all({helper, routes_and_exprs}) do
+    valid_routes = Enum.map(routes_and_exprs, fn {routes, _exrs} -> routes.opts end)
+    route_vars =
+      Enum.map(routes_and_exprs, fn {_routes, exprs} -> :lists.unzip(exprs.binding) end)
+      |> Enum.uniq
+
+    for {_, vars} <- route_vars do
+      arity = Enum.count(vars) + 2
+      vars = Enum.map(vars, fn {field, meta, val} -> {String.to_atom("_#{field}"), meta, val} end)
```
josevalim: We can just generate `{:_, [], nil}` I believe.

lib/phoenix/router/helpers.ex
```diff
@@ -231,6 +240,55 @@ defmodule Phoenix.Router.Helpers do
     end
   end
 
+  def defhelper_catch_all({helper, routes_and_exprs}) do
+    valid_routes = Enum.map(routes_and_exprs, fn {routes, _exrs} -> routes.opts end)
+    route_vars =
+      Enum.map(routes_and_exprs, fn {_routes, exprs} -> :lists.unzip(exprs.binding) end)
```
josevalim: Pipe all the way?

    routes_and_exprs
    |> Enum.map
    |> ...

Gazler:
@josevalim @chrismccord I have squashed all the commits and made the following changes:

 * Pipe all the way for https://github.com/Gazler/phoenix/commit/1034848a2a6724720b591b4f421535bcd309d2a9#diff-d4559021068841a3295eabb4bbcb699dR246
 * Generate {:_, [], nil} for https://github.com/Gazler/phoenix/commit/1034848a2a6724720b591b4f421535bcd309d2a9#diff-d4559021068841a3295eabb4bbcb699dR250

I have also changed the message to be:

 >     No route helper clause for foo_path/2 defined for action :skip. The following foo_path actions are defined under your router

There is no mention of the arity the 2nd time as we can't be certain that the arities for a particular action will be the same as passed in by the user.

chrismccord:
<3 <3 <3

PR #1402, opened 2015-12-11
Add redirect to: :back.

tuvistavie:
I added the option to redirect to the previous page instead of a particular page.
This can be useful for example in an `update` action, when the `edit` page is accessible from
multiple pages.
I followed the Rails behaviour of `:back` which is to raise when no `Referer` header is present.

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -296,7 +296,7 @@ defmodule Phoenix.Controller do
 
   """
   def redirect(conn, opts) when is_list(opts) do
-    url  = url(opts)
+    url  = url(conn, opts)
     html = Plug.HTML.html_escape(url)
     body = "<html><body>You are being <a href=\"#{html}\">redirected</a>.</body></html>"
 
@@ -305,12 +305,13 @@ defmodule Phoenix.Controller do
     |> send_resp(conn.status || 302, "text/html", body)
   end
 
-  defp url(opts) do
+  defp url(conn, opts) do
     cond do
       to = opts[:to] ->
         case to do
           "//" <> _ -> raise_invalid_url()
           "/" <> _  -> to
+          :back     -> referer(conn)
           _         -> raise_invalid_url()
         end
       external = opts[:external] ->
@@ -321,6 +322,12 @@ defmodule Phoenix.Controller do
   end
   defp raise_invalid_url do
     raise ArgumentError, "the :to option in redirect expects a path"
+  end
+  defp referer(conn) do
+    case get_req_header(conn, "referer") do
+      [referer] -> referer
+      _         -> raise ArgumentError, "cannot redirect back without a referer"
+    end
   end
 
   @doc """
--- a/test/phoenix/controller/controller_test.exs
+++ b/test/phoenix/controller/controller_test.exs
@@ -241,6 +241,20 @@ defmodule Phoenix.Controller.ControllerTest do
     end
   end
 
+  test "redirect/2 with to: :back" do
+    conn = conn(:get, "/") |> put_req_header("referer", "/foobar")
+    conn = redirect(conn, to: :back)
+    assert conn.resp_body =~ "/foobar"
+    assert get_resp_content_type(conn) == "text/html"
+    assert get_resp_header(conn, "location") == ["/foobar"]
+    refute conn.halted
+
+    assert_raise ArgumentError, ~r/cannot redirect back without a referer/, fn ->
+      redirect(conn(:get, "/"), to: :back)
+    end
+  end
+
+
   test "redirect/2 with :external" do
     conn = redirect(conn(:get, "/"), external: "http://example.com")
     assert conn.resp_body =~ "http://example.com"


josevalim:
I have mixed feelings. `redirect to: :back` always read weird and it is almost a bad practice in Rails because you eventually always end-up triggering the error bit. So I'd rather right: `redirect_to request.referer || "/other/page"`.

The other issue with referer is that it can redirect to an external page and it may not be obvious nor desired. For all of those reasons, I would rather support a `referer(conn)` function that returns the referer and let the user explicitly decide how it wants to handle it. In other words:

    redirect to: referer(conn) || "/"

tuvistavie:
@josevalim Thanks for your comment. I agree that `redirect to: :back` reads weird.
If exposing a `referer` function, maybe it would be nice to allow the user choose to return any URL, or to return `nil` in case the URL is external.

This would allow to write something like this:

```
redirect to: referer(conn, false) || "/"
redirect external: referer(conn, true) || "/"
```

where we can be sure that the first case will never fail, nor redirect to an external site.
What do you think?

josevalim:
I don't like the examples above because from looking at the code I have no idea what `false` or  `true` would mean. Can we replace those by atoms? How to implement so without looking redundant?

josevalim:
Also, is it guaranteed by browsers that the referer for the current site will always start with a path "/" and external with the whole protocol?

tuvistavie:
>  Can we replace those by atoms? How to implement so without looking redundant?

Allowing a keyword as the second argument would probably read better and be more explicit.

```
referer(conn) # can return anything
referer(conn, ignore_external: true) # returns nil if the request is external
```

> Also, is it guaranteed by browsers that the referer for the current site will always start with a path "/" and external with the whole protocol?

AFAIK, the `Referer` headers always contains a full URL (e.g. http://example.com/path),
so I think the best way to do this is to check if the `Host` domain and the `Referer` domain match, then strip the `scheme://host` part of the `Referer` to get the path that starts with `/`.
I am going to check the HTTP protocol to see if there is not edge-case.

josevalim:
@tuvistavie that complicates things then because you can have multiple domains so it is hard to check semantically which one you want to allow and which one you do not. So I would rather just go with `redirect conn, external: referer(conn) || "/"`

josevalim:
I would also love to hear @chrismccord feedback on this.

tuvistavie:
@josevalim the browser default behavior for subdomains is to treat them as different domains (for cookies, cross domain requests, etc), so I think that handling subdomains as external addresses would be acceptable. Furthermore, the only way to redirect to a subdomain now is to use `external:`, so it would be consistent with the current behavior.

After checking HTTP specs, it seems that `Referer` can be either relative or absolute, and the `Host` header is mandatory so there should not be any edge case as far as I can think of.

Another point, if we were to add a `referer` function, should it be part of `Phoenix.Controller`?
After thinking of it a little more, I thought that `Plug.Conn` could be more appropriate as the function would not depend on any Phoenix feature but only Plug ones.

josevalim:
> After checking HTTP specs, it seems that Referer can be either relative or absolute, and the Host header is mandatory so there should not be any edge case as far as I can think of.

My concern is rather things like myaccount.example.com and main.example.com.

chrismccord:
Thanks for the PR! I also share @josevalim's concerns and agree the only thing I would be for doing is:

    redirect conn, external: referer(conn) || "/"

The issue as @tuvistavie points out is the referrer can be either a path or full url, so you have to know or check to use `:external`. I feel like this would cause the same issues as were laid out in rails earlier, so I'm unsure if it is worth adding. 

> Another point, if we were to add a referer function, should it be part of Phoenix.Controller

yup

tuvistavie:
@josevalim @chrismccord Thanks for the feedback.

> The issue as @tuvistavie points out is the referrer can be either a path or full url,

Yes, if `referer` returns the header as is,

```
redirect to: referer(conn) || ""
```

would not be usable, but I cannot think of a good way to handle this though.

An idea could be to have `referer` return the relative uri when the referer is absolute and
its host matches the `Host` header, but I think it is a little counterintuitive.

Do you think we should stick with

```
redirect external: referer(conn) || ""
```

and leave the potential check for an external referer to the user?

josevalim:
After thinking a little bit more about this, I am wondering if it makes more sense to use the flash or the session to store the path you want to redirect to. This gives you more control (i.e. you would only store the page you are currently on) and don't have any of the drawbacks we are discussing. I.e. I think we should drop this because it is not a good pattern to promote anyway.

tuvistavie:
@josevalim this seems to be a much nicer way to deal with it. I need this functionality in the project I am currently working on, so I am going to try handing this with the session, as you suggested.
I think this PR can be closed. Thank you both of you for your time.

maxim:
This makes a ton of sense. Now we can create a plug tracking the url in flash, and have `redirect(to: :back)` working in terms of that plug. How would you feel about having this much cleaner implementation in phoenix?

tuvistavie:
For the record, I ended up using something like this:

```elixir
defmodule Opencov.Plug.Navigation do
  import Plug.Conn, only: [put_session: 3, get_session: 2]

  @session_key :internal_previous_path

  def init(opts) do
    Keyword.put_new(opts, :excluded_paths, [])
  end

  def call(conn, opts) do
    if conn.method == "GET" and not (conn.request_path in opts[:excluded_paths]) do
      put_previous_path(conn, conn.request_path)
    else
      conn
    end
  end

  def previous_path(conn, opts \\ []) do
    get_session(conn, @session_key) || opts[:default]
  end

  defp put_previous_path(conn, path) do
    put_session(conn, @session_key, path)
  end
end
```

And simply calling `redirect to: Navigation.previous_path(conn, default: "/")` where I needed to redirect back.

paulcsmith:
@tuvistavie Thanks for sharing! That is a really cool way to do it. I love how it solves those weird edge cases where the referer is external or not a GET request. Going to try this out in one of my own apps soon. Thanks for posting.

One thing to consider doing is importing a function in controllers like `redirect_back(conn, default: "path")` so it's a bit easier to use/read. 

```elixir
def redirect_back(conn, default: default_path) do
  Phoenix.Controller.redirect to: Navigation.previous_path(conn, default: default_path)
end
```

Totally a preference thing, but might be a way to make it even nicer.

tuvistavie:
FWIW, as I was starting to use this in multiple project, I improved a little the plug and moved it to a small library with proper tests and documentation, which I just published to hex.
https://github.com/tuvistavie/plug-navigation-history
As it does not depend on Phoenix but only on plug, I did not add the `redirect_back` helper in it, but I did mention it in the README.

augnustin:
i'll go for your lib @danhper thanks!

But I reckon I don't know where to add the 

```
def redirect_back(conn, opts \\ []) do
  Phoenix.Controller.redirect(conn, to: NavigationHistory.last_path(conn, opts))
end
```

in my project, being a Phoenix n00b...

Could you mention it here, and even better, within the doc of you repo?

Cheers

augnustin:
I have succeeded doing the following:

```
# web/controllers/helpers.ex

defmodule MyApp.Controllers.Helpers do
  def redirect_back(conn, opts \\ []) do
    Phoenix.Controller.redirect(conn, to: NavigationHistory.last_path(conn, opts))
  end
end
```

```
# web.ex

defmodule Vae.Web do
  def controller do
    quote do
      import Vae.Controllers.Helpers, only: [redirect_back: 2]
    end
  end
end
```

Cheers

Issue #1314, opened 2015-10-28
Phoenix reports wrong port during startup on port 0


Kavec:
When configuring Phoenix to run Cowboy via port 0, Cowboy correctly receives an ephemeral port from the OS and starts listening on it, but Phoenix reports it's running on port 0 instead of the ephemeral port the http server is listening for new connections on. This requires using command-line tools like `lsof` to hunt down the actual port number.

The error occurs because [CowboyHandler.info/3](https://github.com/phoenixframework/phoenix/blob/6d0dbb3/lib/phoenix/endpoint/cowboy_handler.ex#L100-L104) pulls the port number directly from the configuration (which usually works as expected, port 0 is a bit of an edge case). 

*Something* needs to be asked what port Cowboy is listening on; I've only had a week and a half's worth of Elixir buddying and haven't quite made my way far enough through learning the Phoenix/Cowboy/Elixir/Erlang/OTP ecosystem to make the least intrusive change and submit a PR. :/

chrismccord:
I'd like to see if there's something we can do here that isn't couple'd directly to cowboy. I don't think there's anything in plug that exposes this at the moment. This is up for grabs if someone wants to take a look

druzn3k:
@chrismccord I could try, but I cannot guarantee success (I'm still newbie). If the premise it's ok for you, I will take a look. :+1:

dm1try:
seems [`:ranch.get_port/1` can help](https://github.com/ninenines/cowboy/blob/27ecd7fca1c9111a09beb42ffe7efea0c317fd45/test/cowboy_test.erl#L24)

dm1try:
just checked it:
```elixir
if config[:port] == 0 do
  ref = hd(a)
  ephemeral_port = :ranch.get_port(ref)
  "#{info(scheme, endpoint, config)} (ephemeral port #{ephemeral_port})"
else
  info(scheme, endpoint, config)
end
```
[info] Running Testapp.Endpoint with Cowboy on http://localhost:0 (ephemeral port 45994)
```

chrismccord:
Awesome. We need to expose this at the plug level so it can be adapter independent. Thanks for doing the leg work!

dm1try:
Sure, I didn't know about such thing like an ephemeral port until now, so the "work" was informative :)

I haven't enough time for the implementation, so if anyone want to help - just do it!

scrogson:
I'll take a look into getting this into Plug.

[...]

PR #1404, opened 2015-12-13
Properly handle ephemeral ports in endpoint config

Kavec:
Wound up implementing a fix for #1314. Since @chrismccord wanted it exposed at the plug level, I'll be sending a pull request over that way momentarily. In order to keep the change low-impact, though, I also implemented the direct call to Ranch if the new functionality I'm sending to Plug is unavailable.

Overall, this is a deeper change than the cosmetic logging I originally was complaining about, so please let me know if/where further changes are needed and I'll be glad to make them. In particular, I'm unsure if modifying the integration tests like I have is the best policy-- finding where the tests should actually go was the most difficult part here-- and there may be some knock-on wartiness from coaxing the Endpoint.url call to return the active listening port.

Thanks a ton to @dm1try for doing the initial legwork and pointing us in the right direction!

### DIFF
--- a/lib/phoenix/endpoint/cowboy_handler.ex
+++ b/lib/phoenix/endpoint/cowboy_handler.ex
@@ -89,14 +89,20 @@ defmodule Phoenix.Endpoint.CowboyHandler do
   @doc """
   Callback to start the Cowboy endpoint.
   """
-  def start_link(scheme, endpoint, config, {m, f, a}) do
+  def start_link(scheme, endpoint, config, {m, f, [ref | _] = a}) do
+    # nb on [ref | _] above; ref is used by Ranch to identify its listeners,
+    # defaulting to plug.HTTP and plug.HTTPS and overridable by users.
+    # See the following for more information:
+    # https://github.com/elixir-lang/plug/blob/6ef08ce/lib/plug/adapters/cowboy.ex#L39-L47
+    # https://github.com/elixir-lang/plug/blob/6ef08ce/lib/plug/adapters/cowboy.ex#L164-L166
+    # https://github.com/ninenines/ranch/blob/236d0f8/src/ranch.erl#L88-L95
     case apply(m, f, a) do
       {:ok, pid} ->
-        Logger.info info(scheme, endpoint, config)
+        Logger.info info(scheme, endpoint, ref)
         {:ok, pid}
 
       {:error, {:shutdown, {_, _, {{_, {:error, :eaddrinuse}}, _}}}} = error ->
-        Logger.error [info(scheme, endpoint, config), " failed, port already in use"]
+        Logger.error [info(scheme, endpoint, ref), " failed, port already in use"]
         error
 
       {:error, _} = error ->
@@ -104,8 +110,16 @@ defmodule Phoenix.Endpoint.CowboyHandler do
     end
   end
 
-  defp info(scheme, endpoint, config) do
-    port = config[:port]
+  # TODO: remove this when we depend on plug 1.1
+  defp get_listener_port(ref) do
+    case function_exported?(Plug.Adapters.Cowboy, :info, 1) do
+      true  -> Plug.Adapters.Cowboy.info(ref)[:port]
+      false -> :ranch.get_port(ref)
+    end
+  end
+
+  defp info(scheme, endpoint, ref) do
+    port = get_listener_port(ref)
     host = endpoint.config(:url)[:host] || "localhost"
     "Running #{inspect endpoint} with Cowboy on #{scheme}://#{host}:#{port}"
   end


lib/phoenix/endpoint/adapter.ex
```diff
@@ -182,6 +182,7 @@ defmodule Phoenix.Endpoint.Adapter do
   end
 
   defp build_url(endpoint, url) do
+    url = Keyword.put(url, :port, url[:port])
```
josevalim: This doesn't seem to do anything? :)

lib/phoenix/endpoint/adapter.ex
```diff
@@ -197,7 +198,7 @@ defmodule Phoenix.Endpoint.Adapter do
       end
 
     scheme = url[:scheme] || scheme
-    host   = url[:host]
+    host   = url[:host] || "localhost"
```
josevalim: host should always be set.

lib/phoenix/endpoint/cowboy_handler.ex
```diff
@@ -86,17 +86,35 @@ defmodule Phoenix.Endpoint.CowboyHandler do
   defp default_for(Phoenix.Transports.WebSocket), do: Phoenix.Endpoint.CowboyWebSocket
   defp default_for(_), do: nil
 
+  # Bake in retrieving the actual port from the adapter if possible, otherwise
+  # fallback to lower-level/less portable behavior.
+  defmacrop get_listener_port(ref) do
+    case :info in Plug.Adapters.Cowboy.__info__(:functions) do
```
josevalim: Please use `function_exported?/3`.

lib/phoenix/endpoint/cowboy_handler.ex
```diff
@@ -86,17 +86,35 @@ defmodule Phoenix.Endpoint.CowboyHandler do
   defp default_for(Phoenix.Transports.WebSocket), do: Phoenix.Endpoint.CowboyWebSocket
   defp default_for(_), do: nil
 
+  # Bake in retrieving the actual port from the adapter if possible, otherwise
+  # fallback to lower-level/less portable behavior.
+  defmacrop get_listener_port(ref) do
```
josevalim: Why is this a macro? :)

lib/phoenix/endpoint/cowboy_handler.ex
```diff
@@ -86,17 +86,35 @@ defmodule Phoenix.Endpoint.CowboyHandler do
...
+
   @doc """
   Callback to start the Cowboy endpoint.
   """
-  def start_link(scheme, endpoint, config, {m, f, a}) do
+  def start_link(_scheme, endpoint, config, {m, f, a}) do
     case apply(m, f, a) do
       {:ok, pid} ->
-        Logger.info info(scheme, endpoint, config)
+        listen_port = a |> hd |> get_listener_port()
```
josevalim: endpoint.url != scheme/endpoint being started. The same endpoint can be started under 5 different host + scheme combination but the canonical url is just one.

lib/phoenix/endpoint/cowboy_handler.ex
```diff
@@ -104,9 +122,5 @@ defmodule Phoenix.Endpoint.CowboyHandler do
     end
   end
 
-  defp info(scheme, endpoint, config) do
```
josevalim: Same thing here. endpoint.url != scheme/endpoint being started. For example, we can start the same endpoint under https and http, with your changes, it wll show the same url though, because `endpoint.url` would always be the same.

lib/phoenix/endpoint/adapter.ex
josevalim: This is the line that always set the host: https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/endpoint/adapter.ex#L124

lib/phoenix/endpoint/adapter.ex
Kavec: That line is definitely vestigal, removing it shortly!

lib/phoenix/endpoint/adapter.ex
Kavec: Got it, removing the extra check.

lib/phoenix/endpoint/cowboy_handler.ex
Kavec: To call attention to it; my hope is that it'll be removed once plug is certain to be exporting the info function. I can switch it over to a standard function if you'd like and, if you're okay with leaving it as-is, I've written in a little "remove me" note for the next commit on this PR.

test/phoenix/integration/endpoint_test.exs
```diff
@@ -7,7 +7,7 @@ defmodule Phoenix.Integration.EndpointTest do
   alias Phoenix.Integration.AdapterTest.DevEndpoint
 
   Application.put_env(:endpoint_int, ProdEndpoint,
-      http: [port: "4807"], url: [host: "example.com"], server: true)
```
josevalim: We are changing existing tests with this. We shouldn't do it.

lib/phoenix/endpoint/cowboy_handler.ex
Kavec: That... is unhappily surprising. Is there a good spot to drop tests in to showcase this behavior while I'm reverting the logic back?

josevalim: To repeat what I said on IRC, the same phoenix application can run on 10 different servers on 10 different ports and the endpoint.url would be the same, "example.com:80" which would be proxied to those servers. So we should change the code below to print something like this:

    Running #{inspect endpoint} with Cowboy using #{scheme} on port #{port}

Kavec:
Reduced scope of changes as per the discussion with @josevalim here and on IRC.

lexmag:
@Kavec please squash everything into single commit. :bowtie:

Kavec:
Aye aye, will do the same treatment to the other PR as well.

lib/phoenix/endpoint/cowboy_handler.ex
```diff
@@ -92,7 +92,7 @@ defmodule Phoenix.Endpoint.CowboyHandler do
   def start_link(scheme, endpoint, config, {m, f, a}) do
     case apply(m, f, a) do
       {:ok, pid} ->
-        Logger.info info(scheme, endpoint, config)
+        Logger.info info(scheme, endpoint, hd(a))
```
chrismccord: what is `hd(a)` in this context? Let's name it here or pattern match it as a variable so it's clear

PR #1407, opened 2015-12-14
(configurable) Endpoint Handler Behaviour

zampino:
- allow configuring Endpoint Handler via application config,
  to pass a custom childspec to Phoenix.Endpoint.Server supervisor

- defaults to Phoenix.Endpoint.CowboyHandler

### DIFF
--- a/lib/phoenix/endpoint/handler.ex
+++ b/lib/phoenix/endpoint/handler.ex
@@ -1,0 +1,22 @@
+defmodule Phoenix.Endpoint.Handler do
+  @moduledoc """
+  API for exporting webserver plug adapter child spec
+
+  A handler will need to implement a child_spec/3
+  function which takes
+
+    * the schme of the endpoint :http or :https
+    * phoenix top-most endpoint module
+    * phoenix app configuration for the specified scheme
+
+  it has to return a child_spec
+  """
+  use Behaviour
+
+  @doc """
+  provides the children specification to be passed
+  to Phoenix.Endpoint.Server supervisor
+  """
+  @callback child_spec(scheme :: atom, endpoint :: module, config :: Keyword.t) :: Supervisor.Spec.spec
+
+end
--- a/lib/phoenix/endpoint/server.ex
+++ b/lib/phoenix/endpoint/server.ex
@@ -1,7 +1,6 @@
 defmodule Phoenix.Endpoint.Server do
   # The supervisor for the underlying handlers.
   @moduledoc false
-  @handler Phoenix.Endpoint.CowboyHandler
 
   use Supervisor
   require Logger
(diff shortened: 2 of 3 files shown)


lib/phoenix/endpoint/handler.ex
```diff
@@ -0,0 +1,22 @@
...
+
+  it has to return a child_spec
+  """
+  use Behaviour
+
+  @doc """
+  provides the children specification to be passed
+  to Phoenix.Endpoint.Server supervisor
+  """
+  @callback child_spec(scheme :: atom, endpoint :: atom, config :: Keyword.t) :: Supervisor.Spec.spec
```
zampino: is it ok to treat a Module as an atom for type description in the callback signature?

lib/phoenix/endpoint/server.ex
```diff
@@ -30,6 +29,7 @@ defmodule Phoenix.Endpoint.Server do
   defp default(config, otp_app, port) do
```
zampino: shall I better test this function asserting for the default value of `config` at `:handler` ?
would `Kernel#apply` work for testing private functions?

lib/phoenix/endpoint/handler.ex
chrismccord: You can use `Module.t`, but atom would work

lib/phoenix/endpoint/server.ex
chrismccord: apply only works for public functions

lib/phoenix/endpoint/handler.ex
josevalim: There is no `Module.t` iirc.

lexmag: There is [`module`](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/kernel/typespec.ex#L107) type available.

chrismccord: apparently I've used `Module.t` or dreamt using it without compiler issues. Good to know :)

zampino: Ok, I'll use `module` type

lib/phoenix/endpoint/server.ex
zampino: any hints on how/what to test this?
shall we make `default/3` public?

lib/phoenix/endpoint/handler.ex
```diff
@@ -0,0 +1,22 @@
...
+
+  it has to return a child_spec
+  """
+  use Behaviour
+
+  @doc """
+  provides the children specification to be passed
+  to Phoenix.Endpoint.Server supervisor
+  """
+  @callback child_spec(scheme :: atom, endpoint :: module, config :: Keyword.t) :: Supervisor.Spec.spec
```
josevalim: `defcallback` if we want to remain compatible with Elixir 1.0.

lib/phoenix/endpoint/server.ex
```diff
@@ -30,6 +29,7 @@ defmodule Phoenix.Endpoint.Server do
...
     config =
       config
+      |> Keyword.put_new(:handler, Phoenix.Endpoint.CowboyHandler)
```
josevalim: Instead of defining it here, just use add a key here:

https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/endpoint/adapter.ex#L108

This particular clause is only invoked for things defined dynamically.

josevalim:
The PR is perfect. I have just added two tiny comments.

josevalim:
I will do those small changes locally, thank you @zampino!

zampino:
He, great! 

After your comment here I would have moved the key :handler at root of the config, as it probably would make less sense to have different handlers for :http and :https....

PR #1412, opened 2015-12-17
Enhance instrumentation docs in Phoenix.Instrument

whatyouhide:
This PR adds documentation to the `Phoenix.Endpoint.Instrument` module which gives a detailed overview of the instrumentation API we added to Phoenix. Let me know how it looks!

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -240,30 +240,88 @@ defmodule Phoenix.Endpoint do
   instrumenter module interested in that event will have to export
   `render_view/3`.
 
+  **Note**: since the configuration for the list of instrumenters is specified
+  at compile time but it's used inside Phoenix itself, if you change this
+  configuration you'll have to recompile Phoenix manually:
+
+      $ mix deps.compile phoenix
+      $ mix compile
+
   ### Callbacks cycle
 
   The way event callbacks are called is the following.
 
     1. The event callback is called *before* the event happens (in this case,
-       before the view is rendered). The callback is called with the following
-       arguments:
-
-           MyInstrumenter.render_view(:start, compile_meta, runtime_meta)
-
-       `compile_meta` is a map of compile-time metadata (like the file and line
-       where the instrumentation is being done). `runtime_meta` is a term that
-       is passed on by the caller of the instrumentation. The result of this
-       call is stored and later passed to the after callback.
+       before the view is rendered) with the atom `:start` as the first
+       argument; see the "Before clause" section below.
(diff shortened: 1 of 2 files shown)


josevalim:
This is a private module, we should not expose it. We can either add this
to Phoenix.Endpoint (which I think we did?) or move it to a guide?

Other than that, the docs are excellent! :D


-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

chrismccord:
Ah sorry, I told @whatyouhide to open up the module with a `@moduledoc`. We can put on in `Phoenix.Endpoint`, but it's a quite a chunk of docs. I'd prefer to see it live in the codebase vs a separate guide. This would be easier once we have everything on ex_doc, but if it's b/w guide or docs, let's move the docs to `Phoenix.Endpoint` with  an Instrumentation section

whatyouhide:
I completely overlooked the fact that I already wrote quite a lot of docs in `Phoenix.Endpoint`, so lots of things were duplicated between `Phoenix.Endpoint` and `Phoenix.Endpoint.Instrument`. I moved all the docs to `Phoenix.Instrument`, so this should be ready for review.

lib/phoenix/endpoint.ex
```diff
@@ -240,30 +240,81 @@ defmodule Phoenix.Endpoint do
...
+            end
+
+            def event_callback(:stop, _time_diff, start_time) do
+              stop_time = :erlang.monotonic_time(:micro_seconds)
+              do_something_with_diff(stop_time - start_time)
+            end
+          end
+
+  The return value of each "before" event callback will be stored and passed to
+  the corresponding "after" callback.
```
josevalim: Should we add a tiny note saying it is not guaranteed the after event will be invoked? And also say the instrumenter is invoked in the same process as the one running the request so we should be careful with blocking actions?

josevalim:
One tiny suggestions and we are good to go!

lib/phoenix/endpoint.ex
whatyouhide: +1 on the same process thing, but I think we always call "after" callbacks? https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/endpoint/instrument.ex#L67-L72

josevalim: We don't always call because a link may break.

whatyouhide:
@josevalim made the changes, ready for review

josevalim:

chrismccord:
Thanks Andrea!

Issue #977, opened 2015-06-22
Router: Allow route to match all HTTP verbs


tisba:
While trying to build a simple HTTP echo server to play around with phoenix, I have to write 

```elixir
get "/echo/*rest", ApiController, :echo
post "/echo/*rest", ApiController, :echo
put "/echo/*rest", ApiController, :echo
patch "/echo/*rest", ApiController, :echo
delete "/echo/*rest", ApiController, :echo
options "/echo/*rest", ApiController, :echo
connect "/echo/*rest", ApiController, :echo
trace "/echo/*rest", ApiController, :echo
head "/echo/*rest", ApiController, :echo
```

I'd like to have something similar to routes generated by `resources`:

```elixir
match "/echo/*rest", ApiController, :echo, :all
```

This could also allow

```elixir
match "/echo/*rest", ApiController, :echo, only: [:get, :post]
```

Gazler:
:+1+ for this.

I think this could be implemented pretty easily by doing something like:

```
  defmacro match(path, controller, action) do
    do_match path, controller, action, [], do: nil
  end

  #...other macro implementations

  def do_match(path, controller, action, options, do: context) do
  for verb <- @http_methods do
    method = verb |> to_string |> String.upcase
      quote do
        var!(add_route, Phoenix.Router).(
          Scope.route(__MODULE__, unquote(method), unquote(path), unquote(controller),
                                  unquote(action), unquote(options))
        )
      end
    end
  end
```

chrismccord:
I think an HTTP echo server is about the only useless where this comes up, so I don't think it's worth introducing. I would prefer for the users to have their routes listed explicitly, as match can introduce bad behaviors. Like @Gazler said you could write a macro to do this in the cases where it is needed. So I'd rather wait until we have a lot of reports of need first. Thanks!

Gazler:
@chrismccord Isn't this something that could be useful in the same way that a rails engine is? You mount a plug at a particular endpoint and all request for that path get routed to that plug?

As you said, you could provide a macro in your project to do that.

chrismccord:
> You mount a plug at a particular endpoint and all request for that path get routed to that plug?

Yes, but it wouldn't be via match, it would be with a normal plug.

chrismccord:
[...]

PR #1419, opened 2015-12-20
Allow forwarding twice to a plug with different options

joshprice:
I ran into this issue (https://github.com/joshprice/plug_graphql/issues/7) whereby one plug cannot be forwarded to twice, even with different configurations.

There are 2 possible workarounds (and perhaps more):

1. Wrap the plug in another plug which encapsulates its configuration (the options)
2. Use the `get` or `post` macros instead of forward

I went with (2) mainly because it's more concise, but I lose the ability to handle *all* HTTP verbs at that path. The wrapper workaround is arguably more correct but requires a bit more redundant code.

The other possible solution is to allow the configuration to form part of the identity of the plug.

I think this is in keeping with this warning about forwarding twice (https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/router.ex#L635-L638) at least as far as I understand it.

>  Note, however, that we don't advise forwarding to another
>  endpoint. The reason is that plugs defined by your app
>  and the forwarded endpoint would be invoked twice, which
>  may lead to errors.

This PR is a WIP, currently getting this test failure, which I haven't figured out the cause for yet:

```elixir
  1) test helpers cascade script name across forwards based on main router (Phoenix.Router.ForwardTest)
     test/phoenix/router/forward_test.exs:103
     Assertion with == failed
     code: page_path(conn, :stats) == "/admin/stats"
     lhs:  "/stats"
     rhs:  "/admin/stats"
     stacktrace:
       test/phoenix/router/forward_test.exs:108
```

### DIFF
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -650,9 +650,10 @@ defmodule Phoenix.Router do
   defmacro forward(path, plug, plug_opts \\ [], router_opts \\ []) do
     router_opts = Keyword.put(router_opts, :as, nil)
 
-    quote unquote: true, bind_quoted: [path: path, plug: plug] do
-      path_segments = Route.forward_path_segments(path, plug, @phoenix_forwards)
-      @phoenix_forwards Map.put(@phoenix_forwards, plug, path_segments)
+    quote unquote: true, bind_quoted: [path: path, plug: plug, plug_opts: plug_opts] do
+      plug_with_opts = {plug, plug_opts}
+      path_segments = Route.forward_path_segments(path, plug_with_opts, @phoenix_forwards)
+      @phoenix_forwards Map.put(@phoenix_forwards, plug_with_opts, path_segments)
       unquote(add_route(:forward, :*, path, plug, plug_opts, router_opts))
     end
   end
--- a/test/phoenix/router/forward_test.exs
+++ b/test/phoenix/router/forward_test.exs
@@ -84,7 +84,7 @@ defmodule Phoenix.Router.ForwardTest do
       end
     end
 
-    assert_raise ArgumentError, ~r{`Phoenix.Router.ForwardTest.ApiRouter` has already been forwarded}, fn ->
+    assert_raise ArgumentError, ~r(`\{Phoenix.Router.ForwardTest.ApiRouter, \[\]\}` has already been forwarded), fn ->
       Code.eval_quoted(router)
     end
   end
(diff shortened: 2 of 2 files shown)


chrismccord:
This was an intentional limitation, though I I'll have to review the code again to remember the details. I believe it has to do with building the route helpers and script_name segments. @josevalim do you remember?

chrismccord:
The # 1 workaround is the correct and easiest solution to this today. I'll do some digging later to see why exactly we limit

josevalim:
We have also introduced match, which solves `2`.

chrismccord:
`match` still requires a single verb, not a catch-all

joshprice:
I'm trying to find docs on the `match`, no luck yet. Where is this defined @josevalim? Does it support multiple verbs? I really want to handle all verbs so I can return meaningful errors from the plug itself (also everything is self-contained in the plug that way). 

Stumbled across this which seemed potentially relevant: https://github.com/phoenixframework/phoenix/issues/977

joshprice:
Could you please confirm that https://github.com/phoenixframework/phoenix/blob/master/test/phoenix/router/forward_test.exs#L107-L108 is asserting the correct path? My reading of the code indicates that this doesn't look right (or at least it's correctness isn't exactly obvious ;)

chrismccord:
The way it works is like this: We know that the `AdminDashboard` is forwarded at `"/admin"` from `Router`. When we pass a "blank" `%Plug.Conn{}` to `AdminDashboard.path_path`, we have no cascading forward path. If we grab a `%Plug.Conn{}` that has passed through the `Router` (by calling GET /status on Router), then we should get that prefix based on `Router`'s forward info for `AdminDashboard`. Make sense?

chrismccord:
Since this has stagnated, we can revisit later when we tackle #1608. Thanks!

tpitale:
This would be really useful to have. The use case we've come across is for "namespacing" Absinthe Graphql to have different `scope`/`pipeline` for different authentication/authorization scopes.

chrismccord:
Until the next major release, the best solution now is to define a new module plug per Absinthe endpoint, which is a small cost to get the feature you would like.

joshprice:
@chrismccord is something changing in 1.4 that makes this possible? If so that would be awesome

MrAlexLau:
@chrismccord - is this what you mean by defining a new module plug per Absinthe endpoint?

```
defmodule MyApp.AbsinthePlug do
  @behaviour Plug

  defdelegate init(opts), to: Absinthe.Plug
  defdelegate call(call, default), to: Absinthe.Plug
end
```

Then in your router you can reference this module:
```
scope "/alternate-route/graphql" do
  forward("/", MyApp.AbsinthePlug, schema: MyApp.Schema)
end
```

rmoorman:
@MrAlexLau , for what it's worth, I also use several module plug wrappers to use more than one of the canned absinthe graphiql plugs at the same time. There are several ways to skin the cat and I would guess that your approach should work.
Here is an example of what that looks like (more along the lines of the OP's mentioned workaround 1 here https://github.com/graphql-elixir/plug_graphql/issues/7):
````
  defmodule FooWeb.GraphiQL.Simple do
    use Plug.Builder
    plug Absinthe.Plug.GraphiQL, [
      schema: FooWeb.GraphQL.Schema,
      interface: :simple,
      json_codec: Jason,
      socket: FooWeb.UserSocket,
    ]
  end

  defmodule FooWeb.GraphiQL.Advanced do
    use Plug.Builder
    plug Absinthe.Plug.GraphiQL, [
      schema: FooWeb.GraphQL.Schema,
      interface: :advanced,
      json_codec: Jason,
      socket: FooWeb.UserSocket,
    ]
  end

  defmodule FooWeb.GraphiQL.Playground do
    use Plug.Builder
    plug Absinthe.Plug.GraphiQL, [
      schema: FooWeb.GraphQL.Schema,
      interface: :playground,
      json_codec: Jason,
      socket: FooWeb.UserSocket,
    ]
  end
````
Which lets me use `Absinthe.Plug.GraphiQL` more than once with different options.

**Comment by mathieuprog** [comment]
FYI it is possible to use `match/5` with the catch-all `:*` to overcome the limitation:

```
match(:*, "/foo", Hello)
match(:*, "/bar", Hello)
```

Issue #1240, opened 2015-09-30
POST functionality inconsistent with PUT, GET, and DELETE in referenced models


iguberman:
Hope you find this a reasonable complaint... 

**Background:**
```console
$mix phoenix.new --no-html gps_data
$mix ecto.create
$mix phoenix.gen.json Datum data lat:float lon:float data:map
$mix phoenix.gen.json Tag tags name:string value:string datum_id:references:data
```
Modify models/data.ex like so to create one-to-many datum-tags association:

```elixir
schema "data" do
    field :lat, :float
    field :lon, :float
    field :data, :map

    has_many :tags, GpsData.Tag

    timestamps
end
```

Modify required_fields in models/datum.ex:
```elixir
@required_fields ~w(lat lon data tags)
```

**Issue:**

Create a new datum object by POSTing a json that includes tags list in the datum object -- I actually didn't expect this to work but it did -- Phoenix is AWESOME!!! :)  

```json
{"datum": {
   "lon":55.12345678912345,
   "lat":88.12345678912345,
   "data":{"key":"info", "value":"A lot more info" }, 
   "tags":[
        {"name":"responsible", "value":"Neo"}, 
        {"name":"user","value":"nanderson"}, 
        {"name":"occupation", "value":"hacker"}
    ] 
 } 
}
```

```console
$curl -i -X POST -H 'Content-Type:application/json' -d '{"datum": {"lon":55.12345678912345,"lat":88.12345678912345,"data":{"key":"info", "value":"A lot more info" }, "tags":[{"value":"Neo","name":"responsible"}, {"name":"user","value":"nanderson"}, {"name":"occupation", "value":"hacker"}] } }' http://localhost:4000/api/data > out.html
```

Works great!   I got all the proper associations created in the DB: 3 tag entries and 1 datum entry!  Now I got my hopes up and expect that all the other methods work like this too (NOTE: I won't query tags separately and I won't even add them to routes.ex because from a business perspective they really are part of datum object)

Now I run GET:
```console
$curl -i -X GET http://localhost:4000/api/data/1
{"data":{"lon":55.12345678912345,"lat":88.12345678912345,"id":4,"data":{"value":"A lot more info","key":"info"}}}
```
No tag info :( 

Run PUT -- definitely expect it to work with JSON object of the same format I used  in POST:
```console
[...]

PR #1422, opened 2015-12-21
Add instructional on_delete: :nothing to gen.model

maxim:
Based IRC chat with @josevalim in reference to https://github.com/phoenixframework/phoenix/issues/1240. Let me know if I misunderstood the intent.


josevalim:
I think this is nudge in the the right direction we need, thank you @maxim!

PR #1424, opened 2015-12-23
Use Phoenix.HTML tag helpers in layout template

nettofarah:
This is mostly a style change. So we don't need to interpolate
static_path/2 inside HTML tags anymore.

It will also be a good step in helping new Phoenix users to get
acquainted with Phoenix.HTML.

### DIFF
--- a/installer/templates/new/web/templates/layout/app.html.eex
+++ b/installer/templates/new/web/templates/layout/app.html.eex
@@ -8,7 +8,7 @@
     <meta name="author" content="">
 
     <title>Hello <%= application_module %>!</title>
-    <link rel="stylesheet" href="<%%= static_path(@conn, "/css/app.css") %>">
+    <%%= tag :link, rel: "stylesheet", href: static_path(@conn, "/css/app.css") %>
   </head>
 
   <body>
@@ -30,6 +30,6 @@
       </main>
 
     </div> <!-- /container -->
-    <script src="<%%= static_path(@conn, "/js/app.js") %>"></script>
+    <%%= content_tag :script, "", src: static_path(@conn, "/js/app.js") %>
   </body>
 </html>


josevalim:
I have mixed feelings so I will leave it up to @chrismccord. :)

josevalim:
In particular, the fact we use `tag` on one and `content_tag` in the other is not obvious (it is obvious now because we know how the previous code was). :)

nettofarah:
@josevalim I agree. The use of two different tags makes it ambiguous.
One idea would be to implement a version of `content_tag` that doesn't take a content string and just standardize on `content_tag` everywhere.

how does that sound?

josevalim:
I think the issue would still be there and it would also make content_tag
inconsistent with itself. :(

-- 


nettofarah:
You could argue that html is inconsistent in the way that `script` tags work.
e.g. Empty `<script></script>` with a source attribute.

We currently don't have a way to represent a script tag, unless we do `<%= content_tag :script, "", ... %>`.

On the other hand, if we use `<%= tag ... %>` everywhere we end up with something like:
`<script ...>` (no closing tag).

I suggested that we could create a `<%= script_tag ...%>` or a more generic `<%= asset_tag ..%>` on https://github.com/phoenixframework/phoenix_html/issues/55

I'm fine with closing the issue if you think there's no clear solution. Though I'm personally not a big fan of how Phoenix currently sets up the initial templates. i.e. `<%= ` inside html attributes.

josevalim:
> You could argue that html is inconsistent in the way that script tags work. e.g. Empty <script></script> with a source attribute.

Right. That's the root issue and exactly why I don't want to hide behind tag and content_tag.

I am also unsure on script_tag because there are many ways we could use it, so it is hard to choose one. That's why Rails shipped with things like javascript_link_tag which helps but has its own problem. I don't think `asset_tag` would work exactly because of what we are discussing. :(

nettofarah:
Yeah, these are fair points.
So maybe there's no easy solution right now?

Do you have any suggestions?

Thanks!

josevalim:
@nettofarah yes, I don't have any suggestions right now, sorry. :(

nettofarah:
@josevalim No problem.

chrismccord:
Thanks for the suggestion @nettofarah, but I'm going to hold off on this one for now.

nettofarah:
No problem, @chrismccord. Thanks for looking into it.

PR #1429, opened 2015-12-27
Refactor to fix dialyzer crash

chrismcg:
The original code triggers an error in dialyzer that causes it to crash. You can verify this by adding: `{:dialyze, "~> 0.2"},` to the phoenix deps and running `mix deps.get && mix compile && mix dialyze` before and after applying this patch.

This also removes the duplication in the case clauses.

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -639,15 +639,15 @@ defmodule Phoenix.Endpoint do
     compile = Phoenix.Endpoint.Instrument.strip_caller(__CALLER__) |> Macro.escape()
 
     quote do
-      case unquote(endpoint_or_conn_or_socket) do
-        %Plug.Conn{private: %{phoenix_endpoint: endpoint}} ->
-          endpoint.instrument(unquote(event), unquote(compile), unquote(runtime), unquote(fun))
-        %Phoenix.Socket{endpoint: endpoint} ->
-          endpoint.instrument(unquote(event), unquote(compile), unquote(runtime), unquote(fun))
-        %{__struct__: struct} when struct in [Plug.Conn, Phoenix.Socket] ->
-          unquote(fun).()
-        endpoint ->
-          endpoint.instrument(unquote(event), unquote(compile), unquote(runtime), unquote(fun))
+      endpoint = case unquote(endpoint_or_conn_or_socket) do
+        %Plug.Conn{private: %{phoenix_endpoint: endpoint}} -> endpoint
+        %Phoenix.Socket{endpoint: endpoint} -> endpoint
+        %{__struct__: struct} when struct in [Plug.Conn, Phoenix.Socket] -> nil
+        endpoint -> endpoint
+      end
+      case endpoint do
+        nil -> unquote(fun).()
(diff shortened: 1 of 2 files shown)


chrismccord:
Thanks for taking a look at this Chris! Do we know why dializyer was crashing with the existing code? I'd prefer to fix dialyze or Elixir before arbitrary changing the code to something that happens to work. I also prefer the current code, so some insight into the cause of the crash would be helpful determining which way to go. Thanks!

chrismcg:
No clue as yet, I've created a smaller case that triggers the same failure in dialyzer here:

https://github.com/chrismcg/dial_test

I don't know anywhere near enough about dialyzer to understand why it causes it to crash. It's not strictly related to the data as the number of arguments to the function called from the case statement matters too. It also doesn't happen with a non-macro version (no warning there either). Hopefully someone who understands dialyzer a lot better will be able to look at it and figure out what's happening!

chrismcg:
https://github.com/fishcakez/dialyze/issues/10 is where the dialyzer issue is being tracked for now

chrismcg:
http://bugs.erlang.org/browse/ERL-68

fishcakez:
We should be able avoid this issue in this specific case on 18+/1.2 by marking the AST with `generated: true`. @josevalim why didn't we add support for this quoted option?

josevalim:
@fishcakez there was a good reason but i can't remember. I believe the official why is to directly add `[generated: true]` to the metadata because `quote` would mark more things than it should.

lib/phoenix/endpoint.ex
```diff
@@ -639,15 +639,15 @@ defmodule Phoenix.Endpoint do
...
-      case unquote(endpoint_or_conn_or_socket) do
-        %Plug.Conn{private: %{phoenix_endpoint: endpoint}} ->
-          endpoint.instrument(unquote(event), unquote(compile), unquote(runtime), unquote(fun))
-        %Phoenix.Socket{endpoint: endpoint} ->
-          endpoint.instrument(unquote(event), unquote(compile), unquote(runtime), unquote(fun))
-        %{__struct__: struct} when struct in [Plug.Conn, Phoenix.Socket] ->
-          unquote(fun).()
-        endpoint ->
-          endpoint.instrument(unquote(event), unquote(compile), unquote(runtime), unquote(fun))
+      endpoint = case unquote(endpoint_or_conn_or_socket) do
```
josevalim: This refactoring looks good specially because we can move this `case` to an external function we could call.

chrismcg:
Moved the case to an external function. However I thought there might be some other place in the code that does this and I found https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/token.ex#L123-L125. I'm thinking there should be one place to do this and it should probably live in Phoenix.Endpoint.

chrismcg:
Though the `get_endpoint` in `Phoenix.Token` works slightly differently as it will raise if there's no endpoint in the conn or socket.

josevalim:

josevalim:
Yes, we could likely extract into a single function or a protocol but let's wait for one more use case before we move on.

chrismcg:
:+1: thanks

PR #1435, opened 2015-12-29
Fix dialyzer warning in put_new_layout

chrismcg:
The @spec said the tuple should be {atom, binary} but in `__using__` the second part of the tuple is the symbol `:app`. Seems like a symbol should be allowed so this commit adjusts the spec.


chrismccord:
<3 <3 <3

PR #1441, opened 2015-12-31
Translated errors.po into japanese(ja/LC_MESSAGES/erros.po)

hykw:
I've translated errors.pot in Japanese, and confirmed it works well, in case of `config :my_app, MyApp.Gettext, default_locale: "ja"` in config/config.exs, as suggested in  http://stackoverflow.com/questions/34538502/how-to-set-locale-for-errors-po


josevalim:
Thank you! However, we don't plan to host different translations. Someone can create a separate repository if they want to and host all of those .po files. Happy new year!

hykw:
OK.
I've hosted the file in my repository(https://github.com/hykw/phoenix-locale_ja) for the time being.

PR #1448, opened 2016-01-04
Fix Phoenix.Endpoint instrument example

nuss-justin:
The example at https://github.com/phoenixframework/phoenix/blob/4d4f1b77/lib/phoenix/endpoint.ex#L332 uses the binary "index.html" as argument to Endpoint.Instrument, but this fails as the generated instrument function(s) have a is_map guard on the runtime argument. This changes the example to use a map instead.


chrismccord:
<3 <3 <3

PR #1449, opened 2016-01-04
Add :connection module to applications list in mix.exs if Ecto is used.

Alamoz:
Starting in Phoenix 1.1.0, an application crashes due to not having the Connection module in the applications list (see https://github.com/bitwalker/exrm/issues/265.) The crash occurs in Phoenix apps that have been deployed as a release; apps running locally on development machines are unaffected.

Adding the Connection module to the mix.exs template will relieve users of experiencing the crash, having to figure out the problem and then adding it to the list. It is only added if Ecto is used.

### DIFF
--- a/installer/templates/new/mix.exs
+++ b/installer/templates/new/mix.exs
@@ -23,7 +23,7 @@ defmodule <%= application_module %>.Mixfile do
   def application do
     [mod: {<%= application_module %>, []},
      applications: [:phoenix<%= if html do %>, :phoenix_html<% end %>, :cowboy, :logger, :gettext<%= if ecto do %>,
-                    :phoenix_ecto, <%= inspect adapter_app %><% end %>]]
+                    :phoenix_ecto, :connection, <%= inspect adapter_app %><% end %>]]
   end
 
   # Specifies which paths to compile per environment.


josevalim:
Thank you! We need to add this to the driver instead of directly to Phoenix. Which database are you using? :)

Alamoz:
Postgres, please see issue https://github.com/bitwalker/exrm/issues/265

josevalim:
Thank you! It has been fixed in Postgrex: https://github.com/ericmj/postgrex/commit/a9f4e553e4815d16d6cfa1063d6fd717c40d234e

Alamoz:
Thank you. Why does it need to be added at the driver level? Is the Connection module only for Postgres?

josevalim:
@Alamoz you must add it to the package that depends directly on `:connection`. In this case, postgrex depends on connection, so postgrex was the one with the bug.

Alamoz:
Interesting. When I add it to mix.exs in the applications list the problem goes away. Thanks again!

josevalim:
Yes, you can add it to your own app for now but the correct fix is to list it on whoever depends on it. :)

PR #1450, opened 2016-01-05
Support broadcasts to individual destination nodes

chrismccord:
Support for scoping a broadcast is needed for Phoenix.Presence features. This PR adds scoped broadcasts via tuple to mirror the `send` api. @josevalim I could use your opinion on the API choices and configuration options. We could also make these changes backwards compatible by only invoking the adapters broadcast/5 if we have a target node, but this PR as written will break existing pubsub adapters. 

My test is also pretty weak, but without standing up a 2nd VM, I think it's the best we've got.

### DIFF
--- a/lib/phoenix/pubsub/pg2.ex
+++ b/lib/phoenix/pubsub/pg2.ex
@@ -30,16 +30,13 @@ defmodule Phoenix.PubSub.PG2 do
   end
 
   @doc false
-  def init([server, opts]) when is_atom(server) do
-    init([{server, node()}, opts])
-  end
-  def init([{server, node_name}, opts]) do
+  def init([server, opts]) do
     pool_size = Keyword.fetch!(opts, :pool_size)
     dispatch_rules = [{:broadcast, Phoenix.PubSub.PG2Server, [server, pool_size]}]
 
     children = [
       supervisor(Phoenix.PubSub.LocalSupervisor, [server, pool_size, dispatch_rules]),
-      worker(Phoenix.PubSub.PG2Server, [{server, node_name}]),
+      worker(Phoenix.PubSub.PG2Server, [server]),
     ]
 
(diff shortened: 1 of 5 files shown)


lib/phoenix/pubsub/pg2.ex
```diff
@@ -26,13 +30,16 @@ defmodule Phoenix.PubSub.PG2 do
   end
 
   @doc false
-  def init([server, opts]) do
+  def init([server, opts]) when is_atom(server) do
```
josevalim: Do we need both init signatures?

lib/phoenix/pubsub/pg2_server.ex
```diff
@@ -24,11 +24,15 @@ defmodule Phoenix.PubSub.PG2Server do
...
-    :ok = :pg2.create(pg2_namespace)
-    :ok = :pg2.join(pg2_namespace, self)
-    {:ok, name}
+  def init({server_name, node_name}) do
+    public_pg2_namespace = pg2_namespace(server_name, :global)
+    private_pg2_namespace = pg2_namespace(server_name, node_name)
+    :ok = :pg2.create(public_pg2_namespace)
+    :ok = :pg2.create(private_pg2_namespace)
+    :ok = :pg2.join(public_pg2_namespace, self)
+    :ok = :pg2.join(private_pg2_namespace, self)
```
josevalim: Because pg2 relies on a fully connected Erlang network, we theoretically don't need a private pg2. We can just send a message using `send {registered_name, node}, message`. Thoughts?

chrismccord:
This is ready now. Should we land this in core before extracting?

josevalim:
> This is ready now. Should we land this in core before extracting?

I guess it doesn't matter? I mean, master is 1.2 so this will never be released anyway. So whatever fits your workflow best.

PR #1457, opened 2016-01-08
Update phoenix.new installed .gitignore

ThomasArts:
When you install phoenix as a mix archive, it becomes logical to create a new git repository for any phoenix project. That git repository should probably ignore mix.lock.

### DIFF
--- a/installer/templates/static/bare/.gitignore
+++ b/installer/templates/static/bare/.gitignore
@@ -3,6 +3,7 @@
 /db
 /deps
 /*.ez
+mix.lock
 
 # Generate on crash by the VM
 erl_crash.dump


sevenseacat:
why??

ThomasArts:
Well, I don't want to put mix.lock under version control. Is that strange?
Note that I am not an Elixir expert, so there might be a good reason to put mix.lock under version control, but in general the first thing I would do after a 
mix phoenix.new myproject
is to make a git repo of myproject. In that git repo I do not want mix.lock. Thus, I add it to my .gitignore. This can be done automatic with the proposed change. Hence the pull request.

Cheers
Thomas

sevenseacat:
I think it's very strange, as you'd want to ensure that you're using the same known-working versions of libraries in all environments. Others may have different opinion though.

Gazler:
The lock file should be checked in.

More information on the `mix.lock` file is available at http://elixir-lang.org/getting-started/mix-otp/dependencies-and-umbrella-apps.html#external-dependencies

The relevant section is:

 > You will notice that when you add a dependency to your project, Mix generates a mix.lock file that guarantees repeatable builds. The lock file must be checked in to your version control system, to guarantee that everyone who uses the project will use the same dependency versions as you.

ThomasArts:
Thank you all for the feedback. I'll check in lock files, disregard my pull request.

/Thomas

PR #1469, opened 2016-01-13
Support --binary-id when generating scaffold

hubertlepicki:
Currently if you generate a project with `--binary-id` option, this project is properly using binary IDs almost everywhere by default.

This is, however, not true about the generated tests for the controllers - for both HTML and JSON.

This pull requests fixes the tests so the generated tests are respecting global `binary_id` generators setting. Tests scaffolded using phoenix.gen.json or phoenix.gen.html after this patch do pass correctly when `--binary-id` is used.

### DIFF
--- a/lib/mix/tasks/phoenix.gen.html.ex
+++ b/lib/mix/tasks/phoenix.gen.html.ex
@@ -25,14 +25,20 @@ defmodule Mix.Tasks.Phoenix.Gen.Html do
   information on attributes and namespaced resources.
   """
   def run(args) do
-    {opts, parsed, _} = OptionParser.parse(args, switches: [model: :boolean])
+    switches = [binary_id: :boolean, model: :boolean]
+
+    {opts, parsed, _} = OptionParser.parse(args, switches: switches)
     [singular, plural | attrs] = validate_args!(parsed)
+
+    default_opts = Application.get_env(:phoenix, :generators, [])
+    opts = Keyword.merge(default_opts, opts)
 
     attrs   = Mix.Phoenix.attrs(attrs)
     binding = Mix.Phoenix.inflect(singular)
     path    = binding[:path]
     route   = String.split(path, "/") |> Enum.drop(-1) |> Kernel.++([plural]) |> Enum.join("/")
     binding = binding ++ [plural: plural, route: route, attrs: attrs,
+                          binary_id: opts[:binary_id],
                           inputs: inputs(attrs), params: Mix.Phoenix.params(attrs),
                           template_singular: String.replace(binding[:singular], "_", " "),
                           template_plural: String.replace(plural, "_", " ")]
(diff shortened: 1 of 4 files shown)


josevalim:
:shipit:

chrismccord:
<3 <3 <3

brado:
@josevalim @hubertlepicki @chrismccord  I'm not seeing any change of behavior on the phoenix master or v1.1.3 per the issue mentioned above. I'm getting the exact same test error on controller tests generated using mix phoenix.gen.html as before using the master. The pull request referenced seems to change the output of the generator to include test data for binary ID's of "11111111-1111-1111-1111-111111111111". That seems to work fine, as I see that sample data in my controller tests. But the problem is in the handling of it. I receive this error when I run the test:
```
 1) test renders page not found when id is nonexistent (Shrek.CarControllerTest)
     test/controllers/car_controller_test.exs:35
     expected response status to be 404, but got 500 from:
     
     ** (FunctionClauseError) no function clause matching in Mongo.Ecto.ObjectID.dump/1
     
     stacktrace:
       (mongodb_ecto) lib/mongo_ecto/objectid.ex:25: Mongo.Ecto.ObjectID.dump("11111111-1111-1111-1111-111111111111")
       (elixir) lib/enum.ex:1473: Enum."-reduce/3-lists^foldl/2-0-"/3
       (elixir) lib/enum.ex:1151: Enum."-map_reduce/3-lists^mapfoldl/2-0-"/3
       (ecto) lib/ecto/repo/queryable.ex:91: Ecto.Repo.Queryable.execute/5
       (ecto) lib/ecto/repo/queryable.ex:15: Ecto.Repo.Queryable.all/4
       (ecto) lib/ecto/repo/queryable.ex:55: Ecto.Repo.Queryable.one!/4
       (shrek) web/controllers/car_controller.ex:32: Shrek.CarController.show/2
       (shrek) web/controllers/car_controller.ex:1: Shrek.CarController.action/2
       (shrek) web/controllers/car_controller.ex:1: Shrek.CarController.phoenix_controller_pipeline/2
       (shrek) lib/phoenix/router.ex:261: Shrek.Router.dispatch/2
       (shrek) web/router.ex:1: Shrek.Router.do_call/2
```
I must be doing something wrong...if you guys could advise, I'd appreciate it. Thx.

michalmuskala:
It looks like we're using always UUID here, but `binary_id` is not always a UUID.

That said I think we should keep the current behaviour, probably with a note somewhere. Only the Ecto adapter knows what `binary_id` stands for, but we don't have access to it when running generators. UUID seems to be a reasonable choice given 4 out of 5 adapters are using it as `binary_id`.

brado:
Thanks for the reply, @michalmuskala. I'm not sure how it is fixed, and whether change would be required in phoenix, ecto, etc_mongodb, or all, but either this issue needs to be reopened, or a new one needs to be created (I'm not sure what the preference is) -- because bottom line, a generated resource shouldn't result in errors of any kind. If I need toads a new issue, let me know, and I'll do so. Thx.

PR #1475, opened 2016-01-18
Don't include Ecto gettext translations if --no-ecto

stevedomin:
If the project is generated with --no-ecto we don't want to include the
changeset error translations in .po* files.

* If the project is generated with `--no-html` and `--no-ecto` should we completely remove the `ErrorHelpers` module?
* The `phoenix.gen.json` task relies on the `ErrorHelpers.translate_error` to exist. It won't if the project is generated with `--no-ecto`, is that ok?

### DIFF
--- a/installer/templates/new/priv/gettext/en/LC_MESSAGES/errors.po
+++ b/installer/templates/new/priv/gettext/en/LC_MESSAGES/errors.po
@@ -4,7 +4,7 @@
 ## --merge` or `mix gettext.merge` to merge POT files into PO files.
 msgid ""
 msgstr ""
-"Language: en\n"
+"Language: en\n"<%= if ecto do %>
 
 ## From Ecto.Changeset.cast/4
 msgid "can't be blank"
@@ -96,4 +96,4 @@ msgstr[1] ""
 msgid "must be equal to %{count}"
 msgid_plural "must be equal to %{count}"
 msgstr[0] ""
-msgstr[1] ""
+msgstr[1] ""<% end %>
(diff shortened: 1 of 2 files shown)


josevalim:
> If the project is generated with --no-html and --no-ecto should we completely remove the ErrorHelpers module?

I would say that no.

> The phoenix.gen.json task rely on the ErrorHelpers.translate_error to exist. It won't if the project is generated with --no-ecto, is that ok?

I would say we don't need to worry about those cases because the generators assume Ecto is being used. so that's not the only thing that is going to fail :)

josevalim:
Great job btw!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

stevedomin:
Ok, I'll fix the specs and we can merge then.

stevedomin:
@josevalim `phoenix.gen.html` relies on `ErrorHelpers.error_tag/2`, which itself relies on `ErrorHelpers.translate_error/1` so I'm thinking I'll keep the latter around even if you generate your project with `--no-ecto` but change the comment we have [here](https://github.com/phoenixframework/phoenix/blob/master/installer/templates/new/web/views/error_helpers.ex#L21-L28) to not mention Ecto.

What do you think?

josevalim:
Sounds good!

stevedomin:
Thanks @josevalim, still wanted to tweak the comment but couldn't come up with anything good

josevalim:
Ah, sorry, forgout about that!

stevedomin:
No worries @josevalim I'll send another PR>

PR #1487, opened 2016-01-21
Deprecate ConnTest.conn and add ConnTest.build_conn

### LINKED ISSUE #1486: Deprecate Phoenix.ConnTest.conn/0 and add Phoenix.ConnTest.build_conn/0
We discussed this in IRC https://botbot.me/freenode/elixir-lang/2016-01-21/?msg=58418573&page=10

There is often confusion around this because people often create a conn with `conn/0` and assign it to a `conn` variable. Switching to `build_conn/0` should help make this more clear.

Should we also deprecate `conn/3` and add `build_conn/3` so there aren't two function names for building connections?

paulcsmith:
Closes #1486.

### DIFF
--- a/lib/phoenix/test/conn_test.ex
+++ b/lib/phoenix/test/conn_test.ex
@@ -14,21 +14,21 @@ defmodule Phoenix.ConnTest do
   the preferred way to test anything that your router dispatches
   to.
 
-      conn = get conn(), "/"
+      conn = get build_conn(), "/"
       assert conn.resp_body =~ "Welcome!"
 
-      conn = post conn(), "/login", [username: "john", password: "doe"]
+      conn = post build_conn(), "/login", [username: "john", password: "doe"]
       assert conn.resp_body =~ "Logged in!"
 
   As in your application, the connection is also the main abstraction
-  in testing. `conn()` returns a new connection and functions in this
+  in testing. `build_conn()` returns a new connection and functions in this
   module can be used to manipulate the connection before dispatching
   to the endpoint.
 
   For example, one could set the accepts header for json requests as
   follows:
 
-      conn()
+      build_conn()
       |> put_req_header("accept", "application/json")
       |> get("/")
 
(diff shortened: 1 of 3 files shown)


lib/phoenix/test/conn_test.ex
```diff
@@ -108,9 +108,20 @@ defmodule Phoenix.ConnTest do
...
+
+  @doc """
+  Deprecated version of build_conn/0. Use build_conn/0 instead
+  """
+  @spec build_conn() :: Conn.t
   def conn() do
-    conn(:get, "/", nil)
+    IO.puts :stderr, """
+    warning: using conn/0 to build a connection is deprecated. Use build_conn/0 instead.
+    """
```
josevalim: `IO.write` to avoid duplicated new lines and pls include the stacktrace as well.

```elixir
IO.write :stderr, """
warning: ...
#{Exception.format_stacktrace}
"""
```

test/phoenix/test/conn_test.exs
```diff
@@ -51,16 +52,40 @@ defmodule Phoenix.Test.ConnTest do
     :ok
   end
 
-  test "conn/0 returns a new connection" do
-    conn = conn()
+  test "conn/0 returns a new connection and issues a deprecation warning" do
```
josevalim: You can kill those two tests, we will be fine. :)

josevalim:
perfect, i have added some comments :D

lib/phoenix/test/conn_test.ex
paulcsmith: I didn't know about that. Very nice! I'll include the stack trace :+1:

test/phoenix/test/conn_test.exs
paulcsmith: Yeah I wasn't sure about that either. I'll remove them :)

paulcsmith:
Thanks for the review @josevalim. I believe I addressed everything :)

paulcsmith:
CI is failing, but it passes locally. I did notice that I get the same failure as CI when I switch to Elixir 1.1.1 so maybe CI needs to run on the latest elixir, or maybe there is really a problem with Phoenix on 1.1.1?

josevalim:
We were using 1.2 only syntax. Fixed in master now. I will let @chrismccord merge this. :)

chrismccord:
<3 <3 <3

PR #1471, opened 2016-01-22
Default Router macros `plug_opts` to empty list

scrogson:


test/phoenix/router/routing_test.exs
```diff
@@ -34,6 +39,8 @@ defmodule Phoenix.Router.RoutingTest do
     match :move, "/move", UserController, :move
 
     get "/users/:user_id/files/:id", UserController, :image
+    get "/standard", StandardPlug
```
chrismccord: let's use a function plug here instead so we don't have to create another module

josevalim:
I am :-1:. The most common usage is to include an action. With this change, folks will write `get "/foo", PostController` and it won't work. I think being explicit here is worth it. :)

scrogson:
@josevalim explicit as in `get "/foo", MyPlug, []` ???

chrismccord:
Yes, exactly

josevalim:
Yes, because in the majority of cases not passing the options will lead to the wrong result.

scrogson:
Cool...let's close this then.

PR #1497, opened 2016-01-25
Added a JSON API generator.

eibrahim:
This is almost identical to the JSON generator but uses the JSON API format (which works with Ember's JSONAPIAdapter).  Usage:

`mix phoenix.gen.jsonapi User users email:string encrypted_password:string`


josevalim:
Thank you but I don't believe this should be in Phoenix. I would recommend adding it to projects like JSON API: https://github.com/jeregrine/jsonapi  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

eibrahim:
thanks man.  I will give it a shot.

PR #1502, opened 2016-01-27
Clear channels after disconnect

jfrolich:
If you do not do this, the socket will end up with stale channels, that cause issues when the socket is reconnected.
This fix solved reconnection issues for me, but there might be a better solution.


chrismccord:
Can you describe more what issues you are having? You should not clear this collection on reconnect. The channels on the client will all rejoin automatically after disconnect/reconnects, by design. What problems are you running into?

chrismccord:
@jfrolich if you could provide more info and answers to my questions above, we can revisit. Thanks!

PR #1516, opened 2016-02-02
The LiveReloader and CodeReloader are now decoupled 

caseyhelbling:
There are now two separate config options `code_reloading` and `live_reloading` to be overridden in the Endpoint.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,8 @@
 # Changelog
 
 ## 1.2.0-dev
+* Enhancments
+  * [Endpoint] Live reloader is now broken out to its own config option. Set `live_reloader = true` in your endpoint to turn it on.
 
 ## 1.1.4 (2016-1-25)
 
--- a/installer/templates/new/config/dev.exs
+++ b/installer/templates/new/config/dev.exs
@@ -10,6 +10,7 @@ config :<%= application_name %>, <%= application_module %>.Endpoint,
   http: [port: 4000],
   debug_errors: true,
   code_reloader: true,
+  live_reloader: true,
   check_origin: false,
   watchers: <%= if brunch do %>[node: ["node_modules/brunch/bin/brunch", "watch", "--stdin"]]<% else %>[]<% end %>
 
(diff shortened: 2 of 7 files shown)


josevalim:
Thank you @caseyhelbling!

Unfortunately this introduces duplicated options: `live_reloader` and `live_reload`. What is the use case here? When would you want one but not the other?

chrismccord:
Thank you @caseyhelbling, but the live reloading can be disabled by simply not plugging it. I haven't seen the need for a separate config like José said, so I think sticking with less config options is better. Thanks!

caseyhelbling:
Our team sees code_reload and live_reload as having fundamentally different responsibilities. Specifically, some developers want to make changes that are only picked up when they manually refresh the browser. The front end is in React and the live reloading screws things up inadvertently. Maybe the answer is to just ask them to turn off live reloading in their browsers and not add this complexity to the framework? 

Also, good point about the duplicate options. I was following the `code_reloading` example but I suppose this could be a sub-option under `live_reload`. Maybe something like... 

```
config :ampifiu, Ampifiu.Endpoint,
  live_reload: [
    enabled: true,
    patterns: [
      ~r{priv/static/.*(js|css|png|jpeg|jpg|gif|svg)$},
      ~r{web/views/.*(ex)$},
      ~r{web/templates/.*(eex)$}
    ]
  ]
```

caseyhelbling:
@chrismccord Roger that. Thanks for considering.

chrismccord:
> Specifically, some developers want to make changes that are only picked up when they manually refresh the browser.

In these case, they can adust the live_reload `:patterns` accordingly to include and exclude the files they want to trigger a refresh and the ones they don't?

josevalim:
@caseyhelbling you can also do something like this in your endpoint):

```elixir
if Application.get_env(:my_app, :live_reloading, false) do
  plug LiveReload
end
```

And then configure your files accordingly. You don't need to wait for Phoenix! :D

caseyhelbling:
Cool - we'll give that a try. Thanks!

chrismccord:
@josevalim good point, but the simplest solution instead of the above would be to set `live_reload: [patterns: []]`, or simply delete the `live_reload` config which defaults the patterns to an empty list. In those cases, the plug will noop, no need to introduce extra config for what is already supported.

PR #1530, opened 2016-02-08
Add :sample_binary_id option for generators

### LINKED ISSUE #1527: Generators produce controller tests with ID's which fail against Mongo object ID's.
I asked about this on the mongo_ecto project, but it appears the matter is upstream in the nature of the Phoenix generator. Whatever the case, it causes the generated controller tests from pheonix.gen.html and phoenix.gen.json to fail.

I am using mongo_ecto, which talks to MongoDB and uses ObjectIDs as it ID type. If I run this command: 
```
mix phoenix.gen.html Car cars name:string description:string
```
the generated controller test file will contain this test: 
```
  test "renders page not found when id is nonexistent", %{conn: conn} do
    assert_error_sent 404, fn ->
      get conn, car_path(conn, :show, "11111111-1111-1111-1111-111111111111")
    end
  end
```
This test fails, because the ID string is invalid. If this ID is changed to this: 
```
  test "renders page not found when id is nonexistent", %{conn: conn} do
    assert_error_sent 404, fn ->
      get conn, car_path(conn, :show, "56a537b63e0eee155869a051")
    end
  end
```
Then the test succeeds. In the discussion on the mongo_ecto project, while the issue is specific to MongoDB's ID requirements, the answer I received is that because of the nature of the Phoenix generators, this is not something that can be fixed in the mongo_ecto code, but has to be fixed upstream. 

I confess I don't have the knowledge of the Phoenix or mongo_ecto codebase to speak further on the matter, but if someone can point me to the place in the Phoenix source which is in play here, I'd be happy to take a look and brainstorm on a solution. At any rate, it does constitute a bug, and some provision should be made to either fix it outright in Phoenix, or to expose functionality to allow it to be fixed downstream in mongo_ecto.

michalmuskala:
Closes  #1527

### DIFF
--- a/installer/lib/phoenix_new.ex
+++ b/installer/lib/phoenix_new.ex
@@ -264,11 +264,15 @@ defmodule Mix.Tasks.Phoenix.New do
       append_to path, "config/config.exs", """
 
       # Configure phoenix generators
-      config :phoenix, :generators,
-        migration: #{adapter_config[:migration]},
-        binary_id: #{adapter_config[:binary_id]}
+      config :phoenix, :generators#{kw_to_config(generator_config(adapter_config))}
       """
     end
+  end
+
+  defp generator_config(adapter_config) do
+    adapter_config
+    |> Keyword.take([:binary_id, :migration, :sample_binary_id])
+    |> Enum.filter(fn {_, value} -> not is_nil(value) end)
   end
 
   defp copy_static(_app, path, binding) do
@@ -418,7 +422,8 @@ defmodule Mix.Tasks.Phoenix.New do
      test_begin: "",
      test_restart: "Mongo.Ecto.truncate(#{module}.Repo, [])",
      binary_id: true,
-     migration: false}
+     migration: false,
+     sample_binary_id: "111111111111111111111111"}
   end
   defp get_ecto_adapter(db, _app, _mod) do
     Mix.raise "Unknown database #{inspect db}"
--- a/priv/templates/phoenix.gen.json/controller_test.exs
+++ b/priv/templates/phoenix.gen.json/controller_test.exs
@@ -21,9 +21,9 @@ defmodule <%= module %>ControllerTest do
       "<%= k %>" => <%= singular %>.<%= k %><% end %>}
   end
 
-  test "does not show resource and instead throw error when id is nonexistent", %{conn: conn} do
+  test "renders page not found when id is nonexistent", %{conn: conn} do
     assert_error_sent 404, fn ->
-      get conn, <%= singular %>_path(conn, :show, <%= if binary_id do %>"11111111-1111-1111-1111-111111111111"<% else %>-1<% end %>)
+      get conn, <%= singular %>_path(conn, :show, <%= inspect sample_id %>)
     end
   end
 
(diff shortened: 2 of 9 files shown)


installer/lib/phoenix_new.ex
```diff
@@ -408,7 +409,8 @@ defmodule Mix.Tasks.Phoenix.New do
       prod: [database: "db/#{app}_prod.sqlite"],
       test_begin: "Ecto.Adapters.SQL.begin_test_transaction(#{module}.Repo)",
       test_restart: "Ecto.Adapters.SQL.restart_test_transaction(#{module}.Repo, [])",
-      migration: true}
+      migration: true,
+      sample_binary_id: "11111111-1111-1111-1111-111111111111"}
```
josevalim: Let's not add those. If there is no sample_binary_id, we can't default to the one above.

priv/templates/phoenix.gen.json/controller_test.exs
```diff
@@ -21,9 +21,9 @@ defmodule <%= module %>ControllerTest do
       "<%= k %>" => <%= singular %>.<%= k %><% end %>}
   end
 
-  test "does not show resource and instead throw error when id is nonexistent", %{conn: conn} do
+  test "renders page not found when id is nonexistent", %{conn: conn} do
```
josevalim: "raises with page not found" or "errors with page not found"?

test/mix/tasks/phoenix.gen.html_test.exs
```diff
@@ -191,6 +194,19 @@ defmodule Mix.Tasks.Phoenix.Gen.HtmlTest do
...
+        Mix.Tasks.Phoenix.Gen.Json.run ["User", "users"]
+
+        assert_file "test/controllers/user_controller_test.exs", fn file ->
+          assert file =~ ~S|user_path(conn, :show, "abcd")|
+        end
+      end
+    end
+  end
+
+
```
josevalim: :scissors:

josevalim:

priv/templates/phoenix.gen.json/controller_test.exs
michalmuskala: I made the message exactly the same as in `phoenix.gen.html` for this test. I can change it in the other place, or in both to something else.

josevalim: Ok, let's leave both as is then. :)

chrismccord:
Can we get a a tiny doc in the phoenix_new moduledocs? Maybe even a "Third-pary Extension" section. Ready to go after that!

josevalim:
Ping! :D

installer/lib/phoenix_new.ex
michalmuskala: This general config is used for mysql, postgresql and mssql. Should I give this sample id explicitly in each adapter?

michalmuskala:
@chrismccord I'm not sure phoenix_new is the best place to document those, because it's not configurable by the end user. On the other hand I can't find a plcace to document all the generator config values for phoenix.

installer/lib/phoenix_new.ex
josevalim: No. We should only specify if if binary_id is given and it does not use UUID in its implementation.

michalmuskala:
Updated

josevalim:

Issue #1521, opened 2016-02-04
Controller.accepts handles errors by itself


yolk:
Since this [commit](https://github.com/phoenixframework/phoenix/commit/aea71ff031c9bcaec7ceabb9a17b6a6a826bd8cc) Controller.accepts doesn't raise an Phoenix.NotAcceptableError expection anymore, it rather reponds with an 406 and halts the connection.

Is there any reason for handling it this way? Because this behaviour isn't consistent with the way exception are handled elsewhere (e.g. Plug.Parsers). This prevents me from handling the error with render_errors and the ErrorView. Shouldn't there be an consistent way to handle errors/exceptions?

Btw. The docs still mention the raised exception on [line 81](https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/controller.ex#L81)

josevalim:
I have updated the docs for now. I think we can revert it back to raising as long as we handle it in the render_errors layer.

yolk:
Thanks for the quick reply and fix in the docs. Should I revert the changes in a pull request?

josevalim:
Please do. :)

yolk:
Any idea how to handle the case of an invalid format in the request for the error rendering? Should I just catch the raised exception in RenderErrors.fetch_format/2 and rerender the error with the first accepted format? 

Right now it renders a plain 406 if you throw an error and your request accepts no mime types accepted by RenderErrors: https://github.com/phoenixframework/phoenix/blob/master/test/phoenix/endpoint/render_errors_test.exs#L167

I guess this actually hides the original error (a 500 in this case) behind the 406 and would only make sense if the 406 orginated from the original requested resource. "unknown/unkown" would always raise an NotAcceptableError and this situation wouldn't accure. A better example would be an request for an pdf or rss/atom feed, which is an acceptable mime type for the original resource, but in most cases not for rendering an error.

josevalim:
What does the HTTP spec says about 500s? Should they be parseable according to the accepts header? If so, we have no other option than return 406.

yolk:
I couldn't find anything about how to handle a error case combined with incompatible accept header; but if I understand RFC 2616 correctly it seems to be acceptable to serve another content type in case nothing matches the accept header:

> Note: HTTP/1.1 servers are allowed to return responses which are
      not acceptable according to the accept headers sent in the
      request. In some cases, this may even be preferable to sending a
      406 response. User agents are encouraged to inspect the headers of
[...]

PR #1533, opened 2016-02-09
Raise not acceptable error & fall back to first acceptable format on error

yolk:
As discussed in #1521:

1. Raise Phoenix.NotAcceptableError in Phoenix.Controller.accepts instead of halting the connection with a 406 status
2. Rescue Phoenix.NotAcceptableError in RenderErrors.fetch_format/2 and render original error with first acceptable format.

I am not sure if this is the best place to rescue the error. And: is the halted:true-line at https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/endpoint/render_errors.ex#L69 still needed?


josevalim:

PR #1535, opened 2016-02-09
Phoenix.NotAcceptableError contains accepts field

yolk:
This is useful to render a helpful error entity for the client which includes
the acceptable mime types.

RFC 2616: "Unless it was a HEAD request, the response SHOULD include an entity containing a list of available entity characteristics and location(s) from which the user or user agent can choose the one most appropriate."

Follow up from #1521


josevalim:

Issue #1537, opened 2016-02-09
No way to filter params in Channel logs


stevedomin:
Currently this is no way of filtering params in Channel logs. It would be useful to reuse the `filter_params` config to filter out sensitive information in the channel logs.

This came through the mailing list: https://groups.google.com/forum/#!topic/phoenix-talk/EZFmddtju9o

stevedomin:
I'll start working on it.

chrismccord:
:+1:  thanks!

stevedomin:
This issue can be closed now that #1539 has been merged

PR #1539, opened 2016-02-10
Filter payload on channel join

stevedomin:
I'm not entirely happy with this PR but I thought I'd gather feedback before going any further:

* Although I think we should be DRY and not duplicate the `filter_value/2` in two different places `Phoenix.Logger.Filter` doesn't feel like the right module name for that. Any suggestions?
* I kept the tests for filtering in `test/phoenix/controller/logger_test.exs`, although they are a bit redundant with the tests I have in `test/phoenix/logger/filter_test.exs` I still feels better having them here.
* That being said I should probably have equivalent tests for `lib/phoenix/socket/transport.ex`

Let me know your thoughts and I'll fix the PR.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -4,4 +4,6 @@ use Mix.Config
 config :logger, :console, colors: [enabled: false]
 
 # Use higher stacktrace depth.
-config :phoenix, :stacktrace_depth, 20
+config :phoenix,
+  stacktrace_depth: 20,
+  filter_parameters: ["password"]
--- a/lib/phoenix/controller/logger.ex
+++ b/lib/phoenix/controller/logger.ex
@@ -4,6 +4,10 @@ defmodule Phoenix.Controller.Logger do
   @behaviour Plug
   import Phoenix.Controller
 
+  alias Phoenix.Endpoint.Instrument
+
+  @filter_parameters Application.get_env(:phoenix, :filter_parameters)
+
   @moduledoc """
   Plug to handle request logging at the controller level.
 
@@ -46,24 +50,7 @@ defmodule Phoenix.Controller.Logger do
   defp params(%Plug.Conn.Unfetched{}), do: "[UNFETCHED]"
   defp params(params) do
     params
-    |> filter_values(Application.get_env(:phoenix, :filter_parameters))
+    |> Instrument.filter_values(@filter_parameters)
     |> inspect()
   end
-
-  defp filter_values(%{__struct__: mod} = struct, _filter_params) when is_atom(mod) do
-    struct
-  end
-  defp filter_values(%{} = map, filter_params) do
-    Enum.into map, %{}, fn {k, v} ->
-      if is_binary(k) && String.contains?(k, filter_params) do
-        {k, "[FILTERED]"}
-      else
-        {k, filter_values(v, filter_params)}
-      end
-    end
-  end
-  defp filter_values([_|_] = list, filter_params) do
-    Enum.map(list, &filter_values(&1, filter_params))
-  end
-  defp filter_values(other, _filter_params), do: other
 end
--- a/lib/phoenix/socket/transport.ex
+++ b/lib/phoenix/socket/transport.ex
@@ -102,12 +102,14 @@ defmodule Phoenix.Socket.Transport do
   """
 
   require Logger
+  alias Phoenix.Endpoint.Instrument
   alias Phoenix.Socket
   alias Phoenix.Socket.Message
   alias Phoenix.Socket.Reply
 
   @protocol_version "1.0.0"
   @client_vsn_requirements "~> 1.0"
+  @filter_parameters Application.get_env(:phoenix, :filter_parameters)
 
   @doc """
   Provides a keyword list of default configuration for socket transports.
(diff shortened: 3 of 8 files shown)


stevedomin:
This fixes https://github.com/phoenixframework/phoenix/issues/1537

josevalim:
I would move them to Phoenix.Endpoint.Instrumenter. Not ideal but it is the least weird place. Also, we will make the controller logger an instrumenter event at some point before 1.2.0, so it fits a tiny bit. :) Thank you! :heart:

stevedomin:
@josevalim I've pushed a new version taking into account your comment. It feels a bit weird to have that here but I trust your judgement anyway.

What about the testing situation, do you want me to add more tests?

josevalim:
I think you can simplify the tests for the controller logger and add at least one test for the transport one. :) Basically, we just need a single test for each case to guarantee we are filtering.

stevedomin:
@josevalim removed some of the tests in `test/phoenix/controller/logger_test.exs` but I can't figure out a nice way to do the transport test. I tried to follow the advice you gave me on IRC but I didn't get anywhere. Doing it in `test/phoenix/test/channel_test.exs` doesn't really help because `ChannelTest.join/3` doesn't hit the transport layer.

josevalim:
@stevedomin one last attempt is to write a test here: https://github.com/phoenixframework/phoenix/blob/master/test/phoenix/integration/websocket_test.exs

This is an integration test, so the transport layer is definitely invoked. Indeed, my suggestion wouldn't invoke the transport code, sorry about that.

stevedomin:
Thought about that, shall I add one in both the long-polling and websocket test?

josevalim:
Yes please.

stevedomin:
@josevalim ready for another review

lib/phoenix/socket/transport.ex
```diff
@@ -201,10 +202,11 @@ defmodule Phoenix.Socket.Transport do
     if channel = socket.handler.__channel__(topic, socket.transport_name) do
       socket = %Socket{socket | topic: topic, channel: channel}
 
+      filtered_payload = Instrument.filter_values(msg.payload, Application.get_env(:phoenix, :filter_parameters))
```
chrismccord: Should we bake this into a module attribute? We can document recompile instructions and it's something that shouldn't change often.

stevedomin: `@filter_parameters` in the `Instrument` module?

chrismccord: You would bake it into the transport module and logger module

stevedomin: Would it be a bad pattern having the attribute declared in the `Instrument` module and then calling `Instrument.filter_values(msg.payload)`?

chrismccord: it locks the api to a single filter_params list, so passing it in as an argument is preferred

config/config.exs
```diff
@@ -4,4 +4,6 @@ use Mix.Config
 config :logger, :console, colors: [enabled: false]
 
 # Use higher stacktrace depth.
-config :phoenix, :stacktrace_depth, 20
\ No newline at end of file
+config :phoenix,
+  stacktrace_depth: 20,
+  filter_parameters: ["password"]
```
stevedomin: I had to add this to get the test to pass, although it's already defined here: https://github.com/phoenixframework/phoenix/blob/master/mix.exs

What's the reason for that? I'm unclear on the relation between the two

stevedomin:
I've updated my PR to use a module attribute.

I'm not sure if it's related to my changes but I'm seeing a few failures the first time I run the test. Here's the log: https://gist.github.com/stevedomin/5b1b8ec3fa350f1ed009

config/config.exs
josevalim: That configuration won't be available until after the app is compiled, so it makes sense.

lib/phoenix/controller/logger.ex
```diff
@@ -4,6 +4,10 @@ defmodule Phoenix.Controller.Logger do
   @behaviour Plug
   import Phoenix.Controller
 
+  alias Phoenix.Endpoint.Instrument
+
+  @filter_parameters Application.get_env(:phoenix, :filter_parameters)
```
josevalim: @chrismccord the downside of this solution is that, if you change the filter_parameters configuration, the option won't propagate unless you recompile all of Phoenix. So I would honestly move away from the module attribute because I don't think this cons > pros. But we already gave @stevedomin enough work, i will fix it after merging. :)

josevalim:

lib/phoenix/controller/logger.ex
stevedomin: I wouldn't have mind fixing it but thanks for doing it @josevalim!

I have two questions:

1. Is there a significant performance penalty with calling `Application.get_env/2` each time compared to using a module attribute?

2. While working on this PR I kept wondering if Phoenix should somehow wrap `Logger` with its own module so that we can ensure no logs are left unfiltered. I can see some downsides with this approach but I was curious what your opinion was.

josevalim: 1. No, it is a quick ETS lookup

2. I think we are fine because we call it from two places only. If more start popping up, then we should review it.

PR #1543, opened 2016-02-11
refactor(Phoenix): use Elixir 1.2 alias syntax

Gazler:
In places where it is more readable to do so, the elixir 1.2 alias
syntax is used instead of spreading the alias over multiple lines.

### DIFF
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -172,8 +172,7 @@ defmodule Phoenix.Channel do
   is invoked. This practice is not encouraged though.
   """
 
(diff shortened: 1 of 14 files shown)


josevalim:
Looks good but it will break PRs. :( In particular, sasa is working on transport changes and this would break it. I would say we should fix it as we go through the code.

Gazler:
@josevalim I didn't think it would have too much of an impact, however, I tried Sasa's branch and there were conflicts.

Since this is only a refactor, I'd rather not burden contributors with merge conflicts, so I have closed the PR.

PR #1546, opened 2016-02-13
Enable {:system, "ENV_VAR"} on :url and :static_url endpoint config options

zberkom:
The :port endpoint config allows you to pass in a ENV variable as
{:system, "ENV_VAR"}, but the configuration of :url and :static_url
require that you use System.get_env("ENV_VAR"), which isn't consistent.

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -119,8 +119,9 @@ defmodule Phoenix.Endpoint do
       task automatically sets this to `true`.
 
     * `:url` - configuration for generating URLs throughout the app.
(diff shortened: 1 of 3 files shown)


chrismccord:
Thanks! Unfortunately we are waiting on a comprehensive solution around these kinds of problems with env specific config and releases. Port was added as an edge case, but I would prefer to wait and see how mix and exrm handle this scenario before going this route. Thanks!

zberkom:
@chrismccord No problem, thanks for considering the PR, looking forward to the solution mix and exrm come up with.

PR #1552, opened 2016-02-16
Move brunch packages to `devDependencies`.

lucasmazza:
Brunch related dependencies are meant for app development only, and we can leave
the `dependencies` group for runtime deps like `phoenix_html` and other packages
added by developers during development.


josevalim:

PR #1553, opened 2016-02-16
Add a disclaimer to static/app.css for new project

stevedomin:
Several people have been confused by the inclusion of Bootstrap and
a few CSS rule-set in static/app.css when creating new project. This
commit addresses that by expliciting stating that the content of this
file is only here for convenience and can be safely deleted.


josevalim:

PR #1555, opened 2016-02-16
Add Presence generator. Closes #1523

### LINKED ISSUE #1523: Add presence generator

chrismccord:
Ready for review. I wasn't sure how much docs to include on the generated Presence module. Too much?

### DIFF
--- a/lib/mix/tasks/phoenix.gen.presence.ex
+++ b/lib/mix/tasks/phoenix.gen.presence.ex
@@ -1,0 +1,60 @@
+defmodule Mix.Tasks.Phoenix.Gen.Presence do
+  use Mix.Task
+
+  @shortdoc "Generates a Presence tracker"
+
+  @moduledoc """
+  Generates a Presence tracker for your application.
+
+      mix phoenix.gen.presence
+
+      mix phoenix.gen.presence MyPresence
+
+  The only argument is the module name of the Presence tracker,
+  which defaults to Presence.
+
+  A new file will be generated in:
+
+    * web/channels/presence.ex
+
+  Where `presence.ex` is the snake cased version of the module name provided.
+  """
+  def run([]) do
+    run(["Presence"])
+  end
+  def run([alias_name]) do
+    binding = Mix.Phoenix.inflect(alias_name) ++ [otp_app: Mix.Phoenix.otp_app()]
+    files = [
+      {:eex, "presence.ex", "web/channels/#{binding[:path]}.ex"},
+    ]
+    Mix.Phoenix.copy_from paths(), "priv/templates/phoenix.gen.presence", "", binding, files
+
+    Mix.shell.info """
+
+    *Required* post-install setup:
+
+
+    First, add configuration for your #{binding[:module]} tracker,
+    in config/config.exs:
+
+        config :my_app, #{binding[:module]},
+          pubsub_server: #{inspect Module.concat(binding[:base], PubSub)}
+
+
+    Next, add your new module to your supervision tree,
+    in lib/#{binding[:otp_app]}.ex:
+
+        children = [
+          ...
+          supervisor(#{binding[:module]}, []),
+        ]
+
+    You're all set! See the Phoenix.Presence docs for more details:
+    http://hexdocs.pm/phoenix/Phoenix.Presence.html
+    """
+  end
+
+  defp paths do
+    [".", :phoenix]
+  end
+end
--- a/priv/templates/phoenix.gen.presence/presence.ex
+++ b/priv/templates/phoenix.gen.presence/presence.ex
@@ -1,0 +1,66 @@
+defmodule <%= module %> do
+  @moduledoc """
+  Provides presence tracking to channels and processes.
+
+  See the [`Phoenix.Presence`](http://hexdocs.pm/phoenix/Phoenix.Presence.html)
+  docs for more details.
+
+  ## Usage
+
+  Presences can be tracked in your channel after joining:
+
+      defmodule <%= base %>.MyChannel do
+        alias <%= base %>.Presence
+        ...
+        def join("some:topic", _params, socket) do
+          send(self, :after_join)
+          {:ok, assign(socket, :user_id, ...)}
+        end
+
+        def handle_info(:after_join, socket) do
+          :ok = Presence.track(socket, socket.assigns.user_id, %{
+            online_at: inspect(:os.timestamp())
+          })
+          push socket, "presences", Presence.list(socket)
+          {:noreply, state}
+        end
+      end
+
+  In the example above, `Presence.track` is used to register this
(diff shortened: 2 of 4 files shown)


lib/mix/tasks/phoenix.gen.presence.ex
```diff
@@ -0,0 +1,60 @@
...
+
+    Mix.shell.info """
+
+    *Required* post-install setup:
+
+
+    First, add configuration for your #{binding[:module]} tracker,
+    in config/config.exs:
+
+        config :my_app, #{binding[:module]},
```
michalmuskala: I think we could provide real name of the application here

lib/mix/tasks/phoenix.gen.presence.ex
```diff
@@ -0,0 +1,60 @@
...
+  def run([alias_name]) do
+    binding = Mix.Phoenix.inflect(alias_name) ++ [otp_app: Mix.Phoenix.otp_app()]
+    files = [
+      {:eex, "presence.ex", "web/channels/#{binding[:path]}.ex"},
+    ]
+    Mix.Phoenix.copy_from paths(), "priv/templates/phoenix.gen.presence", "", binding, files
+
+    Mix.shell.info """
+
+    *Required* post-install setup:
```
josevalim: We should probably follow the same format as in `phoenix.gen.html`.

lib/mix/tasks/phoenix.gen.presence.ex
```diff
@@ -0,0 +1,60 @@
...
+
+
+    First, add configuration for your #{binding[:module]} tracker,
+    in config/config.exs:
+
+        config :my_app, #{binding[:module]},
+          pubsub_server: #{inspect Module.concat(binding[:base], PubSub)}
```
josevalim: We can actually find out exactly what this is in the app by reading the config. If there is no name configured, we should probably not even generate the presence fule. Also, can we generate this config inside the presence stuff?

    use Phoenix.Presence, pubsub_server: Foo.Bar

It would be one less step.

priv/templates/phoenix.gen.presence/presence.ex
```diff
@@ -0,0 +1,66 @@
...
+        def join("some:topic", _params, socket) do
+          send(self, :after_join)
+          {:ok, assign(socket, :user_id, ...)}
+        end
+
+        def handle_info(:after_join, socket) do
+          :ok = Presence.track(socket, socket.assigns.user_id, %{
+            online_at: inspect(:os.timestamp())
+          })
+          push socket, "presences", Presence.list(socket)
```
josevalim: What if we call it presence_state to mirror the presence_diff?

priv/templates/phoenix.gen.presence/presence.ex
```diff
@@ -0,0 +1,66 @@
...
+
+  See the [`Phoenix.Presence`](http://hexdocs.pm/phoenix/Phoenix.Presence.html)
+  docs for more details.
+
+  ## Usage
+
+  Presences can be tracked in your channel after joining:
+
+      defmodule <%= base %>.MyChannel do
+        alias <%= base %>.Presence
```
josevalim: Let's include `use <%= base %>.Web, :channel`, the alias and get rid of the `...` below.

priv/templates/phoenix.gen.presence/presence.ex
```diff
@@ -0,0 +1,66 @@
...
+  In the example above, `Presence.track` is used to register this
+  channel's process as a presence for the socket's user ID, with
+  a map of metadata. Next, the current presence list for
+  the socket's topic is pushed to the client as a `"presences"` event.
+
+  Finally, a diff of presence join and leave events will be sent to the
+  client as they happen in real-time with the "presence_diff" event.
+  See `Phoenix.Presence.list/2` for details on the presence datastructure.
+
+
```
josevalim: :scissors:

priv/templates/phoenix.gen.presence/presence.ex
```diff
@@ -0,0 +1,66 @@
...
+  ## Fetching Presence Information
+
+  Presence metadata should be minimized and used for ephemeral state
+  like a users "online" or "away" status. Somtimes you'll want to
+  extend the metadata with information from the database, such as a user's
+  name or proifle information. You can do this by overriding `fetch/2`:
```
josevalim: Typo proifle.

priv/templates/phoenix.gen.presence/presence.ex
```diff
@@ -0,0 +1,66 @@
...
+            where: u.id in ^Map.keys(entries),
+            select: {u.id, u}
+
+        users = query |> Repo.all |> Enum.into(%{})
+
+        for {key, %{metas: metas}} <- entries, into: %{} do
+          {key, %{metas: metas, user: users[key]}}
+        end
+      end
+
```
josevalim: Let's add a quick summary about what this function is supposed to receive and what it is supposed to return. Then we explain how it works.

priv/templates/phoenix.gen.presence/presence.ex
chrismccord: :+1:

priv/templates/phoenix.gen.presence/presence.ex
chrismccord: We can't use web :channel because it uses Phoenix.Channel

josevalim: @chrismccord exactly. this code *is* defining a channel :P

chrismccord: derp, I was thinking you wanted the presence module to use web :channel. :+1:

princemaple:
Is Presence already in phoenix.js but not in Phoenix? I recalling seeing something like that in the source code.

priv/templates/phoenix.gen.presence/presence.ex
```diff
@@ -0,0 +1,66 @@
+defmodule <%= module %> do
+  @moduledoc """
+  Provides presence tracking to channels and processes.
```
allyraza: should read as "Provides presence tracking for channels and processes."

priv/templates/phoenix.gen.presence/presence.ex
```diff
@@ -0,0 +1,66 @@
...
+
+  See the [`Phoenix.Presence`](http://hexdocs.pm/phoenix/Phoenix.Presence.html)
+  docs for more details.
+
+  ## Usage
+
+  Presences can be tracked in your channel after joining:
```
allyraza: typo Presences

allyraza:
maybe we should also mention that user need to add phoenix_presence in their mix.exs deps or are we going to to include that in default app template

PR #1559, opened 2016-02-18
Added an issue template to help users open helpful bug reports

jerel:
Since I watch this repo I've noticed that a large number of issues that are opened are support requests so I decided to submit this issue template for your consideration. You can see the rendered version here: https://github.com/jerel/phoenix/issues/new (click Preview)

### DIFF
--- a/ISSUE_TEMPLATE.md
+++ b/ISSUE_TEMPLATE.md
@@ -1,0 +1,19 @@
+### Precheck
+
+* For help and support, use Stack Overflow or [phoenix-talk](https://groups.google.com/forum/#!forum/phoenix-talk).
+* For proposing a new feature, please start a discussion on [phoenix-core](https://groups.google.com/forum/#!forum/phoenix-core).
+* For bugs, do a quick search and make sure the bug has not yet been reported.
+* All checked? Be nice and have fun!
+
+### Environment
+
+* Elixir version (elixir -v):
+* Phoenix version (mix deps):
+* NodeJS version (node -v):
+* NPM version (npm -v):
+* Operating system:
+
+### Expected behavior
+
+
+### Actual behavior


ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,17 @@
...
+
+### Expected behavior
+
+
+### Actual behavior
+
+
+* [x] I understand that support requests should go to the [phoenix-talk](https://groups.google.com/forum/#!forum/phoenix-talk) mailing list and that this issue tracker is for bugs only.
+
+
```
Gazler: Maybe snip these new lines at the end?

ISSUE_TEMPLATE.md
```diff
@@ -0,0 +1,17 @@
+### Environment
+
+* Elixir version (elixir -v):
+* Phoenix version:
```
Gazler: `mix deps` can be used to list dependencies.

Gazler:
:+1: This should help cut down on the number of support issues raised.

Tica2:
:+1: This is cool!
And if we have mix task that provided this output?
I think you forgot `erl -version` 

In this way we can include more details later if we need (with parameters or not)
We do not need to ask again and again ...
Just write in this file to include this output ... :)

Tica2:
@jerel Include this command in file: 
`mix hex.info`

In my case is:
```
Hex:    0.10.3
Elixir: 1.2.1
OTP:    18.1

Built with: Elixir 1.2.1 and OTP 18.2.2

A new Hex version is available (0.10.4), please update with `mix local.hex`
Registry file available (last updated: 2016-02-18 17:09:01)
Size: 1461kB (compressed 474kb)
Packages #: 1545
Versions #: 6876
```

stevedomin:
I feel like the volume of new issue on the Phoenix repo is still low enough that it doesn't justify having that template.

This kind of information would be useful for a lot of questions on the mailing list but not here, and although I agree there are still a lot of people raising support issue here but I don't think that's the solution.

josevalim:
I am :+1: on this. I think the versions we ask in the template above are fine, we can always ask the erlang info if needed. I would just add this to the top:

```
### Precheck

* For help and support, use [phoenix-talk](https://groups.google.com/forum/#!forum/phoenix-talk) or Stack Overflow. 
* For proposing a new feature, please start a discussion on the [phoenix-core](https://groups.google.com/forum/#!forum/phoenix-core) mailing list.
* For bugs, do a quick search and make sure the bug has not yet been reported.
* All checked? Be nice and have fun!
```

I am afraid it may be too much but I think it can be a good starting point. Thoughts?

chrismccord:
@josevalim :+1: on all points. I was worried it might be too much, but we've been getting more and more noise so let's try it out and adjust as we go. @stevedomin can you make José's changes and I will merge? Thanks!

chrismccord:
s/ @stevedomin / @jerel

josevalim:
In fact I am planning to do the same to Elixir and get rid of the Contributing.md file (which unfortunately nobody reads).

jerel:
@chrismccord ok, ready for review. I also pulled the last line (the checkbox) since it was basically an echo of José's first line.

chrismccord:
<3 <3 <3

michalmuskala:
I would consider keeping the Contributing file. Some do read it, and it gives much more information than the issue template.

lexmag:
Maybe we should move this file (also `CONTRIBUTING.md`?) into `.github/` folder?

As the GitHub blog says:
> If you're worried about the added clutter in the root directory of your project, we also added support for a `.github/` folder. You can put `CONTRIBUTING.md`, `ISSUE_TEMPLATE.md`, and `PULL_REQUEST_TEMPLATE.md` files in `.github/` and everything will work as expected.

lexmag:
Also I'm :+1: for keeping `CONTRIBUTING.md` file.

PR #1567, opened 2016-02-23
Format the lists in the mix file

drapergeek:
This is one of the first changes I make when setting up a new phoenix project. In particular, it's helpful when adding a new dependency to be able to add new entry on a newline. It makes for a really clean read on a PR to see what has changed.

### DIFF
--- a/installer/templates/new/mix.exs
+++ b/installer/templates/new/mix.exs
@@ -2,19 +2,21 @@ defmodule <%= application_module %>.Mixfile do
   use Mix.Project
 
   def project do
-    [app: :<%= application_name %>,
-     version: "0.0.1",<%= if in_umbrella do %>
-     build_path: "../../_build",
-     config_path: "../../config/config.exs",
-     deps_path: "../../deps",
-     lockfile: "../../mix.lock",<% end %>
-     elixir: "~> 1.2",
-     elixirc_paths: elixirc_paths(Mix.env),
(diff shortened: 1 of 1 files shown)


jerodsanto:
:+1: I do this every time as well.

paulcsmith:
I would also love to see this. Makes it easier to sort dependencies and applications and makes diffs super clear.

josevalim:
The `[foo: ...]` is the style "promoted" by Elixir and it is the one you will see on inspect, getting started guides, etc. We don't have plans for changing it in Phoenix as long as Elixir plans to stay the same.

josevalim:
To be clear, two things:

1. I am personally OK with breaking the applications list into one entry per line. I believe that would help with the diffs @paulcsmith mentioned.

2. Regarding the style, my advice would be to try to get used with how Elixir styles data structures. I am aware it may be unconventional to some folks depending on their background, but Elixir is more data-driven than many other languages, which means having the `{`, `[` and so on in their respective line is too verbose and hard to follow.

Indenting according to the container is what makes most logical sense because that's exactly how we indent long function calls:

```elixir
[app: :foo,
 version: "0.0.1"]

# is equivalent to

foobar(baz, bat,
       boz, bot)
```

Such indentation style is common in other communities, Clojure is the most recent one that comes to mind.

heavyk:
i often times write my elixir code with elixir formatting especially when it's an argument to a function (easier to read), hewever I prefer my list of deps to be one per line too.

PR #1568, opened 2016-02-26
Adding HAML mode and templates for phoenix.gen.html

parndt:
### What this does
This is my first pass at adding HAML template support for the `phoenix.gen.html` generator.  There are a few issues, detailed below the example.

### Example

```
$ mix phoenix.gen.html MyHamlThing my_haml_things name:string --template=haml
* creating web/controllers/my_haml_thing_controller.ex
* creating test/controllers/my_haml_thing_controller_test.exs
* creating web/views/my_haml_thing_view.ex
* creating web/templates/my_haml_thing/edit.html.haml
* creating web/templates/my_haml_thing/form.html.haml
* creating web/templates/my_haml_thing/index.html.haml
* creating web/templates/my_haml_thing/new.html.haml
* creating web/templates/my_haml_thing/show.html.haml
* creating web/models/my_haml_thing.ex
* creating test/models/my_haml_thing_test.exs
* creating priv/repo/migrations/20160226220620_create_my_haml_thing.exs

Add the resource to your browser scope in web/router.ex:

    resources "/my_haml_things", MyHamlThingController

Remember to update your repository by running migrations:

    $ mix ecto.migrate

```

### Known issues
- [ ] All the helpers produce eex, for example https://github.com/phoenixframework/phoenix/blob/04464429d9b958e331b2ffe0f0f5926690ab3b56/lib/mix/tasks/phoenix.gen.html.ex#L106-L139 will produce, in `form.html.haml`, the following:

```haml
= form_for @changeset, @action, fn f ->
  = if @changeset.action do
    .alert.alert-danger
      %p
        Oops, something went wrong! Please check the errors below.

  .form-group
    <%= label f, :name, class: "control-label" %>
    <%= text_input f, :name, class: "form-control" %>
    <%= error_tag f, :name %>

  .form-group
    = submit "Submit", class: "btn btn-primary"
```

I wasn't sure of the best way to fix this as the eex output is all hardcoded in there.  Would love some suggestions/feedback please!

- [ ] No tests for the template switch yet


priv/templates/phoenix.gen.html/haml/edit.html.haml
```diff
@@ -0,0 +1,7 @@
+%h2
+  Edit <%= template_singular %>
```
knewter: can you just do `#{template_singular}` here instead of the eex?.  Not sure about that in this haml implementation...

priv/templates/phoenix.gen.html/haml/edit.html.haml
```diff
@@ -0,0 +1,7 @@
...
+
+= render "form.html", changeset: @changeset,
+                      action: <%= singular %>_path(@conn, :update, @<%= singular %>)
```
knewter: ditto the above, and I'll stop repeating this question.

priv/templates/phoenix.gen.html/haml/form.html.haml
```diff
@@ -0,0 +1,13 @@
+= form_for @changeset, @action, fn f ->
+  = if @changeset.action do
+    .alert.alert-danger
+      %p
+        Oops, something went wrong! Please check the errors below.
+<%= for {label, input, error} <- inputs, input do %>
```
knewter: why the eex brackets here?

knewter:
@parndt I really like this :)  Made a few notes but they apply to various places in the commit re eex brackets

knewter:
woopsie-daisy, it's been explained to me.  removing the inline comments now...

priv/templates/phoenix.gen.html/haml/new.html.haml
```diff
@@ -0,0 +1,7 @@
+%h2
+	New <%= template_singular %>
```
knewter: tab?

doomspork:
It'd be great if Phoenix could offload this onto the template library somehow; We recently introduced similar generators in [Phoenix Slime](https://github.com/slime-lang/phoenix_slime) for Slim/Slime template support.

priv/templates/phoenix.gen.html/haml/new.html.haml
parndt: good catch - my editor disobeyed me!

parndt:
@doomspork I also wondered about creating `$ mix phoenix.gen.html.haml` but I figured that users would rather just remember the one mix command.  Perhaps this is a flawed assumption!

parndt:
@doomspork I also didn't want to have to reimplement all of the functionality that is the same between the generators (i.e. the markup independent stuff)

josevalim:
Thank you @parndt.

The reason I am :-1: on this patch is because all of those variable bindings are completely internal to Phoenix. I am not comfortable in exposing then in any way because there were not designed to be exposed but rather to be specific to EEx.

I do like the suggestion of calling something like "phoenix.gen.html.#{template}", passing all args. If no template is given, we just generate the eex stuff we have right now (i.e. no need for phoenix.gen.html.eex).

Thoughts?

chrismccord:
The only way we could support something like this is if the generator is not template-engine aware, ie we can't check in .haml or .slim files into Phoenix. I'm not sure how we'd accomplish this in a way that is better than having each lib simply provide their own generators, but I am open to ideas.

parndt:
@josevalim I like the compromise you suggest and it lines up with what @doomspork is saying too.  

Should the generator live inside phoenix, or some other library?  If another library, it could just import Mix.Tasks.Phoenix.Gen.Html and overload the functions that had to change i.e. the problematic helpers?

chrismccord:
The issue is also the tests that are generated, or anything we provide in the future. I'm hesitant to go that route because then we have a versioned API that we are coupled to with other engines, ie if we add `notifications.html.eex` to the generator, or simply change the output, it would add eex files to projects and break tests. Given how small the html generator is, my preference right now is to have each lib roll and maintain their generators and @doomspork has done. They can even use the model generators and such.

parndt:
@chrismccord would you be happy for this to live in chrismccord/phoenix_haml ?

chrismccord:
Yes, if someone other than me would like to maintain it :)

parndt:
@chrismccord great. I will regroup and target my PR there instead.

Thanks everyone! :heart: :blue_heart: :purple_heart:

parndt:
I've opened chrismccord/phoenix_haml#32

Issue #1564, opened 2016-02-22
Migrate to Ecto 2.0


josevalim:
* [x] Remove `Mix.Task.run` from test_helper.exs and move it to mix.exs aliases
* [x] Replace test_transaction by Ecto sandbox (async: false should set shared mode)
* [x] Use `%{}` instead of `:empty` in generators (but only if Ecto version is 2.0.0)
* [x] Use `cast/3 + validate_required` instead of `cast/4` and get rid of scrub_params (but only if Ecto version is 2.0.0)
* [x] No pool_size: 1 limit for test (Ecto should warn)
* [x] Update gettext messages (length should no longer use count) (see #1589)

stevedomin:
@josevalim: happy to help with any of this if you me to

nathany:
>  Remove Mix.Task.run from test_helper.exs and move it to mix.exs aliases

For those using the beta of Ecto 2, what should test_helper look like?

josevalim:
@nathany your test_helper should remove both Mix.Task.run. Instead, in your mix.exs, you should have something like this inside `def aliases`:

    "test": ["ecto.create --quiet", "ecto.migrate", "test"]

nathany:
Thanks José. Making that change solved an error we were seeing in CI:

```
[error] Postgrex.Protocol (#PID<0.6145.0>) failed to connect: ** (Postgrex.Error) FATAL
(invalid_catalog_name): database “my_app_test" does not exist
```

nathany:
Possibly another TODO is to use Allowances for Channel testing for handle_in's use of Task.start? https://github.com/phoenixframework/phoenix/blob/master/test/phoenix/test/channel_test.exs#L59

We had to put the following in ChannelCase:

```elixir
setup do
  :ok = Ecto.Adapters.SQL.Sandbox.checkout(MyApp.Repo)
  Ecto.Adapters.SQL.Sandbox.mode(MyApp.Repo, {:shared, self()})
end
```

josevalim:
Yes, that's documented in `Ecto.Adapters.SQL.Sandbox`. :) My proposal is to actually have this in test cases:

```elixir
setup tags do
  :ok = Ecto.Adapters.SQL.Sandbox.checkout(MyApp.Repo)
  unless tags[:async] do
    Ecto.Adapters.SQL.Sandbox.mode(MyApp.Repo, {:shared, self()})
  end
end
```

This way you will use a shared connection whenever your tests are sync.

[...]

PR #1570, opened 2016-02-29
#1564 Migrate to Ecto 2.0 - replace test_transaction by ecto sandbox

aphillipo:
Covers a very small part of #1564 but might be useful.  Tests are also updated and pass. 

Is the sandbox supported for all database drivers?

This is my first little attempt to help so I hope I haven't messed anything up too badly :+1:


josevalim:
Thank you! Your PR looks great but we cannot merge until Ecto 2.0 is out, which we are still not sure when it will happens. I will close this for now but circle back later when we are ready!

aphillipo:
Ah okay dude - I'll leave it then - I was just hoping to be able to get ahead of the curve and start using Ecto 2.0 beta on a brand new project.  It's very difficult to do with phoenix as things stand.

I'll give up and revert back to Ecto 1.x for now.

PR #1583, opened 2016-03-10
Add typespec for Phoenix.Template.template_not_found/2

ciastek:
Without this typespec, running Dialyzer on fresh Phoenix project gives
following warnings:
`layout_view.ex:2: Function template_not_found/2 only terminates with explicit exception`
`page_view.ex:2: Function template_not_found/2 only terminates with explicit exception`

This pull request has two problems:
- added typespec uses built-in `binary` type instead of `Phoenix.Template.name` one.
- typespec is added inside the quoted code of the `__using__` macro. I'm not sure, if it's correct place for it.

### DIFF
--- a/lib/phoenix/template.ex
+++ b/lib/phoenix/template.ex
@@ -132,6 +132,7 @@ defmodule Phoenix.Template do
       By default it raises but can be customized
       to render a particular template.
       """
+      @spec template_not_found(Phoenix.Template.name, map) :: no_return
       def template_not_found(template, assigns) do
         {root, names} = __templates__
         raise UndefinedError,


josevalim:
> added typespec uses built-in binary type instead of Phoenix.Template.name one.

Can we use it though?

> typespec is added inside the quoted code of the __using__ macro. I'm not sure, if it's correct place for it.

this is perfect as is. :)

ciastek:
@josevalim Amended, using Template.name. Didn't know I can use it that way inside a quote.

lib/phoenix/template.ex
```diff
@@ -132,6 +132,7 @@ defmodule Phoenix.Template do
       By default it raises but can be customized
       to render a particular template.
       """
+      @spec template_not_found(Template.name, map) :: no_return
```
josevalim: It would be best to use `Phoenix.Template.name`. This way definitely works but, since it depends on the alias defined at the top, it is a little brittle (and we won't fail to compile if a typespec is wrong). So using the full name is more guaranteed. Thank you :)

ciastek: Done. Thanks for your patience.

josevalim:
Thank you for your contributions!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #1584, opened 2016-03-10
Warn on router paths without leading slash

### LINKED ISSUE #1438: Router paths that don't start with forward slash should raise
This regression popped in somewhere

tracehelms:
### What's this PR do?
This shows a warning anytime you declare a path that doesn't start with a `/` in a `Router`. This is only for consistency in router files. Not including the `/` still works. Closes #1438.

#### Scope.route/7
All of the router macros (http verbs, trace, options, connect, match, and resources) go through the `Scope.route/7` function, so I thought that'd be a good place to put the validation. Otherwise, all slashes are removed and reconstructed in the `Scope.join_path/2` function which is called inside `Scope.route/7`.

#### Scope.push/2
The `scope` macro doesn't go through `Scope.route/7`, so it gets its own validation. I wasn't sure how to do it via pattern matching with the `opts` in `push(module, opts)`, so I just used an `if` statement. Any feedback here?

### Gotchas
I was torn on splitting up the `router_test` and `scope_test` that I added. On one hand, it seemed like the two test suites were already broken up this way. On the other hand, I had to remove the `async: true` in order to break up the tests this way. Do you think this is alright?

It seems like `test/phoenix/router/scope_test.exs` isn't a unit test suite, but integration. Also the [module name is inconsistent with the filename](https://github.com/phoenixframework/phoenix/blob/master/test/phoenix/router/scope_test.exs#L1). I would've added these tests as unit tests but I had no idea how to set that up.

All feedback welcome. Thanks for the awesome framework!

### DIFF
--- a/lib/phoenix/router/scope.ex
+++ b/lib/phoenix/router/scope.ex
@@ -18,7 +18,7 @@ defmodule Phoenix.Router.Scope do
   @doc """
   Builds a route based on the top of the stack.
   """
-  def route(module, kind, verb, path, plug, plug_opts, opts) do
+  def route(module, kind, verb, "/" <> _ = path, plug, plug_opts, opts) do
     private = Keyword.get(opts, :private, %{})
     assigns = Keyword.get(opts, :assigns, %{})
     as      = Keyword.get(opts, :as, Phoenix.Naming.resource_name(plug, "Controller"))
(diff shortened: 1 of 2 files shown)


josevalim:
Maybe we can get rid of those tests? Chris, how do you envision this working. Would we emit those warnings forever or would we eventually raise?

tracehelms:
Hey @chrismccord, anything I need to do to make this PR mergeable?

chrismccord:
We can kill the tests. Whenever Phoenix 2.0 ships we will raise. @tracehelms so please remove the special moduels and IO capture and make the test async and we're all set. Thanks!!!

tracehelms:
@chrismccord Tests removed and commits squashed.

chrismccord:
<3 <3 <3

avocade:
FYI, test fails on latest master for "paths must start with /":

```elixir
  1) test paths must start with / (Phoenix.Router.RoutingTest)
     test/phoenix/router/routing_test.exs:51
     Expected exception CompilerError but got ArgumentError (argument error)
     stacktrace:
       (stdlib) binary.erl:244: :binary.split/3
       (plug) lib/plug/adapters/test/conn.ex:128: Plug.Adapters.Test.Conn.split_path/1
       (plug) lib/plug/adapters/test/conn.ex:24: Plug.Adapters.Test.Conn.conn/4
       test/support/router_helper.exs:35: RouterHelper.call/5
       test/phoenix/router/routing_test.exs:52
```

Issue #1587, opened 2016-03-12
Set minimum length to phoenix.gen.secret


riverrun:
At the moment, you can run `mix phoenix.gen.secret` with any non-negative number, and there is no minimum length. You can, for example, run `mix phoenix.gen.secret 0`, and it will give you an empty string. I think we should be setting a minimum for this, perhaps at 32, which seems to be the minimum in the gen.secret test.

A possible fix would be to add a guard to the `random_string` function:

    defp random_string(length) when length > 31 do
      :crypto.strong_rand_bytes(length) |> Base.encode64 |> binary_part(0, length)
    end
    defp random_string(_), do: Mix.raise "that's too short!"

chrismccord:
:+1 PR?

riverrun:
I can send a PR, but maybe it's easier if you just add those few lines yourself.

PR #1588, opened 2016-03-12
Set minimum length to mix phoenix.gen.secret

riverrun:
Regarding #1587 set a minimum length to mix phoenix.gen.secret.

I've set the minimum length to 32 characters, and it will raise on anything shorter.


chrismccord:
<3 <3 <3

PR #1589, opened 2016-03-13
Change errors to match Ecto nomalised format

amencarini:
As we are changing Ecto error messages from strings to tuples and the key for `validate_number` (see elixir-lang/ecto#1308), this updates Phoenix to be aware of those changes.

### DIFF
--- a/installer/templates/new/web/views/error_helpers.ex
+++ b/installer/templates/new/web/views/error_helpers.ex
@@ -21,15 +21,20 @@ defmodule <%= application_module %>.ErrorHelpers do
     # Because error messages were defined within Ecto, we must
     # call the Gettext module passing our Gettext backend. We
     # also use the "errors" domain as translations are placed
-    # in the errors.po file. On your own code and templates,
-    # this could be written simply as:
+    # in the errors.po file.
+    # Ecto will pass the :count keyword if the error message is
+    # meant to be pluralized.
+    # On your own code and templates, depending on whether you
+    # need the message to be pluralized or not, this could be
+    # written simply as:
     #
     #     dngettext "errors", "1 file", "%{count} files", count
+    #     dgettext "errors", "is invalid"
     #
-    Gettext.dngettext(<%= application_module %>.Gettext, "errors", msg, msg, opts[:count], opts)
-  end
-
-  def translate_error(msg) do
-    Gettext.dgettext(<%= application_module %>.Gettext, "errors", msg)
+    if count = opts[:count] do
+      Gettext.dngettext(<%= application_module %>.Gettext, "errors", msg, msg, count, opts)
+    else
+      Gettext.dgettext(<%= application_module %>.Gettext, "errors", msg, opts)
+    end
   end
 end
--- a/installer/templates/new/priv/gettext/en/LC_MESSAGES/errors.po
+++ b/installer/templates/new/priv/gettext/en/LC_MESSAGES/errors.po
@@ -77,27 +77,17 @@ msgstr[0] ""
 msgstr[1] ""
 
 ## From Ecto.Changeset.validate_number/3
-msgid "must be less than %{count}"
-msgid_plural "must be less than %{count}"
-msgstr[0] ""
-msgstr[1] ""
+msgid "must be less than %{number}"
+msgstr ""
 
-msgid "must be greater than %{count}"
-msgid_plural "must be greater than %{count}"
-msgstr[0] ""
-msgstr[1] ""
+msgid "must be greater than %{number}"
+msgstr ""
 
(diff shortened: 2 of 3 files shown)


installer/templates/new/web/views/error_helpers.ex
```diff
@@ -26,10 +26,9 @@ defmodule <%= application_module %>.ErrorHelpers do
     #
     #     dngettext "errors", "1 file", "%{count} files", count
     #
-    Gettext.dngettext(<%= application_module %>.Gettext, "errors", msg, msg, opts[:count], opts)
-  end
-
-  def translate_error(msg) do
-    Gettext.dgettext(<%= application_module %>.Gettext, "errors", msg)
+    case Keyword.has_key?(opts, :count) do
```
josevalim: What about `if count = opts[:count] do`?

josevalim: We also need to move the comment above inside the `if`.

josevalim:
Thank you! I have added some comments. Note this is part of #1564. We should only merge this after Ecto 2.0 is out. :)

installer/templates/new/web/views/error_helpers.ex
amencarini: :+1: for the `if`!
re: the comment, it seems to apply whether we have `:count` defined or not. Maybe I can add another example without count and amend the comment to clarify what's going on with `:count`?

josevalim: The comment uses dngettext, that's why I think it should apply to only one branch.

amencarini: Yeah, that's why I was thinking of adding a second example:

```
  @doc """
  Translates an error message using gettext.
  """
  def translate_error({msg, opts}) do
    # Because error messages were defined within Ecto, we must
    # call the Gettext module passing our Gettext backend. We
    # also use the "errors" domain as translations are placed
    # in the errors.po file.
    if count = opts[:count] do
      # Ecto will pass the :count keyword if the error message is
      # meant to be pluralised. On your own code and templates,
      # this could be written simply as:
      #
      #     dngettext "errors", "1 file", "%{count} files", count
      #
      Gettext.dngettext(Decimals.Gettext, "errors", msg, msg, count, opts)
    else
      # If pluralization is not needed, on your own code and 
      # templates, this could be written simply as:
      #
      #     dgettext "errors", "is invalid"
      #
      Gettext.dgettext(Decimals.Gettext, "errors", msg, opts)
    end
  end
```

josevalim: What if we keep it at the top and just have two examples?

```elixir
dngettext "errors", "1 file", "%{count} files", count
dgettext "errors", "is invalid"
```

amencarini:
Changed the comment, hope that clarifies it!

whatyouhide:
:clap:

lexmag:
@amencarini please squash everything into single commit. :bowtie:

amencarini:
@lexmag sorry about that, squashed!

josevalim:
Thank you! This will be merged when we effectively start the work in migrating to Ecto 2.0.

amencarini:
Sorry to come back on this, but just wanted to double check before opening another PR on Ecto: Am I correct in thinking that this change also affects `cast` and relation errors? They haven't received the same "tuple" treatment yet.

josevalim:
@amencarini it should affect everything, yeah.

amencarini:
Cool, opened a PR back on Ecto with the errors I spotted, https://github.com/elixir-lang/ecto/pull/1325

installer/templates/new/web/views/error_helpers.ex
```diff
@@ -21,15 +21,20 @@ defmodule <%= application_module %>.ErrorHelpers do
...
     #
     #     dngettext "errors", "1 file", "%{count} files", count
+    #     dgettext "errors", "is invalid"
     #
-    Gettext.dngettext(<%= application_module %>.Gettext, "errors", msg, msg, opts[:count], opts)
-  end
-
-  def translate_error(msg) do
-    Gettext.dgettext(<%= application_module %>.Gettext, "errors", msg)
+    if count = opts[:count] do
```
stevedomin: Should that be `number` instead of `count`?

josevalim: It is count. The Number error messages should not be pluralized.

stevedomin: Oops, I see, thanks @josevalim!

josevalim:
Merged to my upcoming branch, thank you! 

PR #1595, opened 2016-03-15
Start Repo process before accepting HTTP requests

vanstee:
There's a very tiny race condition here on startup, where if you make an
HTTP request right as the Endpoint process has started, but before the
Repo process has started, you'll receive an error saying that the Repo
process hasn't started yet. To avoid these errors let's start the Repo
process first since most Endpoints will most likely depend on access to
the database.


josevalim:

Issue #1477, opened 2016-01-18
Add subscribe/3 to Phoenix.Channel


chrismccord:
Notes:

* Fastlane by default
* Phoenix.Channel.Server should not leak external %Broadcast{}'s to handle_info

PR #1600, opened 2016-03-17
Allow subscribe to multiple topics within the same channel

jfrolich:
This will add the capability of:

- Broadcasting to other topics
- Subscribing to other topics within the same channel
- Unsubscribing to these topics

In some applications it can be helpful to do more targeted pubsub within the same channel. There are several advantages:

- Simplifies the client logic (we do not have to join and manage so many channels).
- We can use the same callbacks, if we have differently scoped channels that are receiving the same messages
- The server does not need to have a running process for each topic

I found this capability very helpful in a larger application that does all communications through channels.

I am quite new to Elixir, so pardon my code :). I also think the test can perhaps be more elegant.

### DIFF
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -291,6 +291,7 @@ defmodule Phoenix.Channel do
       fastlane: {socket.transport_pid,
                  socket.serializer,
                  socket.channel.__intercepts__()})
+    Phoenix.Channel.Server.push socket.transport_pid, socket.topic, "phx_subscribe", %{topic: topic}, socket.serializer
   end
 
   @doc """
@@ -298,6 +299,7 @@ defmodule Phoenix.Channel do
   """
   def unsubscribe(topic, socket) do
     Phoenix.PubSub.unsubscribe(socket.pubsub_server, self(), topic)
+    Phoenix.Channel.Server.push socket.transport_pid, socket.topic, "phx_unsubscribe", %{topic: topic}, socket.serializer
   end
 
 
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -282,6 +282,26 @@ defmodule Phoenix.Channel do
   end
 
   @doc """
+  Subscribe this channel to another topic, so it will receive broadcasts with
+  that particular topic
+  """
+  def subscribe(topic, socket) do
+    Phoenix.PubSub.subscribe(socket.pubsub_server, self(), topic,
+      link: true,
+      fastlane: {socket.transport_pid,
+                 socket.serializer,
+                 socket.channel.__intercepts__()})
+  end
+
+  @doc """
+  Unsubscribe from a topic
+  """
+  def unsubscribe(topic, socket) do
+    Phoenix.PubSub.unsubscribe(socket.pubsub_server, self(), topic)
+  end
+
+
+  @doc """
   Broadcast an event to all subscribers of the socket topic.
 
   The event's message must be a serializable map.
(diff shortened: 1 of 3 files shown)


josevalim:
Thank you @jfrolich. Because this can be all done directly, as you just implemented, by either calling the endpoint or the pubsub system, I am :-1:. I.e. it is quite easy for users to provide this functionality in their own, so we don't need to extend the channel API!

jfrolich:
Fixed a bug so that the client will pick up on the extra subscriptions, previously this was not the case.

The problem of a custom implementation on top of phoenix using the pubsub system:
- It's not possible to directly broadcast to another slice of the channel users, but need to subscribe to a pubsub topic and then push the message out in every process (more complex than just broadcasting to a (sub)topic. Another way to do it is to create more subtopics, but then the same events end up being spread across different channels, so this adds a lot of complexity on the client (listening to the same event across multiple channels, and opening closing all the different channels).
- It is not possible to utilize the fastlane, so worse performance

We already have the ability to have subtopics as separate channels (for instance for a chat room), utilizing the same logic. This allows to subscribe to subtopics at the same time (within the same channel). This makes it easy to broadcast something out to for instance only a single user (multiple connections), or a specific group of users (say chatroom admins) without creating channels for that use case for all the user groups.

To sketch an insight of the problem we can run into. I might for instance have a resource that is photos. And a channel `PhotosChannel`, in this case when I have different audiences in terms of users, for instance only myself, or a group of users I am part of, I need to create for each audience a different channel and subscribe to all those channels to get the right updates. Then when I want to subscribe a collection of photos, or a single photo, I need to create channels for that as well. You can see that for all these combinations the numbers of channels I need to manage will increase quickly. For the same resource we have a combination of channels all over the place.

This change will make that a lot easier. In the Photos channel I am now just able to subscribe or unsubscribe to a number of sub-channels. For instance the channel can update my logged in instances, or I can be subscribed to a specific resource id and receive the updates for that resource. 

Though quite simple this adjustment makes channels quite a bit more versatile in my opinion. Of course I can keep it as a hack on top of phoenix in my own codebase but I think this has the potential benefit a lot of use cases other people will run into. At least I hope there will be a way to address my sketched use cases using channels in an easier way independent of implementation :)

chrismccord:
> It's not possible to directly broadcast to another slice of the channel users, but need to subscribe to a pubsub topic and then push the message out in every process (more complex than just broadcasting to a (sub)topic.

You actually get this today by subscribing to a topic in the channel, but you need to forward it yourself in handle_info, which we're going to fix in https://github.com/phoenixframework/phoenix/issues/1477

So after #1477, you get everything you want without client changes :) Thanks!

jfrolich:
Ah right! Great! Good to see this in the roadmap, and great that there will be a sanctioned solution for this. And I learned some phoenix internals today :)

jfrolich:
One issue with the current implementation in 1.2: when I subscribe to external events within a main topic, it will reach the client as the external topic. So client side I still have to subscribe to all the external topics. The point of external topics is mainly to simplify this and not write code to subscribe to all these channels, now it still means I have to write code to subscribe to all these topics. In my solution in this pull request I wrote a small bit of javascript to make sure the external topics are picked up by the main topic that called the external subscription. And a message "phx_subscribe" that informs the client what external topics are tracked. So that each broadcast to the external topic will be picked up correctly.

Correct me if I'm wrong. But I still do not see how to create an elegant solution to this in 1.2 with the current implementation on the client side.

chrismccord:
You can skip the `phx_subscribe` dance. You'll want to pick up the external topic broadcasts in `handle_info` and forward them to the client. If your broadcast use descriptive names, you can forward the message event + payload directly, or you can rewrite the event names to give proper context. for example:

```elixir
def join("bid:" <> id, %{"product_ids" => prod_ids}, socket) 
  when is_list(prod_ids) and length(prod_ids) <= 100 do
  for prod_id <- prod_ids, do: MyApp.Endpoint.subscribe("product:#{prod_id}")

  {:noreply, socket}
end

def handle_info(%Phoenix.Socket.Broadcast{"product:" <> _, event: ev, payload: pay}, socket) do
  push socket, ev, payload
  {:noreply, socket}
end
```

Here we are subscribing to multiple product topics on a single bidding channel/topic per user,  and when we get a product event, we simply forward it to the client. If your event names have context, like "new_product" or then the bidding channel on the client can pick up events naturally for other topics, i.e.:

```javascript
bidChannel.on("new_product", ...)
```

If you need to add context, then you can rewrite the event, i.e.:

```elixir
def handle_info(%Phoenix.Socket.Broadcast{"product:" <> _, event: ev, payload: pay}, socket) do
  push socket, "product_" <> ev, payload
  {:noreply, socket}
end
```

Here, we prefixed all product events with "product_", so now broadcasted events like "price_increase" will get extra context on the client, i.e.:

```javascript
bidChannel.on("product_price_increase", ...)
```

Does that accommodate your needs?

jfrolich:
Yes, that will do, although this means some extra code to manually forward all events. I still think there is something to say for a subscribe method that automatically masquerades all events to the main channel. Plus I think this approach will also be less performant, with the interception/forwarding code.

hunterboerner:
Same problem here. What's the point of having this if the topic isn't remapped?

PR #1613, opened 2016-03-23
Document dispatch behaviour for struct params

novaugust:
Took me a little bit to debug what was going wrong with my tests where I was being lazy and passing a model as a param. Ended up finding that the Plug test adapter does not normalize structs, and thought that documenting that in Phoenix might be useful for others who run into the situation.

### DIFF
--- a/lib/phoenix/test/conn_test.ex
+++ b/lib/phoenix/test/conn_test.ex
@@ -202,6 +202,9 @@ defmodule Phoenix.ConnTest do
       set the content-type to multipart. The map or list may contain
       other lists or maps and all entries will be normalized to string
       keys
+
+    * a struct - unlike other maps, a struct will be passed through as-is
+      without normalizing its entries
   """
   def dispatch(conn, endpoint, method, path_or_action, params_or_body \\ nil)
   def dispatch(%Plug.Conn{} = conn, endpoint, method, path_or_action, params_or_body) do


chrismccord:
Really a struct shouldn't be passed imo, so it would be better to guard against structs specifically because we can't encode/decode them on the other end. Thoughts?

novaugust:
> Really a struct shouldn't be passed imo

Yeah, that was some serious developer laziness. Fixed it by doing a `Map.take` on the model, which is more realistic as far as what the client will be providing.

As far as guarding against structs though, all the actual normalization of the params is happening outside of Phoenix down inside of Plug, so there'd have to be an extra step added to `dispatch` that verifies the params don't contain a struct. I figure if Plug [has tests for that functionality](https://github.com/elixir-lang/plug/blob/master/test/plug/adapters/test/conn_test.exs#L28-L34), there's a reason someone would do it...

> because we can't encode/decode them on the other end

Sorry, what are you talking about here?

chrismccord:
> Sorry, what are you talking about here?

Passing a struct doesn't make sense because on the controller side we assume the payload will be "decoded" as a raw map, so there is nothing in place that would decode a payload back into its struct form from any client, unless you specifically had a custom plug parser for it.

josevalim:
There is one struct that makes sense passing here: `Plug.Upload`. Maybe we should mention it with an example?

chrismccord:
> There is one struct that makes sense passing here: Plug.Upload. Maybe we should mention it with an example?

How do you think any other struct should be handled?

josevalim:
I am not sure. In theory any parser could insert its own structs, so I would think we should allow other structs to pass through as well.

chrismccord:
<3 <3 <#

novaugust:
hah, <# back at you Chris

PR #1614, opened 2016-03-24
Fix regex to match urls in compressed stylesheets

hamann:
This PR fixes the regex to match urls in compressed stylesheets generated by brunch or webpack, otherwise stylesheets will reference non-digested versions of assets

### DIFF
--- a/lib/phoenix/digester.ex
+++ b/lib/phoenix/digester.ex
@@ -112,7 +112,7 @@ defmodule Phoenix.Digester do
     end
   end
 
-  @stylesheet_url_regex ~r{(url\(\s*)(\S+)(\s*\))}
+  @stylesheet_url_regex ~r{(url\(\s*)(\S+?)(\s*\))}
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you. You are changing the fixture but you are not changing any of the tests. Can we get a failing test please? Also, should we break the CSS over multiple lines?

hamann:
I did not change the tests because they failed with the additional line in the fixture, after that I changed the regex and the tests succeeded. Do you mean I should break these steps in two different commits? And regarding your question about breaking the CSS over multiple lines - what's the point here, can you give me an example?

josevalim:
@hamann thanks, that makes sense. Regarding the multiple lines, instead of:

```css
.foo{...}.bar{...}.baz{...}
```

Please break it into:

```css
.foo{}
.bar{}
.baz{}
```

Mostly to avoid really long columns.

hamann:
@josevalim ok, done

josevalim:

PR #1617, opened 2016-03-25
Fix encoding of spaces in URL path components

### LINKED ISSUE #1616: Incorrect encoding in URL path component
### Environment

* Elixir version (elixir -v): 1.2.1
* Phoenix version (mix deps): 1.1.4

### Expected behavior

When generating a URL that contains spaces as part of the path component, the path component should be encoded using percent encoding. Spaces in particular should be encoded as `%20`, for example if you want to have the `title` of a record in the URL instead of its `id`.

### Actual behavior

Spaces are encoded as `+` using the `application/x-www-form-urlencoded` encoding. These are however treated as literal plus signs when parsing URLs back, even by Phoenix itself.

---

This originally came up in this StackOverflow question: http://stackoverflow.com/questions/36221552/url-to-accept-string-instead-of-id-with-spaces

The fix is probably to replace `URI.encode_www_form` with `URI.encode` in https://github.com/phoenixframework/phoenix/blob/96f58407736a2dd3bd2e4a58cd1efbfe9acc15f3/lib/phoenix/router/helpers.ex#L294-L309, I will immediately start working on a small PR.

padde:
Fixes #1616

### DIFF
--- a/lib/phoenix/router/helpers.ex
+++ b/lib/phoenix/router/helpers.ex
@@ -299,12 +299,12 @@ defmodule Phoenix.Router.Helpers do
   end
 
   defp expand_segments([{:|, _, [h, t]}], acc),
-    do: quote(do: unquote(expand_segments([h], acc)) <> "/" <> Enum.map_join(unquote(t), "/", &URI.encode_www_form/1))
+    do: quote(do: unquote(expand_segments([h], acc)) <> "/" <> Enum.map_join(unquote(t), "/", fn(s) -> URI.encode(s, &URI.char_unreserved?/1) end))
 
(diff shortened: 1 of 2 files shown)


Wobbley:
Excuse my complete ignorance, but how do the phoenix release cycles work? When will this be up in Hex, or can i monkeypatch it?

padde:
@Wobbley not sure about Phoenix's release cycle, this is my first PR :-) In the meanwhile, you can use this branch by forcibly pointing your `mix.exs` to it:

    {:phoenix, github: "padde/phoenix", branch: "1616-fix-path-component-encoding", override:true}

josevalim:
This will be out with Phoenix 1.2. :)

PR #1619, opened 2016-03-26
Fix typos in Phoenix.Presence example docs

napcs:
The Phoenix.Presence examples in both the source and the generated file use `state` instead of `socket`, and also return a tuple instead of simply `:ok` as stated. This corrects all four instances.


chrismccord:
Thanks!

PR #1629, opened 2016-04-04
Fix placement of period within quotes in static cache manifest warning.

keichan34:
This may seem a bit pedantic, but it's potentially confusing because the current warning message looks like the period is part of the file name.


josevalim:
Great catch, thank you!

PR #1637, opened 2016-04-09
fix variable scoping; let,var,const

lukeed:
Your use of `let` works, but technically is the incorrect declaration to use. :) 

* Any variable that isn't mutable or isn't reassigned should be `const`.
* If either of those conditions are true, `let`.
* If the mutable or reassigned variable should be accessed outside its current function scope, use `var`.

### DIFF
--- a/web/static/js/phoenix.js
+++ b/web/static/js/phoenix.js
@@ -356,7 +356,7 @@ export class Channel {
     if(!this.joinedOnce){
       throw(`tried to push '${event}' to '${this.topic}' before joining. Use channel.join() before pushing events`)
     }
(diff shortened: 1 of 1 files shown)


chrismccord:
I'm still unsure if I'd like to adopt this usage of `const` vs `let`. Thanks though!

lukeed:
No problem. I used to do the same thing, but after a second look through es6, i realized it's the spec and not really up to my own discretion. =)

While `let` is easier to write and nicer to look at, `const` immediately let's me know what I should expect to happen with a value when I come back to a project with fresh eyes. $0.02

PR #1641, opened 2016-04-14
Add clickable link to server info

sotojuan:
This PR changes the message displayed by running `mix phoenix.server` to include a clickable link. For example, it'd be:

`Running Hello.Endpoint with Cowboy on http://localhost:4000`

Thought this would be a nice feature—let me know what you think. I might've missed some stuff like other domains. Thanks!

### DIFF
--- a/lib/phoenix/endpoint/cowboy_handler.ex
+++ b/lib/phoenix/endpoint/cowboy_handler.ex
@@ -112,6 +112,6 @@ defmodule Phoenix.Endpoint.CowboyHandler do
 
   defp info(scheme, endpoint, ref) do
     port = :ranch.get_port(ref)
-    "Running #{inspect endpoint} with Cowboy using #{scheme} on port #{port}"
+    "Running #{inspect endpoint} with Cowboy using #{scheme} on http://localhost:#{port}"
   end
 end
--- a/lib/phoenix/endpoint/cowboy_handler.ex
+++ b/lib/phoenix/endpoint/cowboy_handler.ex
@@ -112,6 +112,6 @@ defmodule Phoenix.Endpoint.CowboyHandler do
 
   defp info(scheme, endpoint, ref) do
     port = :ranch.get_port(ref)
-    "Running #{inspect endpoint} with Cowboy using #{scheme} on http://localhost:#{port}"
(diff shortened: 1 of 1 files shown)


lib/phoenix/endpoint/cowboy_handler.ex
```diff
@@ -112,6 +112,6 @@ defmodule Phoenix.Endpoint.CowboyHandler do
 
   defp info(scheme, endpoint, ref) do
     port = :ranch.get_port(ref)
-    "Running #{inspect endpoint} with Cowboy using #{scheme} on port #{port}"
+    "Running #{inspect endpoint} with Cowboy using #{scheme} on http://localhost:#{port}"
```
josevalim: If we don't include `http://`, is it still clickable?

chrismccord:
Should we be printing the IP the server is bound to instead of localhost?

lib/phoenix/endpoint/cowboy_handler.ex
sotojuan: On ITerm 2.9 it doesn't :(

sotojuan:
IP probably makes more sense—let me know what you decide on.

chrismccord:
I think we tried this before but the issue is cowboy binds to 0.0.0.0 by default, but plug does not expose an interface to get this information. So we have a couple choices:

1) Assume localhost for dev and call it complete
2) Check the `http: [ip: ..]` config. If it's explicitly set, use it and fallback to localhost

Thoughts?

josevalim:
@chrismccord i would keep it simple because we removed the last one due to being too complex. if you are deploying, you are likely not caring about that message anyway.

lib/phoenix/endpoint/cowboy_handler.ex
josevalim: Can you then please interpolate the scheme too? Something like:

    Running #{inspect endpoint} with Cowboy using #{scheme}://localhost:#{port}

Thank you!

chrismccord: that would be perfect :+1:

sotojuan: Got it :sunglasses:

sotojuan:
Any reason why it fails? :0

PR #1647, opened 2016-04-15
Support reading the host from the environment at runtime

ericsullivan:
I'm deploying the same code for multiple clients with different URLs. I can solve this with multiple builds, but being able to specify it from the environment (like the port) would be excellent.


josevalim:

PR #1653, opened 2016-04-19
Simplify websocket_handle

scrogson:


jeregrine:
👍

chrismccord:
The goal is to eventually be able to handle them differently, so I'd prefer separate clauses even with the single line of duplication. Thanks!

PR #1657, opened 2016-04-20
Fix a few typos in phoenix.new installer

anderkonzen:
While I was reading the book I found these extra spaces when I ran the installer.


chrismccord:
Thanks!

PR #1661, opened 2016-04-21
Configurable template lookup pattern

andrejsm:
This PR makes possible to configure templates lookup pattern.

If `:pattern` on View is set to `"**/*"` phoenix will load templates from subfolders as well. This reveals possibility to keep structured templates in subfolders and still render them via single View module.

For example:
```elixir
# web/views/news_view.ex
defmodule MyApp.NewsView do
  use MyApp.Web, use: :view, pattern: "**/*"
end
```
with folder structure:
```
web/templates/news
|-- article.html.eex
|-- cards
    |-- user.html.eex
    |-- preview.html.eex
```
you could do:
```eex
<%= render MyApp.NewsView, "article.html", @article_assigns %>
<%= render MyApp.NewsView, "cards/user.html", @user_assigns %>
<%= render MyApp.NewsView, "cards/peview.html", @preview_assigns %>
```


chrismccord:
Thanks! A few notes:

- Let's support either `:path` or `:pattern`, but not both. Default will be to look for `:path`, otherwise `:pattern` will be used as a Path.wildcard lookup
- Let's ditch the "component" naming as it can confuse the code. While this PR would enable you to make your own `include_component` thing, really the goal is to simply render templates from a subdir, which it seems like all your `include_compontent` would do? In any case, lets drop the word "component" from anywhere in this PR
- The default generated options will remain the same, using the `:path`.

Thanks agains!

installer/templates/new/web/web.ex
```diff
@@ -48,9 +48,11 @@ defmodule <%= application_module %>.Web do
     end
   end
 
-  def view do
+  def view(opts \\ []) do
```
josevalim: Let's please drop the changes to this file. You are welcome to do those changes in your application but I don't think they belong to all apps by default.

lib/phoenix/view.ex
```diff
@@ -14,9 +14,11 @@ defmodule Phoenix.View do
   Phoenix defines the view template at `web/web.ex`:
 
       defmodule YourApp.Web do
-        def view do
+        def view(opts \\ []) do
```
josevalim: Let's drop these changes as well. :)

installer/templates/new/web/web.ex
andrejsm: Not sure then how to pass options to Phoenix.Template

chrismccord: You can pass the options in your own app by modifying web.ex to suit your needs or calling `use Phoenix.View` directly with the options you need

andrejsm: Ok, now I see your vision. Just make load path configurable but not out of the box.

andrejsm:
@chrismccord could you clarify a bit on `:path` option. Would path option ignore `:root` or `:path` is appended on `:root`? Same regarding auto resolution of view's folder?

chrismccord:
Sorry when I said "path", I mean't root. So everything I said with s/path/root :)

chrismccord:
ping. Any progress on this?

aphillipo:
Is there a way in an umbrella project to have a shared templates folder for all apps within the umbrella - ideally it would make it possible to override the templates in another project (e.g. a login system etc.).  This would make things incredibly flexible... just a thought?  If this PR doesn't support that I can take a look at it?

chrismccord:
> Is there a way in an umbrella project to have a shared templates folder for all apps within the umbrella

Templates can live in any valid file path, so this is already supported today if you needed it

> ideally it would make it possible to override the templates in another project

This would be instead handled by your own view, via `put_view(MyApp.MyLoginView)`. So you'd override the view, not the templates directly.

chrismccord:
@andrejsm As we are nearing a release, how are we looking on this one? Is this something you'll be able to finish up soon?

andrejsm:
@chrismccord I'll do my best and get back on this tomorrow. Having really busy times. Lack of Elixir devs in my team steals all my free time.

installer/templates/new/web/web.ex
lessless: Won't then  using `Phoenix.View` with a `pattern` argument cause an exception? 
https://github.com/phoenixframework/phoenix/pull/1661/files#diff-5fe767297b48fba5e13a940eb9681b9cR92

andrejsm: You will have to manually add this functionality

lib/phoenix/template.ex
```diff
@@ -199,14 +222,14 @@ defmodule Phoenix.Template do
       Returns the template root alongside all templates.
       """
       def __templates__ do
-        {@template_root, unquote(names)}
+        {@template_root, @lookup_pattern, unquote(names)}
```
josevalim: I don't think we need to return the pattern here because it doesn't seem like nothing uses it?

lib/phoenix/template.ex
```diff
@@ -126,9 +145,11 @@ defmodule Phoenix.Template do
   @doc false
   defmacro __using__(options) do
     root = Dict.fetch! options, :root
+    pattern = Dict.get options, :pattern, "*"
 
     quote do
       @template_root Path.relative_to_cwd(unquote(root))
+      @lookup_pattern unquote(pattern)
```
josevalim: Let's call it `@template_pattern`. :)

installer/templates/new/web/web.ex
lessless: I see. Then it must be removed as well, right?

lib/phoenix/template.ex
lessless: Isn't it a  `glob` https://en.wikipedia.org/wiki/Glob_(programming) ? ;) :P

josevalim: Glob is when you include the asterisk. You don't need to have an asterisk in your pattern. You could use "{foo,bar,baz}.html" if you want to select some templates, for example.

lessless:
Just today I sent a 'feature-request' letter to the phoenix-core mailing list to discuss alike functionality and was very surprised to see a pending PR on that matter. 👍

lib/phoenix/template.ex
andrejsm: It was used here https://github.com/phoenixframework/phoenix/pull/1661/files#diff-a3b14cef44a70d6236e4c4c22d6fb30dR181
but think will be gone with upcoming changes

lessless:
@chrismccord regarding

> Let's support either :path or :pattern, but not both. Default will be to look for :path, otherwise :pattern will be used as a Path.wildcard lookup

If it's possible, I would give my vote to support botha and for ability to combine them. 

I.e. set pattern in the `web.ex`  and root in each particular  view:

```
# web.ex
def view do
  quote do
    use Phoenix.View, pattern: './templates/*'
```
and
```
defmodule MyApp.Letter.View do
  use Reflowmail.Web, which: :view, root: `web/letter`
end
```
wdyt?

andrejsm:
You can achieve that by modifying `__using__` macro. Just prepend your `:root` opt to a `:pattern`

chrismccord:
@lessless I think it could be confusing to see both and folks might use them interchangeably. For example, it would be easy to look at the code and think that you are saying "mount the template root here, and only match this pattern within that root path". Let's use `patterns: [..]` which requires a list of patterns. This would solve the requirements you want to be able to mount multiple dirs, but it leaves the default path separate.

lessless:
Can you please explain what are you thinking of? List of patterns?

andrejsm:
Still work in progress.

This commit makes clear why `:pattern` should be used along with `:root` or other prop which defines a _base_.

If you run test you get errors:
```
1) test renders views defined on pattern (Phoenix.ViewTest)
     test/phoenix/view_test.exs:22
     ** (Phoenix.Template.UndefinedError) Could not render "user.html" for MyApp.PatternView, please define a matching clause for render/2 or define a template at "test/fixtures/templates/cards/**". The following templates were compiled:

     * test/fixtures/templates/cards/media/video.html
     * test/fixtures/templates/cards/user.html

  2) test compiles templates from pattern (Phoenix.TemplateTest)
     test/phoenix/template_test.exs:87
     ** (Phoenix.Template.UndefinedError) Could not render "user.html" for Phoenix.TemplateTest.PatternView, please define a matching clause for render/2 or define a template at "test/phoenix/../fixtures/templates/cards/**". The following templates were compiled:

     * test/phoenix/../fixtures/templates/cards/media/video.html
     * test/phoenix/../fixtures/templates/cards/user.html
```

When defining `:pattern` we somehow must declare base and strip it out of template _match names_. Any ideas on this? I suggest such scenario - `:root` is mandatory and `:pattern` is optional and defaults to `/*`. The `:root` gets stripped out of path. Which gives us:
```elixir
use Phoenix.Template, root: "../shared_templates/whatever", pattern: "{media,users}/**/*"

# would compile

* media/video.html
* media/image.html
* users/admins/super.html
* users/admins/redactor.html
* users/guest.html
* ...
```

@josevalim are you ok with changes I made in `Phoenix.View`? I'm not sure what was your intention when you commented on previous changes in the module and asked to drop the changes. I believe changes I made are ok.

chrismccord:
I see the issue now with trying to allow either or. We still need `:root` to be able to know what prefix to drop off the template names, otherwise, all matched patterns would include the path relative to cwd, such as

    render("lib/my_app/lib/a_view/a_dir/file.html")

So we'll need to rather require `:root` now, and allow patterns to be used along with it, which is what you've probably been trying to get me to understand. Let's hold on this for a moment so that I can take a proper look once I recover from jet lag tomorrow. Thanks!

lessless:
Actually, isn't https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/template.ex#L326 an implicit template?

chrismccord:
Hey @andrejsm, I've tacked this in https://github.com/phoenixframework/phoenix/pull/1711

Thanks so much for giving this one a go, but it was trickier than I was thinking so by the time I figured out what needed to be done, I"d already solve it. Thanks!

PR #1670, opened 2016-04-27
Omit `ecto_repos` config when generating with `--no-ecto`

misfo:
I generated an app today (from `master`) w/ the `--no-ecto` option and I noticed that the `config.exs` file still contained the `ecto_repos: [MyApp.Repo]` config.

Thanks for the project!

### DIFF
--- a/installer/templates/new/config/config.exs
+++ b/installer/templates/new/config/config.exs
@@ -5,12 +5,12 @@
 # is restricted to this project.
 use Mix.Config
 
-# General application configuration
-config :<%= application_name %>,<%= if namespaced? do %>
-  app_namespace: <%= application_module %>,<% end %>
-  ecto_repos: [<%= application_module %>.Repo]
+<%= if namespaced? or ecto do %># General application configuration
+config :<%= application_name %><%= if namespaced? do %>,
+  app_namespace: <%= application_module %><% end %><%= if ecto do %>,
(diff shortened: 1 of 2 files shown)


installer/templates/new/config/config.exs
```diff
@@ -6,9 +6,9 @@
 use Mix.Config
 
 # General application configuration
-config :<%= application_name %>,<%= if namespaced? do %>
-  app_namespace: <%= application_module %>,<% end %>
-  ecto_repos: [<%= application_module %>.Repo]
+config :<%= application_name %><%= if namespaced? do %>,
+  app_namespace: <%= application_module %><% end %><%= if ecto do %>,
+  ecto_repos: [<%= application_module %>.Repo]<% end %>
```
chrismccord: If the app isn't namespaced and not using ecto, we'll be left with `config :my_app` where we should have generated nothing

chrismccord:
I have left one small comment. Otherwise this is good to go. Thanks!

installer/templates/new/config/config.exs
misfo: Ah, gotcha. Fixed in 78633ea3d4f1436a650cd8c9785a82e0233ffbe2

misfo:
@chrismccord I addressed that comment. Lemme know if it looks good now...

chrismccord:
<3 <3 <3

PR #1673, opened 2016-04-27
Continue instrumenting the controller/channel

### LINKED ISSUE #1455: Continue instrumenting the controller/channel
We need to instrument the action dispatch and use the information in logging instead of `Phoenix.Controller.Logger`. We also need to revisit how to enable/disable logger because the API today is through `use Phoenix.Controller, log: false` which is not accessible through `web/web.ex`.

chrismccord:
Closes #1455

@josevalim ready for review. Is this what you had in mind?


lib/phoenix/controller/instrumenter.ex
```diff
@@ -23,28 +16,25 @@ defmodule Phoenix.Controller.Logger do
...
-
-      ["Processing by ", module, ?., action, ?/, ?2, ?\n,
+      controller = inspect(module)
+      action = conn |> action_name() |> Atom.to_string()
+      ["Processing by ", controller, ?., action, ?/, ?2, ?\n,
         "  Parameters: ", params(conn.params), ?\n,
         "  Pipelines: ", inspect(conn.private[:phoenix_pipelines])]
     end
-
-    conn
```
jeregrine: Do you still need to return conn here?

chrismccord: no. The return value is what is passed to the `:stop` event, but we aren't handling any context in that case so can return anything

josevalim:
@chrismccord it is pretty much like above. I would call the module something like Phoenix.Logger or Phoenix.Instrumenter.Logger though because we want to also move channel logging there.

lib/phoenix/controller/instrumenter.ex
josevalim: So I would return `:ok` and match on `:ok` too.

chrismccord:
> . I would call the module something like Phoenix.Logger or Phoenix.Instrumenter.Logger though because we want to also move channel logging there.

Good call. I forgot to pluck the channel logging out. Will do

test/phoenix/test/channel_test.exs
```diff
@@ -353,6 +370,7 @@ defmodule Phoenix.Test.ChannelTest do
     _ = close(socket)
   end
 
+  @tag :capture_log
```
josevalim: Maybe `@moduletag :capture_log`?

chrismccord: 👍 yes definitely

PR #1687, opened 2016-05-04
Add a test for re-reading manifest.json on config change.

zacharydenton:
I thought there was an issue with `Endpoint.static_path(some_path)` not getting updated when there was a config change. This test proves that this isn't the case.

What was actually happening in my case was that `Endpoint.config_change` never got called, and thus the cache was never purged, causing `Endpoint.static_path` to return the stale cached paths. In order to force `config_change` to be called, [I added my app :version to the `Endpoint` config](https://github.com/bitwalker/exrm/issues/206#issuecomment-212538439).

### DIFF
--- a/test/fixtures/manifest_upgrade.json
+++ b/test/fixtures/manifest_upgrade.json
@@ -1,0 +1,1 @@
+{"foo.css":"foo-ghijkl.css"}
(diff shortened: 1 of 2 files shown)


josevalim:
:+1:

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

PR #1689, opened 2016-05-04
Code cosmetics

rrrene:
Hi there,

this PR contains two minor fixes:

One removes extraneous white-space and the other one rewrites an if/else construct so that the if-condition is no longer negated (which I think makes it easier to reason about the else block).

Keep up the good work! :+1:

### DIFF
--- a/lib/phoenix/presence.ex
+++ b/lib/phoenix/presence.ex
@@ -102,7 +102,7 @@ defmodule Phoenix.Presence do
   """
   alias Phoenix.Socket.Broadcast
 
-  @type presences :: %{ String.t => %{metas: [map]}}
+  @type presences :: %{String.t => %{metas: [map]}}
   @type presence :: %{key: String.t, meta: map}
   @type topic :: String.t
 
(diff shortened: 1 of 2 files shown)


chrismccord:
Thanks! While I appreciate the thought, we don't accept purely cosmetic PRs <3

rrrene:
I believe this is not just cosmetics (although I foolishly named the branch and therefore PR like that; it's late where I live).

While not an important change, it transforms a code snippet akin to this:

    if !condition do
      # Block A: gets called when condition is not met
    else
      # Block B: gets called when condition is not not met o_O
    end

into this:

    if condition do
      # Block B: now gets called when condition is met
    else
      # Block A: now gets called when condition is not met 
    end

I believe the second code snippet is easier to reason about when encountering it, which was what prompted me to send the PR in the first place. Anyhow, thx for considering! :+1:

josevalim:
Yeah, I do agree it improves the meaning, it was an unfortunate branch name. :)

PR #1690, opened 2016-05-04
Fixed Small Typo

MarcusSorealheis:
So that users who start with this as a base have the correct syntax. `config:`


MarcusSorealheis:
Actually, not a typo.

PR #1694, opened 2016-05-08
Stringify parameter keys pushed to a channel in test

tokafish:
This uses code lifted from `Plug.Adapters.Test.Conn` to stringify map keys pushed to a channel under test. See: https://groups.google.com/forum/#!topic/phoenix-core/keDSxYg_jSQ

### DIFF
--- a/lib/phoenix/test/channel_test.ex
+++ b/lib/phoenix/test/channel_test.ex
@@ -241,7 +241,7 @@ defmodule Phoenix.ChannelTest do
     if endpoint = Module.get_attribute(__CALLER__.module, :endpoint) do
       quote do
         Transport.connect(unquote(endpoint), unquote(handler), :channel_test,
-                          unquote(__MODULE__), NoopSerializer, unquote(params))
+                          unquote(__MODULE__), NoopSerializer, Phoenix.ChannelTest.__stringify__(unquote(params)))
       end
     else
       raise "module attribute @endpoint not set for socket/2"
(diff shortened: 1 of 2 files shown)


tokafish:
Upon further investigation I'm noticing [connect](https://hexdocs.pm/phoenix/Phoenix.ChannelTest.html#connect/2) also does not stringify map keys. Would it be appropriate to alter this PR to include that function as well?

josevalim:
> Upon further investigation I'm noticing connect also does not stringify map keys. Would it be appropriate to alter this PR to include that function as well?

Yes, please. :)

tokafish:
@josevalim Updated. My macro-fu isn't amazing, so any comments are appreciated

lib/phoenix/test/channel_test.ex
```diff
@@ -239,9 +239,12 @@ defmodule Phoenix.ChannelTest do
   """
   defmacro connect(handler, params) do
     if endpoint = Module.get_attribute(__CALLER__.module, :endpoint) do
+      stringified_params =
+        Code.eval_quoted(params) |> elem(0) |> stringify_params |> Macro.escape
```
josevalim: This won't work if you have, for example, a variable in parameter. The proper solution would be to do something like:

    SomeModule.stringify_keys(unquote(params))

The question is where to define stringify_keys so we can access it for such but without importing it into the user scope

tokafish: yeah. I hesitated to make it public in ChannelTest. It could be inlined? What do you think?

josevalim: One option is to call it `__stringify__` and add a `@doc false`. This way it won't be imported nor show up in docs. It can be used both by macros and regular functions.

tokafish: @josevalim Thanks, I didn't know about that rule for importing. I've made those changes.

josevalim:

Issue #1696, opened 2016-05-08
New project UserSocket and docs should use singular topic convention


chrismccord:
Up for grabs. Easy starter for someone so chime in to claim it :)

chrismccord:
ie,` "rooms:*"` => `"room:*"`

riacataquian:
Can I grab this one @chrismccord? :)

chrismccord:
It's all yours! Thanks!

riacataquian:
@chrismccord Are we going to remove the required topic name when generating new channels?

chrismccord:
Yes. It can be optional now and default to the snake-cased module name. You can use `Phoenix.Naming.underscore`

chrismccord:
@riacataquian Bah, it looks like we already tackled this in https://github.com/phoenixframework/phoenix/commit/1bbda32a57310539c9ba8c3c88045276a37fac2c

Sorry to send you down the wrong path! The next starter issue I'll try to ping and it's yours. Thanks for the willingness to help! :)

scrogson:
Yeah, I tackled that a few weeks ago ;)

riacataquian:
Sure @chrismccord! :) But it seems to me that the docs are not yet fixed? Perhaps I can do that. Example here: https://github.com/phoenixframework/phoenix/blob/c47a5b160382ff179a94fe88e080d2833b7d627e/lib/phoenix/channel.ex

chrismccord:
@riacataquian you are right. Please go ahead! We also want to make sure the generated UserSocket module uses "room:*"

riacataquian:
Ok on it, @chrismccord. 👍

scrogson:
Thank you @riacataquian! ❤️ 💛 💚

riacataquian:
My pleasure! 👍

PR #1699, opened 2016-05-09
Updates docs per singular topic convention for channels

riacataquian:
Makes use of singular topic names for docs and tests per convention for naming channels as mentioned here: [#1696](https://github.com/phoenixframework/phoenix/issues/1696)


scrogson:
:+1:

PR #1700, opened 2016-05-10
Remove handle_out/3 from generated channel

iamvery:
## Problem

The generated test of handle_out/3 was passing for the wrong reason. The generated version of handle_out/3 was never invoked, because the "broadcast" event was not being intercepted.

## Solution

~~Add `intercept` call to generated channel module, so that the test continues to pass, now for the right reason. That reason being, the generated version of handle_out/3 is invoked which pushes the payload back to the socket.~~

The handle_out/3 function is not needed in many cases and is generally considered an advanced feature. No need to generate it with all channels.

h/t @nathanl for pairing

### DIFF
--- a/priv/templates/phoenix.gen.channel/channel.ex
+++ b/priv/templates/phoenix.gen.channel/channel.ex
@@ -22,14 +22,6 @@ defmodule <%= module %>Channel do
     {:noreply, socket}
   end
 
-  # This is invoked every time a notification is being broadcast
-  # to the client. The default implementation is just to push it
-  # downstream but one could filter or change the event.
-  def handle_out(event, payload, socket) do
-    push socket, event, payload
-    {:noreply, socket}
-  end
-
   # Add authorization logic here as required.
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you @iamvery! I think the best solution may be actually to remove `handle_out`. I don't believe it must be there in the generated channel code... it is mostly an advanced feature at this point. Let's see what @chrismccord thinks.

iamvery:
That works for me! I thought it might have been left in there for illustration. Happy to rebase ❤️

chrismccord:
Thanks! We should actually rather just not generate the `handle_out` clause at all because it shouldn't be used in many cases, so I am 👍 on José's thoughts. Can you amend the PR to remove it?

iamvery:
Will do!

iamvery:
Amended the changes and updated the issue to reflect the decision.

chrismccord:
<3 <3 <3

PR #1704, opened 2016-05-14
Improving error message when invalid redirect to: path

martinos:
When we have an invalid `redirect to:` path, we were getting an error like this:

```
** (exit) an exception was raised:
    ** (ArgumentError) the :to option in redirect expects a path
```
Which was not very helpful. Which didn't give much information about the issue.

I have made a small change that makes it more clear. With this pull request we get a message like this:

```
** (exit) an exception was raised:
    ** (ArgumentError) the :to option in redirect expects a path but was "http://localhost:4000/login"
```
Which gives a better idea where the error came from.

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -308,9 +308,9 @@ defmodule Phoenix.Controller do
     cond do
       to = opts[:to] ->
         case to do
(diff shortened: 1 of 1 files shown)


chrismccord:
<3 <3 <3

Issue #1707, opened 2016-05-16
Publish Phoenix.js to npm


nicholaswyoung:
I now have control of the `phoenix` [module name on npm.](https://groups.google.com/forum/#!topic/phoenix-core/yaK6t0eoqYU)

@chrismccord @josevalim, etc core team members: drop your npm usernames here so I can add them to the registry.

aphillipo:
How do people feel about solving this problem for all packages - i.e. quickly knocking up something with a github login that can take paths of JS tagged in a repo/release and automatically mirror these changes to npm.  This could work for any packages and you could specify all the metadata in the UI.  Just seems better than relying on whoever it is to remember, creating a bug every time  someone forgets etc.  My time is limited right now but I'll try to find time if people like the idea; I'd rather people stole the idea though :-D

josevalim:
That's awesome @nicholaswyoung, thank you! I am @josevalim on npm as well.

nicholaswyoung:
@josevalim 👍  

Do we want to publish with Phoenix.js in it's own repo (`phoenix-core` would rely on the `npm` package, that way.) I'd say this approach is standard in the Node world, and in other language systems that interface with JS.

But I can also investigate how we might publish from within the `phoenix-core` repo, if that's your preference.

josevalim:
I think we should just publish from this repository. It already has the .js files and package.json. It is just a matter of adding Chris to NPM and updating the release instructions to include the NPM publishing step.

chrismccord:
Great work! I'm still unsure about the repo setup, so for now I'd prefer to keep everything as is, but publish to npm as we release

nicholaswyoung:
That's easy enough, though I continue my warning that it isn't exactly convention. (And I could see a event where the JS package needs a bugfix and version bump on its own.... and could be confusing to newcomers.)

@chrismccord I assume you have the same username on npm as here?

chrismccord:
@nicholaswyoung yup

nicholaswyoung:
On it. Expect an incoming PR with necessary changes, then you should be able to run `npm publish` at will. :)

nicholaswyoung:
**Update:** should we do the same with `phoenix_html`? To my mind, it doesn't make sense to publish one but not the other. [It's available, after all.](https://www.npmjs.com/package/phoenix-html)

chrismccord:
Unsure. It's like 5 LOC, but then again we're talking node :-P

chrismccord:
[...]

PR #1709, opened 2016-05-17
Prepare JS modules for npm

nicholaswyoung:
Once merged, #1707 should be resolved. WIP, but should be resolved today.

### DIFF
--- a/installer/templates/static/brunch/package.json
+++ b/installer/templates/static/brunch/package.json
@@ -6,8 +6,8 @@
     "watch": "brunch watch --stdin"
   },
   "dependencies": {
-    "phoenix": "file:<%= brunch_deps_prefix %><%= phoenix_path %>"<%= if html do %>,
-    "phoenix_html": "file:<%= brunch_deps_prefix %>deps/phoenix_html"<% end %>
+    "phoenix": "~<%= phoenix_version %>"<%= if html do %>,
+    "phoenix_html": "~<%= phoenix_version %>"<% end %>
   },
   "devDependencies": {
     "babel-brunch": "~6.0.0",
--- a/installer/lib/phoenix_new.ex
+++ b/installer/lib/phoenix_new.ex
@@ -201,6 +201,7 @@ defmodule Mix.Tasks.Phoenix.New do
                phoenix_dep: phoenix_dep(phoenix_path),
                phoenix_path: phoenix_path,
                phoenix_static_path: phoenix_static_path(phoenix_path),
+               phoenix_version: @version,
                pubsub_server: pubsub_server,
                secret_key_base: random_string(64),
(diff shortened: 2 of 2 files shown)


installer/templates/static/brunch/package.json
```diff
@@ -6,8 +6,8 @@
     "watch": "brunch watch --stdin"
   },
   "dependencies": {
-    "phoenix": "file:<%= brunch_deps_prefix %><%= phoenix_path %>"<%= if html do %>,
-    "phoenix_html": "file:<%= brunch_deps_prefix %>deps/phoenix_html"<% end %>
+    "phoenix": "~<%= phoenix_version %>"<%= if html do %>,
+    "phoenix_html": "~<%= phoenix_version %>"<% end %>
```
chrismccord: this will be phoenix_html's version, not phoenix

josevalim: I don't think we should depend on the NPM package by default. Using the one inside deps/ is the simplest approach because developers just update Phoenix. If we put them apart, developers will now need to update Phoenix, then Phoenix may be incompatible with their Phoenix NPM package, then they need to update the package, etc. I think we should still depend on "file:deps/phoenix" and leave the NPM package for those who care about not having a file dependency.

nicholaswyoung: @chrismccord Ah, right. I'll correct that.
@josevalim So publish `phoenix` to npm, but don't depend on it? Since they're already using npm to manage dependencies (`brunch` comes in via npm, correct?) and it's a planned part of the new release workflow, I don't see a downside.

chrismccord:
👍 to what José said. We can publish to npm as we release, but keep the local path dep for simplicity for default projects. Thanks!

nicholaswyoung:
@chrismccord How do you propose handling the publish workflow?

chrismccord:
We will add a line to RELEASE.md that says "publish phoenix and phoenix_html packages to npm" and we are done

nicholaswyoung:
@chrismccord `phoenix_html` should be renamed to `phoenix-html` before publishing, to match up with normal `npm` userland conventions... just FYI.

chrismccord:
As much as it might drive you crazy, I'm going to stick with Elixir conventions for the package names :)

nicholaswyoung:
I don't really care one way or the other. But don't be surprised to see pushback, because you might. And just like Elixir has community guidelines, npm does too. So the criticism will be justified.

chrismccord:
It's an underscore. I've dealt with far worse disagreements. The long answer is we have shipped the dependency with the elixir name for many months now, so it makes sense to keep it. To each their own, but if phoenix_html vs phoenix-html is the worst thing we do, then we're doing well!

PR #1715, opened 2016-05-19
Support System.cmd options for watchers instead of root config

### LINKED ISSUE #1695: Allow a file path to be passed to watcher config
We'll also deprecate the `:root` option on endpoint config

chrismccord:
Closes #1695

### DIFF
--- a/installer/templates/new/config/dev.exs
+++ b/installer/templates/new/config/dev.exs
@@ -12,7 +12,7 @@ config :<%= application_name %>, <%= application_module %>.Endpoint,
   code_reloader: true,
   check_origin: false,
   watchers: <%= if brunch do %>[node: ["node_modules/brunch/bin/brunch", "watch", "--stdin",
-                    cd: Path.dirname(__DIR__)]]<% else %>[]<% end %>
+                    cd: Path.expand("../", __DIR__)]]<% else %>[]<% end %>
 
 
 <%= if html do %># Watch static and templates for browser reloading.
--- a/installer/templates/new/config/dev.exs
+++ b/installer/templates/new/config/dev.exs
@@ -11,7 +11,9 @@ config :<%= application_name %>, <%= application_module %>.Endpoint,
   debug_errors: true,
   code_reloader: true,
   check_origin: false,
-  watchers: <%= if brunch do %>[node: ["node_modules/brunch/bin/brunch", "watch", "--stdin"]]<% else %>[]<% end %>
+  watchers: <%= if brunch do %>[node: ["node_modules/brunch/bin/brunch", "watch", "--stdin",
+                    cd: Path.dirname(__DIR__)]]<% else %>[]<% end %>
+
 
 <%= if html do %># Watch static and templates for browser reloading.
 config :<%= application_name %>, <%= application_module %>.Endpoint,
--- a/lib/phoenix/endpoint/adapter.ex
+++ b/lib/phoenix/endpoint/adapter.ex
@@ -61,12 +61,26 @@ defmodule Phoenix.Endpoint.Adapter do
   defp watcher_children(_mod, conf, server?) do
     if server? do
       Enum.map(conf[:watchers], fn {cmd, args} ->
-        worker(Phoenix.Endpoint.Watcher, [root!(conf), cmd, args],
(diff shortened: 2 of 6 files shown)


installer/templates/new/config/dev.exs
```diff
@@ -11,7 +11,9 @@ config :<%= application_name %>, <%= application_module %>.Endpoint,
   debug_errors: true,
   code_reloader: true,
   check_origin: false,
-  watchers: <%= if brunch do %>[node: ["node_modules/brunch/bin/brunch", "watch", "--stdin"]]<% else %>[]<% end %>
+  watchers: <%= if brunch do %>[node: ["node_modules/brunch/bin/brunch", "watch", "--stdin",
+                    cd: Path.expand("../", __DIR__)]]<% else %>[]<% end %>
```
josevalim: nitpick: are you aligning this with `node: ` or the first list element? because it should totally be the first list element. :P

lib/phoenix/endpoint/adapter.ex
```diff
@@ -61,13 +61,19 @@ defmodule Phoenix.Endpoint.Adapter do
...
       end)
     else
       []
     end
   end
+  defp watcher_args(cmd, cmd_args, conf) do
+    case Enum.split_while(cmd_args, &is_binary(&1)) do
+      {args, []}   -> [cmd, args, [cd: root!(conf)]]
+      {args, opts} -> [cmd, args, opts]
+    end
```
josevalim: Counter proposal:

```elixir
{args, opts} = Enum.split_while(cmd_args, &is_binary/1)
{args, Map.put_new_lazy(opts, :cd, fn -> root!(conf) end)}
```

lib/phoenix/endpoint/watcher.ex
```diff
@@ -2,46 +2,48 @@ defmodule Phoenix.Endpoint.Watcher do
...
-    Task.start_link(__MODULE__, :watch, [root, to_string(cmd), args])
+  def start_link(cmd, args, opts) do
+    Task.start_link(__MODULE__, :watch, [to_string(cmd), args, opts])
   end
 
-  def watch(root, cmd, args) do
-    :ok = validate(root, cmd, args)
+  def watch(cmd, args, opts) do
+    merged_opts = Keyword.merge(
+      [cd: File.cwd!(), into: IO.stream(:stdio, :line), stderr_to_stdout: true], opts)
```
josevalim: I am not sure if we can run into issues with the `File.cwd!` here. Maybe if you need it later on, you can call `inspect opts[:cd] || File.cwd!`?

installer/templates/new/config/dev.exs
chrismccord: It indeed aligns with the first list element as it sits. Generated output:

```
  watchers: [node: ["node_modules/brunch/bin/brunch", "watch", "--stdin",
                    cd: Path.expand("../", __DIR__)]]
```

lib/phoenix/endpoint/adapter.ex
chrismccord: 👍

lib/phoenix/endpoint/watcher.ex
chrismccord: 👌

michalmuskala:
I find it quite wired that you pass the arguments and options in a single list that ends being binaries with a keyword at the end. Wouldn't it be better to pass a tuple with arguments and opts separated?

```elixir
watchers: [node: {~w(node_modules/brunch/bin/brunch watch --stdin),
                  [cd: Path.expand("../", __DIR__)]}]
```

It ends up only sightly more verbose (actually you could omit the brackets around `cd:`), and I feel like is more explicit in what is happening.

chrismccord:
I also went back and forth on the tuple for list keyword thing. I'm open to it, but not sold either way. @josevalim any second thoughts?

chrismccord:
Maybe we should just go with the args to apply to System.cmd?

```elixir
watchers: [
  ["node", ["node_modules/brunch/bin/brunch", "watch", "--stdin"], [cd: Path.expand("../", __DIR__)]]
]
```

Not pretty but it's all we are doing after all and there's no dancing around it

josevalim:
The current format is not ambiguous and less verbose than the options above so I would stick with the current format. :)

PR #1717, opened 2016-05-19
Update errors_on in model_case to refelct Ecto 2.0.0 errors changes

jamescmartinez:
In Ecto 2.0.0 (currently in RC), errors take on a different format and look like this:

```elixir
[username: {"can't be blank", []}]
```
[see docs for 2.0.0-rc.5](https://hexdocs.pm/ecto/2.0.0-rc.5/Ecto.Changeset.html)

whereas before, they looked like this:
```elixir
[username: "can't be blank"]
```
[see docs for 1.1.7](https://hexdocs.pm/ecto/Ecto.Changeset.html)

The current function for `errors_on` will not work with this new error format, so I've updated it to parse the errors first so it can be used the same way as before, like this:

 ```elixir
assert {:username, "can't be blank"} in errors_on(%User{}, %{username: nil})
```

### DIFF
--- a/installer/templates/ecto/model_case.ex
+++ b/installer/templates/ecto/model_case.ex
@@ -59,5 +59,6 @@ defmodule <%= application_module %>.ModelCase do
   """
   def errors_on(struct, data) do
     struct.__struct__.changeset(struct, data).errors
+    |> Enum.map(fn {field, {message, _opts}} -> {field, message} end)
   end
 end
--- a/installer/templates/ecto/model_case.ex
+++ b/installer/templates/ecto/model_case.ex
@@ -59,6 +59,6 @@ defmodule <%= application_module %>.ModelCase do
   """
   def errors_on(struct, data) do
     struct.__struct__.changeset(struct, data).errors
-    |> Enum.map(fn {field, {message, _opts}} -> {field, message} end)
+    |> Ecto.Changeset.traverse_errors(&<%= application_module %>.ErrorHelpers.translate_error/1)
   end
 end

### CODE REVIEW

**Comment by chrismccord** [comment]
Thanks for this! Let's have errors_on use `MyApp.ErrorHelpers` to traverse and translate the errors like you would do in your application:

```elixir
def errors_on(struct, data) do
  struct.__struct__.changeset(struct, data)
  |> Ecto.Changeset.traverse_errors(&MyApp.ErrorHelpers.translate_error/1)
end
```

scrogson:
So...I've got a slight issue with this helper and have been meaning to bring it up. I think this helper should instead take a changeset as the first argument rather than the struct. This help only works in the case that someone has just gone with the default generated changeset in the model we generate. However, in the event that someone has renamed the changes function, or changed the arity, this will no longer work.

So I'd propose that we change the API to:

```elixir
def errors_on(%Ecto.Changeset{} = changeset) do
  Ecto.Changeset.traverse_errors(changeset, &MyApp.ErrorHelpers.translate_error/1)
end
```

Thoughts?

chrismccord:
@scrogson 👍 from me. I'm not sure why we didn't do this from the beginning. @josevalim ?

jamescmartinez:
Happy to make this change, just let me know which you decide on. 👍

josevalim:
I agree with @scrogson. However, this helper will likely not exist when we get rid of "model" in upcoming Phoenix versions, so I would merge this PR and keep it working on structs for now knowing well that @scrogson's feedback will be addressed by the next version.

chrismccord:
@josevalim sounds good, but we should still have it use traverse_errors, no?

josevalim:
@chrismccord yup!

chrismccord:
@jamescmartinez ok so we can roll with 

```elixir
def errors_on(struct, data) do
  struct.__struct__.changeset(struct, data)
  |> Ecto.Changeset.traverse_errors(&MyApp.ErrorHelpers.translate_error/1)
end
```

And we're all set for now. Please make the change and I will merge. Thanks!

jamescmartinez:
I've made the change, but changed `MyApp` to `<%= application_module %>` from your example.

jamescmartinez:
The test that failed:

screen shot 2016-05-20 at 7 42 17 am

chrismccord:
<3 <3 <3

PR #1724, opened 2016-05-23
Fix a typo in the conn() deprecation notice; improve the build_conn() typespec

durub:
New to contributing, so I don't know if this is too minor for a pull request, but here goes.

Stumbled upon it while testing a Plug and had to do a double take because of the "@spec build_conn()" above the conn function. Even though it is minor, I think it's worth fixing.

If I made a mistake with the typespecs, let me know!


chrismccord:
<3 <3 <3

PR #1725, opened 2016-05-24
add mime as standalone dependency

rixmann:
Current plug has ugly deprecated warnings when using Plug.MIME.
Using mime here as a seperate dependency omits those warnings and is backwards compatible with old plug versions.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -28,7 +28,7 @@ defmodule Phoenix.Mixfile do
 
   def application do
     [mod: {Phoenix, []},
-     applications: [:plug, :poison, :logger, :eex],
+     applications: [:plug, :poison, :logger, :eex, :mime],
      env: [stacktrace_depth: nil,
            template_engines: [],
            format_encoders: [],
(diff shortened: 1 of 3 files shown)


Gazler:
If we depend on it directly then we should probably also include it in https://github.com/rixmann/phoenix/blob/0c99a20fbeb478c2fe071c94eaeefdfab93b674f/mix.exs#L31

rixmann:
you are right, i forgot that

jeregrine:
👍

mix.exs
```diff
@@ -52,7 +52,9 @@ defmodule Phoenix.Mixfile do
 
      # Test dependencies
      {:phoenix_html, "~> 2.4", only: :test},
-     {:websocket_client, git: "https://github.com/jeremyong/websocket_client.git", only: :test}]
+     {:websocket_client, git: "https://github.com/jeremyong/websocket_client.git", only: :test},
+     {:mime, "~> 1.0"},
```
chrismccord: please move this up as it's sitting underneath "test dependencies" :)

chrismccord:
<3 <3 <3

rixmann:
:)

PR #1729, opened 2016-05-25
Update documentation for endpoint.ex

arcz:
Secrets are used for encryption rather than encoding data. 🔑


chrismccord:
we don't encrypt cookies by default, so I'm sure of the wording. I'd like something that doesn't lead to that conclusion. We sign by default, but the secret is also used to encrypt if you choose. Can you update the wording to better describe?

josevalim:
We should probably say it is used for "encrypting and signing data. For example, cookies and tokens are signed by default using this value but they may also be encrypted if desired."

arcz:
I updated the text using what @josevalim proposed

chrismccord:
Perfect. Thanks!!

PR #1731, opened 2016-05-26
Add transform option

itrelease:
This will allow to transform payload before send it to channels

### DIFF
--- a/web/static/js/phoenix.js
+++ b/web/static/js/phoenix.js
@@ -397,7 +397,9 @@ export class Channel {
   // Overridable message hook
   //
   // Receives all events for specialized message handling
-  onMessage(event, payload, ref){}
+  onMessage(event, payload, ref){
+    return payload;
+  }
 
   // private
 
@@ -419,9 +421,9 @@ export class Channel {
     if(ref && [close, error, leave, join].indexOf(event) >= 0 && ref !== this.joinRef()){
       return
     }
-    this.onMessage(event, payload, ref)
+    let handledPayload = this.onMessage(event, payload, ref)
     this.bindings.filter( bind => bind.event === event)
-                 .map( bind => bind.callback(payload, ref))
+                 .map( bind => bind.callback(handledPayload, ref))
   }
 
   replyEventName(ref){ return `chan_reply_${ref}` }
@@ -455,9 +457,6 @@ export class Socket {
   //
   //   params - The optional params to pass when connecting
   //
-  //   transform - The optional function that transforms somewhow
-  //               payload before put it into the channel
-  //
   // For IE8 support use an ES5-shim (https://github.com/es-shims/es5-shim)
   //
   constructor(endPoint, opts = {}){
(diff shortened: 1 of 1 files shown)


chrismccord:
Can you describe the usecase that you have that requires this feature? Thanks!

itrelease:
@chrismccord I need keys in payload object to be in camelCase

chrismccord:
How about we instead modify the channel.onMessage contract to return the payload? This way you could transform the payload on a channel by channel basis and we don't introduce any new apis. The diff would be a single line change:

```diff
- this.onMessage(event, payload, ref)
+ payload = this.onMessage(event, payload, ref)
```

If you wanted this to apply to all messages, you could write your own function to build new channel for you:

```javascript
let newChannel = (socket, topic, params = {}) => {
  let chan = socket.channel(topic, params)
  chan.onMessage = (event, paylaod, ref) => { return camelize(payload) }
  return chan
}
...
let roomChannel = newChannel(socket, "rooms:lobby")
```

Thoughts?

itrelease:
@chrismccord but with this approach we still need to do this for each channel right?

itrelease:
@chrismccord but it doesn't affect payload in each individual event, like:
```
channel.on('[event_name]', payload => {
  // here payload still with original data
});
```
right?

chrismccord:
No, it would allow you to rewrite the payload. By default we would return it unchanged, but what we pass to the channel.on(..) handlers is the rewritten payload

itrelease:
@chrismccord hmm that's odd, I've tried that but still I get original data

itrelease:
@chrismccord sorry, I was passing original data :) yeah your approach works for me

chrismccord:
Great. Please go ahead with the changes!

itrelease:
@chrismccord should default `onMessage` return just payload or the contract have to be object with payload in it like:
```
onMessage(event, payload, ref){
    return { event, payload, ref };
}
```
?

chrismccord:
I think just the payload is fine, unless we want to allow customizing events. I suppose if you want your event names to be customized, you need it as well, but the issue with that is phoenix has some special events that you can't coerce, like phx_error and phx_close, so I'm hesitant to expose transforming the event name. Same with ref, it's not something you'd ever want to transform

web/static/js/phoenix.js
```diff
@@ -397,7 +397,9 @@ export class Channel {
   // Overridable message hook
   //
   // Receives all events for specialized message handling
-  onMessage(event, payload, ref){}
+  onMessage(event, payload, ref){
+    return payload;
+  }
```
chrismccord: single line and kill the semicolon :)

itrelease:
@chrismccord fair enough

chrismccord:
Thanks!

chvanikoff:
Glad I found this, got into a situation where I have to emulate socket.io behaviour and this change will definitely easy my job. Thanks @itrelease !

PR #1733, opened 2016-05-28
Potential fix for issue 1605

GriffinMB:
If the key_id doesn't end with "_id", define_field will be set to false to avoid auto-generation, and will instead be set in the foreign_key option.

`mix phoenix.gen.html Room rooms author:references:users` will now set:

`belongs_to :author, Module.Author, define_field: false, foreign_key: :author`

`mix phoenix.gen.html Room rooms author_id:references:users` will still follow default behavior.

### DIFF
--- a/priv/templates/phoenix.gen.model/model.ex
+++ b/priv/templates/phoenix.gen.model/model.ex
@@ -3,7 +3,7 @@ defmodule <%= module %> do
 
   schema <%= inspect plural %> do
 <%= for {k, _} <- attrs do %>    field <%= inspect k %>, <%= inspect types[k] %><%= schema_defaults[k] %>
-<% end %><%= for {k, _, m, _} <- assocs do %>    belongs_to <%= inspect k %>, <%= m %>
+<% end %><%= for {k, k_id, m, _} <- assocs do %>    belongs_to <%= inspect k %>, <%= m %><%= if !String.ends_with?(Atom.to_string(k_id), "_id") do %>, define_field: false, foreign_key: <%= inspect k %><% end %>
 <% end %>
     timestamps
   end


GriffinMB:
This passes all tests locally, and the failed test doesn't seem to have anything to do with my updates! Let me know if there's anything else I need to do.

priv/templates/phoenix.gen.model/model.ex
```diff
@@ -3,7 +3,7 @@ defmodule <%= module %> do
 
   schema <%= inspect plural %> do
 <%= for {k, _} <- attrs do %>    field <%= inspect k %>, <%= inspect types[k] %><%= schema_defaults[k] %>
-<% end %><%= for {k, _, m, _} <- assocs do %>    belongs_to <%= inspect k %>, <%= m %>
+<% end %><%= for {k, k_id, m, _} <- assocs do %>    belongs_to <%= inspect k %>, <%= m %><%= if !String.ends_with?(Atom.to_string(k_id), "_id") do %>, define_field: false, foreign_key: <%= inspect k %><% end %>
```
josevalim: Will this work as well? If we don't define the field... then we won't be able to use it in queries.

GriffinMB: Hmm, perhaps not. Reading the comments (here: https://github.com/elixir-lang/ecto/blob/master/lib/ecto/schema.ex#L674), I was under the impression that, when false, define_field simply didn't set the foreign key automatically, but it could then be set properly with the foreign_key option. 

I did a quick new-project test and ran the automated tests, so thought it was working correctly. I'll do another test and make sure I didn't mess something up.

GriffinMB: You're right, that's not the way the foreign_key option works. With the check for "_id" being in Ecto, the only other solution I see within Phoenix (without much larger changes) is to force the id at the end. Whether that's done automatically or through warning errors. 

I'll go ahead and close this PR.

PR #1736, opened 2016-05-30
call the standalone MIME library in all places

rixmann:
The Plug.MIME library was still called in some places, which generates ugly warning messages with current Plug.


josevalim:

PR #1741, opened 2016-06-02
Channel receive instrumentation

tokafish:
See https://groups.google.com/forum/#!topic/phoenix-core/SIHXEWEoFbA

Example output:

> [info] RECEIVED message on MyApp.ApiChannel
>   Parameters: %{"example" => "Hi"}
>   Ref: 20784513-6CF4-441D-9807-17E10B8D175E

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -345,6 +345,8 @@ defmodule Phoenix.Endpoint do
       `"index.html"` - and the `:format` key - for the format of the template.
     * `:phoenix_channel_join` - the joining of a channel. The `%Phoenix.Socket{}`
       and join params are passed as runtime metadata via `:socket` and `:params`.
+    * `:phoenix_channel_receive` - the receiving of a message over a channel. The `%Phoenix.Socket{}`,
+      payload, event, and ref are passed as runtime metadata via `:socket`, `:params`, `:event`, and `:ref`.
 
   ### Dynamic instrumentation
 
--- a/lib/phoenix/channel/server.ex
+++ b/lib/phoenix/channel/server.ex
@@ -219,9 +219,12 @@ defmodule Phoenix.Channel.Server do
 
   def handle_info(%Message{topic: topic, event: event, payload: payload, ref: ref},
                   %{topic: topic} = socket) do
-    event
-    |> socket.channel.handle_in(payload, put_in(socket.ref, ref))
-    |> handle_result(:handle_in)
+    Phoenix.Endpoint.instrument socket, :phoenix_channel_receive,
+      %{ref: ref, event: event, params: payload, socket: socket}, fn ->
+      event
+      |> socket.channel.handle_in(payload, put_in(socket.ref, ref))
+      |> handle_result(:handle_in)
+    end
   end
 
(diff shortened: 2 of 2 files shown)


chrismccord:
Thanks, but we decided against logging/instrumenting every incoming event out of the box. If you need such a setup you can define a single `handle_in` clause that logs + instruments and delegates the event to a private function on the module, such as `defp incoming("event1", params, socket)`

tokafish:
One thing I noticed in putting this together: It looks like `Phoenix.Endpoint.Instrument` adds `Phoenix.Logger` as an instrumenter, regardless of the app's config. Should the phoenix.new generator create a config with `Phoenix.Logger` explicitly included, giving the user the option to remove it?

tokafish:
@chrismccord Thanks. @josevalim seemed to suggest that you might be open to this in the link above, which is why I submitted it. Might I ask, if there was no runtime cost to adding this instrumentation (for example, it could be turned off based on the configured log level), would you reconsider this functionality?

josevalim:
@chrismccord, we decided to wait until someone needed it. I guess the time has arrived? :)

I think we should instrument this one because it is impossible for the users to do so (opposite to calling broadcast) and if nobody is "listening", it should be no-cost, right?

josevalim:
To be clear, I think we should merge but without the Phoenix.Logger instrumentation for it.

chrismccord:
It's possible by define a catch-all handle_in yourself and wrapping in instrument, but I can be onboard with instrumenting the event. @tokafish can you remove the logger bit?

tokafish:
@chrismccord @josevalim Thanks much! Removed the logger bit and added a tiny bit of documentation. Let me know if there's anything else I can do.

tokafish:
Also, I'm curious if anyone has a suggestion for instrumenting broadcast/push within an app. Since using Phoenix.Channel imports it, I can't define functions named `push` and `broadcast`. I guess the functions could be named differently, or called within an unimported module, but it just feels (slightly) awkward to me.

josevalim:
You can even unimport them in your App.Web and bring the ones you want.

tokafish:
That would be awesome! Forgive my ignorance, how do I unimport? :)

tokafish:
For anyone who's curious to the solution, simply importing Phoenix.Channel again, with an `except` clause, was how to unimport a function 😄

lib/phoenix/endpoint.ex
```diff
@@ -345,6 +345,8 @@ defmodule Phoenix.Endpoint do
       `"index.html"` - and the `:format` key - for the format of the template.
     * `:phoenix_channel_join` - the joining of a channel. The `%Phoenix.Socket{}`
       and join params are passed as runtime metadata via `:socket` and `:params`.
+    * `:phoenix_channel_receive` - the receiving of a message oover a channel. The `%Phoenix.Socket{}`,
```
jquadrin: oover

tokafish: @jquadrin thanks!

tokafish:
@chrismccord Just wanted to check in and see if there's anything I should do to help me this forward. Thanks!

chrismccord:
<3 <3 <3

PR #1749, opened 2016-06-07
Properly encode named params.

chrismccord:
Route helpers encode named param segments with
URI.encode_www_form, but when named params are
received from new requests, the values were not
decoded, leading to a mismatch between how the
helpers encode params and how the route decodes
them. For example, previous behavior:

    profile_path(conn, :show, "my name")
    => "/users/my+name"
    params
    => %{"id" => "my+name"}

Fixed behavior:

    profile_path(conn, :show, "my name")
    => "/users/my+name"
    params
    => %{"id" => "my name"}

@josevalim I'd appreciate a sign-off on this one

### DIFF
--- a/test/phoenix/router/routing_test.exs
+++ b/test/phoenix/router/routing_test.exs
@@ -77,6 +77,12 @@ defmodule Phoenix.Router.RoutingTest do
 
     conn = call(Router, :get, "/spaced users/hello matey")
     assert conn.params == %{"id" => "hello matey"}
+
+    conn = call(Router, :get, "/users/a%20b")
+    assert conn.params == %{"id" => "a b"}
+
+    conn = call(Router, :get, "/backups/a%20b/c%20d")
+    assert conn.params == %{"path" => ["a b", "c d"]}
   end
 
   test "get to custom action" do
--- a/lib/phoenix/router/helpers.ex
+++ b/lib/phoenix/router/helpers.ex
@@ -269,11 +269,14 @@ defmodule Phoenix.Router.Helpers do
     raise ArgumentError, message: String.strip(message)
   end
 
+  @doc false
+  def encode_param(str), do: URI.encode(str, &URI.char_unreserved?/1)
+
   defp expand_segments([]), do: "/"
   defp expand_segments(segments) when is_list(segments),
     do: expand_segments(segments, "")
   defp expand_segments(segments) do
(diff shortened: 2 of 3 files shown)


test/phoenix/router/routing_test.exs
```diff
@@ -77,6 +77,9 @@ defmodule Phoenix.Router.RoutingTest do
 
     conn = call(Router, :get, "/spaced users/hello matey")
     assert conn.params == %{"id" => "hello matey"}
+
+    conn = call(Router, :get, "/users/a+b")
+    assert conn.params == %{"id" => "a b"}
```
jeregrine: Do you need to add another test to cover the list format here https://github.com/phoenixframework/phoenix/pull/1749/files#diff-4b61711a8ead48b802b5b21df56705a4R116

chrismccord: Done. Thanks!

josevalim:
@chrismccord we should not use `encode_www_form` when encoding path segments, only for the query string bits. So this fix is incorrect, we already call `URI.decode` when matching routes so we don't need to do anything else there. The fix is to guarantee that something like: profile_path(conn, :show, "my name") actually emits `"/my%20path"` and that `profile_path(conn, :show, "my path", foo: "my param")` emits `/my%20path?foo=my+param` (although emitting `/my%20path?foo=my%20param` in this case would also be fine).

josevalim:
@chrismccord  to be more precise, the bug is here: https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/router/helpers.ex#L276

It needs to use `URI.encode(string, &URI.char_unreserved?/1)` as in the other functions.

chrismccord:
@josevalim I thought we intentionally encoded the named params that way, per this commit?
https://github.com/phoenixframework/phoenix/commit/9d0847654c5bb169350dbba71c622489c1da6cdb

josevalim:
Great we can use the new function in more places :D

On Wednesday, June 8, 2016, Chris McCord <notifications@github.com> wrote:

> Merged #1749 <https://github.com/phoenixframework/phoenix/pull/1749>.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/phoenixframework/phoenix/pull/1749#event-685849134>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAAlbrPIbwaPtH9tByVvn8xi0juTU9qnks5qJtAogaJpZM4IwKqf>
> .
>


-- 

PR #1751, opened 2016-06-10
Add only one index

joeri:
`mix phoenix.gen.model Score scores score:integer user_id:references:users:unique` currently creates two indices (with the same name)

When running the created migration this gives an error

```
00:32:35.055 [info]  == Running EcPredictions.Repo.Migrations.CreateScore.change/0 forward

00:32:35.056 [info]  create table scores

00:32:35.077 [info]  create index scores_user_id_index

00:32:35.078 [info]  create index scores_user_id_index
** (Postgrex.Error) ERROR (duplicate_table): relation "scores_user_id_index" already exists
```

This pull requests filters out the normal index you get with an association.

Also adds a test for the fact that an index is added when adding a belongs_to field


josevalim:

PR #1752, opened 2016-06-11
Fix binary-id generator location in config/config.exs

Gazler:
Previously the config for the phoenix generators was appended to the
file. This is technically not a problem, however the comment for the
`import_config "#{Mix.env}.exs" explicitly states that it should be at
the bottom of the file.

Before:

    # Import environment specific config. This must remain at the bottom
    # of this file so it overrides the configuration defined above.
    import_config "#{Mix.env}.exs"

    # Configure phoenix generators
    config :phoenix, :generators,
      binary_id: true

After:

    # Configure phoenix generators
    config :phoenix, :generators,
      binary_id: true

    # Import environment specific config. This must remain at the bottom
    # of this file so it overrides the configuration defined above.
    import_config "#{Mix.env}.exs"

The spacing in the template is deliberate to ensure there is only a
single space between the blocks.

### DIFF
--- a/installer/lib/phoenix_new.ex
+++ b/installer/lib/phoenix_new.ex
@@ -196,6 +196,18 @@ defmodule Mix.Tasks.Phoenix.New do
         :error -> adapter_config
       end
 
+
+    generator_config =
+      case get_generator_config(adapter_config) do
+        []               -> nil
+        generator_config ->
+          """
+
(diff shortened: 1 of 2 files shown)


Gazler:
I wanted to PR this first because it makes the template slightly more complicated.

josevalim:
:shipit:

PR #1755, opened 2016-06-15
Cleanup usage of Dict

gmile:
Since Dict is being deprecated, clean up the usage of it

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -734,7 +734,6 @@ defmodule Phoenix.Controller do
 
   defp to_map(assigns) when is_map(assigns), do: assigns
   defp to_map(assigns) when is_list(assigns), do: :maps.from_list(assigns)
-  defp to_map(assigns), do: Dict.merge(%{}, assigns)
 
   defp template_name(name, format) when is_atom(name), do:
     Atom.to_string(name) <> "." <> format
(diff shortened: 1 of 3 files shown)


lib/phoenix/controller.ex
```diff
@@ -734,7 +734,6 @@ defmodule Phoenix.Controller do
 
   defp to_map(assigns) when is_map(assigns), do: assigns
   defp to_map(assigns) when is_list(assigns), do: :maps.from_list(assigns)
-  defp to_map(assigns), do: Dict.merge(%{}, assigns)
```
gmile: I'm very uncertain about this change, but the tests passed.

lib/phoenix/view.ex
```diff
@@ -250,7 +250,7 @@ defmodule Phoenix.View do
 
   """
   def render_existing(module, template, assigns \\ []) do
-    render(module, template, Dict.put(assigns, :render_existing, {module, template}))
+    render(module, template, Keyword.put(assigns, :render_existing, {module, template}))
```
josevalim: I think there is no guarantee it is a keyword on this one. I would suggesting using:

    put_in assigns[:render_existing], {module, template}

josevalim:
Looks great, I have added just one tiny comment.

@chrismccord we want to include this in v1.2.

gmile:
@josevalim thanks, I just pushed a fix to that one

josevalim:

josevalim:

PR #1760, opened 2016-06-17
Use put_req_header correctly in recycle documentation

radar:
The first arg for `put_req_header` should be the `conn` itself.

### DIFF
--- a/lib/phoenix/test/conn_test.ex
+++ b/lib/phoenix/test/conn_test.ex
@@ -85,8 +85,7 @@ defmodule Phoenix.ConnTest do
       conn = post conn, "/login"
 
       # We can also recycle manually in case we want custom headers
-      conn = recycle(conn)
-      conn = put_req_header("x-special", "nice")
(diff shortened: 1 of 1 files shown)


Gazler:
Good spot. Perhaphs the code should assign the result back to `conn` to reflect the other sample lower down.

Could you change `recycle` to `recycle()` so that @whatyouhide doesn't come in and change it after please? :)

josevalim:
@Gazler we will only warn if a variable becomes a function call, in this case it can't happen, so we are fine (that's a common misconception :D).

Gazler:
@josevalim Thanks for clarifying. @radar Sorry for the misinformation!

radar:
All good! 👍👍

chrismccord:
I actually would prefer if we changed the null-amity call to `recycle()`. Please make that change then we are all set. Thanks!

radar:
Done :)

Gazler:
Thanks <3

PR #1764, opened 2016-06-21
Add an `includes?` method to Presence to check presences by key

ProGM:
I searched around if there was a method like this, but I couldn't find it.

An example usage:
I have a chat and I want to check if a certain user is online in the channel. If it isn't, for instance, send him an email.

Currently, I had to do something like:

```elixir
Presence.list("topic").has_key?(user_id)
```

With this method:

```elixir
Presence.includes?("topic", user_id)
```

Maybe it could be renamed to `exists?` or `any?`.

### DIFF
--- a/lib/phoenix/presence.ex
+++ b/lib/phoenix/presence.ex
@@ -160,6 +160,11 @@ defmodule Phoenix.Presence do
 
       def fetch(_topic, presences), do: presences
 
+      def includes?(%Phoenix.Socket{topic: topic}, key), do: includes?(topic, key)
+      def includes?(topic, key) do
+        Phoenix.Presence.includes?(__MODULE__, topic, key)
+      end
+
       def list(%Phoenix.Socket{topic: topic}), do: list(topic)
       def list(topic) do
         Phoenix.Presence.list(__MODULE__, topic)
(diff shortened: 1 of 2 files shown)


lib/phoenix/presence.ex
```diff
@@ -205,6 +210,13 @@ defmodule Phoenix.Presence do
   end
 
   @doc """
+  Returns true if there's a certain presence for a topic.
+  """
+  def includes?(module, topic, key) do
```
josevalim: It would be nice to push this to the tracker. If we are still traversing the whole list, we are not gaining much IMO.

lib/phoenix/presence.ex
```diff
@@ -160,6 +160,11 @@ defmodule Phoenix.Presence do
 
       def fetch(_topic, presences), do: presences
 
+      def includes?(%Phoenix.Socket{topic: topic}, key), do: includes?(topic, key)
```
josevalim: I would call it `has_key?` (that's Elixir's convention)

chrismccord:
This should be pushed upstream to `Phoenix.Tracker` like José said, and then we can think about delegating from Presence, or not. For example, we expose functions that let you grab members by topic or key from `Phoenix.Tracker.State`, but we need to provide them off tracker

ProGM:
Uhm. Would you accept a PR on https://github.com/phoenixframework/phoenix_pubsub ? I can try.

chrismccord:
Exposing this has been on my internal list of todos following 1.2 stable, so hang tight. Thanks!

ProGM:
@chrismccord Any news about this feature? It would be great!

Thank you

chrismccord:
You can track progress here https://github.com/phoenixframework/phoenix/issues/1777

ProGM:
Great, thanks!

PR #1770, opened 2016-06-23
Add .svg to default gzippable_exts

aaronrenner:
This change allows `.svg` files to be gzipped along with the rest of the text based assets (like js and css). Please let me know what you think and if I need to update anything else.


josevalim:

PR #1772, opened 2016-06-24
Add missing backtick to Endpoint docs

colinrymer:


colinrymer:
:grin: And I thought Jose was fast with my bug report for phoenix_html!

Thanks for all your work on this! ❤️ ❤️

PR #1774, opened 2016-06-27
Add Endpoint.version comment for OTP hot upgrade

orikremer:
As per https://github.com/bitwalker/exrm/issues/206,
set the Endpoint.version to trigger code change callback when performing hot upgrade.

### DIFF
--- a/installer/templates/new/config/prod.exs
+++ b/installer/templates/new/config/prod.exs
@@ -55,10 +55,11 @@ config :logger, level: :info
 #
 #     config :<%= application_name %>, <%= application_module %>.Endpoint, server: true
 #
-# You will also need to set the application root to `.` in order
+# You will also need to set the application root to `.` and the endpoint version in order 
 # for the new static assets to be served after a hot upgrade:
 #
 #     config :<%= application_name %>, <%= application_module %>.Endpoint, root: "."
+#     config :<%= application_name %>, <%= application_module %>.Endpoint, version: Mix.Project.config(:version)
 
 # Finally import the config/prod.secret.exs
 # which should be versioned separately.


orikremer:
Related PR in phoenix guides.

https://github.com/phoenixframework/phoenix_guides/pull/528

josevalim:
Thank you. I am not a big fan of this approach unfortunately but I don't
have better alternatives in mind. :) I am OK with it being a guide PR
though as long as we explain it as a way to bust the cache on every
deployment.

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

chrismccord:
I'm still not sure how setting the version in this case works. @josevalim where do we reference this value?

josevalim:
Setting the version guarantees at least a configuration changes which
forced the config_change mechanism that recompiled assets.

On Monday, June 27, 2016, Chris McCord <notifications@github.com> wrote:

> I'm still not sure how setting the version in this case works. @josevalim
> <https://github.com/josevalim> where do we reference this value?
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/phoenixframework/phoenix/pull/1774#issuecomment-228778645>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAAlbnmSzGyXAzilnX6OUMI_qsXiOS1tks5qP-p4gaJpZM4I_HN4>
> .
>


-- 


chrismccord:
ah I see. Yeah it's not great, but I also don't know of a better solution. @orikremer please open up a guide PR and we'll include it there. Thanks!

orikremer:
@chrismccord @josevalim Thanks for the feedback.
I've opened two PRs, this one to add to prod.exs template and another ([528](https://github.com/phoenixframework/phoenix_guides/pull/528)) in the guides.
Reason I've added it also to prod.exs template is for completeness: we already include configuration to correctly serve static assets after hot upgrade:
``` 
# You will also need to set the application root to `.`  in order 
# for the new static assets to be served after a hot upgrade:

#     config :<%= application_name %>, <%= application_module %>.Endpoint, root: "."
```
not having version would result in not loading the new manifest and potentially not serving assets correctly.

orikremer:
@chrismccord @josevalim thoughts on last comment?

josevalim:
@orikremer I have removed the `root: "."` bit because it is no longer required as of Phoenix v1.2. :)

orikremer:
Thanks @josevalim. That's great that `root` config is no longer required however the ```Endpoint.version``` config (or similar) **is** required to get assets working in OTP releases.
Since there's a section for OTP releases configuration in prod.exs it makes sense to me that it'll appear there.

Happy to leave it only as guide change if you think that's enough.

PR #1781, opened 2016-06-28
Change generated changeset default to :empty

kenips:
Hey guys, I've just started learning about Phoenix and so far it's been an amazing ride. Thanks for creating this! I'm just running to a discrepancy with the generator versus what's been said in the Phoenix book. I guess the generated model should use `:empty` as the default?


scrogson:
Hi @kenips,

The generators are using the syntax for Phoenix 1.2 and Ecto 2.0 now (which are not covered in the Programming Phoenix book).

kenips:
I see, thanks @scrogson!

PR #1785, opened 2016-06-29
Improve phoenix.new with Mix.shell.cmd/2 and error handling

nimrev:
In this pull request I've changed the phoenix.new script to now use Mix.shell.cmd/2 instead of :os.cmd/1, because according to the comments it caused some problems on Windows systems. This is possible now because the Mix.shell.cmd/2 function supports the 'quiet: true' option.

I've also added error handling where the return code of the shell command is checked. 
Because I reside in China at the moment, on numerous occasions my 'mix deps.get' command failed because of the Chinese firewall randomly dropping connections without the phoenix.new script informing me that the command failed.

Up for debate is still how to properly handle the error. Instead of printing an error message we could of course also stop the script, but this might be a little bit too intrusive since just running the command again might solve the problem.

Let me know what you think,

Bas

### DIFF
--- a/installer/lib/phoenix_new.ex
+++ b/installer/lib/phoenix_new.ex
@@ -382,12 +382,7 @@ defmodule Mix.Tasks.Phoenix.New do
 
   defp cmd(cmd) do
     Mix.shell.info [:green, "* running ", :reset, cmd]
-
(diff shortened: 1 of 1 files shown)


josevalim:

Issue #1784, opened 2016-06-29
Add possibility to turn off phoenix.digest query string ?vsn=d


VladShcherbin:
* Phoenix version (mix deps): 1.2

Currently, when we run `phoenix.digest`, it adds `?vsn=d` to the links in the files. It also adds fingerprint to the filename, so this `?vsn=d` is not necessary.

It is suggested to use only fingerprints in filename and to remove query string from links ([link](https://gtmetrix.com/remove-query-strings-from-static-resources.html)).

Can we have an option to disable this query string so it won't be added into files and to the links, currently it seems to be hardcoded.

VladShcherbin:
Seems related to [issue](https://github.com/phoenixframework/phoenix/issues/1499) and [commit](https://github.com/phoenixframework/phoenix/commit/ee34b6bcbd46c204974d43a801993b5ddf79d5a3)

josevalim:
You still need the query string to tell the server to cache for long periods of times.

josevalim:
Sorry, premature send. Therefore, we can have an option to disable the query string (if we don't have one yet) but the default should be to leave it enabled.

VladShcherbin:
@josevalim yes, a config option with default _enabled_ would be awesome.

I have nginx pointing to phoenix app, so I think if will work fine.

Thank you.

josevalim:
Can you send a PR please?

On Wednesday, June 29, 2016, Vlad Shcherbin <notifications@github.com>
wrote:

> @josevalim <https://github.com/josevalim> yes, a config option with
> default *enabled* would be awesome.
>
> I have nginx pointing to phoenix app, so I think if will work fine.
>
> Thank you.
>
> —
> You are receiving this because you were mentioned.
> Reply to this email directly, view it on GitHub
> <https://github.com/phoenixframework/phoenix/issues/1784#issuecomment-229376471>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAAlbifvC1oUSTeqn7sgMZjMehQ5a5bLks5qQoNogaJpZM4JBIt5>
> .
>


-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

VladShcherbin:
@josevalim I would love to, but I'm still not that good with `elixir`, sorry.

chrismccord:
> It is suggested to use only fingerprints in filename and to remove query string from links

[...]

PR #1788, opened 2016-07-02
add support for --no-version-param swtich to phoenix.digest

allyraza:
https://github.com/phoenixframework/phoenix/issues/1784

### DIFF
--- a/lib/mix/tasks/phoenix.digest.ex
+++ b/lib/mix/tasks/phoenix.digest.ex
@@ -4,6 +4,7 @@ defmodule Mix.Tasks.Phoenix.Digest do
 
   @shortdoc "Digests and compress static files"
   @recursive true
+  @version_param true
 
   @moduledoc """
   Digests and compress static files.
@@ -32,16 +33,23 @@ defmodule Mix.Tasks.Phoenix.Digest do
     * app-eb0a5b9302e8d32828d8a73f137cc8f0.js
     * app-eb0a5b9302e8d32828d8a73f137cc8f0.js.gz
     * manifest.json
+
+  ## version_param
+
+  Generated links include `version_param` by default 
+  an option `--no-version-param` can be sent to exlcude 
+  the version parameter.
   """
 
   def run(args) do
-    {opts, args, _} = OptionParser.parse(args, aliases: [o: :output])
+    {opts, args, _} = OptionParser.parse(args, aliases: [o: :output], switches: [version_param: :boolean])
     input_path  = List.first(args) || @default_input_path
     output_path = opts[:output] || input_path
+    version_param = Keyword.get(opts, :version_param, @version_param)
 
     {:ok, _} = Application.ensure_all_started(:phoenix)
 
-    case Phoenix.Digester.compile(input_path, output_path) do
+    case Phoenix.Digester.compile(input_path, output_path, version_param) do
       :ok ->
         # We need to call build structure so everything we have
         # generated into priv is copied to _build in case we have
--- a/lib/phoenix/digester.ex
+++ b/lib/phoenix/digester.ex
@@ -20,8 +20,8 @@ defmodule Phoenix.Digester do
     * `input_path` - The path where the assets are located
     * `output_path` - The path where the compiled/compressed files will be saved
   """
-  @spec compile(String.t, String.t) :: :ok | {:error, :invalid_path}
-  def compile(input_path, output_path) do
+  @spec compile(String.t, String.t, Bool.t) :: :ok | {:error, :invalid_path}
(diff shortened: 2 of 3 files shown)


josevalim:
@allyraza thank you! I think we should likely not include `?vsn=d` in the digested assets. I don't remember the reason why we include it though, so we would need to investigate that. I think we do it so we generate proper links but we need to split on ? before calculating the links themselves.

allyraza:
@josevalim can we hold this for a bit I would like to add the docs for --no-version-param switch

lib/mix/tasks/phoenix.digest.ex
```diff
@@ -35,13 +36,14 @@ defmodule Mix.Tasks.Phoenix.Digest do
   """
 
   def run(args) do
-    {opts, args, _} = OptionParser.parse(args, aliases: [o: :output])
+    {opts, args, _} = OptionParser.parse(args, aliases: [o: :output], switches: [version_param: :boolean])
     input_path  = List.first(args) || @default_input_path
     output_path = opts[:output] || input_path
+    version_param = opts[:version_param] || @version_param
```
josevalim: This means version_param will never be false because it will always be `anything || true`. We need to make the tests fail and fix this accordingly by using `Keyword.get/3`.

josevalim:
@allyraza i have added one tiny comment. Once it is addressed, we are good to go!

lib/mix/tasks/phoenix.digest.ex
```diff
@@ -4,6 +4,7 @@ defmodule Mix.Tasks.Phoenix.Digest do
 
   @shortdoc "Digests and compress static files"
   @recursive true
+  @version_param true
```
lexmag: I think it would be better to inline this instead of having module attribute.

lib/phoenix/digester.ex
```diff
@@ -20,8 +20,8 @@ defmodule Phoenix.Digester do
     * `input_path` - The path where the assets are located
     * `output_path` - The path where the compiled/compressed files will be saved
   """
-  @spec compile(String.t, String.t) :: :ok | {:error, :invalid_path}
-  def compile(input_path, output_path) do
+  @spec compile(String.t, String.t, Bool.t) :: :ok | {:error, :invalid_path}
```
lexmag: `Bool.t` –> `boolean`?

lib/phoenix/digester.ex
```diff
@@ -20,8 +20,8 @@ defmodule Phoenix.Digester do
...
+  def compile(input_path, output_path, version_param) do
```
lexmag: We can use `version_param?` in all the places to improve readability and highlight variable nature.

lib/phoenix/digester.ex
```diff
@@ -144,10 +150,14 @@ defmodule Phoenix.Digester do
 
         case Map.fetch(manifest, manifest_path) do
           {:ok, digested_path} ->
-            url
+            url = url
             |> Path.dirname()
             |> Path.join(Path.basename(digested_path))
-            |> Kernel.<>("?vsn=d")
+            if version_param do
```
lexmag: Maybe we can extract this into a helper function.

lib/phoenix/digester.ex
```diff
@@ -144,10 +150,14 @@ defmodule Phoenix.Digester do
 
         case Map.fetch(manifest, manifest_path) do
           {:ok, digested_path} ->
-            url
+            url = url
```
lexmag:
```elixir
url =
  url
  |> Path.join(Path.basename(digested_path))
  |> Path.dirname()
```
Otherwise it's very unclear where first expression ends.

lib/mix/tasks/phoenix.digest.ex
```diff
@@ -32,16 +33,23 @@ defmodule Mix.Tasks.Phoenix.Digest do
...
+  ## version_param
+
+  Generated links include `version_param` by default 
+  an option `--no-version-param` can be sent to exlcude 
+  the version parameter.
   """
 
   def run(args) do
-    {opts, args, _} = OptionParser.parse(args, aliases: [o: :output])
+    {opts, args, _} = OptionParser.parse(args, aliases: [o: :output], switches: [version_param: :boolean])
```
lexmag: Shall we use `:strict` instead of `:switches`?

josevalim: We don't handle errors for now anyway...

chrismccord:
Thanks for the PR but holding on this one. Please see my comment here https://github.com/phoenixframework/phoenix/issues/1784#issuecomment-230809151

Thanks!

PR #1799, opened 2016-07-08
Set host and secret_key_base from ENVs if provided

henrik:
What do you think about doing something along these lines to make e.g. Heroku and Dokku deploys even easier?

I like that it makes those deploys easier while not changing behaviour if you don't want to use ENVs.

I think the `host` part is more clear cut than the secret key part, since that one is still a bit fiddly: having to unignore the file, probably wanting to remove the fallback value. Maybe we could think of some clever way to make this even easier?

I intend this as the starting point of a discussion - I don't expect that you'd want to merge this as-is.

### DIFF
--- a/installer/templates/new/config/prod.exs
+++ b/installer/templates/new/config/prod.exs
@@ -6,6 +6,8 @@ use Mix.Config
 #
 # You should also configure the url host to something
 # meaningful, we use this information when generating URLs.
+# You can either replace the `host` value below, or
+# set the APP_HOSTNAME environment variable.
 #
 # Finally, we also include the path to a manifest
 # containing the digested version of static files. This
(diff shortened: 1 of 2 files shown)


josevalim:
We should be having these discussions in the mailing lists. :)

The issue with such proposals is that it won't work for releases. People
will think they can use ENV vars but it will be misleading. Unless we solve
this dichotomy it will be hard to favor one or the other.


-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

chrismccord:
This has come up several times in the past and it's something I wish we could solve better, but I'm waiting for a better solution to materialized for Elixir in general around releases and dynamic environments. Until then, we'll have to hold on these kind of changes. Thanks!

henrik:
Alright, thank you!

PR #1807, opened 2016-07-11
Make phoenix.js doc consistent with presence.ex

hopsor:
While trying out the new and exciting Presence feature I've just discovered phoenix.js documentation is wrong. In the example it says you should listen to `presences` room channel event to get the current presence of the room channel after you join it. It seems listening to that event doesn't trigger anything. However changing it to `presence_state` does the job.

Edit: Just realized you can use the event name you prefer in the `handle_info(:after_join)` call. In this case `presence.ex` [doc](https://github.com/phoenixframework/phoenix/blob/f7ea46ea581d96ef40245264085369261d195473/lib/phoenix/presence.ex#L46) says `presence_state` so rather than a documentation fix I'd say this PR is a consistency improvement.


chrismccord:
<3 <3 <3

PR #1810, opened 2016-07-12
Add docs for URI argument in Router url helpers

ephe-meral:
This is a proposal to add a check for the `scheme` field of the Plug.Conn when generating URLs with the Router Helpers.

My use case was the following:
I have a phoenix server running behind an SSL proxy (ELB in this case) and I wanted to return URLs with a scheme based on the protocol the client uses. (Which is set through "x-forwarded-proto")

Thoughts? :smile:

### DIFF
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -71,7 +71,13 @@ defmodule Phoenix.Router do
       "/dynamic/something"
 
   The url generated in the named url helpers is based on the configuration for
-  `:url`, `:http` and `:https`.
+  `:url`, `:http` and `:https`. However, if for some reason you need to manually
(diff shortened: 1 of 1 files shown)


josevalim:
Thank you @ephe-meral! I find this a dangerous feature. :D Ideally, if you serve the website behind https, you should always generate https links. Is there any reason why this wouldn't work for you?

josevalim:
> Although now that I saw how it's implemented, and that it's changeable by passing in a URI instead of a Conn etc., I'd guess that it can also be done client-side. Maybe we can simply improve the documentation so that this feature is more visible?

That sounds like a great idea! :)

ephe-meral:
@josevalim I updated the commit :)

josevalim:

Issue #1776, opened 2016-06-27
Proposal: new error page


rstacruz:
Hey, friends. Let's talk about Phoenix's error page. I think we can do a lot of improvement here.

![image](https://cloud.githubusercontent.com/assets/74385/16389363/2a53dc92-3cd0-11e6-9ab6-6cb0d6bf6227.png)

## Situation
This design has served Phoenix and better_errors well for a while, but it can use some improvement.

- The exception messages are limited to one line (tho more info can be shown on hover). This is kind of cumbersome, considering many Elixir errors can span multiple lines.

- The request info/headers/params/session area (below the source) made sense for better errors, because better_errors allowed you to inspect what's in each stack frame. Not quite so in Phoenix. It feels weird that it's huge and it looks like it changes when you click on a stack frame on the left.

- It feels needlessly verbose. Most times I'm only concerned about what error happened, and the line that triggered it. These need to be presented more prominently.

## Proposal

How about something more like this?

![image](https://cloud.githubusercontent.com/assets/74385/16389501/cdc811fe-3cd0-11e6-8842-e6c86928aa8b.png)

Gone is the weird frame-like scrolling; instead, it scrolls down naturally like a page would. (except for the stack trace... you should be able to scroll it down on its own iframe-like container.)

You'll see that above the fold, most the relevant things are what immediately draws your attention.

> ![image](https://cloud.githubusercontent.com/assets/74385/16389698/b691af3a-3cd1-11e6-9321-f88648018911.png)

By putting all the "more info" stuff at the bottom, we get more space to do more interesting things. How links to search Google for the error? Or maybe link to the relevant hexdocs?

> ![image](https://cloud.githubusercontent.com/assets/74385/16389571/1e3142a0-3cd1-11e6-8ef4-33d23db50e9e.png)

This is a rough draft of course, but if you like where this is going I'd be happy to polish it up and implement it.

## By the way

I'm the author of the original [better_errors error page](https://github.com/charliesome/better_errors), which I believe Phoenix's is based off of :)

josevalim:
This is so elegant, thank you! :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

> The exception messages are limited to one line (tho more info can be shown on hover). This is kind of cumbersome, considering many Elixir errors can span multiple lines.

This has been my biggest pet peeve and you solved it perfectly! I absolutely love the links to search and docs too. Well done!

[...]

PR #1811, opened 2016-07-13
Endpoint: customize Plug.Debugger's styles

rstacruz:
Preemptive PR; customization of Plug.Debugger isn't available until Plug 1.2, but it should be safe to have this now, so I'm opening this PR :)

Related:

- #1776 - new error pages
- https://github.com/elixir-lang/plug/pull/423 - new Debugger design
- https://github.com/elixir-lang/plug/pull/425 - customizable Debugger pages


lib/phoenix/endpoint.ex
```diff
@@ -449,7 +454,7 @@ defmodule Phoenix.Endpoint do
       end
 
       if var!(config)[:debug_errors] do
-        use Plug.Debugger, otp_app: @otp_app
+        use Plug.Debugger, otp_app: @otp_app, style: @debugger_style
```
josevalim: Can we inline the attributes here in this case? That's because using module attributes first requires the binary to be copied to ETS and then into the compiled module. That step is quite harmless in Plug.Debugger because we compile it only once in the project life time but, because we may define multiple endpoits, we want to avoid this cost.

josevalim:
Thank you! 

chrismccord:
😍😍😍

PR #1818, opened 2016-07-18
write actual addr in server's log

fsword:
write actual addr in log

bug describe:

server's log always show `localhost`, even it's setuped by user, like that:

```
[info] Running Demo.Endpoint with Cowboy using http://localhost:4000
```

It will make user confused.


josevalim:

Issue #1815, opened 2016-07-17
Endpoint with wrong otp_app fails silently


michalmuskala:
### Environment

* Elixir version (elixir -v): master
* Phoenix version (mix deps): master

### Expected behavior

When `use Phoenix.Endpoint` is called with an `otp_app` that does not exist (or the given module is not configured) an error is raised.

### Actual behavior

There's no error, the default configuration is silently merged and the application just doesn't work.
When starting with `mix phoenix.server` nothing happens and the app sits there infinitely.

This is very confusing and error prone to simple typos in application or module name.

romul:
@chrismccord Have you already started work on this issue?
I was thinking about adding 

```elixir
def from_env!(otp_app, module, defaults) do
  config = Application.fetch_env!(otp_app, module)
  merge(defaults, config)
end
```
to Phoenix.Config and use it [here](https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/endpoint/adapter.ex#L90).

PR #1828, opened 2016-07-21
Fix behavior of `use Phoenix.Endpoint` with wrong otp_app

romul:
When `use Phoenix.Endpoint` is called with an otp_app that does not exist (or the given module is not configured) an error should be raised.

Details: #1815


chrismccord:
Thanks @romul but I had already had this mostly done last night. I appreciate the help!

romul:
@chrismccord  Ok, no problem.

PR #1832, opened 2016-07-23
Add missing fonts used by Bootstrap glyphicons

kovpack:
## This PR:
- adds missing fonts that are used by standard Bootstrap package for [glyphicons](http://getbootstrap.com/components/#glyphicons)
  - adds missing files for the bare app
  - adds missing files for the app that uses brunch
  - adds corresponding tests

## Problem description
Every new Phoenix app (with or without brunch option) included Twitter Bootstrap 3 minified css files by default. Standard Bootstrap 3 css has glyphicons styles, so you expect it's OK to use them, however, when you try using them, you get broken icons and 404 errors for not found fonts. This causes frustration, because you do not expect that files that are used by css styles are really missing, so try to figure out what is wrong with brunch config first, finally realising that there've never been any fonts in the app, so you have to find, download and add them manually.

### missing fonts

image

### icons are broken

image

## After bugfix:
image


chrismccord:
I am 👎 on this because we try to generate as few files as possible for new projects so dropping fonts in is too much. I agree it may be frustrating to assume fonts are there, but it's a tradeoff I'm willing to make. Thanks!

kovpack:
@chrismccord Yeap, adding a few files is ugly solution, but deliberately keeping bugs in a newly generated app, IMHO, is also not much better solution :) And if the only reason is just to keep small number of files... Hm..

I saw a few options:

- build minified bootstrap css, excluding all glyphicons etc.
  - This approach was once used, as I saw in issues list, but it's prone to errors and is very unreliable: you have to build manually each bootstrap version (and have to remember that you should actually do this), not forgetting to remove glyphicons and **hundreds of related styles** (120 KB minified). Moreover, keeping **custom** bootstrap build is bad idea - noone knows what will be cut off there, or which other bugs it will add, so, basically, in most cases it should be deleted from the project and replaced with a new fresh full-feature build. And to see `Boostrap 3.x.x` copyright in styles and get something else - not a proper solution.

- add glyphicons fonts
  - This was the fastest and safest solution (bug ugly), which could save from bugs related to building custom bootstrap (and issues caused by this), bugs related to not building custom solution (missing fonts, hundreds of redundant styles), and which could save users from frustration and spending their time figuring out what is wrong with newly generated app instead of just writing the app you need (first impression matters if you test something for production use).

- **remove third party css stuff at all**
  - This is the best solution, actually, was the one I first selected, but than thought that it would not be merged because of possible negative side-effects. Moreover, if Twitter Bootstrap specific markup and styles are used in other places - this will influence that parts. 
  - This approach is used, as far as I remember, in Rails, which does not have any 3-rd party lib styles in a fresh app, just some 20 lines of basic inline css in custom pages. And it does not offer nor force you to any particular css framework, its styles or markup in a new clean app. 
  - Actually, there is not much sense in adding some partially broken 3-rd party lib if you just need to add some jumbotron-like block and two lists.

Anyway, I just wanted to help, so no problem, just thoughts that may be helpful to improve Phoenix even more :)

omin:
I came across this problem too. Adding the fonts files is not a big deal but I expected them to be there since they're included in the standard bootstrap library. Anyway, I'd like to see Phoenix removing third party css to make the skeleton project minimal, clean, and extensible.

PR #1836, opened 2016-07-26
Don't copy static files when both --no-brunch & --no-html are present.

### LINKED ISSUE #1835: Skip copying files with --no-brunch on pheonix.new when --no-html is specified
On phoenix.new if you pass `--skip-brunch --no-html` static files are still getting copied. This [appears to be by design](https://github.com/phoenixframework/phoenix/blob/90f8b45163a545b264787b73388cb2356d775c3a/installer/lib/phoenix_new.ex#L303). However with both options specified I don't see why static files are needed.

Not sure about being smart and detecting `--no-html` within `copy_static`. I'm happy if `--no-api` is given with more specific documentation on it doing both options above AND not copying static files over.

Happy to submit a PR for either case. Thanks for all the great work again!

kenips:
Fixes #1835.

### DIFF
--- a/installer/lib/phoenix_new.ex
+++ b/installer/lib/phoenix_new.ex
@@ -67,7 +67,7 @@ defmodule Mix.Tasks.Phoenix.New do
     {:eex,  "new/web/views/page_view.ex",                    "web/views/page_view.ex"},
   ]
 
-  @bare [
+  @static [
     {:text,   "static/bare/.gitignore", ".gitignore"},
     {:text,   "static/app.css",         "priv/static/css/app.css"},
     {:append, "static/phoenix.css",     "priv/static/css/app.css"},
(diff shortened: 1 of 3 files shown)


josevalim:
Thank you @kenips! I am afraid though this will still copy everything in `@bare`:

https://github.com/kenips/phoenix/blob/3f4d34f9b2800d48f899434eeda52cfdf138d389/installer/lib/phoenix_new.ex#L70-L76

Maybe we should break `@bare` into two separate attributes and use them accordingly?

kenips:
Sounds good. Just create `@static` for what `@bare` really is. I'm confused about having `test/mix/tasks/phoenix.new_test.exs` and `installer/test/phoenix_new_test.exs` so I moved the new test case to `installer/` based on the level of tests in each.

kenips:
@josevalim when you have a chance on this.

josevalim:
@kenips sorry, for some reason I missed your follow ups. All merged now! 

PR #1838, opened 2016-07-27
Fix prod.exs to include defaults for exrm builds

noma4i:
It's a shame that to get phoenix just running we need to dig in docs.
Align config with documentation: http://www.phoenixframework.org/docs/advanced-deployment

It's not compulsory but will save hours to a newcomers who will deploy by creating release. Exrm releases are de facto standard way to deploy.


josevalim:
I would prefer to link to the docs in prod.exs because this breaks other
deployment strategies.

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

noma4i:
Instead of linking docs we can leave commented options with explanation. Same approach is done in many places in phoenix.

josevalim:
@noma4i the issue is that if we are going to include references on how to deploy with exrm, with mix (for heroku) and maybe with distillery, it is just way too much. If reading the deployment document is too much, maybe the document should include a TL;DR at the top.

josevalim:
Actually, ignore me. There is a whole section called "Using releases" at the end of this file which covers exactly the scenarios you described above. root: "." is no longer required since Phoenix v1.2.

noma4i:
Fair enough. Btw check docs for 1.2 they look outdated.

PR #1840, opened 2016-07-29
Extract serialization into a separate class

jmerriweather:
Extract serialization out so anyone can swap the serialization method. I'm currently using this for MessagePack serialization. It will fallback to JSON if LongPoll is detected.

This would allow someone to write a different serializer such as:

``` javascript
import msgpack from "msgpack-js-v5"

export default class MsgPackSerializer {

  contentType() { return "application/msgpack" }

  isBinary() { return true }

  binaryType() { return "arraybuffer" }  

  encode(payload, callback) {
    callback(msgpack.encode(payload))
  }

  decode(payload, callback) {
    var fileReader = new FileReader()
    fileReader.onload = function(e) {
      let bytes = new Uint8Array(e.target.result.slice(0, e.target.result.byteLength))
      //console.log("about to decode message, that starts with", bytes);
      let msg = msgpack.decode(bytes)        
      callback(msg)
    }
    fileReader.readAsArrayBuffer(payload)
  }
}
```

An example of chat using MessagePack is here: https://github.com/jmerriweather/phoenix_chat

### DIFF
--- a/web/static/js/phoenix.js
+++ b/web/static/js/phoenix.js
@@ -437,23 +437,6 @@ export class Channel {
   isLeaving(){ return this.state === CHANNEL_STATES.leaving }
 }
 
-class JSONSerializer {
-
-  contentType() { return "application/javascript" }
-
-  isBinary() { return false }
-
-  binaryType() { return undefined }  
-
-  encode(payload, callback) {
-    callback(JSON.stringify(payload))
-  }
-
-  decode(payload, callback) {
-    callback(JSON.parse(payload))
-  }
-}
-
 export class Socket {
 
   // Initializes the Socket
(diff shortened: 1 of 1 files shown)


chrismccord:
Is `isBinary` and `binaryType` necessary? I'm not seeing a real use of it in the code and if we could ditch these then the serializer would simply be a decode and encode function and not a full class. I'm not opposed to a class, but if we can get away with user supplied functions it would be simpler. Thoughts?

jmerriweather:
Yeah on second look `isBinary` and `binaryType` are not necessary and as LongPoll will always be JSON we don't need `contentType` either. So would we pass in the encode and decode functions as options?

I kind of like the idea of someone implementing the serializer for me, then all I have to do is import it. But I guess we could still import the encode and decode function separately. Something like the following:

```javascript
import {encode, decode} from "phoenix-fictional-serializer"

let socket = new Socket("/socket", {params: {user: user}, encode: encode, decode: decode})
```
Or
```javascript
let socket = new Socket("/socket", {params: {user: user}, 
  encode: (payload, callback) => {
    callback(JSON.stringify(payload))
  }, 
  decode: (payload, callback) => {
    callback(JSON.parse(payload))
  }})
```

What do you think?

chrismccord:
I need to do some more thinking on this API since the websocket binaryType attribute only accepts how the information for the onmessage callback should be provided, not the outgoing payload like I thought. I'll get something in place to accommodate this usecase, but I need to spend some time tomorrow thinking about the API so hold tight. Thanks!

jeffutter:
For a slightly different API checkout the PR I opened (and immediately closed #1956 since I didn't notice this one. I didn't handle the long polling over there.

chrismccord:
❤️❤️❤️🐥🔥

Thanks for being patience while I had a chance to really think about this one. Looks great and I added tests to cover the new features thanks to the new suite that @rossta stood up for us!

jmerriweather:
No worries! Really excited to have been able to contribute to the Phoenix Framework! Looking forward to 1.3 👍

PR #1852, opened 2016-08-06
Clarify inline docs, web/web.ex

nicholaswyoung:
I think this edit clarifies the purpose of `web/web.ex`.

### DIFF
--- a/installer/templates/new/web/web.ex
+++ b/installer/templates/new/web/web.ex
@@ -1,6 +1,6 @@
 defmodule <%= application_module %>.Web do
   @moduledoc """
-  A module that keeps using definitions for controllers,
+  A module defining __using__ hooks for controllers,
   views and so on.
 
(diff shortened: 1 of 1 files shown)


installer/templates/new/web/web.ex
```diff
@@ -1,6 +1,6 @@
 defmodule <%= application_module %>.Web do
   @moduledoc """
-  A module that keeps using definitions for controllers,
+  A module containing the base definitions for controllers,
```
chrismccord: How about `A module defining `__using__` hooks for controllers.` ?

nicholaswyoung: Even better.

nicholaswyoung:
Rebased this branch against master, and changed the line @chrismccord and I discussed above.

chrismccord:
<3 <3 <3

PR #1854, opened 2016-08-07
Add documentation to assert_push

knewter:
These assertion macros are potential sharp points, since they are
patterns.  This provides documentation to call this out a bit more
explicitly.


chrismccord:
<3 <3 <3

PR #1856, opened 2016-08-08
Upgrade to v 1.3

PatNowak:


josevalim:
Unless there is a reason to upgrade to 1.3, we should stick with also supporting 1.2.

PatNowak:
Sounds logical for now. Thanks for your comment.

PR #1857, opened 2016-08-08
Remove out of date docs for errors_on

HeroicEric:
With the new error format in Ecto 2, this is no longer true.


josevalim:

PR #1859, opened 2016-08-09
New feature: Force WebSocket transport garbage collection

vheathen:
Allow to force WebSocket transport process garbage collection with `send socket.transport_pid, :garbage_collect`. Could be useful after processing large messages.


chrismccord:
Thanks, but I'm still not convinced this is necessary since greedy memory usage was triggered from a synthetic test. It's also possible today to make your own WebSocket transport based on ours that includes this functionality, so the escape hatch is there today if you need it, but requires a little more work. I'd like to explore different real-world cases of requiring this before we'd support it. Thanks!

chrismccord:
Discussing this with @josevalim currently, so hold tight! :)

chrismccord:
<3 <3 <3

josevalim:

PR #1861, opened 2016-08-11
Use clearer language when recommending bang functions in seeds

lukerollans:
A very minor tweak, but I feel like the proposed change is much clearer. The current text in `seeds.exs` is rather redundant.. "as they will fail if something goes wrong" can be said about literally any failure, anywhere. Specifying that execution will be halted if something goes wrong more clearly conveys the intended use of bang functions


josevalim:

PR #1865, opened 2016-08-12
Update docs to new method of adding a MIME

EduardoBautista:
This is how it is described in the Plug documentation because the old way is
no longer supported.

https://github.com/elixir-lang/plug/blob/master/lib/plug/mime.ex#L6


josevalim:

PR #1867, opened 2016-08-13
push socket should be before Presence.track in doc

kpanic:
Hi,
I just started to play with `Phoenix.Presence`.
 I guess the `push` to the socket has to be done after `Phoenix.Presence.track` in the doc, which I guess it would track the data in his own process and *then* send the presence list to the clients
Thanks!

### DIFF
--- a/priv/templates/phoenix.gen.presence/presence.ex
+++ b/priv/templates/phoenix.gen.presence/presence.ex
@@ -19,10 +19,10 @@ defmodule <%= module %> do
         end
 
         def handle_info(:after_join, socket) do
+          push socket, "presence_state", Presence.list(socket)
           {:ok, _} = Presence.track(socket, socket.assigns.user_id, %{
             online_at: inspect(System.system_time(:seconds))
           })
-          push socket, "presence_state", Presence.list(socket)
           {:noreply, socket}
         end
       end


princemaple:
It seems to me that the original code is correct. You sync the `state` first, then the presence of yourself will be pushed out through a `diff`. And.. it used to be the way you are suggesting here. It's changed in https://github.com/phoenixframework/phoenix/commit/de4e6295cc5cf6f1457911b9fbf82b5e8ffa03ef

josevalim:
Yes, this seems to be the correct one indeed. Otherwise you can have a race condition as well as  between pushing the list and tracking, you get new clients.

kpanic:
ok, then should I correct the `phoenix.gen.presence` here https://github.com/phoenixframework/phoenix/blob/master/priv/templates/phoenix.gen.presence/presence.ex#L21?

josevalim:
The one in phoenix presence and in this pull request are the correct ones. :) We should merge this, let's just wait for @chrismccord!

kpanic:
ah sorry, I misinterpreted your first comment @josevalim :)
Waiting for instructions people ;)

scrogson:
I believe the explanation from @princemaple is correct.

chrismccord:
The current docs are correct. I made this change here:
https://github.com/phoenixframework/phoenix/commit/de4e6295cc5cf6f1457911b9fbf82b5e8ffa03ef

As @princemaple said, you sync the state, then your own presence will be pushed in the presence_diff. There won't be an issue with missing event since we don't handle anything until after join. Once the handle_Info is picked up we are already subscribed to the topic, which will receive presence_diff events. To avoid races, we sync the state to grab all known members, including the presence_diff ones we may have raced from join -> handle_info. Were you observing races or you thought the code looked like it needed swapped?

kpanic:
Hey @chrismccord I had no issues when swapping the push statement before or after `Presence.track`

I thought it was more consistent to document in the same way the push statement in:
https://github.com/phoenixframework/phoenix/blob/master/priv/templates/phoenix.gen.presence/presence.ex#L21
and
https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/presence.ex#L45

chrismccord:
Ah, can you amend the PR to fix the gen.presence docs to match the lib/presence docs? Thanks!

kpanic:
sure, no prob!

chrismccord:
<3 <3 <3

PR #1875, opened 2016-08-19
Add :view key to phoenix_controller_render event metadata

deadtrickster:
Hi,

Currently there no way to distinguish views and this patch fixes this.
When applied metadata looks like this:

```elixir
%{format: "json-api", template: "index.json-api", view: Api.UserView}
```

Can this be applied to 1.2 branch as well? 

Thank you.


josevalim:

PR #1877, opened 2016-08-22
Expose source es6 file too in npm package

princemaple:
Packages published to NPM better include its original source files, as the they are not meant to be served directly. (although can be, through npmcdn or other services)

Including the original es6 code enables all sorts of nice functions in build tools.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "phoenix",
-  "version": "1.2.0",
+  "version": "1.2.1",
   "description": "The official JavaScript client for the Phoenix web framework.",
   "license": "MIT",
   "main": "./priv/static/phoenix.js",
@@ -15,7 +15,7 @@
     "babel-brunch": "~6.0.0",
     "uglify-js-brunch": "~2.0.1"
   },
-  "files": ["README.md", "LICENSE.md", "package.json", "priv/static/phoenix.js"],
+  "files": ["README.md", "LICENSE.md", "package.json", "priv/static/phoenix.js", "web/static/js/phoenix.js"],
   "scripts": {
     "test": "./node_modules/.bin/mocha ./web/test/**/*.js --compilers js:babel-register"
   }


chrismccord:
I'm confused how references `main` as the es6 source will be compatible with non-babel users? Pointing to the compiled priv/ as main allows us to not depend on babel downstream. What am I missing?

princemaple:
So I can see two ways of using the npm package directly.

One is via npmcdn, so you specify a path to the file you want to load, i.e. irrelevant to the `main`. You can do this [https://npmcdn.com/phoenix@1.2.0/priv/static/phoenix.js](https://npmcdn.com/phoenix@1.2.0/priv/static/phoenix.js)

The other is to `npm install` and serve the `node_modules` dir in the public static assets, which is again irrelevant to the definition of `main`.

I don't code node servers so I might have missed other ways of directly using the source code that **is** relevant to what `main` is. For any other ways of using the npm package though, that do require some sort of transpilation, the es6 code would be the superior source. However, if there **are** cases that I've missed, I'll happily revert the `main` change.

princemaple:
> Pointing to the compiled priv/ as main allows us to not depend on babel downstream.

On a second thought, were you actually talking about simply bundling the code without any transpilation? I guess you are correct. I'll remove the second commit.

princemaple:
Force pushed to remove the `main` change

chrismccord:
<3 <3 <3

princemaple:
Thanks!

If you could publish it to NPM you'd make my day :heart:

princemaple:
Hi, just wondering if a new release would be published to NPM anytime soon. Thanks!

nicholaswyoung:
@princemaple @chrismccord I just published to NPM, so that's handled.

There are better methods (that are already standardized in the npm/node/js community), like `jsnext:main` (here are a [couple of](https://github.com/reactjs/redux/blob/master/package.json#L7) [examples](https://github.com/rollup/rollup/wiki/jsnext:main)).

IMO, this is how we should address ES6 support in the longterm. I'll assemble a PR for that shortly.

princemaple:
@nicholaswyoung great, thanks!

PR #1880, opened 2016-08-23
for log

shinriyo:
we would like to prevent log


chrismccord:
Thanks, but I don't believe this needs to be part of the default project as I haven't encountered any npm logs by accident.

PR #1881, opened 2016-08-24
Make errors_on support different changeset fns

xingxing:
1. Let errors_on/3 support different changeset functions via pass the third argument (default :changeset).
2. Let errors_on/1 support the changeset created by manua
3. Update the doc for errors_on


josevalim:
Thank you but this function was meant to be simple on purpose. It is meant to provide guidance rather than be a complete solution that would work across all applications.

PR #1885, opened 2016-08-25
Remove delete! comment in generated controllers

alex88:
Was thinking that maybe it could be worth to remove the comments in the generated controllers, since none of the other methods are commented and the comment is regarding the usage of `delete!`.
The comment is added when migrating controllers to Ecto 2 https://github.com/phoenixframework/phoenix/commit/4428e142ffc8c0b6a2ae4bf639244485c9060b13 but the `delete!` was there even before that


chrismccord:
Thanks! We're making changes around the generators so I'm going to hold on this one.

Issue #1571, opened 2016-02-29
Respect connection host on :force_ssl option without custom domain


tjsousa:
### Precheck

* For help and support, use Stack Overflow or [phoenix-talk](https://groups.google.com/forum/#!forum/phoenix-talk).
* For proposing a new feature, please start a discussion on [phoenix-core](https://groups.google.com/forum/#!forum/phoenix-core).
* For bugs, do a quick search and make sure the bug has not yet been reported.
* All checked? Be nice and have fun!

### Environment

* Elixir version (elixir -v): 1.2.2
* Phoenix version (mix deps): 1.1.4
* NodeJS version (node -v): 5.5.0
* NPM version (npm -v): 3.5.3
* Operating system: OSX 10.11.4

### Expected behavior

I'm trying to force SSL on all connections to an app that can be deployed to different custom domains (currently on Heroku). I'm not setting up any `host` configuration for the app.

[Based on `Plug.SSL` behaviour](https://github.com/elixir-lang/plug/blob/master/lib/plug/ssl.ex#L97), I would expect that the connection's `host` would be used as a default when no custom domain has been configured in Phoenix.

### Actual behavior

The SSL redirection is actually working but I am being forced into `https://localhost` which is not the desired behaviour.

[The fix seems to be around this line](https://github.com/elixir-lang/plug/blob/master/lib/plug/ssl.ex#L97), but I may be missing something that this default accomplishes in another context. Any insight?

josevalim:
What happens if you explicitly set `force_ssl: [host: nil]`? Phoenix automatically copies the URL host, that's why you are seeing the behaviour you have described.

tjsousa:
It worked beautifully.

I checked the source code by my standard lib foo didn't make me realize the `put_new/3` was being used and I could trick it with a `host: nil`. Sorry for wasting your time ;)

sgtpepper43:
I think this fix should be included in the docs, I went down quite the rabbit hole until finally finding this issue.

josevalim:
@sgtpepper43 please do send a PR! :D

PR #1887, opened 2016-08-25
Update Endpoint docs to better explain force_ssl config

sgtpepper43:
Specify that `Plug.SSL` is forwarded `:host` from `:url`'s config
Describe how to use `Plug.SSL`'s default host value by setting to `nil`

See #1571


josevalim:

PR #1893, opened 2016-08-31
Change "Processing by" to "Processing with" in the request log

laserlemon:
…because @scrogson asked us to at ElixirConf 2016.


josevalim:
:+1:

laserlemon:
There's talk of using "Processed by" rather than "Processing with" but because this log line is output when the request starts, I think that the present tense is still most appropriate.

josevalim:

PR #1898, opened 2016-09-02
Fix a module definition typo in the channels docs

gsamokovarov:
Just a little something I stumbled upon reading the channels docs.


chrismccord:
Thanks!

PR #1901, opened 2016-09-03
Allow user definable class name in `error_tag` helper

lukerollans:
Pretty much what it says on the tin! This allows a user to define a specific class name (or names) when rendering `error_tag`.

```elixir
<%= error_tag f, :field, class: "error-message" %>
```

Also generates `error_helper_test.exs` with default coverage in `phoenix.new`

### DIFF
--- a/installer/templates/new/web/views/error_helpers.ex
+++ b/installer/templates/new/web/views/error_helpers.ex
@@ -9,10 +9,10 @@ defmodule <%= application_module %>.ErrorHelpers do
   Generates tag for inlined form input errors.
   """
   def error_tag(form, field, opts \\ []) do
-    opts = Keyword.put_new(opts, :class, "help-block")
+    {class_name, _opts} = Keyword.pop(opts, :class, "help-block")
 
     if error = form.errors[field] do
-      content_tag :span, translate_error(error), class: opts[:class]
+      content_tag :span, translate_error(error), class: class_name
     end
   end
 <% end %>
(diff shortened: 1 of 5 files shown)


installer/templates/new/web/views/error_helpers.ex
```diff
@@ -8,9 +8,11 @@ defmodule <%= application_module %>.ErrorHelpers do
   @doc """
   Generates tag for inlined form input errors.
   """
-  def error_tag(form, field) do
+  def error_tag(form, field, opts \\ []) do
+    opts = Keyword.put_new(opts, :class, "help-block")
```
scrogson: How about using `Keyword.pop/3` here?

```elixir
{class_name, _opts} = Keyword.pop(opts, :class, "help-block")
```

lukerollans: This does seem to make more sense when there's only one possible option, I'll update it

chrismccord:
Thanks @lukerollans ! Since the `error_tag` is a function defined in user-land, we intentionally make it basic to serve as an example for user extension. For those wanting more specialized handling, such as classes in this case, then the code you've produced is exactly the way to go. tldr; the error helper functions are a starting piton for user extension, so we keep them basic. Thanks!!!

lukerollans:
Oh noes. OK, thanks @chrismccord, at least I learned a bit about `phoenix.new`

lukerollans:
@chrismccord FWIW, this work was inspired by having seen this in Slack a few times.

screen shot 2016-09-04 at 8 40 37 am

It seems there's a few users (that I've seen at least) aren't aware that `error_helpers.ex` exists, and given the most common customisation of form output is a class name I figured it was sensible to add this a default despite the auto-generated function being a starting point.

josevalim:
@lukerollans that's a very good point and I think that, in that case, the best solution is to change the `error_tag` implementation indeed, since you can set the class red as a default for everything!

PR #1903, opened 2016-09-06
Add required babel presets to static template

jessedijkstra:
`package.json` is currently missing packages used by babel. These packages are nog longer bundled with babel itself.

### DIFF
--- a/installer/templates/static/brunch/package.json
+++ b/installer/templates/static/brunch/package.json
@@ -15,6 +15,8 @@
     "clean-css-brunch": "~2.0.0",
     "css-brunch": "~2.0.0",
     "javascript-brunch": "~2.0.0",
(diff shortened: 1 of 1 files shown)


chrismccord:
As far as I'm aware, the current package.json builds fine without the presets included as deps? Are you running into cases where the default project fails to build assets?

jessedijkstra:
@chrismccord I'll test it again. Might have to do that I used Node 6.4.

chrismccord:
ping @jessedijkstra

peillis:
This happens using node v4.3.0 Once set the babel dependencies everything else works ok.

chrismccord:
We document that node >= v5 is required. So I think we can close this one. Thanks for taking a look!

PR #1905, opened 2016-09-08
Add js/2 to send down js response

imranismail:


lib/phoenix/controller.ex
```diff
@@ -282,6 +282,21 @@ defmodule Phoenix.Controller do
...
+
+  ## Examples
+
+      iex> js conn, "console.log('Ground Control to Major Tom')"
+
+  """
+  @spec js(Plug.Conn.t, iodata) :: Plug.Conn.t
+  def js(conn, data) do
+    conn
+    |> put_resp_content_type(Plug.MIME.type("js"))
```
josevalim: Maybe we should just hardcode the value Plug.MIME returns here? Similar to how we do for `html/2`?

josevalim: Actually, look at `html/2` implementation as this could be 1LOC. :) Other than that, :+1: from me.

chrismccord:
Thanks for the PR! I am -1 since you can already send json by rendering a "foo.js" template or view function.

josevalim:
@chrismccord I thought this mirrors html/2 though?

chrismccord:
@josevalim we shouldn't have shipped the `html` and `json` functions imo :) and I've resisted adding new ones where we already have a means to do it in the view layer. For json/html, I almost always reccoemnd people avoid them and use a view anyway. In the case of `js`, it would be much cleaner to write js in a view or eex template than within a controller itself. Thoughts?

josevalim:
@chrismccord if I have something quick to return, forcing to go through a view+template seems like too much? I agree people shouldn't have large js snippets in their controller but that doesn't exclude small ones?

chrismccord:
> if I have something quick to return, forcing to go through a view+template seems like too much?

this is the reason we added `json`, which I can see in the real world of needing to send quick json responses *as maps*. For `js`, I would much rather see folks writing javascript in a js.eex file or at least a view function than js strings sprinkled in the controller.

imranismail:
In my case I needed to just send a short js response in to a turbolinks mobile app like so

```js
Turbolinks.visit('/')
```

I think it's really handy for when you don't need a full view for this

chrismccord:
I think the usecases would be rare and I'm only okay with `json` because we are sending elixir terms that are encoded. I think in practice trying to extend or debug javascript strings that are sprinkled in controllers is not going to be what we want to promote. In the cases where users absolutely can't live without sending small js snippets in the controller, then it's just a few line function definition in their project and they have it. I don't disagree it may be nice for rare usecases, but that's the case for implementing yourself especially when it's only a few LOC

chrismccord:
Case in point. Instead of sprinkling js strings everywhere that you're going to have to track down and maintain in controllers, a function that does this in one places accommodates your usecase beautifully:

```elixir
def create(conn, ...) do
  ...
  turbo_visit(conn, "/")
end

def turbo_visit(conn, path) do
  conn
  |> put_resp_content_type("application/javsacript")
  |> send_resp("Turbolinks.visit('#{path}')")
end
```

imranismail:
I understand, that's what I did in `web.ex`. I just thought it would be natural to have it there since there's a html equivalent.

josevalim:
> I understand, that's what I did in web.ex.

Just to be clear: do not define functions in `web/web.ex`. There is a comment at the top explaining why. :)

Issue #1908, opened 2016-09-09
Phoenix.HTML.Safe not implemented for %{}


jpospychala:
hi,
I created model with map field, like:

mix phoenix.gen.html Model models name:string body:map

but the generated code doesnt't work:

$ mix test
.....

  1) test lists all entries on index (Website.ModelControllerTest)
     test/controllers/model_controller_test.exs:8
     ** (Protocol.UndefinedError) protocol Phoenix.HTML.Safe not implemented for %{}
     stacktrace:
       (phoenix_html) lib/phoenix_html/safe.ex:1: Phoenix.HTML.Safe.impl_for!/1
       (phoenix_html) lib/phoenix_html/safe.ex:15: Phoenix.HTML.Safe.to_iodata/1
       (website) web/templates/model/index.html.eex:18: anonymous fn/3 in Website.ModelView.index.html/1
       (elixir) lib/enum.ex:1623: Enum."-reduce/3-lists^foldl/2-0-"/3
       (website) web/templates/model/index.html.eex:14: Website.ModelView."index.html"/1
       (website) web/templates/layout/app.html.eex:20: Website.LayoutView."app.html"/1
       (phoenix) lib/phoenix/view.ex:335: Phoenix.View.render_to_iodata/3
       (phoenix) lib/phoenix/controller.ex:641: Phoenix.Controller.do_render/4
       (website) web/controllers/model_controller.ex:1: Website.ModelController.action/2
       (website) web/controllers/model_controller.ex:1: Website.ModelController.phoenix_controller_pipeline/2
       (website) lib/website/endpoint.ex:1: Website.Endpoint.instrument/4
       (website) lib/phoenix/router.ex:261: Website.Router.dispatch/2
       (website) web/router.ex:1: Website.Router.do_call/2
       (website) lib/website/endpoint.ex:1: Website.Endpoint.phoenix_pipeline/1
       (website) lib/website/endpoint.ex:1: Website.Endpoint.call/2
       (phoenix) lib/phoenix/test/conn_test.ex:224: Phoenix.ConnTest.dispatch/5
       test/controllers/model_controller_test.exs:9: (test)

underlying database is postgres

$ mix deps
* connection 1.0.4 (Hex package) (mix)
  locked at 1.0.4 (connection) a1cae722
  ok
* fs 0.9.1 (Hex package) (rebar)
  locked at 0.9.2 (fs) ed17036c
  ok
* gettext 0.11.0 (Hex package) (mix)
[...]

PR #1913, opened 2016-09-12
Envisage field of type map in phoenix.gen

shhavel:
#1908

### DIFF
--- a/lib/mix/tasks/phoenix.gen.html.ex
+++ b/lib/mix/tasks/phoenix.gen.html.ex
@@ -128,6 +128,8 @@ defmodule Mix.Tasks.Phoenix.Gen.Html do
         {label(key), ~s(<%= datetime_select f, #{inspect(key)}, class: "form-control" %>), error(key)}
       {key, :naive_datetime} ->
         {label(key), ~s(<%= datetime_select f, #{inspect(key)}, class: "form-control" %>), error(key)}
+      {key, :map} ->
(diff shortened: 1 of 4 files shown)


josevalim:
Thank you @shhavel. I am wondering if we should simply omit the map field. A text input into the data field is likely not the best way to handle a map field but on the other hand I like this approach because it shows at least *one* way of working with maps.

Thoughts?

shhavel:
@josevalim I think it is good to point developers to a working example. Map field can also be omitted of course but then you should expect questions like "Why map field is omitted in generated views?".
I suggest to keep it - maybe later someone will come up with another idea and will refactor.

Thanks

shhavel:
@josevalim 
As an alternative `textarea` can be generated for a map field in `phoenix.gen`.
Related required changes:
 - define protocol `Phoenix.HTML.Safe` for `Map` in `phoenix_html`
 - ability to cast JSON `String` into `Map` in `Ecto.Type`

If this is more appropriate solution I can do it.
Thanks

josevalim:
I don't think implementing the HTML safe protocol for
Maps is the way to go. Most will rather traverse it by hand. So I
definitely prefer the current solution.


-- 


josevalim:
Since the type does not integrate it with through all layers, we have decided it is best to add the map type by hand for now, thank you!

MeerKatDev:
it's 2021, and no one seems to know how to handle that, still. is there any working example around?

josevalim:
You can still add the field manually to your schema:

    field :my_field, :map

And similar in your migration.

The discussion here is mostly that, in order to fully automated, we need to include an approach to show and edit a map in the UI, and this is very specific to the purpose the map is being used.

PR #1916, opened 2016-09-15
Add entryPoints comment to brunch-config.js

dropfen:
This comment will not hurt, but save some extra time for people who are not used to brunch.

I've had some misunderstanding with this thing, where I flooded some other thread with it. I hope, this comment will help some people with this essential brunch config opportunity.

### DIFF
--- a/installer/templates/static/brunch/brunch-config.js
+++ b/installer/templates/static/brunch/brunch-config.js
@@ -18,6 +18,11 @@ exports.config = {
       //     "web/static/vendor/js/bootstrap.min.js"
       //   ]
       // }
+      //
+      // Or, you just bundle your application to a single file based on your imports/requires
+      // entryPoints: {
+      //   "web/static/js/app.js": "js/app.js"
+      // }
     },
     stylesheets: {
       joinTo: "css/app.css",


chrismccord:
Thanks! This is what the `autoRequire` feature does as far as I can tell, so I would prefer we stick with the existing config.

josevalim:
Chris, from what I understand, the entryPoint feature makes brunch behave
more like webpack. Instead of automatically including everything based on
the path, it will include only what we explicitly import. I would honestly
try to migrate everything to entry points on the new generators since the
entryPoint approach is much simpler conceptually.

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

dropfen:
Yeah, exactly. Its just simpler and makes more sense for umbrella multi-apps:

```
      'entryPoints': {
        'apps/app/static/js/index.js': 'js/app/bundle.js',
        'apps/admin/static/js/index.js': 'js/admin/bundle.js'
      }
      ...
      autoRequire: {
        'js/app/bundle.js': ['apps/app/static/js/index'],
        'js/admin/bundle.js': ['apps/admin/static/js/index'],
      }
```

chrismccord:
@dropfen sorry, you were right! I will be including these changes in the new generator PR. You can follow along there. Thanks!

dropfen:
cool, thx!

PR #1917, opened 2016-09-15
Replace endpoint manual builder impl with Plug.Builder

iamvery:
## Background

While recently studying the innards of Phoenix, I noticed that there appears to be partial-implementations of Plug.Builder in a few places. Namely the endpoint and controller code. I thought router may be at fault too, but realized it's a little different with pipelining.

## Solution

The endpoint work was straightforward with a couple small caveats. However, the controller proved me be trickier due to the instrumentation that's happening. Rather than trying to figure that all out, I've opted to offer the endpoint changes which may spark inspiration to dig in other places.

## Caveats

Two of them:

1. As of 1.2, Plug.Builder raises an error when nothing is plugged. By default Phoenix.Endpoint has no plugs. In order to keep it happy, a "dummy" function must be plugged. However, this [has been removed from Plug.Builder](https://github.com/elixir-lang/plug/pull/450).
2. Using Plug.Builder also imports Plug.Conn functions which may be undesirable.

## Summary

What do you think? It's nice to reduce the code and eliminate duplication implementation. Is it a valuable change?

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -435,22 +435,13 @@ defmodule Phoenix.Endpoint do
 
   defp plug() do
     quote location: :keep do
-      @behaviour Plug
+      use Plug.Builder
       import Phoenix.Endpoint
 
-      Module.register_attribute(__MODULE__, :plugs, accumulate: true)
       Module.register_attribute(__MODULE__, :phoenix_sockets, accumulate: true)
 
       if force_ssl = Phoenix.Endpoint.__force_ssl__(__MODULE__, var!(config)) do
         plug Plug.SSL, force_ssl
-      end
-
-      def init(opts) do
-        opts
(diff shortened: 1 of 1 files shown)


lib/phoenix/endpoint.ex
```diff
@@ -435,24 +435,15 @@ defmodule Phoenix.Endpoint do
 
   defp plug() do
     quote location: :keep do
-      @behaviour Plug
+      use Plug.Builder
```
iamvery: **Note**: by using Plug.Builder, the Plug.Conn functions are imported. If this is undesirable, we can `import Plug.Conn, only: []` after using Plug.Builder.

josevalim:
Thank you! I think we should wait until there is a new Plug out so we don't need to add workarounds. :)

josevalim:

PR #1921, opened 2016-09-19
Add conn to runtime data in render instrumentation

alanpeabody:
Hi, I was just digging into the instrumentation and thought this would be a nice addition to the event data for `:phoenix_controller_render`. In particular I think it would enable correlation of render/call events and make it much simpler to get any custom data to the render instrumentation in the future.

If this is acceptable I will also update the Endpoint documentation.

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -637,7 +637,7 @@ defmodule Phoenix.Controller do
     view = Map.get(conn.private, :phoenix_view) ||
             raise "a view module was not specified, set one with put_view/2"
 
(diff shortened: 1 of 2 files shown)


josevalim:
@whatyouhide can you think of any reason for us to *not* do this?

whatyouhide:
@josevalim not really, I mean this data is gonna go directly to each instrumenter but instrumenters can just ignore it. Can't see downsides :)

josevalim:
@alanpeabody please add docs and we should be good to go! 🎉

alanpeabody:
I don't know exactly why the tests failed, but I believe it is un-related to this PR.

Issue #1876, opened 2016-08-21
Add mix task to clean older versions of assets


aaronjensen:
## Problem

Depending on your method of deployment, asset directories can either grow indefinitely or be emptied unsafely and prematurely.

This will lead to either running out of disk space (and then needing to do an unsafe purge of some sort) or users not being served older assets that they may need while running an older version of a single page app with dynamic asset loading.

Here are two scenarios where the strategy of deploying an entirely new priv/static directory is broken:

### Single Page App

1. User A loads Version A of a SPA and downloads Asset A1.
1. You deploy Version B of your SPA which includes Asset B1 and B2.
1. User B navigates to another "page" in your single page app, which triggers a load of Asset A2. 1. They get a 404 because you are currently serving Version B which does not include old assets.

### Regular apps (narrow window)

1. User A loads Version A of your app
1. Just as they're loading it, you're deploying Version B which includes a new logo file, Logo B.
1. User A's browser tries to load Logo A, but it's 404 because only Logo B is available.

Both of these scenarios could be mitigated with a CDN as long as someone has loaded the assets at least once. We cannot rely on that, however. 

Both scenarios could also be mitigated by uploading and serving assets from something like S3, but that should be an unnecessary piece of infrastructure, not a necessary one.

See initial discussion here: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/phoenix-core/QRU-WUogUyI/ZG13JK3xDQAJ

## How does Rails solve this?

Rails solves this by storing more information in their manifest, specifically:

1. Old files
2. mtimes of those files

They then provide a rake task that can delete anything older than the past N versions as long as it is also older than T seconds ago.

## Proposed Solution

- [ ] Mimic Rails/Sprocket's manifest file format, at least the parts that we need:

    ```json
    {
      "files": {
        "application-8c357e6c2c428e1cb27cf646c6882f7e79326fbb63dc6602997f42d12f273ede.js": {
          "logical_path": "application.js",
          "mtime": "2016-08-21T07:55:52-07:00",
          "size": 369053,
          "digest": "8c357e6c2c428e1cb27cf646c6882f7e79326fbb63dc6602997f42d12f273ede",
          "integrity": "sha256-jDV+bCxCjhyyfPZGxogvfnkyb7tj3GYCmX9C0S8nPt4="
        },
        "application-af04b226fd7202dfc532ce7aedb95a0128277937e90d3b3a3d35e1cce9e16886.css": {
[...]

PR #1922, opened 2016-09-19
Improve asset manifest

RobinClowers:
This the first half a fix for https://github.com/phoenixframework/phoenix/issues/1876. Part 2 will be a new mix task which cleans old versions of assets.

### DIFF
--- a/lib/phoenix/digester.ex
+++ b/lib/phoenix/digester.ex
@@ -52,7 +52,7 @@ defmodule Phoenix.Digester do
                    manifest_join(file.relative_path, file.digested_filename))
     end)
 
-    manifest_content = Poison.encode!(entries, [])
+    manifest_content = Poison.encode!(%{latest: entries}, [])
     File.write!(Path.join(output_path, "manifest.json"), manifest_content)
 
     entries
--- a/lib/phoenix/digester.ex
+++ b/lib/phoenix/digester.ex
@@ -31,7 +31,7 @@ defmodule Phoenix.Digester do
         |> filter_files
         |> Enum.map(&digest/1)
 
-      digests = load_digests(input_path, output_path)
+      digests = load_digests(output_path)
       manifest = generate_manifest(digested_files, digests, output_path)
 
       Enum.each(digested_files, &(write_to_disk(&1, manifest, output_path)))
@@ -56,8 +56,8 @@ defmodule Phoenix.Digester do
     |> Enum.map(&(map_digested_file(&1, output_path)))
   end
 
-  defp load_digests(input_path, output_path) do
-    manifest_path = Path.join(input_path, "manifest.json")
+  defp load_digests(output_path) do
+    manifest_path = Path.join(output_path, "manifest.json")
     if File.exists?(manifest_path) do
       manifest_path
       |> File.read!
--- a/lib/phoenix/digester.ex
+++ b/lib/phoenix/digester.ex
@@ -30,7 +30,8 @@ defmodule Phoenix.Digester do
         |> filter_files
         |> Enum.map(&digest/1)
 
-      manifest = generate_manifest(digested_files, output_path)
+      digests = load_digests(input_path) || %{}
+      manifest = generate_manifest(digested_files, digests, output_path)
 
       Enum.each(digested_files, &(write_to_disk(&1, manifest, output_path)))
     else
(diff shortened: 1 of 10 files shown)


RobinClowers:
Not sure why the PR build is failing, it passed before: https://travis-ci.org/phoenixframework/phoenix/builds/161093495. I fixed a warning and forced pushed and now this unrelated test is failing.

RobinClowers:
@josevalim Here are the changes to the asset manifest we discussed, I'll be opening another PR with the mix task shortly.

lib/phoenix/digester.ex
```diff
@@ -46,18 +47,48 @@ defmodule Phoenix.Digester do
...
+      |> generate_new_digests
+      |> Map.merge(digests)
+    manifest_content = Poison.encode!(%{latest: entries, version: 1, digests: digests}, [])
     File.write!(Path.join(output_path, "manifest.json"), manifest_content)
 
     entries
   end
 
+  defp generate_new_digests(files) do
+    Enum.reduce(files, %{}, fn (file, acc) ->
```
aaronjensen: `Map.new/2` would be nice here

lib/phoenix/digester.ex
```diff
@@ -68,17 +99,23 @@ defmodule Phoenix.Digester do
   end
 
   defp map_file(file_path, input_path) do
+    {:ok, stats} = File.stat(file_path)
     %{absolute_path: file_path,
       relative_path: Path.relative_to(file_path, input_path) |> Path.dirname(),
       filename: Path.basename(file_path),
+      mtime: stats.mtime,
```
aaronjensen: I believe rails just uses the current time for new files. It doesn't read mtime from the file itself. I vaguely remember that they used to do that and changed it for some reason. I could be wrong on that, but the safest thing may be to do what they do.

lib/phoenix/digester.ex
```diff
@@ -46,18 +47,48 @@ defmodule Phoenix.Digester do
     |> Enum.map(&(map_file(&1, input_path)))
   end
 
-  defp generate_manifest(files, output_path) do
+  defp load_digests(input_path) do
+    manifest_path = Path.join(input_path, "manifest.json")
```
aaronjensen: I believe the `manifest.json` is in the output path, where this ultimately writes it to, which means that this doesn't need the `input_path` which means it can probably be called only by `generate_manifest` (which can then take its original two arguments only.

lib/phoenix/digester.ex
```diff
@@ -46,18 +47,48 @@ defmodule Phoenix.Digester do
...
     entries = Enum.reduce(files, %{}, fn (file, acc) ->
       Map.put(acc, manifest_join(file.relative_path, file.filename),
                    manifest_join(file.relative_path, file.digested_filename))
     end)
 
-    manifest_content = Poison.encode!(entries, [])
+    digests =
+      files
+      |> generate_new_digests
+      |> Map.merge(digests)
```
aaronjensen: It might be nice to name this `old_digests` so it's easier to follow

lib/phoenix/digester.ex
```diff
@@ -46,18 +47,48 @@ defmodule Phoenix.Digester do
...
 
-  defp generate_manifest(files, output_path) do
+  defp load_digests(input_path) do
+    manifest_path = Path.join(input_path, "manifest.json")
+    if File.exists?(manifest_path) do
+      manifest_path
+      |> File.read!
+      |> Poison.decode!
+      |> Access.get("digests")
+    end
```
aaronjensen: I believe this needs to return `%{}` and never nil if I'm reading correctly. This means that it needs an else clause and it needs to account for the file format: `Access.get("digests", %{})`

test/phoenix/digester_test.exs
```diff
@@ -28,7 +28,34 @@ defmodule Phoenix.DigesterTest do
       Path.join(output_path, "manifest.json")
       |> File.read!()
       |> Poison.decode!()
-    assert json["phoenix.png"] =~ ~r"phoenix-[a-fA-F\d]{32}.png"
+
+    assert json["latest"]["phoenix.png"] =~ ~r"phoenix-[a-fA-F\d]{32}.png"
+    assert json["version"] == 1
+  end
+
+  test "includes existing digests in new manifest" do
```
aaronjensen: I'd suggest adding a test for upgrading from the old manifest format.

lib/phoenix/digester.ex
gringocl: You are correct.  Sprockets uses `Time.now`.  I've switched over the new digest mapping to use the current time but am still using the file `mtime` in the case of files added switching from the old manifest format.

lib/phoenix/digester.ex
gringocl: This is correct, however in the case that we are building a digest from an old manifest file we still need the `input_path` to read existing digested files.

aaronjensen:
I would use Now for those too. Safer.

On Thu, Sep 22, 2016 at 10:46 AM, Miles Starkenburg <notifications@github.com> wrote:
@gringocl commented on this pull request.




In lib/phoenix/digester.ex [https://github.com/phoenixframework/phoenix/pull/1922] :

>      %{absolute_path: file_path,
       relative_path: Path.relative_to(file_path, input_path) |> Path.dirname(),
       filename: Path.basename(file_path),
+      mtime: stats.mtime,


You are correct. Sprockets uses Time.now . I've switched over the new digest mapping to use the current time but am still using the file mtime in the case of files added switching from the old manifest format.

lib/phoenix/digester.ex
gringocl: 💚 5e30aee2aa5f2bbd4bb2e74d543b59d9332dd223

lib/phoenix/digester.ex
gringocl: 💚  5e30aee2aa5f2bbd4bb2e74d543b59d9332dd223

lib/phoenix/digester.ex
gringocl: 💚  5e30aee2aa5f2bbd4bb2e74d543b59d9332dd223

test/phoenix/digester_test.exs
gringocl: 💚 5e30aee2aa5f2bbd4bb2e74d543b59d9332dd223

lib/phoenix/digester.ex
aaronjensen: Why is that? existing digested files should be in the output path as well, right?

gringocl: Nevermind me. I was mistaken!

gringocl: 💚  1916513ba907518ada8940fd70171dcb4492ddac

josevalim:

PR #1923, opened 2016-09-19
Add mix digest clean

RobinClowers:
This is a follow on to https://github.com/phoenixframework/phoenix/pull/1922, and should fix https://github.com/phoenixframework/phoenix/issues/1876. It depends on the previous PR, so the diff you see here is for both. The changes here are all in this commit: 4b4ed06.


josevalim:

Issue #1742, opened 2016-06-03
Consider adding drainstop to Phoenix.Endpoint.Handler behaviour


chrismccord:
Ref https://gist.github.com/aaronjensen/33cc2aeb74746cac3bcb40dcefdd9c09

We'd need to make this part of the behavior that can be optionally supported by adapters, and it's also worth considering as part of plug.

jquadrin:
Given that ranch is OTP compliant we can just specify a shutdown option for the child specification of the connection supervisor in Ranch. This will follow standard shutdown behavior, `:inifinity` isn't the best idea for graceful shutdown of workers but a user-specified wait time would suffice.

aaronjensen:
My understanding is that when shutting down the [ranch_cons_sup](https://github.com/ninenines/ranch/blob/master/src/ranch_conns_sup.erl) sends an [exit signal to its children](https://github.com/ninenines/ranch/blob/master/src/ranch_conns_sup.erl#L245) which, as I understand it, if the [children are not trapping exists, they will be killed immediately](https://groups.google.com/d/msg/phoenix-core/ig8SrIn_uUQ/NH3GtgskBgAJ). In other words, you have to actually wait for the children to die themselves. Maybe I'm misunderstanding something though.

I'm not sure where the mention of `:infinity` is coming from, the gist has an explicit timeout that defaults to `5000`.

jquadrin:
Yes, if you are not trapping exits then you cannot depend on ranch's graceful shutdown which follows normal OTP behavior. 

Your workaround is required because of [these points](https://groups.google.com/d/msg/phoenix-core/ig8SrIn_uUQ/CqiL086uBwAJ).

chrismccord:
We are going to hold on this one and for now the best solution is to release it as a separate package for those that need it. Because it requires trapping exits of all requests, we feel the tradeoffs need to be carefully considered so it's not something that should be in core. Thanks!

aaronjensen:
Ok, sounds good. For what it's worth, it doesn't actually require trapping exits of all requests if you want to do it in the way that I did it, which seems to work just fine even if it's not the most "clean" thing in the world.

derekkraan:
Maybe [this small package](https://hex.pm/packages/ranch_connection_drainer) will help someone trying to get connection draining working in Phoenix.

PR #1925, opened 2016-09-20
Support graceful close

fishcakez:


josevalim:
Beautiful! Maybe we should make it part of plug and then change Phoenix to
use it?

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

fishcakez:
I didn't make this PR to plug because there would be edge cases where it could break when starting cowboy as part of the ranch supervision tree.

fishcakez:
Also note that can not guarantee it will work in future versions of ranch/cowboy, i.e. cowboy 2.

josevalim:
@fishcakez so should we add even add it to Phoenix? :)

fishcakez:
@josevalim I don't know :smile:, last time this came up was https://github.com/phoenixframework/phoenix/issues/1742. The method here is reliable and would always occur before exit signals so no need to trap. However both rely equally on ranch listener sub tree, which was not discussed.

fishcakez:
I think this is too fragile in the longer term for the core maintainers to guarantee continuous support. I think it would better to contribute this to ranch 1.2/cowboy 2.0 and await plug support for that. Certainly it would be nice to see some contribution in that direction. I will leave the branch on my github so it can be cherry-pick'ed by anyone in need of the feature - there will be no support for it from me.

PR #1926, opened 2016-09-20
Future JavaScript, done right

nicholaswyoung:
For starters, [many files are automatically ignored by npm](https://docs.npmjs.com/files/package.json), so we don't need to manually specify the `files` that we wish to include.

Secondly, `main` and `jsnext:main` are all that's required of phoenix.js. These basic fields in `package.json` allow us to be compatible with the widest selection of client side bundlers.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,9 +1,10 @@
 {
   "name": "phoenix",
-  "version": "1.2.1",
+  "version": "1.2.2",
   "description": "The official JavaScript client for the Phoenix web framework.",
   "license": "MIT",
   "main": "./priv/static/phoenix.js",
+  "jsnext:main": "./web/static/js/phoenix.js",
   "repository": {
     "type": "git",
     "url": "git://github.com/phoenixframework/phoenix.git"
@@ -15,7 +16,6 @@
     "babel-brunch": "~6.0.0",
     "uglify-js-brunch": "~2.0.1"
   },
-  "files": ["README.md", "LICENSE.md", "package.json", "priv/static/phoenix.js", "web/static/js/phoenix.js"],
   "scripts": {
     "test": "./node_modules/.bin/mocha ./web/test/**/*.js --compilers js:babel-register"
   }


package.json
```diff
@@ -15,7 +16,6 @@
     "babel-brunch": "~6.0.0",
     "uglify-js-brunch": "~2.0.1"
   },
-  "files": ["README.md", "LICENSE.md", "package.json", "priv/static/phoenix.js", "web/static/js/phoenix.js"],
```
josevalim: Wouldn't it be better to keep it explicit anyway?

chrismccord: These are my thoughts as well. A single line in the config that we almost never need to update seems like a  win as far as what's being included?

nicholaswyoung: Not really. The above lines (`main` and `jsnext:main`) are all that's needed. This is duplication, actually.

`README` (and it's variants) are already picked up by default (and always have been). `package.json` has to be part of the bundle (per the spec), and the other files are extraneous.

josevalim: @nicholaswyoung then we are subject to whatever is considered extraneous by npm. I would prefer to be explicit here, it is only a single line of code, and it will save us from accidentally including all of Phoenix one day in npm.

nicholaswyoung: @josevalim That concern is imagined, IMO, but fine. This is the last npm related issue I'll bring to your attention, if at all possible.

josevalim:
@nicholaswyoung sorry, can you please give some feedback on where this discussion went sour? The pull request is welcome. Why is our interest in keeping one line of code explicit reason enough to close the pull request altogether?

chrismccord:
@nicholaswyoung can you re-open and amend with the jsnext change?

package.json
josevalim: > That concern is imagined, IMO, but fine.

Btw, I just tried it locally, and removing the "files" configuration includes all of the files in lib, which are Elixir files, in the npm package.

nicholaswyoung: @josevalim I stand corrected, then.

nicholaswyoung:
@josevalim Sorry, I don't mean to "leave in a huff." I do, however, feel that it's been an uphill climb for anything npm related. Coming over from the npm community, I hoped to help the project become a slightly better citizen of that community, having experienced frustration from that perspective.

What I've been met with is some willingness, yes, but also a fair bit of opposition. It took several attempts before I was able to get the module published, and I don't think it's been smooth at all since.

It seems (to me) that attempts to enhance the npm experience for consumers of phoenix.js are met with the opinion that we need to do a bare minimum for that community, favoring Elixir convention where it may differ from JS, and never questioning further. It comes off as disdain for a inferior language, which irks me. As a Elixirist who came over from the JS world, I don't have that disdain &mdash; or whatever it is &mdash; and tend to see npm as an ecosystem where a different set of best practices apply (ones that are relevant to their community).

I guess it's the pretense that everything regarding npm is to be viewed with suspicion, or at very least is unnecessary (as I've been told several times, too). I want everyone from the JavaScript community writing Elixir too!

@chrismccord I'll try to get to that tomorrow. It's time to wind down for the night.

nicholaswyoung:
@josevalim I guess I'm open to contributing in the future, so long as I can be assured this isn't (and won't be) the dominant attitude going forward. That we can have technical discussions that take into account correctness for the relevant community.

If that approach is at odds with the team's, then there's an incompatibility. But if I go into this thinking everything is an automatic no, then why even try.

josevalim:
> What I've been met with is some willingness, yes, but also a fair bit of opposition. It took several attempts before I was able to get the module published, and I don't think it's been smooth at all since.

I agree there was a lot of opposition early on but my understand is that our attitude started to change since a couple important events happened on the JavaScript side of things, such as Bower getting retired, NPM support being added to Brunch, the snatching of the phoenix name on NPM, etc.

I also see the disdain aspect and I understand how that is frustrating. It is something we should improve at. However, I don't believe anyone acted with disdain at any moment during this particular discussion. The reason to keep "files" was never suspicion of JavaScript or NPM. We [do the same for Hex too](https://github.com/phoenixframework/phoenix/blob/master/mix.exs#L62-L63). Being explicit is a trait we promote in general, and it is going to show up in both Elixir and JavaScript code.

nicholaswyoung:
Oh no, this discussion was fairly civil. But you can't deny that the attitude is commonplace, even boosted by prominent community members.

Churn isn't a good thing in the JavaScript community, but it's essential. I wouldn't blame your skepticism on the community at large and changes it hnderwent, when it isn't the root cause. Attitude towards people should be tops, even if you disagree with what a community is currently doing. (That point is kinda irrelevant, but understandable.)

I'm honestly still going to keep contributions to a minimum, while this attitude change proves itself. But I'm grateful that you acknowledged that we can all do better, which is indeed true.

PR #1927, opened 2016-09-20
Enhance new generators

chrismccord:
WIP
- [x] Add MyApp.Application
- [x] ensure all model references are gone
- [x] remove errors_on
- [x] add `cd assets` to npm install instructions
- [x] `mix phoenix.routes` should check for `MyApp.Web.Router` and fallback to `MyApp.Router` 
- [x] ensure repo is added to ecto_repos, in apps/my_app_web
- [x] add phx.new.web / phx.new.ecto
- [x] add phx.gen.schema (needs docs)
- [x] add phx.gen.html (needs docs)
- [x] add mix phx.gen.json
- [x] add action_fallback
- [x] add function plug support to action_fallback
- [x] normalize MyApp.Web app module across single and umbrella apps
- [x] ensure gen.\* cannot run from umbrella root
- [x] Remove shortdoc + deprecate 1.2 generators
- [ ] reconcile --module and --app options when combined with --umbrella
- [x] add phx.gen.channel|presence|secret


antipax:
As discussed in IRC, `mix phoenix.routes` is still looking for `MyApp.Router` rather than `MyApp.Web.Router`.

chrismccord:
Thanks for bringing that up! `mix phoenix.route`s will check for `MyApp.Web.Router` and fallback to `MyApp.Router`

installer/templates/phx_umbrella/apps/app_name_web/lib/templates/page/index.html.eex
```diff
@@ -1,5 +1,36 @@
-<div id="container" class="row marketing">
+<div class="jumbotron">
+  <h2><%%= gettext "Welcome to %{name}", name: "Phoenix!" %></h2>
```
jeroenvisser101: Maybe leave the exclamation mark inside the text to translate (Spanish for instance does special things with that)?

josevalim:
¡Si! Good call.

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

installer/templates/phx_umbrella/apps/app_name_web/lib/templates/page/index.html.eex
```diff
@@ -1,5 +1,36 @@
-<div id="container" class="row marketing">
+<div class="jumbotron">
+  <h2><%%= gettext "Welcome to %{name}!", name: "Phoenix!" %></h2>
```
jeroenvisser101: @chrismccord sorry for nitpicking, but this one now has double

installer/lib/phx_new/app.ex
```diff
@@ -0,0 +1,43 @@
+defmodule Mix.Tasks.Phx.New.App do
+  use Mix.Task
+  use Mix.Tasks.Phx.New.{Project, Generator}
+
+  @prefix "phx_umbrella/apps/app_name"
```
OvermindDL1: This is called @prefix....

installer/lib/phx_new/app.ex
```diff
@@ -0,0 +1,43 @@
...
+
+  template :new, [
+    {:eex, "#{@pre}/config/config.exs",      "apps/app_name/config/config.exs"},
```
OvermindDL1: Yet, these are called @pre?

chrismccord: @OvermindDL1 this was a 2am checkpoint commit last night. Untested and not currently building. Thanks for taking a look! :)

niccolox:
hi, 

any ideas on when this branch will become a full version release? 

when its ready? 

is there a timebox? a set cadence?

chrismccord:
> when its ready?

You knew the answer. I'm working on it as we speak. Will ship when ready, no sooner, no later! :)

> is there a timebox? a set cadence?

Anytime I have that's not client time, family time, general free time, or sleeping

niccolox:
just comparing the branches now, looks great

shankardevy:
How to try the new `phx.gen.schema` generator?

I removed my stable `phoenix.new` archive from `~/.mix/archive` folder and created a new phoenix umbrella project from the installer folder in phoenix source code folder. However, since the new project created using the `phx.new` command still refers to the stable version of Phoenix in its mix file, I don't get the new mix commands like `phx.gen.schema` in my newly created project. 

As a workaround, I modified the `mix.exs` in my apps/my_app_web and changed the phoenix dependency to use my local phoenix folder. Now I get the new `phx.gen.schema` from the root of the umbrella project but I still don't get that task when I am in `apps/my_app` which doesn't have `phoenix` as dependency. 

I know this new generator is WIP for 1.3 and is not ready. Still any help to test out a dev version?

chrismccord:
The gen.schema|html|json generators only work today in the context of a single application with a dependency on phoenix. A later release with smarter multi-application aware generators will be explored for umbrellas, but 1.3 will ship with html/json/schema generators strictly in the context of a single appliaction.

lib/phoenix/test/conn_test.ex
```diff
@@ -538,6 +538,31 @@ defmodule Phoenix.ConnTest do
...
+  Uses the provided `%Plug.Conn{}`s router matched in the previous request.
+  Raises if the response's location header is not set.
+
+  ## Examples
+
+      assert redirected_to(conn) =~ "/posts/123"
+      assert %{id: "123"} = redirected_params(conn)
+  """
+  @spec redirected_params(Conn.t) :: map
+  def redirected_params(%Plug.Conn{} = conn) do
```
chrismccord: cc @josevalim (ignore my last mention as it shows the entire PR)

lib/phoenix/router.ex
```diff
@@ -357,6 +358,10 @@ defmodule Phoenix.Router do
                  unquote(exprs.host)) do
         unquote(exprs.dispatch)
       end
+      def __match_params__(_conn, unquote(exprs.verb_match), unquote(exprs.path),
+                 unquote(exprs.host)) do
+        unquote(exprs.binding)
+      end
```
chrismccord: It feels like we should standardize privately injected apis under the `__phoenix__(..` or `__router__` convention, but the ship may have sailed with our other injections at this point, i.e. `__routes__`. Thoughts?

lib/phoenix/test/conn_test.ex
josevalim: :+1:

lib/phoenix/test/conn_test.ex
```diff
@@ -538,6 +538,31 @@ defmodule Phoenix.ConnTest do
...
+    router = Phoenix.Controller.router_module(conn)
+    %URI{path: path, host: host} = conn |> redirected_to() |> URI.parse()
```
josevalim: If `host` is nil then we should do `host = host || conn.host`.

lib/phoenix/router.ex
```diff
@@ -357,6 +358,10 @@ defmodule Phoenix.Router do
                  unquote(exprs.host)) do
         unquote(exprs.dispatch)
       end
+      def __match_params__(_conn, unquote(exprs.verb_match), unquote(exprs.path),
```
josevalim: I don't like that we are making the router larger by adding all of those new clauses. But it seems we can't use the dispatch one because it invokes the pipe_through stuff?

About the name, one other option is to make this API public. The reason I like `__routes__` is because if one day we want to move the generators out of Phoenix it won't be a problem.

wojtekmach:
@chrismccord is `--no-gettext` option up for grabs? let me know if/when you need any help with it, I'd be happy to give it a shot.

josevalim:
I think we should honestly move `--no-gettext` to a separate issue and we can discuss its merits once this is done. :)

shankardevy:
>  ensure gen.* cannot run from umbrella root

Since `gen.*` mix tasks are part of `phoenix`, there is no way to call `gen.schema` from the ecto app in umbrella. Since this is a restriction from how mix tasks work, I think `gen.*` should be allowed to run from umbrella root. 

Pseudo code as follows

```elixir
# from umbrella root
mix phx.gen.schema Account User users name  # creates schema in apps/ecto_app/lib/

# from umbrella root
mix phx.gen.schema Account User users name --ecto-app=my_ecto_app # creates schema in apps/my_ecto_app/lib

```

The same way `phx.gen.html` and `phx.gen.json` can also automatically use the default naming convention to find the `ecto_app` or use `--ecto-app` to override the default name.

What do you think of this? If this is acceptable, I can take a stab at submitting a PR.

chrismccord:
@wojtekmach we'll wait on no-gettext for the future issue, but what I could use is a thorough review of the PR, along with testing out the new features under different scenarios. Making sure aliases are correct, generated tests pass, etc. Everything is shaping up on my end and it's time to do final sweep to find any rough edges. Let me know if you come across anything. Thanks!

installer/lib/phx_new.ex
```diff
@@ -46,123 +49,87 @@ defmodule Mix.Tasks.Phx.New do
...
+      extra = if mix?, do: [], else: ["$ mix deps.get"]
 
-        if binding[:ecto], do: print_ecto_info()
+      brunch? = maybe_cd(project.web_path, fn -> install_brunch(install?) end)
 
-        if not brunch?, do: print_brunch_info()
-      end)
+      print_mix_info(project.project_path, extra)
+      if Project.ecto?(project), do: print_ecto_info()
+      if not brunch?, do: print_brunch_info(project, starting_dir)
```
wojtekmach: I noticed this line is causing issues when doing:

```
cd apps && mix phx.new.ecto hello
```

```
    lib/phx_new.ex:129: anonymous fn/3 in Mix.Tasks.Phx.New.prompt_to_install_deps/1
    (elixir) lib/file.ex:1162: File.cd!/2
    (mix) lib/mix/task.ex:294: Mix.Task.run_task/3
    (mix) lib/mix/cli.ex:58: Mix.CLI.run_task/2
    (elixir) lib/code.ex:370: Code.require_file/2
```

(looks like `brunch?` is then `nil` and so it fails on `if not nil, do:`)

I think we should use `if Project.brunch?(project)` here instead however it probably still needs to be modified as it returns `true` when doing `phx.new.ecto`

wojtekmach:
I noticed an issue with running installer in "dev mode":

```
$ cd /path/to/phoenix
$ cd installer
$ mix phx.new dev_app --dev
Fetch and install dependencies? [Yn] (Y)
* error command failed to execute, please run the following command again after installation: "cd assets && npm install && node node_modules/brunch/bin/brunch build"
```

jeregrine:
I'm running through the different generators now as well FYI.

lib/mix/tasks/phx.gen.html.ex
```diff
@@ -0,0 +1,138 @@
+Code.require_file("../../../installer/lib/phx_new/generator.ex", __DIR__)
```
wojtekmach: @chrismccord regarding build failures, compiler is handing on these:

```
Compiling lib/mix/tasks/phx.gen.json.ex (it's taking more than 10s)
Compiling lib/mix/tasks/phx.gen.schema.ex (it's taking more than 10s)
Compiling lib/mix/tasks/phx.gen.html.ex (it's taking more than 10s)
```

I was able to reproduce this locally by doing `rm -rf _build && MIX_ENV=test mix compile`. I noticed that if I commented out everything in these 3 mix tasks except for `Code.require_file` line it was still hanging, but if I commented out this one as well it went through. Couldn't find the root cause, but maybe it will point you into the right direction.

jeregrine:
Noticed a few issues with alias's and module names in gen.json and gen.html. I will be submitting a pr shortly to fix those.

jeregrine:
Looks like the phx.gen.html|json are putting migrations in the priv directory of the `App.Web` app, and I believe they need to be in the priv directory of the top level `App` application.

lib/mix/tasks/phx.gen.html.ex
josevalim: We should not use require_file on compiled code. The installer directory is likely not going to be there once we push the package.

wojtekmach: right. I just double-checked something, and it seems compilation gets through if archive was installed (I used: `cd installer && MIX_ENV=prod mix archive.build && mix archive.install`). That's probably why I haven't seen it before.

installer/lib/phx_new/generator.ex
```diff
@@ -95,4 +88,187 @@ defmodule Mix.Tasks.Phx.New.Generator do
...
+    |> Enum.map(fn _ -> ".." end)
+    |> Path.join()
+  end
+
+  defp phoenix_path(_path, false) do
+    "deps/phoenix"
+  end
+
+  defp phoenix_dep("deps/phoenix"), do: ~s[{:phoenix, "~> 1.2.0"}]
+  # defp phoenix_dep("deps/phoenix"), do: ~s[{:phoenix, github: "phoenixframework/phoenix", override: true}]
```
chrismccord: indeed. Right now it should point at this PR branch. At the moment you have to manually update the generated project and point it, but I'll bump the dep to GitHub for now

priv/templates/phx.gen.html/schema_access.ex
```diff
@@ -0,0 +1,97 @@
...
+
+      iex> delete_<%= schema.singular %>(<%= schema.singular %>)
+      {:error, %Ecto.Changeset{}}
+  """
+  def delete_<%= schema.singular %>(%<%= inspect schema.alias %>{} = <%= schema.singular %>) do
+    Repo.delete(<%= schema.singular %>)
+  end
+
+  @doc """
+  Returns an `%Ecto.Changeset{}` for tracking <%= schema.singular %> changes.
```
chrismccord: I think it still makes sense to import since it may be common to expose functions that operate on changesets, but could go either way since it is not used by the generated code. // @josevalim

chrismccord:
@jeregrine For 1.3, the `gen.*` generators will not be multi-app aware. So if you run gen.html|json in apps/my_app_web/, then it will do the same thing as running the generator in a non-umbrella app where ecto is assumed to be in the app. After 1.3, I'd like to explore expanding these to support multi-app code generation, but it's tricky.

installer/lib/phx_new/umbrella.ex
```diff
@@ -1,138 +1,54 @@
...
-
-    {app, Module.concat([opts[:module] || Macro.camelize(app)]), app_path}
+  def prepare_project(%Project{app: app} = project) when not is_nil(app) do
+    project
+    |> put_app()
+    |> put_web()
+    |> put_root_app()
   end
+  defp put_app(project) do
+    project_path = Path.expand(project.base_path <> "_umbrella")
```
chrismccord: I went back and forth on the generated project name. The only other consideration was `foo_platform`. I can see how postfixing the name with _umbrella could be counterintuitive, but I think it's more important that we don't dup the app name as the root level folder name. // @josevalim

jeregrine:
@chrismccord ah, so this PR is just addressing the new structure change not including handling of umbrella applications. Should have read the above closer. Should we put a mention in the readme that if you are using the gen.* in umbrella applications it won't work correctly?

installer/lib/phx_new/umbrella.ex
wojtekmach: just to clarify, I think automatically adding a suffix works well, however it behaves counterintuitively when a given path already exists in the system (e.g. giving `~/src/foo` or `.`, I expect to generate project in these directories)

chrismccord:
Correct. To be clear, the extent of our umbrella handling in this PR (and for 1.3.0) is:

- support --umbrella in phx.new
- new `mix phx.gen.web` for generating a phoenix project inside an umbrella (no ecto)
- new `mix phx.gen.ecto` for generating an elixir project inside an umbrella, with ecto

The `mix phx.gen.html|json|schema` will not be "umbrella aware", but we want to explore it in a future release. So they will work correctly for umbrella apps, in that they will generate what the docs say when run within the application, but they won't try to be crafty and put files across multiple umbrella apps. So they would still be usable if you had an umbrella app using phoenix and ecto within the same project. Make sense?

jeregrine:
@chrismccord yep. They mostly work outside of that one PR, and the issue of the migrations going to the wrong spot. If we had a way to know if an app is in an umbrella and find its parent then it would Just Work TM. 

I will test the ecto and web routes further.

priv/templates/phx.gen.html/schema_access.ex
josevalim: @chrismccord I would also keep for backwards compat purposes. The imports won't hurt. :)

olivermt:
Hi guys!

https://github.com/phoenixframework/phoenix/blob/0f3472ca34adb30815f5a1891a8471b05bf7a1aa/priv/templates/phx.gen.json/controller.ex#L14-L17

You have forgotten to remove new() from the json controller generator it seems.

olivermt:
@chrismccord when you are already overhauling all this, what is the chance that you could make templates overrideable? I use the generators a LOT when prototyping and it would be such a great tool to be able to add in some standard security stuff, some different patterns on how updates works etc etc

josevalim:
@olivermt given this is already a lot of work, I don't think we should put even more work on top of it. 😃

olivermt:
@josevalim I would be interested in putting in the work if I just got instructions on where you would like to have it on a conceptual level. As in where would you put the config and how is it perpetuated down to the actual template rendering into a file etc.

I would be happy to take on the plumbing. Imagine being able to set a config :phoenix, generator: [json_template: "priv/my_ocd_pleasing_template.ex"]

chrismccord:
@olivermt the generators will check your apps priv dir for the template before using phoenix's, but it's not something that is well documented. It's also going to be brittle so I am okay improving the docs on how to override but we should make the docs clear that it's not something we will ensure backwards compatibility for.

olivermt:
That is *great* news! Thank you @chrismccord

lib/mix/tasks/phx.gen.channel.ex
```diff
@@ -0,0 +1,59 @@
+defmodule Mix.Tasks.Phoenix.Gen.Channel do
```
wojtekmach: @chrismccord I believe this should be `Mix.Tasks.Phx.Gen.Channel` to match the filename and the `@moduledoc` below should be updated as well

lib/mix/tasks/phx.gen.channel.ex
```diff
@@ -0,0 +1,59 @@
...
+  @moduledoc """
+  Generates a Phoenix channel.
+
+      mix phoenix.gen.channel Room
+
+  Accepts the module name for the channel
+
+  The generated model will contain:
+
+    * a channel in web/channels
```
wojtekmach: it should be `lib/web/channels`. I can fix this (and the mix task name I mentioned above) in a PR later this evening.

chrismccord: please do! Thanks

wojtekmach:
@chrismccord sorry if it's rehashing a discussion you already have had (I couldn't find anything about it online)

- In phoenix v1.3 we'll have: `mix phx.new my_app` => `cat my_app/lib/application.ex`
- In Elixir v1.4 we have: `mix new my_app --sup` => `cat my_app/lib/my_app/application.ex`

would it make sense for phx v1.3 match project structure generated by elixir v1.4? I kinda like shallower project structure tbh, but I also think there's some value in matching what mix generates by default.

lib/mix/tasks/phoenix.gen.json.ex
```diff
@@ -23,7 +19,10 @@ defmodule Mix.Tasks.Phoenix.Gen.Json do
   with `--no-model`. Read the documentation for `phoenix.gen.model`
   for more information on attributes and namespaced resources.
   """
+  use Mix.Task
+
   def run(args) do
+    IO.puts :stderr, "mix phoenix.gen.json is deprecated. Use phx.gen.html instead."
```
jamilabreu: Should this be: "Use phx.gen.json instead."?

chrismccord:
@wojtekmach the dir structure was discussed internally and we are moving forward with the shallower approach that doesn't dup the otp_app dir, with @josevalim's blessing.

josevalim:
@chrismccord wait, I could swear you were against the shallower approach and you convinced me to not make it shallow? To be clear, I thought we were going with `lib/foo_my_app/application.ex`?

Gazler:
@josevalim @chrismccord There was some discussion here about it https://github.com/phoenixframework/phoenix/pull/1871#issuecomment-239914269

chrismccord:
@josevalim haha, and I swear you specifically said the shallow approach is the way to go. I will make the change to lib/my_app/ to follow mix.

josevalim:
@chrismccord my concern was always nesting. Your concern was always being inconsistent. :) I think you convinced me at the end but I am unsure now. :P

test/mix/tasks/phoenix.gen.channel_test.exs
```diff
@@ -14,7 +14,7 @@ defmodule Mix.Tasks.Phoenix.Gen.ChannelTest do
   end
 
   test "generates channel" do
-    in_tmp "generates channel", fn ->
+    in_tmp "generates deprecated channel", fn ->
```
wojtekmach: damn that makes sense, sorry! :D

chrismccord: No worries. `in_tmp` has been pesky lately. I'm not sure why this would conflict given the tests were both async false, but giving it a different dir name resolved some test issues. Thanks for the PR!

chrismccord:
Dir structure now follows mix in all cases 👌🏻

jeregrine:
💖

PR #1931, opened 2016-09-21
fixed a typo

hykw:
Fixed a minor typo.


chrismccord:
`preceeded` is actually correct in this case. Thanks for the contribution!

PR #1932, opened 2016-09-21
Allow non-keyword list in Endpoint http config

talklittle:
For example:

      http: [{:port, 4000}, :inet6]


josevalim:
Thank you! Could you please write a test? It could be as simple as changing one of the endpoint configuration examples we have to include a non keyword configuration.

PR #1933, opened 2016-09-21
Allow non-keyword list in Endpoint http config

talklittle:
For example:

      http: [{:port, 4000}, :inet6]

Redo of #1932 plus a test


josevalim:

PR #1934, opened 2016-09-21
Installer: Listen on IPv6 on prod

talklittle:


installer/templates/new/config/prod.exs
```diff
@@ -12,7 +12,7 @@ use Mix.Config
 # manifest is generated by the mix phoenix.digest task
 # which you typically run after static files are built.
 config :<%= app_name %>, <%= app_module %>.Endpoint,
-  http: [port: {:system, "PORT"}],
+  http: [port: {:system, "PORT"}] ++ [:inet6],
```
josevalim: I believe having `:inet6` at the beginning is more readable, such as: `http: [:inet6, port: {:system, "PORT"}]`. Could you please amend this and the other code samples? Thank you!

talklittle: Fixed. I had no idea that syntax existed!

josevalim:

PR #1937, opened 2016-09-23
made missing parameter error message more human readable

jbhatab:
I was passing params to a create action and got this error.

screen shot 2016-09-22 at 7 37 06 pm

I eventually figured it out but found the error message to be unclear. I put what I thought was more human readable. Open to other message ideas or if I'm missing some common terminology concepts that I'm not considering.

### DIFF
--- a/lib/phoenix/controller/pipeline.ex
+++ b/lib/phoenix/controller/pipeline.ex
@@ -155,8 +155,8 @@ defmodule Phoenix.Controller.Pipeline do
 
   @doc false
   def __catch__(:error, :function_clause, controller, action,
-                [{controller, action, [%Plug.Conn{} | _], _loc} | _] = stack) do
-    args = [controller: controller, action: action]
+                [{controller, action, [%Plug.Conn{} = conn | _], _loc} | _] = stack) do
+    args = [controller: controller, action: action, params: conn.params]
     reraise Phoenix.ActionClauseError, args, stack
   end
   def __catch__(kind, reason, _controller, _action, stack) do
--- a/lib/phoenix/exceptions.ex
+++ b/lib/phoenix/exceptions.ex
@@ -44,8 +44,12 @@ defmodule Phoenix.ActionClauseError do
   def exception(opts) do
     controller = Keyword.fetch!(opts, :controller)
     action = Keyword.fetch!(opts, :action)
-    msg = "bad request to #{inspect controller}.#{action}, " <>
-          "no matching action clause to process request"
+    params = Keyword.fetch!(opts, :params)
+    msg = "Could not find a matching #{inspect controller}.#{action} clause " <>
+          "to process request. This typically happens when there is a " <>
+          "parameter mismatch but may also happen when any of the other " <>
+          "action arguments do not match. The request parameters are:\n\n" <>
+          "#{inspect params}"
     %Phoenix.ActionClauseError{message: msg}
   end
 end


chrismccord:
Thanks! This isn't necessarily only missing params. ActionClauseError is a failed match on the controller action, which could be missing params, failed match on a params value, or a failed match on the %Plug.Conn{}. It also will be raised for customized action clauses that may or may not contain the params as a argument at all. That said, I agree we can make it more helpful. How about?:

    no controller action defined that matches the provided arguments, 
    such as the %Plug.Conn{} or map of params

Please amend the PR and I'll get things merged. Thanks!

jbhatab:
@chrismccord I'm liking this direction. Maybe I'm a little too novice since I come from a rails background, but it is still a little confusing if that was the first time I saw that. I think it's the 'no controller action part that tricks my head. I immediately think that I'm just missing the method completely instead of just needing to tweak it or the request parameters.

```
controller action <action name> is undefined or the request parameters do not match the action parameters
```

I'm not sure if this is perfect, but thanks for the quick response.

lukerollans:
@jbhatab The error message might be raised for situations other than a parameters mismatch, though.

I think something along the lines of ..

> no controller action defined which matches the provided arguments,
> please ensure your parameters map and %Plug.Conn{} match an action clause

jbhatab:
@lukerollans What other situations are there? I figured since it was the Phoenix.MissingParamError that we could make it pretty specific.

lukerollans:
@jbhatab The situations which Chris mentioned above

josevalim:
Given that this error will most likely be raised due to a parameter
mismatch, why don't we focus on it but still acknowledge the remaining
cases?

    could not find a matching Controller.action clause to process request.
This typically happens when there is a parameter mismatch but may
also happen when any of the other action arguments do not match. The
request parameters are:

        #{inspect params}



-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

jbhatab:
@josevalim @chrismccord  Really loving that direction.  I pushed up a new change for this, but here's some remaining thoughts.

1. "This typically happens when there is a parameter mismatch but may also happen when any of the other action arguments do not match" Wouldn't parameter mismatch be the same thing as the second part of that sentence? Just not sure what the difference is.

2. I got parameters from the catch method in the absolute wonkiest way and I'm 100% sure I did it in a non ideal way. It got the error and I attached what it looks like below but not sure how I can extract params from stack in a more elegant way.

```
params = stack |> Enum.at(0) |> elem(2) |> Enum.at(0) |> Map.get(:body_params)
```

3. I'm really confused why me passing the wrong parameters is triggering the ActionClauseError. Shouldn't this be triggering the MissingParamError since its triggered by scrub_params? Or is that only if scrub_params is explicitly called?

Here's the new error in terminal.

screen shot 2016-09-23 at 4 09 32 pm

thanks again for helping out with this :DDDDD

shhavel:
@jbhatab 
> "This typically happens when there is a parameter mismatch but may also happen when any of the other action arguments do not match" Wouldn't parameter mismatch be the same thing as the second part of that sentence? Just not sure what the difference is.

`parameter mismatch` - mismatch of the `conn.params` - the second argument of controller action by default (if not overridden)
`any of the other action arguments` - other than second argument of controller action. This probably may happen only if controller action was overridden (customized action clauses)

> I got parameters from the catch method in the absolute wonkiest way and I'm 100% sure I did it in a non ideal way. It got the error and I attached what it looks like below but not sure how I can extract params from stack in a more elegant way.

> params = stack |> Enum.at(0) |> elem(2) |> Enum.at(0) |> Map.get(:body_params)

You can get parameters from `%Plug.Conn{}`:

```elixir
def __catch__(:error, :function_clause, controller, action,
              [{controller, action, [%Plug.Conn{} = conn | _], _loc} | _] = stack) do
  args = [controller: controller, action: action, params: conn.params]
  reraise Phoenix.ActionClauseError, args, stack
end
``` 

>  I'm really confused why me passing the wrong parameters is triggering the ActionClauseError. Shouldn't this be triggering the MissingParamError since its triggered by scrub_params? Or is that only if scrub_params is explicitly called?

In my understanding:
`ActionClauseError` means unprovided case, like 404 page
`MissingParamError` means validation error caused by [scrub_params/2](https://hexdocs.pm/phoenix/Phoenix.Controller.html#scrub_params/2), like 400 or 422 page

jbhatab:
@shhavel Thanks for the response. The explanation makes sense but for some reason the error message still feels a little off in terms of clarity. I think its much better than before though so not a big deal.


I fixed it with the better way of getting the params. Should be good to go now.

chrismccord:
<3 <3 <3

Issue #1920, opened 2016-09-18
Phoenix sends a 204 response after 404 error response


alex88:
### Environment

* Elixir version (elixir -v): 1.3.2
* Phoenix version (mix deps): 1.2.1
* Operating system: Osx

So I was developing an android app, I had 2 consecutive requests, one was returning 404 (controller not yet implemented) and the other one 200, at least this by the phoenix logs.

However the async callback was showing a 204 response. No logs was showing a 204 so I started sniffing traffic with wireshark. I've seen using "follow http stream" that phoenix was sending a 404 and a 204 for the first request and then a 200 for the second request.

I've tried to recreate a sample app and the issue can be reproduced (you can find the app at https://github.com/alex88/phoenix-204-issue-app).

An example of what happens is at https://gist.github.com/alex88/9709396b6d324a7e433c845559534ed4 if you make the 2 requests, you see after the 404 the 204 response https://gist.github.com/alex88/9709396b6d324a7e433c845559534ed4#file-404_http_stream

Update: I've just tried to run the app in production and it doesn't happen, so it's a dev only issue

josevalim:
Can you reproduce it on something else than the android device? With curl or any other HTTP client?

alex88:
If you look at the gist, it's reproduced locally with httpie

josevalim:
Oh, I missed that. Beautiful!


-- 


*José Valimwww.plataformatec.com.br
<http://www.plataformatec.com.br/>Founder and Director of R&D*

PR #1938, opened 2016-09-23
Fix extra 204 response when 404 error and debug_errors enabled

shhavel:
Potential fix for #1920 

extra_204_response_on_404

On screenshot displayed traffic for 2 get requests which render 404 page.
First request was made before fix and contains one extra 204 response,
second request was made after fix and contains only one response.


josevalim:
Excellent catch! I now understand why this happens, we will discuss solutions soon. :)

josevalim:

PR #1939, opened 2016-09-24
Fix experimental generators

wojtekmach:
* Use correct web template paths for live reload
* Use ecto 2.1-rc in umbrella domain app
* Fix umbrella templates


wojtekmach:
There's a bit overlap with #1927 so if preferred I can change base branch of this PR to be `cm-phx-gen-next`

chrismccord:
Thanks @wojtekmach for catching all my oversights :)
<3 <3 <3

PR #1941, opened 2016-09-25
Fix flaky installer tests

wojtekmach:
Looks like this change fixes the random build failures.

I'm still not exactly sure what's causing this, but it looks like calling the task directly (rather than going through Mix.TasksServer) fixes it - some race conditions, silent failure maybe? At the very least I was able to get a consistent repro test, see this commit message & content: https://github.com/wojtekmach/phoenix/commit/552b00a55c1924ae82f1fb0957d8ed1c4014dfde - perhaps someone more experienced might have better luck tracking it down :-)

### DIFF
--- a/installer/lib/phoenix_new.ex
+++ b/installer/lib/phoenix_new.ex
@@ -166,7 +166,7 @@ defmodule Mix.Tasks.Phoenix.New do
 
     case argv do
       [] ->
-        Mix.Task.run "help", ["phoenix.new"]
+        Mix.Tasks.Help.run ["phoenix.new"]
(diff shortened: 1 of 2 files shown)


wojtekmach:
ugh, the build failed but at least for a different reason - looks like a different flaky test (that also uses `capture_io`)

chrismccord:
Thanks! @josevalim can you review wrt to capture_io races and this fix?

josevalim:
This looks good to me.

chrismccord:
❤️❤️❤️

PR #1942, opened 2016-09-26
Fix brunch crash when file is removed and readded

ThijsWouters:
Brunch version 2.7.4 has a bug that when a file is removed and then readded, brunch crashes. This is fixed in version 2.7.7.

https://github.com/brunch/brunch/issues/1332


ThijsWouters:
Fixed. Closing.

PR #1956, opened 2016-10-02
Add Options For Transport Serializer

jeffutter:
This adds options to set and a default implementation for the websocket transport serializer. The backend supports this here: https://hexdocs.pm/phoenix/Phoenix.Transports.WebSocket.html . However the javascript had no way to set a matching serializer on the front-end.

Example Usage:

On your phoenix backend, with a socket like this:

```elixir
defmodule Ui.UserSocket do
  use Phoenix.Socket

  channel "thermostat", Ui.ThermostatChannel

  transport :websocket, Phoenix.Transports.WebSocket, serializer: Ui.WebSocketSerializerMsgPack
 
  def connect(_params, socket) do
    {:ok, socket}
  end

  def id(_socket), do: nil
end
```

And a serializer like this (using `Msgpax` to encode in msgpack rather than JSON):

```elixir
defmodule Ui.WebSocketSerializerMsgPack do
  @moduledoc false
@behaviour Phoenix.Transports.Serializer

  alias Phoenix.Socket.Reply
  alias Phoenix.Socket.Message
  alias Phoenix.Socket.Broadcast

  @doc """
  Translates a `Phoenix.Socket.Broadcast` into a `Phoenix.Socket.Message`.
  """
  def fastlane!(%Broadcast{} = msg) do
    message = %Message{
      topic: msg.topic,
      event: msg.event,
      payload: msg.payload
    }
    |> Map.from_struct
    |> Msgpax.pack!
    |> IO.iodata_to_binary

    {:socket_push, :binary, message}
  end

  @doc """
  Encodes a `Phoenix.Socket.Message` struct to JSON string.
  """
  def encode!(%Reply{} = reply) do
    message = %Message{
      topic: reply.topic,
      event: "phx_reply",
      ref: reply.ref,
      payload: %{status: reply.status, response: reply.payload}
    }
    |> Map.from_struct
    |> Msgpax.pack!
    |> IO.iodata_to_binary

    {:socket_push, :binary, message}
  end
  def encode!(%Message{} = msg) do
    message = msg
    |> Map.from_struct
    |> Msgpax.pack!
    |> IO.iodata_to_binary

    {:socket_push, :binary, message}
  end

  @doc """
  Decodes JSON String into `Phoenix.Socket.Message` struct.
  """
  def decode!(message, _opts) do
    message
    |> Msgpax.unpack!
    |> Phoenix.Socket.Message.from_map!()
  end
end
```

You can now use the following on the other end:

```js
import {Socket} from "./phoenix";
import msgpack from 'msgpack-lite';

const serializer = {
  encode: (data) => {
    return msgpack.encode(data)
  },
  decode: (data) => {
    let raw_binary_data = new Uint8Array(data);
    return msgpack.decode(raw_binary_data);
  }
};

let socket = new Socket("/socket", {params: {token: window.userToken}, serializer: serializer})
socket.connect()
```

To actually use this with Messagepack, you need to make one more change to phoenix.js. You need to set `this.conn.binaryType = 'arraybuffer';` around line 526. It would be nice to be able to set this without messing with phonenix.js I'll try to come up with another solution for that and open a separate PR, but being able to set the encoder/decoder is useful without that.

### DIFF
--- a/web/static/js/phoenix.js
+++ b/web/static/js/phoenix.js
@@ -447,6 +447,9 @@ export class Socket {
   // opts - Optional configuration
   //   transport - The Websocket Transport, for example WebSocket or Phoenix.LongPoll.
   //               Defaults to WebSocket with automatic LongPoll fallback.
+  //   serializer - The methods for encoding and decoding your messages for the socket.
+  //                This is usful when implementing an alternative serializer on your
+  //                backend transport.
   //   timeout - The default timeout in milliseconds to trigger push timeouts.
   //             Defaults `DEFAULT_TIMEOUT`
   //   heartbeatIntervalMs - The millisec interval to send a heartbeat message
@@ -474,6 +477,10 @@ export class Socket {
     this.ref                  = 0
     this.timeout              = opts.timeout || DEFAULT_TIMEOUT
     this.transport            = opts.transport || window.WebSocket || LongPoll
+    this.serializer           = opts.serializer || {
+      encode: (data) => JSON.stringify(data),
+      decode: (data) => JSON.parse(data)
+    }
     this.heartbeatIntervalMs  = opts.heartbeatIntervalMs || 30000
     this.reconnectAfterMs     = opts.reconnectAfterMs || function(tries){
       return [1000, 2000, 5000, 10000][tries - 1] || 10000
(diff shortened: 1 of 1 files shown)


jeffutter:
Oh, :facepalm: it looks like someone else contributed this already here #1840 . And much more thoroughly. I'll close this PR.

PR #1957, opened 2016-10-03
Fix spacing between generated Edit and Back links.

mattneel:
Just some simple span tags around it fixes it on the 3.x and 4.x CDN versions of Bootstrap I've tested.

### DIFF
--- a/priv/templates/phoenix.gen.html/edit.html.eex
+++ b/priv/templates/phoenix.gen.html/edit.html.eex
@@ -3,4 +3,4 @@
 <%= "<%=" %> render "form.html", changeset: @changeset,
                         action: <%= singular %>_path(@conn, :update, @<%= singular %>) <%= "%\>" %>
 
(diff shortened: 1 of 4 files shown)


josevalim:
@mattneel thank you! can you please do so for all relevants views in here? https://github.com/mattneel/phoenix/tree/fb29de14ce66bef128bae28244bed4915cf891c4/priv/templates/phoenix.gen.html

mattneel:
Done. Weird Travis-CI fail on OTP 19 though.

josevalim:
Thank you! Btw, quick question that I realized just now. Wouldn't it be semantically better to use `<p>` instead of `<span>`?

mattneel:
<p> Will cause a linebreak, which I don't think is the intention. Wrapping with span applies the Bootstrap padding correctly.

josevalim:

PR #1958, opened 2016-10-05
Accept media-range with empty subtype

LuzianW:
HTTP Accept headers with media range without subtype should be accepted if the Plug accepts a media-type matching the base type specified in the header.

eg. Accept text/* should be accepted for text/html, text/plain etc.

RFC: https://www.w3.org/Protocols/rfc2616/rfc2616-sec14.html

This change doesn't affect the order in which types are currently processed in the content negotiation, just adds the ability to match media types with an empty subtype.


josevalim:

PR #1963, opened 2016-10-12
yarn package manager

LostKobrakai:
Dynamically determine which node package manager will be used. It does use the faster yarn to install dependencies if installed.

### DIFF
--- a/installer/lib/mix/tasks/phx.new.ex
+++ b/installer/lib/mix/tasks/phx.new.ex
@@ -145,7 +145,8 @@ defmodule Mix.Tasks.Phx.New do
       webpack_pending = install_webpack(install?, project)
       Task.await(compile, :infinity)
 
-      if Project.webpack?(project) and !elem(node_package_manager(), 0) do
+      {pkm_available?, _} = node_package_manager()
+      if Project.webpack?(project) and !pkm_available? do
         print_webpack_info(project, generator)
       end
 
--- a/installer/lib/mix/tasks/phx.new.ex
+++ b/installer/lib/mix/tasks/phx.new.ex
@@ -145,7 +145,7 @@ defmodule Mix.Tasks.Phx.New do
       webpack_pending = install_webpack(install?, project)
       Task.await(compile, :infinity)
 
-      if Project.webpack?(project) and !System.find_executable("npm") do
+      if Project.webpack?(project) and !elem(node_package_manager(), 0) do
         print_webpack_info(project, generator)
       end
 
@@ -175,9 +175,12 @@ defmodule Mix.Tasks.Phx.New do
   defp install_webpack(install?, project) do
     assets_path = Path.join(project.web_path || project.project_path, "assets")
     webpack_config = Path.join(assets_path, "webpack.config.js")
+    {pkm_available?, pkm_install} = node_package_manager()
+    cd_into_assets_directory = "cd #{relative_app_path(assets_path)}"
+    initial_assets_compilation = "node node_modules/webpack/bin/webpack.js --mode development"
 
-    maybe_cmd "cd #{relative_app_path(assets_path)} && npm install && node node_modules/webpack/bin/webpack.js --mode development",
-              File.exists?(webpack_config), install? && System.find_executable("npm")
+    maybe_cmd Enum.join([cd_into_assets_directory, pkm_install, initial_assets_compilation], " && "),
+              File.exists?(webpack_config), install? && pkm_available?
   end
 
   defp install_mix(install?) do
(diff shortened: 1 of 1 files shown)


josevalim:
Thank you @LostKobrakai but I would like to wait a bit more time. Yarn is one day old and we should wait for it to get a bit more mileage before integrating it into Phoenix. If you prefer to use yarn, you can always skip the dependency installation step and then run yarn afterwards.

jeroenvisser101:
@josevalim Yarn is a lot faster though, and since it's now already 22 days old... Phoenix wouldn't force you anywhere, but if you have yarn installed locally, it would use yarn instead of npm, sounds alright to me?

nekath:
Time passed and we see a lot of traffic around yarn. (more than 460k downloads of the package). I think we should reconsider adding integration with yarn.

brandonmikeska:
Any news on yarn, especially if we only use it if we detect it, that way it doesn't affect anyone who isn't using it?

chrismccord:
No existing plans for support

ryanwinchester:
Guys. If you want to use yarn, just use yarn. ¯\_(ツ)_/¯

That's what I do (and every other sane person).

brandonmikeska:
Yeah I don't think anyone is making a big deal out of it :), just seeing if there were plans down the road to take that 5 key strokes down to one

josevalim:
@LostKobrakai would you be interested in rebasing this PR so we can finally merge it? You need to rebase because master uses webpack. :) Thank you.

LostKobrakai:
I can certainly do that in the next few days.

pragmaticivan:
Hey @LostKobrakai maybe it would require more checks for that feature.

npm nowadays also creates a `package-lock.json` for version consistence.

It would also require to check for `yarn.lock` to determine which one to use.

I think yarn still don't check `package-lock.json` content. https://github.com/yarnpkg/yarn/issues/3614

josevalim:
@pragmaticivan this is when just creating a new project, so there won't be any lock.

pragmaticivan:
@josevalim ahh my bad, makes sense, thanks!

installer/lib/mix/tasks/phx.new.ex
```diff
@@ -145,7 +145,7 @@ defmodule Mix.Tasks.Phx.New do
       webpack_pending = install_webpack(install?, project)
       Task.await(compile, :infinity)
 
-      if Project.webpack?(project) and !System.find_executable("npm") do
+      if Project.webpack?(project) and !elem(node_package_manager(), 0) do
```
Gazler: Could we avoid using `elem` here? The pattern matching approach used in `install_webpack` is much clearer.

Another option is to return the package manager command or `nil` from `node_package_manager()`

```
if Project.webpack?(project) and !node_package_manager() do

defp node_package_manager() do
  cond do
    System.find_executable("yarn") -> 
      "yarn"
    System.find_executable("npm") ->
      "npm install"
    true ->
      nil
  end
end
```

LostKobrakai: I really like the explicit return of "npm install" for the command logged into the console if npm/yarn is not available, but maybe it's enough to let `install_webpack` default to that command.

Gazler: @LostKobrakai that's a good point. In that case, I'd go for:

```
   {pkm_available?, _} = node_package_manager()
    if Project.webpack?(project) and !pkm_available? do
```

Fudoshiki:
I use yarn, but new npm@6 seems pretty fast

OvermindDL1:
Yeah I would absolutely opt for *not* installing another package manager since npm is 'already' one and is 'already' required by yarn anyway, while yarn adds nothing that I've seen (please correct me if it does otherwise!  I'm quite curious) except some speed (and maybe not in comparison to npm anymore anyway) but even then deps are downloaded generally only once per project anyway.

LostKobrakai:
Can we please not discuss the usefulness of yarn here? The installer does just a few lines of checking if yarn is installed and if so uses that instead of npm, otherwise it'll still fallback to npm. Also this pr is here since over a year now. So if the phoenix team decides to merge it it's certainly not without giving it the proper thought.

chrismccord:
Thanks @LostKobrakai! I have decided not to move forward on this feature. It would be an implicit lookup and another thing to go wrong outside of the team's direct control. npm has had improvements as well since yarn's release, and users are free as always to use yarn on their new projects themselves. Thanks!

Mdsp9070:
I understand it, however npm stills so dam slow! Any chance of going back on this decision? Yeah, I could start a new project and then remove npm-related files but thats not a really good user experience.

ryanwinchester:
> I understand it, however npm stills so dam slow! Any chance of going back on this decision? Yeah, I could start a new project and then remove npm-related files but thats not a really good user experience.

It's such a minimal amount of work to change it. (I even swap out all the webpack stuff because I don't even use webpack directly and that's not bad, either). If you are creating enough new phoenix projects that this significantly impacts your productivity, write a script that does what you want.

npm is generally installed alongside nodejs and is the default package manager. The team decided to default to the default, so please just let this die.

PR #1965, opened 2016-10-13
Phoenix.Param for maps

iamvery:
## Background

The documentation for `Phoenix.Param` appears to be out of sync with the implementation. It says:

> By default, [Phx] implements this protocol for [...] maps. For maps
> and structs, a key :id is looked up.

However the Map [implementation](https://github.com/phoenixframework/phoenix/blob/655f45bf9cf6cc0b1d85af3d1c9c43afbcacaf1a/lib/phoenix/param.ex#L75-L80) of the protocol raises an error.

## This Solution

The most obvious and easiest solution would be to update the documentation to match the implementation (it [was added](https://github.com/iamvery/phoenix/commit/9f668b05b57e9bc160b30397177a7e96e87b4fb3#diff-bfdd7e816187d8bfe18e7eb889727fcb) before [map support was removed](https://github.com/iamvery/phoenix/commit/870e779d593f19871d32f48e4f2440f25ced939d#diff-bfdd7e816187d8bfe18e7eb889727fcb)) , but rather than taking the easy road, I'd like to inquire more about why the protocol is not supported for maps.

The commit https://github.com/iamvery/phoenix/commit/870e779d593f19871d32f48e4f2440f25ced939d#diff-bfdd7e816187d8bfe18e7eb889727fcb seems to suggest that it was in order to support Plug 0.12. If that were the case, though, I would expect failing tests with map support added back. Is there some incompatibility that is not made obvious by the test suite?

Can you provide more context as to why maps should not be convertible to param? Thanks so much! ❤️

### DIFF
--- a/lib/phoenix/param.ex
+++ b/lib/phoenix/param.ex
@@ -10,9 +10,9 @@ defprotocol Phoenix.Param do
   Phoenix knows how to extract the `:id` from `@user` thanks
   to this protocol.
 
-  By default, Phoenix implements this protocol for integers,
-  binaries, atoms, maps and structs. For maps and structs, a
-  key `:id` is looked up.
+  By default, Phoenix implements this protocol for integers, binaries, atoms,
+  and structs. For structs, a key `:id` is assumed, but you may provide a
+  specific implementation.
 
   Nil values cannot be converted to param.
 


josevalim:
There are two reasons, both related to explicitness.

1. It is harder to make an assumption where the primary key is going to be in a map. With structs, for non-primary keys, you can always implement the Phoenix.Param protocol and it is guaranteed we will have a proper conversion. For maps, you need to depend on one particular key, so I would prefer if developers used `map.id` instead of simply `map`.

2. Plug knows how to encode maps: `%{key: "value"}` becomes `key=value`. So there is a chance a map is going to be encoded differently depending on the argument. So instead of avoiding the confusion in bullet 1 and also the confusion with plug encoding, I would prefer to to not support the to_param one.

So I think the way to go would be to update the docs. :)

iamvery:
Thank you so much for taking the time to write a great explanation! I'm happy to update the branch with the documentation changes :).

It's still a little curious to me why it's OK to assume a primary key in a struct and not a map. I do understand that there would be no way to specify an _alternate_ implementation for maps, but even so... somehow strikes me as a little off balance.

thanks again!

josevalim:
The difference is that you can change it for a struct by implementing the protocol. So the behaviour is always specific to each struct. With a map, it is the same global behaviour for all maps.

iamvery:
If you don't mind, I have one more proposal before I push up the doc changes (thank you for bearing with me)...

What if we were to remove support for `Any` struct? You mentioned explicitness, so it might be nice to at the very least require users to explicitly `@derive` the protocol. That way there is no base case for maps _or_ structs without an explicit decision. Maybe there would be some concern for struct-generating libs like ecto, but perhaps the explicitness is worth it?

This internal struggle for him is coming from some awkwardness writing the docs. Here's what I've got right now:
> By default, Phoenix implements this protocol for integers, binaries, atoms, and structs. For structs, a key `:id` is looked up. Maps are not supported as no assumption is made about the primary key stored in it.

To me that immediately raises the question "if you know how to look it up in a struct, why can't you just do 'what i mean' with maps too?" On the flip side if you're wanting to drive users to using `map.id` instead of implicitly finding `:id` in the map, why not suggest that same policy for structs without an implementation? The point might even be further driven with an error on the default implementation that tells them to do just that.

I hope this doesn't seem annoying and pedantic to you! I'm not meaning to be a bother 😞

josevalim:
No problem at all. To rehash what we said on IRC, while being explicit is one part of it, the main difference is that structs can be customized while maps do not. That's why we are ok with making assumptions for structs, because they can't  overridden, but they can't for maps.

josevalim:

Issue #1936, opened 2016-09-22
mix phoenix.digest should handle source maps


aaronjensen:
When building sourcemaps w/ webpack you get something like:

```
//# sourceMappingURL=bundle.js.map
```

at the bottom of your js file. `mix phoenix.digest` will digest and rename the `bundle.js` and the `bundle.js.map` but it does not update the `sourceMappingURL`.

Is this a responsibility of phoenix? Is this not an issue w/ the brunch build for some reason? Ideally it'd be handled by phoenix.

josevalim:
We should handle that. A PR is welcome! :D

-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

chrismccord:
To be clear, we have source maps in dev by virtue of brunch. I don't think digest needs anything in particular? If `brunch build --prod` strips them out by default, we could ship a config that includes them, but I think this is up to the user. I don't think the digest task itself should be concerned with this since it's asset builder specific? Or am I missing something?

aaronjensen:
@chrismccord the digest task breaks sourcemaps by virtue of not changing the `sourceMappingURL` if you want them in production.

It all depends on the user's workflow for production sourcemaps. Some upload them to their error tracker, others serve them behind an ip blocker, etc. I'd suggest having a default that doesn't break them if the user wants them would be good.

chrismccord:
ah derp, I see. Can this be as simple as updating the url via a regex just like we do with css? I am 👍 if someone want's to take this one!

ntilwalli:
I think it should also preserve the file name convention.  My `base.css.map` file gets digested to `base.css-e73a374a4c0650448f26e781ad616aa3.map`.  The hash string should be appended to the token before the first period, not the token before the last period imho.

Szetty:
hi :) I think I would take this issue, @chrismccord can u assign it to me?

chrismccord:
@Szetty It's all yours :) Thanks!

aaronjensen:
It looks like this was done and merged, thanks @Szetty!

PR #1968, opened 2016-10-14
Source mapping url digest

Szetty:
#1936 solved, also I think we will need to digest the contents from *.map files too, I can do that too, but I need a feedback if it's okay

### DIFF
--- a/lib/phoenix/digester.ex
+++ b/lib/phoenix/digester.ex
@@ -185,35 +185,43 @@ defmodule Phoenix.Digester do
   end
 
   defp digested_contents(file, manifest) do
-    if Path.extname(file.filename) == ".css" do
-      digest_asset_references(file, manifest)
-    else
-      file.content
+    case Path.extname(file.filename) do
+      ".css" -> digest_asset_references(:css, file, manifest)
+      ".js" -> digest_asset_references(:js, file, manifest)
+      _ -> file.content
     end
   end
 
   @stylesheet_url_regex ~r{(url\(\s*)(\S+?)(\s*\))}
   @quoted_text_regex ~r{\A(['"])(.+)\1\z}
 
-  defp digest_asset_references(file, manifest) do
+  defp digest_asset_references(:css, file, manifest) do
     Regex.replace(@stylesheet_url_regex, file.content, fn _, open, url, close ->
       case Regex.run(@quoted_text_regex, url) do
         [_, quote_symbol, url] ->
-          open <> quote_symbol <> digested_url(url, file, manifest) <> quote_symbol <> close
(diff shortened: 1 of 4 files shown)


lib/phoenix/digester.ex
```diff
@@ -185,35 +185,43 @@ defmodule Phoenix.Digester do
   end
 
   defp digested_contents(file, manifest) do
-    if Path.extname(file.filename) == ".css" do
-      digest_asset_references(file, manifest)
-    else
-      file.content
+    case Path.extname(file.filename) do
```
josevalim: I think we have two options. Or we pattern match here and then call `css_asset_references` and `js_asset_references` or we call `digest_asset_references(Path.extname(file.filename), ...)` and pattern match inside each function clause. Today we are doing a mixture which is a bit unclear.

lib/phoenix/digester.ex
```diff
@@ -185,35 +185,43 @@ defmodule Phoenix.Digester do
...
+          open <> quote_symbol <> digested_url(url, file, manifest, true) <> quote_symbol <> close
         nil ->
-          open <> digested_url(url, file, manifest) <> close
+          open <> digested_url(url, file, manifest, true) <> close
       end
     end)
   end
 
-  defp digested_url("/" <> relative_path, _file, manifest) do
+  @javascript_sourceMap_regex ~r{(//#\s*sourceMappingURL=)(\S+)}
```
josevalim: `@javascript_source_map_regex`?

lib/phoenix/digester.ex
```diff
@@ -185,35 +185,43 @@ defmodule Phoenix.Digester do
...
+
+  defp digest_asset_references(:js, file, manifest) do
+    Regex.replace(@javascript_sourceMap_regex, file.content, fn _, sourceMapText, url ->
```
josevalim: The var should be `source_map_text`.

josevalim:
@Szetty In my opinion it is also ok to digest the contents of the `.map` files. Can you think of any reason not to? Shouldn't we digest so we are able to point back to the proper JS file?

Szetty:
thx for commenting @josevalim, there is absolutely no reason not to digest the contents of the .map files, just that I didn't know if we should open a new issue for this, to be able to track separate tasks, or I can implement it in this pull request, because they have some connection.

Talking about this from CONTRIBUTING.md : "Good pull requests - patches, improvements, new features - are a fantastic help. They should remain focused in scope and avoid containing unrelated commits."

josevalim:
Yes, let's do two PRs. That would be better.

josevalim:

PR #1973, opened 2016-10-17
Do not import phonix_html when generating with --no-html

philipgiuliani:
Currently when generating a phoenix app with `--no-html`, it will still try to import the phoenix_html javascript dependency.

```
Resolving deps of web/static/js/app.js failed. Could not load module 'phoenix_html' from '.../web/static/js'. Possible solution: add 'phoenix_html' to package.json and `npm install`.
```

### DIFF
--- a/installer/templates/assets/brunch/app.js
+++ b/installer/templates/assets/brunch/app.js
@@ -11,7 +11,7 @@
 //
 // If you no longer want to use a dependency, remember
 // to also remove its path from "config.paths.watched".
-import "phoenix_html"
+<%= if html do %>import "phoenix_html"<% end %>
 
 // Import local files
 //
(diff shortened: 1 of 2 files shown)


josevalim:
Can you also change: https://github.com/phoenixframework/phoenix/blob/master/installer/templates/static/brunch/app.js#L14

philipgiuliani:
Done! 👍

josevalim:
Haha, oops, sorry. :)

josevalim:

installer/templates/assets/brunch/app.js
```diff
@@ -11,7 +11,7 @@
 //
 // If you no longer want to use a dependency, remember
 // to also remove its path from "config.paths.watched".
```
philipgiuliani: Should i move the beginning of the `if` to this line, so it does not create 2 newlines?

PR #1977, opened 2016-10-22
Fix unclosed backquotes in Token docs

wsmoak:
Fix warning when generating docs, backquotes were not closed correctly.


chrismccord:
❤️❤️❤️

PR #1986, opened 2016-10-26
replace `milli_seconds` with `milliseconds`

sasa1977:
Elixir spec for `System.time_unit` doesn't include underscore versions (though they still work at runtime). Therefore, in a Phoenix dependent project dialyzer will generate errors (e.g. when invoking `Phoenix.Token.sign`).

Looking at Elixir docs, it seems that the decision to not support underscore style of units is deliberate. Therefore, it seems reasonable to use Elixir style of time units in Phoenix.

### DIFF
--- a/lib/phoenix/token.ex
+++ b/lib/phoenix/token.ex
@@ -161,5 +161,5 @@ defmodule Phoenix.Token do
     KeyGenerator.generate(secret_key_base, salt, key_opts)
   end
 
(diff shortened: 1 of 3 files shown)


josevalim:
The underscore version is wrong, so we normalized it to `:milliseconds` but OTP 19.1 normalized it to `:millisecond`. This means yet another change is coming in the future. Although this is great as is, thank you!

josevalim:

PR #1989, opened 2016-10-31
Adding foreign key to the generated belongs to

### LINKED ISSUE #1605: Invalid migration generated by gen.model and run-time errors after omitting `_id` postfix in association column name
### Environment
- Phoenix version: 1.1.4, _master_
- Ecto version: 1.1.4
### Steps to reproduce

I've created two models with those commands:
- `mix phoenix.gen.html User users foo:string` - _not important_, just for clarity.
- `mix phoenix.gen.html Room rooms author:references:users`

Second command is crucial, because **I forgot to add** `_id` postfix - it should look more like this one:
- `mix phoenix.gen.html Room rooms author_id:references:users`
### Actual behavior

So the fragment of generated migration looks like the following:

``` elixir
  # ...
    create table(:rooms) do
      # It should be `author_id`...
      add :author, references(:users, on_delete: :nothing)

  # ... and similar issue for column name for index creation :(
```

This would not be a problem - migrations will be applied successfully, app will compile without any errors - however it will crash during the run-time execution.

It happens because `belongs_to` will automatically add `_id` postfix to the column name and it will look for a non-existing column e.g. doing `Repo.all(Room)`:

**It is not a big deal** - it can be fixed manually, by changing column name in migration. However _IMHO_ it can lead to confusion and if you're using generators you'd expect to have working state immediately after using a command.
### Expected behavior

Created migration could look like this:

``` elixir
defmodule InteractiveTetris.Repo.Migrations.CreateRoom do
  use Ecto.Migration

  def change do
    create table(:rooms) do
      add :author_id, references(:users, on_delete: :nothing)

      timestamps
    end
    create index(:rooms, [:author_id])

  end
end
```

Or at least it should trigger _a warning_ that we're expecting certain naming convention for column names in references - it looks like it trims postfix automatically [here](https://github.com/phoenixframework/phoenix/blob/master/lib/mix/tasks/phoenix.gen.model.ex#L170).

This can be also reformulated to a different question:

> Is _gen.model_'s association syntax compliant with both formats - the reference name (`author`) and column name (`author_id`) - or just with column name in certain format (`author_id`)?

If you'll find this _"bug"_ worth fixing - I'd like to provide a fix for that. :neckbeard:
If not - feel free to close it. :wink:

rustedgrail:
FIX #1605

Adding a foreign_key to the belongs to so that the key can be changed without having to update the migrations

With help from @trevoke

### DIFF
--- a/priv/templates/phoenix.gen.model/migration.exs
+++ b/priv/templates/phoenix.gen.model/migration.exs
@@ -5,7 +5,7 @@ defmodule <%= base %>.Repo.Migrations.Create<%= scoped %> do
     create table(:<%= plural %><%= if binary_id do %>, primary_key: false<% end %>) do
 <%= if binary_id do %>      add :id, :binary_id, primary_key: true
 <% end %><%= for {k, v} <- attrs do %>      add <%= inspect k %>, <%= inspect v %><%= migration_defaults[k] %>
-<% end %><%= for {_, i, _, s} <- assocs do %>      add <%= inspect i %>, references(<%= inspect(s) %>, on_delete: :nothing<%= if binary_id do %>, type: :binary_id<% end %>)
+<% end %><%= for {_, i, _, s} <- assocs do %>      add <%= if(String.ends_with?(inspect(i), "_id"), do: inspect(i), else: inspect(i) <> "_id") %>, references(<%= inspect(s) %>, on_delete: :nothing<%= if binary_id do %>, type: :binary_id<% end %>)
 <% end %>
       timestamps()
     end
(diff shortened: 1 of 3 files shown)


kelvinst:
The code LGTM. But I would prefer if the `foreign_key` was set just in the cases it is needed, more specifically just when the reference field does not end with `_id`.

josevalim:
I agree we should only emit the foreign key when necessary and we should likely add a test case based on the example described in #1605.

Thanks for the PR so far!

Trevoke:
Made a change so that it adds `_id` only if the provided field does not already end with `_id`. Note that this can lead to things like `userid_id`.

I had to modify code in both eex files, which means there might be a better place for this code, should future changes occur.

chrismccord:
❤️❤️❤️🐥🔥

PR #1990, opened 2016-11-01
Update Presence update/3 and update/4 @spec

pdawczak:
In this PR the change to `update/3` and `update/4` is introduced.

It is a mirror of change in the other Pull Request of PubSub's Tracker, which [accepts a function as a `meta` updater](https://github.com/phoenixframework/phoenix_pubsub/pull/52).

This is little bit delayed change, but I hope it helps!

Best regards,
Paweł


chrismccord:
<3 <3 <3

PR #1991, opened 2016-11-03
fix assigns typespec

sasa1977:


chrismccord:
<3 <3 <3

sasa1977:
With special thanks to @fishcakez who discovered the root cause of the problem :-)

PR #1993, opened 2016-11-05
Add configurable logging to channels

### LINKED ISSUE #1984: Channel logging is insufficient to understand source of errors
### Environment
- Elixir version (elixir -v):

```
Erlang/OTP 18 [erts-7.3] [source] [64-bit] [smp:4:4] [async-threads:10] [hipe] [kernel-poll:false] [dtrace]

Elixir 1.3.2
```
- Phoenix version (mix deps):

```
➜  delorean git:(myron/increase-timeout) cat mix.lock | grep phoenix
  "phoenix": {:hex, :phoenix, "1.2.1", "6dc592249ab73c67575769765b66ad164ad25d83defa3492dc6ae269bd2a68ab", [:mix], [{:cowboy, "~> 1.0", [hex: :cowboy, optional: true]}, {:phoenix_pubsub, "~> 1.0", [hex: :phoenix_pubsub, optional: false]}, {:plug, "~> 1.1", [hex: :plug, optional: false]}, {:poison, "~> 1.5 or ~> 2.0", [hex: :poison, optional: false]}]},
  "phoenix_html": {:hex, :phoenix_html, "2.7.0", "19e12e2044340c2e43df206a06d059677c59ea1868bd1c35165438d592cd420b", [:mix], [{:plug, "~> 1.0", [hex: :plug, optional: false]}]},
  "phoenix_live_reload": {:hex, :phoenix_live_reload, "1.0.5", "829218c4152ba1e9848e2bf8e161fcde6b4ec679a516259442561d21fde68d0b", [:mix], [{:fs, "~> 0.9.1", [hex: :fs, optional: false]}, {:phoenix, "~> 1.0 or ~> 1.2-rc", [hex: :phoenix, optional: false]}]},
  "phoenix_pubsub": {:hex, :phoenix_pubsub, "1.0.0", "c31af4be22afeeebfaf246592778c8c840e5a1ddc7ca87610c41ccfb160c2c57", [:mix], []},
```
- NodeJS version (node -v):

Not using node.  The front end is an entirely different code base so we have no assets.
- NPM version (npm -v):

See above.
- Operating system:

OS X locally, Ubuntu in prod.
### Expected behavior

The logging I get from Phoenix/Plug for normal HTTP requests is quite good.

```
2016-10-25T06:26:16.540 request_id=5i4mgbo84cui7sm06fea456g957lim12 module=Plug.Logger [info] GET /delorean-api/on-page/prod.4089487.831342/summary/google.en-US/week
# ...
2016-10-25T06:26:16.597 request_id=5i4mgbo84cui7sm06fea456g957lim12 campaign=prod.4089487.831342 module=Plug.Logger [info] Sent 200 in 56m
```

Each request is tagged with an id, and the input is logged (the `GET /path` bit), and then later the response and response time is logged.  I can clearly see what each request and response was.

For channels, I expect like a similar level of logging:
- Everything associated with specific channel should have a unique channel id (similar to the `request_id` for HTTP) that allows you to easily find all the log messages related to that channel.
- Every time a client sends an event, the input should be logged (similar to the `GET /path` bit above) and the response logged.  Having all of this would help me be able to see how a channel got in a specific state by seeing all the events the client sent to it.
### Actual behavior

The logging I see for phoenix channel interactions is much more limited than what I expected.  Looking through my production logs, I see this:

```
2016-10-25T06:50:10.659 module=Phoenix.Logger [info] JOIN keyword_analysis to DeloreanAPI.KeywordAnalysisChannel
  Transport:  Phoenix.Transports.WebSocket
  Parameters: %{}
2016-10-25T06:50:10.660 module=Phoenix.Socket.Transport [info] Replied keyword_analysis :ok
```

...which is the input and output for the join request, but that's about it.  I'm not seeing any similar messages for the events sent by clients, and no `channel_id` attached to each log message. (I realize I'd have to configure my logger to include the `channel_id` for it to appear in my logs, but Phoenix isn't even setting it via `Logger.metadata`).

chrismccord:
/cc @josevalim 

We can also ditch the clauses in the Phoenix.Logger that match on the "phoenix" topic since they are there to silence the live reloader channel, which can now simply do `log: false` internally.

### DIFF
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -247,6 +247,17 @@ defmodule Phoenix.Channel do
         push socket, ev, payload
         {:noreply, socket}
       end
+
+  ## Logging
+
+  By default, channel `"join"` and `"handle_in"` events are logged, using
+  the level `:info` and `:debug`, respectively. Logs can be customized per
+  event type or disabled by setting the `:log_join` and `:log_handle_in`
+  options when using `Phoenix.Channel`. For example, the following
+  configuration logs join events as `:info`, but disables logging for
+  incoming events:
+
+      use Phoenix.Channel, log_join: :info, log_handle_in: false
   """
   alias Phoenix.Socket
   alias Phoenix.Channel.Server
@@ -279,15 +290,21 @@ defmodule Phoenix.Channel do
               {:shutdown, :left | :closed} |
               term
 
-  defmacro __using__(_) do
+  defmacro __using__(opts \\ []) do
     quote do
+      opts = unquote(opts)
       @behaviour unquote(__MODULE__)
       @on_definition unquote(__MODULE__)
       @before_compile unquote(__MODULE__)
       @phoenix_intercepts []
+      @phoenix_log_join Keyword.get(opts, :log_join, :info)
+      @phoenix_log_handle_in Keyword.get(opts, :log_handle_in, :debug)
 
       import unquote(__MODULE__)
       import Phoenix.Socket, only: [assign: 3]
+
+      def __channel__(:log_join), do: @phoenix_log_join
+      def __channel__(:log_handle_in), do: @phoenix_log_handle_in
 
       def code_change(_old, socket, _extra), do: {:ok, socket}
 
--- a/lib/phoenix/socket/transport.ex
+++ b/lib/phoenix/socket/transport.ex
@@ -227,9 +227,17 @@ defmodule Phoenix.Socket.Transport do
     |> do_dispatch(msg, socket)
   end
 
-  defp do_dispatch(nil, %{event: "phx_join", topic: topic} = msg, socket) do
-    if channel = socket.handler.__channel__(topic, socket.transport_name) do
-      socket = %Socket{socket | topic: topic, channel: channel}
+  defp build_channel_socket(%Socket{} = socket, channel, topic) do
+    %Socket{socket |
+            topic: topic,
+            channel: channel,
+            log_join: channel.__channel__(:log_join),
+            log_handle_in: channel.__channel__(:log_handle_in)}
+  end
+
+  defp do_dispatch(nil, %{event: "phx_join", topic: topic} = msg, base_socket) do
+    if channel = base_socket.handler.__channel__(topic, base_socket.transport_name) do
+      socket = build_channel_socket(base_socket, channel, topic)
 
       case Phoenix.Channel.Server.join(socket, msg.payload) do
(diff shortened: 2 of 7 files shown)


lib/phoenix/channel.ex
```diff
@@ -279,16 +293,19 @@ defmodule Phoenix.Channel do
...
               term
 
-  defmacro __using__(_) do
+  defmacro __using__(opts \\ []) do
     quote do
       @behaviour unquote(__MODULE__)
       @on_definition unquote(__MODULE__)
       @before_compile unquote(__MODULE__)
       @phoenix_intercepts []
+      @log unquote(opts[:log] || [join: :info, handle_in: :info])
```
josevalim: `bind_quoted: opts` and the value should be accessed at runtime. We should also do `Keyword.get(opts, :log, [...])` otherwise `log: false` would never work?

josevalim: Actually, I am not a big fan of this API. What does an empty list mean? If I configure only one of them, like `[join: :info]`, does it mean I don't  want to log the other?

Also controllers pass the log_level explicitly as the instrumentation metadata. Is there any chance we could do something similar here?

chrismccord:
The goal is to be able to configure each kind of event separately. In prod I may want to log all joins as today, but all channel incoming events would be too noisy, where in dev, logging handle_in's would be helpful. What kind of API are you thinking? Separate confit keys per event?

josevalim:
@chrismccord let's talk over skype whenever you have the chance.

lib/phoenix/socket/transport.ex
```diff
@@ -227,9 +227,17 @@ defmodule Phoenix.Socket.Transport do
...
   end
 
-  defp do_dispatch(nil, %{event: "phx_join", topic: topic} = msg, socket) do
-    if channel = socket.handler.__channel__(topic, socket.transport_name) do
-      socket = %Socket{socket | topic: topic, channel: channel}
+  defp build_channel_socket(%Socket{} = socket, channel, topic) do
+    %Socket{socket |
+            topic: topic,
+            channel: channel,
+            log_join: channel.__channel__(:log_join),
```
josevalim: What if we introduce a private field to the `socket` and store this there? It is going to be counter productive I think to introduce new fields to the socket every time we introduce a new log operation. If we don't want to introduce a generic private field, I would recommend at least grouping the log operations:

    logs: channel.__channel__(:log)

logs will be a map and we should only log if there is a matching key. For example, `%{join: :info}` means only log join with level `:info`.

PR #1996, opened 2016-11-06
Tokens with optional json payload support

### LINKED ISSUE #1995: Phoenix.Token.sign use json payload
### Phoenix.Token.sign Docs says 

> The data can be read by clients, but the message is signed to prevent tampering.

### Expected behavior

Its expectable, that the signed data can be _easily_ read by clients. Nowadays we are speaking mostly about js clients, or other services which can parse json.. (they all do)

### Actual behavior

The current behavior is, that the token payload is actually a BERT => Binary Erlang Term.
Yes, there are js libraries, which could be used to parse it, but.. does it makes sense? I mean Poison is already part of phoenix anyway. We could use `Poison.encode` instead of `:erlang.term_to_binary` [here](https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/token.ex#L99) 

----

Additionally, we could put a simple Phoenix Token helper into phoenix.js itself, to let the developer decode the token, if needed.

Do you agree, that its better to switch to json string, instead of BERT, to make the tokens readable everywhere?

dropfen:
Currently the payload is encoded as a binary_erlang_term.. In other words, its not directly parsable on the client, like a jwt.

This patch makes it possible to provide an option `json: true` to Phoenix.Token.sign & Phoenix.Token.verify

Note: If the you have created a token with json payload, it must be verified with json: true option, otherwise an Exception will raise, because the :erlang.binary_to_term will not be able to parse, obviously 

But this patch also provides the documentation for the new option and the note above.
Resolves #1995


josevalim:
Thank you but I am :-1:. If you want to use JSON or JWT, it would be preferrable to use a library that was designed for so instead of slowly adding features to Phoenix that makes it work like a JSON library but not quite.

dropfen:
This statement in the docs, made me confused
> The data can be read by clients, but the message is signed to prevent tampering.

Phoenix.Token is *readable on the client, but not quite*

I disagree, that that kind of patches makes phoenix behave like a JSON library. Phoenix has json support in other places as well, its a web framework, and json is the de-facto the default data format on the web.

Yes, I agree that if you need JWT, you should use a JWT library. But I am also seeing a great potential in Phoenix.Token as well. Its simple, and already part of the system, so why forcing people going other ways if they could built on top it.

Well, anyway thank you for the review.

josevalim:
> The data is readable on the client, but not quite

I believe the point of this line is to say to not store confidential information. Any client can read it as long as it implemented the Erlang deserialization format. :) Improvements to the docs are always welcome.

dropfen:
> Any client can read it as long as it implemented the Erlang deserialization format. :)
I knew you have a great sense of humor :)

josevalim:
;)

Just to be double clear though: the point of that line is not to say clients can consume it programatically (i.e. I am not asking you to implement Erlang deserialization in JS) but to say the information in there is not encrypted. I will improve the docs to be make it clearer.

dropfen:
Yeah, the line is clear, I was just thinking the PR would bring some value to the Token Module. 
It adds only *10LOC* but offers some features like system -> system communication, usage of microservices. It's not like I don't like BERT. Its just data for me nothing more, and its fine for communication between processes, because you can store tuples in it. But json is de-facto a standard, and here it's optional as well. So without `json: true` its still the erlang term without breaks.

webdeb:
@josevalim I would like to make a another PR which addresses the initial issue, but instead of providing `json: true` I think it's better to provide the an `encoder: MyEncoder` option, which is more flexible and general. `MyEncoder` would have to implement `term_to_binary` / `binary_to_term` functions, so by default the `:erlang` module could be used. Does it make sense that way?

PR #1999, opened 2016-11-11
Add socket.js tests

rossta:
This PR adds a few mocha tests for the `Socket` object in `phoenix.js`.

I'm opening this PR as a potential starting point for building out a suite of unit tests for the javascript functionality provided by `phoenix.js`. Though only just a few tests, there are already a few key decision points where I'm looking for feedback before I get too far.

Though `phoenix.js` touches a few browser APIs, I decided to stick with the Node.js context for the tests for simplicity. This means introducing a few libraries, e.g., `jsdom`, `jsdom-global`, `mock-socket`, to mock out those dependencies. We could certainly run the tests in a browser-based runner, e.g., karma, testem, etc., instead (or do both!), though the approach I went with was consistent with the existing presence tests. Looking for comments.

I've only added a few tests to get a proof-of-concept going for the mocking approach. If we like this direction, let me know your thoughts on how to tackle remaining tests: merge small sets of tests as I go vs build out an entire suite in one PR. Certainly, if you have any comments on the tests I've written so far, I'd be happy to address.

Another decision would be when to add the node tests to the travis build.

Finally, there's never enough love to go around, so thanks for all your hard work on Phoenix! 
💜💜💜

### DIFF
--- a/web/static/js/phoenix.js
+++ b/web/static/js/phoenix.js
@@ -724,7 +724,7 @@ export class Ajax {
       this.xdomainRequest(req, method, endPoint, body, timeout, ontimeout, callback)
     } else {
       let req = window.XMLHttpRequest ?
-                  new XMLHttpRequest() : // IE7+, Firefox, Chrome, Opera, Safari
+                  new window.XMLHttpRequest() : // IE7+, Firefox, Chrome, Opera, Safari
                   new ActiveXObject("Microsoft.XMLHTTP") // IE6, IE5
       this.xhrRequest(req, method, endPoint, accept, body, timeout, ontimeout, callback)
     }
--- a/web/test/socket_test.js
+++ b/web/test/socket_test.js
@@ -251,19 +251,35 @@ describe("connectionState", () => {
     assert.equal(socket.connectionState(), "closed")
   })
 
-  it("returns the valid connection readyState", () => {
+  it("returns connecting", () => {
     socket.connect()
 
     socket.conn.readyState = 0
     assert.equal(socket.connectionState(), "connecting")
+    assert.ok(!socket.isConnected(), "is not connected")
+  })
+
+  it("returns open", () => {
+    socket.connect()
 
     socket.conn.readyState = 1
     assert.equal(socket.connectionState(), "open")
+    assert.ok(socket.isConnected(), "is connected")
+ })
+
+  it("returns closing", () => {
+    socket.connect()
 
     socket.conn.readyState = 2
     assert.equal(socket.connectionState(), "closing")
+    assert.ok(!socket.isConnected(), "is not connected")
+  })
+
+  it("returns closed", () => {
+    socket.connect()
 
     socket.conn.readyState = 3
     assert.equal(socket.connectionState(), "closed")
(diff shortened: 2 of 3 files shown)


chrismccord:
This looks great! And we surely could use unit tests for proper client coverage. 

> I've only added a few tests to get a proof-of-concept going for the mocking approach. If we like this direction, let me know your thoughts on how to tackle remaining tests: merge small sets of tests as I go vs build out an entire suite in one PR.

Please continue with the tests, using this PR, and we can merge when we're in a more complete state. Thanks so much!

rossta:
@chrismccord I've now completed a first pass at unit tests for the `Socket` class. I'm pretty sure this gives us coverage for all the Socket functions, either explicitly or implicitly. There may be a few pathways missing (like testing against a mock `ActiveXObject` object), but this may be good enough as a start. 

I opted for duplication in areas like setup and teardown of tests in the interest of local readability and perhaps to help highlight dependencies and coupling for future refactoring. 

I'm interested in tackling tests other exported classes in `phoenix.js` perhaps in separate PRs.

Looking forward to your feedback!

web/static/js/phoenix.js
```diff
@@ -724,7 +724,7 @@ export class Ajax {
       this.xdomainRequest(req, method, endPoint, body, timeout, ontimeout, callback)
     } else {
       let req = window.XMLHttpRequest ?
-                  new XMLHttpRequest() : // IE7+, Firefox, Chrome, Opera, Safari
+                  new window.XMLHttpRequest() : // IE7+, Firefox, Chrome, Opera, Safari
```
rossta: Because the `jsdom-global` package gives us `window` as a property of `global`, `window` does not represent the global context. That's why I needed to make this change in `phoenix.js` to make sure we can locate the fake `XMLHttpRequest` in our node tests.

web/test/socket_test.js
```diff
@@ -0,0 +1,739 @@
...
+    })
+  })
+})
+
+describe("protocol", () => {
+  beforeEach(() => {
+    socket = new Socket("/socket")
+  })
+
+  it("returns wss when location.protocal is https", () => {
```
roperzh: Really minor typo: "protocal" should be "protocol"

web/test/socket_test.js
```diff
@@ -0,0 +1,739 @@
...
+    jsdom.changeURL(window, "https://example.com/");
+
+    assert.equal(socket.protocol(), "wss")
+  })
+
+  it("returns ws when location.protocal is http", () => {
```
roperzh: Really minor typo: "protocal" should be "protocol"

chrismccord:
😍😍😍🐥🔥

You are my hero! Thank you so much for paying down my lack of js test debt. This will make future iterations must nicer. Looks great and I'm sure now we've improved the bus factor on the js client :) Welcome to the club!

rossta:
w00t! Thanks!

I'd be interested in working on some additional tests for `Channel` next if you think that would be helpful.

chrismccord:
Yes, please do! That would be extremely helpful. We particularly could use tests on the Push join/receive/timeout chains.

PR #2000, opened 2016-11-14
Refactor log options into private field

chrismccord:
@josevalim thoughts on __socket__/1? Prefer something else?


josevalim:

PR #2003, opened 2016-11-18
Fix IE Bug

jfrolich:
If you set xhr.timeout in IE11 before calling open it raises an InvalidStateError.
Relevant documentation: https://msdn.microsoft.com/en-us/library/cc304105(v=vs.85).aspx
This is the reason LongPoll does not work in IE11.


chrismccord:
❤️❤️❤️🐥🔥

PR #2005, opened 2016-11-19
Handle empty json in json_response

migore:
I was trying to send an empty response according to [this](http://www.phoenixframework.org/docs/controllers#section-sending-responses-directly). But in my tests, I was getting `** (CaseClauseError) no case clause matching: {:error, :invalid}`

According to poison `{:error, :invalid}` is a possible response in decode (https://github.com/devinus/poison/blob/master/lib/poison.ex#L66). However, it's just not being handled in `json_response`. I added a clause to handle this scenario.

I ended up using [response/2](https://hexdocs.pm/phoenix/Phoenix.ConnTest.html#response/2) to make the assertion in my test. Should the error message suggest to use this other function?


chrismccord:
<3 <3 <3

Issue #1892, opened 2016-08-31
Improve Phoenix.Token verify examples


chrismccord:

byelipk:
Happy to work on this documentation. This is what I've come up with so far:

```Elixir
defmodule Phoenix.Token do
  
  ...

  @doc """
  Given a secret key base, a salt, a token, and optional arguments `verify/4`
  will establish the authenticity and integrity of the token and then decode
  and return the original data.

  ## Example

  In this scenario we will create a token and provide it to a client application.
  The data we will encode is a `user_id`, our `secret key base` will be in
  the form of a string, and we'll use a cryptographic `salt` of "random" data.

      user_id = 99
      secret  = "kjoy3o1zeidquwy1398juxzldjlksahdk3"
      salt    = "random"

      token   = Token.sign(secret, salt, user_id)

  When the client requests a resource, it sends the server a token to validate
  the request. The server can now verify if the token sent by the client is valid
  by using `verify/4`:

      Token.verify(secret, salt, token)
      => {:ok, 99}

  This time the client sent a valid token because `verify/4` returned a
  tuple of type `{:ok, user_id}`.

  If the client had sent an expired or otherwise invalid token `verify/4` would have
  returned an error instead:

      Token.verify(secret, salt, expired, max_age: 1209600)
      => {:error, :expired}

      Token.verify(secret, salt, invalid)
      => {:error, :invalid}

  ## Options

    * `:max_age` - verifies the token only if it has been generated
      "max age" ago in seconds. A reasonable value is 2 weeks (`1209600`
      seconds);
    * `:key_iterations` - option passed to `Plug.Crypto.KeyGenerator`
      when generating the encryption and signing keys. Defaults to 1000;
    * `:key_length` - option passed to `Plug.Crypto.KeyGenerator`
      when generating the encryption and signing keys. Defaults to 32;
    * `:key_digest` - option passed to `Plug.Crypto.KeyGenerator`
      when generating the encryption and signing keys. Defaults to `:sha256';

  """
  def verify(context, salt, token, opts \\ [])
  def verify(context, salt, token, opts) ...
  def verify(_context, salt, nil, _opts)   ...
  
  ...

end
```

ksolo:
@chrismccord @byelipk 
Is this issue now closed?

byelipk:
@ksolo Not to my knowledge. I'm new to Elixir/Phoenix and haven't received any feedback on this. If there are fixes to make, let me know.

[...]

PR #2013, opened 2016-11-29
Tweaked `Phoenix.Token` documentation.

bill-mybiz:
* I found it very confusing to have "user" as the salt argument.
  * Changed the salt from "user" to "tokensalt123".
  * Added documentation briefly describing the salt with a link
    to wikipedia entry.
  * Added documentation briefly describing the fact that the third
    argument can be any Elixir term which was very useful for me
    but not obvious since the functions do not have specs.

### DIFF
--- a/lib/phoenix/token.ex
+++ b/lib/phoenix/token.ex
@@ -15,8 +15,8 @@ defmodule Phoenix.Token do
   the id from a database. For example:
 
       iex> user_id = 1
-      iex> token = Phoenix.Token.sign(MyApp.Endpoint, "tokensalt123", user_id)
-      iex> Phoenix.Token.verify(MyApp.Endpoint, "tokensalt123", token)
+      iex> token = Phoenix.Token.sign(MyApp.Endpoint, "user salt", user_id)
+      iex> Phoenix.Token.verify(MyApp.Endpoint, "user salt", token)
       {:ok, 1}
 
   In that example we have a user's id, we generate a token and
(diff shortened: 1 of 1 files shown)


lib/phoenix/token.ex
```diff
@@ -15,7 +15,7 @@ defmodule Phoenix.Token do
   the id from a database. For example:
 
       iex> user_id = 1
-      iex> token = Phoenix.Token.sign(MyApp.Endpoint, "user", user_id)
+      iex> token = Phoenix.Token.sign(MyApp.Endpoint, "tokensalt123", user_id)
       iex> Phoenix.Token.verify(MyApp.Endpoint, "user", token)
```
kelvinst: Agree with you for the salt changing, but you just need to fix this line too. 😉

bill-mybiz: Ah, yep. :eyes:

bill-mybiz:
I was unaware of #1892. I just have been changing ibGib to be more dynamic and less server-side rendered, so I'm switching to channels. In learning about how to use `Token`, I was slightly :confused: when reading the documentation.

bill-mybiz:
I have now changed the other references to the "user" argument (there were a couple of spots doh!)

josevalim:
I think calling the salt "user" is completely fine. It is better to give a semantically relevant value to the salt, such as token, as a value that you have remember. The salt does not have to be a randomized token/value.

bill-mybiz:
I'm not particularly against having a semantically relevant/memorable value to the salt, but in my (anecdotal) experience, it was distracting enough for me to want to do a PR to change it for the example's purposes. 

I think this stems from it not being self-documenting code: If you _already_ know what it is, then yes, it's fine. But if you have no idea what it is (as the POV of a learner or new reader of your code), then you have to guess at it. Perhaps "usersalt" or "user salt" would be a better one (I almost used this in the first place). 

I'm just giving you my take on having recently gone through the learning process. I also wanted to improve the docs since the [google results for "elixir phoenix channel authentication"](https://www.google.com/search?client=ubuntu&channel=fs&q=elixir+phoenix+channel+authentication&ie=utf-8&oe=utf-8) gives [this article](https://quickleft.com/blog/creating-game-lobby-system-phoenix-websockets/) as the first result after Guardian and the Phoenix.Token issue #699. And one of the first lines in the whole article, which I really disagreed with and disliked, is this:

>Be wary that Phoenix and Elixir are still under heavy development and prone to backwards-incompatable changes. 

But that is the most useful post I could find - but I figure the better the official docs are, the less people would come across that (IMO) grossly inaccurate statement.

josevalim:
@bill-mybiz I think using "user salt" would be better then! :+1:

chrismccord:
<3 <3 <3

PR #2019, opened 2016-12-02
Introduce current_path and current_url

### LINKED ISSUE #2002: Introduce current_path and current_url
Today those need to be built by hand. See [discussion here](https://elixirforum.com/t/how-to-get-the-current-url/2390/6).

chrismccord:
@josevalim I wasn't sure the best place for these to live. My first goto was the Router.Helpers, but there is no need for us to inject the functions, and the module itself is private (moduledoc false). They are also more discoverable on Phoenix.Controller, but we'll have to import them in web.ex/view, or have folks import as needed. Wdyt?

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -1121,43 +1121,57 @@ defmodule Phoenix.Controller do
     conn.request_path <> "?" <> URI.encode_query(params)
   end
 
-  @doc """
+  @doc ~S"""
   Returns the current request URL, with and without query params.
 
-  See `current_path/1` for details on how the request path
-  is generated. By default, the connection's endpoint will be
-  used for URL generation, but the endpoint may also be
-  explicitly provided to geneerate a URL with the connection's
-  request path, but for an endpoint that the connection did
-  not originate from.
+  The connection's endpoint will be used for URL generation.
+  See `current_path/1` for details on how the request path is generated.
 
   ## Examples
 
       iex> current_url(conn)
-      "www.endpoint1.example.com/users/123?existing=param"
+      "https://www.example.com/users/123?existing=param"
 
       iex> current_url(conn, %{new: "param"})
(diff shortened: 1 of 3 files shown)


josevalim:
It looks great to me. My only question is if we should expose the version
that accepts the endpoint. I can't think of a use case for it, so maybe we
should make that clause private?

chrismccord:
I also went back and forth on the need for the endpoint variant. I'm fine killing it, but the reason I added it was this. Imagine we don't have it, and someone who needs it will do:

    Endpoint.url() <> current_path(conn)

There no problem, unless the request path is "/" without query params, then they will end up with `"example.com/"` with a trailing slash, which is something our `current_url` handles. I could go either way, and it's always easy for us to add later if people are needing it. I'll let you decide to kill it with :fire: or not :)

josevalim:
@chrismccord I believe we should always include the "/". It is a required part of URLs. So we can nuke the endpoint one if that is the reason of its existence. Btw, why are we not including the scheme? Such as http or https, similar to how our router helpers would do?

lib/phoenix/controller.ex
```diff
@@ -1092,6 +1092,73 @@ defmodule Phoenix.Controller do
...
+    current_url(conn, endpoint_module(conn), params)
+  end
+  def current_url(%Plug.Conn{} = conn, endpoint) when is_atom(endpoint) do
+    Path.join(endpoint.url(), current_path(conn))
+  end
+  def current_url(%Plug.Conn{} = conn, endpoint, %{} = params) when is_atom(endpoint) do
+    Path.join(endpoint.url(), current_path(conn, params))
+  end
+
+
```
josevalim: ✂️ maybe?

chrismccord:
Scheme is definitely needed. Will make the change, nuke endpoint, and always include /

PR #2020, opened 2016-12-02
fix brunch-config.js in installer to be compatible with 2.9.1

Gazler:
In brunch 2.8.2 the following is valid in autoRequire:

  "js/app.js": ["app"]

In brunch 2.9.1 this has to be:

  "js/app.js": ["js/app"]

Otherwise the following error will be logged and the JavaScript will not
execute:

 > Uncaught Error: Cannot find module 'js/app' from '/'

The difference in the compiled JavaScript prior to this change was:

    $ diff /tmp/app-broken.js /tmp/app-working.js
    151c151
    < require.register("js/app.js", function(exports, require, module) {
    ---
    > require.register("app.js", function(exports, require, module) {

I'd appreciate if someone else could validate the solution just to make sure it isn't only happening to me.

### DIFF
--- a/installer/templates/assets/brunch/brunch-config.js
+++ b/installer/templates/assets/brunch/brunch-config.js
@@ -53,7 +53,7 @@ exports.config = {
 
   modules: {
     autoRequire: {
-      "js/app.js": ["app"]
+      "js/app.js": ["js/app"]
     }
   },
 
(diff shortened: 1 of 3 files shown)


chrismccord:
<3 <3 <#

PR #2027, opened 2016-12-07
Add IncorrectParamError to handle incorrect values in request params

gliush:
Quite often when you have a reach API you need to create exception to handle incorrect param values in the incoming requests.
So instead of having everyone to create a new exception - it's better to create a common one.

### DIFF
--- a/lib/phoenix/exceptions.ex
+++ b/lib/phoenix/exceptions.ex
@@ -15,6 +15,22 @@ defmodule Phoenix.NotAcceptableError do
   defexception message: nil, accepts: [], plug_status: 406
 end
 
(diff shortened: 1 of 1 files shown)


michalmuskala:
I'm not sure it makes sense to create an exception that's not used by phoenix itself. The discoverability would be very poor.

gliush:
What if I add couple of tests? Though, I don't yet understand how to create a good one.

gliush:
And also a notion in documentation might be added some value

chrismccord:
Thanks! I don't think this is something we need in Phoenix. You could add an exception to your own app with a plug_status and use this, but I'd prefer to leave 400 handling to non-raising code. Thanks!

PR #2029, opened 2016-12-08
Add command for generating a self-signed cert to dev.exs

Gazler:
The command has been written using a heredoc so that the multi-line
command can be easily copied and pasted to the terminal. The openssl
command has been deliberately constructed so that it does not require
any user input to minimize confusion for the end user.

Using the heredoc is probably a bad idea, I have included it so you can copy
and paste it in for testing to make sure it works. It works fine on Linux, I am
not sure if openssl is different on OS X


josevalim:
I would do just one very big line I will be able to copy and paste into the terminal (no heredocs).

Gazler:
@josevalim That's fine. I guess if people are annoyed by the long line they can easily remove it.

chrismccord:
❤️❤️❤️🐥🔥

PR #2030, opened 2016-12-10
removed destructured 'ref' from push.matchReceive

elbow-jason:
Was going over the JS and spotted an unused variable.

Cheers.


chrismccord:
I'd rather leave it as self documenting function. Thanks!

PR #2036, opened 2016-12-15
Added default unprocessable entity behavior on create and update action templates when there is no match for the namespace on user's payload

mustafaturan:
Added default unprocessable entity behavior on create and update action templates when there is no match for the namespace on user's payload

**Description**
This PR adds two new pattern matches for create/update action templates to handle RESTful pattern. 

**Story**
When a user forgets to add namespace on payload, on create/update endpoints(generally on API endpoints), the user gets HTTP status 400 with text "Internal server error", instead of HTTP status code 422 and missing fields. In the RESTFul approach, we should response with 422 instead of 400 and valid JSON. 

**Tasks**
- [x] Add create action to the controller and controller test templates with any params and return 422
- [x] Add update action to the controller and controller test templates with any params and return 422

### DIFF
--- a/priv/templates/phoenix.gen.html/controller.ex
+++ b/priv/templates/phoenix.gen.html/controller.ex
@@ -25,6 +25,10 @@ defmodule <%= module %>Controller do
         render(conn, "new.html", changeset: changeset)
     end
   end
+  def create(conn, _) do
+    changeset = <%= alias %>.changeset(%<%= alias %>{}, %{})
+    render(conn, "new.html", changeset: changeset)
+  end
 
   def show(conn, %{"id" => id}) do
     <%= singular %> = Repo.get!(<%= alias %>, id)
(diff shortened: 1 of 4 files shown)


josevalim:
Thank you @mustafaturan!

While this seems to be the correct default, I would prefer to have this as part of Phoenix parameter handling instead of introducing boilerplate on every action that receives parameters. We should either consider making the action mismatch error use status 422 or provide some sort of configuration, but I wouldn't change the generators.

chrismccord:
I agree with José. We already provide this functionality with the ActionClauseError, but it only 400's. To get what you want, you could define call/2, invoke super, and rescue from `Phoenix.ActionClauseError`, sending your own 422 response. I'm open to configuration suggestions if you have any, but it would have to be programmatically done on a controller by controller basis. Even if the status was configurable, you'd still need to customize your json response, so rescuing might be the best option in general.

mustafaturan:
Hi @joeellis and @chrismccord 

Thanks for the comments. I see your points but on create/update functions; handling this condition either by 'pattern matching with creating new functions' or 'accepting all params and handling the params inside the function' will be my choice in a project. Because in my case, error messages should change depending on input and should give a more descriptive information to the client.

`Phoenix.ActionClauseError` is generic and IMHO rescuing from `Phoenix.ActionClauseError` will increase complexity. But good to know that I have this option with call/2.

PR #2037, opened 2016-12-17
Add summary and examples for `Phoenix.Token.verify/4` (#1892)

byelipk:
This PR adds a new summary and example usage to the `Phoenix.Token.verify/4` documentation as discussed in #1892. 

It also tries to remain consistent with the theme of #2013  by swapping out "user" for "user salt" and creating a detailed example scenario.

### DIFF
--- a/lib/phoenix/token.ex
+++ b/lib/phoenix/token.ex
@@ -113,7 +113,44 @@ defmodule Phoenix.Token do
   end
 
   @doc """
-  Decodes the original data from the token and verifies its integrity.
+  Given a secret key base, a salt, a token, and optional arguments `verify/4`
+  will establish the authenticity and integrity of the token and then decode
+  and return the original data.
+
+  ## Examples
+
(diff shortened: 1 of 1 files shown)


ksolo:
@byelipk I think this reads clearly and like the change.

@chrismccord @josevalim Thoughts? Can this be merged in?

lib/phoenix/token.ex
```diff
@@ -113,7 +113,44 @@ defmodule Phoenix.Token do
   end
 
   @doc """
-  Decodes the original data from the token and verifies its integrity.
```
josevalim: The first line of the documentation must always be a short summary. So we should keep the line above or right a new one. :)

josevalim:
Thank you @byelipk, I have added one comment!

byelipk:
Thanks @josevalim, I reverted back to the original summary to keep it short and simple.

josevalim:

PR #2041, opened 2016-12-22
Add Elixir 1.4 to travis allowed failures

devonestes:
It looks like we'll get 1.4 soon, and I thought adding the release
candidate to the travis matrix as an allowed failure might help in
finding and squashing any bugs in the release candidate, as well as
give folks time to make any updates or address any deprecations that
might be coming in 1.4.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -5,6 +5,15 @@ matrix:
       elixir: 1.3.2
     - otp_release: 19.0
       elixir: 1.3.2
+    - otp_release: 18.3
+      elixir: 1.4.0-rc.1
+    - otp_release: 19.0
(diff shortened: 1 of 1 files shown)


michalmuskala:
It looks like the failures on travis are only related to a different error message format from `OptionParser`. I think it should be possible to fix them and add those builds as normal ones.

Gazler:
@michalmuskala This issue has been fixed on the new generators branch in this commit https://github.com/phoenixframework/phoenix/commit/d5970eca74c07a9e576a8e61c5a3a89581aa8369

chrismccord:
Thanks! I added the fix from d5970ec and removed the `allowed_failures` bit since we pass across all versions.

PR #2042, opened 2016-12-24
Update the model_case.ex template errors_on function to accept any changeset function, defaulting to :changeset.

ptlu:
Found this useful when creating multiple changeset functions to handle different scenarios.  Means that testing error messages are easy to handle in a consistent way.


josevalim:
Hi @ptlu, that function is meant to be kept simple on purpose. Developers are free to improve it in their apps. Thank you!

PR #2045, opened 2016-12-26
Test and document Router catch-all verb matches

christopheradams:
This adds some tests and a note in the Router documentation about using `:*` as an HTTP method catch-all in the `match` macro. The behavior already existed but was not documented. Thanks @chrismccord for pointing this out!

Addresses issue #977.


chrismccord:
❤️❤️❤️🐥🔥

cjhowe7:
Thanks!

PR #2057, opened 2017-01-06
Adds Presence.get_by_key with default

adkron:
Sometimes we want just a single piece of information out of the Presence
tracker. In order to facilitate that we want to be able to grab from the
tracker with a key and not get the entire map. I mapped this to the same
behavior as `Map.get` so it would have the ability to give a default if
one doesn't exist. By default `Map.get` returns `nil` if something isn't
in the map. Since we are expecting map to come back from Presence I
decided to make this default into an empty map.

I'm not sure if this is what was looked for in issue ##1777, but that is what I was trying to address. Is this the right track?

### DIFF
--- a/lib/phoenix/presence.ex
+++ b/lib/phoenix/presence.ex
@@ -105,6 +105,7 @@ defmodule Phoenix.Presence do
   @type presences :: %{ String.t => %{metas: [map()]}}
   @type presence :: %{key: String.t, meta: map()}
   @type topic :: String.t
+  @type key :: String.t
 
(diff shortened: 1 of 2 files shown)


adkron:
I just saw a comment from @josevalim about this needing to be in tracker. I can try my hand at implementing it there. Is this repo and phoenix_pub_sub released in lock step with each other? To expose it here and implement it in tracker we will need to have it in both repos at the same time.

chrismccord:
Phoenix.Tracker needs to expose an api that allows us to do this efficiently. This solution works, but requires fetching all presences to get a single one, which is not ideal. I will tackle the work on the Phoenix.PubSub side to enable this. Thanks!

PR #2060, opened 2017-01-09
Make Presence.handle_diff overridable

max-vogler:
I'm using `Presence` to track online counts in a high performance application, where I would like to save the bandwidth consumed by the `presence_diff`s (my client's don't need them).

I didn't find an intuitive solution for that. Instead I had to copy the `Presence` module and change `handle_diff`.

Allowing to override `handle_diff`would allow this and other custom `presence_diff`s.

### DIFF
--- a/lib/phoenix/presence.ex
+++ b/lib/phoenix/presence.ex
@@ -172,7 +172,7 @@ defmodule Phoenix.Presence do
         {:ok, state}
       end
 
-      defoverridable fetch: 2
+      defoverridable [fetch: 2, handle_diff: 2]
     end
   end
 


chrismccord:
There's also an easy way to accomplish this today by adding the following in your channel(s):

```elixir
intercept ["handle_diff"]

def handle_out("handle_diff", _, socket), do: {:noreply, socket}
```

This way, any elixir process can track presences changes instead of the all or nothing approach that looping handle_diff would cause. 

If you really wanted to override handle_diff, I think the better way would be to implement the Phoenix.Presence behavior (which just calls into Phoenix.Tracker):
https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/presence.ex

I say this because I can't think of a good reason to override, unless you really know what you are doing with the internals, in which case making your own behavior is probably the way to go instead of us exposing it for overrides. I would simply intercept and drop the messages at the cost of slightly more work with message sending. Thanks!

max-vogler:
@chrismccord, thanks for the great explanation ❤️! Indeed, I tackled the problem at the wrong module. On a minor note, `handle_diff`didn't do the job, but `presence_diff` did!

```elixir
intercept ["presence_diff"]

def handle_out("presence_diff", _, socket), do: {:noreply, socket}
```

PR #2047, opened 2016-12-27
Add channel js tests

rossta:
This PR builds on #1999 by adding JavaScript tests for the `Channel` class in `phoenix.js`. Tests should cover most the key behavior around joining/leaving channels, timeouts, and pushing data to channel topics.

PR #2062, opened 2017-01-13
[UNTESTED] Match * to any receive event in phoenix.js

bopjesvla:
Enables:

```js
channel.push('new:thing', thing)                                                                                                                                                             
       .receive('error', err => this.errors = err.errors)                                                                                                                                               
       .receive('ok', _ => e.target.reset())                                                                                                                                                            
       .receive('*', _ => findButton(e.target).disabled = false)                                                                                                                                                                      
```

I don't think channel behavior is tested at all so proceed with some caution.

### DIFF
--- a/priv/static/phoenix.js
+++ b/priv/static/phoenix.js
@@ -272,7 +272,7 @@ var Push = function () {
           ref = _ref.ref;
 
       this.recHooks.filter(function (h) {
-        return h.status === status;
+        return h.status === status || h.status === "*";
       }).forEach(function (h) {
         return h.callback(response);
       });
--- a/web/static/js/phoenix.js
+++ b/web/static/js/phoenix.js
@@ -242,7 +242,7 @@ class Push {
   // private
 
   matchReceive({status, response, ref}){
-    this.recHooks.filter( h => h.status === status )
+    this.recHooks.filter( h => h.status === status || h.status === "*" )
                  .forEach( h => h.callback(response) )
   }
 


rossta:
FWIW, there are channel.js tests waiting for review in #2047

chrismccord:
what is the usecase here, to avoid having the disable state changed in two places? I still would rather write:

```js
resetSubmit(button){
  findButton(button).disabled = false
}

onSubmitError(err, button){
  this.errors = err
  resetSubmit(button)
}

someFunction(){
  channel.push('new:thing', thing)                                                                                                                                                             
    .receive('error', err => onSubmitError(err, e.target))                                                                                                                                               
    .receive('ok', _ => resetSubmit(e.target))          
}                                                                                                                                                  
```

I'm not convinced a catch-all is needed for receive hooks and it also introduces a special event, "*". Thanks!

bopjesvla:
The alternative doesn't handle timeouts, and makes it much less clear that the button is re-enabled regardless of what happened. Why would you want to move a global response to every individual hook?

Point taken about the special event, though. Would you accept a revision that takes `.receive(fn)` instead of `.receive('*', fn)`?

chrismccord:
> The alternative doesn't handle timeouts, and makes it much less clear that the button is re-enabled regardless of what happened. 

Right, you'd need to handle it explicitly in your case. UI concerns may be different as well for success, error, and network timeout cases, so it make not make sense in all cases to catchall and re-enable. Often you'll have some specific UI hook that needs to happen, i.e. "Sorry it looks like your network is having problems", and it makes sense there to re-enable the button state in my opinion

My concern is mainly around semantics of the receive. Today only a single receive "clause" is invoked and then the receive chain is done. This PR would introduce a receive hook that is invoked with other receive hooks without cancelling the chain. So if we supported something like `receive(fn)` then I would prefer it to be the only one invoked, which is not what you want.

Keep in mind you could do something like this to handle all cases with your existing code:

```js
let onReceive = (callback) => {
  findButton(e.target).disabled = false
  return callback || function(){}
}

channel.push("new:thing", thing)
  .receive("ok", onReceive(err => this.errors = err.errors))
  .receive("error", onReceive(resp => e.target.reset()))
  .receive("timeout", onReceive())
```

bopjesvla:
That is what I'm doing at the moment, but when the UI changes are more complex it makes sense to split the global clean-up from the individual response handling as not to lose oversight.

> Today only a single receive "clause" is invoked and then the receive chain is done.

`push.receive('ok', fn1).receive('ok', fn2)` calls both functions, but I see what you mean. Rename the global hook to `always`?

PR #2065, opened 2017-01-15
Avoid running compilers not present in the application's compiler list.

ciaran:
This would cause an error in applications which did not configure their `reloadable_compilers` option,
and did not have a default compiler (gettext) present.


josevalim:

PR #2070, opened 2017-01-22
Remove unnecessary Phoenix.Param.Any.deriving/3 public function

lexmag:
It was in use before in Phoenix.Param.Map, but is not anymore
after the dropping of Elixir v1.1 support.


josevalim:

PR #2075, opened 2017-01-26
Adds clarification to what term means for flash

adkron:
A few times I have seen people ask how to put more than one
message under the same flash key. Adding the idea of a what
a term in here may further what a term means in the specs and
keep people from needing to search as far.

### DIFF
--- a/lib/phoenix/test/conn_test.ex
+++ b/lib/phoenix/test/conn_test.ex
@@ -276,7 +276,9 @@ defmodule Phoenix.ConnTest do
   defdelegate get_flash(conn, key), to: Phoenix.Controller
 
   @doc """
-  Puts the given value under key in the flash storage.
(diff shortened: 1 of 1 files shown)


adkron:
This is a documentation change. I'm not sure why travis is failing.

chrismccord:
Since our generators, and the majority of applications use a single message for a given key, I would prefer to keep the docs from mentioning this specific use case. If were are to change the docs, It should be at the Phoenix.Controller level, using an  couple example calls underneath an `## Examples` section. Can you amend the PR to show a couple example calls in Phoenix.Controller.put_flash/2? Thanks!

adkron:
I didn't even notice that I put this in the test file. I will add the examples in the controller. Thanks for the feedback.

adkron:
Closing in favor of #2077

Issue #2074, opened 2017-01-26
mix phoenix.digest.clean should probably call build_structure


aaronjensen:
`mix phoenix.digest` [calls `Mix.Project.build_structure()`](https://github.com/phoenixframework/phoenix/blob/master/lib/mix/tasks/phoenix.digest.ex#L46-L50) after it digests to ensure that what is generated in priv is copied to _build.

Should we also call it after `mix phoenix.digest.clean` so that things deleted in priv will also be deleted in _build?

josevalim:
Yes please!

aaronjensen:
Ok, I'll fix that and a few other nasty bugs in digest.clean. Pull request coming soon.

PR #2076, opened 2017-01-27
Fix digest clean

aaronjensen:
This addresses #2074 and fixes several other bugs with the new digest/digest.clean.

Previously:

1. `mtime` was updated to the current time for all assets, including those that were not digested this time around.
2. If a file was deleted manually, it would not be removed from the manifest the next time digest was run.
3. In manifests affected by some of the other bugs this pull fixes, it was possible that clean would delete the latest version of an asset.
4. Cleaned assets were not removed from the manifest.
5. Cleaned assets were not also removed from the `_build` directory when `build_embedded` was true (#2074)
6. The documentation for `digest.clean` was incorrect.

I added test cases to cover these and some other edge cases that were exposed while I was refactoring. Please let me know if there's any changes you'd like me to make.

Thanks!


josevalim:

PR #2077, opened 2017-01-27
Adds example usage of storing a list in flash

adkron:
#A few times I have seen people ask how to put more than one
message under the same flash key. Adding the idea of a what
a term in here may further what a term means in the specs and
keep people from needing to search as far. I didn't want to go too
far with examples so I stopped at the list.


chrismccord:
Thanks @adkron! I think storing other terms in the flash like lists is an edge case for most folks, so I don't feel it needs to be included in the examples, but I appreciate your efforts to make things better for phoenix users! <3 <3 <3

PR #2079, opened 2017-02-02
Update faulty @doc for Phoenix.Controller.get_flash/1

odiumediae:
Fix the documentation that wrongly states the function `Phoenix.Controller.get_flash/1` would return `nil` if no flash messages are set, even though `conn.private.phoenix_flash` is an empty map by default.


chrismccord:
❤️❤️❤️🐥🔥

PR #2080, opened 2017-02-03
Fix phx.gen.html context_test.exs template

wojtekmach:
cc @chrismccord


chrismccord:
Whoops, good catch! ❤️❤️❤️🔥🐥

PR #2082, opened 2017-02-06
Remove javascript-brunch because it is deprecated

ali-l:
Bump brunch to 2.10, as this is the minimum required to work without javascript-brunch


chrismccord:
❤️❤️❤️🐥🔥

PR #2086, opened 2017-02-12
Fix name clash with web app manifest file

bryannaegele:
#### Problem

`Phoenix.Digester` overwrites a `manifest.json` file placed by the user in the `static` folder when generating a manifest for use by `Endpoint` caching mechanism. Phoenix should not overwrite this file.

#### Background

This is an attempt to not get in the way of PWAs by changing the name for the manifest file generated by `mix phoenix.digest` from `manifest.json` to `cache_manifest.json`. `manifest.json` is a standard name file for the (Web App Manifest)[https://developer.mozilla.org/en-US/docs/Web/Manifest] file, though other names are proposed. While naming may be in flux, we should definitely get out of the way of existing standards.

MDN currently (references)[https://developer.mozilla.org/en-US/docs/Web/Manifest] the name of the file to be `manifest.manifest`, however Google has and continues to (reference)[https://developers.google.com/web/fundamentals/engage-and-retain/web-app-manifest/] `manifest.json` for the file.

Since the purpose of the generated manifest file is ultimately for use by the `cache_static_manifest` setting in `Endpoint` we can be more descriptive with the file name while not overwriting a `manifest.json` file purposefully placed by the user in the `static` folder for their PWA.

Thanks to @chrismccord for providing insight on what the purpose of the manifest was.

If this is satisfactory, I can put a PR up for the guides to reconcile the name change.


josevalim:
This looks good, thanks! However I believe there is no need to rename function names and variables in Phoenix.Digester.

bryannaegele:
I went back and forth on whether to change those to be specific to the task or leave them as is to preserve the API. 

I can revert the changes to the function and variable names, but I think the documentation changes should keep the references to being cache specific to clarify intent. Does that sound good?

josevalim:
It sounds good, yes!

bryannaegele:
Updated per feedback.

test/fixtures/manifest.json
```diff
@@ -1,14 +1,33 @@
 {
```
josevalim: I believe this file and the one above are here by accident?

bryannaegele: This one was an accident. 

The one above could be removed if we only want to run the tests on files created by the generator.

josevalim: Oh, I see. The other one is actual manifest. Yeah, we can keep it.

bryannaegele: Yes. I mainly left it in as an example for anyone digging in the tests, but it's also nice for preventing a regression and to represent another common filetype in the compression/manifest creation process. I could see addition of other common filetypes in the future, such as fonts. It really just depends on the desired testing strategy.

bryannaegele:
Build failure appears to be unrelated to the last commit on this.

josevalim:

PR #2087, opened 2017-02-12
Add @shortdoc for phx.new.web and phx.new.ecto

shankardevy:
Without the `@shortdoc` in both these tasks, they don't show up in `mix help`. This PR fixes it.

### DIFF
--- a/installer/lib/phx_new/web.ex
+++ b/installer/lib/phx_new/web.ex
@@ -28,6 +28,10 @@ defmodule Mix.Tasks.Phx.New.Web do
   use Phx.New.Generator
   alias Phx.New.{Project}
 
+  @version Mix.Project.config[:version]
+  @shortdoc "Creates a new Phoenix v#{@version} application inside an umbrella application"
+
+
   @pre "phx_umbrella/apps/app_name_web"
 
   template :new, [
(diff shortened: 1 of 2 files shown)


installer/lib/phx_new/web.ex
```diff
@@ -28,6 +28,10 @@ defmodule Mix.Tasks.Phx.New.Web do
   use Phx.New.Generator
   alias Phx.New.{Project}
 
+  @version Mix.Project.config[:version]
+  @shortdoc "Creates a new bare Phoenix v#{@version} application without database integration inside an umbrella application"
```
josevalim: The shortdoc needs to be shorter. :) Also, no need for that many new lines afterwards.

shankardevy:
@josevalim 

shortened the `@shortdoc` ;-)

Regarding the linebreaks, I added it to keep it even with:
https://github.com/phoenixframework/phoenix/blob/master/installer/lib/phx_new.ex#L56-L58

I thought those extra line breaks are to distinguish @shortdoc from the rest of the code. Let me know if those line breaks have to removed.

1. no line break after @shortdoc
2. or 1 line break
3. or 2 line breaks as it's now.

Please advise on the above.

chrismccord:
I already addressed these locally, so no need to merge this one. Thanks for your other PRs and going thru the new generators! <3 <3 <3

PR #2089, opened 2017-02-12
Add missing .gitignore in umbrella app

shankardevy:
`mix phx.new app_name --umbrella` creates a new umbrella app without `.gitignore` file in the root directory. This PR adds a test and fixes the same.

### DIFF
--- a/installer/lib/phx_new/umbrella.ex
+++ b/installer/lib/phx_new/umbrella.ex
@@ -7,6 +7,7 @@ defmodule Phx.New.Umbrella do
   template :new, [
     {:eex,  "phx_umbrella/config/config.exs", :project, "config/config.exs"},
     {:eex,  "phx_umbrella/mix.exs",           :project, "mix.exs"},
+    {:eex,  "phx_umbrella/gitignore",         :project, ".gitignore"},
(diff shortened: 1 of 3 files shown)


shankardevy:
looks like the CI failure is not related to any bug with PR code.

installer/lib/phx_new/umbrella.ex
```diff
@@ -7,6 +7,7 @@ defmodule Phx.New.Umbrella do
   template :new, [
     {:eex,  "phx_umbrella/config/config.exs", :project, "config/config.exs"},
     {:eex,  "phx_umbrella/mix.exs",           :project, "mix.exs"},
+    {:eex,  "phx_umbrella/gitignore",         :project, ".gitignore"},
```
wojtekmach: while at it, I think it would be good to remove node_modules from the gitignore [1] as it's no longer required here

[1] https://github.com/phoenixframework/phoenix/blob/6011906e79b3a7391e0762826fca6ecfa1d197c7/installer/templates/phx_umbrella/gitignore#L6

shankardevy: good catch @wojtekmach. Update .gitignore template.

chrismccord:
❤️❤️❤️🐥🔥

PR #2090, opened 2017-02-14
Align and bump npm package versions.

mareksuscak:
In https://github.com/phoenixframework/phoenix/pull/2082 we forgot to remove deprecated `javascript-brunch` package and align the versions in both installer templates. 

Shall we add a comment to both files that there are two of them to mitigate the reduce the likelihood of doing the same mistake again?

### DIFF
--- a/installer/templates/assets/brunch/package.json
+++ b/installer/templates/assets/brunch/package.json
@@ -11,9 +11,9 @@
   },
   "devDependencies": {
     "babel-brunch": "6.0.6",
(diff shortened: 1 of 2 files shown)


chrismccord:
The previous 1.2 generators are being deprecated so we don't need to worry about maintaining the brunch builds going forward. Thanks!

steakknife:
FYI: [css-brunch](https://github.com/brunch/css-brunch) is also deprecated. 😢

PR #2091, opened 2017-02-14
Update DB setup instructions in the readme file.

mareksuscak:
I have noticed that there’s a mix alias that helps set up the database incl. seeds which might be helpful.

### DIFF
--- a/installer/templates/phx_umbrella/apps/app_name_web/README.md
+++ b/installer/templates/phx_umbrella/apps/app_name_web/README.md
@@ -3,7 +3,7 @@
 To start your Phoenix app:
 
   * Install dependencies with `mix deps.get`
-  * Create and migrate your database with `mix ecto.create && mix ecto.migrate`
+  * Setup your database with `mix ecto.setup`
   * Install Node.js dependencies with `cd assets/ && npm install`
   * Start Phoenix endpoint with `mix phoenix.server`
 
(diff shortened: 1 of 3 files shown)


installer/templates/phx_umbrella/apps/app_name_web/README.md
```diff
@@ -3,7 +3,7 @@
 To start your Phoenix app:
 
   * Install dependencies with `mix deps.get`
-  * Create and migrate your database with `mix ecto.create && mix ecto.migrate`
+  * Setup your database with `mix ecto.setup`
```
wojtekmach: in umbrella, `ecto.setup` alias is defined in the domain logic app and aliases aren't shared between apps, so this does not work out of the box

mareksuscak: oh so what do you suggest then?

wojtekmach: Yeah, I should have written something. Honestly I don't know what's the best answer here, I'm sure maintainers will chime in. I like the `ecto.setup` change in non-umbrella project though!

chrismccord:
Thanks! I would prefer to to keep the existing instructions since they apply to all applications. It also introduces users to tasks they will necessarily need to know about. `mix ecto.setup` is a convenient alias, but I believe the current instructions are best for now. Thanks!

PR #2093, opened 2017-02-14
Fix module name issue with new phx.gen.json and phx.gen.html.

jeregrine:
When I tested gen.json|html it would generate module names like `defmodule App.Web.Web.PostController do`. Now it correctly generates `defmodule App.Web.PostController do`. 

I also went ahead and made sure we weren't matching on Module.split\1 anywhere else.


chrismccord:
❤️❤️❤️🐥🔥

PR #2094, opened 2017-02-14
Use const for immutable bindings.

mareksuscak:
It is a common misconception that ES2015 `const` keyword creates constants or immutable vars but that is not true. The following code is perfectly valid:

```js
const foo = {};
foo.bar = false;

const baz = [];
baz.push('item')
```

The truth is that `const` creates **an immutable binding** which means the following code would throw:

```js
const foo = {}
foo = [] // throws “Assignment to constant variable.”
```

Some eslint configs use [import/no-mutable-exports](https://github.com/benmosher/eslint-plugin-import/blob/24e2c9d27110c04a31323c8cee1c9509a4c8a969/docs/rules/no-mutable-exports.md) rule which causes a linting error in the socket.js file.

Hence I propose to change `let` to `const`.

### DIFF
--- a/installer/templates/assets/brunch/socket.js
+++ b/installer/templates/assets/brunch/socket.js
@@ -5,7 +5,7 @@
 // and connect at the socket path in "lib/my_app/endpoint.ex":
 import {Socket} from "phoenix"
 
-let socket = new Socket("/socket", {params: {token: window.userToken}})
+const socket = new Socket("/socket", {params: {token: window.userToken}})
 
 // When you connect, you'll often need to authenticate the client.
 // For example, imagine you have an authentication plug, `MyAuth`,
@@ -54,7 +54,7 @@ let socket = new Socket("/socket", {params: {token: window.userToken}})
 socket.connect()
 
 // Now that you are connected, you can join channels with a topic:
-let channel = socket.channel("topic:subtopic", {})
+const channel = socket.channel("topic:subtopic", {})
 channel.join()
   .receive("ok", resp => { console.log("Joined successfully", resp) })
   .receive("error", resp => { console.log("Unable to join", resp) })
--- a/installer/templates/static/brunch/socket.js
+++ b/installer/templates/static/brunch/socket.js
@@ -5,7 +5,7 @@
 // and connect at the socket path in "lib/my_app/endpoint.ex":
 import {Socket} from "phoenix"
 
-let socket = new Socket("/socket", {params: {token: window.userToken}})
+const socket = new Socket("/socket", {params: {token: window.userToken}})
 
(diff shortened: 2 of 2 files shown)


chrismccord:
Honestly, I prefer the current generated code but I appreciate the thoughts. Thanks!

eduardonunesp:
This PR was interesting, @chrismccord, use `const` instead of `let` is something very appreciated by the JS community, I think it's a good idea to reconsider this PR. Although you prefer the current generated code, unfortunately this though isn't a good justification to not use `const` instead of `let`.

chrismccord:
To expand. The JS community is still not 100% in unison on const/let usage. I personally agree const is for a constant reference not a constant value, but I am coming from the position that the generated phoenix js may be the first exposure to es20xx as a newcomer. It's a balance of throwing too much at a new user, and in this case, the distinction of const vs let in that file isn't clear in general. If we go with `const`, there isn't a single `let` declaration, which will be confusing for someone who has not used es20xx before. I'd rather push folks towards block scoped variable binding and then let them consider the merits of constant reference vs mutable reference after they level up.

mareksuscak:
> The JS community is still not 100% in unison on const/let usage.

I don't think that's the case anymore. The community is catching up with [most](https://github.com/webpack/webpack/pulls?q=is%3Apr+is%3Aclosed+label%3A%22Refactor+%28ES5+%3D%3E+ES6%29%22) [significant](https://github.com/facebookincubator/create-react-app/search?utf8=%E2%9C%93&q=const+&type=Code) [open source](https://github.com/babel/babel/search?p=32&q=const&utf8=%E2%9C%93) projects advocating the change. [So](https://medium.com/javascript-scene/javascript-es6-var-let-or-const-ba58b8dcde75#.xv7sm0tka) [are](https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Statements/const) [the influencers](https://github.com/airbnb/javascript#references) on the JS scene.

> It's a balance of throwing too much at a new user, and in this case, the distinction of const vs let in that file isn't clear in general. If we go with const, there isn't a single let declaration, which will be confusing for someone who has not used es20xx before.

I totally get what you're saying and it certainly is a solid justification. My thoughts were that replacing `let` with `const` to convey the best practice shouldn't be that big of a deal. We already impose the burden of learning fairly complicated ES modules spec on new users so why not to do that while following the best practices in all other areas? We're encouraging them to use ES2015+ instead of ES5 anyway.

Having said that, I respect that you were not comfortable merging it just yet and I hope you will revisit this idea later. Thank you for thoughts.

eduardonunesp:
Thanks @chrismccord by the expand of your thoughts, as @mareksuscak said. Thanks guys.

PR #2095, opened 2017-02-14
Fixing issue where the Repo is the incorrect module

jeregrine:
It was generating a new schema with the following alias:
```elixir
alias App.Web.Repo
```

Which is incorrect in the case of an umbrella application. So I'm not sure this is the correct fix but if you use phx.gen it will set the repo to override the default.

### DIFF
--- a/lib/mix/tasks/phx.gen.schema.ex
+++ b/lib/mix/tasks/phx.gen.schema.ex
@@ -103,6 +103,10 @@ defmodule Mix.Tasks.Phx.Gen.Schema do
       Mix.raise "mix phx.gen.schema can only be run inside an application directory"
     end
     {opts, parsed, _} = OptionParser.parse(args, switches: @switches)
+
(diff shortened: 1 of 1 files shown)


chrismccord:
Can you sure how you are are running the command? For example, inside which umbrella app? The issue is wifi you run `$ mix phx.gen.schema ..`. inside a non-phoenix app, then the task won't be found unless you are depending on phoenix itself for that app. In the case you are inside your phoenix web project, then it will use the namespace of whatever the application uses, in which case MyApp.Web.Repo would be "correct". I think the correct solution is to only use `phx.gen.schema` in the context of a web project which depends on phoenix, and when outside of that, you can call ecto's generators directly, such as `ecto.gen.migration|repo`. Perhaps ecto needs a gen.schema command? Thanks for the PR. Closing, but feel free to follow up if I'm not making sense. <3 <3 <3

jeregrine:
yea this was the umbrella issue. No worries, ez close.

PR #2096, opened 2017-02-15
Allows atom statuses in `redirected_to`

matthewlehner:
Fixes `redirected_to` when passed an atom instead of an integer for the status. The docs for `redirected_to/2` give the following example, which doesn't currently work. This PR fixes it.

```elixir
      assert redirected_to(conn, :moved_permanently) =~ "/foo/bar"
```


chrismccord:
❤️❤️❤️🐥🔥

Issue #2103, opened 2017-02-19
phoenix default page should be responsive


nhooyr:
### Expected behavior
When scaling down the default phoenix framework page, it should respond appropriately to screen resizes. However, it does not. It looks awkward when resizing to a small size.

### Actual behavior
![screen shot 2017-02-19 at 3 19 40 am](https://cloud.githubusercontent.com/assets/10180857/23100556/4eaeded4-f652-11e6-98a7-049186a626d0.png)

The logo gets clipped and the `Get Started` part ends up above the logo.

josevalim:
Someone please send a PR that improves this in case someone is using the Phoenix page as a template. We probably need to scale the logo and make sure it sits on top of the menu.

josevalim:
Closing it in favor of the PR #2105!

PR #2105, opened 2017-02-19
Make the default page responsive

dwightwatson:
This was made in response to #2103. Please excuse any mistakes as this is my first PR to Phoenix.

This tweaks the default CSS as little as possible so that the framework logo is scaled on mobile devices. The page looks as it already does on larger devices.

I'm more than happy to pop in standard Bootstrap navigation which would collapse by default if it's something you'd be interested in. Thanks!

### DIFF
--- a/installer/templates/static/phoenix.css
+++ b/installer/templates/static/phoenix.css
@@ -34,6 +34,7 @@ form.link, form.button {
   height: 27px;
   background-size: 100%;
   display: inline-block;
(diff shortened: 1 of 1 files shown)


dwightwatson:
Whoops - had saved the screenshot but then forgot to attach it.

Here is is on at the iPhone 6 screen size in the Chrome dev tools:

image

PR #2107, opened 2017-02-21
Add graceful exit contract to transports

chrismccord:
ref: https://github.com/phoenixframework/phoenix_pubsub/issues/68

Previously a channel could gracefully terminate
using the stop semantics of a regular genserver;
however when restarting an application for deploys
the shutdown of the transport and channel processes
would be indistinguisable from an intentional channel
shutdown and would cause clients to incorrectly not
reconnect after server restart. This commit adds a
{:graceful_exit, channel_pid, %Phoenix.Socket.Message{}}
contract to distinguish an intentional channel exit from
what should be regarded as an error condition on the client.

cc @josevalim . Longer term, Saša's design will be a better solution so we can avoid more duplication across transports, but in the interest of shipping 1.3, the `{:graceful_exit, ...` can be handled inline.

### DIFF
--- a/lib/phoenix/transports/long_poll_server.ex
+++ b/lib/phoenix/transports/long_poll_server.ex
@@ -112,6 +112,11 @@ defmodule Phoenix.Transports.LongPoll.Server do
     end
   end
 
+  def handle_info({:graceful_exit, channel_pid, %Phoenix.Socket.Message{} = msg}, state) do
+    new_state = delete(state, msg.topic, channel_pid)
+    publish_reply(msg, new_state)
+  end
+
   def handle_info({:subscribe, client_ref, ref}, state) do
     broadcast_from!(state, client_ref, {:subscribe, ref})
     {:noreply, state}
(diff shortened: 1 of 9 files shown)


lib/phoenix/transports/long_poll_server.ex
```diff
@@ -112,6 +112,11 @@ defmodule Phoenix.Transports.LongPoll.Server do
     end
   end
 
+  def handle_info({:graceful_exit, channel_pid, %Phoenix.Socket.Message{} = msg}, state) do
+    new_state = delete(state, msg.topic, channel_pid)
```
josevalim: OH. If we are deleting the entry from the state then we no longer need to unlink in the channel/server above. Can you please remove the unlink? Other than that this is good to go!

chrismccord: If we go that route then the EXIT will still propagate to the transport and potentially cause failures depending on how the EXIT is handled. For example, it would cause the long poll server to terminate today.

josevalim: @chrismccord I see. Ok, so ship it!

PR #2108, opened 2017-02-21
Update docs of channel.ex

sysashi:
Reply with created record as specified on line 74
instead of replying with changeset.


chrismccord:
❤️❤️❤️🐥🔥

PR #2113, opened 2017-02-23
fix code order. follows code from web module order

yordis:
This follows the order from here https://github.com/phoenixframework/phoenix/blob/cm-phx-gen-next/installer/templates/phx_umbrella/apps/app_name_web/mix.exs#L23

### DIFF
--- a/installer/templates/phx_umbrella/apps/app_name/mix.exs
+++ b/installer/templates/phx_umbrella/apps/app_name/mix.exs
@@ -20,8 +20,8 @@ defmodule <%= app_module %>.Mixfile do
   #
   # Type `mix help compile.app` for more information.
   def application do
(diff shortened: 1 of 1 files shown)


josevalim:
We are getting rid of the :applications key so no need for ordering, thank you!

yordis:
@josevalim sorry I am confused on something, we are getting rid of `:applications` but what happen with `:extra_applications`?

Which it's weird that I have this `extra_applications: [:logger]]` in every single OTP app 😄

chenxsan:
@yordis Guess you could check it here http://elixir-lang.org/blog/2017/01/05/elixir-v1-4-0-released/#application-inference

yordis:
@chenxsan what should I check there?

PR #2114, opened 2017-02-23
Require Elixir v1.4 + rely on the new application inference

jamilabreu:


mix.exs
```diff
@@ -6,7 +6,7 @@ defmodule Phoenix.Mixfile do
   def project do
     [app: :phoenix,
      version: @version,
-     elixir: "~> 1.3",
+     elixir: "~> 1.4",
```
josevalim: Let's keep the version here `~> 1.3` so we keep people upgrading a more breathing room but people starting a new app start fresh.

mix.exs
```diff
@@ -34,7 +34,7 @@ defmodule Phoenix.Mixfile do
 
   def application do
     [mod: {Phoenix, []},
-     applications: [:plug, :poison, :logger, :eex],
+     extra_applications: [:logger, :eex],
```
josevalim: Therefore, let's revert this change BUT leave a TODO so we update it as soon as 1.3 is out.

josevalim:
Thank you! I have added two comments! If you fix them you will likely fix travis too. :)

jamilabreu:
✅

jamilabreu:
One check passing so far - others caused I believe by mismatch in versions

josevalim:

PR #2115, opened 2017-02-23
Bug fix channel join with timeout behavior

### LINKED ISSUE #2035: phoenix.js odd timeout behavior
### Environment

* Elixir version (elixir -v): 1.4.0-rc.1
* Phoenix version (mix deps): 1.2.1
* NodeJS version (node -v): 7.2.0
* NPM version (npm -v): 3.10.9
* Operating system: Windows 10 Pro 64 bits

# First behavior
1. Join a channel
2. Wait for the timeout to happen
3. Connect the socket

### Expected behavior
One channel join request is sent to the server
### Actual behavior
Two join requests are sent to the server, the reply to the first one is ignored because it timed-out before it was even sent

# Second behavior
1. Put a Process.sleep(5000) on the join callback of the channel
2. Join the channel
3. Wait 6 secs
4. Connect the socket

### Expected behavior
One channel join request is sent to the server
### Actual behavior
Multiple join requests are sent and some are sent with a ref value of null as seen on the screenshot below:
http://i.imgur.com/7QSRd76.png

# Repoducible Example
Reproducible demo of the second behavior: https://github.com/ghalle/bugdemo
To test the first one, comment out the Process.sleep and increase the setTimeout to 12 seconds

# Investigation
The first behavior happens because the `startTimeout` is called as soon as the `join` is called rather than when it is actually sent through the socket, I suggest `startTimeout` only be called when the message is actually sent through the socket

The issue with the ref being null seems to be caused by `startTimeout` returning early because a `timeoutTimer` already exists, since the ref is set inside `startTimeout` the message is being sent with a null ref.

UPDATE: Dug a little deeper and found out that this is caused by the rejoinTimer calling `rejoinUntilConnected` faster than the timeout of `startTimeout`. The default values for the `reconnectAfterMs` function used by the rejoinTimer are `[1000, 2000, 5000, 10000]` while the default join timeout is `10000` so if it takes a few secs for the server to reply to the join, the client will try to rejoin again before there is a reply or before it times out

rossta:
Fixes #2035 

- [x] - Test for joining after socket and channel delay
- [x] - Test for joining after socket delay only
- [x] - Test for retries with backoff
- [x] - Implement fix without introducing API change (i.e., without adding callbacks/promises, etc.)

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -256,7 +256,7 @@ class Push {
     this.timeoutTimer = null
   }
 
-  startTimeout(){ if(this.timeoutTimer){ return }
+  startTimeout(){ if(this.timeoutTimer){ this.cancelTimeout() }
     this.ref      = this.channel.socket.makeRef()
     this.refEvent = this.channel.replyEventName(this.ref)
 
--- a/assets/test/channel_test.js
+++ b/assets/test/channel_test.js
@@ -132,6 +132,43 @@ describe("join", () => {
       assert.equal(spy.callCount, 1)
     })
 
+    it("retries with backoff after timeout", () => {
+      const spy = sinon.spy(socket, "push")
+      const timeout = joinPush.timeout
+      let callCount = 1
+
+      socket.connect()
+      helpers.receiveSocketOpen()
+
+      channel.join()
+      assert.equal(spy.callCount, callCount)
+
+      clock.tick(timeout-100)
+      assert.equal(spy.callCount, callCount)
+
+      clock.tick(100)
+      assert.equal(spy.callCount, callCount)
+
+      clock.tick(1000)
+      assert.equal(spy.callCount, ++callCount)
+
+      clock.tick(2000)
+      assert.equal(spy.callCount, ++callCount)
+
+      clock.tick(5000)
+      assert.equal(spy.callCount, ++callCount)
+
+      clock.tick(10000)
+      assert.equal(spy.callCount, ++callCount)
+
+      joinPush.trigger("ok", {})
+      assert.equal(channel.state, "joined")
+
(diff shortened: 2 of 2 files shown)


rossta:
I've created failing tests for the unexpected behavior caused by multiple, overlapping channel join requests when connecting to socket is delayed as described in #2035. Fix is still pending.

chrismccord:
I'm currently looking at this now. I do believe there is a bug, but I also think the dup joins received on the server are correct. If the join times out on the client, the server won't know, so when the client rejoins, it should kill the existing channel.

rossta:
OK, feel free to pull and modify these tests if you think they're helpful. 

If I recall correctly from looking at this previously, it seems like the channel join timeout shouldn't start until the socket push has actually occurred; when socket isn't connected, pushes are buffered but the channel join currently assumes the request is sent immediately.

rossta:
Finally got back around to addressing this issue. Since "multiple joins" is correct as discussed, the fix turns out to be as simple as cancelling the existing push timeout when a new one is scheduled. This ensures the corresponding channel is synced with the most recent join attempt and prevents join attempts with `null` refEvents, as observed in #2035.

ssboisen:
@chrismccord is this going to get merged?

chrismccord:
I have a V2 version of the channel protocol that may touch this area, so standby while I wrap things up on that side. Thanks!

ssboisen:
alright, thanks! Would that protocol change be part of the final 1.3.0 or does it require a major bump?

chrismccord:
It is backwards compatible and will ship with 1.3.0

chrismccord:
Sorry for letting this hang for so long. This issue has been taking care of on master. Thanks!

PR #2122, opened 2017-02-27
Remove mentions of "ecto models" from mix tasks documentation

christos:
I believe (newbie alert!), that these two mentions of "ecto models" in the new `phx` mix tasks documentation were overlooked in #1927.

This PR fixes that.


josevalim:

Issue #2124, opened 2017-03-01
`mix phoenix.new dev_app --dev` fails inside-phoenix-directory test


bryanstearns:
Following the [instructions in the README](https://github.com/phoenixframework/phoenix#generating-a-phoenix-project-from-unreleased-versions), I attempted to generate a new project within the installer directory.

Note that this is using the *old* generator (which is what the README says to use); the new generator works fine (and is actually what I wanted, of course, but didn't remember to use the new name until I started tracking down this problem).

### Environment

* Elixir version (elixir -v): 1.4.2
* Phoenix version (mix deps): current master, c5d5b6a8
* NodeJS version (node -v): 7.6.0
* NPM version (npm -v): 4.1.2
* Operating system: macOS 10.12.3 Sierra

### Expected behavior

A new project would be created.

### Actual behavior

`mix phoenix.new dev_app --dev` failed with `** (Mix) --dev projects must be generated inside Phoenix directory`.

This appears to be happening because the old generator code moved down two directory levels in [this commit](https://github.com/phoenixframework/phoenix/commit/0bda09e33faab0e417da03f322f82c3264f499ea), but the relative path calculation when setting `@phoenix` in `Mix.Tasks.Phoenix.New` (which uses `__DIR__`) didn't change.

PR coming in a minute :-)

PR #2125, opened 2017-03-01
Update relative path to installer directory (#2124)

bryanstearns:
`installer/lib/mix/tasks/phoenix.new.ex` got moved down two directory levels without updating the `@phoenix` path calculation, which uses `__DIR__`, causing `mix phoenix.new foo --dev` to fail.


chrismccord:
❤️❤️❤️🐥🔥

PR #2128, opened 2017-03-01
Remove model switch from phx.gen.html.ex

jamilabreu:
Does it make sense to add a “-no-schema” option to phx.gen.html/json? May be useful in umbrella apps where schema goes in another app/folder.


chrismccord:
❤️❤️❤️🐥🔥

PR #2134, opened 2017-03-02
Replace errors_on/2 with errors_on/1

wojtekmach:
As discussed on https://github.com/phoenixframework/phoenix/commit/d77cddbb0dc00d6f1edcbcb21b7f266a3d0614a2#commitcomment-21075635

### DIFF
--- a/installer/templates/phx_ecto/data_case.ex
+++ b/installer/templates/phx_ecto/data_case.ex
@@ -36,19 +36,18 @@ defmodule <%= app_module %>.DataCase do
   end
 
   @doc """
-  A helper that converts the changeset error messages
-  for a given field into a list of strings for assertion:
+  A helper that transform changeset errors to a map of messages.
 
       changeset = Blog.create_user(%{password: "short"})
-      assert "password is too short" in errors_on(changeset, :password)
+      assert "password is too short" in errors_on(changeset).password
+      assert %{password: ["password is too short"]} = errors_on(changeset)
 
   """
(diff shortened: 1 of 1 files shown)


wojtekmach:
Another idea is to actually have `errors_on/1` instead, that returns the map. We'd support the original use case, but also have an ability to inspect all errors:

```elixir
assert "password is too short" in errors_on(changeset).password
assert %{password: ["password is too short"]} = errors_on(changeset)
```

installer/templates/phx_ecto/data_case.ex
```diff
@@ -44,10 +44,11 @@ defmodule <%= app_module %>.DataCase do
...
   """
   def errors_on(changeset, field) do
-    for {message, opts} <- Keyword.get_values(changeset.errors, field) do
+    Ecto.Changeset.traverse_errors(changeset, fn {message, opts} ->
       Enum.reduce(opts, message, fn {key, value}, acc ->
         String.replace(acc, "%{#{key}}", to_string(value))
       end)
-    end
+    end)
+    |> Map.get(field, [])
```
josevalim: Credo has a check that does not allow a single pipe. So we should rather pipeline the whole way, starting with changeset or break it apart using a variable.

josevalim: Actually, let's go with `errors_on/1` that you have proposed. This way we get rid of the pipe and it is slightly more general purpose.

josevalim:

PR #2135, opened 2017-03-02
Fix ecto generator config path for umbrella apps

moomerman:
When running the new phoenix 1.3 umbrella generator, the path to the database config
file is incorrect.  Eg.

$ mix phx.new /tmp/hello --umbrella

before:

```
Before moving on, configure your database in config/dev.exs and run:

    $ mix ecto.create
```

after:

```
Before moving on, configure your database in apps/hello/config/dev.exs and run:

    $ mix ecto.create
```

A side-effect of this fix is that when you run the phx.new.web generator
the configuration message is no longer displayed (because there is no app_path)
which is correct since the web-only app doesn't have any database config.

### DIFF
--- a/installer/lib/mix/tasks/phx.new.ex
+++ b/installer/lib/mix/tasks/phx.new.ex
@@ -131,7 +131,7 @@ defmodule Mix.Tasks.Phx.New do
         Task.await(compile, :infinity)
 
         print_mix_info(project.project_path, extra)
-        if Project.ecto?(project), do: print_ecto_info()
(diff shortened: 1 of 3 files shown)


josevalim:
@moomerman can you please include a test? You should be able to assert on the messages of the existing web and umbrella generators. Let me know if you have any questions.

moomerman:
@josevalim added tests to ensure:

- the previous non-umbrella app path is unchanged
- the new umbrella app path is correct
- the web-only one should not have the message

Let me know if I've missed anything.

josevalim:

PR #2141, opened 2017-03-02
Remove deprecated phoenix.digest.clean

aaronjensen:
This was never released, so there is no need to deprecate it.


chrismccord:
I *just* killed it on master myself. Thanks!!!

PR #2147, opened 2017-03-03
Mention conn.path_params in CHANGELOG

christopheradams:
I have seen at least [one question](https://elixirforum.com/t/access-conn-path-params-in-phoenix-app/3902) asking why `conn.path_params` doesn't work in Phoenix 1.2 even with Plug 1.3.

Now that support was added in dad413f, we might as well document it, at least in the CHANGELOG. :)

This PR also fixes capitalization of the previous line to match the rest of them.


josevalim:

PR #2159, opened 2017-03-05
Hide precheck in issue template from appearing in issues

LostKobrakai:
Html comments allow the precheck part of the issue template to remain visible for the issue creator, but it's not going to be visible as part of the issue when rendered by github.


chrismccord:
<3 <3 <3

PR #2161, opened 2017-03-05
installer: fix reference to missing module in config/prod.exs

develop7:
Generated `config/prod.exs` referred non-existing `Demo.Web.Endpoint` module, while it should refer `endpoint_module`

Fixes prompt `MIX_ENV=prod mix phx.server` crash on start with reason 

```
…{undef,[{'Elixir.Demo.Web.Endpoint',load_from_system_env…
```


josevalim:
Good catch! Can you also please fix phx_single/config/prod.exs?
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

josevalim:
I went ahead and fixed it since we are working on a new release. :) Thank you!

develop7:
Thanks!

PR #2162, opened 2017-03-05
Don't generate trailing whitespace in context tests

smadeja:
The template for context tests currently generates trailing spaces, and if the generator was called with no resource attributes, also an empty line. This fixes that behaviour.

### before/after with resource attributes
accounts_test_exs_bak_ _phoenix_and_event_controller_test_exs_ _phestiment

### before/after without resource attributes
accounts_test_exs_bak_ _phoenix_and_dev_app_ _-bash


chrismccord:
<3 <3 <3

PR #2166, opened 2017-03-07
Fix create function example in context generator template

VladShcherbin:
Create function has one param while the produced examples have two.

Current produced examples:

```elixir
iex> create_letter(letter, %{field: value})
{:ok, %Letter{}}

iex> create_letter(letter, %{field: bad_value})
{:error, %Ecto.Changeset{}}
```


chrismccord:
Thanks!

PR #2171, opened 2017-03-08
not-web app secret config generated with phx.new --umbrella is not ignored

Joe-noh:
`mix phx.new --umbrella` creates two application in an umbrella project, say `my_app` and `my_app_web`. I found that `prod.secret.exs` in `apps/my_app/config` is not listed in `gitignore`. So the file possibly accidentally be under version controlled. I guess this is not intended.

In this pull request, I add a line simply ignoring `config/*.secret.exs`. Would you let me know If there is any other good solution or some reasons that it is not ignored.

Thanks.

### DIFF
--- a/installer/lib/phx_new/ecto.ex
+++ b/installer/lib/phx_new/ecto.ex
@@ -6,15 +6,16 @@ defmodule Phx.New.Ecto do
   @pre "phx_umbrella/apps/app_name"
 
   template :new, [
-    {:eex, "#{@pre}/config/config.exs",           :app, "config/config.exs"},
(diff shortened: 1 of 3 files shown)


Joe-noh:
noticed that `apps/my_app/.gitignore` is not generated. Is there another problem?

josevalim:
Thank you @Joe-noh! Can you please use this `.gitignore` instead: https://github.com/phoenixframework/phoenix/blob/master/installer/templates/phx_assets/bare/gitignore?

There is no need to define a new one in the project, thank you!

Joe-noh:
@josevalim 

Thank you for your response! I added last two commits!

josevalim:

PR #2178, opened 2017-03-09
Update commands in the watcher's logger

smadeja:
These would normally show up when starting the server in the root
directory, and `npm install` will fail unless it's run in `/assets`.


chrismccord:
Thanks!

PR #2183, opened 2017-03-12
Support signed_at option for Token.sign

tsubery:
This parameter can be useful for integration testing or to deal with edge cases such as day light savings when using short-lived tokens.

I can imagine some might prefer ```System.monotonic_time``` over ```System.system_time```. With this change it would be possible to make that choice. Might make sense to implement a complementary parameter to verify. At the moment we can use max_age to adjust time on ```verify```.

I also added some tests for key options, It might be redundant but better err on the side of too many tests when dealing with security related matters. These test make sure we are at least sending the parameters with correct spelling and provide reasonably secure defaults.


josevalim:
I would prefer to always default to monotonic_time now that we require Elixir v1.3 and not provide an option.

josevalim:
Actually, please ignore my previous comment. :)

chrismccord:
❤️❤️❤️🐥🔥

PR #2188, opened 2017-03-15
Deprecate global format encoders and move config per app

### LINKED ISSUE #2186: :format_encoders leaks into other Umbrella Phoenix applications
In one of my Umbrella Applications I'm using the following config:
```
config :phoenix, :format_encoders,
  json: Blah.blah
```

My team has recently started working on adding another Phoenix application to the Umbrella. Upon trying to call their Controller which is using `json(conn, ...)` it was erroring trying to access `Blah.blah`.

I'm struggling to figure out how to have these configured individually :(

### Environment

* Elixir version (elixir -v): Elixir 1.4.0
* Phoenix version (mix deps): phoenix 1.3.0-rc.0
* Operating system: Mac OSX Sierra

I've omitted Node as neither Phoenix apps use brunch/node/npm

### Expected behavior
Individually configured format_encoders per Umbrella application.

### Actual behavior
Whichever configuration is loaded last seems to take precedence.

chrismccord:
@josevalim I will need a second set of eyes on this one as it's a bit complex to support since we relied on the global config in a couple places. I'm not thrilled about having to inject `__otp_app__` into both the endpoints and views, but I couldn't come up with a better option. Thoughts?

P.S. What we decide here will also be applied to the global template_engines config, since it has the same issues.

### DIFF
--- a/lib/phoenix/template/eex_engine.ex
+++ b/lib/phoenix/template/eex_engine.ex
@@ -10,7 +10,7 @@ defmodule Phoenix.Template.EExEngine do
   end
 
   defp engine_for(name) do
-    case Phoenix.Template.format_encoder(name) do
+    case Phoenix.Template.format_encoder(Mix.Phoenix.otp_app(), name) do
       Phoenix.Template.HTML ->
         unless Code.ensure_loaded?(Phoenix.HTML.Engine) do
           raise "Could not load Phoenix.HTML.Engine to use with .html.eex templates. " <>
--- a/lib/phoenix/view.ex
+++ b/lib/phoenix/view.ex
@@ -132,6 +132,7 @@ defmodule Phoenix.View do
 
       @doc "The resource name, as an atom, for this view"
       def __resource__, do: @view_resource
+      def __otp_app__, do: unquote(Mix.Phoenix.otp_app())
     end
   end
 
@@ -329,7 +330,7 @@ defmodule Phoenix.View do
   Renders the template and returns iodata.
   """
   def render_to_iodata(module, template, assign) do
-    render(module, template, assign) |> encode(template)
+    render(module, template, assign) |> encode(module, template)
   end
 
   @doc """
@@ -339,8 +340,8 @@ defmodule Phoenix.View do
     render_to_iodata(module, template, assign) |> IO.iodata_to_binary
   end
 
-  defp encode(content, template) do
-    if encoder = Template.format_encoder(template) do
+  defp encode(content, view, template) do
+    if encoder = Template.format_encoder(view.__otp_app__(), template) do
       encoder.encode_to_iodata!(content)
     else
       content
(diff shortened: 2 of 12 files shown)


michalmuskala:
Why not make it an endpoint configuration? This would be far less awkward and more powerful, especially that we try to move towards more runtime-friendly configurations across both phoenix & ecto - this would allow it to be configured at runtime in the init callback or in the endpoint config.

lib/phoenix/template/eex_engine.ex
```diff
@@ -10,7 +10,7 @@ defmodule Phoenix.Template.EExEngine do
   end
 
   defp engine_for(name) do
-    case Phoenix.Template.format_encoder(name) do
+    case Phoenix.Template.format_encoder(Mix.Phoenix.otp_app(), name) do
```
josevalim: We shouldn't access the mix config in our app code. Mix won't be available in prod when using releases.

lib/phoenix/view.ex
```diff
@@ -132,6 +132,7 @@ defmodule Phoenix.View do
 
       @doc "The resource name, as an atom, for this view"
       def __resource__, do: @view_resource
+      def __otp_app__, do: unquote(Mix.Phoenix.otp_app())
```
josevalim: Even though this is compile time, if you want to depend on the otp app, you need to pass it explicitly.

josevalim:
@michalmuskala the view does not require an endpoint to work.

josevalim:
@chrismccord I don't like this change as well and I am not sure if we can fix it without coupling the views to something that it is not coupled to today.

I would probably say this is a won't fix and suggest the user to implement their own `blah_json` in controller or use a separate extension in views.

lib/phoenix/template/eex_engine.ex
chrismccord: I thought it was okay since the engine is only compiled time, but would prefer to avoid it

chrismccord:
I'm going to take a second pass at this and see what falls out. I don't like that our solutions today can be stomped, probably surprisingly, but a dep or another app that configures their format encoders.

josevalim:
> I'm going to take a second pass at this and see what falls out. I don't like that our solutions today can be stomped, probably surprisingly, by a dep or another app that configures their format encoders.


A dep cannot change another dep. This is only the case with umbrella because they are all explicitly loading each others configs. And this is not a problem exclusive to Phoenix, anything in app is going to fail like that. The `:mime` library, `:logger`, etc.

Folks need to understand that all configs are shared and put global configuration in their umbrella config, not in the app.

chrismccord:
That is a very good point comparing to logger and mine. I'd love to avoid the global config in this case, but since it is introducing complexity and coupling you've convinced me to wait and see if it becomes a larger issue. We've had 1 complaint since 1.0, so worth waiting a bit longer 👍

PR #2194, opened 2017-03-17
fix path to router.ex in generators for an umbrella application

glanotte:
When running the `phx.gen.html` or `phx.gen.json` generators, the path to the router.ex file is displayed as:

```
lib/dev_app_web/web/router.ex
```
where it is now located in

```
lib/dev_app_web/router.ex
```

This makes use of the `Mix.Phoenix.web_prefix()` function to point to the correct directory for the `router.ex` file whether in an umbrella project or not.


chrismccord:
❤️❤️❤️🐥🔥

PR #2203, opened 2017-03-21
Improve confusing example

gradenb:


gradenb:
I found this example to be confusing since it specified two guards and the documentation says "The second plug will always run, except for the index action." I spent some time trying to understand the purpose of the first guard. I believe it may have been a copy/paste mistake?

Also, I seem to be unable to get either of these lines to actually work in my dev_app. When I copy either of the lines from the documentation:
````elixir
plug :authenticate, usernames: ["jose", "eric", "sonny"] when action in [:show, :edit]
plug :authenticate, usernames: ["admin"] when not action in [:index]
````

I am told that `when` is not a function:
````
== Compilation error on file lib/dev_app/web/controllers/user_controller.ex ==
** (CompileError) lib/dev_app/web/controllers/user_controller.ex:29: undefined function when/2
    (elixir) expanding macro: Kernel.@/1
    lib/dev_app/web/controllers/user_controller.ex:29: DevApp.Web.UserController (module)
    (phoenix) expanding macro: Phoenix.Controller.Pipeline.plug/2
    lib/dev_app/web/controllers/user_controller.ex:29: DevApp.Web.UserController (module)
    (elixir) lib/kernel/parallel_compiler.ex:117: anonymous fn/4 in Kernel.ParallelCompiler.spawn_compilers/1
````

I am only able to get it to work when I remove the `usernames` argument. I'm not sure if I am doing something wrong or if there is an additional syntax error here.

Gazler:
Great spot. Thanks <3

PR #2204, opened 2017-03-21
Change the order of text when creating new phoenix application

mrunks:
In the console the following text "We are all set! Run your Phoenix application mix phoenix.server"
comes before "Before moving on, configure your database in config/dev.exs and run mix ecto.create"

Since mix ecto.create needs to run first, I changed the order of when they display in the console.

### DIFF
--- a/installer/lib/mix/tasks/phoenix.new.ex
+++ b/installer/lib/mix/tasks/phoenix.new.ex
@@ -248,11 +248,11 @@ defmodule Mix.Tasks.Phoenix.New do
       brunch? = install_brunch(install?)
       extra   = if mix?, do: [], else: ["$ mix deps.get"]
 
-      print_mix_info(path, extra)
-
       if binding[:ecto] do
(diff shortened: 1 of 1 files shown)


chrismccord:
I prefer the current approach because folks often miss the ecto notice, which is required to move forward. Thanks!

MichaelDimmitt:
@chrismccord 
Why not put it in both locations?
1) mix ecto.create, now present in the correct order
2) mix ecto.create is repeated for the folks who often miss the ecto notice.
this is not included in the current pull request.
interested at your thoughts.
<pre>
We are all set! Run your Phoenix application:

    $ cd test_api
    $ mix deps.get
    $ mix ecto.create
    $ mix phoenix.server

You can also run your app inside IEx (Interactive Elixir) as:

    $ iex -S mix phoenix.server

Before moving on, configure your database in config/dev.exs and run:

    $ mix ecto.create
</pre>

PR #2205, opened 2017-03-21
Use Parantheses for render calls

zimt28:
Use parentheses when calling `render` to keep it consistent.

* EEX templates don't use parentheses
* EX(S) files use parentheses for function calls


chrismccord:
Thanks! Since we have churn in these areas right now, I would prefer to avoid stylistic PRs.

PR #2206, opened 2017-03-21
Fix syntax errors in controller plug examples

gradenb:
I'm not sure if there is a more preferred syntax here, but these example didn't work for me unless I wrapped the parameters in brackets. I'm told that `when` is not a function.

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -90,8 +90,8 @@ defmodule Phoenix.Controller do
   `plug/2` in controllers supports guards, allowing a developer to configure
   a plug to only run in some particular action:
 
-      plug :authenticate, usernames: ["jose", "eric", "sonny"] when action in [:show, :edit]
-      plug :authenticate, usernames: ["admin"] when not action in [:index]
+      plug :authenticate, [usernames: ["jose", "eric", "sonny"]] when action in [:show, :edit]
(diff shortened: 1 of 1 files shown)


lib/phoenix/controller.ex
```diff
@@ -90,8 +90,8 @@ defmodule Phoenix.Controller do
   `plug/2` in controllers supports guards, allowing a developer to configure
   a plug to only run in some particular action:
 
-      plug :authenticate, usernames: ["jose", "eric", "sonny"] when action in [:show, :edit]
-      plug :authenticate, usernames: ["admin"] when not action in [:index]
+      plug :authenticate, %{ usernames: ["jose", "eric", "sonny"] } when action in [:show, :edit]
+      plug :authenticate, %{ usernames: ["admin"] } when not action in [:index]
```
chrismccord:
```elixir
plug :authenticate, [usernames: ["jose", "eric", "sonny"]] when action in [:show, :edit]
plug :authenticate, [usernames: ["admin"]] when not action in [:index]
```

chrismccord:
One small comment, then we are golden. Thanks!

gradenb:
@chrismccord Thanks. I fixed up the commit. I think it should be good now.

chrismccord:
❤️❤️❤️🐥🔥

PR #2210, opened 2017-03-22
Don't raise misleading AlreadySentErrors and use Plug.Conn.AlreadySentError alias

gradenb:
Phoenix.Controller aliases Plug.Conn.AlreadySentError, but only took advantage of the alias in one place.

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -436,7 +436,7 @@ defmodule Phoenix.Controller do
     put_private(conn, :phoenix_view, module)
   end
 
-  def put_view(_conn, _module), do: raise AlreadySentError
+  def put_view(%Plug.Conn{}, _module), do: raise AlreadySentError
 
   @doc """
   Stores the view for rendering if one was not stored yet.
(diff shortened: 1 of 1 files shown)


gradenb:
I think `put_view/2`, `put_new_view/2`, and `put_layout_formats/2` have potential to raise misleading errors if a bad type is passed in. They all look similar to this:

````
  @spec put_view(Plug.Conn.t, atom) :: Plug.Conn.t
  def put_view(%Plug.Conn{state: state} = conn, module) when state in @unsent do
    put_private(conn, :phoenix_view, module)
  end

  def put_view(_conn, _module), do: raise AlreadySentError
````

I think if I mistakenly do `put_view(:not_a_conn, :view)` then I will get an `AlreadySentError`.
I'd like to fix this if this actually behaves as I suspect it does. I'm still learning elixir, but it doesn't sem that the @spec actually enforces the types correct?

`put_layout/2` and `put_new_layout/2` do not appear to have this issue since they lack the "catch all" clause.

chrismccord:
❤️❤️❤️🐥🔥

PR #2220, opened 2017-03-28
added nanobox.md as part of another PR for the phoenix_guides

sdomino:
Added Nanobox as an alternative when deploying Phoenix applications (this is part of a PR submitted on phoenix_guides)

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -95,6 +95,7 @@ defmodule Phoenix.Mixfile do
 
      "deployment/deployment.md": [group: "Deployment"],
      "deployment/heroku.md": [group: "Deployment"],
+     "deployment/nanobox.md": [group: "Deployment"],
      "deployment/exrm_releases.md": [group: "Deployment"],
 
      "bonus_guides/upgrading_phoenix.md": [group: "Bonus Guides"],


chrismccord:
Since I'm not too familiar with nanobox, I'm not comfortable making it an official guide. I think it's a great blog post reference though! Thanks!

sdomino:
Thanks Chris, I understand the hesitation completely. If you find yourself with 5 minutes sometime, I'd love to show you Nanobox. We were recently added to the [Elixir Installation guide](http://elixir-lang.org/install.html#nanobox), and thought it would make sense to be part of Phoenix as well.

We love Elixir and Phoenix and are trying to find meaningful ways to contribute. If you feel like there is anything in particular the community could benefit from, we'd love to help!

I can go ahead and close my [other PR](https://github.com/phoenixframework/phoenix_guides/pull/670) if it's easier for you.

ryanwinchester:
@chrismccord I think instructions for using paid 3rd party services should have a separate space and perhaps just include links to the service's own guides instead of making them part of the framework's documentation. That would also remove the burden for things like maintaining docs for 3rd-party services as the services change.

Part of the reason I think this, is that having heroku instructions but not including nanobox instructions is not exactly even-handed (unless maybe heroku is sponsoring phoenix development, but that kind of info should be included?)

Just my $0.02

jeregrine:
The Heroku guides exist because we got a considerable amount of questions from people in IRC/Slack/Forums and at the time Heroku was the only Paas around. 

Its position in the guides is purely to solve a persistent question. Not promote Heroku.

sdomino:
@ryanwinchester thanks for the comment.

Part of the reason we thought to add Nanobox was because we saw Heroku, and thought it might be appropriate. I like your idea and also agree that Heroku is much more ubiquitous than Nanobox ATM, and so I can see why they would be included vs. anyone else.

PR #2233, opened 2017-04-03
feature(header): header in the socket

rsilvestre:


rsilvestre:
@chrismccord : I absolutely need the header send during the handshake in the websocket connection to make some control in the user_socket to allow user to upgrade http connection in web socket. In Node JS there is some possibilities to do it but it's not possible to do it with phoenix. I'm not advanced enough to do it properly and I would like some advice.

for the moment I Hacked the Websocket transport plugin to do It, but It should be possible to do something in a better way. I was thinking about letting the possibility to be able have the socket and the connection available in some way in the same time accessible somewhere in the user_socket or somewhere else. I read somewhere that someone argued that the web socket is a transport layer and I totally agree with it. But the handshake is an upgrade of a http request. So for my point of view, the details of this first connection should still be available during all the life of the websocket connection.

What do you think if we changed something to let to de developper the opportunity to access to the connection during the handshake to do what ever we what ?

Regards

josevalim:
@rsilvestre can't you pass what you need to control as a parameter in the websocket connection instead of a header?

rsilvestre:
@josevalim in fact no. I have a nginx before the service and it adds x-forwarded-for x-forwarded-proto, x-real-ip in the header. I need those data to validate the user and I also need user-agent and others data to make some interesting monitoring to understand how the users use my webservice. 

So it could be interesting to increase the amount of data source during the `user_socket` connection to allow user/assign data to the socket. Like :
```
connect(_conn, _params, _socket)
```

OvermindDL1:
It would also be highly highly useful to get the cookie header in the websocket as well, would be much nicer for authentication instead of hardcoding in a token to the page source some-how...

josevalim:
@OvermindDL1 websockets send the cookie across origins, it is dangerous to rely on cookie for websockets authentication. You can be caught off guard even when using the proper allow origin mechanisms. This is exactly why hiding the connection sometimes it is useful, less likely to run into pitfalls.

chrismccord:
> What do you think if we changed something to let to de developper the opportunity to access to the connection during the handshake to do what ever we what ?

Please see my other responses on this for detailed remarks. We would like to use Saša's work around making transports more extensive so you can accomplish this more easily without as much duplication, but it is lower priority for us over other efforts. In the meantime, you can dup Phoenix.Transports.WebSocket and customize it for your needs. In the future, please to do not send feature requests are discussions to the Github repo. Thanks!

rsilvestre:
Hi @chrismccord, 

My first idea was to dup web socket transport without success. Probably I miss something. I suppose that the Saša's work is exactly what I'm looking for. Where can I found this work and when the integration of the Saša's code is suppose to append, where can I found the prioritised roadmap?

PR #2236, opened 2017-04-04
Improve websocket check_origin documentation to be more explicit

axelson:
Since I spent 20 minutes tracking down why check_origin wasn't working
as expected I'm submitting this as a PR to improve the documentation
(since I never encountered the (very nice!) error defined in
transport.ex)

I'm not sure if the formatting would come out 100% but wanted to at least get the conversation about the content started.

### DIFF
--- a/lib/phoenix/transports/websocket.ex
+++ b/lib/phoenix/transports/websocket.ex
@@ -27,8 +27,8 @@ defmodule Phoenix.Transports.WebSocket do
       check_origin: ["https://example.com",
                      "//another.com:888", "//other.com"]
 
-      Note: it is necessary to set `check_origin: false` to support native
-      mobile apps.
+      Note: To connect from a native app be sure to either have the native app
+      set an origin or allow any origin via `check_origin: false`
 
     * `:code_reloader` - optionally override the default `:code_reloader` value
       from the socket's endpoint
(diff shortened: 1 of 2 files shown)


lib/phoenix/transports/websocket.ex
```diff
@@ -22,7 +22,13 @@ defmodule Phoenix.Transports.WebSocket do
       origin header is present. It defaults to true and, in such cases,
       it will check against the host value in `YourApp.Endpoint.config(:url)[:host]`.
       It may be set to `false` (not recommended) or to a list of explicitly
-      allowed origins
+      allowed origins.
+
+      check_origin: ["https://example.com",
+                     "//another.com:888", "//other.com"]
+
+      Note: it is necessary to set `check_origin: false` to support native
```
OvermindDL1: This is not true, a native app can set that it is connecting 'from' an origin anyway, even when it is not.  Maybe instead word it as:
```markdown
Note: To connect from a native app be sure to either have the native app set an origin or allow any origin via `check_origin: false`
```

axelson: Ah, I was going off of https://github.com/davidstump/SwiftPhoenixClient/issues/21#issuecomment-157050293 which is quite likely outdated by this point.

Updated the wording.

OvermindDL1: Might be a limitation of that specific client, but I know I've connected from Python without issue by setting the origin.

axelson:
I don't see a way to rebuild this on travis-ci but the tests are actually passing (they just temporarily failed due to the nodejs.org outage yesterday).

chrismccord:
<3 <3 <3

axelson:
Thanks! Glad that I was able to help improve the documentation :)  🎊

PR #2244, opened 2017-04-10
Arity typo on list_<schema> context test name?

bryanstearns:
The generated context test is named "list_xxx/1 returns all xxxs", but list_xxx takes no arguments.

/0, right?


chrismccord:
❤️❤️❤️🐥🔥

PR #2245, opened 2017-04-11
Fixed newline in migration template

EugZol:
Fixed code style in migration template: extra newline before `end` removed, and only one newline at the end of file left.

Was before:

```elixir
defmodule YahtzeePhoenix.Repo.Migrations.CreateThing do
  use Ecto.Migration

  def change do
    create table(:things) do
      add :name, :string

      timestamps()
    end

  end
end
# Two newlines at the end of file
```
Is now:

```elixir
defmodule YahtzeePhoenix.Repo.Migrations.CreateThing do
  use Ecto.Migration

  def change do
    create table(:things) do
      add :name, :string

      timestamps()
    end
  end
end # Single newline at then end of file
```

### DIFF
--- a/priv/templates/phoenix.gen.model/migration.exs
+++ b/priv/templates/phoenix.gen.model/migration.exs
@@ -8,9 +8,8 @@ defmodule <%= base %>.Repo.Migrations.Create<%= scoped %> do
 <% end %><%= for {_, i, _, s} <- assocs do %>      add <%= if(String.ends_with?(inspect(i), "_id"), do: inspect(i), else: inspect(i) <> "_id") %>, references(<%= inspect(s) %>, on_delete: :nothing<%= if binary_id do %>, type: :binary_id<% end %>)
 <% end %>
       timestamps()
-    end
-<%= for index <- indexes do %>
-    <%= index %><% end %>
+    end<%= if length(indexes) > 0 do %>
+
+    <%= Enum.join(indexes, "\n    ") %><% end %>
   end
 end
-


priv/templates/phoenix.gen.model/migration.exs
```diff
@@ -8,9 +8,8 @@ defmodule <%= base %>.Repo.Migrations.Create<%= scoped %> do
 <% end %><%= for {_, i, _, s} <- assocs do %>      add <%= if(String.ends_with?(inspect(i), "_id"), do: inspect(i), else: inspect(i) <> "_id") %>, references(<%= inspect(s) %>, on_delete: :nothing<%= if binary_id do %>, type: :binary_id<% end %>)
 <% end %>
       timestamps()
-    end
-<%= for index <- indexes do %>
-    <%= index %><% end %>
+    end<%= if length(indexes) > 0 do %>
```
chrismccord: if we move the `for` comprehension up to the same line as `end`, we can avoid the if check and enum join dance

chrismccord: ping :)

EugZol: @chrismccord sorry, missed the notification :) If we move `for` up to the previous line, then the result is:

```
  def change do
    create table(:somethings) do
      add :name, :string

      timestamps()
    end
    create unique_index(:somethings, [:name])
  end
```

(Blank line between `create table(...)` section and indexes is missing.`)

chrismccord:
❤️❤️❤️🐥🔥

Issue #2240, opened 2017-04-07
mix phx.gen.context issues with nesting


klintmane:
### Environment

* Elixir version (elixir -v): 1.4
* Phoenix version (mix deps): 1.3.0-rc
* NodeJS version (node -v):
* NPM version (npm -v):
* Operating system: Ubuntu 16.10

### Reproducing
![screenshot at 2017-04-07 18-56-05](https://cloud.githubusercontent.com/assets/9338536/24810819/f6cf9388-1bc3-11e7-821f-98bae7954223.png)

### Expected behavior
Running `mix phx.gen.context Site.Blog Post title:string content:string` should create the following directory structure:
![screenshot at 2017-04-07 18-37-45](https://cloud.githubusercontent.com/assets/9338536/24810093/5b4d8232-1bc1-11e7-9aa2-7051d1f88c29.png)
Schema naming should adhere to `snake_case`, like below:
![screenshot at 2017-04-07 18-38-37](https://cloud.githubusercontent.com/assets/9338536/24810150/996c89be-1bc1-11e7-8d48-193655ba4eaa.png)

### Actual behavior
The generated directory structure is mirrored like below:
![screenshot at 2017-04-07 18-34-30](https://cloud.githubusercontent.com/assets/9338536/24810592/161c5452-1bc3-11e7-9610-08d122fe0f9b.png)
The schema name is:
![screenshot at 2017-04-07 18-51-33](https://cloud.githubusercontent.com/assets/9338536/24810635/3e8e057a-1bc3-11e7-9bb7-80f0a761bfba.png)
Related to the above, generated migrations create invalid atoms for table names
![screenshot at 2017-04-07 18-53-52](https://cloud.githubusercontent.com/assets/9338536/24810741/a9bb3610-1bc3-11e7-8433-18d0006173aa.png)

lalo2302:
The generator in the image differs from the expected behavior running command

lalo2302:
This is because the table name is generated via `Macro.underscore` which according to the docs it converts `"Site.Blog"` to `"site/blog"`.

A short solution could be to trim all `.` characters before underscoring it, or a longer one to create/use another naming solution as the docs suggests:

> This function was designed to underscore language identifiers/tokens, that’s why it belongs to the Macro module. Do not use it as a general mechanism for underscoring strings as it does not support Unicode or characters that are not valid in Elixir identifiers.

[...]

PR #2250, opened 2017-04-14
Fix nested context module name (#2240)

lalo2302:
#2240 

1. To make the table name I made a new method (`Context.underscore/1`) that underscores the context name without `.`, to prevent the `\` character as a result of `Phoenix.Naming.underscore/1`.

```elixir
def underscore(context_name) do
  context_name
  |> String.replace(".","")
  |> Phoenix.Naming.underscore
end
```

2. The basename of the context is going to be the last word of a list split by `.` in case the module name is a nested context.
3. The directories will use `Phoenix.Naming.underscore/1` like before to take advantage of the `\` character.

### DIFF
--- a/lib/mix/phoenix/context.ex
+++ b/lib/mix/phoenix/context.ex
@@ -25,12 +25,12 @@ defmodule Mix.Phoenix.Context do
     ctx_app   = opts[:context_app] || Mix.Phoenix.context_app()
     base      = Module.concat([Mix.Phoenix.context_base(ctx_app)])
     module    = Module.concat(base, context_name)
-    alias     = module |> Module.split() |> tl() |> Module.concat()
-    basename  = Phoenix.Naming.underscore(context_name)
-    dir       = Mix.Phoenix.context_lib_path(ctx_app, basename)
+    alias     = module |> Module.split() |> List.last() |> Module.concat(nil)
+    basename  = context_name |> String.split(".") |> List.last |> Phoenix.Naming.underscore
+    dir       = Mix.Phoenix.context_lib_path(ctx_app, Phoenix.Naming.underscore(context_name))
     test_dir  = Mix.Phoenix.context_app_path(ctx_app, "test")
     file      = Path.join([dir, basename <> ".ex"])
-    test_file = Path.join([test_dir, basename <> "_test.exs"])
+    test_file = Path.join([test_dir, Phoenix.Naming.underscore(context_name) <> "_test.exs"])
     generate? = Keyword.get(opts, :context, true)
 
     %Context{
--- a/lib/mix/tasks/phx.gen.context.ex
+++ b/lib/mix/tasks/phx.gen.context.ex
@@ -99,7 +99,8 @@ defmodule Mix.Tasks.Phx.Gen.Context do
     {opts, parsed, _} = parse_opts(args)
     [context_name, schema_name, plural | schema_args] = validate_args!(parsed)
 
-    table = Keyword.get(opts, :table, Phoenix.Naming.underscore(context_name) <> "_#{plural}")
+    table_name = context_name |> Phoenix.Naming.underscore |> String.replace("/", "_")
+    table = Keyword.get(opts, :table, "#{table_name}_#{plural}")
     schema_module = inspect(Module.concat(context_name, schema_name))
 
     schema = Gen.Schema.build([schema_module, plural | schema_args] ++ ["--table", table], opts, __MODULE__)
--- a/test/mix/tasks/phx.gen.context_test.exs
+++ b/test/mix/tasks/phx.gen.context_test.exs
@@ -41,6 +41,33 @@ defmodule Mix.Tasks.Phx.Gen.ContextTest do
     end
   end
 
+  test "new nested context", config do
+    in_tmp_project config.test, fn ->
+      schema = Schema.new("Site.Blog.Post", "posts", [], [])
+      context = Context.new("Site.Blog", schema, [])
+
+      assert %Context{
+        alias: Blog,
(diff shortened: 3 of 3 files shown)


lib/mix/tasks/phx.gen.context.ex
```diff
@@ -99,7 +99,7 @@ defmodule Mix.Tasks.Phx.Gen.Context do
     {opts, parsed, _} = parse_opts(args)
     [context_name, schema_name, plural | schema_args] = validate_args!(parsed)
 
-    table = Keyword.get(opts, :table, Phoenix.Naming.underscore(context_name) <> "_#{plural}")
+    table = Keyword.get(opts, :table, (Phoenix.Naming.underscore(context_name) |> String.replace("/","_")) <> "_#{plural}")
```
lalo2302: I did a cleaner solution. I hope this will do it. Thanks for the suggestion José.

kelvinst: I would just change `Phoenix.Naming.underscore(context_name)` to `context_name |> Phoenix.Naming.underscore`. Prefer "pure" values over function calls when starting the pipe chain, [here is the link](https://github.com/rrrene/elixir-style-guide#pipe-chains) the the styleguide definition.

lalo2302:
I have 3 options. Which one looks better?

1. This one I think is too long
```elixir
table = Keyword.get(opts, :table, (context_name |> Phoenix.Naming.underscore |> String.replace("/","_")) <> "_#{plural}")
```

2. Looks better, but something is still missing
```elixir
table = Keyword.get(opts, :table, (context_name
                                  |> Phoenix.Naming.underscore
                                  |> String.replace("/","_")) <> "_#{plural}")
```

3. Looks very nice, but I don't know if it is bloated and confusing by being the 3rd argument on `Keyword.get`
```elixir
table = Keyword.get(opts, :table, (context_name
                                  |> Phoenix.Naming.underscore
                                  |> String.replace("/","_")) 
                                  <> "_#{plural}")
```

lib/mix/tasks/phx.gen.context.ex
```diff
@@ -99,7 +99,10 @@ defmodule Mix.Tasks.Phx.Gen.Context do
     {opts, parsed, _} = parse_opts(args)
     [context_name, schema_name, plural | schema_args] = validate_args!(parsed)
 
-    table = Keyword.get(opts, :table, Phoenix.Naming.underscore(context_name) <> "_#{plural}")
+    table = Keyword.get(opts, :table, (context_name
```
lalo2302: It looks better this way.

lib/mix/phoenix/context.ex
```diff
@@ -25,12 +25,12 @@ defmodule Mix.Phoenix.Context do
     ctx_app   = opts[:context_app] || Mix.Phoenix.context_app()
     base      = Module.concat([Mix.Phoenix.context_base(ctx_app)])
     module    = Module.concat(base, context_name)
-    alias     = module |> Module.split() |> tl() |> Module.concat()
-    basename  = Phoenix.Naming.underscore(context_name)
-    dir       = Mix.Phoenix.context_lib_path(ctx_app, basename)
+    alias     = module |> Module.split() |> List.last() |> Module.concat(nil)
```
josevalim: Are those changes necessary?

lib/mix/tasks/phx.gen.context.ex
josevalim: Instead of inline such a huge expression, please move it a variable. Something like:

    table_prefix =  context_name |> Phoenix.Naming.underscore |> String.replace("/", "_")
    table = Keyword.get(opts, :table, "#{table_prefix}_#{plural}")

lib/mix/phoenix/context.ex
lalo2302: Without them, the alias for a nested context like `Site.Blog` would be `Site.Blog`, and for what I've read it should be only `Blog`, but tell me if I'm wrong 😊. 

I took as an example the code on `lib/mix/phoenix/schema.ex`
```elixir
%Schema{
...
alias: module |> Module.split() |> List.last() |> Module.concat(nil),
...
```

lib/mix/phoenix/context.ex
```diff
@@ -25,12 +25,12 @@ defmodule Mix.Phoenix.Context do
...
+    basename  = context_name |> String.split(".") |> List.last |> Phoenix.Naming.underscore
+    dir       = Mix.Phoenix.context_lib_path(ctx_app, Phoenix.Naming.underscore(context_name))
```
lalo2302: In the case of a context being `Site.Blog`, after underscoring it turns into `site/blog` which gives us the desired path

lib/mix/tasks/phx.gen.context.ex
lalo2302: Done. But I don't know why the Travis CI build failed, maybe it is expecting this:
```elixir
table_name = 
    context_name 
    |> Phoenix.Naming.underscore 
    |> String.replace("/", "_")
```
instead of:
```elixir
table_name = context_name |> Phoenix.Naming.underscore |> String.replace("/", "_")
```

At the job log it crashes within the integration endpoint test
```
test adapters starts on configured port and serves requests and stops for dev (Phoenix.Integration.EndpointTest)
     test/phoenix/integration/endpoint_test.exs:136
     Assertion with =~ failed
```
I'm sorry for the trouble, it's my first contribution.

chrismccord:
Merged in https://github.com/phoenixframework/phoenix/commit/20e2826587b25227e107a83fbb56ce7f4d0c731f  ❤️❤️❤️🐥🔥

PR #2252, opened 2017-04-17
fix doc of errors_on function

chenxsan:
This is code of the generated `Accounts.create_user`:

```
  @doc """
  Creates a user.

  ## Examples

      iex> create_user(%{field: value})
      {:ok, %User{}}

      iex> create_user(%{field: bad_value})
      {:error, %Ecto.Changeset{}}

  """
  def create_user(attrs \\ %{}) do
    %User{}
    |> user_changeset(attrs)
    |> Repo.insert()
  end
```
It will return `{:error, changeset}` when errors happen, not `changeset`.


josevalim:

Issue #2253, opened 2017-04-17
Generated brunch-config.js has incorrect links to brunch docs


smpallen99:
The follow links do not work in my project's generated brunch-config.js file. I used `mix phx.new` to generate the project.

```
https://github.com/brunch/brunch/blob/master/docs/config.md#files
https://github.com/brunch/brunch/tree/master/docs#concatenation
``` 

I think these links should be updated to point to `brunch.io`. For example `http://brunch.io/docs/config#-files`

chrismccord:
Good catch! Can you send a PR? Thanks!

smpallen99:
Yes, I can send a PR Chris

PR #2256, opened 2017-04-17
Update generated brunch-config.js brunch docs references #2253

smpallen99:
Updated the files link and removed the link to concatenation to be consistent with the version in static


chrismccord:
<3 <3 <3

Issue #2257, opened 2017-04-18
mix phx.gen.schema error about array type is unclear


kuon:
When creating a schema with `mix.phx.gen.schema` the `array` type support is unclear.

For example:

```
$ mix phx.gen.schema Awesome.Test awesome_tests foo:jsonb
** (Mix) Unknown type `:jsonb` given to generator. The supported types are: array, binary, boolean, date, decimal, float, integer, map, naive_datetime, references, string, text, time, utc_datetime, uuid

$ mix phx.gen.schema Awesome.Test awesome_tests foo:array
** (Mix) Unknown type `array` given to generator
```

The first error (which is expected) lists `array` as supported type, but when `array` is used as a type, it is not accepted. The first error message should inform the user about the `array:element_type` syntax.

Tested on phoenix 1.3 master installed from 3e46a129187aa6b5d0734fdae75594df496bea68

chrismccord:
You need to pass the type of the array elements, for example ` mix phx.gen.schema Awesome.Test awesome_tests foo:array:string`. I agree we can improve the error in this case. Thanks for the report!

kuon:
Yes I found the actual syntax in the doc. This issue is just about improving the error message, cheers!

jeroenvisser101:
@chrismccord @gavlak this is fixed now right?

PR #2259, opened 2017-04-20
Improve Mix.Phoenix.Schema array error message

gavlak:
This PR solves #2257 

When you generate schema with key of type array, but without the type of the array, the following error message is now raised:

```
$ mix phx.gen.schema Post posts settings:array
** (Mix) Phoenix generators expect the type of the array to be given.
For example:

    mix phx.gen.schema Post posts settings:array:string
```


almielczarek:
Also, I may be missing something, but I do not think that is the place to add the check for an array. The way it looks to me is that every time anyone passes the type :array to the generator it will match on your new function clause and raise.

I don't mean to be annoying, but I started working on this and didn't have an internet connection to see that you did too, and I found that doing the check by `validate_attr` caused mix to raise only when it should.

gavlak:
I messed that up and noticed I'm quite a few commits behind. I'll fix that.

As for the build failing. When I tested it on my machine it passed just fine, but `Phoenix.Integration.WebSocketTest` seems to fail sometimes. That's the only failed test on Travis too.

I also saw your other comment and I think you're right. Pattern matching `validate_attr!` looks much nicer, clearer and makes more sense than checking the value in `schema_type`. I'll change that too.

almielczarek:
I deleted that comment because I found the function pipeline was raising before hitting `validate_attrs!` causing the validation to never be run. So I think that if you're not changing anything else here you have to put the check where you did to make sure it gets run.

And once again, I could be missing something, but I still think it should be checked by `validate_attrs!`, but then we need to either change the order that functions are called here or stop raising by `schema_type` and instead calling `validate_attrs` where it would normally raise so the array check happens first. Either way if you pass an unsupported type it'll raise saying it is an unknown type.

If that wasn't clear I'll try to rephrase it.

chrismccord:
This looks good. Can you add a test case? Thanks!

gavlak:
Done.

chrismccord:
❤️❤️❤️🐥🔥

PR #2261, opened 2017-04-21
Fix the typo in phx.router (should be phx.routes)

hubertlepicki:
This fixes the warning message that says:

"mix phoenix.router is deprecated. Use phx.router instead."

The mix task is called `phx.routes`.

### DIFF
--- a/lib/mix/tasks/phoenix.routes.ex
+++ b/lib/mix/tasks/phoenix.routes.ex
@@ -23,7 +23,7 @@ defmodule Mix.Tasks.Phoenix.Routes do
   """
 
   def run(args, base \\ Mix.Phoenix.base()) do
-    IO.puts :stderr, "mix phoenix.router is deprecated. Use phx.router instead."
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/phoenix.routes.ex
```diff
@@ -23,7 +23,7 @@ defmodule Mix.Tasks.Phoenix.Routes do
   """
 
   def run(args, base \\ Mix.Phoenix.base()) do
-    IO.puts :stderr, "mix phoenix.router is deprecated. Use phx.router instead."
+    IO.puts :stderr, "mix phoenix.router is deprecated. Use phx.routes instead."
```
wojtekmach: `mix phoenix.routes is deprecated. Use phx.routes instead.` :D

hubertlepicki:
whops. will fix 2nd typo when around keyboard later

lib/mix/tasks/phoenix.routes.ex
hubertlepicki: I have updated the other typo

josevalim:

PR #2262, opened 2017-04-22
Adding ESLint to maintain consistent styling and prevent bugs in JS files

cvoege:
# ESLint
This pull request adds an [ESLint](http://eslint.org) ruleset to better manage the code quality and style choices of the Javascript that comes bundled with Phoenix. ESLint is a linter for Javascript that does several things for project maintainers. It ensures a consistent styles is maintained across all files by, for example, displaying an error on improper indentation of an if block. It also prevents bugs by displaying errors when a variable is used without being declared, preventing simple misspellings from making it into the final build.

Rather than have to check over pull requests by hand to make sure they conform to the unoffical styles set by precendent, ESlint will automatically all JS files during the build step. If at some point in the future the project maintainers decided to change a style rule, like for example requiring semi-colons, one change to the ESLint file would make it clear exactly what changes need to be made to conform to the new styles. Even better, ESLint can auto fix them so certain style changes can be done in a matter of seconds across all JS files in the entire repo.

ESLint lets you specify the ruleset, which is done in the .eslintrc.js file. This ruleset extends the [recommended](http://eslint.org/docs/rules/) linting rules, with some overrides to fit the existing style choices.

## Inconsistent styles: 

The phoenix.js file was mostly consistent in how thing were styled, with a few exceptions. The finding of these exceptions shows where eslint shines, as manual code review didn't pick these up. In the future, eslint will prevent all of these inconsistencies should this pull be merged.

I went through the trouble of finding all the correct linting rules for based my reading of the existing phoenix.js file which was mostly consistent to one set of styles, with a few exceptions. In these exceptions, I chose the rule setting that I thought fit the file best. Here are the rules that I chose in those situations where the phoenix.js file was not wholly consistent. I'd like to hear if there are any disagreements on these choices.

* No semi-colons 
* Only use block in arrow functions when it is needed
* Don't require parenthesis around single arrow function paramater
* Don't use var's outside their block
* "One true brace style" with single line blocks allowed
* No comma dangle
* Indentation of 2 spaces, including switch statements
* Required spacing around keywords such as if/else/try/catch/
* No empty functions. If you want to create a noop function, you need to put a comment in it so that it's clear to future developers. The minifier remove the comment from production.
* No unnecessary parenthesis
* No reassigning of parameters.
* Don't throw literals as errors. Dev tools are usually better at giving feedback for errors when they are thrown in the format 'throw new Error("Something something something")'
* No using variables before they are declared. This includes classes like Timer. Ideally the solution to this is to move classes like "Timer" into their own files, which would make for much cleaner, more modular code later.
* No more "var". var is used 4 times, 2 of them can be replaced with let with no trouble, and the last one is an anti pattern using a variable outside its scope.
* Use double quotes for all strings, unless the double quote character is present in the string.
* Always have a space before blocks. This one was very inconsistent across the file, something like 50/50 between space or no space. I think space looks a lot prettier and easier to reason about, so I selected that, but it can be changed and auto-fixed by eslint at any time.
* Never use spaces in right after opening parenthesis and before closing parenthesis, e.g. (1+2) does not raise an error but ( 1 + 2 ) does. Spaces we're never used except for around arrow function parameters in map() and filter() calls. So I went with no opening and closing space, since other function parameter declarations and uses of parenthesis did not have these spaces.

For most of those rules the choice was pretty obvious, as the entire file would follow that rule except for one issue. For example, there was a single semi-colon, but it was not a necessary semi-colon. It can be safely deleted. ESLint lets you know when semi-colons can be safely deleted, and when they are required to make your code work. I prefer to always use semi-colons, but ESLint lets you write code without them and not fear the few cases where you're going to end up needing one.

## What we need to do now

At this point this pull request does **not** include changes to the actual Javascript files to make them fully conform to the rules I set out (though they do for the most part already follow the rules). All this pull request does right now is create the ruleset with the necessary dependencies to run it. If the maintainers of this project approve of the rules that I've derived, I'll commit once again to this branch and finish the job by fixing all the linting errors (most of which can be autofixed by ESLint) and adding the "npm run lint" call to the Travis build.

The rule that causes the most errors is the "space-before-blocks" rule, which was very inconsistent across the file. Basically the choice is between "class Klass {" and "class Klass{". I chose the former, because I think it's more readable, but the rule could be changed if the maintainers disagree.

### Future goals:

For these kinds of rules I didn't have enough info to set rules for them yet, but we should commit to setting them up in the future because they are great for ensuring clean code.

* Maximum line length (max-len)
* Maximum number of lines per file (max-lines)
* Prefer const over let (prefer-const). const is preferable since a developer can guarantee that their variable won't get changed by another developer in a later commit.
* No magic numbers. This library uses lots of magic numbers for things like default timeouts, which is somewhat acceptable, but using constants in CAPTIAL_CASE can be more readable.

I've been wanting to contribute to Phoenix for quite a while, but haven't had the time to delve into it enough to feel comfortable making changes. I do however know JS and the JS ecosystem very well, so I thought this would be a good starter to get me into the community :)


cvoege:
I have another branch in my repo [eslint_fixes](https://github.com/cvoege/phoenix/tree/eslint_fixes) which contains the all the linting errors with fixes applied and with `npm run lint` added to the travis build. If this ruleset is agreeable by the project managers here, I can merge that into this same pull request.

Looking forward to hearing what you guys think about this!

-Colton

chrismccord:
Hey Colton,

Thank you for the very thorough explanation of this PR and taking the time to help make phoenix better! At this time I'm not ready to adopt a particular js linter, but I will keep it in mind for the future. Thanks again!

cvoege:
Hey @chrismccord,

Can I ask why you're not ready to adopt a JS linter? They don't have any lock in and at any given point you could remove it, switch to a different one, or run multiple at the same time. All this does is enforce the styles that the JS file already follows. 

ESLint is also a very mature library with use at Google, AirBnB, and several other large companies, and is widely regarded as the most advanced and useful JS linter. Using it will save a lot of time in the future when reviewing pull requests to the Javascript in this codebase, as you won't have to make sure they indent properly or forget to declare variables, because ESLint will do that for you.

Thanks for the consideration!

ohmfox:
I really like this idea. Linter's make it a lot easier to manage JS code especially in a project this large. I'd love to know why the project doesn't want to use a linter.

lperiodbose:
Phoenix doesn't specify a default _Elixir_ linter. It's a reasonable choice not to proscribe a JavaScript one.

cvoege:
Hey @lperiodbose, thanks for the reply. I feel as though Elixir, as a language, is much less in need of a linter since it's designed to be beautiful. Javascript on the other hand is much more unwieldy, it was designed with several philosophies in mind and they aren't exactly merged together beautifully. I say this even though I love the JS language, and I think JS shines a lot brighter when you add a linter which restricts the language styling to be exactly what you want it to be.

That was my goal with this pull request, to give you guys a ruleset to enforce the styles that you already use. It's a win for everyone in my opinion. The major developers on this repo never have to check a JS pull request for styling anymore, just features and tests. It also makes it more approachable to get started making pull requests on the library because the styling will be done for you to stay in line with the repos existing ideals. It doesn't add any weight when someone installs Phoenix, but it increases confidence that the JS code is bug free.

PR #2268, opened 2017-04-27
Fix function argument inconsistencies in docs for scope

joshprice:
Looks like the docs didn't match the macro definitions.

I also added examples to demonstrate the usage of various scope functions to try and make this clearer.


chrismccord:
❤️❤️❤️🐥🔥

Issue #2274, opened 2017-05-05
Shouldn't `error_tag` show all errors for a given field?


nathanl:
### Expected behavior

If multiple validations have called `Ecto.Changeset.add_error(changeset, :name)`, `error_tag(form, :name)` should show all of them.

### Actual behavior

It only shows the first one.

## Proposed fix

Change [this code](https://github.com/phoenixframework/phoenix/blob/4420a23586712239b6cea2f0e20750f357a35b1f/installer/templates/phx_web/views/error_helpers.ex#L11-L15) from:

```elixir
  def error_tag(form, field) do
    if error = form.errors[field] do
      content_tag :span, translate_error(error), class: "help-block"
    end
  end
```

to something like:

```elixir
  def error_tag(form, field) do
    errors = Keyword.get_values(form.errors, field)
    if Enum.any?(errors) do
      Enum.map(errors, fn (error) ->
        content_tag :span, translate_error(error), class: "help-block"
      end)
    end
  end
```

## Visual difference

Here's a form with silly example errors. I've added `<%= inspect f.errors %>` at the top of the form to make clear what's happening.

With the current code, `<%= error_tag f, :name %>` shows only the first error on `:name`:
![before](https://cloud.githubusercontent.com/assets/338814/25753794/e2632c98-318a-11e7-99ce-6feb1bd73565.png)

With my suggested code, `<%= error_tag f, :name %>` shows all errors on `:name`:
![after](https://cloud.githubusercontent.com/assets/338814/25753803/e817c770-318a-11e7-80ee-f25dd33e7919.png)

Maybe it would be better to concatenate these errors instead of giving each its own tag, but only showing one definitely seems wrong to me.

What would you think of a change like this?

josevalim:
Yes, this would be a good call. I don't think you need the Enum.any check.
You can return an empty list just fine I think as it is an IO list?
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

nathanl:
Cool, thanks! Will make a PR.

PR #2275, opened 2017-05-05
`error_tag` shows all errors on field, not just first

nathanl:
Previously, if a field had multiple errors, only the first would be
displayed.

With this change, each error on the field is displayed in its own tag.

Fixes https://github.com/phoenixframework/phoenix/issues/2274


nathanl:
The build error on this appears to be nondeterministic and unrelated to the PR.

josevalim:

philippneugebauer:
@josevalim In which part of Phoenix is this fix included? I'm using Phoenix 1.3 but I can see all errors in the changeset, but just one is shown in the view

josevalim:
Look at the error_helpers.ex file in your application and change it accordingly. This was a change to the generator, not in Phoenix itself.

PR #2284, opened 2017-05-08
Fixing example for presence fetch/2 callback

schrockwell:
Need to convert presences' string keys to integers for user lookup


chrismccord:
Thanks for the PR! This is now outdated after Gazler's update, but much appreciated!

PR #2290, opened 2017-05-11
Update presence.ex description of an example

seivan:
Unless I misunderstood, it was a bit confusing as the description does not follow the code example in order.


chrismccord:
<3 <3 <3

PR #2291, opened 2017-05-12
use Mix.Dep.Umbrella.cached/0 instead of loaded/0

Gazler:
This fixes the issues mentioned in [1] where the code reloaded would be
in the wrong directory. `cached/0` works but is Elixir 1.4 exclusive, so
the deps are retrieved conditionally depending on if the function exist.

[1] https://github.com/elixir-lang/elixir/issues/6104


josevalim:

PR #2299, opened 2017-05-16
Fix location for files in presence and channel generators

Gazler:
When in an umbrella and the `context_app` config is set, the files
should go in directories without the web prefix.

In the channels test, the test_prefix is specified directly so that the
files ends up in:

`/test/channels/foo_test.exs` instead of
`test/phoenix_web/channels/foo_test.exs`


chrismccord:
😍😍😍

PR #2302, opened 2017-05-16
Fix phx.gen.html templates and shell instructions with --web

britto:
Currently, when the `--web` option is used, the templates generated are inconsistent with the controller code regarding namespaced path helpers. While the controller correctly follows the convention of `<namespace>_<resource>_path(conn, ...)`, the templates are always generated with `<resource>_path(conn, ...)`, regardless of `--web`.

As an example, when we run the generator as follows:

```
mix phx.gen.html Blog Post posts title:string --web Admin
```

We expect all path helpers to be generated as `admin_post_path(@conn)`. The same goes for the router scope, to be compatible it should include an `as` directive – like `scope "/admin", App.Web.Admin, as: :admin` – otherwise both controller and templates will not compile.

Now the templates follow the same convention, as well as the shell instructions. So, when the instructions are followed, the generated code compiles correctly.


chrismccord:
❤️❤️❤️🐥🔥

PR #2303, opened 2017-05-16
Remove all functions from mix tasks from docs

michalmuskala:
The functions should not be called manually (only through `Mix.Task.run`).
This also makes the task docs much more concise and less cluttered.

### DIFF
--- a/lib/mix/tasks/phoenix.digest.ex
+++ b/lib/mix/tasks/phoenix.digest.ex
@@ -32,6 +32,7 @@ defmodule Mix.Tasks.Phoenix.Digest do
     * cache_manifest.json
   """
 
+  @doc false
   def run(args) do
     IO.puts :stderr, "mix phoenix.digest is deprecated. Use phx.digest instead."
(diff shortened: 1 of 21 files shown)


chrismccord:
😍😍😍

josevalim:
I am -1 on this change. Run is part of the contract, I don't think we
should hide it. If it is important to hide because of ex_doc changes, then
we should do it in Mix itself.
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

chrismccord:
No one will be calling these directly, and all relevant docs are on the moduledoc, so I don't see the value in exposing them. Really it doesn't matter either way, so I'll let @josevalim revert if he pleases.

michalmuskala:
It's only partially related to the PR I opened in ex_docs. We could automatically hide the `run` function there, but I'm not 100% sure it's a good idea for all the cases. In the case of phoenix tasks, I'm convinced that exposing the functions wouldn't bring any benefit.

PR #2310, opened 2017-05-17
Remove plural name arg from phx.gen.embedded

mbuhot:
I was surprised reading the docs for `mix phx.gen.embedded` saying that it required a plural name which is used for the 'table name'.

Embedded schemas aren't linked to a table, and may be used to validate data coming from other sources (request params, message queues, etc.)

This PR removes the required plural name argument from `phx.gen.embedded` and de-couples `phx.gen.embedded` from `phx.gen.schema`.

So the interface is just:

```
mix phx.gen.embedded Blog.Post title:string views:integer
```


chrismccord:
😍😍😍

mbuhot:
Thanks @chrismccord @josevalim !

PR #2311, opened 2017-05-17
Fix spacing in generated embedded schemas

britto:
The current template generates inlined fields, which is invalid Elixir. Now we have one line per field.

```
$ mix phx.gen.embedded Blog.Postr title:string body:text views:integer
```

Before:

```elixir
  embedded_schema do
    field :body, :string    field :title, :string    field :views, :integer
  end
```

After:

```elixir
  embedded_schema do
    field :body, :string
    field :title, :string
    field :views, :integer
  end
```


josevalim:

PR #2312, opened 2017-05-19
Fix outdated links: from elixir-lang.org to hexdocs.pm

StanBright:
There are a number of links pointing to elixir-lang.org that
are redirected to hexdocs.pm currently. Simply replace the old ones
to the new targets.

e.g.
http://elixir-lang.org/docs/stable/elixir/Application.html
=>
https://hexdocs.pm/elixir/Application.html

### DIFF
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -72,7 +72,7 @@ defmodule Phoenix.Router do
 
   The URL generated in the named URL helpers is based on the configuration for
   `:url`, `:http` and `:https`. However, if for some reason you need to manually
-  control the URL generation, the url helpers also allow you to pass in a [`URI`](https://hexdocs.pm/elixir/URI.html)
+  control the URL generation, the url helpers also allow you to pass in a `URI`
   struct:
 
       uri = %URI{scheme: "https", host: "other.example.com"}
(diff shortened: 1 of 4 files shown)


lib/phoenix/router.ex
```diff
@@ -72,7 +72,7 @@ defmodule Phoenix.Router do
 
   The URL generated in the named URL helpers is based on the configuration for
   `:url`, `:http` and `:https`. However, if for some reason you need to manually
-  control the URL generation, the url helpers also allow you to pass in a [`URI`](http://elixir-lang.org/docs/stable/elixir/URI.html)
+  control the URL generation, the url helpers also allow you to pass in a [`URI`](https://hexdocs.pm/elixir/URI.html)
```
josevalim: Here we can just use `URI`.

ExDoc will take care of generating the proper URL when docs are built. :)

josevalim:
Thank you, I have added just one small comment. :)

lib/phoenix/router.ex
StanBright: I'm not sure I understand :)

Do you mean replacing
```[`URI`](https://hexdocs.pm/elixir/URI.html)```
with
``` `URI` ```
?

josevalim: Yes!

josevalim:

Issue #2322, opened 2017-05-26
Generating schemas with phx.gen.{context, json, html} with references, is unclear


MartinElvar:
Hello :-) 

Excuse the bit unclear title, could't come up with anything more describing, feel free to correct it. :)

### Environment

* Elixir version (elixir -v): 1.4.2
* Phoenix version (mix deps): 1.3.0-rc.2
* NodeJS version (node -v): v7.4.0
* NPM version (npm -v): 4.0.5
* Operating system: MacOS

### To the case

Today i started a new project with Phoenix 1.3 RC 2, and i started generating some schemas, i used the mix phx.gen.context to get myself a Tenant schema.

```mix phx.gen.context Accounts Tenant tenants name:string logo:string```

All good!

I migrated, and got to my user schema straight away.

```mix phx.gen.context Accounts User users name:string email:string tenant:references```

I could't remember how to state my reference, phoenix helped me out with a fantastic error message 👍 .

```
** (Mix) Phoenix generators expect the table to be given to tenant:references.
For example:

    mix phx.gen.schema Comment comments body:text post_id:references:posts
```

I corrected my command as a result.

```mix phx.gen.context Accounts User users name:string email:string tenant_id:references:tenants password_digest:string```

Great success, but then i ran mix ecto.migrate, and got.

```
13:50:27.956 [info]  == Running Pharma.Repo.Migrations.CreatePharma.Accounts.User.change/0 forward

13:50:27.956 [info]  create table accounts_users
** (Postgrex.Error) ERROR 42P01 (undefined_table): relation "tenants" does not exist
```

Hmm weird i thought. But i quickly realised that in phoenix 1.3, the context is prefixed to the table names, so i quickly altered my migration and my schema file.

```Elixir
  schema "accounts_users" do
    field :email, :string
    field :name, :string
    field :password_digest, :string
    field :picture, :string
    - field :tenant_id, :id
    + field :accounts_tenant_id, :id

    timestamps()
  end
```

```Elixir
  use Ecto.Migration

  def change do
    create table(:accounts_users) do
      add :name, :string
      add :email, :string
      add :password_digest, :string
      add :picture, :string
      - add :tenant_id, references(:tenants, on_delete: :nothing)
      + add :accounts_tenant_id, references(:accounts_tenants, on_delete: :nothing)
      
      timestamps()
    end

    - create index(:users, [:tenant_id])
    + create index(:accounts_users, [:accounts_tenant_id])
  end
```

[...]

PR #2323, opened 2017-05-26
Improve error messages for references, with context consideration

MartinElvar:
As suggested in #2322 .


josevalim:
Hi @MartinElvar, we are currently discussing not namespacing the table, as this is causing confusion in many other areas. Thank you!

josevalim:
Hi @MartinElvar, I have decided to keep things simpler and not namespace the table names. Thank you for reporting an issue though, it was one of the many examples that led us into this simpler direction.

MartinElvar:
Fantastic @josevalim, great you guys found a even better solution. 😃

PR #2328, opened 2017-05-29
Update comment for Socket's id/1 function

Trevoke:
I *think* this was the intent behind the comments?


chrismccord:
Thanks! In this case you don't have a handle to all the users `%Socket{}`'s, and rather need to build the topic by hand per the existing docs. Thanks for the PR

Trevoke:
Hmm... So isn't that a situation where the `id/1` function might come in handy, assuming you don't have a reference to the socket (of course this means `id/1` cannot take a socket as argument) but need to create the topic, so that you can keep the knowledge of "what the user's topic is" in that single place?

PR #2334, opened 2017-05-31
Update instructions for token authentication in comments

mudphone:
The instructions currently refer to the old way to passing passing
parameters on `socket.connect`. If followed, this results in a
warning message in the JS console that "passing params to connect is
deprecated. Instead pass :params to the Socket constructor", see
[`phoenix.js L533`](https://github.com/phoenixframework/phoenix/blob/master/assets/js/phoenix.js#L533).

This confused me as I was reading through this, so I've created an
update that points to the correct location for passing the token.


chrismccord:
❤️❤️❤️

mudphone:
🏖 🎉

PR #2336, opened 2017-06-02
Replace 'umbrella application(s)' with 'umbrella project(s)'

webdeb:
Sometimes it's called `umbrella application`, sometimes `umbrella project`.

IMHO, it doesn't really matter how its called, but it also makes sense to call umbrellas as projects, cause it feels natural for a project to have multiple apps, docs, marketing department other sub-projects.

However, the main purpose of this PR is to avoid such fails ;)

https://travis-ci.org/phoenixframework/phoenix/jobs/237944489#L344

### DIFF
--- a/installer/lib/mix/tasks/phx.new.web.ex
+++ b/installer/lib/mix/tasks/phx.new.web.ex
@@ -3,7 +3,7 @@ defmodule Mix.Tasks.Phx.New.Web do
   Creates a new Phoenix web project within an umbrella project.
 
   It expects the name of the otp app as the first argument and
-  for the command to be run inside your umbrella application's
+  for the command to be run inside your umbrella project
   apps directory:
 
       $ cd my_umbrella/apps
@@ -11,7 +11,7 @@ defmodule Mix.Tasks.Phx.New.Web do
 
   This task is inteded to create a bare Phoenix project without
   database integration, which interfaces with your greater
-  umbrella application(s).
+  umbrella project(s).
 
   ## Examples
 
--- a/installer/test/phx_new_ecto_test.exs
+++ b/installer/test/phx_new_ecto_test.exs
@@ -15,7 +15,7 @@ defmodule Mix.Tasks.Phx.New.EctoTest do
   test "new without args" do
     in_tmp_umbrella_project "new without args", fn ->
       assert capture_io(fn -> Mix.Tasks.Phx.New.Ecto.run([]) end) =~
-             "Creates a new Ecto project within an umbrella application."
+             "Creates a new Ecto project within an umbrella project."
     end
   end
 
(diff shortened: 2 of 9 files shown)


josevalim:
FWIW, "umbrella project" is the parent project and "umbrella applications" are the child applications. That's a very small distinction though, so we should probably go with "umbrella projects" and "child applications" or "umbrella child applications" when the umbrella context is not implied.

StanBright:
@webdeb as you've noted, it feels natural for a project to have many applications. In that regards, it makes sense to refer to the children as "applications", and not "projects".

installer/lib/mix/tasks/phx.new.web.ex
```diff
@@ -3,15 +3,15 @@ defmodule Mix.Tasks.Phx.New.Web do
...
+  for the command to be run inside your umbrella project
   apps directory:
 
       $ cd my_umbrella/apps
       $ mix phx.new.web APP [--module MODULE] [--app APP]
 
   This task is inteded to create a bare Phoenix project without
   database integration, which interfaces with your greater
-  umbrella application(s).
+  umbrella project(s).
```
webdeb: -> umbrella child application(s)

installer/test/phx_new_ecto_test.exs
```diff
@@ -15,7 +15,7 @@ defmodule Mix.Tasks.Phx.New.EctoTest do
   test "new without args" do
     in_tmp_umbrella_project "new without args", fn ->
       assert capture_io(fn -> Mix.Tasks.Phx.New.Ecto.run([]) end) =~
-             "Creates a new Ecto project within an umbrella application."
+             "Creates a new Ecto project within an umbrella project."
```
webdeb: -> Creates a new Ecto application within an umbrella project.

installer/test/phx_new_web_test.exs
```diff
@@ -17,7 +17,7 @@ defmodule Mix.Tasks.Phx.New.WebTest do
   test "new without args" do
     in_tmp_umbrella_project "new without args", fn ->
       assert capture_io(fn -> Mix.Tasks.Phx.New.Web.run([]) end) =~
-             "Creates a new Phoenix web project within an umbrella application."
+             "Creates a new Phoenix web project within an umbrella project."
```
webdeb: -> Creates a new Phoenix web application within an umbrella project.

webdeb:
Hey guys, this was exactly my intention, to call umbrella application -> umbrella project, and umbrella childs as umbrella applications. I find that the suggested wording `umbrella child application` makes it much more clear, even if it feels a bit long at the first glance. It's obviously the `child` keyword.

In the meanwhile I've added some suggestions to the diff. If its ok, I'll commit them. Thanks for your attention!

webdeb:
By checking all the occurrences of the wording its clear, that the consolidation and the needed fixes is a more mature thing, and would involve a lot more changes on very different places. I think for now the pragmatic solution would be to fix the actual tests, and leave other places as is.

This is why I am closing this PR for now, in favour of #2349 
However, the discussion was worth it, thanks again!

PR #2352, opened 2017-06-15
No need to install brunch globally

raspo:
A global installation of brunch is not required as it can be executed using *npm scripts*. 
It also prevents any potential problems when there is a mismatch between the local version of brunch and the globally installed one.
I added two scripts for the most common tasks `watch` and `build`.

Let me know if I'm missing something.

### DIFF
--- a/package.json
+++ b/package.json
@@ -23,6 +23,8 @@
   "files": ["README.md", "LICENSE.md", "package.json", "priv/static/phoenix.js", "assets/js/phoenix.js"],
   "scripts": {
     "test": "./node_modules/.bin/mocha ./assets/test/**/*.js --compilers js:babel-register -r jsdom-global/register",
-    "docs": "documentation build assets/js/phoenix.js -f html -o doc/js"
+    "docs": "documentation build assets/js/phoenix.js -f html -o doc/js",
+    "watch": "brunch watch",
+    "build": "brunch build"
   }
 }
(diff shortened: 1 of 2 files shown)


package.json
```diff
@@ -23,6 +23,8 @@
   "files": ["README.md", "LICENSE.md", "package.json", "priv/static/phoenix.js", "assets/js/phoenix.js"],
   "scripts": {
     "test": "./node_modules/.bin/mocha ./assets/test/**/*.js --compilers js:babel-register -r jsdom-global/register",
-    "docs": "documentation build assets/js/phoenix.js -f html -o doc/js"
+    "docs": "documentation build assets/js/phoenix.js -f html -o doc/js",
+    "watch": "brunch watch",
+    "build": "brunch build"
```
OvermindDL1: The `./node_modules/.bin` directory is added to the `$PATH` by npm when it runs scripts, so just `brunch watch` and such is fine.

webdeb: oh, I did't knew, thx. And what if another `brunch` version is already installed globally?

raspo: @webdeb  The locally installed package will always be used by the scripts

raspo:
The build is failing but I really can't understand why :/
Is there something I need to do?

webdeb:
@raspo No, there are some other Tests failing, not your fault.

chrismccord:
❤️❤️❤️🐥🔥

PR #2354, opened 2017-06-16
Add phoenix to umbrella/app_name/mix.exs

webdeb:
This adds phoenix to the main application in an umbrella project. 
The purpose of this, is to make the generators available inside the **ecto application**.
phoenix will be included for **dev** ENV only, and only if **ecto** is enabled.

*Additionally*: Use elixir 1.5 list style

### DIFF
--- a/installer/templates/phx_umbrella/apps/app_name/mix.exs
+++ b/installer/templates/phx_umbrella/apps/app_name/mix.exs
@@ -31,8 +31,11 @@ defmodule <%= app_module %>.Mixfile do
   #
   # Type `mix help deps` for examples and options.
   defp deps do
-    [<%= if ecto do %>{:<%= adapter_app %>, ">= 0.0.0"},
-     {:ecto, "~> 2.1"}<% end %>]
+    [<%= if ecto do %>
+     {:<%= adapter_app %>, ">= 0.0.0"},
+     {:ecto, "~> 2.1"},
+     {:phoenix, "~> 1.3.0-rc", only: :dev}
+    <% end %>]
   end
 
   # Aliases are shortcuts or tasks specific to the current project.<%= if ecto do %>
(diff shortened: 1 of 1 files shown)


webdeb:
I think about this PR as a HOTFIX, because I believe it would make much more sense to move schema specific generators to ecto and use a proxy from phx.gen.schema -> ecto.gen.schema when generating with phx.gen.json or phx.gen.html. The ecto standalone generators could provide much more specific commands with respect to the models inside the given application.

chrismccord:
Thanks for the PR! I understand the task not being available in all apps can be confusing, but I think the best solution is to require the user to run the command within a given phoenix project. This solves the task issue obviously and is a natural place for the command when you think about it. Since phoenix is the web layer, the phoenix generators should be run from the web layer of your app. We introduced the `context_app` configuration so phoenix is "context aware" and can generate files within a separate umbrella app than the web one, but I think running the generators from the web apps is the way to go. Thanks!

webdeb:
Ah, ok. I did't knew about the `context_app` option, thanks. @chrismccord what do you think about the suggestion to put the schema generator into `ecto`?

chrismccord:
I don't think the generators belong in echo because phoenix's schema generator are necessarily tied to other features like the context generator. Coupling to ecto's version of the schema generator would make deviations more difficult in the future.

webdeb:
So, what is phoenix then? the web layer? 
Not only. 

Yes It's a **web framework**
- routing
- channels
- controllers
- templates

And also a **generator** for your
- project
- context
- endpoint
- schema

What I see, is that the web-layer is only one part of phoenix.
The other is the generator part which could be used separately. I believe, that it would make sense split it, and use separately. Maybe in future there will be a `phxgen` API, so that you can define and package your phxgens easily and share them on hex

Anyway, thanks @chrismccord, these are just some ideas for **v3** or so, lol.

And last but not least, if you only have to type `mix phxgen foo` or `mix gen foo` or `mix xgen foo` instead of `mix phx.gen foo` you see, that you'll probably save some of your time by not typing the `.`, and you will have more time to read Issues, PRs and discuss further ideas with some crazy people, or just have more time for things you really like `.` )

PR #2356, opened 2017-06-17
Progress output during code reloading

ciaran:
This is an attempt at adding browser feedback during code reload compilations.

I’m sure this will need at least some touch up work, but I’d like to make sure the approach is valid and have someone else test it before continuing.


lib/phoenix/code_reloader/server.ex
```diff
@@ -178,14 +192,31 @@ defmodule Phoenix.CodeReloader.Server do
...
     original_gl = Process.group_leader
-    {:ok, proxy_gl} = Proxy.start()
+    {:ok, proxy_gl} = Proxy.start(callback)
     Process.group_leader(self(), proxy_gl)
+    stderr =
+      case Process.whereis(:standard_error) do
+        nil ->
+          nil
+        pid ->
+          Process.unregister(:standard_error)
```
josevalim: This is not safe. Imagine you have processes running alongside your Phoenix application and they decide to write something to stderr, they can now randomly fail.

lib/phoenix/code_reloader/server.ex
```diff
@@ -13,8 +13,16 @@ defmodule Phoenix.CodeReloader.Server do
     GenServer.call(__MODULE__, :check_symlinks, :infinity)
   end
 
-  def reload!(endpoint) do
-    GenServer.call(__MODULE__, {:reload!, endpoint}, :infinity)
+  def reload!(conn, callback) do
```
josevalim: Hrm, I don't like that we went from a process that works only on the endpoint to a process that expects a connection to work. The ConnRegistry feels unnecessary, as it feels an indirection around the proxy.

Here is my suggestion: because we only run one code_reloader/server.ex, we can also start the proxy.ex when we start the code server, in the same supervision tree. Now that the proxy is always running, it will keep a list of pids it should proxy to. Then the compiler will do the following:

```elixir
original_gl = Process.group_leader
Proxy.capture(original_gl)
try do
  {..., Proxy.flush()}
after
  Proxy.uncapture()
end
```

Then, to redirect the processing to the browser, you will simply do:

```elixir
Proxy.forward_to(self())
```

Incorporating the registry functionality directly into the proxy. Then the proxy will forward to those processes when necessary and the CodeReloader will use `receive` to use those messages to render, instead of relying on callbacks.

This would allow us for example to even have routes such as `/__compile__` where you could be connected to the proxy all the time, regardless if you are also triggering the code reloader.

Does it make sense?

ciaran: The reason the ConnRegistry was required is that we are potentially modifying a `conn`, in that we change its state to `:chunked` (using `Plug.Conn. send_chunked`).

So I think merging the registry into Proxy is fine but I think we can't avoid having to track connections, so I believe `Proxy.un/capture` will have to receive/return a connection also (if the user wants to receive messages).

josevalim:
Why can't we keep all state in the connection process itself? And start
receiving messages there instead?
-- 


ciaran:
I’ve started changing to the approach you suggested, but there's one thing I'm unsure how to handle.

I have something like this on the CodeReloader side, which I believe is what you are suggesting:

```elixir
Proxy.forward_to(self())

CodeReloader.Server.async_compile()

receive do # Message sent by Proxy
  {:chars, msg} -> puts chars; receive again
  :done -> # sent on Proxy.uncapture, stop receive loop
end
```

So this has the server perform the compilation, and the proxy is forwarding output to the connection process.

However I'm not sure how best to have the server communicate the compilation result back to the plug with the current setup.

josevalim:
@ciaran one option is to do the connection writing in a task:

```
conn = set_conn_to_chunked(conn)
chunker = Task.async(fn -> chunk(conn) end)
Proxy.forward_to(chunker.pid)
res = CodeServer.Server.recompile!
Task.shutdown(chunker)
case res do
```

ciaran:
Pushed some work on the changes discussed above.

Additionally to avoid the standard error issue, Proxy now takes over stderr at startup,
and during capture it runs a child proxy which forwards group leader output.

lib/phoenix/code_reloader/server.ex
```diff
@@ -44,10 +44,10 @@ defmodule Phoenix.CodeReloader.Server do
 
   def handle_call({:reload!, endpoint}, from, state) do
     compilers = endpoint.config(:reloadable_compilers)
-    backup = load_backup(endpoint)
-    froms  = all_waiting([from], endpoint)
+    backup    = load_backup(endpoint)
```
josevalim: Please don't align `=`. It generates code churn. :)

lib/phoenix/code_reloader.ex
```diff
@@ -49,17 +51,201 @@ defmodule Phoenix.CodeReloader do
...
+  defp send_feedback?(conn),
+    do: accepts_html?(conn) and not is_ajax?(conn)
+
+  defp feedback_started?(conn),
+    do: conn.state == :chunked
+
+  defp accepts_html?(conn) do
+    conn
+    |> get_req_header("accept")
+    |> Enum.any?(&String.contains?(&1, "html"))
```
josevalim: `String.ends_with?(&1, "html")`?

lib/phoenix/code_reloader/proxy.ex
```diff
@@ -1,48 +1,125 @@
...
+
+    state = Map.put(state, :output, "")
+    state = Map.put(state, :captured, nil)
+    state = Map.put(state, :pids, [])
+
+    {:reply, :ok, state}
+  end
+
+  def handle_call(:flush, _from, state) do
+    broadcast state.pids, :flush
```
josevalim: This message does not seem to be used anywhere?

lib/phoenix/code_reloader.ex
```diff
@@ -49,17 +51,201 @@ defmodule Phoenix.CodeReloader do
...
-        |> halt()
+        |> send_chunked(200)
+        |> Plug.Conn.chunk(compile_template())
+      conn
+    else
+      conn
     end
   end
 
+  defp header_template(title) do
```
josevalim: Can't we use the existing template instead of building a new one?

lib/phoenix/code_reloader.ex
ciaran: I believe `contains` is better here, unless we want to start properly reading the accept field.

E.g. here is the `accept` value I'm testing against:
```"text/html,application/xhtml+xml,application/xml;q=0.9,image/webp,*/*;q=0.8"```

We could parse this properly, but since we only care if HTML is anywhere in the list (or even XHTML, in this case), I think the `String.contains` is the best option.

**Comment by ciaran** [comment]
I’ve added interval-based autoscrolling during output too.

I think it’s tricky to really do this properly (would be nicer to not scroll if user has scrolled up) without some larger cross-browser code, so I’m not convinced about keeping this.

**Comment by josevalim** [comment]
@ciaran yes, let's keep it simple. Let's make sure the code is tidy for now before we end-up adding more and more features.

**Comment by ciaran** [comment]
I think what’s there is pretty solid now.

There is a fair amount of duplicated code in `GroupLeaderProxy`/`Proxy`, but I think trying to avoid this would just add complication.

File: lib/phoenix/code_reloader.ex [hunk 7]
```diff
@@ -49,26 +51,86 @@ defmodule Phoenix.CodeReloader do
...
   """
   def call(conn, opts) do
-    case opts[:reloader].(conn.private.phoenix_endpoint) do
-      :ok ->
+    {conn, res, output} = call_reloader(conn, opts[:reloader])
+
+    case {res, feedback_started?(conn)} do
+      {:ok, true} ->
+        Plug.Conn.chunk(conn, "Done!<br />\n")
+        Plug.Conn.chunk(conn, ~s(<meta http-equiv="refresh" content="0">\n))
```
josevalim: I am worried about the impact of this line on developer using Windows. Mix tracks mtimes and they are very fragile on Windows. Some users report recompilation of some files on every access and this refresh may put those applications in an infinite loop.

lib/phoenix.ex
```diff
@@ -48,7 +48,8 @@ defmodule Phoenix do
 
     children = [
       # Code reloading must be serial across all Phoenix apps
-      worker(Phoenix.CodeReloader.Server, [])
+      worker(Phoenix.CodeReloader.Server, []),
+      worker(Phoenix.CodeReloader.Proxy, []),
```
josevalim: Shouldn't the proxy be started first since the server calls it?

lib/phoenix/code_reloader/colors.ex
```diff
@@ -0,0 +1,15 @@
+defmodule Phoenix.CodeReloader.Colors do
+  defp css_color(n) when is_binary(n),
+    do: String.to_integer(n) |> css_color()
+  defp css_color(31), do: :red
+  defp css_color(33), do: "#b3b30a"
```
josevalim: We need to emit an empty binary if we don't know the number. Also note they can be combined (red + bold). Some unit tests here would be nice.

lib/phoenix/code_reloader/colors.ex
```diff
@@ -0,0 +1,15 @@
...
+
+  def to_html(string) do
+    string = Plug.HTML.html_escape(string)
```
josevalim: We are already calling `html_escape` on the input before calling this function. Won't this lead to double escaping?

lib/phoenix/code_reloader/proxy.ex
```diff
@@ -1,48 +1,120 @@
...
+# This proxy is spawned at startup and replaces (and forwards to)
+# the standard_error process.
 defmodule Phoenix.CodeReloader.Proxy do
   @moduledoc false
   use GenServer
 
-  def start() do
-    GenServer.start(__MODULE__, "")
+  def start_link() do
+    stderr = Process.whereis(:standard_error)
```
josevalim: It is worth noting that this process always runs... including in production. And I am not sure if that's ok.

lib/phoenix/code_reloader/group_leader_proxy.ex
```diff
@@ -0,0 +1,64 @@
+# Spawned by the Proxy when capture is started,
+# to capture stdout of the compilation process.
+# Restores the original group leader when terminated after compilation.
+# Output is forwarded to the parent proxy with the :stdout message.
+defmodule Phoenix.CodeReloader.GroupLeaderProxy do
```
josevalim: What if you make this a regular process started via `start_link` inside `proxy` so you can re-use the handle_info and put_chars logic?

josevalim: Something like this:

```elixir
defp handle_io(msg, device) do
  case msg do
    {:io_request, ..., ...}
    _ ->
  end
end
```

then the group leader proxy is:

```
Task.start_link(fn ->
  replace group leader...
  loop("")
end)

defp loop(arg1, arg2, arg3) do
  receive do
    :stop -> ...
    msg ->
      buffer = handle_io(...)
      loop(buffer)
```

and the `handle_info` of the proxy also calls handle_io when necessary.

lib/phoenix/code_reloader.ex
scrogson: Maybe we can guard with `match?({:unix, _}, :os.type)` ?

lib/phoenix/code_reloader/group_leader_proxy.ex
ciaran: 86adec55a8ff50000cd9cb7903ce4dfd27a8cb34 has my attempt at this refactoring.

lib/phoenix/code_reloader/proxy.ex
ciaran: I’ve not made any changes wrt. this as I’m unsure the best way to solve it.

Perhaps have the calls to `Proxy.capture` check if it has been started yet and then start it at first call if necessary?

lib/phoenix/code_reloader.ex
ciaran: I’ve added an `is_unix?` check to the `send_feedback?` test, so this feature won’t be active on Windows.

lib/phoenix/code_reloader/colors.ex
ciaran: I’ve rewritten the Colors module to handle more complex inputs (combined/overwritten styles etc.) and added some tests.

lib/phoenix/code_reloader/proxy.ex
ciaran: The Proxy is now only started when the Server needs it (2a36dc00eafd90dd6095fa06f10947a7d0e87f48)

josevalim:
Hi @ciaran!

Sorry for the delay on this. @chrismccord and I have discussed this issue and we have agreed that the fact there is not a good moment to capture standard error. When we did it dynamically we could cause errors in running systems. Doing it at startup means it affects production. Alternatively we could introduce another configuration option... but we are not convinced either.

My suggestion is to move this to a separate package. It will be able to develop faster than it would be in Phoenix (which has stalled for 6 months because of us) and maybe we can solve this issue faster if we have more eyes looking at it.

In any case, thanks for the PR! Very appreciated. :heart:

ciaran:
I think I had initially tried to create this as a separate package, but there were some hardcoded references to the built-in code reloader within Phoenix which meant it wasn't possible at the time. 

So I would like to move to a separate package but first I will need to make a PR which removes this restriction (and wait for this to be in a release).

josevalim:
@ciaran I think we would need a way to redirect the compilation output somewhere of your choice. That sounds doable to me and much smaller in scope. :)

axelson:
If it was possible to redirect the compilation output to somewhere else of your choice then it would be easier to show compilation errors on the Phoenix error page :+1:

PR #2360, opened 2017-06-19
Prevent non-atoms being used as plugs

Gazler:
When a non-atom is passed to the `plug` macro, an error is raised at
compile time.

This is base on https://github.com/phoenixframework/phoenix_guides/issues/696#issuecomment-309361956

This does however, eliminate a valid use case:

```elixir
defmodule TestPlug do
  defmacro foo, do: :fetch_session
end
```

Is there a way to determine if an AST is a macro? @josevalim If you want this, I can also port it to `Plug.Builder`

### DIFF
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -462,6 +462,11 @@ defmodule Phoenix.Router do
   See `pipeline/2` for more information.
   """
   defmacro plug(plug, opts \\ []) do
+    if not is_atom(plug) do
+      raise ArgumentError, "the plug must expects an atom (for example `:fetch_session`) " <>
+        "or a module (for example `Plug.Logger`). Got #{Macro.to_string(plug)}"
+    end
+
     quote do
       if pipeline = @phoenix_pipeline do
         @phoenix_pipeline [{unquote(plug), unquote(opts), true}|pipeline]
--- a/test/phoenix/router/pipeline_test.exs
+++ b/test/phoenix/router/pipeline_test.exs
@@ -134,4 +134,18 @@ defmodule Phoenix.Router.PipelineTest do
     conn = call(Router, :get, "/browser/hello")
     assert conn.assigns[:params] == %{"id" => "hello"}
   end
+
+  test "pipeline with a non-atom plug raises" do
+    assert_raise(ArgumentError, fn ->
+      Code.eval_string("""
+        defmodule Phoenix.Router.PipelineTest.RouterWithException do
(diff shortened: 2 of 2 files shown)


lib/phoenix/router.ex
```diff
@@ -462,6 +462,11 @@ defmodule Phoenix.Router do
   See `pipeline/2` for more information.
   """
   defmacro plug(plug, opts \\ []) do
+    if not is_atom(plug) do
+      raise ArgumentError, "the plug must expects an atom (for example `:fetch_session`) " <>
+        "or a module (for example `Plug.Logger`). Got #{Macro.to_string(plug)}"
+    end
+
     quote do
```
josevalim: Do this:

```elixir
quote bind_quoted: [plug: plug, opts: opts] do 
  if not is_atom(plug) do ... end
  if pipeline = @phoenix_pipeline do
     @phoenix_pipeline [{plug, opts, true}|pipeline]
```

michalmuskala:
If we call `Macro.expand` on the argument before checking it's an atom, macros, module attributes, etc should work just fine.

lib/phoenix/router.ex
josevalim: This way we still maintain valid use cases as well as the runtime capabilities.

josevalim:
@michalmuskala no need for Macro.expand, just check it inside the quote. :)

Gazler:
@josevalim This will result in an `UndefinedFunctionError` which is what was happening before, since the call to `TestPlug.foo` will happen at compile time.

josevalim:
This is valid code today:

    foo = :foo
    plug foo

I don't think we should stop supporting it.
-- 


Gazler:
@josevalim I wasn't suggesting we stop supporting it. The goal here is to provide a better error if someone does:

`plug MyModule.some_function`

Which would only work if `MyModule.some_function` returned an atom or a module.

Currently this will fail to compile with the following error:

 > ** (UndefinedFunctionError) function MyModule.some_function/0 is undefined or private

Or, more likely:

> ** (UndefinedFunctionError) function MyModule.some_function/0 is undefined or private. Did you mean one of:
>
>     * some_function/2


It would be better if we could point them towards one of the two valid values (atom or module name) for the plug macro.

josevalim:
I know, what I am saying is that it is not possible to fix the given
example without breaking the snippet sample I sent or special case it.
-- 


Gazler:
Closing as I don't think there is an easy way to fix this.

PR #2362, opened 2017-06-22
Add multi flash message support

JensDebergh:
When generating a new project some hard coded `get_flash` calls get inserted into the layout template. This results is some manual labor for the developer to add missing flash messages.

Adding `get_flash/1` inside the the view module and looping inside the template can generate multiple alerts without hardcoding.

```
       # controller
	def contact(conn, _params) do
		conn
		|> put_flash(:success, "Hello Phoenix!")
		|> put_flash(:info, "Hello Phoenix!")
		|> render("contact.html")
	end
```

screen shot 2017-06-22 at 8 46 07 am

Note: I'm having issues finding the failing test. If anyone could point me into the right direction

### DIFF
--- a/installer/templates/new/web/templates/layout/app.html.eex
+++ b/installer/templates/new/web/templates/layout/app.html.eex
@@ -22,8 +22,9 @@
         <span class="logo"></span>
       </header>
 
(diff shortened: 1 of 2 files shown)


chrismccord:
I think multiple flash messages is an exception from the norm so I'm okay with folks having to do some manual work once they require it. Thanks!

PR #2363, opened 2017-06-22
Refactor phoenix.js

webdeb:
Use ES6 where possible, rely on wide used standards, use let only for variables, which are going to be changed. Otherwise use `const` and so on. 

I hope it provides also better readability.

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -180,9 +180,14 @@
  */
 
 const VSN = "1.0.0"
-const SOCKET_STATES = {connecting: 0, open: 1, closing: 2, closed: 3}
 const DEFAULT_TIMEOUT = 10000
 const WS_CLOSE_NORMAL = 1000
+const SOCKET_STATES = {
+  connecting: 0,
+  open: 1,
+  closing: 2,
+  closed: 3
+}
 const CHANNEL_STATES = {
   closed: "closed",
   errored: "errored",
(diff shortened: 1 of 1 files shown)


assets/js/phoenix.js
```diff
@@ -255,34 +264,37 @@ class Push {
...
-
   // private
-
-  matchReceive({status, response, ref}){
-    this.recHooks.filter( h => h.status === status )
-                 .forEach( h => h.callback(response) )
+  matchReceive({ status, response, ref }) {
+    this.recHooks.forEach(h =>
+      h.status === status && h.callback(response)
+    )
```
webdeb: Remove unnecessary filter loop

chrismccord:
We don't accept purely stylistic PRs as it introduces churn and potential conflicts for progress going forward. Thanks!

aphillipo:
This a little more efficient (o(n) vs o(2n)) as well as being a stylistic change. It’s probably not necessary to loop twice here ever that I can think of. If recHooks can never be large then probably it’s not worth changing but if it can be a few hundred items maybe consider it.

webdeb:
The whole `phoenix.js` feels a bit messy, and inconsistent in its styling, this reduces readability IMHO.

However, #2368 removes the unnecessary double-loops.

odyright:
Hello, consider if we use eslint, what would be the best config...?

webdeb:
@odyright I would stick with the **eslint:recommended**, then add your own as you go.

PR #2364, opened 2017-06-22
Use webpack

dparnell:
Hi,
  First I just want to say thank you so much for making phoenix. It has been a joy to work with :)
I've been building quite a few sites using phoenix and Vue.js recently which required me to manually install webpack each time. Since I'm lazy I decided to see how hard it would be to update `phx.new` to produce a webpack environment rather than brunch and this pull request is the result.

Thanks for making web development fun again :)

Daniel


josevalim:
Thank you @dparnell. You can find previous discussions on the past, but the TL;DR is that we don't plan to replace brunch by something else nor support multiple options. The best way to go is to indeed replace it or have some sort of phx.new.webpack as a separate package you would run afterwards.

dparnell:
Thanks. I'll take a look and see how hard it would be to make `phx.new.webpack` that would be a nice way to do it also :)

PR #2368, opened 2017-06-22
remove double loops in phoenix.js

webdeb:


chrismccord:
Thanks! I have made changes to the js client that conflict with this branch. I'm okay with the existing filter + map as you won't have large numbers of filtered results to map over, and in practice it will almost always be a single channel instance so I'm 👎 on this change.

PR #2376, opened 2017-07-03
Add typed param feature

lancer26:
Found this [discuss](https://elixirforum.com/t/typed-url-parameters-in-phoenix/2633), I think a lot of people need typed parameter feature. Hope this will be help.


josevalim:
If many developers need it, we recommend it to be built as a separate plug so anyone can bring to their projects when necessary. Thank you.

PR #2380, opened 2017-07-07
Fix documentation on format encoders

jayjun:
`:format_encoders` example shows `Phoenix.HTML.Engine` set as the encoder for HTML.

However, `Phoenix.Template.HTML` is the module that implements `encode_to_iodata!/1`, not `Phoenix.HTML.Engine`.


josevalim:

PR #2383, opened 2017-07-13
Add moduledoc to application file

acconrad:
It was added [in the general umbrella app template](https://github.com/phoenixframework/phoenix/blob/master/installer/templates/phx_umbrella/apps/app_name/lib/app_name/application.ex) but was never added to the other application file templates. Credo will fail with a default Phoenix app and since it was already in 1 file, seemed harmless to add to the other template files and will get Credo to pass with a stock `phx.new` install.

### DIFF
--- a/installer/templates/phx_single/lib/app_name/application.ex
+++ b/installer/templates/phx_single/lib/app_name/application.ex
@@ -1,4 +1,12 @@
 defmodule <%= app_module %>.Application do
+  @moduledoc """
+  The <%= app_module %> Application Service.
+  
+  The <%= app_name %> system business domain lives in this application.
+  
+  Exposes API to clients such as the `<%= app_module %>.Web` application
+  for use in channels, controllers, and elsewhere.
+  """
   use Application
 
   # See https://hexdocs.pm/elixir/Application.html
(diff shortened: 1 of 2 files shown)


chrismccord:
Thanks for the PR! I think docs around this would better live in the main "App" module, rather than the supervision tree. That said, I'm not sure about loading up the terms "service" will be helpful for folks.

Mehonoshin:
Hey, guys.
I wonder why this PR was not merged.
Maybe we can add `@moduledoc false` at least here?
The reason why I ask is that a newly generated phoenix app does not pass `credo` code style checks.

@chrismccord WDYT about `credo`? If you accept it's style guide, then maybe we should follow it at new apps?

josevalim:
I would be happy with @moduledoc false. Please send a new PR?

Credo is great but it is not the job of Phoenix to impose it.
-- 


Mehonoshin:
@josevalim 👍

Mehonoshin:
Looks like it has already been done, so versions after `1.3.1` will generate valid apps.

fertapric:
If it helps, I've been using this doc to document my `MyApp.Application` modules:

```
@moduledoc """
Defines an Elixir (Erlang/OTP) application.

See `Application` for more information.
"""
```

PR #2385, opened 2017-07-15
Add :force option to Mix.Phoenix.copy_from/4 params

mogadget:
Add an option :force to copy_from/4 parameters.  This will overwrite an existing file in the target directory without a shell prompt [Yn] confirmation.

### DIFF
--- a/lib/mix/phoenix.ex
+++ b/lib/mix/phoenix.ex
@@ -25,8 +25,12 @@ defmodule Mix.Phoenix do
 
   Files are evaluated against EEx according to
   the given binding.
+
+  ## Options
+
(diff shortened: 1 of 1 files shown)


chrismccord:
This options list never appears to make it down to the `copy_from` function from any call site? Can you verify mix cli args are properly passed down? Thanks!

mogadget:
@chrismccord im calling `copy_this` from another function instead of cli.  I thought of re-using this function not just through cli.

chrismccord:
Ah I see. This is for your own custom purposes. Right now, this module is `@moduledoc false`, so it is a private API that we can't guarantee backwards compatibility on. For your use case, I would add your own `copy_from` function that does what you need. Thanks!

PR #2387, opened 2017-07-17
Add channel protocol V2 

chrismccord:
Adds version 2 of the channel wire protocol to resolve a race condition that was discovered when a client times out and rejoins. Backwards compatible with V1 clients.

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -473,62 +473,77 @@ export class Channel {
   isLeaving(){ return this.state === CHANNEL_STATES.leaving }
 }
 
-  /** Initializes the Socket
-   *
-   *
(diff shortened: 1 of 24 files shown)


assets/js/phoenix.js
```diff
@@ -473,62 +493,77 @@ export class Channel {
...
-   * @param {number}  opts.longpollerTimeout - The maximum timeout of a long poll AJAX request.
-   *
-   * Defaults to 20s (double the server long poll timer).
-   *
-   * @param {Object}  opts.params - The optional params to pass when connecting
-   *
-   *
-  */
-export class Socket {
+let Encoder = {
```
scrogson: Since this has both `encode` and `decode` functions, should we call this `Serializer` or `Codec` instead?

chrismccord: 👍🏻

PR #2388, opened 2017-07-18
Update Channel Terminate documentation

Azolo:
I was a little bit confused by this statement:

> `terminate/2`, however, won't be invoked in case of errors nor in case of exits. This is the same behaviour as you find in Elixir abstractions like `GenServer` and others.

Mainly because the way I understand things, it just isn't true. Not only that, but it is [tested to be untrue][1].

I am submitting a PR because it's nice if you can be lazy and just merge a change. But I feel like there was some other warning that was trying to be conveyed.

[1]: https://github.com/phoenixframework/phoenix/blob/5ef5d112bebb1bb241e4696b03b4ce8ed77b5ec8/test/phoenix/test/channel_test.exs#L261

### DIFF
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -164,8 +164,8 @@ defmodule Phoenix.Channel do
   If any of the callbacks return a `:stop` tuple, it will also
   trigger terminate with the reason given in the tuple.
 
-  `terminate/2`, however, won't be invoked in case of errors nor in
-  case of exits. This is the same behaviour as you find in Elixir
+  `terminate/2`, however, won't be invoked in the case of a linked process
+  non-normal exit. This is the same behaviour as you find in Elixir
   abstractions like `GenServer` and others. Typically speaking, if you
   want to clean something up, it is better to monitor your channel
   process and do the clean up from another process.  Similar to GenServer,


michalmuskala:
The rules of when terminate is or isn't called are pretty complex 😞 

This document sums it up: https://gist.github.com/mrallen1/806fe5506132260574af33e99dadd499

I think the primary goal of this documentation should be to make it clear that it might not be called in some cases and that it behaves exactly like `terminate/2` in `GenServer`.

Azolo:
Yeah, you're right. Any exit signal will cause it to not be run unless you're trapping exits.

But I don't think that the rules of when it is called and when it isn't invoked are confusing as much as the difference between `Kernel.exit` and `Process.exit` is. Using `Kernel.exit(:my_reason)` in a callback will invoke terminate while `Process.exit(self(), :my_reason)` will not.

I think the only time terminate won't be invoked is from an untrapped exit signal, an untrappable `:kill` exit signal, or a `:brutal_kill` system termination. However that statement requires you to know about exit signals and how supervision works with the `:sys` module. Of course I may be missing something else, but I don't think am.

chrismccord:
The statement in the docs is correct, assuming you are not trapping exits, which is the default behavior, no? Can you tweak your doc changes to maintain our current docs, but clarify with trapping exits? Thanks!

Azolo:
@chrismccord Sorry for being confrontational, but the way I read the docs is that doing something like:

```elixir
def handle_in(_,_,_), do: exit(:blah)
```

or

```elixir
def handle_in(_,_,_), do: raise "Blah"
```

Then terminate won't be called in either case.

Which is why I am confused, because terminate will be called in both of those cases whether exits are being trapped or not.

The only case where exits being trapped matters is when I do:

```elixir
def handle_in(_,_,_), do: Process.exit(self(), :blah)
```

Azolo:
Closing this because as the discussion this is confusing and even good understanding of the rules is hard to explain.

PR #2395, opened 2017-07-21
fix docs: phx does not prepend context name to table names anymore.

aesmail:
Phoenix used to generate tables like `<context_name>_<table_name>` but not anymore.
The docs for `phx.gen.context` state that the generator will generate normal table names without prepending the context name, which is the actual behavior.
This PR is just to fix the docs for `phx.gen.html` and `phx.gen.json`.


josevalim:

PR #2397, opened 2017-07-24
Fix CHANGELOG version typo

nickurban:
Looks like @chrismccord  typed rc.2 by mistake instead of rc.3.


scrogson:
Hey @nickurban looks like there was another similar PR (#2398). I saw that one first and merged it. Sorry.

PR #2407, opened 2017-07-25
Phoenix.ConnTest.recycle/1 resets Plug.Conn.host

mobileoverlord:
In the current implementation:

`build_conn/0` defaults to generating a conn where the host is set to `"www.example.com"` because it passes `"/"` which `URI.parse` will return `nil`. You can trace that to here

https://github.com/elixir-plug/plug/blob/master/lib/plug/adapters/test/conn.ex#L21

You can override this by using `build_conn/3` and pass 
`build_conn(:get, "http://localhost/", nil)`

This causes the conn to set `host` to `"localhost"`. The problem is when using `build_conn/3` in conjunction with `recycle/1`. Currently, `recycle/1` starts off by calling `build_conn/0`. 

https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/test/conn_test.ex#L467

This PR adds `build_conn/1` which takes a conn and derive the method and host to call `build_conn/3`

I am unsure if this is the right way of handling the situation, but I believe that its a bug if recycle does not derive off the initial conn.

Example of usage:
```elixir 
  defp active_login(admin) do
    build_conn(:get, "http://localhost/", nil)
      # do stuff...
      |> recycle()
  end
```
Expected: `%Conn{host: "localhost"}`
Received: `%Conn{host: "www.example.com"}`

This needs a test or two, but I wanted to get the ball rolling to see if this is how you guys wanted to handle this.

### DIFF
--- a/lib/phoenix/test/conn_test.ex
+++ b/lib/phoenix/test/conn_test.ex
@@ -465,6 +465,7 @@ defmodule Phoenix.ConnTest do
   @spec recycle(Conn.t) :: Conn.t
   def recycle(conn) do
     build_conn()
+    |> Map.put(:host, conn.host)
     |> Plug.Test.recycle_cookies(conn)
     |> copy_headers(conn.req_headers, ~w(accept))
   end
--- a/test/phoenix/test/conn_test.exs
+++ b/test/phoenix/test/conn_test.exs
@@ -160,6 +160,11 @@ defmodule Phoenix.Test.ConnTest do
     assert conn.cookies == %{"req_cookie"  => "req_cookie",
                              "over_cookie" => "pos_cookie",
                              "resp_cookie" => "resp_cookie"}
+
+    conn =
+      build_conn(:get, "http://localhost/", nil)
+      |> recycle()
+    assert conn.host == "localhost"
   end
 
   test "ensure_recycled/1" do


lib/phoenix/test/conn_test.ex
```diff
@@ -464,7 +479,8 @@ defmodule Phoenix.ConnTest do
   """
   @spec recycle(Conn.t) :: Conn.t
   def recycle(conn) do
-    build_conn()
+    build_conn(conn)
```
josevalim: Instead of changing `build_conn`, wouldn't be better to have a function called `copy_host` or `copy_request_uri` or similar? That's how we handle copy_headers, for example.

josevalim: In any case, we should hold until after v1.3 is out, because this may break the tests of people relying on the old, although arguably worse, behaviour.

mobileoverlord:
I updated this to copy the host and params which would be set originally from `build_conn/3`. As I agree that this may backwards break peoples tests, that seems like a valid use case to get it in for 1.3 since it signifies breaking change.

josevalim:
@mobileoverlord if you are copying `params`, then you should not be using recycle. The goal of recycle is to prepare for a new request and keep only the information that would be resent on the new request, such as cookie headers and previous accept.

> As I agree that this may backwards break peoples tests, that seems like a valid use case to get it in for 1.3 since it signifies breaking change.

The problem is that 1.3 is in the last RC and I would be wary to introduce something this late in the game that has not been yet tested. If you need this right now, you can introduce `recycle_with_params` or something similar in your app. :)

LostKobrakai:
I do feel like this is an issue as well. Users won't suddenly switch domains in the browser as well. It just doesn't come up for people using only a single domain.

By now I've gone ahead and replaced `use Phoenix.ConnTest` in `test/support/conn_case.ex` with:
```elixir
      import Plug.Conn
      import Phoenix.ConnTest, except: [recycle: 1]

      def recycle(conn) do
        conn
        |> Phoenix.ConnTest.recycle()
        |> Map.put(:host, conn.host)
      end
```

chrismccord:
❤️❤️❤️🐥🔥

LostKobrakai:
What do you think about backporting such changes to stable 1.3? Without that it's kind of a pain to test things which rely on the host being set.

PR #2415, opened 2017-07-29
Remove flash function inconsistency

odiumediae:
`Phoenix.Controller.clear_flash/1` resets the flash messages to en empty map.

Raising an error via `Phoenix.Controller.get_flash/1` when there are no flash messages is therefore inconsistent. Better set the default value to an empty map and save a needless exception.

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -1224,8 +1224,7 @@ defmodule Phoenix.Controller do
 
   """
   def get_flash(conn) do
-    Map.get(conn.private, :phoenix_flash) ||
-      raise ArgumentError, message: "flash not fetched, call fetch_flash/2"
+    Map.get(conn.private, :phoenix_flash, %{})
   end
 
   @doc """


josevalim:
Sorry but I don't fully follow your logic. The current error exists for those who do not call `fetch_flash`. It is *not* related to clear_flash.

odiumediae:
@josevalim You're of course right, the error is not related to `clear_flash`. However, `get_flash/1` shouldn't throw an error in the first place, since the docs for `get_flash/1` explicitly state `Returns a map of previously set flash messages or an empty map.`.

That means `conn.private.phoenix_flash` is not supposed to be nil, or am I misunderstanding something here? It should return a map of flash messages or an empty map. When `clear_flash` is called it does exactly the same, it sets `conn.private.phoenix_flash` to an empty map. So why doesn't `get_flash/1` do the same in order to be consistent?

If I am being thick or stupid here, please explain.

aphillipo:
@odiumediae looks like the fetch_flash function needs to be called and that will (probably) set things as you expect.  The code uses an unset value to trigger a suggestion to the user, this is okay. Presumably you aren’t seeing the error message as I assume Phoenix calls fetch_flash somewhere?

josevalim:
@odiumediae can you please post your example? This definitely works here:

```elixir
conn = fetch_flash(conn)
get_flash(conn, :hello)
```

The point is that the flash needs to be explicitly loaded or cleared before used, otherwise, you will keep on getting an empty map and later you will be confused why the value you set was not previously loaded.

odiumediae:
Thank you very much @josevalim and @aphillipo!

I solved my problem and I now see what I did wrong and that throwing an exception in `Phoenix.Controller.get_flash/1` is perfectly alright, when fetch_flash isn't being called dbeforehand. 

I forgot to run the connection through the `:browser` pipeline in my router, because I am simultaneously working on a tutorial about pipelines. So from my perspective everything should have worked fine, because it also worked in other routes, so I went through the Phoenix.Controller code and misunderstood the exception.

So this was totally my fault and this pull request can safely be ignored/deleted. Sorry for wasting your time with my stupidity.

josevalim:
Glad to have helped and don't worry about it, have fun!

PR #2416, opened 2017-07-29
change getting started url in the default layout

samdvr:
**Reason for change:**
I did this because http://www.phoenixframework.org/docs url seems to have moved and returns a 404 currently. 

How change solved the issue:
I changed it to http://hexdocs.pm/phoenix/adding_pages.html which was the most appropriate link I found in the docs. I'd be happy to change it to something more useful if the community has a better suggestion.

### DIFF
--- a/installer/templates/phx_web/templates/layout/app.html.eex
+++ b/installer/templates/phx_web/templates/layout/app.html.eex
@@ -16,7 +16,7 @@
       <header class="header">
         <nav role="navigation">
           <ul class="nav nav-pills pull-right">
-            <li><a href="https://hexdocs.pm/phoenix/adding_pages.html">Get Started</a></li>
+            <li><a href="https://hexdocs.pm/phoenix">Get Started</a></li>
           </ul>
         </nav>
         <span class="logo"></span>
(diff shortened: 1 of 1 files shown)


Gazler:
@chrismccord I think we should add a redirect from http://www.phoenixframework.org/docs to https://hexdocs.pm/phoenix to solve this.

chrismccord:
Unfortunately it doesn't appear possible to add an s3 redirect for a root index object :(

chrismccord:
@samdvr Thanks for the PR! Can you update the link to point to `https://hexdocs.pm/phoenix` ?

chrismccord:
Nevermind, got it. The metadata could only be set form the cli for some reason:

```
$ aws s3 cp index s3://phoenixframework.org/docs/index --content-type 'text/html' --acl public-read --website-redirect 'https://hexdocs.pm/phoenix'
```

I think we should still update the link to hexdocs, but redirect is in place.

samdvr:
@Gazler @chrismccord  thanks for the quick response updated the link to [https://hexdocs.pm/phoenix](https://hexdocs.pm/phoenix)

chrismccord:
❤️❤️❤️🐥🔥

PR #2418, opened 2017-07-29
Add an eslint config

davydog187:
Javascript is a quirky, interesting language that needs to be tamed :) I propose adding an eslint config that runs as part of the tests, to ensure that we're writing high quality javascript that avoids common gotchas.

This is an initial proposal, but the configuration is negotiable. I recommend turning on rules such as `no-const`, as well as some other rules that push authors towards functional javascript.

Thanks!

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -283,7 +283,7 @@ class Push {
   /**
    * @private
    */
-  matchReceive({status, response, ref}){
+  matchReceive({status, response}){
     this.recHooks.filter( h => h.status === status )
                  .forEach( h => h.callback(response) )
   }
@@ -529,7 +529,7 @@ export class Channel {
    * @param {Object} payload
    * @param {integer} ref
    */
-  onMessage(event, payload, ref){ return payload }
+  onMessage(event, payload){ return payload }
 
   /**
    * @private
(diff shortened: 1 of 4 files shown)


davydog187:
The added `.eslintrc.json` was generated using `eslint --init` that inspects the existing javascript code.

assets/js/phoenix.js
```diff
@@ -851,7 +855,7 @@ export class LongPoll {
     })
   }
 
-  close(code, reason){
```
davydog187: If the intention here was documented passed values, I can add the relevant jsdoc

assets/js/phoenix.js
```diff
@@ -279,7 +279,7 @@ class Push {
     this.sent         = false
   }
 
-  matchReceive({status, response, ref}){
```
davydog187: same here

Gazler:
This has been discussed before:

https://github.com/phoenixframework/phoenix/pull/2262
https://github.com/phoenixframework/phoenix/pull/2363

davydog187:
Difference is that this adheres to the existing styles and only ensures consistency moving forward

chrismccord:
Hey @davydog187 ,

Sorry this took so long for me to look at. For now I would prefer to wait on adding more JS tooling. I need less JS tooling in my life, not more :). That said, I absolutely see the desire and need for liners in the JS world and would use one on a large JS applications if I was in the setting. Thanks!

davydog187:
Thanks chris!

PR #2420, opened 2017-07-30
Fixed generate typo in phx.gen.html and phx.gen.json docs

w0rd-driven:
I noticed a very small typo getting re-acquainted with mix tasks under the `Web namespace` section (https://hexdocs.pm/phoenix/1.3.0/Mix.Tasks.Phx.Gen.Json.html#module-web-namespace) that appears to affect both `phx.gen.html` and `phx.gen.json`. I looked for similar PR's to see if this was addressed recently and noticed #2406 also didn't cover both files, so I included that as well.


Gazler:
Thanks <3

PR #2422, opened 2017-07-31
Fixed info about timeout duration in phoenix.js

PatNowak:
After VSN 2.0 the comment about timeout is not valid.


chrismccord:
Thanks!

PR #2424, opened 2017-07-31
Make application.ex match upgrade instructions

michaelgundlach:
The [Phoenix 1.2 -> 1.3 upgrade instructions](https://gist.github.com/chrismccord/71ab10d433c98b714b75c886eff17357) tell you to "remove the config_change callback", but the codegen incorrectly creates this callback.

Let me know if I should have created a topic branch per the contributing guidelines; this seemed too trivial for that.

(Additionally, for Elixir 1.5 users application.ex is using a deprecated module (Supervisor.Spec).  Can/should the codegen target different Elixir versions?)


josevalim:
Good catch! The instructions are wrong. The code is correct. ;) @chrismccord can you update the instructions?

Thanks @michaelgundlach!

chrismccord:
Fixed. Thanks!

PR #2429, opened 2017-08-02
Ref generation for channel even handlers

youroff:
Following this discussion: https://groups.google.com/forum/?utm_medium=email&utm_source=footer#!msg/phoenix-core/vHq964tYMqI/Sw_5t4tQAAAJ

This PR introduces ref counter in the channel. Now `channel.on(...)` returns a unique integer, which can be used to remove one exact even handler and not all handlers matching the event as it was before.

```javascript
     ref1 = channel.on("event", do_stuff)
     ref2 = channel.on("event", do_other_stuff)
     channel.off("event", ref1)
     // Since unsubscription, do_stuff won't fire,
     // while do_other_stuff will keep firing on the "event" 
```

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -331,6 +331,7 @@ export class Channel {
     this.params      = params || {}
     this.socket      = socket
     this.bindings    = []
+    this.bindingRef  = 0
     this.timeout     = this.socket.timeout
     this.joinedOnce  = false
     this.joinPush    = new Push(this, CHANNEL_EVENTS.join, this.params, this.timeout)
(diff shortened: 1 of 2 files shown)


assets/js/phoenix.js
```diff
@@ -392,9 +393,30 @@ export class Channel {
     this.on(CHANNEL_EVENTS.error, reason => callback(reason) )
```
blatyo: Can you have this return as well?

blatyo: Also, onClose which is a couple lines above.

chrismccord:
Beautiful. Thank you!

mxs:
This is awesome and much needed. Any indication on when this will be published to npm in the next version?

PR #2431, opened 2017-08-04
Better english grammar

gilbert:
A small but proper fix.


chrismccord:
Thanks Gilbert! I prefer the existing phrase for our tagline but I appreciate the intent to make it better :)

christopheradams:
To be grammatically correct, shouldn't it be "or"? :)

PR #2449, opened 2017-08-20
Improved Endpoint.subscribe/1 Error Message

benwilson512:
Right now if you don't have pubsub configured at at all and attempt to use `Endpoint.subscribe` or similar, you get an error that looks like:

```
    (stdlib) :ets.lookup(nil, :subscribe)
    (phoenix_pubsub) lib/phoenix/pubsub.ex:288: Phoenix.PubSub.call/3
    (absinthe) lib/absinthe/subscription/proxy.ex:19: Absinthe.Subscription.Proxy.init/1
    (stdlib) gen_server.erl:365: :gen_server.init_it/2
    (stdlib) gen_server.erl:333: :gen_server.init_it/6
    (stdlib) proc_lib.erl:247: :proc_lib.init_p_do_apply/3
```

Separately I plan on adding a PR to Phoenix.PubSub to make that a little better, but the advantage of handling this error in Phoenix itself is that we can provide very detailed suggestions for how to recover from the error.

Note: I'm a bit unclear on the best way to test this.

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -478,46 +478,69 @@ defmodule Phoenix.Endpoint do
 
       def __pubsub_server__, do: @pubsub_server
 
+      defp __pubsub_server__! do
+        if @pubsub_server do
+          @pubsub_server
+        else
+          app_module =
+            __MODULE__
+            |> Module.split
+            |> List.first
+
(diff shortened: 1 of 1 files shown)


chrismccord:
❤️❤️❤️🐥🔥

PR #2450, opened 2017-08-21
Fix typo and capitalize HTTP and HTTPS in docs

jayjun:
HTTP and HTTPS are acronyms and should be capitalized.


chrismccord:
❤️❤️❤️🐥🔥

PR #2458, opened 2017-08-24
simple tests to check that generated .gitignore end with newline

jmarca:
Just went through and added something like:

assert_file "phx_blog/.gitignore", ~r/\n$/

every time I found .gitignore in the installer tests.

passes if and only if the new phx installer is installed locally.  Not
sure how to trigger tests to use the repo's version of the installer.


chrismccord:
❤️❤️❤️🐥🔥

PR #2459, opened 2017-08-24
Make sure sockets with unconfigured endpoints raise warning

novaugust:
This brings sockets in line with how connections and raw endpoints are treated by `Phoenix.Token.{sign,verify}`


chrismccord:
❤️❤️❤️🐥🔥

PR #2461, opened 2017-08-26
Reuse common lines in controller tests via test case

wnuqui:
This PR extracted:
- 2 `use` lines
- 1 `import` line
- and a `setup` block 

which are common in `test/phoenix/controller` tests. Extracted file is placed inside `test/support` as `Phoenix.Controller.ConnCase` test case. The `test/phoenix/controller` tests then can just do

```elixir
use Phoenix.Controller.ConnCase
```

instead of the repetitive common lines.

@chrismccord @josevalim What do you think?

### DIFF
--- a/test/phoenix/controller/controller_test.exs
+++ b/test/phoenix/controller/controller_test.exs
@@ -1,14 +1,6 @@
 defmodule Phoenix.Controller.ControllerTest do
-  use ExUnit.Case, async: true
-  use RouterHelper
-
-  import Phoenix.Controller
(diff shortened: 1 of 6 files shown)


chrismccord:
Thanks for effort cleaning up some code! In this case, I would like to avoid code injection in the phoenix test suite since the duplication isn't exhaustive. Thanks!

wnuqui:
Got, it. After a review, the duplicates indeed are not exhaustive. Thanks @chrismccord!

PR #2481, opened 2017-09-07
Generate Epub version of the docs

jeregrine:
If we want to generate epub for the docs we'll need to generate the epub version at the top level of the guides. This will generate an epub that looks sort of like the following:

screen shot 2017-09-07 at 3 35 44 pm

This should finally close up https://github.com/phoenixframework/phoenix_guides/issues/623

We will also want to link to it in the guides/site since hexdocs doesn't natively support the epub version of docs, but it will host it. The URL should be `Phoenix-v1.4.0-dev.epub` for whichever version we deploy. We'll just need to link to https://hexdocs.pm/phoenix/Phoenix-v1.4.0-dev.epub


josevalim:
Beautiful. Should we change the guides to have a reference to the epub document somewhere?

jeregrine:
@josevalim yea that makes sense.

ericmj:
Maybe the epub file should not be versioned so links dont have to be updated on every new version.

Gazler:
Isn't it better to keep the version so that people can download the epub version that is relevant to the version of Phoenix they are using?

josevalim:
@Gazler the version will already be in the hexdocs path.

Gazler:
@josevalim Of course it will! I must be all conferenced out. Looks like there isn't an option for it at the moment:

https://github.com/elixir-lang/ex_doc/blob/master/lib/ex_doc/formatter/epub.ex#L54

I'll open an issue.

Gazler:
@jeregrine Can you please remove the version and update ex_doc to 0.16.3 then we can merge this in?

jeregrine:
@Gazler 👍

Gazler:
@jeregrine I pushed on top of your commit so that the mix.lock changes only impacted ex_doc and earmark.

Thanks <3 <3 :rocket:

Gazler:
One thing about this, currently only the versioned epub link ( https://hexdocs.pm/phoenix/1.3.0/Phoenix.epub ) works. This version does not https://hexdocs.pm/phoenix/Phoenix.epub

We will either need to use an absolute URL for the version or fix it in hexpm.

jeregrine:
Could we just like to ‘Phoenix.epub’

Gazler:
@jeregrine if that was in reference to my ninja deleted comment, it turns out that hexdocs.pm/phoenix/Phoenix.epub totally works, there was just a small delay after publishing.

jeregrine:
:+1:

PR #2482, opened 2017-09-08
Add Phoenix 1.3 controller features to guides

mcelaney:
(resolves https://github.com/phoenixframework/phoenix_guides/issues/803)
Add section on halt()
Add section on action_fallback

### DIFF
--- a/guides/docs/controllers.md
+++ b/guides/docs/controllers.md
@@ -537,3 +537,119 @@ def index(conn, _params) do
   redirect conn, external: redirect_test_url(conn, :redirect_test)
 end
 ```
+
+### Halting the Plug Pipeline
+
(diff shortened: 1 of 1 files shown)


Gazler:
Thanks for the PR! The action_fallback section is perfect, I think we could do with changing the halt example though. it is not a common pattern to call halt() inside a controller action. Instead, we should do something like:

```elixir
defmodule HelloWeb.PostFinder do
  use Plug
  import Plog.Conn

  alias Hello.Blog

  def init(opts), do: opts

  def call(conn, _) do
    case Blog.get_post(conn.params["id"]) do
      {:ok, post} ->
        assign(conn, :post, post)
      {:error, :notfound} ->
        conn
        |> send_resp(404, "Not found")
        |> halt()
    end
  end
end
```

We should also put this example after the `action_fallback` code since we check the responses from Blog.get_post/1 when showing the example code, which would no longer make sense if it was in a separate plug.

mcelaney:
KK - I'll hit it this evening.

mcelaney:
@Gazler Can you have another look when you get a chance?

Gazler:
@mcelaney I totally missed this update, sorry!

Thanks for updating and thanks for the PR! <3

PR #2485, opened 2017-09-09
Provides a common view helpers module for user-define view helpers.

jung-hunsoo:
On developing with Phoenix, I found that there needs a place to define own view helpers callable inside my app. And this user-defined methods are essential for most of the web dev works as well; like full_title(), render_shared(), and more. (I use many partial chunks and they're embedded with render_shared() often.)

Though I named this module as "CommonHelpers", please open a discuss and find the best one instead.
Suffix with '**Helpers**' doesn't seem look nice since they're in "/view" folder. It's also weird to take '**View**' because this module is not intended to work for some specific templates nor it use Phoenix.View inside. Even more this '**Common**' thing should not be generated while View things are generated on task '**phx.gen.***'.

#### Here are some examples :

- GeneralHelpers
- GlobalHelpers
- CommonView
- GeneralView
- GlobalView

### DIFF
--- a/guides/docs/adding_pages.md
+++ b/guides/docs/adding_pages.md
@@ -31,6 +31,7 @@ Most of our work in this guide will be in the `lib/hello_web` directory, which h
 │   └── page
 │       └── index.html.eex
 └── views
+│   ├── common_helpers.ex
 │   ├── error_helpers.ex
 │   ├── error_view.ex
 │   ├── layout_view.ex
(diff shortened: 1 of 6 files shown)


josevalim:
The problem with such views is that it tends to become a "junk drawer". For example, the helpers that you mention may fit better under a NavigationHelper or RenderingHelper. Given we already have an example of using this pattern in Phoenix via the ErrorsHelper, I don't believe we need to another mechanism, especially one that may even potentially push people to the wrong direction.

In any case, thank you for the PR!

jung-hunsoo:
Maybe you're right. People may tend to abuse that module, and I also worried about it. I don't like dragging the issue. But there're pros and cons.

As long as the phx users are implementing frontend with phx, not a Vue, they will use some (or many) helpers; full_title(), render_shared(), active_link(), or so.

Let me show you what the devs are going on;

- https://github.com/tuvistavie/phoenix-active-link
This doc guides the users add codes to the quote of the view function in your web.ex.

- https://medium.com/@stanbright/elixir-phoenix-partials-1df87c5eaea0
This post too.

Regardless of the intended coding guidance, beginners will more likely abuse the 'web.ex' whatever reasons. Most of phoenix developers will eventually make their own common helpers, and every one of them will feel inconvenient for lack of this common module.

Explicit but not be verbose.. If there's a same job many users doing, then it'll be perfect to be provided by framework.

Thanks anyway. I'm really enjoying with phoenix!

PR #2488, opened 2017-09-12
Remove all references of mix.phoenix from guides

jeregrine:
Several documents contained `mix.phoenix` this is find all and replace.

We'll need to devote more time into the testing guides as a whole, to include contexts.

### DIFF
--- a/guides/docs/deployment/deployment.md
+++ b/guides/docs/deployment/deployment.md
@@ -31,7 +31,7 @@ use Mix.Config
 
 # You can generate a new secret by running:
 #
-#     mix phoenix.gen.secret
+#     mix phx.gen.secret
 config :foo, Foo.Endpoint,
   secret_key_base: "A LONG SECRET"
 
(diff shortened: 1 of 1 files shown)


guides/docs/phoenix_mix_tasks.md
```diff
@@ -30,9 +30,9 @@ Note that there are still other tasks under the full phoenix name:
 ```console
 ➜  mix help | grep -i phoenix
 mix local.phoenix      # Updates Phoenix locally
-mix phoenix.gen.html   # Generates controller, model and views for an HTML based resource
-mix phoenix.new        # Creates a new Phoenix v1.2.1 application
-mix phoenix.server     # Starts applications and their servers
+mix phx.gen.html   # Generates controller, model and views for an HTML based resource
+mix phx.new        # Creates a new Phoenix v1.2.1 application
```
Gazler: We should probably remove the version here so we don't need to keep updating it.

guides/docs/phoenix_mix_tasks.md
```diff
@@ -30,9 +30,9 @@ Note that there are still other tasks under the full phoenix name:
 ```console
 ➜  mix help | grep -i phoenix
 mix local.phoenix      # Updates Phoenix locally
-mix phoenix.gen.html   # Generates controller, model and views for an HTML based resource
-mix phoenix.new        # Creates a new Phoenix v1.2.1 application
-mix phoenix.server     # Starts applications and their servers
```
nekath: @jeregrine These 3 lines should not be changed :)

guides/docs/phoenix_mix_tasks.md
Gazler: Actually, we need to keep these as is, since the section states:

 >  Note that there are still other tasks under the full phoenix name:

jeregrine: Maybe we just kill the entire section because the only one that remains on my machine is `phoeinx.new` and `local.phoenix` thats because I have it laying around from previous versions.

Gazler: @jeregrine The others exist when you are in an application using Phoenix. Phoenix 1.3 supports both sets of generators, however the phoenix.* functions are gone in 1.4

Gazler:
@jeregrine ping

mcelaney:
Lemme know if you'd like a hand nudging this over the finish line...

PR #2497, opened 2017-09-13
Update channels.md typo for endpoint.ex

cschar:
In the channels [Tying it all together section](https://hexdocs.pm/phoenix/channels.html#tying-it-all-together).

User is referred to
```
 lib/hello/endpoint.ex
```
but in phoenix 1.3 , path project generator makes is:
```
lib/hello_web/endpoint.ex
```


Gazler:
Thanks <3

PR #2499, opened 2017-09-15
Default `server: true` in `prod.exs`

jesseshieh:
Hi! This pull request is really more of a question than a proposal. I'm not sure why `server: true` is not set here in the `prod.exs` template. Was this a conscious decision?

Lots of newcomers run into this issue when using Distillery/releases and I answer a lot of questions about this in the elixir-lang.slack.com #deployment channel. I can understand maybe why it isn't set for development, but for production I'm not sure.

Anyway, thanks for your consideration! If you like this pull request, I can do it for the other `prod.exs` templates too.

### DIFF
--- a/installer/templates/phx_single/config/prod.exs
+++ b/installer/templates/phx_single/config/prod.exs
@@ -15,6 +15,7 @@ use Mix.Config
 # which you typically run after static files are built.
 config :<%= web_app_name %>, <%= endpoint_module %>,
   load_from_system_env: true,
+  server: true,
   url: [host: "example.com", port: 80],
   cache_static_manifest: "priv/static/cache_manifest.json"
 


chrismccord:
It's not set for prod by default because you may want to run other tasks, for example `mix ecto.migrate` or even iex to debug prod and in those cases you won't want to spin up a phoenix server. I agree this can trip folks up when going to releases, but they will necessarily have to configure their app for building releases so it's something they'll need to become familiar with. If you think we can improve docs in this area please let us know. Thanks!

OvermindDL1:
_/me is wondering if a `release.ex` config file might be useful..._

chrismccord:
Maybe eventually once the release story is ironed out, but I don't think it's something we should ship with today as not everyone is using releases

PhillippOhlandt:
@chrismccord Maybe because it's still a very confusing thing for people new to Elixir. I just had a quick look at the Phoenix guides and it seems like there is no guide for a deployment using OTP releases.

chrismccord:
We link to the distillery deployment guide for Phoenix https://github.com/bitwalker/distillery/blob/master/docs/Phoenix%20Walkthrough.md

PhillippOhlandt:
Ah ok, haven't seen that. Good to know such a guide exists.

PR #2503, opened 2017-09-15
Use String.duplicate/2 instead.

jung-hunsoo:
An Elixir's String.duplicate/2 function can be used at no risk instead of creating our own function.


chrismccord:
❤️❤️❤️🐥🔥

PR #2504, opened 2017-09-16
Phoenix.new: Fix path for brunch invocation

gfiorav:
## Context

Running the `mix phoenix.new` command yields an error when running brunch. The cause is that the execution path for the node module is wrong. The correct path is `node_modules/brunch/.bin/brunch`, instead of `node_modules/brunch/bin/brunch` (missing `.`).

This is a small PR to fix the path so that the `brunch` executable is found and task succeeds.

### DIFF
--- a/installer/lib/mix/tasks/phx.new.ex
+++ b/installer/lib/mix/tasks/phx.new.ex
@@ -178,7 +178,7 @@ defmodule Mix.Tasks.Phx.New do
   defp switch_to_string({name, val}), do: name <> "=" <> val
 
   defp install_brunch(install?) do
-    maybe_cmd "cd assets && npm install && node node_modules/brunch/bin/brunch build",
+    maybe_cmd "cd assets && npm install && node node_modules/brunch/.bin/brunch build",
               File.exists?("assets/brunch-config.js"), install? && System.find_executable("npm")
   end
 


josevalim:
Can you please tell us your npm version? It may be that npm recently changed this, but this was surely working in the past. :)

gfiorav:
Hi @josevalim,

Thanks for looking into this.

My current version is `5.3.0`, but I've used `npm` quite a lot and I would swear this hasn't changed. I'll make some checks with different versions.

webdeb:
I am wondering, why not just add another script task into the package.json:

```
"scripts": {
  "build": "brunch build"
}
```

so running builds would be just:

`npm run build`

gfiorav:
I didn't go through with the tests. Instead I found [a 7 year old commit](https://github.com/npm/npm/blame/latest/lib/build.js#L111) that refers to using `node_modules/.bin/` structure for locally installed packages. I think it's enough evidence, but if you're not convinced say so and i'll run the task with several `npm` versions just to be sure :)

Thanks!

josevalim:
@gfiorav npm uses `.bin` at the root but you are changing the `bin` inside brunch, which does not have a dot. Hence the confusion. :)

And the reason why we use this one and the reason why we don't use scripts is because we had a bad time making .bin and scripts work on Windows.

gfiorav:
@josevalim Ah, you're right, I fired too quickly there!

The issue must have been caused by some other problem, I just tried again and can't reproduce.

Thanks again for looking into it, and sorry for the false alarm!

PR #2507, opened 2017-09-18
Added support for partial downloads

bsidoruk:
A small change  for the function `Phoenix.Controller.send_download\3`.  When calling the function with a file path, it now allows for passing in the `offset` and `length` options so they can be passed on to `send_file`.  This allows for partial content requests.  I also added a couple tests for this.  I'm open to any feedback.  Thanks.


chrismccord:
❤️❤️❤️🐥🔥

Gazler:
@chrismccord @jeregrine Just to note, now that the Phoenix API docs has changed, we will have to back-merge any guide changes for 1.3 into the 1.3 branch.

PR #2509, opened 2017-09-20
Updated module names

Grantimus9:
Updated a small number of module names to include the "Web" suffix - e.g. HelloWeb instead of Hello, so documentation matches what phx.new generates in 1.3.


chrismccord:
❤️❤️❤️🐥🔥

PR #2510, opened 2017-09-21
Update contexts.md, fix rm paths to match latest in ecto.md

ashean:
ashean hello $ rm lib/hello/user.ex
rm: lib/hello/user.ex: No such file or directory
ashean hello $ rm lib/hello/accounts/user.ex
ashean hello $ rm priv/repo/migrations/*_create_user.exs
rm: priv/repo/migrations/*_create_user.exs: No such file or directory
ashean hello $ rm priv/repo/migrations/*_create_users.exs
ashean hello $

### DIFF
--- a/guides/docs/contexts.md
+++ b/guides/docs/contexts.md
@@ -28,8 +28,8 @@ In order to run the context generators, we need to come up with a module name th
 Before we use the generators, we need to undo the changes we made in the Ecto guide, so we can give our user schema a proper home. Run these commands to undo our previous work:
 
 ```console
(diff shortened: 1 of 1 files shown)


chrismccord:
As I mentioned in the commit (https://github.com/phoenixframework/phoenix/commit/88eddbf55a8a489883e02e4265bafaa441643bc6#commitcomment-24471061) we shouldn't use the context namspacing in the Ecto go and will be reverting the commit that added it, in which case the current `rm` commands will be correct. Thanks!

PR #2511, opened 2017-09-21
Revert "add the context module name in phx.gen.schema run"

bryanstearns:
Per @chrismccord [here](https://github.com/phoenixframework/phoenix/commit/88eddbf55a8a489883e02e4265bafaa441643bc6#commitcomment-24471061): The namespacing/context ideas come later in the context guide; using them here without explanation is confusing, and adding an explanation would be premature since we tackle this later.

This reverts commit 88eddbf55a8a489883e02e4265bafaa441643bc6.


chrismccord:
❤️❤️❤️🐥🔥

PR #2512, opened 2017-09-22
Correct plug name

khaledh:
Correcting the reference to the plug that placed `current_author` in assigns.


chrismccord:
❤️❤️❤️🐥🔥

PR #2513, opened 2017-09-22
Add parameter keep strategy

blatyo:
Initial conversation here: https://groups.google.com/forum/#!topic/phoenix-core/drYwLVlnvI4

I started with `discard` and `keep` as names to get things started. I'm fine with any name you want.

There are a couple choices I made when implementing the keep strategy that are worth discussing. First, I didn't want filtering to discard the structure of the data. For example, if there is a deeply nested structure, I didn't want to specify every key in the path to the leaf data. This will make it easier to look at a log and see which key you want to add.

``` elixir
# I implemented this
filter_values(%{"foo" => %{"bar" => 1}}, ["bar"])
#=> %{"foo" => %{"bar" => 1}}

filter_values(%{"foo" => %{"bar" => 1}}, [])
#=> %{"foo" => %{"bar" => "[FILTERED]"}

# Not this
filter_values(%{"foo" => %{"bar" => 1}}, ["bar"])
#=> %{"foo" => "[FILTERED]}"

filter_values(%{"foo" => %{"bar" => 1}}, ["foo", "bar"])
#=> %{"foo" => %{"bar" => 1}}
```

Second, I used `in` to check that the exact value was in the list of params. I did this because, I didn't want keeping something like `"confirm"`, to also keep `"password_confirmation"`. This differs from the discard strategy.

### DIFF
--- a/lib/phoenix/logger.ex
+++ b/lib/phoenix/logger.ex
@@ -15,6 +15,15 @@ defmodule Phoenix.Logger do
   case sensitive.
 
   Phoenix's default is `["password"]`.
+
+  Phoenix can filter all parameters by default and selectively keep
+  parameters. This can be configured like so:
+
+      config :phoenix, :filter_parameters, {:keep, ["id", "order"]}
+
+  With the configuration above, Phoenix will filter all parameters,
+  except those that match exactly `id` or `order`. If a kept parameter
+  matches, all parameters nested under that one will also be kept.
   """
   require Logger
   import Phoenix.Controller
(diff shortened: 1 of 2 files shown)


blatyo:
Failures appear to also be present on master

lib/phoenix/logger.ex
```diff
@@ -48,22 +48,41 @@ defmodule Phoenix.Logger do
 
   @doc false
   def filter_values(values, params \\ Application.get_env(:phoenix, :filter_parameters))
-  def filter_values(%{__struct__: mod} = struct, _filter_params) when is_atom(mod) do
+  def filter_values(values, {:discard, params}), do: discard_values(values, params)
+  def filter_values(values, {:keep, params}), do: keep_values(values, params)
+  def filter_values(values, params), do: filter_values(values, {:discard, params})
```
josevalim: You can call `discard_values` directly here.

blatyo:
I've addressed your comments

chrismccord:
❤️❤️❤️🐥🔥

PR #2514, opened 2017-09-22
More accurate diff for controller actions

khaledh:
This is not correcting any mistakes, but does a couple of things to make a particular diff clearer to the reader:
- shows the previous state for some changed lines, so that it's easy to compare what's changed
- removes some parameters that are no longer necessary after the change


Gazler:
Thanks <3

PR #2515, opened 2017-09-22
Add Seamless Web Worker Support

kturney:
use `self` instead of `window` for access to globals in either web worker or window contexts


chrismccord:
❤️❤️❤️🐥🔥

PR #2520, opened 2017-09-24
Fix broken link in Ecto guide

michaelfich:
I noticed there was a link that was broken in the Phoenix Guides under the [Ecto](https://hexdocs.pm/phoenix/ecto.html) section.  I've updated that link so that it no longer leads to a 404 page.


Gazler:
Thanks <3

PR #2521, opened 2017-09-24
correct table relationship

jduan:
The relationship between posts and users should be many to one, not one to many.


chrismccord:
❤️❤️❤️🐥🔥

jduan:
Thanks for merging!

PR #2525, opened 2017-09-27
Guide Update: Move creating skeleton app

### LINKED ISSUE #2492: Don't recommend generating a skeleton app in the intro
@nisstyre56 raised this issue on the phoenix_guides

---

In the guide located at https://hexdocs.pm/phoenix/installation.html#content it recommends doing mix phx.new web --no-brunch --no-ecto which generates an application without brunch to compile static assets (like JS). It does not say why you would want to generate only a skeleton app, so I'm wondering if it might be a good idea to generate a full app instead.

My rationale is that if you follow the guide and get to the point where you want to implement a channel, it seems to assume that you've setup brunch, so either you have to manually setup static asset compilation with brunch, or regenerate your entire app, both of which are frustrating and interrupt the learning process. I think it would be better to recommend just doing mix phx.new so that the guide can be followed from start to finish without having to get sidetracked.

ksolo:
Closes #2492 

@chrismccord @Gazler @josevalim 
I removed the mention of the skeleton app from the installation page, and added a note on creating an app without ecto and combining it with no-brunch to create a skeleton app. Feedback is welcome!

### DIFF
--- a/guides/docs/introduction/installation.md
+++ b/guides/docs/introduction/installation.md
@@ -97,3 +97,7 @@ This is a Linux-only filesystem watcher that Phoenix uses for live code reloadin
 
 Linux users need to install this dependency. Please consult the [inotify-tools wiki](https://github.com/rvoicilas/inotify-tools/wiki) for distribution-specific installation instructions.
 
+### Our First Phoenix Application
+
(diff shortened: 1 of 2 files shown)


chrismccord:
Perfect! Thank you! ❤️❤️❤️🐥🔥

PR #2533, opened 2017-09-29
Enforce safe local urls with static_path

### LINKED ISSUE #2516: Enforce local uri's for static_path

alexgaribay:
Enforces local paths when using `static_path`.

Resolves #2516.

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -670,6 +670,7 @@ defmodule Phoenix.Endpoint do
       Generates a route to a static file in `priv/static`.
       """
       def static_path(path) do
+        Phoenix.Endpoint.validate_local_url(path)
         Phoenix.Config.cache(__MODULE__, :__phoenix_static__,
                              &Phoenix.Endpoint.Supervisor.static_path/1) <>
         Phoenix.Config.cache(__MODULE__, {:__phoenix_static__, path},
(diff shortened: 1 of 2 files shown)


lib/phoenix/controller.ex
```diff
@@ -405,26 +405,32 @@ defmodule Phoenix.Controller do
...
   end
-  defp validate_local_url(to), do: raise_invalid_url(to)
 
-  @spec raise_invalid_url(term()) :: no_return()
-  defp raise_invalid_url(url) do
-    raise ArgumentError, "the :to option in redirect expects a path but was #{inspect url}"
+  @invalid_local_url_chars ["\\"]
+  @doc false
+  @spec validate_local_url(String.t) :: {:ok, String.t} | {:error, :invalid_local_url | :unsafe_chars}
+  def validate_local_url("//" <> _), do: {:error, :invalid_local_url}
```
josevalim: This means this function will end-up in all controllers since they do import `Phoenix.Controller`. I would say a little bit of code duplication is fine in this case, definitely better than the coupling, just add the logic directly to Phoenix.Endpoint and we should be golden!

alexgaribay:
Code has been updated to reflect your suggestions.

chrismccord:
❤️❤️❤️🐥🔥

PR #2536, opened 2017-09-30
The connection -> a connection

fxn:
"the connection struct" is the kind of language you use for mutable structures. There is one "the" (which must be the one already in context), but "a" suggests returning that one or some other connection struct, which is of course what you normally do.


Gazler:
Thanks <3

Issue #2187, opened 2017-03-15
phoenix.js Socket reconnectTimer is not reset after disconnect()


davidye:
### Environment

* Phoenix version (mix deps): 1.2.1

### Expected behavior

The socket is not connected but is attempting to reconnect via reconnectTimer. In this state, I think it would make sense that calling socket.disconnect() resets reconnectTimer.

### Actual behavior

After calling socket.disconnect(), the socket continues trying to reconnect. I believe it is due to the following line:

https://github.com/phoenixframework/phoenix/blob/62f18c71888ce700cc86933296e619f8a9dee0be/assets/js/phoenix.js#L523

chrismccord:
Fixed on master

PR #2537, opened 2017-09-30
Add a forceDisconnect function to Socket that also resets reconnectTimer in addition to disconnecting

stawberri:
I think I might still be experiencing the issue described in #2187 in v1.3.0, and I can't find anything that was changed with `disconnect` or `reconnectTimer` to resolve this issue.

The line highlighted in that issue is probably not the one causing this problem, as the noop function that `this.conn.onclose` is set to gets overwritten by `connect` to point back to `this.onConnClose()` in the next reconnection attempt.

This line seems to be intended to stop `reconnectTimer` from being scheduled in the first place, so it seems that a good way to resolve this problem would be to reset the timer if it has already been scheduled.

---

I initially tried to fix this by resetting `reconnectTimer` inside of `disconnect`. This was a problem because `reconnectTimer` uses `disconnect` as part of its callback function, and trying to reset `reconnectTimer` inside of `disconnect` would cause `reconnectTimer` to reset itself.

Instead, I think it might be a good idea to implement a new function that resets `reconnectTimer` before calling `disconnect`, and have this one be documented so that people can find it more easily.

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -750,6 +750,17 @@ export class Socket {
   }
 
   /**
+   * Cancels pending reconnect attempts and then disconnects
+   * @param {Function} callback
+   * @param {integer} code
+   * @param {string} reason
+   */
+  forceDisconnect(callback, code, reason){
+    this.reconnectTimer.reset()
+    this.disconnect(callback, code, reason)
+  }
+
+  /**
    *
    * @param {Object} params - The params to send when connecting, for example `{user_id: userToken}`
    */


chrismccord:
Can you outline the exact behavior you're seeing so we can see if this fix is necessary? Maybe include console logs? I'm assuming you're seeing the same reconnect loop after calling disconnect? More info would be helpful. Thanks!

stawberri:
Sorry I neglected to do this earlier, but I made a fresh project to test this with and I'm not seeing my problem. I think my issue might have been a problem with something else after all.

PR #2539, opened 2017-10-01
Update docs to use groups_for_extras for guide grouping

Gazler:
Please don't merge until ExDoc 0.17 is released


josevalim:
v0.17.0 is out!

chrismccord:
Let's point ex_doc at the newly released version and this is good to merge!

PR #2540, opened 2017-10-01
Group modules using ExUnit 0.17 groupings

Gazler:
Please see this issue:
https://github.com/elixir-lang/ex_doc/issues/779
and this PR:
https://github.com/elixir-lang/elixir/pull/6621

This groups the modules in the following way:

    (not in a group)
    Phoenix
    Phoenix.Digester
    Phoenix.Logger
    Phoenix.Naming
    Phoenix.Token

    CHANNELS AND TRANSPORTS
    Phoenix.Channel
    Phoenix.Presence
    Phoenix.Socket
    Phoenix.Socket.Broadcast
    Phoenix.Socket.Message
    Phoenix.Socket.Reply
    Phoenix.Socket.Transport
    Phoenix.Transports.LongPoll
    Phoenix.Transports.Serializer
    Phoenix.Transports.WebSocket

    DEVELOPMENT & TESTING
    Phoenix.ChannelTest
    Phoenix.CodeReloader
    Phoenix.ConnTest

    ENDPOINT & REQUESTS
    Phoenix.Controller
    Phoenix.Endpoint
    Phoenix.Endpoint.CowboyHandler
    Phoenix.Endpoint.Handler
    Phoenix.Router

    PROTOCOLS
    Phoenix.Param

    VIEWS AND TEMPLATING
    Phoenix.Template
    Phoenix.Template.EExEngine
    Phoenix.Template.Engine
    Phoenix.Template.ExsEngine
    Phoenix.Template.HTML
    Phoenix.View

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -26,6 +26,7 @@ defmodule Phoenix.Mixfile do
         extra_section: "GUIDES",
         assets: "guides/docs/assets",
         formatters: ["html", "epub"],
+        groups_for_modules: groups_for_modules(),
         extras: extras(),
         groups_for_extras: groups_for_extras()
       ],
@@ -122,6 +123,54 @@ defmodule Phoenix.Mixfile do
       "Guides": ~r/guides\/docs\/[^\/]+\.md/,
       "Testing": ~r/guides\/docs\/testing\/.?/,
       "Deployment": ~r/guides\/docs\/deployment\/.?/
+    ]
+  end
+
+  defp groups_for_modules do
+    # Ungrouped Modules:
+    #
+    # Phoenix
+    # Phoenix.Channel
+    # Phoenix.Controller
+    # Phoenix.Naming
+    # Phoenix.Param
+    # Phoenix.Presence
+    # Phoenix.Router
+    # Phoenix.Token
+    # Phoenix.View
+
+    [
+      "Endpoint And Plugs": [
+        Phoenix.CodeReloader,
+        Phoenix.Endpoint,
+        Phoenix.Endpoint.CowboyHandler,
+        Phoenix.Endpoint.Handler,
+        Phoenix.Logger,
+      ],
+
+      "Socket And Transport": [
+        Phoenix.Socket,
+        Phoenix.Socket.Broadcast,
+        Phoenix.Socket.Message,
+        Phoenix.Socket.Reply,
+        Phoenix.Socket.Transport,
+        Phoenix.Transports.LongPoll,
+        Phoenix.Transports.Serializer,
+        Phoenix.Transports.WebSocket,
+      ],
+
+      "Templating": [
+        Phoenix.Template,
+        Phoenix.Template.EExEngine,
+        Phoenix.Template.Engine,
+        Phoenix.Template.ExsEngine,
+        Phoenix.Template.HTML,
+      ],
+
+      "Testing": [
+        Phoenix.ChannelTest,
+        Phoenix.ConnTest,
+      ],
     ]
   end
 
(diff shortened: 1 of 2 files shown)


mix.exs
```diff
@@ -112,8 +113,52 @@ defmodule Phoenix.Mixfile do
 
       "deployment/deployment.md": [group: "Deployment"],
       "deployment/heroku.md": [group: "Deployment"]
+      ]
+      |> Enum.map(fn {file, opts} -> {:"guides/docs/#{file}", opts} end)
+  end
+
+  defp groups_for_modules do
+    [
```
josevalim: Let's document the ones that are not in any group at the top.

mix.exs
```diff
@@ -112,8 +113,52 @@ defmodule Phoenix.Mixfile do
...
+        Phoenix.Transports.WebSocket,
+      ],
+
+      "Development & Testing": [
+        Phoenix.ChannelTest,
+        Phoenix.CodeReloader,
+        Phoenix.ConnTest,
+      ],
+
+      "Protocols": [
```
josevalim: I would probably kill this group.

mix.exs
josevalim: Speaking about this group, I would `@moduledoc false` the Phoenix.Digester. Most interact with it through the task.

josevalim:
@Gazler this looks great. I am wondering though if it would be best to keep the "big names" outside of any group and keep them easily accessible. For example:

 * Channel and presence are out (we can rename the group to Sockets & Transport)
 * Phoenix.View is out (we can rename the group to Templating only)
 * Phoenix.Controller and Phoenix.Router are out (we can rename the group to endpoint)

This way the main concepts are quickly accessible and users can scroll down to dig deeper.

josevalim:
This looks great to me! I would just have a comment on the top of the groupings listing all modules that do not belong to any group and then :shipit:!

Gazler:
@josevalim Done. Thanks for the requirement to resync btw! ;)

mix.exs
```diff
@@ -125,6 +126,54 @@ defmodule Phoenix.Mixfile do
     ]
   end
 
+  defp groups_for_modules do
+    # Ungrouped Modules:
+    #
+    # Phoenix
+    # Phoenix.Channel
+    # Phoenix.Controller
+    # Phoenix.Logger
```
josevalim: What do you think about this:

Have an "Endpoint and Plugs" category, move Phoenix.CodeReloader and Phoenix.Logger in there. Rename the "Development and Testing" to testing.

The reason for this is because Phoenix.Logger is really not at the same level as these.

Thoughts?

josevalim:
> @josevalim Done. Thanks for making me rebase btw! 😄

No problem. Glad to keep your git skills in check.

mix.exs
Gazler: I don't think it would be immediately obvious that they would be in the "Endpoint and Plugs" section without prior knowledge that those modules are plugs. However, I think it is very likely that the users who want the docs for those modules will have that prior knowledge.

Gazler:
> > @josevalim Done. Thanks for making me rebase btw! :smile:
> No problem. Glad to keep your git skills in check.

Just for that I've added a bonus commit https://github.com/Gazler/phoenix/commit/a3a2a79c18e91a4b79cfcd81b4cedeca9b7d83e8

mix.exs
josevalim: Well, to help with that, those two modules are used in the endpoint as `plug Plug.CodeReloader` and `plug Phoenix.Logger`. So I think they do fit all together. :)

Gazler:
Had to manually make changes to deploy this to hex since the guides were merged in after 1.3 was released - https://hexdocs.pm/phoenix/overview.html

PR #2542, opened 2017-10-03
Update command in contexts.md docs.

faizmokhtar:
The `user.ex` file created from the previous Ecto guide is inside `lib/hello/account/` folder instead of `lib/hello/`.

This PR change it to delete the `accounts` directory instead.


chrismccord:
We've since updated the Ecto guide to not use the accounts dir so this is no longer needed. Thanks!

faizmokhtar:
Alright. noted 👌

Issue #2469, opened 2017-08-31
Deprecate passing a view in render/3 and render/4 in favor of put_view


josevalim:

michalmuskala:
It should also document using `plug :put_view, View` to change the implicit view used in the whole controller - it's very useful but I don't think it's very popular. I only realised you could do that after reading through the code for how phoenix sets the default view when you `use Phoenix.Controller`.

PhillippOhlandt:
Yeah, the automatic view selection confused the hell out of me when I started using Phoenix. I just accepted that it works how it works now.

jeregrine:
Whomever takes this on should also update the guides, should be a fairly easy find and replace for render. The trickier part will be explaining the new put_view, and removing reference to passing a View to render.

evuez:
I'd love to tackle this one, is it OK if I do a PR with a work in progress? Cause I think it's gonna take me some time given the number of places these functions are referenced.

josevalim:
@evuez yes, please go ahead. :)

stevedomin:
Just curious, what was the reasoning behind this? No strong feeling either way but I'm trying to understand why it was deprecated.

josevalim:
@stevedomin I think our clauses were laid out in a way where you could pass a template where a view was expected or vice-versa because we were matching on the type and that can be a security issue if people are not careful. I mean, people would have to already be doing something wrong, but better to be safer and keep the API simple.

stevedomin:
Makes a lot of sense. Thanks @josevalim

PR #2544, opened 2017-10-04
Deprecate render/3 and render/4 in favor of put_view/2

evuez:
Encourage the use of `put_view/2` and `plug :put_view, View` instead.

See #2469.

### DIFF
--- a/guides/docs/controllers.md
+++ b/guides/docs/controllers.md
@@ -452,7 +452,8 @@ The correct way to render the 404 page from `HelloWeb.PageController` is:
 def index(conn, _params) do
   conn
   |> put_status(:not_found)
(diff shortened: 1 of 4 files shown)


guides/docs/controllers.md
```diff
@@ -452,7 +452,8 @@ The correct way to render the 404 page from `HelloWeb.PageController` is:
 def index(conn, _params) do
   conn
   |> put_status(:not_found)
-  |> render(HelloWeb.ErrorView, "404.html")
+  |> put_view(HelloWeb.ErrorView)
```
evuez: Should we add a "Setting the view" section to introduce `put_view` and `plug :put_view, View` before using it?

chrismccord:
❤️❤️❤️🐥🔥

PR #2547, opened 2017-10-04
Extract Phoenix.Socket.Transport.dispatch to separate module

spscream:
Hi,
I think what extracting of phoenix transport protocol to separate module would give some benefits - e.g. If you mind you could have your own realisation of it and make some custom logic on top of it later. So I extracted responsibility of Phoenix.Transport.dispatch to separate module. And also this extraction separates logic of handling transport itself from handling events and channels logic.

### DIFF
--- a/lib/phoenix/socket/transport.ex
+++ b/lib/phoenix/socket/transport.ex
@@ -120,8 +120,7 @@ defmodule Phoenix.Socket.Transport do
 
   require Logger
   alias Phoenix.Socket
-  alias Phoenix.Socket.Message
-  alias Phoenix.Socket.Reply
+  alias Phoenix.Socket.TransportDispatcher
 
   @protocol_version "2.0.0"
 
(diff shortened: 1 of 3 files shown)


josevalim:
@spscream it is unclear how extracting it to a module would make constructs that was not possible before possible. They are still referenced as a module and as a function, what are we gaining here then? Thanks.

spscream:
Simple example, If you want to implement some command event, which will join you to multiple channels at once - "phx_join_all", how would you implement it without forking phoenix?

To make this possible we could add some custom dispatcher, which we could set in config. In this implementation I haven't made this configurable, but now this is more readable and clear what is the transport part and what is dispatching.

I'm interested in your opinion, should I continue to work on this to make it configurable, or it doesn't worth it.

josevalim:
I would like to see how a configuration option would look like because I
still can't see why the configuration option can't still default to the
current dispatch implementation.

Also, we have old plans to refactor the transport and doing so may make
some of those needs obsolete.
-- 


spscream:
Ok, I see your point, I'm closing this and would try to make some example later, to show what I want to achieve. Thanks

JayDadhania:
> Simple example, If you want to implement some command event, which will join you to multiple channels at once - "phx_join_all", how would you implement it without forking phoenix?

@spscream I've been trying to figure out how to do this but haven't had any luck. Is there a way to achieve this without forking in the current version.

spscream:
@JayDadhania we implemented custom transport for websocket. I made an example project here: https://github.com/spscream/autojoin_example 

I'm not sure if project works and starts succesfully - I copy and paste code from our project and removed some parts not related to autojoins, but idea is clear - to simulate receive of phx_join in transport by the list of autojoin topics and change state of socket accodringly.

PR #2554, opened 2017-10-07
Fixed errors in the documentation in the presence.ex template and gen.presence file

riverrun:
The main changes are in the template where the Presence module should have a `Web` suffix and there should be parentheses after `self`.


chrismccord:
❤️❤️❤️🐥🔥

PR #2555, opened 2017-10-07
use new elixir "not in" syntax in docs

atomkirk:


josevalim:

chrismccord:
Does `left not in right` compile on <= Elixir 1.4? We may consider keeping the docs as is until we go Elixir 1.5+. I wasn't aware of this being valid syntax at all before I saw this PR :)

atomkirk:
I think I read in the update nots that `not action in` is deprecated, so I thought it was a good idea to get people using the new syntax. But you're right, maybe we could wait a bit to start demonstrating it in the docs. For someone not following along, using the syntax in the docs could give a confusing error.

PR #2561, opened 2017-10-14
HTML options for generated error_tag

josephan:
I found it helpful to be able to customize error_tag similar to other html tag helpers.


josevalim:
You are free to do those in your app. We keep it simple on purpose as each app will have different needs. Thanks!

PR #2572, opened 2017-10-17
Fix vertical alignment of `mix phx.routes` sample output in routing guide

phoffer:
The output of `mix phx.routes` is vertically aligned, but one example in the routing guide lost its alignment. This updates it to match the command output.


Gazler:
Thanks <3

PR #2574, opened 2017-10-18
Fix all misaligned code blocks in routes guide

phoffer:
Quite a few of the sample outputs of `mix phx.routes` lost some whitespace, and thus their vertical alignment. All of the updated ones are taken directly from console output.

P.S. Apologies for not fixing all of these in my prior PR. I just happened to pause at that spot in the guide for the night and didn't check the rest of the page.


Gazler:
Thanks <3

PR #2576, opened 2017-10-18
Rule out race by randomizing tmp test dirs

chrismccord:
This was a shot in the dark, but it has resolved our build issues. Shrug 🍻

### DIFF
--- a/installer/test/mix_helper.exs
+++ b/installer/test/mix_helper.exs
@@ -16,36 +16,58 @@ defmodule MixHelper do
     Path.expand("../../tmp", __DIR__)
   end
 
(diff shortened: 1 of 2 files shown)


OvermindDL1:
Why not just use the erlang monotonically incrementing integer?  That way it is always unique and still no chance (little though it may be) of a random clash?

chrismccord:
My hunch was we have races across different builds in the travis build matrix, not within the test suite, so System.unique_integer wouldn't be unique across VM's. Unique int would indeed work to guarantee uniqueness across local tests though, but we ruled out races in the test suite itself. Thanks for the idea!

PR #2577, opened 2017-10-20
guides/contexts: fix path to accounts/user.ex

buhman:
The `hello` path component is missing, yet the path was being referenced relative to `lib/`.

---

Files in other paragraphs are relative to `accounts/`; I'd be happy to change this that way instead, but it isn't consistent through the entire file, so I kept it as is.


Gazler:
Thanks <3

PR #2578, opened 2017-10-20
guides/contexts: add missing edit/2 signature change

buhman:
The generated code includes an `id` pattern match, which is implicitly removed in the diff. Add the actual change for improved clarity.


Gazler:
Thanks <3

PR #2581, opened 2017-10-20
Corrected commands to reset changes made in Ecto guide

Ernmodern:
The Ecto guide command generates the user.ex in an accounts directory and migration with *_created_users.exs (plural users).


chrismccord:
We have since changed the Ecto guide to not generate the files inside an `accounts` dir, so this fix is no longer necessary. Thanks!

PR #2588, opened 2017-10-21
An attemt at a description for templates

jish:
I'm not sure if this is the best description ever, but it is a bit more informative.

### DIFF
--- a/guides/docs/introduction/overview.md
+++ b/guides/docs/introduction/overview.md
@@ -35,7 +35,8 @@ Phoenix is made up of a number of distinct parts, each with its own purpose and 
     - act as a presentation layer
     - define helper functions, available in templates, to decorate data for presentation
  - [Templates](templates.html)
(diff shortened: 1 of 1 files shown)


guides/docs/introduction/overview.md
```diff
@@ -35,7 +35,8 @@ Phoenix is made up of a number of distinct parts, each with its own purpose and
     - act as a presentation layer
     - define helper functions, available in templates, to decorate data for presentation
  - [Templates](templates.html)
-    - are what they sound like :)
+    - files that will eventually be served in a response
```
Gazler: Could we change this to:

> files containing the contents that will be served in a response

I this helps differentiate between a template and something like Plug.Static

jish:
@Gazler I updated the copy with your suggestion.

Gazler:
Thanks <3

PR #2594, opened 2017-10-23
Guides (Controllers): Prepend params with underscore in examples

bolducp:
For functions in the code examples where params are not used, prepend `params` parameter with an underscore.

P.S. Thanks very much for putting together such excellent guides as part of the documentation; I've found them to be tremendously helpful!


chrismccord:
❤️❤️❤️🐥🔥

PR #2595, opened 2017-10-24
Fix typo

davvidbaker:
authorized_page -> authorize_page

FYI, there was only one place that had an errant "d", but the diff is showing the whole line/paragraph.


chrismccord:
❤️❤️❤️🐥🔥

PR #2597, opened 2017-10-24
Update docs for mix tasks containing full phoenix name

seanmadi:
Remove `mix phoenix.gen.html` and `mix phoenix.server`, and change `mix
phoenix.new` to say "Creates a new Phoenix v1.3.0 application".

Freshly installed Phoenix on two machines to confirm that `mix phoenix.gen.html` and `mix phoenix.server` no longer come up as mix tasks, and `mix phoenix.new` lists 1.3.0 as the version instead of 1.2.1. Confirmed that running `mix phoenix.new` creates a v1.3.0-rc project.

### DIFF
--- a/guides/docs/phoenix_mix_tasks.md
+++ b/guides/docs/phoenix_mix_tasks.md
@@ -31,7 +31,7 @@ Note that there are still other tasks under the full phoenix name:
 ➜  mix help | grep -i phoenix
 mix local.phoenix      # Updates Phoenix locally
 mix phoenix.gen.html   # Generates controller, model and views for an HTML based resource
-mix phoenix.new        # Creates a new Phoenix v1.2.1 application
+mix phoenix.new        # Creates a new Phoenix v1.3.0 application
 mix phoenix.server     # Starts applications and their servers
 ```
 


chrismccord:
The gen.html and server tasks are still there. Can you update the PR to change the original source to `s/phoenix/phx` ? Thanks!

seanmadi:
Apologies about removing `phoenix.gen.html` and `phoenix.server`, for some reason when running mix I only get the following on the two machines I am running:

image

As for replacing `phoenix` with `phx`, the phx commands are already listed, but above the section that I edited in the PR (the docs list grepping for `phx` and then `phoenix`).

chrismccord:
The phx_new archive only contains the `phx.new` tasks, but inside an existing phoenix application, the generator tasks will be available.

seanmadi:
Ah, that was it. Sorry, I'm still new to Elixir/Phoenix. Should I change the PR to just change the version on `mix phoenix.new` to from 1.2.1 to 1.3.0 in the docs?

chrismccord:
No worries :)

You can change the original code to s/phoenix/phx and make sure the version is bumped to 1.3 as well. Thanks!

seanmadi:
I have changed the commit to only change the version. I think doing the phoenix/phx replace in the docs is unnecessary because the phx section is above the phoenix that I edited, but I may be misunderstanding you. Apologies if I am.

chrismccord:
❤️❤️❤️🐥🔥

PR #2599, opened 2017-10-24
Guides (Contexts): Fix commands for undoing Ecto Guide changes

bolducp:
I'm currently working through the guides, and I noticed that the commands listed for undoing the changes from the Ecto guide in order to work with the generators in the Contexts guide are slightly off. 

The commands are meant to undo the work of this command from the Ecto Guide (`mix phx.gen.html Accounts User users name:string username:string:unique`), also shown in the screenshot here:

<a href="https://hexdocs.pm/phoenix/ecto.html#content">
screen shot 2017-10-24 at 2 15 39 pm</a>

But the commands currently given in the Context guide don't quite delete the correct files. I wonder if the Ecto guide was perhaps changed at some point, but these commands weren't updated to match. I think this PR should fix the discrepancy.

### DIFF
--- a/guides/docs/contexts.md
+++ b/guides/docs/contexts.md
@@ -28,8 +28,8 @@ In order to run the context generators, we need to come up with a module name th
 Before we use the generators, we need to undo the changes we made in the Ecto guide, so we can give our user schema a proper home. Run these commands to undo our previous work:
 
 ```console
-$ rm lib/hello/user.ex
-$ rm priv/repo/migrations/*_create_user.exs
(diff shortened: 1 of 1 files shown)


chrismccord:
Thanks! This has since been fixed so that the Ecto guides do no place the code under the accounts dir, which makes the current guide correct (not yet published to hex docs). Thanks!

bolducp:
Ah, I see. Thanks for the quick response! I'll check for changes not yet published to hex first next time. 👍

PR #2602, opened 2017-10-25
Return mutated connection

### LINKED ISSUE #2601: Websocket returns multiple http responses (403 and 204) when socket is refused.
### Environment

* Elixir version (elixir -v):1.5.1
* NodeJS version (node -v):
* NPM version (npm -v):
* Operating system:

### Expected behavior
When a socket's connect/2 method returns :error, only a single 403 response should be sent to the client.

### Actual behavior
After the 403, an additional 204 is sent over the tcp connection. 

I have located the problem and will send a pull request which fixes the issue.

markmeeus:
Fixes #2601

By not returning the connection returned by send_resp, cowboy is unable to determine whether a response was allready sent, and send another 204 over the tcp connection.

### DIFF
--- a/lib/phoenix/transports/websocket.ex
+++ b/lib/phoenix/transports/websocket.ex
@@ -90,7 +90,7 @@ defmodule Phoenix.Transports.WebSocket do
           {:ok, socket} ->
             {:ok, conn, {__MODULE__, {socket, opts}}}
           :error ->
-            send_resp(conn, 403, "")
+            conn = send_resp(conn, 403, "")
             {:error, conn}
         end
       %{halted: true} = conn ->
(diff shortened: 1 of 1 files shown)


markmeeus:
The websockets are tested with an actual websocket client, which seems to ignore the additional 204. (As most clients do)
I'm not sure how to approach testing this in an elegant way.

josevalim:
Yup, we understand this is hard to test unless we drop down to TCP, so this looks good as is IMO. Thanks!

markmeeus:
The build seems to fail for elixir 1.5, OTP 20. 
It looks like logging a failed request is broken,  could this be related to my change?

josevalim:
@markmeeus it seems to be rather a race condition between the request process, the logger process and the test process and not related to your PR.

chrismccord:
❤️❤️❤️🐥🔥

PR #2603, opened 2017-10-25
Guides (Contexts): Update docs to reflect correct cast_assoc function arity

bolducp:
I think that `Ecto.Changeset.cast_assoc/2` should read `Ecto.Changeset.cast_assoc/3`, referencing [this Ecto function](https://hexdocs.pm/ecto/Ecto.Changeset.html#cast_assoc/3) shown as `cast_assoc/3` in the Ecto docs.

Note: the diff highlights the whole paragraph, but there are only two changes of one char each.


chrismccord:
❤️❤️❤️🐥🔥

PR #2610, opened 2017-10-27
Add a npm brunch run command and change output in the installer for that

ducky427:


Gazler:
Please see https://github.com/phoenixframework/phoenix/pull/2504

ducky427:
ah ok.

tl;dr - npm scripts don't work well on Windows. Fair enough.

PR #2614, opened 2017-10-31
Add controller and view generators

sebastiandedeyne:
I often want to quickly create a controller or view, not necessarily tied to an entire context/model.

Right now I copy an existing controller/view, and strip it down to a new one. These generators look pretty small, but imo it would be a great workflow improvement to be able to generate these from the command line.

If there's no interest in maintaining this in core, I'll probably move these to an `extra_generators` package or something.

If this gets merged, I'll whip up a docs PR too :)

### DIFF
--- a/lib/mix/tasks/phx.gen.controller.ex
+++ b/lib/mix/tasks/phx.gen.controller.ex
@@ -1,0 +1,64 @@
+defmodule Mix.Tasks.Phx.Gen.Controller do
+  @shortdoc "Generates a Phoenix controller"
+
+  @moduledoc """
+  Generates a Phoenix controller.
+
+      mix phx.gen.controller Posts
+
+  Accepts the module name for the controller
+
+  The generated files will contain:
+
+  For a regular application:
+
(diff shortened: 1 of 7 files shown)


chrismccord:
I'm not convinced this belongs as a feature in core but big :+1: on a separate package. Thanks!

OvermindDL1:
> Right now I copy an existing controller/view, and strip it down to a new one. These generators look pretty small, but imo it would be a great workflow improvement to be able to generate these from the command line.

I do the same copy-cleaning quite routinely.  It would be convenient to get a generator to generate controllers/views (and perhaps Drab commanders and/or other basic things) along with the corresponding test files via a single command.  Those would be generators that I would actually use unlike the current ones so I definitely vote for distributing them as a package and post it to the elixir forums.  :-)

sebastiandedeyne:
In case anyone's interested, here's the package containing these two generators:

https://github.com/sebastiandedeyne/phoenix_extra_generators

I'm open for adding more in the future!

PR #2621, opened 2017-11-03
Cowboy 2 support

Gazler:
I need to tidy up the commits, but all the tests are passing.

It uses cowboy master because we need a fix that will be deployed with 2.1

Reviews appreciated @chrismccord @josevalim 

I'd recommend reviewing the commits individually so you can see the diff between cowboy and cowboy 2.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -3,21 +3,57 @@ matrix:
   include:
     - otp_release: 18.3
       elixir: 1.4
+      env:
+       - COWBOY_VERSION=1.0 ELIXIR_ASSERT_TIMEOUT=2000
     - otp_release: 19.3
       elixir: 1.4
+      env:
+       - COWBOY_VERSION=1.0 ELIXIR_ASSERT_TIMEOUT=2000
+    - otp_release: 19.3
+      elixir: 1.4
+      env:
+       - COWBOY_VERSION=2.0 ELIXIR_ASSERT_TIMEOUT=2000
     - otp_release: 20.0
       elixir: 1.4
+      env:
+       - COWBOY_VERSION=1.0 ELIXIR_ASSERT_TIMEOUT=2000
+    - otp_release: 20.0
+      elixir: 1.4
+      env:
+       - COWBOY_VERSION=2.0 ELIXIR_ASSERT_TIMEOUT=2000
     - otp_release: 20.1
       elixir: 1.4
+      env:
+       - COWBOY_VERSION=1.0 ELIXIR_ASSERT_TIMEOUT=2000
+    - otp_release: 20.1
+      elixir: 1.4
+      env:
+       - COWBOY_VERSION=2.0 ELIXIR_ASSERT_TIMEOUT=2000
     - otp_release: 19.3
       elixir: 1.5
+      env:
+       - COWBOY_VERSION=1.0 ELIXIR_ASSERT_TIMEOUT=2000
+    - otp_release: 19.3
+      elixir: 1.5
+      env:
+       - COWBOY_VERSION=2.0 ELIXIR_ASSERT_TIMEOUT=2000
     - otp_release: 20.0
       elixir: 1.5
+      env:
+       - COWBOY_VERSION=1.0 ELIXIR_ASSERT_TIMEOUT=2000
+    - otp_release: 20.0
+      elixir: 1.5
+      env:
+       - COWBOY_VERSION=2.0 ELIXIR_ASSERT_TIMEOUT=2000
     - otp_release: 20.1
       elixir: 1.5
+      env:
+       - COWBOY_VERSION=1.0 ELIXIR_ASSERT_TIMEOUT=2000
+    - otp_release: 20.1
+      elixir: 1.5
+      env:
+       - COWBOY_VERSION=2.0 ELIXIR_ASSERT_TIMEOUT=2000
 sudo: false
-env:
-  - ELIXIR_ASSERT_TIMEOUT=2000
 before_script:
   - mix deps.get
   - nvm install 6.2 && nvm use 6.2
--- a/lib/phoenix/endpoint/cowboy2_handler.ex
+++ b/lib/phoenix/endpoint/cowboy2_handler.ex
@@ -47,6 +47,7 @@ defmodule Phoenix.Endpoint.Cowboy2Handler do
   It is also important to specify your handlers first, otherwise
   Phoenix will intercept the requests before they get to your handler.
   """
+
   @behaviour Phoenix.Endpoint.Handler
   require Logger
 
@@ -73,12 +74,20 @@ defmodule Phoenix.Endpoint.Cowboy2Handler do
     # Use put_new to allow custom dispatches
     config = Keyword.put_new(config, :dispatch, [{:_, dispatches}])
 
-    {ref, mfa, type, timeout, kind, modules} =
-      Plug.Adapters.Cowboy2.child_spec(scheme, endpoint, [], config)
+    cowboy_supervisor =
+      Plug.Adapters.Cowboy2.child_spec(scheme: scheme, plug: {endpoint, []}, options: config)
 
-    # Rewrite MFA for proper error reporting
-    mfa = {__MODULE__, :start_link, [scheme, endpoint, mfa]}
-    {ref, mfa, type, timeout, kind, modules}
+    %{
+      start: start,
+      id: id,
+      type: type,
+      shutdown: shutdown,
+      restart: restart,
+      modules: modules,
+    } = cowboy_supervisor
+
+    start = {__MODULE__, :start_link, [scheme, endpoint, start]}
+    {id, start, restart, shutdown, type, modules}
   end
 
   defp default_for(Phoenix.Transports.LongPoll), do: Plug.Adapters.Cowboy2.Handler
--- a/lib/phoenix/endpoint/cowboy2_handler.ex
+++ b/lib/phoenix/endpoint/cowboy2_handler.ex
@@ -1,0 +1,113 @@
+defmodule Phoenix.Endpoint.Cowboy2Handler do
+  @moduledoc """
+  The Cowboy2 adapter for Phoenix.
+
+  It implements the required `child_spec/3` function as well
+  as the handler for the WebSocket transport.
+
+  ## Custom dispatch options
+
+  You can provide custom dispatch options in order to use Phoenix's
+  builtin Cowboy server with custom handlers. For example, to handle
+  raw WebSockets [as shown in Cowboy's docs](https://github.com/ninenines/cowboy/tree/2.0.x/examples)).
+
+  The options are passed to both `:http` and `:https` keys in the
+  endpoint configuration. However, once you pass your custom dispatch
+  options, you will need to manually wire all Phoenix endpoints,
+  including the socket transports.
+
+  You will need the following rules:
+
+    * Per websocket transport:
+
+          {"/socket/websocket", Phoenix.Endpoint.Cowboy2WebSocket,
+            {Phoenix.Transports.WebSocket,
+              {MyApp.Endpoint, MyApp.UserSocket, :websocket}}}
+
+    * Per longpoll transport:
+
+          {"/socket/long_poll", Plug.Adapters.Cowboy2.Handler,
+            {Phoenix.Transports.LongPoll,
+              {MyApp.Endpoint, MyApp.UserSocket, :longpoll}}}
+
+    * For the endpoint:
+
+          {:_, Plug.Adapters.Cowboy2.Handler, {MyApp.Endpoint, []}}
+
+  For example:
+
+      config :myapp, MyApp.Endpoint,
+        http: [dispatch: [
+                {:_, [
+                    {"/foo", MyApp.CustomHandler, []},
+                    {"/bar", MyApp.AnotherHandler, []},
+                    {:_, Plug.Adapters.Cowboy2.Handler, {MyApp.Endpoint, []}}
+                  ]}]]
+
+  It is also important to specify your handlers first, otherwise
+  Phoenix will intercept the requests before they get to your handler.
+  """
+  @behaviour Phoenix.Endpoint.Handler
+  require Logger
+
+  @doc """
+  Generates a childspec to be used in the supervision tree.
+  """
+  def child_spec(scheme, endpoint, config) do
+    if scheme == :https do
+      Application.ensure_all_started(:ssl)
+    end
+
+    dispatches =
+      for {path, socket} <- endpoint.__sockets__,
+          {transport, {module, config}} <- socket.__transports__,
+          # Allow handlers to be configured at the transport level
+          handler = config[:cowboy] || default_for(module),
+          do: {Path.join(path, Atom.to_string(transport)),
+               handler,
+               {module, {endpoint, socket, transport}}}
+
+    dispatches =
+      dispatches ++ [{:_, Plug.Adapters.Cowboy2.Handler, {endpoint, []}}]
+
+    # Use put_new to allow custom dispatches
+    config = Keyword.put_new(config, :dispatch, [{:_, dispatches}])
+
+    {ref, mfa, type, timeout, kind, modules} =
+      Plug.Adapters.Cowboy2.child_spec(scheme, endpoint, [], config)
+
+    # Rewrite MFA for proper error reporting
+    mfa = {__MODULE__, :start_link, [scheme, endpoint, mfa]}
+    {ref, mfa, type, timeout, kind, modules}
+  end
+
+  defp default_for(Phoenix.Transports.LongPoll), do: Plug.Adapters.Cowboy2.Handler
+  defp default_for(Phoenix.Transports.WebSocket), do: Phoenix.Endpoint.Cowboy2WebSocket
+  defp default_for(_), do: nil
+
+  @doc """
+  Callback to start the Cowboy2 endpoint.
+  """
+  def start_link(scheme, endpoint, {m, f, [ref | _] = a}) do
+    # ref is used by Ranch to identify its listeners, defaulting
+    # to plug.HTTP and plug.HTTPS and overridable by users.
+    case apply(m, f, a) do
+      {:ok, pid} ->
+        Logger.info info(scheme, endpoint, ref)
+        {:ok, pid}
+
+      {:error, {:shutdown, {_, _, {{_, {:error, :eaddrinuse}}, _}}}} = error ->
+        Logger.error [info(scheme, endpoint, ref), " failed, port already in use"]
+        error
+
+      {:error, _} = error ->
+        error
+    end
+  end
+
+  defp info(scheme, endpoint, ref) do
+    {addr, port} = :ranch.get_addr(ref)
+    addr_str = :inet.ntoa(addr)
+    "Running #{inspect endpoint} with Cowboy2 using #{scheme}://#{addr_str}:#{port}"
+  end
+end
--- a/lib/phoenix/endpoint/cowboy2_websocket.ex
+++ b/lib/phoenix/endpoint/cowboy2_websocket.ex
@@ -1,0 +1,117 @@
+defmodule Phoenix.Endpoint.Cowboy2WebSocket do
+  # Implementation of the WebSocket transport for Cowboy2.
+  @moduledoc false
+
+  @behaviour :cowboy_websocket_handler
+  @connection Plug.Adapters.Cowboy2.Conn
+  @already_sent {:plug_conn, :sent}
+
+  def init({transport, :http}, req, {module, opts}) when transport in [:tcp, :ssl] do
+    conn = @connection.conn(req, transport)
+    try do
+      case module.init(conn, opts) do
+        {:ok, %{adapter: {@connection, req}}, args} ->
+          {:upgrade, :protocol, __MODULE__, req, args}
+        {:error, %{adapter: {@connection, req}}} ->
+          {:shutdown, req, :no_state}
+      end
+    catch
+      kind, reason ->
+        # Although we are not performing a call, we are using the call
+        # function for now so it is properly handled in error reports.
+        mfa = {module, :call, [conn, opts]}
+        {:upgrade, :protocol, __MODULE__, req, {:error, mfa, kind, reason, System.stacktrace}}
+    after
+      receive do
+        @already_sent -> :ok
+      after
+        0 -> :ok
+      end
+    end
+  end
+
+  def upgrade(req, env, __MODULE__, {handler, opts}) do
+    args = [req, env, __MODULE__, {handler, opts}]
+    resume(:cowboy_websocket, :upgrade, args)
+  end
+
+  def upgrade(_req, _env, __MODULE__, {:error, mfa, kind, reason, stack}) do
+    reason = format_reason(kind, reason, stack)
+    exit({reason, mfa})
+  end
+
+  def terminate(_reason,  _req, _state) do
+    :ok
+  end
+
+  def resume(module, fun, args) do
+    try do
+      apply(module, fun, args)
+    catch
+      kind, [{:reason, reason}, {:mfa, _mfa}, {:stacktrace, stack} | _rest] ->
+        reason = format_reason(kind, reason, stack)
+        exit({reason, {__MODULE__, :resume, []}})
+    else
+      {:suspend, module, fun, args} ->
+        {:suspend, __MODULE__, :resume, [module, fun, args]}
+      _ ->
+        # We are forcing a shutdown exit because we want to make
+        # sure all transports exits with reason shutdown to guarantee
+        # all channels are closed.
+        exit(:shutdown)
+    end
+  end
+
+  defp format_reason(:exit, reason, _), do: reason
+  defp format_reason(:throw, reason, stack), do: {{:nocatch, reason}, stack}
+  defp format_reason(:error, reason, stack), do: {reason, stack}
+
+  ## Websocket callbacks
+
+  def websocket_init(_transport, req, {handler, args}) do
+    {:ok, state, timeout} = handler.ws_init(args)
+    {:ok, :cowboy_req.compact(req), {handler, state}, timeout}
+  end
+
+  def websocket_handle({opcode = :text, payload}, req, {handler, state}) do
+    handle_reply req, handler, handler.ws_handle(opcode, payload, state)
+  end
+  def websocket_handle({opcode = :binary, payload}, req, {handler, state}) do
+    handle_reply req, handler, handler.ws_handle(opcode, payload, state)
+  end
+  def websocket_handle(_other, req, {handler, state}) do
+    {:ok, req, {handler, state}}
+  end
+
+  def websocket_info(message, req, {handler, state}) do
+    handle_reply req, handler, handler.ws_info(message, state)
+  end
+
+  def websocket_terminate({:error, :closed}, _req, {handler, state}) do
+    handler.ws_close(state)
+    :ok
+  end
+  def websocket_terminate({:remote, :closed}, _req, {handler, state}) do
+    handler.ws_close(state)
+    :ok
+  end
+  def websocket_terminate({:remote, code, _}, _req, {handler, state})
+      when code in 1000..1003 or code in 1005..1011 or code == 1015 do
+    handler.ws_close(state)
+    :ok
+  end
+  def websocket_terminate(reason, _req, {handler, state}) do
+    handler.ws_terminate(reason, state)
+    :ok
+  end
+
+  defp handle_reply(req, handler, {:shutdown, new_state}) do
+    {:shutdown, req, {handler, new_state}}
+  end
+  defp handle_reply(req, handler, {:ok, new_state}) do
+    {:ok, req, {handler, new_state}}
+  end
+  defp handle_reply(req, handler, {:reply, {opcode, payload}, new_state}) do
+    {:reply, {opcode, payload}, req, {handler, new_state}}
+  end
+end
--- a/lib/phoenix/endpoint/cowboy2_websocket.ex
+++ b/lib/phoenix/endpoint/cowboy2_websocket.ex
@@ -1,26 +1,25 @@
 defmodule Phoenix.Endpoint.Cowboy2WebSocket do
-  # Implementation of the WebSocket transport for Cowboy2.
+  # Implementation of the WebSocket transport for Cowboy.
   @moduledoc false
 
-  @behaviour :cowboy_websocket_handler
+  @behaviour :cowboy_websocket
   @connection Plug.Adapters.Cowboy2.Conn
   @already_sent {:plug_conn, :sent}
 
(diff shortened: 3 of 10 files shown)


lib/phoenix/endpoint/handler.ex
```diff
@@ -32,6 +32,7 @@ defmodule Phoenix.Endpoint.Handler do
           config = endpoint.config(scheme) do
         handler.child_spec(scheme, endpoint, default(config, otp_app, port))
       end
+
```
Gazler: :scissors:

.travis.yml
```diff
@@ -3,21 +3,57 @@ matrix:
   include:
     - otp_release: 18.3
       elixir: 1.4
+      env:
+       - COWBOY_VERSION=1.0 ELIXIR_ASSERT_TIMEOUT=2000
```
Gazler: If anyone knows a better way of doing this, I'm all ears. We don't want the cowboy 2 tests running on OTP 18 as OTP 19 is required to compile cowlib.

lib/phoenix/endpoint/cowboy2_websocket.ex
```diff
@@ -0,0 +1,116 @@
+defmodule Phoenix.Endpoint.Cowboy2WebSocket do
+  # Implementation of the WebSocket transport for Cowboy.
+  @moduledoc false
+
+  @behaviour :cowboy_websocket
+  @connection Plug.Adapters.Cowboy2.Conn
+  @already_sent {:plug_conn, :sent}
+
+  def init(req, {module, opts}) do
+    # Is this too brittle?
```
josevalim: Is this a non-rhetorical question? :P

lib/phoenix/endpoint/cowboy2_websocket.ex
```diff
@@ -0,0 +1,116 @@
...
+  end
+
+  defp format_reason(:exit, reason, _), do: reason
+  defp format_reason(:throw, reason, stack), do: {{:nocatch, reason}, stack}
+  defp format_reason(:error, reason, stack), do: {reason, stack}
+
+  ## Websocket callbacks
+
+  def websocket_init({module, req, opts}) do
+    conn = @connection.conn(req)
```
josevalim: Why do we have the same logic in websocket_init and init?

josevalim: And same logic I mean the module.init call.

lib/phoenix/endpoint/cowboy2_websocket.ex
```diff
@@ -0,0 +1,116 @@
...
+      when code in 1000..1003 or code in 1005..1011 or code == 1015 do
+    handler.ws_close(state)
+    :ok
+  end
+  def terminate(:remote, _req, {handler, state}) do
+    handler.ws_close(state)
+    :ok
+  end
+  def terminate(reason, _req, {handler, state}) do
+    handler.ws_close(state) # do we want to do this?
```
josevalim: I think terminate is responsible to close as well? One is a clean shutdown, the other one is not.

lib/phoenix/endpoint/cowboy2_websocket.ex
```diff
@@ -0,0 +1,116 @@
...
+        {:ok, %{adapter: {@connection, _}}, {_handler, args}} ->
+          {:ok, state, _timeout} = module.ws_init(args)
+          {:ok, {module, state}}
+      end
+    catch
+      kind, reason ->
+        # Although we are not performing a call, we are using the call
+        # function for now so it is properly handled in error reports.
+        mfa = {module, :call, [conn, opts]}
+      {__MODULE__, req, {:error, mfa, kind, reason, System.stacktrace}}
```
josevalim: Wrong indentation.

lib/phoenix/endpoint/cowboy2_websocket.ex
Gazler: Yeah, Is the way I am fetching the config here suitable? In cowboy 2 the timeout is provided here instead of in websocket_init.

lib/phoenix/endpoint/cowboy2_websocket.ex
```diff
@@ -0,0 +1,116 @@
...
+        {_, socket, _} = opts
+        %{websocket: {_, config}} = socket.__transports__
+        timeout = Keyword.fetch!(config, :timeout)
+        {:cowboy_websocket, req, {module, req, opts}, %{idle_timeout: timeout}}
+      {:error, %{adapter: {@connection, req}}} ->
+        {:error, req}
+    end
+  end
+
+  def resume(module, fun, args) do
```
Gazler: :scissors: - don't need this function.

lib/phoenix/endpoint/cowboy2_websocket.ex
```diff
@@ -0,0 +1,116 @@
...
+        {:suspend, __MODULE__, :resume, [module, fun, args]}
+      _ ->
+        # We are forcing a shutdown exit because we want to make
+        # sure all transports exits with reason shutdown to guarantee
+        # all channels are closed.
+        exit(:shutdown)
+    end
+  end
+
+  defp format_reason(:exit, reason, _), do: reason
```
Gazler: Don't need this

lib/phoenix/endpoint/cowboy2_websocket.ex
Gazler: So, we need to do the first inside init here (force SSL, origin check, etc.) - this is the HTTP part of the request (before the upgrade).

We then want to do the Transport.connect part of the code here.

https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/transports/websocket.ex#L77

Perhaps we should split the logic of init in the transport and update the Cowboy 1 adapter to do both?

josevalim: @Gazler yes, we should split the logic somehow. I have been planning to rework this part for almost 2 years now so I am glad I have procrastinated. :P Feel free to break it apart so it works on both, later I will come back and revisit everything. :)

Gazler: @josevalim Turns out splitting this was way trickier than I first anticipated. I pushed a different fix though. Could you check it out please?

Gazler:
@josevalim Turns out splitting this was way trickier than I first anticipated. I pushed a different fix though. Could you check it out please?

Gazler:
Waiting for Cowboy 2.1 before merging this. Please see https://github.com/ninenines/cowboy/commit/c602871f86c795da8463908d12f7bf966bfeec12#commitcomment-25284551

chrismccord:
Looks good! We should update the new project template `mix.exs` to point to cow2. Thanks!

Gazler:
@chrismccord Next on the list.

michalmuskala:
Do we have any benchmarks comparing Phoenix on Cowboy 2 to the previous version?

chrismccord:
Not to my knowledge but I would like to see some # 's if someone has time to run some tests

Gazler:
https://potatosalad.io/2017/08/20/load-testing-cowboy-2-0-0-rc-1 by @potatosalad is relevant

potatosalad:
If it's helpful, I can try to put together some results of HTTP/1.1 performance as well as HTTP/2 performance using cowboy directly as well as the full Phoenix/Plug stack.

If I remember correctly for 2.0.0-rc.1, HTTP/1.1 performance was slightly worse than in cowboy 1.x due to changes in how processes are spawned (I think there's 1 extra layer of parent/child processes in 2.x compared to 1.x).  I could also remember that incorrectly, so I'd need to run the tests again to see.

Anyone else is also welcome to run the same benchmarks, but you'll have to compile a [custom version of h2load](https://github.com/potatosalad/nghttp2/tree/elixirconf2017) to get the "duration" behavior where the benchmark is over a period of time versus a finite number of requests.

chrismccord:
That would be extremely helpful if you have the time! I also remember something along those lines perf wise as the new H2 model will require some more overhead.

josevalim:
> (I think there's 1 extra layer of parent/child processes in 2.x compared to 1.x)

AFAIK Cowboy 1.0 started one process for the whole connection and multiple requests on the same connection used the same process. In Cowboy 2.0, you get one process for the connection and the another process per request.

@potatosalad is there any chance of a port of h2o becoming usable in the future?

potatosalad:
> AFAIK Cowboy 1.0 started one process for the whole connection and multiple requests on the same connection used the same process. In Cowboy 2.0, you get one process for the connection and the another process per request.

I think that's where the extra layer comes from.  Spawning processes has a fairly significant impact on performance under certain circumstances (requests per second load testing being one of them).

> @potatosalad is there any chance of a port of h2o becoming usable in the future?

@josevalim *TL;DR* Possibly.

I stopped experimenting with h2o back in March/April because there were too many pieces missing from OTP at the time to continue development, namely: [`enif_open_resource_type_x`](http://erlang.org/doc/man/erl_nif.html#enif_open_resource_type_x) and the [`enif_ioq_*`](http://erlang.org/doc/man/erl_nif.html#enif_ioq_create) functions which were added in OTP/20 and OTP/20.1.  The ability to call a BEAM function from a NIF would also be a nice-to-have feature.

However, the biggest performance roadblock I ran into is actually from spawning processes.

When I configured the supervision tree to look similar to cowboy, I got performance out of h2o only slightly better than cowboy (for example, 90k req/s from cowboy might result in 110k req/s from h2o).

When I configured everything to run out of a single process running a loop as fast as possible, I was able to get requests per second from h2o that were 4-5x faster than cowboy (and fairly consistent with raw numbers you would get from running h2o standalone).

In general, however, the ~400k req/s results from h2o are very sensitive to minor adjustments.  For example, the difference between using a map versus a tuple for the request object sent from the NIF to a process was the difference between ~300k and ~400k (I don't remember the exact numbers, but I remember shaking my fist and yelling "Why!?" regularly).

So, I'm still not sure what the next steps would be there.  To name a few options I've thought of:

 * Finish [erlang-h2o](https://github.com/potatosalad/erlang-h2o) in its current form because slightly better performance over cowboy is better than nothing.
 * Try to figure out why process spawning in OTP is causing such performance degradation.  If it can be solved, would cowboy's numbers look similar to h2o?
 * Try to add support to OTP for calling BEAM functions from the NIF side of things.  This could potentially open up a whole new set of performance possibilities.
 * Find some sort of middle-ground where processes have an expiration time or limited number of request uses before they die.  This seemed to yield ~200 req/s, but was a rough draft implementation.
 * ???

There are still some potential benefits to a fast-as-cowboy h2o implementation:

 * Not every path needs to actually route to the BEAM side of things.  So, for example, serving static files can happen entirely on the NIF side of things and result in ~400k req/s for those requests.
 * You might not need a reverse proxy like nginx sitting in front of your application.
 * Listening on UNIX domain sockets.
 * Non-OTP SSL/TLS possibilities (like [picotls](https://github.com/h2o/picotls) or [LibreSSL](https://github.com/libressl-portable/portable) + [neverbleed](https://github.com/h2o/neverbleed)).
 * Similar to cowboy's handlers + middleware, h2o supports:
   * [logger](https://github.com/h2o/h2o/blob/v2.2.3/include/h2o.h#L144-L154) &mdash; MAY log a request
   * [filter](https://github.com/h2o/h2o/blob/v2.2.3/include/h2o.h#L132-L142) &mdash; MAY modify a response

     For example, automatic gzip or brotli compression from [`compress.c`](https://github.com/h2o/h2o/blob/v2.2.3/lib/handler/compress.c)

   * [handler](https://github.com/h2o/h2o/blob/v2.2.3/include/h2o.h#L120-L130) &mdash; MAY generate a response
 * A bunch of other features builtin: fastcgi, throttling, caching, mruby, reverse proxying, etc.

PeterDavidCarter:
What are the benefits of Cowboy2? Will it negate the need for an nginx reverse proxy?

chrismccord:
> What are the benefits of Cowboy2? Will it negate the need for an nginx reverse proxy?

Namely HTTP2. Cowboy never required nginx in front of it, though a lot of folks do deploy that way. Please send further questions to the forums or slack where folks will be able to better direct you.

josevalim:
Thanks @potatosalad for the in-depth reply. If all we get is a slightly better performance, then the then it is probably not worth it due to its impact on robustness and maintenance. Maybe the other goals would make it worth but with the OTP team moving out of linked-in drivers and into NIFs, we are also bound to get performance improvements and it may even make the other SSL options simpler to interface with directly from Erlang.

josevalim:
> Try to figure out why process spawning in OTP is causing such performance degradation. If it can be solved, would cowboy's numbers look similar to h2o?

Cowboy 2.0 still has multiple acceptors right?  I wonder if we could have something like "process templates". Since all processes started by each acceptor are mostly equal in terms of group leaders, flags, most of the args, etc, I wonder if we could define the memory layout, flags and structures once, and then spawning a process just needs to copy this shared structure to a new area and then do the non-shared part of the work.

michalmuskala:
I wonder if increasing the initial heap size for the connection processes could have an impact. By default the heap is very tiny, so I would imagine doing pretty much any work in there will trigger at least couple garbage collections very quickly.

potatosalad:
Here are the initial benchmark results from my MacBook Pro i7 16GB running macOS 10.13.1, Erlang OTP/20.1.4, and Elixir 1.5.2.

These load tests aren't perfect and can probably be improved.  All of the code for the setup is available at [potatosalad/phoenix_h2load](https://github.com/potatosalad/phoenix_h2load).

| Stack   | h1 (1.x &mdash; 2.x)      | h2 (2.x)  |
|---------|--------------------------:|----------:|
| phoenix | `13,209` &mdash; `12,173` | `10,178`  |
| plug    | `15,831` &mdash; `18,193` | `16,381`  |
| cowboy  | `17,412` &mdash; `24,972` | `24,320`  |
| h2o     | `52,477`                  | `122,253` |

Here is the graphical comparison of the different stacks (cowboy, plug, and phoenix) using cowboy 1.x and HTTP/1.1 load testing for 1 minute:

cowboy 1.x (h1)

cowboy 2.x and HTTP/1.1 load testing for 1 minute:

cowboy 2.x (h1)

cowboy 2.x and HTTP/2 load testing for 1 minute:

cowboy 2.x (h2)

For comparison, here is cowboy 2.x versus h2o with HTTP/1.1 load testing for 1 minute:

cowboy 2.x vs h2o (h1)

For comparison, here is cowboy 2.x versus h2o with HTTP/2 load testing for 1 minute:

cowboy 2.x vs h2o (h2)

## Basic Load Testing

If you want to run basic load testing without the graphical representation, you can simply clone [potatosalad/phoenix_h2load](https://github.com/potatosalad/phoenix_h2load) and run `make` if you have docker available on your machine.

It will run through the same tests (except h2o) and display the averaged results instead of incremental ones.

By default the tests will run for 1 minute per test, but you can adjust the time spent load testing by setting `H2LOAD_DURATION`, for example:

```bash
H2LOAD_DURATION=10s make
```

You can see example output on [travis](https://travis-ci.org/potatosalad/phoenix_h2load/builds/300283695) or here are the resulting numbers from the travis build:

| Stack   | h1 (1.x &mdash; 2.x)      | h2 (2.x)  |
|---------|--------------------------:|----------:|
| phoenix |  `9,322` &mdash;  `6,813` |  `6,301`  |
| plug    | `13,733` &mdash;  `9,352` |  `9,923`  |
| cowboy  | `16,893` &mdash; `11,733` | `13,326`  |

*Note:* I realized after recording these numbers that the tests were run on travis without kernel polling enabled.  There is a [new build](https://travis-ci.org/potatosalad/phoenix_h2load/builds/300297644) running, but hasn't finished yet at the time of writing.

## Explanation of Process

To repeat the graphical load tests, you'll need to clone [potatosalad/phoenix_h2load](https://github.com/potatosalad/phoenix_h2load) and build a custom version of h2load from [potatosalad/nghttp2@cowboy2](https://github.com/potatosalad/nghttp2/tree/cowboy2):

```bash
git clone https://github.com/potatosalad/nghttp2.git
cd nghttp2
git checkout cowboy2
git submodule update --init
autoreconf -i
automake
autoconf
./configure --enable-app --disable-silent-rules --disable-python-bindings --enable-asio-lib --with-boost --without-jemalloc
make -j8
```

*Note:* You might have to adjust the configure options based on your system.  For example, on macOS with homebrew, I had to use `--with-boost=$(brew --prefix boost)`.

You should end up with an executable located at `src/h2load` where you can run one of the following:

```bash
# For testing cowboy 1.x
export COWBOY_VERSION=1
export MIX_ENV=prod
export PORT=29595
mix do deps.get, deps.compile, compile
elixir --erl "+K true" --name phoenix_h2load@127.0.0.1 --cookie mycookie -S mix phx.server
```

In a separate terminal, the separate stacks can be load tested:

```bash
cd priv
export H2LOAD_BIN="~/path/to/nghttp2/src/h2load"
./h2load.sh $H2LOAD_BIN c1_h1_cowboy h1 29593
./h2load.sh $H2LOAD_BIN c1_h1_plug h1 29594
./h2load.sh $H2LOAD_BIN c1_h1_phoenix h1 29595
```

The cowboy 2.x tests are identical, with the exception of the removal of the `export COWBOY_VERSION=1` line above.  Also, the `h1` can be replaced with `h2` to perform HTTP/2 load testing.

potatosalad:
Kernel polling does seem to make a difference, the [latest results from travis](https://travis-ci.org/potatosalad/phoenix_h2load/builds/300300547) are a little closer to the results from my local machine:

| Stack   | h1 (1.x &mdash; 2.x)      | h2 (2.x)  |
|---------|--------------------------:|----------:|
| phoenix | `13,375` &mdash;  `9,851` |  `8,368`  |
| plug    | `22,832` &mdash; `14,152` | `14,727`  |
| cowboy  | `28,463` &mdash; `18,384` | `19,122`  |

mix.exs
```diff
@@ -56,9 +57,15 @@ defmodule Phoenix.Mixfile do
   end
 
   defp deps do
+    cowboy_version =
```
scrogson: What do you think about pulling this into a function similar to the `lockfile/0`?

```ex
defp cowboy_dep do
  case System.get_env("COWBOY_VERSION") do
    "1" -> {:cowboy, "~> 1.0", optional: true}
    _ -> {:cowboy, github: "ninenines/cowboy", optional: true, override: true}
  end
end
```

Or

```ex
defp cowboy_dep("1"), do: ...
defp cowboy_dep(_), do: ...

defp deps do
  [
    cowboy_dep(System.get_env("COWBOY_VERSION")),
    ...
  ]
end
```

potatosalad:
I realized after the fact that my initial load test was done in a suboptimal environment (I had a few virtual machines running in the background while load testing).

So, after modifying the cowboy 1.x handler to be more similar to plug, I ran the load testing again on a clean machine.

Here are the updated results from [potatosalad/phoenix_h2load](https://github.com/potatosalad/phoenix_h2load):

| Stack   | h1 (1.x &mdash; 2.x)      | h2 (2.x)  |
|---------|--------------------------:|----------:|
| phoenix | `41,509` &mdash; `33,617` | `25,362`  |
| plug    | `56,005` &mdash; `49,927` | `46,765`  |
| cowboy  | `59,112` &mdash; `62,195` | `69,203`  |
| h2o     | `152,433`                 | `359,611` |

Here is a breakdown of the measured requests per second by stack.

*Note:* The highest results are in **BOLD**.

| phoenix             | AVG          | MIN          | MAX          |
|---------------------|-------------:|-------------:|-------------:|
| **cowboy 1.x (h1)** | **`41,509`** | **`38,635`** | **`59,146`** |
|   cowboy 2.x (h1)   |   `33,617`   |   `31,756`   |   `44,369`   |
|   cowboy 2.x (h2)   |   `25,362`   |   `23,815`   |   `33,802`   |

| plug                | AVG          | MIN          | MAX          |
|---------------------|-------------:|-------------:|-------------:|
| **cowboy 1.x (h1)** | **`56,005`** |   `41,774`   | **`82,214`** |
|   cowboy 2.x (h1)   |   `49,927`   |   `40,236`   |   `63,394`   |
|   cowboy 2.x (h2)   |   `46,765`   | **`41,979`** |   `64,274`   |

| cowboy              | AVG          | MIN          | MAX           |
|---------------------|-------------:|-------------:|--------------:|
|   cowboy 1.x (h1)   |   `59,112`   |   `31,395`   | **`124,998`** |
|   cowboy 2.x (h1)   |   `62,195`   |   `45,229`   |   `92,683`    |
| **cowboy 2.x (h2)** | **`69,203`** | **`61,716`** |   `94,772`    |

Here is the graphical comparison of the different stacks (cowboy, plug, and phoenix) using cowboy 1.x and HTTP/1.1 load testing for 1 minute:

cowboy 1.x (h1)

cowboy 2.x and HTTP/1.1 load testing for 1 minute:

cowboy 2.x (h1)

cowboy 2.x and HTTP/2 load testing for 1 minute:

cowboy 2.x (h2)

For comparison, here is cowboy 2.x versus h2o with HTTP/1.1 load testing for 1 minute:

cowboy 2.x vs h2o (h1)

For comparison, here is cowboy 2.x versus h2o with HTTP/2 load testing for 1 minute:

cowboy 2.x vs h2o (h2)

Gazler:
@potatosalad Thanks so much for this, these benchmarks are great!

I have a small commit that brings Plug performance more inline with Cowboy from my tests. Would you be able to try it out of possible? https://github.com/Gazler/plug/commits/gr-improve-handler

Would you also be able to advise how to run the tester? I got this from my test (by the way how do you make the pretty charts from your output?!):

https://gist.github.com/Gazler/d8f7371e2fbabf5d2fac491e2ccd6464

However it is failing on:

```
 GEN    cowboy2-h1-test
[cowboy 2.x] load testing cowboy (h1) for 5s
Error response from daemon: Container 89171e9381fedb5e7ce3a602a5e78c504650db0cd6863125f4cb454c9fa90505 is not running
Makefile:108: recipe for target 'cowboy2-h1-test' failed
make: *** [cowboy2-h1-test] Error 1
```

potatosalad:
@Gazler That's a curious failure, I'm not sure why one would run while the other one fails.  They both use the same docker container.  I opened an issue over at potatosalad/phoenix_h2load#1 so as not to hijack this pull request too much.

I'll see if I can run the tests again with the new plug changes, but if I can't get it done this week I should be able to beginning of next week.

lib/phoenix/endpoint/cowboy2_handler.ex
```diff
@@ -0,0 +1,122 @@
...
+        http: [dispatch: [
+                {:_, [
+                    {"/foo", MyApp.CustomHandler, []},
+                    {"/bar", MyApp.AnotherHandler, []},
+                    {:_, Plug.Adapters.Cowboy2.Handler, {MyApp.Endpoint, []}}
+                  ]}]]
+
+  It is also important to specify your handlers first, otherwise
+  Phoenix will intercept the requests before they get to your handler.
+  """
```
narrowtux: I made changes to the original in #2643

It fixes formatting (the code blocks don't render completely) and shows how to get the live-reload websocket in there too. 

If that PR gets pulled, can you do the same changes here as well?

Gazler: Yeah, I'll be sure to update it, thanks. Please shout at me if I forget!

idyll:
Is there anything I can help with to get this merged into master?

Gazler:
Stopped at about the point we got to the benchmarks. Found the cause
though, so hoping to get back to it at the weekend. Any help with
https://github.com/elixir-plug/plug/issues/629 would be appreciated

idyll:
I have been testing things out and I am pretty sure that multipart uploads are failing with H2 (and not HTTP/1.1) possibly due to an issue with plug? I am still digging into this but wanted to report here in case some knows what's going on.

```
[debug] ** (Plug.Parsers.ParseError) malformed request, a RuntimeError exception was raised with message "invalid multipart, body terminated too soon"
    (plug) lib/plug/conn.ex:919: Plug.Conn.next_multipart/3
    (plug) lib/plug/conn.ex:885: Plug.Conn.read_part_body/8
    (plug) lib/plug/parsers/multipart.ex:74: Plug.Parsers.MULTIPART.parse_multipart_headers/5
    (plug) lib/plug/parsers/multipart.ex:53: Plug.Parsers.MULTIPART.parse_multipart/5
    (plug) lib/plug/parsers/multipart.ex:43: Plug.Parsers.MULTIPART.parse_multipart/2
    (plug) lib/plug/parsers/multipart.ex:15: Plug.Parsers.MULTIPART.parse/5
    (plug) lib/plug/parsers.ex:221: Plug.Parsers.reduce/4
```

Gazler:
Can you make a reproducable example and push the repo? I'll take a look at
it then.

Ilyes512:
Hey, I got a couple of questions:
How can I use this version of phoenix that uses cowboy 2. I already tried a couple of things, but I believe the generated mix.exs is not up-to-date (still says cowboy version 1). Besides being able to run it, is there an ETA of when this will be release to stable? (would this be in version 1.4 of Phoenix?).

Gazler:
@Ilyes512 You can use it by generating a new phoenix project as normal, then setting the following deps in `mix.exs`:

```elixir
      {:phoenix, github: "phoenixframework/phoenix", branch: "gr-cowboy2", override: true},
      {:plug, github: "elixir-plug/plug", override: true},
      {:cowboy, "~> 2.1", override: true}
```

Then go into `config/config.exs` and add the following to the endpoint config:

```elixir
  handler: Phoenix.Endpoint.Cowboy2Handler,
```
I imagine that the Cowboy 2 handler will be included in Phoenix 1.4, we will likely need to have some beta releases of Plug and Phoenix before it can be considered stable though.

idyll:
To help with the TLS issue on localhost I have registered `phx.sh`. I am going to point it and all subdomains to 127.0.0.1.

I will host with DNSimple. @josevalim and @chrismccord let me know if you want to be invited to make changes to it.

chrismccord:
@idyll that's a fantastic idea! @Gazler do you think this would be a nice solution to TLS on localhost? If we go this route, we'd need the phoenix org to have control over such a domain so we can promote it in the guides and book without risk of someone else nabbing an expired domain and redirecting in the future. @idyll would you be open to gifting this domain to the phoenix org? Thanks! <3

idyll:
Totally, Chris. You can have control over it.
Lets make a plan to set this up in a way so that I will still pay for it. ;)

idyll:
@Gazler here's a simple project that you can use to see the multipart issue.
https://github.com/idyll/sunshine

If you connect on port 4001 you will be using H2 and image uploads will fail. If you connect on 4002, the browser will downgrade you to HTTP/1.1 and image uploads will work.

Gazler:
@idyll This should be fixed on Plug master. Thanks for raising it and providing a sample repo!

In terms of the phx.sh domain, what's the advantage here over generating a self-signed certificate? Could you expand on the usage for me please?

chrismccord:
Assuming it Just Works, the advantage would be a thing for folks to get instant TLS support on localhost without jumping through any hoops (unless there's an automated way for us to generate self signed certs for local use). I'll let @idyll chime in though – I'm admittedly less well versed here that both of you :)

josevalim:
Although a downside is that it requires internet connection during dev? It
is not a major problem but we need to make sure we guide the users if
connection is not available.
-- 


Ilyes512:
Just today I experimented with self-signed certificate and cowboy2. Only requirement is that you have openssl installed:

```
$ which openssl
/usr/bin/openssl
```
It returns `openssl not found` if you don't have openssl installed.

At the root of your phoenix app create a new cert dir and move into it:
```
$ mkdir cert && cd cert
```

Create a self-signed cert (this will generate 2 files: a key and a certificate):
```
$ openssl req -newkey rsa:2048 -nodes -keyout example.key -x509 -days 365 -out example.crt -subj "/C=NL/ST= /L= /O= /CN= "
```

Now you should just need to point to the files from Phoenix/Plug/Cowboy.

idyll:
The inspiration for the `phx.sh` comes from `pow` and `puma-dev` from the Rails community. Apparently though, this week, new broke that Google is going to mess up `.dev` in Chrome because they technically own it. 

`.dev` was a bad idea anyways, because a certificate for `*.dev` is never actually valid. So if you've got a complicated local dev setup for Docker or using a proxy the certs can't be trusted. But `phx.sh` or `*.phx.dev` can be valid.

Right now the example shipped in the Phoenix template generates a snake oil cert for `www.example.com`. This lines right up with that practice. Instead you'd generate a snake oil cert for `*.phx.sh` or `phx.sh` and away you'd go. This could be really nice for people using local proxies or similar. You'd still be prompted to accept the certificate in Safari, but at least your trusting the cert for a hostname that matches. It worries me a little what browsers may do in the future for a hostname mismatch. We just don't know.

Taking that example a bit further we could encourage people to use a subdomain on their project. So that if they do need to revoke a cert from their keychain, at least its clear which one.

We may want to use this for Phoenix to use internally, for testing. We can add an encrypted private key and a cert, signed by a real CA, to the test suites. You won't need a special setup for this to work. Just the password for the key.

idyll:
Hmmmm. What if we setup let's encrypt and then issued certs to projects as they needed them...
I think we can use dns-01:

https://community.letsencrypt.org/t/how-to-use-dns-01-challenge/28593

mix.exs
```diff
@@ -57,8 +58,8 @@ defmodule Phoenix.Mixfile do
 
   defp deps do
     [
-      {:cowboy, "~> 1.0", optional: true},
-      {:plug, "~> 1.3.3 or ~> 1.4"},
+      cowboy_dep(System.get_env("COWBOY_VERSION")),
+      {:plug, "~> 1.5.0-rc.0", override: true},
```
Gazler: @josevalim @chrismccord If we are happy with specifying this version for now then this can be merged.

I'll do a new PR updating the generator so we can release a new archive and people can begin trying this out.

josevalim:
Thanks @Gazler! I don't think we need new generators, do we? Rather we need to include instructions on how to use cowboy 2? I assume we only need three things:

1. Change the cowboy version in the mix.exs
2. Set the Cowboy2 handler in config/config.exs
3. Generate a certificate

I don't think we can skip 1, but can we autodetect cowboy2 on step 2? And did we have any progress regarding certificates or in providing a more guided experience to developers?

Gazler:
Sorry, I wasn't clear,  I meant to say installer not generators. I don't
think we want anything like `mix phx.new foo --cowboy2`.

We can either leave 1.x as default and have a comment about 2.1 in the
mix.exs file. Also a comment on changing the handler in config/config.exs

Or we can go ahead and make 2.x default on new projects.

josevalim:
Agreed. I would have cowboy1 as the default still but have some docs (maybe on the guides) that teach how to use HTTP 2. I think you should go ahead and merge this, we will continue iterating on master. :)

lib/phoenix/endpoint/cowboy2_handler.ex
narrowtux: Just a heads up, it was pulled today.

Gazler: Thanks for the reminder!

PR #2624, opened 2017-11-07
fix init return type in Presence.init/1

flaviogrossi:
The phoenix presence init/1 return type does not match the one declared in the callback declaration, causing an error when dialyzer is used.

This small PR should fix it.

### DIFF
--- a/lib/phoenix/presence.ex
+++ b/lib/phoenix/presence.ex
@@ -108,7 +108,7 @@ defmodule Phoenix.Presence do
   @type topic :: String.t
 
   @callback start_link(Keyword.t) :: {:ok, pid()} | {:error, reason :: term()} :: :ignore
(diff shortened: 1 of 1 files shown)


lib/phoenix/presence.ex
```diff
@@ -108,7 +108,7 @@ defmodule Phoenix.Presence do
   @type topic :: String.t
 
   @callback start_link(Keyword.t) :: {:ok, pid()} | {:error, reason :: term()} :: :ignore
-  @callback init(Keyword.t) :: {:ok, pid()} | {:error, reason :: term}
+  @callback init(Keyword.t) :: {:ok, map()} | {:error, reason :: term}
```
chrismccord: Can you make this `{:ok, state :: term}` ? Then we are all set. Thanks!

chrismccord:
❤️❤️❤️🐥🔥

flaviogrossi:
Done!

PR #2628, opened 2017-11-09
Include documentation bug reports

febeling:
I found a passage in the routing documentation that I thought needs improvement and came here to check how to report and possibly contribute a patch, once the answer was researched.

Then the guidelines seems to rule out non-code issues being opened. I brought up the question where to raise doc issues on #elixir-lang. The view was offered that documentation was as important as code and needs to be handled similarly wrt issues. 

This is an attempt to reflect that in the contrib guide. Please consider this suggestion.


chrismccord:
❤️❤️❤️🐥🔥

PR #2629, opened 2017-11-11
Explicit js api for updating Channel reconnect join params

davydog187:
By default, Phoenix channels send up exactly the same join payload up to the server on a reconnect. This is fine for simple cases, but in more complex scenarios, you might have additional data to send to `join` in the case of reconnect.

This PR aims to provide an explicit API for updating the join params. The closest discussion that I could about this problem was [here](https://groups.google.com/forum/#!topic/phoenix-core/YrfjvctJW_0), but was referring to the socket payload.

Happy to add additional tests, or change function names if there are better alternatives. 🍍 🍍 🍍

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -270,6 +270,16 @@ class Push {
   }
 
   /**
+   * Updates the Push payload for subsequent resends
+   *
+   * @param {Object} payload
+   * @returns {Push}
+   */
+  updatePayload(payload) {
+    this.payload = payload;
+  }
+
+  /**
    * @private
    */
   reset(){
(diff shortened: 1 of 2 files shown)


davydog187:
I wanted to give more justification to this PR since its been sitting here.

As it currently stands, there is not an obvious way for app developers to update the reconnect params for channel joins. This PR intents to make an obvious and explicit API for doing so to remove confusion and simplify the process.

chrismccord:
❤️❤️❤️🐥🔥

chrismccord:
FYI this feature has been replaced in master in favor of passing a function to socket and channel params, for example:

```javascript
let socket = new Socket("/socket", {params: function(){ return myParams })
```

PR #2633, opened 2017-11-13
Update the docs for bypass_through to be more explicit.

digitalcake:
I found it hard to understand that invoking the route/endpoint was related to the use of `get("/")` and ran into a `** (ArgumentError) flash not fetched, call fetch_flash/2` error on my tests related to the fact I had a flash dependency in my test but was not invoking the router.  That led me to write a post in the elixir forums. https://elixirforum.com/t/bypass-through-3-not-working/10070/2
 
It was later made clear that I needed to invoke the router via `get("/")`. 

I made this pr in hopes to guide people to this same solution when searching for `flash not fetched, call fetch_flash/2` since I was unable to resolve my issue by searching alone and felt the docs missed explicitly calling out why the use of `get("/")` was needed.

### DIFF
--- a/lib/phoenix/test/conn_test.ex
+++ b/lib/phoenix/test/conn_test.ex
@@ -495,7 +495,7 @@ defmodule Phoenix.ConnTest do
   Useful for unit testing Plugs where Endpoint and/or
   router pipeline plugs are required for proper setup.
   
-  Note the use of `get("/")` following `bypass_through` in the examples bellow. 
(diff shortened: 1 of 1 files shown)


chrismccord:
Merged in https://github.com/phoenixframework/phoenix/commit/f3a6bceba26ff550aff3cd2746d8b38d6b7f1ccb 

❤️❤️❤️🐥🔥

PR #2639, opened 2017-11-21
Add .eot and .ttf to gzippable_exts

kimlai:
This change makes `mix phx.digest` compress font files by default.
`.woff` files are still ignored since they're already compressed.


chrismccord:
❤️❤️❤️🐥🔥

PR #2640, opened 2017-11-21
Fix direct link to endpoint API docs

rossta:
Autolinking of the text inside ticks within link was causing markdown to be output to text of link (image). Also fixes the url fragment to the Endpoint API section.

endpoint_ _phoenix_v1_3_0


Gazler:
Thanks <3

PR #2643, opened 2017-11-22
Update documentation on the http dispatch option

narrowtux:
* Added `.Web` to all the `MyApp` module names since phoenix 1.3 or later generates with that name
 * Added the needed route for the live_reload socket

### DIFF
--- a/lib/phoenix/endpoint/cowboy_handler.ex
+++ b/lib/phoenix/endpoint/cowboy_handler.ex
@@ -23,39 +23,53 @@ defmodule Phoenix.Endpoint.CowboyHandler do
   You will need the following rules:
 
     * Per websocket transport:
-      
-          {"/socket/websocket", Phoenix.Endpoint.CowboyWebSocket,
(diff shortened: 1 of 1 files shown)


narrowtux:
Weirdly, a unit test failed (only in one elixir version) but this PR doesn't even change code

chrismccord:
Can you update the alias to use `MyAppWeb` rather than the `.Web`? Also please kill the added whitespace on the newlines. Thanks!

narrowtux:
Ok, I added the whitespace because the markdown parser produces this weirdness: 

image

I will change the module names though.

narrowtux:
image
found another way to format the code blocks, now they render correctly

narrowtux:
I also added a line about the init/1 callback in the endpoint module, which might help readers if they wonder why their changes to `config/*.exs` aren't applied (happened to me)

narrowtux:
@chrismccord is there still something missing for this to be pulled?

chrismccord:
❤️❤️❤️😍🔥

Issue #2641, opened 2017-11-21
Broken link to MySQL Guide


manju16832003:
### Broken link to MySQL guide

Refer to https://github.com/elixir-lang/ex_doc/issues/795

chrismccord:
Thanks for the report! Can you send a PR to fix the link?

obahareth:
I think this issue should be closed since #2644 has been merged.

PR #2644, opened 2017-11-23
Add 'Using MySQL' doc file

bowyern:
Create the "Using mysql" page referenced by the Ecto docs. Should address #2641.


Gazler:
Thanks for the PR. Rather than adding a new page for Using MySQL, I think it could be added as a section to the bottom of the Ecto guide instead. If you could do the following (either on this PR or in a new one) that would be great:

 * Change `# Using MySQL` to `## Using MySQL`
 * Copy the contents to the bottom of https://github.com/phoenixframework/phoenix/blob/master/guides/docs/ecto.md
 * Update the link in the Ecto guide to be [ecto.html#using-mysql]
 * Remove the using_mysql.md file

guides/docs/ecto.md
```diff
@@ -447,3 +447,80 @@ In addition to inserts, we can also perform updates and deletes with `Repo.updat
...
+  . . .
+  # Specifies your project dependencies.
+  #
+  # Type `mix help deps` for examples and options.
+  defp deps do
+    [
+      {:phoenix, "~> 1.3.0"},
+      {:phoenix_pubsub, "~> 1.0"},
+      {:phoenix_ecto, "~> 3.2"},
+      {:mariaex, ">= 0.0.0"},
```
radar: Should this version be locked down, like the others? I would suggest:

```
{:mariaex, "~> 0.8.3"}
```

bowyern: That was what I thought to do initially, but I noticed that creating a project with --database=mysql uses the included dep line.

Gazler:
Thanks <3

PR #2646, opened 2017-11-23
Learning guide link now talks about Phoenix + other projects

radar:
The link to the learning guide from the overview page mentioned a "Learning Elixir and Erlang guide", but the `learning.html` page is neither of these things. I think the text should be updated to more accurately reflect the content of those pages.

I've also added a link to the [elixir-lang.org learning](https://elixir-lang.org/learning.html) page too, as I think that contains the information that the original learning link should've been pointing to.


Gazler:
Great addition. Thanks <3

PR #2649, opened 2017-11-27
Optimize Phoenix.Controller.current_path/1

entertainyou:
Directly use query_string from Plug.Conn instead of encode query_params.


chrismccord:
❤️❤️❤️🐥🔥

PR #2650, opened 2017-11-27
Fix the generated code of the ecto repo 

jfayad:
So that the `url:` param in the config file is being properly picked up if we want to hard code the config. in the dev. env.

With the previous code, even if we were populating the `url:` param in the dev.exs file we would have errors being thrown as the `DATABASE_URL` would be empty and would also empty the param for ecto thus creating an error.

This new code simply checks if the env. variable exists, if it does then it uses it to populate the config. otherwise it ignores it.

### DIFF
--- a/installer/templates/phx_ecto/repo.ex
+++ b/installer/templates/phx_ecto/repo.ex
@@ -6,6 +6,10 @@ defmodule <%= app_module %>.Repo do
   DATABASE_URL environment variable.
   """
   def init(_, opts) do
-    {:ok, Keyword.put(opts, :url, System.get_env("DATABASE_URL"))}
+	  if url = System.get_env("DATABASE_URL") do
+	    {:ok, Keyword.put(opts, :url, url)}
+	  else
+	    {:ok, opts}
+	  end
   end
 end
--- a/installer/templates/phx_ecto/repo.ex
+++ b/installer/templates/phx_ecto/repo.ex
@@ -6,10 +6,10 @@ defmodule <%= app_module %>.Repo do
   DATABASE_URL environment variable.
   """
   def init(_, opts) do
-	  if url = System.get_env("DATABASE_URL") do
-	    {:ok, Keyword.put(opts, :url, url)}
-	  else
-	    {:ok, opts}
-	  end
+    if url = System.get_env("DATABASE_URL") do
+      {:ok, Keyword.put(opts, :url, url)}
+    else
+      {:ok, opts}
+    end
   end
 end


chrismccord:
If it's missing (returns `nil`) it should be fine in dev, no? I have no issues on my end with this configuration in the dev env. Can you confirm?

chrismccord:
Ah, sorry – I misunderstood your issue. If you populate dev.exs with a url, this always overwrites which could indeed be confusing. Let me consult with @josevalim . Thanks!

jfayad:
Indeed that's the issue.

When I populate `url: ecto://my_db_url` in dev.exs for instance, it gets ignored because of the former template.

This new code fixed my issue, tested with a url provided in the config. file and with one provided as an env. variable and both worked.

Thx

installer/templates/phx_ecto/repo.ex
```diff
@@ -6,6 +6,10 @@ defmodule <%= app_module %>.Repo do
   DATABASE_URL environment variable.
   """
   def init(_, opts) do
-    {:ok, Keyword.put(opts, :url, System.get_env("DATABASE_URL"))}
+	  if url = System.get_env("DATABASE_URL") do
+	    {:ok, Keyword.put(opts, :url, url)}
+	  else
+	    {:ok, opts}
+	  end
```
webdeb: LGTM, I also had to modify it in my project ...maybe you could just fix the indentation to spaces?

josevalim:
I thought Ecto already ignored the `:url` if it was set to nil?

installer/templates/phx_ecto/repo.ex
jfayad: I've reviewed the code formatting, should be good now

jfayad:
@josevalim not sure what you mean exactly by "ignoring the value", the issue I faced was related to the fact that even if I declared a `url:` in my config file, it wasn't being picked up as the `init` of the `MyApp.Repo` was overriding it... this was leading to an unpredictable behaviour with random values showing up in the repo.config(). Simply making sure that the value wouldn't be overriden by the init function if no Env. variable exist fixed the issue. Thus this PR.

webdeb:
@josevalim looks like the initial `:url` configuration from config.exs/dev.exs is forced to be overridden by [repo_init](https://github.com/elixir-ecto/ecto/blob/1178f41be9bef3f2819adfbe8cfa81809fb1555f/lib/ecto/repo/supervisor.ex#L24)

josevalim:

PR #2653, opened 2017-11-28
Fix function arity in Views guide

cvortmann:
`HelloWeb.PageView` defines `render/2` functions but the documentation says `render/3`.


josevalim:

PR #2654, opened 2017-11-28
Remove diff of PageController create method in guide

fkschmidt:
A change is not necessary since the bootstrapped code already has no id param


josevalim:

PR #2659, opened 2017-12-01
Use 0.1.0 for the initial version in templates

lexmag:
Semantic versioning starts at 0.1.0.

Also has been fixed in Elixir some time ago: https://github.com/elixir-lang/elixir/pull/4788.


chrismccord:
❤️❤️❤️🐥🔥

PR #2665, opened 2017-12-10
Feat/log on invalid cowboy version

Gazler:
This relies on the Cowboy 2 PR, so that should be merged first. I discovered this could happen quite easily when upgrading an app using the gr-cowboy2 branch.

If a user updates their cowboy dependency without updating the relevant
hander in their config then the warning will be displayed. This should
help minimize issues when users simply update the Cowboy version without
also updating the config.

### DIFF
--- a/lib/phoenix/endpoint/handler.ex
+++ b/lib/phoenix/endpoint/handler.ex
@@ -78,15 +78,17 @@ defmodule Phoenix.Endpoint.Handler do
 
     If you wish to use Cowboy 1, please update mix.exs to point to the \
     correct Cowboy version:
-    
+
         {:cowboy, "~> 1.0"}
-    
+
     If you want to use Cowboy 2, then please remove the :handler option \
     in your config.exs file or set it to:
 
         handler: Phoenix.Endpoint.Cowboy2Handler
 
     """)
+
+    raise "aborting due to handler mismatch"
   end
   defp warn_on_different_handler_version(_user, _autodetected, _endpoint), do: nil
 end
--- a/lib/phoenix/endpoint/handler.ex
+++ b/lib/phoenix/endpoint/handler.ex
@@ -16,6 +16,8 @@ defmodule Phoenix.Endpoint.Handler do
   """
   @callback child_spec(scheme :: atom, endpoint :: module, config :: Keyword.t) :: Supervisor.Spec.spec
 
+  alias Phoenix.Endpoint.{CowboyHandler, Cowboy2Handler}
+
   use Supervisor
   require Logger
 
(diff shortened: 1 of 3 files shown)


josevalim:
I think we should go ahead and autodetect the handler. Basically:

1. autodetect the handler to see if we are using cowboy1 or cowboy2
2. if the user gave a handler, warn if user_handler is cowboy1 and autodetect is cowboy2
3. then return `user_handler || autodetected_handler`

Thoughts?

Gazler:
@josevalim sounds good to me. Should I keep the log message the same as in this PR?

lib/phoenix/endpoint/cowboy_handler.ex
```diff
@@ -114,4 +116,23 @@ defmodule Phoenix.Endpoint.CowboyHandler do
...
+        Logger.warn("""
+        This handler is for Cowboy 1.x and you are using Cowboy #{version}
+
+        If you wish to use Cowboy 1.x, please update mix.exs to point to the
+        correct Cowboy version. If you want to use Cowboy 2, then please set:
+
+            handler: Phoenix.Endpoint.Cowboy2Handler,
+
+        In your config.exs file.
+
```
josevalim: I think you don't need this trailing newline.

lib/phoenix/endpoint/cowboy_handler.ex
```diff
@@ -114,4 +116,23 @@ defmodule Phoenix.Endpoint.CowboyHandler do
     addr_str = :inet.ntoa(addr)
     "Running #{inspect endpoint} with Cowboy using #{scheme}://#{addr_str}:#{port}"
   end
+
+  defp log_on_invalid_cowboy_version(endpoint) do
+    case Application.spec(:cowboy, :vsn) do
+      [?1 | _] -> nil
+      version ->
+        Logger.warn("""
+        This handler is for Cowboy 1.x and you are using Cowboy #{version}
```
josevalim: "This handler" is not clear. Probably something like this:

```
You have specified Phoenix.Handlers.Cowboy for cowboy v1.x in the
:handler configuration of your Phoenix endpoint #{inspect endpoint}
but your mix.exs has fetched cowboy v2.x.
```

The rest of the message is perfect.

lib/phoenix/endpoint/cowboy_handler.ex
```diff
@@ -114,4 +116,23 @@ defmodule Phoenix.Endpoint.CowboyHandler do
...
+        correct Cowboy version. If you want to use Cowboy 2, then please set:
+
+            handler: Phoenix.Endpoint.Cowboy2Handler,
+
+        In your config.exs file.
+
+        """)
+    end
+  end
+
```
chrismccord: ✂️

Gazler:
@josevalim I've updated this to autodetect. It is very unlikely that the CowboyHandler can be used when the Cowboy version is 2.x - so maybe we should raise instead of warning? The only situation I can think of is that someone has their own fork which has a different non-1.x version for Cowboy.

Gazler:
@josevalim @chrismccord Are you both happy with this approach?

josevalim:
@Gazler I think we should raise in all cases, since it will certainly fail.

PR #2666, opened 2017-12-12
Remove -y option from mix phx.new command

cvortmann:
That option gives an error message:

```
** (Mix) Invalid option: -y
```


Gazler:
Thanks <3 - not sure how that ended up in there.

PR #2669, opened 2017-12-12
Fix for dializer with `build_conn`

OvermindDL1:
Currently the typespec for `build_conn` requires passing in the params as a binary/list/map, yet a nil is being passed in as the default and as it is not the list nil of `[]` but rather the atom `nil` it is making dialyzer very unhappy, fixed to the empty nil list of `[]` makes dialyzer happy again.


chrismccord:
❤️❤️❤️🐥🔥

PR #2670, opened 2017-12-12
Match error message key and value to what's expected in test

cvortmann:
Also reformatted some snippets to match a following example and the community style guide for [multiline pipelines](https://github.com/christopheradams/elixir_style_guide#pipe-operator).

### DIFF
--- a/guides/docs/testing/testing_controllers.md
+++ b/guides/docs/testing/testing_controllers.md
@@ -270,7 +270,7 @@ test "Responds with user info if the user is found", %{conn: conn} do
     |> get(user_path(conn, :show, user.id))
     |> json_response(200)
 
-  expected = %{ "data" => %{"email" => user.email, "name" => user.name} }
+  expected = %{"data" => %{"email" => user.email, "name" => user.name}}
 
   assert response == expected
 end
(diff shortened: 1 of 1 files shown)


guides/docs/testing/testing_controllers.md
```diff
@@ -431,7 +435,7 @@ def show(conn, %{"id" => id}) do
     nil ->
       conn
       |> put_status(:not_found)
-      |> json(%{error: "User not found"})
+      |> json(%{errors: "User not found."})
```
josevalim: It doesn't make sense to say errors and then return a string. Maybe better to revert the change to this line for now?

cvortmann: @josevalim it's only matching what is in the test:

```elixir
test "Responds with a message indicating user not found", %{conn:  conn} do
  response =
    conn
    |> get(user_path(conn, :show, -1 ))
    |> json_response(404)

  expected = %{ "errors" => "User not found." }

  assert response == expected
end
```

I can update both to return only a string message, what you think?

josevalim: I see. Can we fix the test in this case? We should also remove the spaces after %{ and before }. Tks!

cvortmann: I just realize that, by [JSON definition](http://json.org/), the API must either return a Javascript _Object_ or an _Array_. If we move to returning just *text* we may need to also change the text of the guide.

I'd rather then make that change in a separate PR, so I can review the text together with the code and make sure it makes sense.

This one still stay "weird" but makes the tests green 😄

josevalim:

PR #2674, opened 2017-12-14
Use text response for show test when user not found

cvortmann:
As per @josevalim comment in https://github.com/phoenixframework/phoenix/pull/2670.


cvortmann:
@josevalim is there a way I can label these PRs with the **guides** label?

chrismccord:
❤️❤️❤️🐥🔥

PR #2678, opened 2017-12-15
Include get_author!/1 function on definitions

cvortmann:
The snippet included changes to `get_author!/1` but only mentioned after the snippet was presented.


cvortmann:
@Gazler 👍 . Thanks, I would have never seen it 😄

PR #2679, opened 2017-12-15
Fix "phoenix_live_reload/issues/67"

zhengxiaoyao0716:
Sometimes the built file will be empty after live reload. [#67](https://github.com/phoenixframework/phoenix_live_reload/issues/67)

### DIFF
--- a/lib/phoenix/code_reloader/server.ex
+++ b/lib/phoenix/code_reloader/server.ex
@@ -27,9 +27,10 @@ defmodule Phoenix.CodeReloader.Server do
     if not checked? and Code.ensure_loaded?(Mix.Project) do
       priv_path = "#{Mix.Project.app_path}/priv"
 
-      case File.read_link(priv_path) do
+      case :file.read_link(priv_path) do
         {:ok, _} ->
           :ok
+
         {:error, _} ->
           if can_symlink?() do
             File.rm_rf(priv_path)
--- a/lib/phoenix/code_reloader/server.ex
+++ b/lib/phoenix/code_reloader/server.ex
@@ -26,17 +26,17 @@ defmodule Phoenix.CodeReloader.Server do
   def handle_call(:check_symlinks, _from, checked?) do
     if not checked? and Code.ensure_loaded?(Mix.Project) do
       priv_path = "#{Mix.Project.app_path}/priv"
+
       case File.read_link(priv_path) do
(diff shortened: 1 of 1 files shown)


lib/phoenix/code_reloader/server.ex
```diff
@@ -78,10 +80,27 @@ defmodule Phoenix.CodeReloader.Server do
...
+    do: " On Windows, you can started the shell with \"Run as Administrator\" at least once time, " <> 
+        "to allow phoenix create symlinks and resolve this problem."
   defp os_symlink(_),
     do: ""
 
+  defp can_symlink() do
+    build_path = Mix.Project.build_path()
+    symlink = Path.join(Path.dirname(build_path), "__phoenix__")
+
+    case File.ln_s(build_path, symlink) do
```
josevalim: Isn't this a recursive symlink? I don't think there is anything wrong with it but it would probably be best to avoid it. Maybe we should try to symlink app_path/priv to priv, which is the symlink we would build at the end of the day.

lib/phoenix/code_reloader/server.ex
```diff
@@ -78,10 +80,27 @@ defmodule Phoenix.CodeReloader.Server do
   end
 
   defp os_symlink({:win32, _}),
-    do: " On Windows, such can be done by starting the shell with \"Run as Administrator\"."
+    do: " On Windows, you can started the shell with \"Run as Administrator\" at least once time, " <> 
+        "to allow phoenix create symlinks and resolve this problem."
   defp os_symlink(_),
     do: ""
 
+  defp can_symlink() do
```
josevalim: We can name this function `can_symlink?` and have it return a boolean.

lib/phoenix/code_reloader/server.ex
josevalim: Ignore me, we are using the same link as before. :)

zhengxiaoyao0716: Yes, and I think this is not the best way too. Those codes that create a temp symlink and then remove them maybe just a opportunistic way.
What the `can_symlink` function actually need to do is just, find that is the task are running 
 with `Administrator` and have permission to create symlinks.
But since those codes exist in the old version and the have works well, I became lazy to find other solutions, just move them into the `can_symlink` function.

PR #2697, opened 2017-12-29
Fix non-working example usage in phx.server doc

### LINKED ISSUE #2690: `mix phx.server --no-deps-check` still checks dependencies
### Environment

* Elixir version (elixir -v): 1.5.3
* Operating system: macOS 10.12.6

### Expected behavior

Create a new Phoenix app:

```
$ mix phx.new my_app
…
Fetch and install dependencies? [Yn] Y
…
$ cd my_app
```

Nuke one of the dependencies:

```
$ rm -rf deps/poison
```

Start the app with `--no-deps-check`, as mentioned in the doc for `mix phx.server`:

```
$ mix phx.server --no-deps-check
```

The app should start (and eventually fail when it tries to use the broken dependency).

### Actual behavior

Mix refuses to start the app, as if `--no-deps-check` had not been specified:

```
$ mix phx.server --no-deps-check
Unchecked dependencies for environment dev:
* poison (Hex package)
  the dependency is not available, run "mix deps.get"
** (Mix) Can't continue due to errors on dependencies
```

It appears Mix automatically invokes the `deps.loadpaths` task, but without passing on the arguments:

```
$ MIX_DEBUG=1 mix phx.server --no-deps-check
** Running mix loadconfig (inside Hellophoenix145.Mixfile)
** Running mix phx.server --no-deps-check (inside Hellophoenix145.Mixfile)
** Running mix deps.loadpaths (inside Hellophoenix145.Mixfile)
Unchecked dependencies for environment dev:
* poison (Hex package)
  the dependency is not available, run "mix deps.get"
** (Mix.Error) Can't continue due to errors on dependencies
    (mix) lib/mix.ex:291: Mix.raise/1
    (mix) lib/mix/tasks/deps.loadpaths.ex:26: Mix.Tasks.Deps.Loadpaths.run/1
    (mix) lib/mix/task.ex:301: Mix.Task.run_task/3
    (mix) lib/mix/task.ex:324: Mix.Task.get_task_or_run/3
    (mix) lib/mix/task.ex:283: Mix.Task.run_task/3
    (mix) lib/mix/cli.ex:75: Mix.CLI.run_task/2
```

As it turns out, Mix needs to do `deps.loadpaths` for the task `phx.server` to even be available, so it looks like by the time execution reaches `Mix.Tasks.Phoenix.Server#run`, it is too late to disable the dependency check. A workaround is to explicitly invoke `deps.loadpaths` before `phx.server`:

```
$ mix do deps.loadpaths --no-deps-check, phx.server
[info] Running Hellophoenix145Web.Endpoint with Cowboy using http://0.0.0.0:4000
```

I don't know if there is a way to make `phx.server --no-deps-check` work as expected, but I think at least it should not be the example given in the documentation given (unless I'm missing something) it does not have the desired effect.

jonathanperret:
The `--no-deps-check` flag has no effect when used with the `phx.server` task, because Mix always runs the dependency check before loading tasks from dependencies. This replaces the `--no-deps-check` example to a working one using `--no-compile`, and suggests a workaround to run `phx.server` without checking dependencies.

Fixes #2690.

### DIFF
--- a/lib/mix/tasks/phx.server.ex
+++ b/lib/mix/tasks/phx.server.ex
@@ -12,11 +12,18 @@ defmodule Mix.Tasks.Phx.Server do
   For additional information, refer to the documentation for
   `Mix.Tasks.Run`.
 
-  For example, to run `phx.server` without checking dependencies:
+  For example, to run `phx.server` without recompiling:
 
-      mix phx.server --no-deps-check
+      mix phx.server --no-compile
 
   The `--no-halt` flag is automatically added.
+
+  Note that the `--no-deps-check` flag cannot be used this way,
+  because Mix needs to check dependencies to find `phx.server`.
+
+  To run `phx.server` without checking dependencies, you can run:
+
+      mix do deps.loadpaths --no-deps-check, phx.server
   """
 
   @doc false


josevalim:

PR #2698, opened 2017-12-30
Removed reasigned conn's in phx.gen.json controller test [Proposal]

Hajto:
Since I am probably abusing generators I often encountered a problem when generated test would fail because conn was overwritten.

I often use Guardian or other similar library to handle user authentication. Most of the time needed data is carried somewhere in `conn` struct. In current form conn is overwritten which often means that mandatory data is lost when conn is being reused. (`Guardian`'s case)

That's why I often have to change their names to not collide and to always blank conn is being used. I also do think overwriting params is bad practice so this change would carry over some good practices.

If you think this change could be beneficial conducting similar operation on `gen.html` would also make sense.

This is not a huge deal but it would make life easier (at least for me).


chrismccord:
I prefer the rebinding in this case but you are free to write your tests without rebinding as needed. Thanks!

Issue #2704, opened 2018-01-09
Unable to reply asynchronously to a socket push with just an atom status


van-mronov:
### Environment

* Elixir version (elixir -v): 1.5.2
* Phoenix version (mix deps): 1.3.0
* Operating system: macOS

### Expected behavior
`Phoenix.Channel.reply/2` replies with an empty payload in case of the second argument is an atom (represented status).

### Actual behavior
According to the documentation for `Phoenix.Channel.reply/2` the type of the second argument is an atom or a tuple: `@type reply :: status :: atom | {status :: atom, response :: map}`.
But it supports only tuple as reply now and fails if pass an atom:
`(FunctionClauseError) no function clause matching in Phoenix.Channel.reply/2`

### Possible solutions
Reply with an empty payload like `handle_in` callbacks.

josevalim:
Can you please provide the code you are trying out as well as the full message with stacktrace?

josevalim:
Nevermind, I just saw the PR. ;)

PR #2705, opened 2018-01-09
Support an atom as reply in `Phoenix.Channel.reply/2` (#2704)

van-mronov:
* Reply with an empty payload in case of the second arg is an atom
* Add test for reply/2 replying just status to transport


josevalim:

PR #2706, opened 2018-01-09
Honor ExUnit default receive timeouts in Channel tests

GregMefford:
When running in CI, sometimes the tests run slowly and cause these Channel tests to fail simply because it took more than 100 ms for the reply to come. This makes the tests almost always work fine locally but flakey in CI.

When using a standard ExUnit `assert_receive`, the default timeout can be adjusted to 1000ms like so:

```elixir
config :ex_unit, :assert_receive_timeout, 1000
```

Since these Channel test macros are always explicitly passing the timeout to the underlying `assert_receive` or `refute_receive`, we lose the option to configure the default globally and have to set a higher default on each test individually.

This PR honors the same `config :ex_unit, :assert_receive_timeout` configuration setting so that [the explicit value that's passed down is set to the default value](https://github.com/elixir-lang/elixir/blob/v1.5.3/lib/ex_unit/lib/ex_unit/assertions.ex#L353). This feels like the path of least surprise to me, but does couple Phoenix to ExUnit in a way that might not be appropriate. It would also be possible to define another `defmacro assert_push(event, payload)` head that _doesn't_ explicitly set the timeout, but I think that would require more duplication of code.

### DIFF
--- a/lib/phoenix/test/channel_test.ex
+++ b/lib/phoenix/test/channel_test.ex
@@ -436,7 +436,8 @@ defmodule Phoenix.ChannelTest do
   In the assertion above, we don't particularly care about
   the data being sent, as long as something was sent.
 
-  The timeout is in milliseconds and defaults to 100ms.
+  The timeout is in milliseconds and defaults to the :assert_receive_timeout
+  set on the :ex_unit application (which defaults to 100ms).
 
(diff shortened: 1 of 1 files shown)


josevalim:

PR #2708, opened 2018-01-14
Document the need to restart the server

sharils:
Hot-reload doesn't work for the for channel


sharils:
I also found that http://phoenixchat.herokuapp.com/ is showing an error. 
image where should I report that?

josevalim:
If you refresh the page, it should definitely trigger the channel to
recompile.
-- 

PR #2709, opened 2018-01-14
Update docs

sharils:


guides/docs/channels.md
```diff
@@ -292,7 +292,10 @@ Now our `conn.assigns` contains the `current_user` and `user_token`.
 Next we need to pass this token to JavaScript. We can do so inside a script tag in `web/templates/layout/app.html.eex`, as follows:
 
 ```html
-<script>window.userToken = "<%= assigns[:user_token] %>";</script>
+<head>
+  ...
+  <script>window.userToken = "<%= assigns[:user_token] %>";</script>
+</head>
```
chrismccord: We don't need the head tag in this case, so can you drop this change? If we are going to make a change here we should show the script tag appears above the app.js script, which is located before </body

sharils: Hi @chrismccord - No problem and here you are! I've dropped that commit and added another version right above the app.js script. My goal is to make it clear about where to add the script tag. When I went through the example, it didn't work at the first time because I added it to be right before the `</body>` tag which is after the app.js script thus it didn't work out then I switched to the `<head>` version. I believe making this clear is going to help other people in the future. :) Thanks for pointed out a version better than the `<head>` version. I didn't like that myself.

Gazler:
Thanks <3

guides/docs/channels.md
chrismccord: perfect! Thanks!

Issue #2712, opened 2018-01-15
Phoenix instrumenters and http status code inside phoenix_controller_call


kpanic:
hi everyone,

### Environment

* Elixir version (elixir -v): Elixir 1.6.0-rc.0 (182c730) (compiled with OTP 19)
* Phoenix version (mix deps): phoenix 1.3.0
* Operating system: linux

### Description

When trying to use the instrumenters feature, if I create an instrumenter like:

```elixir
defmodule Instrumenter.Example.My do
  def phoenix_controller_call(:start, _compile, conn) do
    conn
  end

  def phoenix_controller_call(:stop, _time_diff, %{conn: conn}) do
    # conn.status is always `nil` at this stage in the case of
    # phoenix_controller_call
    # However for phoenix_controller_render the conn.status will be there
    IO.inspect(conn.status, label: "http_status from the instrumenter")
    conn
  end
end
```
and I add it to the endpoint configuration with the `instrumenters: [Instrumenter.Example.My]` configuration.
when an http endpoint is hit then the `phoenix_controller_call/3` functions are invoked correctly, however `conn.status` will be always `nil` in the `:stop` stage. Is this intended?

This does not  happen when using `phoenix_controller_render/3`
Thanks!

chrismccord:
This seems like a bug. Let me take a look. Thanks!

chrismccord:
I see the confusion here, but in this case what is happening is by design. The instrumenter passes the return value from the `:start` callback to the last arg of the `:stop` callback as a means of passing information between the two, but the `conn` in this case is the *original* connection, not the plug connection after the controller has finished executing. Our instrumentation today doesn't support receiving the value of the executed block being instrumented, so you'll need to use something like `Plug.Conn.registered_before_send` to receive e the status of the conn.

kpanic:
I imagined that, however I wanted to double check. 
In my project I "worked around" by using pattern matching `phoenix_controller_render/3` which contains the http status in the conn struct when we are at `:stop` "stage"

Good to know about `Plug.Conn.registered_before_send`!

Also FYI I thought (initially) that the `phoenix_controller_call/3` would invoke fully a controller's function, that's what confused me indeed.
Do you think it would be worth to improve the phoenix documentation?

Thanks a lot!

PR #2714, opened 2018-01-15
Elaborate more about phoenix_controller_call doc

kpanic:
hi, this is a try to improve the documentation of `:phoenix_controller_call`
See issue https://github.com/phoenixframework/phoenix/issues/2712 for more context.

If you have better wording of the **NOTE** in this PR, please drop a comment.
Also please bear with me since I am not an English native speaker :)

If you think that it's too detailed this **NOTE**, then I will close the PR

Thanks!

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -314,6 +314,14 @@ defmodule Phoenix.Endpoint do
 
     * `:phoenix_controller_call` - it's the whole controller pipeline.
       The `%Plug.Conn{}` is passed as runtime metadata.
+
+      **NOTE:** `%Plug.Conn{}` will be passed as runtime metadata to the
+        corresponding `phoenix_controller_call/3` in the "after" event callback
+        from the "before" event callback.
+
+        This means that some `%Plug.Conn{}` values might be not populated, for
+        example `%Plug.Conn{}.status` will not be updated even if the whole
+        controller pipeline will be executed.
     * `:phoenix_controller_render` - the rendering of a view from a
       controller. The map of runtime metadata passed to instrumentation
       callbacks has the `:view` key - for the name of the view, e.g. `HexWeb.ErrorView`,


chrismccord:
Thanks for the PR! In this case I think the note is too specific since the issue you ran into was confusion on what is passed from the before call to the after call. In your case, you were passing a piece of the runtime metadata down to your after clause, which just happened to be a map with %Plug.Conn{} in it. If you'd like to take a 2nd pass, you can clarify the behavior in the `# Instrumentation` section, otherwise feel free to close. I appreciate the PR!

kpanic:
Hey Chris,  thanks for the answer :) From my understanding the instrumenter for `phoenix_controller_call` is defined here https://github.com/phoenixframework/phoenix/blob/2d19cac62cdb54a59e85c08a68798e1b8714adbb/lib/phoenix/controller/pipeline.ex#L26

And `phoenix_controller_call` suggested to me that I would be able to hook into `:start` / `:stop` "events" and I would have the `conn` updated *after* the execution of `phoenix_controller_call(:start, ...)` -- so in `phoenix_controller_call(:stop, time_diff, data)` I would have the time difference of the whole controller execution and `data` which, in this case, I would expect that it would be an updated `conn` structure with the proper `conn.status`.

That was my first  interpretation. Would you please shed some light on this interpretation? That would help  me to refine the doc I think.

thanks for the patience :)

chrismccord:
Where you are going wrong is the `:stop` callback receives the value you return from the `:start` callback, so it's entirely controller by your `:start` implementation.

kpanic:
mmm but this `instrument` call is done at phoenix level. and it's phoenix that is passing the `conn` if I am not mistaken

chrismccord:
to be clear, Phoenix passes the conn in the runtime metadata to your `:start` callbacks. What those callbacks return (your code), is passed to the `:stop` callbacks.

kpanic:
yeah, this I understood, thank you.
I am just wondering if this makes sense (to not have an updated `conn`), from the functionality point of view, in those callbacks

chrismccord:
The instrumenter doesn't pass the result of the code because it may raise and the exception is caught, so we can't guarantee `:stop` values should we try to pass them

Issue #2031, opened 2016-12-12
Nice to have: Record uninterpolated route in `conn`


palm86:
For metrics purposes (having Exometer in mind) it would be nice to know which route was matched in addition to the actual URI that was hit.

In other words, both `GET api/entities/1` and `GET api/entities/2` might match the route `GET api/entities/:id`, but that fact is not recorded in the `conn`. There are of course several less than ideal ways of establishing which route was matched, but recording it as it is matched seems the simplest (if that is possible).

I'd be happy to give it a shot, if you feel such an addition could be useful.

josevalim:
To be clear, you want to know the pattern that was match on? Why that would be useful to record?

palm86:
Yes, that is the idea.

To record endpoint hits with Exometer one could do this:

`:exometer.update([:myapp, :api, conn.method, conn.request_path], 1)`

but if `conn.request_path` contains variables like `:id`, it would look like a separate endpoint was hit for every value of `id`. In some cases that may be desirable, but I think in the most common case, you would want to consider requests to `GET api/entity/1` and `GET api/entity/2` as having hit the same endpoint.

I realize that using the controller name and action gives similar information, so one could do this:

`:exometer.update([:myapp, :api, conn.private.phoenix_controller, conn.private.phoenix_action], 1)`

(inspect or string conversion omitted for clarity)

but multiple routes could point to the same controller and action and different plugs may have been applied.

josevalim:
> I realize that using the controller name and action gives similar information

That's exactly what I would suggest. controller+action+verb should be enough to gather the uniqueness information you need for your exometer reports.

palm86:
My concern in that case, in addition to the ones provided in the previous post, is that it pollutes my metrics with implementation details.

franc:
I can imagine having different routes pointing to the same action. 
If one is being deprecated, I would like to know when I can safely kill the old route.

`/api/horrible/:id/endpoint` vs `/api/v2/better/:id`

Also, could be a simple way of doing A/B testing analysis

josevalim:
@palm86 I would argue "/api/v2/better/:id" is also an implementation detail. The path and verb, however, are not as they are part of the request.

@franc If the goal is to do A/B testing, I would prefer to use an explicit API than implicitly rely on the route pattern. For example:

```elixir
get "/api/v2/better/:id", Controller, :action, assigns: %{ab: {:experiment, :a}}
[...]

PR #2716, opened 2018-01-17
Expose router path as defined in router

aquarhead:
I'm interested in this when someone posted a related question in #prometheus channel on elixir-lang slack, because currently we're doing something similar (for metrics) in a (non-Elixir) project. But I just realized [the exact situation has been discussed a year ago](https://github.com/phoenixframework/phoenix/issues/2031)... I'll submit this PR anyway since it's a relatively simple change and I think this is the only way for Instrumenter (e.g. [prometheus-phoenix](https://github.com/deadtrickster/prometheus-phoenix)) to have access to this information.

Posting this in case anything changed since a year ago. And also if anyone else needs such functionality this can serve as a reference or inspiration.

### DIFF
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -264,13 +264,14 @@ defmodule Phoenix.Router do
   end
 
   @doc false
-  def __call__({%Plug.Conn{private: %{phoenix_router: router, phoenix_bypass: {router, pipes}}} = conn, _pipeline, _dispatch}) do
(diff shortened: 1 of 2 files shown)


chrismccord:
This is not something we have exposed for the reasons discussed in the previous issue. We'd need more work than this to expose a public data structure, and it would cause more issues for us in the future since we would be locked to the public API. Since most of the information you need for metrics is available by lifting other data off the conn, I'm not yet ready to go this route. Thanks!

PR #2721, opened 2018-01-18
Conditionally generate child specs based on elixir version

### LINKED ISSUE #2713: Conditionally generate new Elixir 1.5 child spec
The project generator will check the Elixir version and produce the new format. We will officially continue to support back to Elixir 1.3.

chrismccord:
closes #2713

### DIFF
--- a/installer/lib/phx_new/generator.ex
+++ b/installer/lib/phx_new/generator.ex
@@ -102,6 +102,7 @@ defmodule Phx.New.Generator do
       end
 
     binding = [
+      elixir_version: elixir_version(),
       app_name: project.app,
       app_module: inspect(project.app_mod),
       root_app_name: project.root_app,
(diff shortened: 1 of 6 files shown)


installer/lib/phx_new/generator.ex
```diff
@@ -127,11 +128,31 @@ defmodule Phx.New.Generator do
...
       adapter_config: adapter_config,
       generators: nil_if_empty(project.generators ++ adapter_generators(adapter_config)),
-      namespaced?: namespaced?(project)]
+      namespaced?: namespaced?(project),
+    ]
 
     %Project{project | binding: binding}
   end
 
+  def stub_elixir_version(vsn, func) do
```
josevalim: Shouldn't this be moved to the test directory?

installer/lib/phx_new/generator.ex
```diff
@@ -127,11 +128,31 @@ defmodule Phx.New.Generator do
...
+    if configured_version() do
+      raise RuntimeError, "race condition detected when trying to stub elixir version"
+    end
+
+    try do
+      Application.put_env(:phoenix, :installer_elixir_vsn, vsn)
```
josevalim: We should probably read from `:phx_new` since the app in this case is `:phx_new`.

josevalim:
Two tiny comments. Everything else is perfect!

Issue #2605, opened 2017-10-25
Provide integration with Elixir v1.6 formatter


josevalim:
Depends on https://github.com/elixir-lang/elixir/issues/6644.

michallepicki:
As elixir approaches 1.6.0 I tried the formatter on a default `phx.new` project [in this commit](https://github.com/michallepicki/learning_dialyzer_with_phoenix/commit/cebb509eed45afbbe1b7d65927200ad545c6dfd2). I also looked at one of my projects. At first glance, questionable parens:
- `socket`, `recv`
- `plug`, `pipe_through`, `forward`, `match`, `resources`
- `get`, `post`, `put`, `delete`
- `render`, `json`, `text`, `html`
- Ecto's `add`, `create`, `create_if_not_exists`, `drop`, `drop_if_exists`, `execute`,  `modify`, `remove`, `rename`

chrismccord:
Yes we will need special formatter config for new projects for DSL macros such as the router

josevalim:
@michallepicki a PR is definitely appreciated. I don't think `render`, `json`, `text` and `html` should make the cut though. I would mostly restrict to module-level DSLs and testing concepts.

The ones about Ecto should be added to Ecto. A PR would definitely be appreciated.

michallepicki:
Sure, I'm on it.

michallepicki:
I compiled an initial list of functions [here](https://github.com/michallepicki/phoenix/commit/ca504570b33a8a3b4303d15c2a5821143cc988cf). I followed the Phoenix Docs and the way all of the functions were written there (I skipped some cases where I assumed the Docs just didn't get updated).
1. I am still unsure about whether we shouldn't also include `Phoenix.Controller`'s `html` etc, `Phoenix.View`'s `render`s,  `Phoenix.ChannelTest` and `Phoenix.ConnTest` replicas of `socket`, `broadcast`, `get`, `put` (but in this case we already have some of them on the list and only arity may sometimes differ so I can include these).
2. I am generally (a little bit) disappointed with the formatter and the fact that I can't specify an exact function by providing the full name with package and module, e.g. `"Phoenix.Router.forward": 2` because of possible function name overlaps.
3. The things that (I assume) are left to do:
   - Do the same in Ecto and Plug
   - Generate a `.formatter.exs` with `phx.new` which imports appropriate settings from `:ecto`, `:phoenix` depending on task arguments

chrismccord:
[...]

PR #2722, opened 2018-01-19
Add .formatter import from ecto

taiansu:
For Elixir 1.6 format feature, [Ecto will export formatter config](https://github.com/elixir-ecto/ecto/pull/2385) in the future. I propose to add the `.formatter.exs` to import the configuration accordingly while generating the `Single` and `Web` projects.

I've done some experiment that `mix format` feature seems work as normal while Ecto doesn't have the `.formatter.exs`, in other words, not exporting the configuration.

### DIFF
--- a/installer/lib/phx_new/single.ex
+++ b/installer/lib/phx_new/single.ex
@@ -66,6 +66,7 @@ defmodule Phx.New.Single do
 
   template :bare, [
     {:text, "phx_assets/bare/gitignore", :project, ".gitignore"},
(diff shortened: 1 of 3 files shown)


taiansu:
Uppps. I forgot to add the testing part, will open another one.

josevalim:
Also please see #2605 and coordinate with @michallepicki :)

taiansu:
Sure, thanks a lot!

PR #2724, opened 2018-01-23
Update the Cowboy 2 dep to be >= 2.2.1

### LINKED ISSUE #2686: Pages over a certain length served with H2 hang in Safari (and _just_ Safari)
If you setup a new Phoenix project from master with H2 support and create a page that's length is over 18000 chars the page will hang when loading in Safari.

The server sends back a 200 OK (at least that is logged) but Safari never displays the data. Chrome works, Firefox works, curl works. It's *just* Safari. And only in production.

The thing that has me perplexed is that the failure *only* happens when the `MIX_ENV=prod`. If you run in dev mode the page loads. It only fails in production mode.

I have created a repo with a project that reproduces this issue.

Grab it here:
   https://github.com/idyll/snowday

If you run the above repo you will see that going to the following two pages will hang Safari:

 - https://snowday.phx.sh/test/one
 - https://snowday.phx.sh/test/two

However going to these two smaller pages works:

 - https://snowday.phx.sh/test/three
 - https://snowday.phx.sh/test/four

And again, it works in dev. To replicate `MIX_ENV=prod mix phx.server`

I have an open radar with Apple but I am not convinced this is 100% on Safari. (This is not a Webkit issue, the Webkit team doesn't handle the H2 part.)

idyll:
Cowboy 2.2.1 includes a fix that addresses an H2 issue for the Safari web browser.

Fixes #2686
(there is a discussion of the issue there along with sample code that demonstrates the problem)

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -75,7 +75,7 @@ defmodule Phoenix.Mixfile do
   end
 
   defp cowboy_dep("1" <> _), do: {:cowboy, "~> 1.0", optional: true}
-  defp cowboy_dep(_), do: {:cowboy, ">= 2.2.1", optional: true}
+  defp cowboy_dep(_), do: {:cowboy, "~> 2.2.1 or ~> 2.3", optional: true}
 
   defp lockfile() do
     case System.get_env("COWBOY_VERSION") do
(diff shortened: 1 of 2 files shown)


josevalim:

PR #2727, opened 2018-01-27
Raise an exception for unknown actions found only/except scope

### LINKED ISSUE #2723: Raise on unknown :only and :except options in resources
Otherwise you may try to select or skip an action that does not exist and it can be quite puzzling why it doesn't work.

See original discussion here: https://stackoverflow.com/questions/48358986/cant-create-a-nested-resource-in-phoenix/48361577#48361577

starbelly:
Closes #2723

Initial pass. 

 Notes:

 1. Raising ArgumentError, but maybe something else is in order
 2. The error message surely could be better
 3. I formatted if the router out of habit now 👍 ... but this may be undesirable right now, if so, let me know and I will rollback and re-patch the file. 
 4. Did not feel this specifically needed to be documented, but perhaps I am wrong.

### DIFF
--- a/lib/phoenix/router/resource.ex
+++ b/lib/phoenix/router/resource.ex
@@ -21,41 +21,79 @@ defmodule Phoenix.Router.Resource do
     * :collection - the context for collection routes
 
   """
-  defstruct [:path, :actions, :param, :route, :controller, :route, :member, :collection, :singleton]
+  defstruct [
+    :path,
+    :actions,
+    :param,
+    :route,
+    :controller,
+    :route,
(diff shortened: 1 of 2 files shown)


josevalim:
@starbelly yes, please don't format the file. it makes it very hard to review the code and it will make it hard to track changes. it should be done in a separate commit and in a separate pull request. :)

chrismccord:
Closed via https://github.com/phoenixframework/phoenix/commit/54a5b691430dbfc9cf882351dcb8c02ef9797a29

PR #2734, opened 2018-01-30
Use Jason instead of Poison for json encoding

### LINKED ISSUE #2693: Investigate using Jason instead of Poison
This will also require changing phoenix_ecto / ecto.

chrismccord:
Closes #2693

### DIFF
--- a/installer/templates/phx_single/config/config.exs
+++ b/installer/templates/phx_single/config/config.exs
@@ -25,8 +25,7 @@ config :logger, :console,
   metadata: [:request_id]
 
 # Use Jason for JSON parsing in Phoenix<%= if ecto do %> and Ecto<% end %>
-config :phoenix, :format_encoders, json: Jason
-<%= if ecto do %>
+config :phoenix, :json_library, Jason<%= if ecto do %>
 config :ecto, :json_library, Jason
 <% end %>
 # Import environment specific config. This must remain at the bottom
--- a/installer/templates/phx_umbrella/apps/app_name/config/config.exs
+++ b/installer/templates/phx_umbrella/apps/app_name/config/config.exs
@@ -1,6 +1,6 @@
 use Mix.Config
 
-<%= if ecto do %>config :<%= app_name %>, ecto_repos: [<%= app_module %>.Repo]<% end %>
-<%= if ecto do %>config :ecto, :json_library, Jason<% end %>
+<%= if ecto do %>config :<%= app_name %>, ecto_repos: [<%= app_module %>.Repo]
+config :ecto, :json_library, Jason<% end %>
 
 import_config "#{Mix.env}.exs"
--- a/lib/phoenix.ex
+++ b/lib/phoenix.ex
@@ -55,12 +55,21 @@ defmodule Phoenix do
     Supervisor.start_link(children, strategy: :one_for_one, name: Phoenix.Supervisor)
   end
 
-  @doc false
-  # TODO remove Poison default in 2.0
-  def json do
-    :phoenix
-    |> Application.fetch_env!(:format_encoders)
-    |> Keyword.get(:json, Poison)
+  # TODO: swap Poison default with Jason in 2.0
+  # from there we can ditch explicit config for new projects
+  @doc """
+  Returns the configured JSON encoding library for Phoenix.
+
+  To customize the JSON library, including the following
+  in your `config/config.exs`:
+
+      config :phoenix, :json_library, Jason
+  """
+  def json_library do
+    case Application.fetch_env(:phoenix, :json_library) do
+      {:ok, module} -> module
+      :error -> Poison
+    end
   end
 
   defp warn_on_missing_format_encoders do
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -266,7 +266,7 @@ defmodule Phoenix.Controller do
   @doc """
   Sends JSON response.
 
-  It uses the configured `:format_encoders` under the `:phoenix`
+  It uses the configured `:json_library` under the `:phoenix`
   application for `:json` to pick up the encoder module.
 
   ## Examples
@@ -276,7 +276,7 @@ defmodule Phoenix.Controller do
   """
   @spec json(Plug.Conn.t, term) :: Plug.Conn.t
   def json(conn, data) do
-    response = Phoenix.json().encode_to_iodata!(data)
+    response = Phoenix.json_library().encode_to_iodata!(data)
     send_resp(conn, conn.status || 200, "application/json", response)
   end
 
--- a/lib/phoenix/template.ex
+++ b/lib/phoenix/template.ex
@@ -93,8 +93,7 @@ defmodule Phoenix.Template do
   New encoders can be added via the format encoder option:
 
       config :phoenix, :format_encoders,
-        html: Phoenix.Template.HTML,
-        json: Jason
+        html: Phoenix.Template.HTML
 
   """
 
@@ -251,7 +250,7 @@ defmodule Phoenix.Template do
   end
 
   defp default_encoders do
-    [html: Phoenix.Template.HTML, json: Phoenix.json(), js: Phoenix.Template.HTML]
+    [html: Phoenix.Template.HTML, json: Phoenix.json_library(), js: Phoenix.Template.HTML]
   end
 
   @doc """
--- a/lib/phoenix/transports/v2/longpoll_serializer.ex
+++ b/lib/phoenix/transports/v2/longpoll_serializer.ex
@@ -12,7 +12,7 @@ defmodule Phoenix.Transports.V2.LongPollSerializer do
     encoded =
       %Message{topic: msg.topic, event: msg.event, payload: msg.payload}
       |> to_list()
-      |> Phoenix.json().encode_to_iodata!()
+      |> Phoenix.json_library().encode_to_iodata!()
 
     {:socket_push, :text, encoded}
   end
@@ -23,7 +23,7 @@ defmodule Phoenix.Transports.V2.LongPollSerializer do
   Encoding is handled downstream in the LongPoll controller.
   """
   def encode!(msg) do
-    {:socket_push, :text, msg |> to_list() |> Phoenix.json().encode_to_iodata!()}
+    {:socket_push, :text, msg |> to_list() |> Phoenix.json_library().encode_to_iodata!()}
   end
   defp to_list(%Reply{} = reply) do
     [reply.join_ref, reply.ref, reply.topic, "phx_reply",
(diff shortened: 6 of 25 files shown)


guides/docs/controllers.md
```diff
@@ -131,7 +131,7 @@ end
 ```
 Assuming we had a route for `get "/our_path/:id"` mapped to this `show` action, going to `/our_path/15` in your browser should display `Showing id 15` as plain text without any HTML.
 
-A step beyond this is rendering pure JSON with the `json/2` function. We need to pass it something that the [Poison library](https://github.com/devinus/poison) can parse into JSON, such as a map. (Poison is one of Phoenix's dependencies.)
+A step beyond this is rendering pure JSON with the `json/2` function. We need to pass it something that the [Jason library](https://github.com/michalmuskala/jason) can parse into JSON, such as a map. (Jason is one of Phoenix's dependencies.)
```
michalmuskala: Should this be "encode" instead of "parse"?

lib/phoenix/transports/v2/longpoll_serializer.ex
```diff
@@ -12,7 +12,7 @@ defmodule Phoenix.Transports.V2.LongPollSerializer do
     encoded =
       %Message{topic: msg.topic, event: msg.event, payload: msg.payload}
       |> to_list()
-      |> Poison.encode!()
+      |> Jason.encode!()
```
michalmuskala: We might use `encode_to_iodata!` here as well, like in the websocket serializer.

lib/phoenix/controller.ex
```diff
@@ -265,7 +265,7 @@ defmodule Phoenix.Controller do
 
   defp get_json_encoder do
     Application.get_env(:phoenix, :format_encoders)
-    |> Keyword.get(:json, Poison)
+    |> Keyword.get(:json, Jason)
```
josevalim: This is backwards incompatible, I am afraid.

test/phoenix/transports/websocket_serializer_test.exs
```diff
@@ -5,10 +5,10 @@ defmodule Phoenix.Tranports.WebSocketSerializerTest do
...
-  @v1_fastlane_json [123, [[34, ["topic"], 34], 58, [34, ["t"], 34], 44, [34, ["ref"], 34], 58, "null", 44, [34, ["payload"], 34], 58, [34, ["m"], 34], 44, [34, ["event"], 34], 58, [34, ["e"], 34]], 125]
-  @v2_msg_json [91, ["null", 44, "null", 44, [34, ["t"], 34], 44, [34, ["e"], 34], 44, [34, ["m"], 34]], 93]
+  @v1_msg_json ["{\"", [[], "event"], "\":", [34, [], "e", 34], ",\"", [[], "payload"], "\":", [34, [], "m", 34], ",\"", [[], "ref"], "\":", "null", ",\"", [[], "topic"], "\":", [34, [], "t", 34], 125]
+  @v1_reply_json ["{\"", [[], "event"], "\":", [34, [], "phx_reply", 34], ",\"", [[], "payload"], "\":", ["{\"", [[], "response"], "\":", "null", ",\"", [[], "status"], "\":", "null", 125], ",\"", [[], "ref"], "\":", [34, [], "null", 34], ",\"", [[], "topic"], "\":", [34, [], "t", 34], 125]
+  @v1_fastlane_json ["{\"", [[], "event"], "\":", [34, [], "e", 34], ",\"", [[], "payload"], "\":", [34, [], "m", 34], ",\"", [[], "ref"], "\":", "null", ",\"", [[], "topic"], "\":", [34, [], "t", 34], 125]
+  @v2_msg_json [91, "null", 44, "null", 44, [34, [], "t", 34], 44, [34, [], "e", 34], 44, [34, [], "m", 34], 93]
```
michalmuskala: Would it make sense to flatten the iodata before assertion in the test? It should be less brittle in case of changes in the library.

lib/phoenix/template.ex
```diff
@@ -104,7 +104,7 @@ defmodule Phoenix.Template do
 
   alias Phoenix.Template
 
-  @encoders [html: Phoenix.Template.HTML, json: Poison, js: Phoenix.Template.HTML]
+  @encoders [html: Phoenix.Template.HTML, json: Jason, js: Phoenix.Template.HTML]
```
josevalim: Also backwards incompatible.

lib/phoenix/transports/v2/longpoll_serializer.ex
josevalim: We should start using encode_to_iodata here.

lib/phoenix/transports/v2/longpoll_serializer.ex
```diff
@@ -23,7 +23,7 @@ defmodule Phoenix.Transports.V2.LongPollSerializer do
   Encoding is handled downstream in the LongPoll controller.
   """
   def encode!(msg) do
-    {:socket_push, :text, msg |> to_list() |> Poison.encode!()}
+    {:socket_push, :text, msg |> to_list() |> Jason.encode!()}
```
josevalim: Encode to iodata.

lib/phoenix/controller.ex
chrismccord: Yeah, I was going to ask how we should handle it. We can make it an explicit config, but then we'd still need to ship with the phoenix dep, which is not ideal

josevalim: We cannot simply change to Jason as that is backwards incompatible, since Jason is not fully compatible with Poison.

In my opinion, we should expose this function behind `Phoenix.json_encoder` (it can be @doc false) and use it throughout the codebase but we should still default to poison. In new apps, we generate:

    config :phoenix, :format_encoders, json: Jason

josevalim: Another reason to this clean up is that Ecto can only migrate to Jason by default in 3.0. So I am afraid that making this easily achievable is the best way to go for now.

chrismccord:
Recap:

- Require explicit configuration of json format_encoder for Phoenix to boot. (may also default to Poison)
- Use explicit config for new applications that configures `Jason` in user's mix config.
- Add `:jason` to new application's deps.
- Add `:jason` as optional dep in phoenix (tests)
- Drop Poison from Phoenix deps

chrismccord:
@josevalim updated as discussed. Can you take another look? Thanks!

lib/phoenix.ex
```diff
@@ -47,10 +47,33 @@ defmodule Phoenix do
     import Supervisor.Spec
 
     children = [
+      # Ensure format encoders are loaded before starting
+      worker(Task, [fn -> ensure_format_encoders() end], restart: :transient),
```
chrismccord: I'm not thrilled about this approach but it's the only way to get reasonable error messages out of `start`. The downside is the default max_restarts shows the runtime error three times. Better ideas?

michalmuskala: Would making it a `temporary` worker solve this?

josevalim: Starting a worker sounds like too much.

@chrismccord if you are not happy with the error message when raising directly inside application, you have two options:

  1. Log and return `{:error, :missing_format_encoder}`
  2. Log and let the app boot anyway

lib/phoenix/template.ex
josevalim: FWIW, this is still backwards incompatible. :)

test/phoenix/transports/websocket_serializer_test.exs
josevalim: :+1:

installer/templates/phx_single/config/config.exs
```diff
@@ -24,6 +24,9 @@ config :logger, :console,
   format: "$time $metadata[$level] $message\n",
   metadata: [:request_id]
 
+# Use Jason for JSON parsing in Phoenix
+config :phoenix, :format_encoders, json: Jason
```
josevalim: We also need to generate the configuration for Ecto:

    config :ecto, :json_library, Jason

chrismccord: Is this deprecated? Should we generate `config :#{db_adapter}, :json_library, Jason` instead?

chrismccord:
@josevalim can you take a final pass? I generated the "deprecated" for 3.0 Ecto config, but we can update to the ecto adapter config once 3.0 ships.

installer/templates/phx_umbrella/apps/app_name/config/config.exs
```diff
@@ -1,5 +1,6 @@
 use Mix.Config
 
 <%= if ecto do %>config :<%= app_name %>, ecto_repos: [<%= app_module %>.Repo]<% end %>
+<%= if ecto do %>config :ecto, :json_library, Jason<% end %>
```
josevalim: We can probably merge the two conditionals into one.

test/phoenix/transports/websocket_serializer_test.exs
josevalim: Ping!

test/test_helper.exs
```diff
@@ -3,6 +3,10 @@ Code.require_file("support/router_helper.exs", __DIR__)
 # Starts web server applications
 Application.ensure_all_started(:cowboy)
 
+# Clear format encoder cache from startup and inject Jason
+Application.delete_env(:phoenix, :compiled_format_encoders)
```
josevalim: Should we also cache `Phoenix.json`? Also, should we make it public?

Alternatively we can read `Phoenix.json` at compile time.

josevalim:
@chrismccord I have added some comments. Regarding Ecto 3.0, we have no estimate for its release, so I wouldn't worry about it.

Should we make `Phoenix.json` public? One advantage of doing so is that we don't need to hardcode Jason in the user application as well. For example, in Plug.Parsers we would pass `Phoenix.json`. One less place to change at the end of the day. Thoughts?

test/test_helper.exs
chrismccord: We can't read Phoenix.json at compile time because the application needs to be started to fetch application env, but isn't started yet ¯\_(ツ)_/¯

josevalim: No, we don't need to start the application to fetch the environment, only to fetch the defaults in mix.exs. But if the default is in the module, like in this case here, it should be dine.

Any configuration from the user application will also be available.

In any case, we can separate this discussion in two:

  1. Shall we make this public?
  2. Shall we cache it? At runtime or compile time?

chrismccord: Recap:

Introduce public `Phoenix.json_library/0` with `config :phoenix, :json_library` configuration. No need to cache as runtime lookup is simple ets call.

PR #2737, opened 2018-02-01
Add ability to route socket mounts by host

### LINKED ISSUE #2703: Feature request: Routing sockets based on host
In plug, we can route requests based on host

```
forward "/", host: "api.", to: MyApp.API.Router
```

We are using `phoenix` together with `absinthe` to build our GraphQL api:s. We host different schemas on different subdomains, and we are now looking into GraphQL subscriptions. But it seems sockets are only routable on path and not host.

chrismccord:
Closes #2703

@josevalim I'm not thrilled with the complexity of cowboy dispatches to support this, but let me know what you think!

### DIFF
--- a/lib/phoenix/endpoint/cowboy2_handler.ex
+++ b/lib/phoenix/endpoint/cowboy2_handler.ex
@@ -82,8 +82,13 @@ defmodule Phoenix.Endpoint.Cowboy2Handler do
       Application.ensure_all_started(:ssl)
     end
 
+    dispatches =
+      build_dispatches(endpoint, Plug.Adapters.Cowboy2.Handler, %{
+        Phoenix.Transports.LongPoll => Plug.Adapters.Cowboy2.Handler,
+        Phoenix.Transports.WebSocket => Phoenix.Endpoint.Cowboy2WebSocket
+      })
+
     # Use put_new to allow custom dispatches
-    dispatches = build_dispatches(endpoint, __MODULE__, Plug.Adapters.Cowboy2.Handler)
     config = Keyword.put_new(config, :dispatch, dispatches)
 
     cowboy_supervisor =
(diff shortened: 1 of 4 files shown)


lib/phoenix/endpoint/cowboy2_handler.ex
```diff
@@ -113,9 +102,43 @@ defmodule Phoenix.Endpoint.Cowboy2Handler do
...
-  defp default_for(Phoenix.Transports.LongPoll), do: Plug.Adapters.Cowboy2.Handler
-  defp default_for(Phoenix.Transports.WebSocket), do: Phoenix.Endpoint.Cowboy2WebSocket
-  defp default_for(_), do: nil
+  @doc false
+  def build_dispatches(endpoint, cowboy_handler, plug_handler) do
+    socket_dispatches =
+      for {path, socket, opts} <- endpoint.__sockets__,
+          {transport, {module, config}} <- socket.__transports__,
+          # Allow handlers to be configured at the transport level
+          handler = config[:cowboy] || default_for(cowboy_handler, module),
```
Gazler: Could we pass a map into this function instead to avoid referencing the `Phoenix.Endpoint.CowboyHander` in `default_for`?

something like:

```elixir
handler = config[:cowboy] || Map.get(default_handlers, module),
```


```elixir
  default_handlers = %{
    Phoenix.Transports.LongPoll => Plug.Adapters.Cowboy2.Handler,
   Phoenix.Transport.WebSocket => Phoenix.Endpoint.Cowboy2WebSocket
  }
dispatches = build_dispatches(endpoint, __MODULE__, Plug.Adapters.Cowboy2.Handler, default_handlers)
```

Then in the Cowboy1 handler we can have:

```elixir
  default_handlers = %{
    Phoenix.Transports.LongPoll => Plug.Adapters.Cowboy.Handler,
   Phoenix.Transport.WebSocket => Phoenix.Endpoint.CowboyWebSocket
  }
  Cowboy2Handler.build_dispatches(endpoint, __MODULE__, Plug.Adapters.Cowboy.Handler, default_handlers)
```

josevalim: Good call. Just remember to `@doc false` it. It would help us write some unit tests too!

PR #2738, opened 2018-02-01
Add an overview and diagram to the 'Moving Parts' section

nathanl:
I had a hard time following the "Moving Parts" section initially, and thought a diagram could clarify how things work.

I made this diagram using [Monodraw](https://monodraw.helftone.com/). The Monodraw file is much easier to edit and update than the exported text, but weighs a few KB. If you like this diagram, you can decide whether you also want the Monodraw file checked in.

### DIFF
--- a/guides/docs/channels.md
+++ b/guides/docs/channels.md
@@ -12,9 +12,42 @@ Phoenix ships with a JavaScript client that is available when generating a new P
 
 ## The Moving Parts
 
+### Overview
+
+To start communicating, a client connects to a socket using a transport (eg, Websockets or long polling) and joins one or more channels using that single network connection.
+One channel server process is created per client, per topic.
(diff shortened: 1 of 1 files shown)


nathanl:
This is purely a documentation change; the test failures are unrelated.

nathanl:
> I'm not wild about including a proprietary (mac only too!) format in the repository 

Yep, totally understand. I'll hang on to the file in case changes are needed, but worst case, it wouldn't be terribly hard to recreate this using another tool.

chrismccord:
Beautiful! Thanks you! This has definitely been an area of confusion for folks so having the new text an d diagram will definitely help. Since we're just using the ascii output from Monodraw, I'm fine adding it as is. If we need to tweak it in the future we can recreate with something else, but I don't think it will change too often. ❤️❤️❤️❤️❤️❤️🐥🔥

nathanl:
🎆 🎆

PR #2740, opened 2018-02-02
A working first pass at using npm scripts instead of brunch.

baldmountain:
Based on discussion on elixir-forum about brunch not being maintained I figured rather than griping about it I'd contribute some code. This PR adds an npm option to generate npm scripts to maintain javascript and css assets. This isn't meant to be the final solution. I'm not an expert at JavaScript/CSS build pipelines so I may have done a few things wrong. It's a starting point for conversation and contributions from folks who know more.

- All tests pass.
- The default is to use npm scripts rather than brunch.
- It generates a valid working project (at least with no-ecto) and includes watching app.js and app.css.
- Using npm should imply --no-brunch but isn't handled properly
- Probably some other issue too.

### DIFF
--- a/installer/README.md
+++ b/installer/README.md
@@ -4,4 +4,4 @@ Provides `phx.new` installer as an archive. To build and install it locally:
 
     $ cd installer
     $ MIX_ENV=prod mix archive.build
-    $ mix archive.install
+    $ MIX_ENV=prod mix mix archive.install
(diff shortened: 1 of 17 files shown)


chrismccord:
Thanks! I'm not ready at this point to make a change with our js tooling but we can use this as reference should we decide to use npm scripts in the future.

danschultzer:
The discussion can be found here: https://elixirforum.com/t/potentially-removing-brunch-from-the-phoenix-new-template-generator/2454

Thanks, @baldmountain! We got a production app where we badly need to get away from brunch, and I had plans to make a working example too. We'll try this out, and I'll report back asap 😄

The production app we got going has a complex but common setup of assets that should make it an ideal testing ground!

PR #2741, opened 2018-02-02
Fix issue when umbrella apps don’t use `web` in the name

### LINKED ISSUE #2656: `mix phx.new.web app_name` and mix `mix phx.gen.html` conflicts if app_name is not suffixed with `_web`
### Environment

* Elixir version (elixir -v): 1.5.2
* NodeJS version (node -v): 8.4.0
* NPM version (npm -v): 5.5.1
* Operating system: macOS Sierra 10.12.6

### Expected behavior
 `mix phx.new.web app_name` and mix `mix phx.gen.html` work fine together, no matter what the app_name is.

If for example the app name is `foo_web` and you run `mix phx.new.web foo_web` and `mix phx.gen.html Account User users name: string` everything works as expected.

### Actual behaviour
 `mix phx.new.web app_name` and mix `mix phx.gen.html` conflicts if app_name is not suffixed with `_web`

If for example the app name is `foo` and you run `mix phx.new.web foo` and `mix phx.gen.html Account User users name: string` the controller contains the namespace `FooWeb.UserController` and uses `use FooWeb, :controller` however `FooWeb` does not exist since the app name is `Foo`. The PageController (generated with the creation of the app) does this right however.

Same problem occurs in `FooWeb.UserView`

### Possible solutions
- Force the developer to suffix his app_name with `_web` when using `mix phx.new.web`, raise if not.
- Automatically suffix app_name with `_web` when creating new web app. (Or prompt developer for it)
- Remove suffixing with `_web` when building context struct: https://github.com/phoenixframework/phoenix/blob/d36148ae7b6570d45227abaabd2af9740beff542/lib/mix/phoenix/context.ex#L77-L84

alexgaribay:
When Phoenix is added to an umbrella app with a name that doesn't contain `web`, it maintains that name with the generators when using a separate context app.

Fixes #2656.

### DIFF
--- a/lib/mix/phoenix/context.ex
+++ b/lib/mix/phoenix/context.ex
@@ -76,10 +76,15 @@ defmodule Mix.Phoenix.Context do
 
   defp web_module do
     base = Mix.Phoenix.base()
-    if String.ends_with?(base, "Web") do
-      Module.concat([base])
-    else
-      Module.concat(["#{base}Web"])
+    cond do
+      Mix.Phoenix.context_app() != Mix.Phoenix.otp_app() ->
+        Module.concat([base])
+
+      String.ends_with?(base, "Web") ->
+        Module.concat([base])
+
+      true ->
+        Module.concat(["#{base}Web"])
     end
   end
 end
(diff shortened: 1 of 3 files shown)


chrismccord:
❤️❤️❤️🐥🔥

PR #2742, opened 2018-02-04
Add some type safety for presence state

edouardmenayde:
I spent yesterday afternoon debugging my code only to later found out I used an array as a presence state maintener instead of a plain old object. I think it's something that either should be checked at runtime (given JS has no type enforcement and the library is not using typescript...) or explicitely stated in the doc.
I went with the first for my pull request because I guessed why not ?


chrismccord:
Thanks! I can imagine it was a frustrating experience, but this one is tricky because of js types and check `!isArray` only accommodates this single case so I'm not convinced we should add it.

PR #2745, opened 2018-02-06
Remove support for Elixir 1.4

scrogson:
Based on changes in this commit: https://github.com/phoenixframework/phoenix/commit/49def42a9220198527807176f84ba92d131daa72

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,34 +1,6 @@
 language: elixir
 matrix:
   include:
-    - otp_release: 18.3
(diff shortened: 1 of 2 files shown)


chrismccord:
We need to decide how we still want to test elixir 1.4, given that existing apps will be supported on  Phoenix 1.4, but the generator for new apps will require 1.5+ // @josevalim

josevalim:
Working on it in #2748. :)

PR #2746, opened 2018-02-06
Guides: correct exunit async feature description

novaugust:
Per exunit docs, `async` doesn't parallelize individual tests: it only parallelizes the case modules themselves.

from exunit [main namespace docs](https://hexdocs.pm/ex_unit/ExUnit.html)
```
  # 3) Notice we pass "async: true", this runs the test case
  #    concurrently with other test cases. The individual tests
  #    within each test case are still run serially.
  use ExUnit.Case, async: true
```

and again, from the actual [Case docs](https://hexdocs.pm/ex_unit/ExUnit.Case.html#content)
>  :async - configure this specific test case to run in parallel with other test cases. May be used for performance when this test case does not change any global state. Defaults to false.

### DIFF
--- a/guides/docs/testing/testing.md
+++ b/guides/docs/testing/testing.md
@@ -96,7 +96,7 @@ defmodule HelloWeb.ErrorViewTest do
 end
 ```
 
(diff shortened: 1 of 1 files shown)


josevalim:

PR #2748, opened 2018-02-06
Fix the build matrix in regard to Elixir versions

josevalim:


josevalim:
The only failure we got was an intermittent one. We need to look into it.

josevalim:

Gazler:
Just for reference, I have used something like this to replicate the failure locally:

`while COWBOY_VERSION=1.0 mix test; do :; done`

PR #2751, opened 2018-02-06
Remove Plug.Static from endpoint.ex template when both --no-brunch and --no-html are passed

fertapric:
When both `--no-brunch` and `--no-html` options are passed, the [directory `priv/static` is not generated](https://github.com/phoenixframework/phoenix/blob/49def42a9220198527807176f84ba92d131daa72/installer/lib/phx_new/single.ex#L116). However, the template for `<app>_web/endpoint.ex` is configured to always include `plug Plug.Static`, which doesn't seem correct.


josevalim:
I am :-1:. Not having html/brunch does not imply there will be no other static asset or even that another javascript framework won't be used. If we remove Plug.Static, trying to add any other asset can be very confusing, so I would rather prefer it to be removed manually. Thank you.

PR #2752, opened 2018-02-08
Minor doc tweaks [ci skip]

fxn:
Minor documentation tweaks I did in passing while reading this file.

* "Pubsub" → "PubSub": It is the normal spelling, and the one used by the package itself.
* "pubsub" → "pub/sub": Also the conventional spelling in lowercase.
* Fixed-width font for `gettext` (as it is customary, and consistent with [their official docs](https://www.gnu.org/software/gettext/manual/gettext.html)).
* Links `gettext`.
* A missing article.


josevalim:

PR #2753, opened 2018-02-09
Use new plug init mode to reduce compile-time plug deps

### LINKED ISSUE #2622: Find a way to not add compile time dependencies to plugs during development
By calling `init` at runtime, instead of compile time.

chrismccord:
Closes #2622

@josevalim once you push plug 1.5.rc2 I'll update the plug dep and this will be ready to go. Thanks!


lib/phoenix/config.ex
```diff
@@ -9,6 +9,13 @@ defmodule Phoenix.Config do
   require Logger
   use GenServer
 
+  @doc """
+  Returns the `:init_mode` to pass to `Plug.Builder.compile/3`.
+  """
+  def plug_init_mode do
+    Application.get_env(:phoenix, :plug_init_mode, :compile)
+  end
```
josevalim: Nitpick: is this the best place to have this function? It feels very unrelated to this module. The only other option I can think is the `Phoenix` module but with `@doc false` in it.

josevalim:
1.5.0-rc.2 released.

lib/phoenix/config.ex
chrismccord: I went back and forth on this. My initial thought was a doc'd false function on `Phoenix`, but I didn't want that to start becoming a dumping ground anytime we need to look up a Phoenix config value, and `Phoenix.Config` literally is doc'd `Handles Phoenix configuration.`, so seem like a good place ¯\_(ツ)_/¯ 

I'll let  you make the call. LMK and I'll swap it :)

josevalim: `Phoenix.Config` is rather configuration for Phoenix projects/endpoints. You need to starts its GenServer and what not. We don't do it for Phoenix itself. :) So I would go with `Phoenix` unless we can think of something else. :)

PR #2754, opened 2018-02-10
Add support for context_app with schema generator

### LINKED ISSUE #2699: Support --context-app option for phx.gen.schema
To be consistent with _phx.gen.html_ options, please support _--context-app_ option as well as _context module_ parameter for _phx.gen.schema_ generator(and other applicable generators too?).

Thank you.

alexgaribay:
Adds context_app support for generating schema files based on `:context_app` in the config and `--context_app` in the CLI.

Fixes #2699

### DIFF
--- a/lib/mix/phoenix.ex
+++ b/lib/mix/phoenix.ex
@@ -276,6 +276,9 @@ defmodule Mix.Phoenix do
         via cli option:
 
             mix phx.gen.[task] --context-app some_app
+
+        Note: cli option only works when `context_app` is not set to `false`
+        in the config.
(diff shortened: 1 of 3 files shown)


chrismccord:
Looks great! Can you add a test case where context-app is passed as an argument to Gen.Schema.run? Will be good to go after that. Thanks!

chrismccord:
I cansee why you skipped the cli test just like I did for the other tasks as we check for the dependency to actually exist. Looks perfect. Thanks! ❤️❤️❤️🐥🔥

PR #2756, opened 2018-02-12
Support :infinity for Phoenix.Token max age

### LINKED ISSUE #2755: Allow silencing Phoenix.Token warning
### Environment

* Elixir version (elixir -v): 1.7.0-dev (c42df09f3)

### Expected behavior

We're stuck for some time having to support tokens issued with `Phoenix.Token` with an infinite expiration timeout. I understand the reasoning behind the warning in [`Phoenix.Token.verify/4`](https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/token.ex#L179) perfectly, it's unfortunately very annoying when I can't do much about it.

It would be great, if there was a way to silence the warning - I propose passing `max_age: :infinity` as an explicit way to opt-out from the token age checking. This should be the least disruptive and should make sure that the person using it understands the trade-offs.

### Actual behavior

We're spammed with the warnings, since they are issued at runtime.

michalmuskala:
Closes #2755 

If would be awesome if this could be backported to the 1.3 branch if there's a release planned.

### DIFF
--- a/lib/phoenix/token.ex
+++ b/lib/phoenix/token.ex
@@ -171,22 +171,11 @@ defmodule Phoenix.Token do
   def verify(context, salt, token, opts) when is_binary(salt) and is_binary(token) do
     secret = context |> get_key_base() |> get_secret(salt, opts)
 
-    max_age_ms =
-      if max_age_secs = opts[:max_age] do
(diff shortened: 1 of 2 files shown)


josevalim:

PR #2759, opened 2018-02-14
Move gettext under "Test dependencies" section in mix.exs

fertapric:
`gettext` is used `only: :test` so I thought it should be moved under the "Test dependencies" section.


chrismccord:
❤️❤️❤️🐥🔥

PR #2761, opened 2018-02-15
Support subdomains in the Router Helpers. Closes #2462

### LINKED ISSUE #2462: Support subdomains in the Router Helpers
### Environment

* Elixir version (elixir -v): Elixir 1.5.1
* Phoenix version (mix deps):  phoenix 1.2.3
### Expected behavior

Url generated on the subdomain should point to the subdomain

### Actual behavior
Url generated on the subdomain points to the host specified in the endpoint configuration

This further development of a host-based routing discussed in https://github.com/phoenixframework/phoenix/issues/489:
while  router itself supports subdomains, generating correct urls still requires manual work - user has to:
-  deduct current host via `conn.host`
- construct `%URI{}` structure with that host 
- pass URI to the `_url` helper

```
uri = %URI{host: conn.host}
MyAppWeb.Router.Helpers.page_url(uri, :show, "hello")
"https://other.example.com/pages/hello"
```

On the surface adding this piece of functionality is as simple as:

```
  def url(_router, %Conn{private: private, host: host}) do

    private.phoenix_endpoint.url
    |> URI.parse()
    |> Map.put(:host, host)
    |> URI.to_string()
  end
```

But it feels more like plumbing than a proper solution and I want to ask whether team has any strategy/vision/plans for the subdomains support in the Phoenix?

Thanks.

chrismccord:
@josevalim we have some overlap here and need to reconcile the `current_url` docs and behavior with this new feature. Let's chat tomorrow about it.

### DIFF
--- a/lib/phoenix/router/helpers.ex
+++ b/lib/phoenix/router/helpers.ex
@@ -9,14 +9,12 @@ defmodule Phoenix.Router.Helpers do
   @doc """
   Callback invoked by url generated in each helper module.
   """
-  def url(router, %Conn{private: %{phoenix_router_url: %URI{} = uri}}) do
-    url(router, uri)
-  end
-  def url(_router, %Conn{private: %{phoenix_router_url: "" <> url}}) do
-    url
-  end
-  def url(_router, %Conn{private: %{phoenix_endpoint: endpoint}}) do
-    endpoint.url()
+  def url(router, %Conn{private: private}) do
+    case private do
+      %{phoenix_router_url: %URI{} = uri} -> url(router, uri)
+      %{phoenix_router_url: url} when is_binary(url) -> url
+      %{phoenix_endpoint: endpoint} -> endpoint.url()
+    end
   end
 
   def url(_router, %Socket{endpoint: endpoint}) do
--- a/lib/phoenix/router/helpers.ex
+++ b/lib/phoenix/router/helpers.ex
@@ -9,12 +9,18 @@ defmodule Phoenix.Router.Helpers do
   @doc """
   Callback invoked by url generated in each helper module.
   """
-  def url(_router, %Conn{private: private}) do
-    private.phoenix_endpoint.url
+  def url(router, %Conn{private: %{phoenix_router_url: %URI{} = uri}}) do
+    url(router, uri)
+  end
+  def url(_router, %Conn{private: %{phoenix_router_url: "" <> url}}) do
+    url
+  end
+  def url(_router, %Conn{private: %{phoenix_endpoint: endpoint}}) do
+    endpoint.url()
   end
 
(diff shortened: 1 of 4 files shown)


lib/phoenix/router/helpers.ex
```diff
@@ -9,20 +9,26 @@ defmodule Phoenix.Router.Helpers do
   @doc """
   Callback invoked by url generated in each helper module.
   """
-  def url(_router, %Conn{private: private}) do
-    private.phoenix_endpoint.url
+  def url(router, %Conn{private: %{phoenix_router_url: %URI{} = uri}}) do
+    url(router, uri)
+  end
+  def url(_router, %Conn{private: %{phoenix_router_url: "" <> url}}) do
```
josevalim: This is equivalent to `is_binary(url)`, right?

lib/phoenix/router/helpers.ex
```diff
@@ -9,20 +9,26 @@ defmodule Phoenix.Router.Helpers do
   @doc """
   Callback invoked by url generated in each helper module.
   """
-  def url(_router, %Conn{private: private}) do
```
josevalim: Since this is a callback... maybe it should be `__url__`? It seems there is no reason to call it in public, is there?

josevalim: Nevermind. This whole module is `@moduledoc false`. :)

lib/phoenix/router/helpers.ex
```diff
@@ -9,20 +9,26 @@ defmodule Phoenix.Router.Helpers do
...
-    private.phoenix_endpoint.url
+  def url(router, %Conn{private: %{phoenix_router_url: %URI{} = uri}}) do
+    url(router, uri)
+  end
+  def url(_router, %Conn{private: %{phoenix_router_url: "" <> url}}) do
+    url
+  end
+  def url(_router, %Conn{private: %{phoenix_endpoint: endpoint}}) do
+    endpoint.url()
   end
```
josevalim: Let's get all of those three clauses and put them in a case. Something like:

```elixir
def url(_router, %Conn{private: private}) do
  case private do
    ...
```

The reason is that the formatter won't keep those clauses semantically grouped as we have right now and the current approach may end-up doing multiple accesses on private at runtime.

michalmuskala:
How does this feature interact with nested routers and `forward` macro?

josevalim:
I have added two minor comments. Feature-wise it looks perfect, I just pushed some docs. :)

josevalim:
It should just work because the forward stuff is ablut paths only.
-- 


lib/phoenix/router/helpers.ex
chrismccord: yeah. I'll change it to is_binary when I rewrite the defs to a cause hough

PR #2762, opened 2018-02-15
Authorization request header is now persisted during conn recycling

### LINKED ISSUE #2452: Request headers aren't persisted during controller tests
### Issue

Controller tests that use the `conn` object with request headers only send those headers for the first request. If the `conn` instance is used to make a second request then any request headers don't get sent to the app.

The Phoenix generators `phx.gen.json` and `phx.gen.html` create controller tests that follow this pattern of making two requests using the same `conn`. When I started to implement auth within my app I discovered that only the first request sends my `authorization` request header.

I did some digging and think that this is where all request headers (apart from `accepts`) are being dropped: https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/test/conn_test.ex#L469

### Example app

I've put a bare-bones example app together with a few controller tests replicating this issue: https://github.com/rosswilson/phoenix_req_header_issue

### Expected behaviour

A conn that has been configured with request headers should use those header values for any subsequent requests, not just the first.

I do wonder whether this is intentional, given the below comment. If that is the case, what is the recommended approach for including req headers for multiple requests within a single test? Should I be using `build_conn()` each time?

### Actual behaviour

Only the first request includes the configured headers.

### Environment

* Elixir version (elixir -v): Elixir 1.5.0
* NodeJS version (node -v): 8.2.1
* NPM version (npm -v): 5.3.0
* Operating system: Mac OS 10.12.6

rosswilson:
When `conn` gets recycled during tests, only the `accept` request header got persisted. This PR adds `authorization` to that list.

Fixes #2452.

### DIFF
--- a/lib/phoenix/test/conn_test.ex
+++ b/lib/phoenix/test/conn_test.ex
@@ -461,7 +461,7 @@ defmodule Phoenix.ConnTest do
     build_conn()
     |> Map.put(:host, conn.host)
     |> Plug.Test.recycle_cookies(conn)
-    |> copy_headers(conn.req_headers, ~w(accept))
+    |> copy_headers(conn.req_headers, ~w(accept authorization))
   end
 
   defp copy_headers(conn, headers, copy) do
--- a/test/phoenix/test/conn_test.exs
+++ b/test/phoenix/test/conn_test.exs
@@ -143,28 +143,44 @@ defmodule Phoenix.Test.ConnTest do
     refute conn.private.phoenix_recycled
   end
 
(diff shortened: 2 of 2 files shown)


chrismccord:
❤️❤️❤️🐥🔥

Cosaquee:
Any information when this will be released?

chrismccord:
When 1.4 is ready. Note you can easily duplicate the Phoenix.ConnTest function in your app and apply these changes until release :)

Cosaquee:
@chrismccord thank you!

PR #2765, opened 2018-02-16
Rework the introduction and organization of the Channels guide

nathanl:
Having gotten lost in some of these abstractions myself, I tried to make
this introduction as concrete and clear as possible. The examples and
new diagram at the top should help readers quickly grasp what channels
can be used for and the flexibility and scalability they offer.

Beyond that, I added a few clarifications and examples, and reordered
the parts so that PubSub is last, since it's the least visible to an
application developer.

### DIFF
--- a/guides/docs/channels.md
+++ b/guides/docs/channels.md
@@ -1,16 +1,44 @@
 # Channels
 
-Channels are a really exciting and powerful part of Phoenix that allow us to easily add soft-realtime features to our applications. Channels are based on a simple idea - sending and receiving messages. Senders broadcast messages about topics. Receivers subscribe to topics so that they can get those messages. Senders and receivers can switch roles on the same topic at any time.
+Channels are an exciting part of Phoenix that enable soft real-time communication with and between connected clients.
+Some possible use cases include:
 
-Since Elixir is based on message passing, you may wonder why we need this extra mechanism to send and receive messages. With Channels, neither senders nor receivers have to be Elixir processes. They can be anything that we can teach to communicate over a Channel - a JavaScript client, an iOS app, another Phoenix application, our watch. Also, messages broadcast over a Channel may have many receivers. Elixir processes communicate one to one.
+- Chat rooms
+- Breaking news, like "a goal was scored" or "an earthquake is coming"
+- Events in multiplayer games
+- Monitoring sensors and controlling lights
+- Notifying a browser that a page's CSS or JavaScript has changed (this is handy in development)
 
-The word "Channel" is really shorthand for a layered system with a number of components. Let's take a quick look at them now so we can see the big picture a little better.
+Conceptually, Channels are pretty simple.
+Clients connect and subscribe to one or more topics, whether that's `public_chat` or `updates:user1`.
+Any message sent on a topic, whether from the server or from a client, is sent to all clients subscribed to that topic, like this:
 
-## JS Documentation
+                                                        ┏━━━━━━━━━━━━━━━━┓
+                                                    ┏━━▶┃ Mobile Client  ┃
+                                                    ┃   ┗━━━━━━━━━━━━━━━━┛
+                             ┏━━━━━━━━━━━━━━━━━━━┓  ┃
+      ┏━━━━━━━━━━━━━━━━┓━━━━▶┃                   ┃  ┃   ┏━━━━━━━━━━━━━━━━┓
+      ┃ Browser Client ┃     ┃ Phoenix Server(s) ┃━━╋━━▶┃ Desktop Client ┃
+      ┗━━━━━━━━━━━━━━━━┛◀━━━━┃                   ┃  ┃   ┗━━━━━━━━━━━━━━━━┛
+                             ┗━━━━━━━━━━━━━━━━━━━┛  ┃
+                                                    ┃   ┏━━━━━━━━━━━━━━━━┓
+                                                    ┗━━▶┃   IoT Client   ┃
+                                                        ┗━━━━━━━━━━━━━━━━┛
 
-Phoenix ships with a JavaScript client that is available when generating a new Phoenix project. The documentation for the JavaScript module is available at [https://hexdocs.pm/phoenix/js/](https://hexdocs.pm/phoenix/js/)
+Channels can support any kind of client: a browser, native app, smart watch, embedded device, or anything else that can connect to a network.
+All the client needs is a suitable library; see the [Client Libraries](#client-libraries) section below.
+Each client library communicates using one of the "transports" that Channels understands.
+Currently, that's either Websockets or long polling, but other transports may be added in the future.
+
+Unlike stateless HTTP connections, Channels support long-lived connections, each backed by a lightweight BEAM process, working in parallel and maintaining its own state.
+
+This architecture scales well; Phoenix Channels [can support millions of subscribers with reasonable latency on a single box](http://phoenixframework.org/blog/the-road-to-2-million-websocket-connections), passing hundreds of thousands of messages per second.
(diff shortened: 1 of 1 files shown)


guides/docs/channels.md
```diff
@@ -47,14 +77,21 @@ The message flow looks something like this:
...
+
+Within the handler, you can authenticate and identify a socket connection and set default socket assigns.
 
 ### Channel Routes
 
-These are defined in Socket handlers, such as `lib/hello_web/channels/user_socket.ex`, which makes them distinct from other routes. They match on the topic string and dispatch matching requests to the given Channel module. The star character `*` acts as a wildcard matcher, so in the following example route, requests for `sample_topic:pizza` and `sample_topic:oranges` would both be dispatched to the `SampleTopicChannel`.
+Channel routes are defined in socket handlers, such as `lib/hello_web/channels/user_socket.ex` in the example app below.
+They match on the topic string and dispatch matching requests to the given Channel module.
+
+The star character `*` acts as a wildcard matcher, so in the following example route, requests for `elixir:ecto` and `elixir:phoenix` would both be dispatched to the `ElixirTopicChannel`.
```
Gazler: It may be worth changing the example here, as readers may be confused by these topics due to their usage elsewhere in the framework.

Perhaps we could go with `room:lobby`, `room:123` and `RoomTopicChannel instead?

Gazler:
This is great! Thanks @nathanl. Looks like you are really putting Monodraw to good use!

fertapric:
I'm also a Monodraw user and I've been having some issues rendering non-ASCII diagrams in code blocks of the documentation. Here is a screenshot of the generated docs:

screen shot 2018-02-20 at 10 54 46 am

nathanl:
@fertapric Yikes! Thanks for noticing that. I'll try to fix it soon.

nathanl:
@fertapric I switched these diagrams to be ASCII-only, and they render much better. (If anyone else wants to try it, run `mix docs`).

I also rearranged the larger one so that it doesn't require horizontal scrolling.

Here's how they look for me (I checked on MacOS in Chrome, Firefox and Safari):

first

second

guides/docs/channels.md
```diff
@@ -1,60 +1,127 @@
...
+- Events in multiplayer games
+- Monitoring sensors and controlling lights
+- Notifying a browser that a page's CSS or JavaScript has changed (this is handy in development)
+
+Conceptually, Channels are pretty simple.
+Clients connect and subscribe to one or more topics, whether that's `public_chat` or `updates:user1`.
+Any message sent on a topic, whether from the server or from a client, is sent to all clients subscribed to that topic, like this:
+
+<pre>
+<code>
```
nathanl: I put in these `<pre><code>` tags manually because if I just indented, the resulting `<code>` tags had `highlightjs` and tried to highlight the diagrams as Elixir, with inconsistent coloring.

guides/docs/channels.md
```diff
@@ -1,60 +1,127 @@
...
+
+- Chat rooms
+- Breaking news, like "a goal was scored" or "an earthquake is coming"
+- Events in multiplayer games
+- Monitoring sensors and controlling lights
+- Notifying a browser that a page's CSS or JavaScript has changed (this is handy in development)
+
+Conceptually, Channels are pretty simple.
+Clients connect and subscribe to one or more topics, whether that's `public_chat` or `updates:user1`.
+Any message sent on a topic, whether from the server or from a client, is sent to all clients subscribed to that topic, like this:
```
fertapric: I think it would be nice to explain what happens when a client is subscribed to a topic and sends a message to that same topic.

fertapric:
@nathanl they look good! 👏 

However, I'm not sure if the first diagram is really needed, because it's a visual support for a paragraph that, in my opinion, is not hard to understand:
> Clients connect and subscribe to one or more topics, whether that's `public_chat` or `updates:user1`.
> Any message sent on a topic, whether from the server or from a client, is sent to all clients subscribed to that topic, like this:

Regarding the second diagram, I would like to suggest a different diagram:

```
  Sending client
        │
        │
    Transport
        │
        │
  Channel route
        │
        │
  Channel.Server
  Sending Client
     Topic 1
        │
        │
   Local PubSub
        │
        ├── Channel.Server
        │   Browser Client
        │       Topic 1
        │          │  
        │          └── Transport
        │                  │
        │                  └── Browser Client
        │
        ├── Remote PubSub
        │         │
        │         └── Channel.Server
        │               IoT Client
        │                Topic 1
        │                   │
        │                   └── Transport
        │                           │
        │                           └── IoT Client
        └── Remote PubSub
                 │
                 ├── Channel.Server
                 │     Phone Client
                 │       Topic 1
                 │          │
                 │          └── Transport
                 │                  │
                 │                  └── Phone Client
                 │
                 └── Channel.Server
                      Watch Client
                        Topic 1
                           │
                           └── Transport
                                   │
                                   └── Watch Client
```

In my opinion, the main advantages of the new diagram are:

  * it reads from top to bottom and from left to right, which I think it's easier to follow
  * It takes less horizontal space, which should help when reading the docs using mobile devices
  * It does not require Monogram to update it 

Said that, I'm ok if you decide to go with the current diagrams :)

nathanl:
> However, I'm not sure if the first diagram is really needed, because it's a visual support for a paragraph that, in my opinion, is not hard to understand

It's true; it's not necessary. I think it makes it more obvious what's happening, but don't feel strongly.

> I would like to suggest a different diagram...

That works, too. I personally prefer the one I drew because:

- It's clearer that messages travel *over* a transport, not *to* one
- I like seeing that the sending client has a "Topic 2" server that is bypassed by the route
- It's prettier IMO. 😄

I think that updating either diagram in a text editor will be somewhat annoying but doable.

However, I'm happy either way. @Gazler what do you think?

Gazler:
We can keep the first diagram, I think it adds value.

For the second diagram, I prefer the original one and find it easier to understand. I think we may just have to live with the fact they are tricky to read on mobile.

fertapric:
It sounds good to me 👌

Gazler:
@nathanl I think we still want to change https://github.com/phoenixframework/phoenix/pull/2765#discussion_r169262771 - other than that, I think this is good to go.

guides/docs/channels.md
nathanl: @fertapric Good point. Explained in latest commit.

guides/docs/channels.md
nathanl: Yep - fixed in latest commit.

nathanl:
@Gazler Updated!

guides/docs/channels.md
```diff
@@ -1,60 +1,127 @@
...
+                              +-------------------+  |
+                                                     |            +----------------+
+                                                     +--Topic X-->|   IoT Client   |
+                                                                  +----------------+
+</code>
+</pre>
+
+Channels can support any kind of client: a browser, native app, smart watch, embedded device, or anything else that can connect to a network.
+All the client needs is a suitable library; see the [Client Libraries](#client-libraries) section below.
+Each client library communicates using one of the "transports" that Channels understands.
```
fertapric: Maybe `Channels understand`? 

In the document `Channels` is treated as a plural concept (see a few lines below "Channels support..."), but in this line is treated as a singular name.

fertapric:
@Gazler Nitpick: the commit message says differently `Channels understands -> Channel understands`, but the change is `Channels understands -> Channels understand` :)

Gazler:
Worst day ever!

Gazler:
Thanks @nathanl this is fantastic! <3 <3 

Thanks for reviewing @fertapric

PR #2766, opened 2018-02-17
Resolve presence race and introduce new Presence API

### LINKED ISSUE #2255: Race condition when presence is left shortly after presence state retrieval
### Environment

* Elixir version (elixir -v): Elixir 1.4.0, erts-8.2)
* Phoenix version (mix deps): phoenix 1.3.0-rc.1
* NodeJS version (node -v): v7.7.2
* NPM version (npm -v): 4.4.1
* Operating system: OS X 10.12.3

### Expected behavior

Using the default setup for channels and presence, any leaves happening concurrently with a new channel's connection should result in a client side presence state without the left presence (either through it not being included in `presence_state` or it being removed shortly thereafter through a `presence_diff`.

### Actual behavior

Some presence changes, that are applied after the newly joined channel has retrieved its current presence state, will be sent to the client before the presence_state event is sent. When this happens, any such presence leaves will be ignored by the default js syncDiff. Once the presence_state event is received, the left presences will be kept even though they are not present anymore.

### Easiest way to reproduce locally

1. Introduce a delay in a `Presence.fetch/2` callback, that scales linearly with the number of presences. Example: `def fetch(_topic, presences) do Process.sleep(1000 * Enum.count presences); presences end`
2. Open a few tabs to make sure the presence state contains 5 or more items.
3. Reload a tab, triggering a new channel connection.
4. Within a few seconds, close one or more of the other tabs.
5. Observe that the newly opened tab recevies the leave events before it's presence_state, and that it keeps those left presences until the next reload.

A runnable example of the above is available at https://github.com/alvinlindstam/phoenix_social_presence/tree/race-condition

I understand that the one second delay is an extreme example, but a fetch function that takes longer for many presences than for a small diff is hardly unlikely. I'm not sure if this would be possible to reproduce without a fetch function, but I can't se any reason why a presence change handled just after a presence state fetch by a new channel couldn't be sent to and handled by the transport process before the channel process itself sends the state to the transport.

chrismccord:
Fixes #2255, for real this time :)

@josevalim and @alvinlindstam I would appreciate thoughts on this new API. It keeps the "lower level" existing API and simplifies the knowledge and code required to handle presence events. At the same time it also resolves the race condition in #2255 by tracking the joinRef or "session id" of the channel connection. The new API looks like this (taken from Alvin's example app):

```javascript
const renderPresences = (presence) => {
  const listed_presences = presence.list((user, {metas: metas}) => {
    return {user: user, ...}
  })
  ...
}

let channel = socket.channel("user:all", {})
let presence = new Presence(channel)

presence.onSync(() => {
  renderPresences(presence)
})

channel.join()...
```

Note: I still need to update the presence.js docs with the new API, but I'm awaiting feedback. Thanks!

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -109,21 +109,29 @@
  * The `Presence` object provides features for syncing presence information
  * from the server with the client and handling presences joining and leaving.
  *
- * ### Syncing initial state from the server
+ * ### Syncing state from the server
  *
- * `Presence.syncState` is used to sync the list of presences on the server
- * with the client's state. An optional `onJoin` and `onLeave` callback can
- * be provided to react to changes in the client's local presences across
- * disconnects and reconnects with the server.
+ * To sync presence state from the channel, first instantiate an object,
+ * passing your channel in to track lifecycle events:
  *
- * `Presence.syncDiff` is used to sync a diff of presence join and leave
- * events from the server, as they happen. Like `syncState`, `syncDiff`
- * accepts optional `onJoin` and `onLeave` callbacks to react to a user
- * joining or leaving from a device.
+ * ```javascript
+ * let channel = new socket.channel("some:topic")
+ * let presence = new Presence(channel)
+ * ```
+ *
+ * Next, use the `presence.onSync` callback to react to state changes
+ * from the server. For example, to render the list of users every time
+ * the list changes, you could write:
+ *
+ * ```javascript
+ * presence.onSync(() => {
+ *   myRenderUsersFunction(presence.list())
+ * })
+ * ```
  *
  * ### Listing Presences
  *
- * `Presence.list` is used to return a list of presence information
+ * `presence.list` is used to return a list of presence information
  * based on the local state of metadata. By default, all presence
  * metadata is returned, but a `listBy` function can be supplied to
  * allow the client to select which metadata to use for a given presence.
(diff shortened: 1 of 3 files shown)


OvermindDL1:
What about when you have different presences coming down a single channel (as I do)?

chrismccord:
Not sure I follow? How are you handling it today that this api would not allow? (Keep in mind the "old" api is still there, but you would need to resolve the race yourself)

OvermindDL1:
The standard naming conventions that presence uses by default I don't at all, thus if I wanted to use the new API defined here I'd also imagine being able to pass in different names other than `presence_state` and `presence_diff` as, for example, I instead use `notifications_state`/`notifications_diff` and `members_state`/`members_diff`.  I.E. having the ability to pass in different message names than the ones it is currently hard-coded to would let me use this new API (and remove all my manual synching code, whoo).  :-)

Basically I want to be able to do this:

```javascript
let notifications_presence = new Presence(channel, "notifications_state", "notifications_diff")
```

Or maybe just this, which would just alias to the above (for the simple case where you follow the `<name>_state`/`<name>_diff` pattern as I do):

```javascript
let notifications_presence = new Presence(channel, "notifications")
```

chrismccord:
Oh I see. We can easily support those as options. Thanks!

jeregrine:
👍 I don't love that it uses internal state to keep track of it. But this is JavaScript, and its probably best to keep with how they handle it.

josevalim:
@chrismccord do we have docs for update? Also, do we generate some javascript in the generators or do we show some javascript in them? If so they need to be updated too.

chrismccord:
@josevalim I was waiting on feedback for the API before finalizing docs. If all looks good from your side I'll touch up the docs

josevalim:
Good for me!

OvermindDL1:
@chrismccord And just a reminder ping on the message naming options.  ^.^

christianevans214:
Hey there! Sorry to butt in on this PR, but is there an ETA on when this new presence API will be out via NPM?

PR #2771, opened 2018-02-21
Allow Websocket Transport to copy conn fields to params

mobileoverlord:
When authenticating websockets, sometimes its necessary to have access to the headers from the initial HTTP request prior to the connection being upgraded. 

The PR would allow the Websocket transport to take a list of conn fields which it would copy into the params before the websocket upgrade.

For Example, 

```elixir
## Transports
  transport :websocket, Phoenix.Transports.WebSocket, conn_fields: [:req_headers]

def connect(params, socket) do
    %{req_headers: req_headers} = params
    {:ok, socket}
  end
```

### DIFF
--- a/lib/phoenix/transports/websocket.ex
+++ b/lib/phoenix/transports/websocket.ex
@@ -34,6 +34,14 @@ defmodule Phoenix.Transports.WebSocket do
     * `:code_reloader` - optionally override the default `:code_reloader` value
       from the socket's endpoint
 
+    * `:conn_fields` - optionally copy extra fields from the conn to the params
+      for use in connecting the socket.
+
+      For example, if you need access to the request headers, you can set
+      
+      conn_fields: [:req_header]
+
+
   ## Serializer
 
   By default, JSON encoding is used to broker messages to and from clients.
@@ -82,6 +90,7 @@ defmodule Phoenix.Transports.WebSocket do
       conn
       |> code_reload(opts, endpoint)
       |> fetch_query_params()
+      |> copy_conn_fields(opts)
       |> Transport.transport_log(opts[:transport_log])
       |> Transport.force_ssl(handler, endpoint, opts)
       |> Transport.check_origin(handler, endpoint, opts)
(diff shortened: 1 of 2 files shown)


josevalim:
Which fields would you need? It is unlikely we will accept such a general construct because there are security implications and it makes channels less generic.

mobileoverlord:
The primary reason would be the request headers. Our example case is when SSL is being terminated upstream and the headers are being rewritten with authentication information.

josevalim:
Sorry, I meant to ask: which headers do you need?

mobileoverlord:
What we are doing is TLS client auth on an edge device which then forwards fields from the certificate into the request headers. This is done by some providers, in our case we are using the cloudflare. 

https://blog.cloudflare.com/introducing-tls-client-auth/

The certificate is going to contain the serial number of a Nerves device after successful authentication which we later need for identifying the device on the other end of the socket. Since the forwarding configuration is handled by cloudflare, we felt the implementation needed flexibility to allow any req header to be fetched. Giving access to everything in the conn may be overkill, but I fail to see the security implications considering you have full access to the conn everywhere else in phoenix.

I do agree that there would need to be an example showing that if a socket implements more then one transport that the extra socket params would only be available on connections from the websocket transport. I don't think that this makes things any less generic where as this could be implemented as an added feature to any conn backed transport connection without altering the overall `Transport` API.

josevalim:
We need a way to make this work then but without copying whole connection fields. My suggestion is to allow *headers* to be copied and explicitly disallow the "Cookie" and "Authentication" header to be used in any shape.

@chrismccord where would those headers be stored? We can pass them as a third argument to connect/3 or store them in the socket only during the connection (they are removed when sent to the channel).

michalmuskala:
I think another common requested thing is the `remote_ip` field and I think it is very much related to that. If anything, it's useful for logging, but there might be other uses.

chrismccord:
I would prefer to see a metadata key on the socket or as a third argument to connect/3 that the transports can provide, rather than calling it headers or similar. That way we don't get so coupled directly to the transport and users could match as needed on it. Thoughts?

rmoorman:
@chrismccord when using a third argument to connect (`connect/3`), would you suggest to give access to the conn struct or still just a limited set of data?
Also, `connect/3` would be in charge of making some sense out of the headers (or another field of the conn) using this kind of flow, which feels a bit after the fact for me maybe. I think that a plug could be more appropriate to take some decision based on the conn, before the socket/transport code is even touched. Having a plug handle this and having a way to (re) use some assigns from the conn could also be a solution. But then the endpoint logic would have to be restructured a little too.

p.s. It would be really great to be able to get some additional request (-based) data into my sockets/channels.

bobics:
:+1: This would also be useful to us -- our use case is IP based rate limiting on join / messages (unfortunately, passing in the IP in as a message param isn't really an option since that can easily be modified).

mobileoverlord:
This is being pushed forward in https://github.com/phoenixframework/phoenix/pull/2889

PR #2779, opened 2018-03-02
Replace brunch with webpack 4 in the installers

Gazler:
Since brunch is no longer maintained, webpack has been chosen as a replacement.
webpack 4 reduces a lot of the config required, the included configuration is
designed so that the compiled assets act the same way as they did with bruch (js
goes in `priv/static/js/app.js` and css in `priv/static/css/app.js`. Any assets
in `assets/static` are copied over to `priv/static`.

All references to brunch have been replaced by webpack. The `--no-bruch` flag is
now `--no-webpack`.

Stylistically, webpack is written with a lower case "w" - so this has
been used in any comments/documentation.


Gazler:
This might need testing on Windows.

chrismccord:
❤️❤️❤️🐥🔥

webdeb:
Wow!

drozdzynski:
@Gazler From where you got information about no longer maintained Brunch?

Gazler:
@drozdzynski That was poor wording on my part, I had meant to go back and change it before this PR was merged. I have updated my original comment to help clarify this.

It's not brunch specifically (although there hasn't been a huge amount of activity on the brunch main repo either recently), but the libraries and eco-system around it that seem to be lacking in maintenance updated.

skovmand:
Have you considered this statement in the [extract-text-webpack-plugin](https://github.com/webpack-contrib/extract-text-webpack-plugin) repo? 
"Since webpack v4 the [extract-text-webpack-plugin](https://github.com/webpack-contrib/extract-text-webpack-plugin) should not be used for css. Use mini-css-extract-plugin instead."

Gazler:
@skovmand Yeah, we changed it in this PR https://github.com/phoenixframework/phoenix/pull/2831

PR #2782, opened 2018-03-03
Validate params

arcz:
I doubt this will be accepted but I would like to at least touch this topic and discuss it with the community. This code sets Ecto as a dependency which I suspect is unfortunately a no-go for this PR. It defines new `Phoenix.Controller.validate_params/3` function which purpose is to transform string params into valid arguments accepted by domain functions.

There are a lot of useful functions for presenting data to user be it json or html but there is nothing more than pattern matching on actions for controlling params coming to action.
I'm talking about missing * in this data flow:

```
request -> controller -> * -> call domain -> present response
```

It is hard to call domain functions with the current featureset or I didn't discover it yet.
Lets say there a is a function which is to be called from controller:

```elixir
defmodule Report do
  def build(dates) do
    ...
  end
end
```

As written in Programming Erlang, I'm following the principle that functions shouldn't fix its arguments and rather assume that it is a caller responsibility to pass them correctly. When I'm in control of the arguments and call it from iex or automated job which sends it over the email I would call it like that
```elixir
iex> Report.build([~D[2018-03-01], ~D[2018-03-02]])
```

When calling on behalf of a user in controller action, it is a hard job to do it right. With the `validate_params` I can do:

```elixir
def index(conn, params) do
  schema = %{dates: {{:array, :date}, :required}}

  case validate_params(schema, params) do
    {:ok, %{dates: dates}} ->
      render_200 conn, Report.build(dates)
    {:error, changeset} ->
      render_422 conn, changeset
  end
end
```

What I like about this solution is that we are explicit about the input to the controller instead or map of strings.
Actually problem is still there even in basic CRUD application.

```elixir
def show(conn, %{"id" => id}) do
  case Repo.get(Foo, id) do
    nil -> send_resp(conn, 404, "not found")
    foo -> json(conn, %{id: foo.id})
  end
end
```
`Repo.get` will accept here string which is fine as primary key can be actually a string but in this context is is integer. If passed `1` or `"1"` it works fine, but one can easily pass `"foo"` and get an exception. Actually now I've checked and looks like Phoenix catches `Ecto.Query.CastError` exception and returns 400. But still, catching an exception here for an argument which can be validated upfront is a bit sneaky. It is good that Phoenix handles Ecto case but domain can call different database library or service.

I use this approach in production for a few projects and it has been solid. I hope that I didn't miss something, please let me know if I'm doing this wrong.

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -1399,6 +1399,72 @@ defmodule Phoenix.Controller do
     Phoenix.Router.Helpers.url(router_module(conn), conn) <> current_path(conn, params)
   end
 
+  @doc """
+  Validates params according to the given `schema`.
+  Schema is a map where key is an atom and corresponds
+  to outside param key. Value is one of the three possibilities:
+
+  1. `type`
+  2. `{type, :required}`
+  3. `{type, default_value}`
+
+  Where type is a castable Ecto type.
+  When `params` are missing a key which was marked `:required` in `schema`,
+  result will be an error tuple.
+  A `changeset_fun` can be passed as optional last argument,
+  which will be applied to the resulting changeset.
+
+  ## Examples
+      iex> schema = %{id: {:integer, :required}}
+      iex> validate_params(schema, %{"id" => "1"})
+      {:ok, %{id: 1}}
+
+      iex> schema = %{year: {:integer, Date.utc_today.year}}
+      iex> validate_params(schema, %{})
+      {:ok, %{year: 2018}}
+
+      iex> schema = %{quarter: {:integer, :required}}
+      iex> transformer = &Ecto.Changeset.validate_inclusion(&1, :quarter, 1..4)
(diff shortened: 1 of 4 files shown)


LostKobrakai:
Using schemaless changesets or embedded schemas one can easily have a changeset to transform/validate incoming string input into their need domain types. To me your proposed additional function only adds a new abstractions above already known ways to configure a changeset without providing much benefit beyond that from what I can tell. So why would I use your proposed function instead of a simple helper function, where I define a schemaless changeset like it's done for all other changesets in a project?

chrismccord:
In the future please follow the issue template guidelines and use the mailing list to discuss feature requests.  You may also consider wrapping up these ideas in a library and seeking feedback from the community on the forums and elsewhere. Thanks!

arcz:
@LostKobrakai the point here was to have declarative way on hand and avoid some boilerplate, but now I see it is not really universal solution and should have put it on mailing list/forum instead as Chris said. Thanks for feedback :)

PR #2786, opened 2018-03-06
Add .DS_Store to gitignore

inoas:


michalmuskala:
Since this is something OS-specific, I think it should be generally handled by the global `.gitignore` file on the machine and not per project.

josevalim:
Yup, OS specific stuff goes to your global gitignore.

PR #2787, opened 2018-03-07
Add additional information on TLS config and HSTS to the Endpoint guilde.

idyll:
This PR adds additional information on configuring HTTPS including a link to the Erlang options that can be set for TLS.

It also provides details around HSTS, the implications of turning it on, and how to turn it off for localhost.

### DIFF
--- a/guides/docs/endpoint.md
+++ b/guides/docs/endpoint.md
@@ -134,7 +134,7 @@ Without the `otp_app:` key, we need to provide absolute paths to the files where
 Path.expand("../../../some/path/to/ssl/key.pem", __DIR__)
 ```
 
-If you require further customization to the TLS versions or cyphers used you can include additional `https:` configuration. For example to disable older versions of TLS which are now considered insecure you could add `versions: [:'tlsv1.2']`. More information on the available settings is available in the (Erlang SSL docs)[http://erlang.org/doc/man/ssl.html}] (see "SSL OPTION DESCRIPTIONS - SERVER SIDE").
+If you require further customization to the TLS versions or ciphers used you can include additional `https:` configuration. For example to disable older versions of TLS which are now considered insecure you could add `versions: [:'tlsv1.2']`. More information on the available settings is available in the (Erlang SSL docs)[http://erlang.org/doc/man/ssl.html}] (see "SSL OPTION DESCRIPTIONS - SERVER SIDE").
 
 ### Force SSL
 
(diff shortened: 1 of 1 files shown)


guides/docs/endpoint.md
```diff
@@ -150,6 +152,16 @@ config :my_app, MyApp.Endpoint,
   force_ssl: [rewrite_on: [:x_forwarded_proto], host: nil]
 ```
 
+### HSTS
+
+HSTS or "strict-transport-security" is a mechanism that allows a website to declare itself as only accessable via a secure connection (HTTPS). It was introduced as a mechanism to in
```
PhillippOhlandt: > It was introduced as a mechanism to in

Sentence suddenly ends.

idyll: Apparently I need more sleep. Good catch. :)

guides/docs/endpoint.md
```diff
@@ -134,6 +134,8 @@ Without the `otp_app:` key, we need to provide absolute paths to the files where
 Path.expand("../../../some/path/to/ssl/key.pem", __DIR__)
 ```
 
+If you require further customization to the TLS versions or cyphers used you can include additional `https:` configuration. For example to disable older versions of TLS which are now considered insecure you could add `versions: [:'tlsv1.2']`. More information on the available settings is available in the (Erlang SSL docs)[http://erlang.org/doc/man/ssl.html}] (see "SSL OPTION DESCRIPTIONS - SERVER SIDE").
```
Gazler: I think it is `cipher` and not `cypher` in the context of tls.

Gazler:
@idyll One small change then this is good to merge.

guides/docs/endpoint.md
```diff
@@ -150,6 +152,16 @@ config :my_app, MyApp.Endpoint,
...
+
+This can result in unexpected behaviour if you access your application on localhost, for example `https://localhost:4000`, as from that point forward and traffic coming from localhost will be expected to be encrypted, except port 80 which will be redirected to port 443. This has the potential to disrupt traffic to any other local servers or proxies that you may be running on your computer. Other applications or proxies on localhost will refuse to work unless the traffic is encrypted.
+
+If you do invertenlty turn on HSTS for localhost you may need to rest the cache on your browser before it will accept any HTTP traffic from localhost. For Chrome you need to `Empty Cache and Hard Reload` which is available from the reload menu that appears when you click and hold the reload icon from the Developer Tools Panel. For Safari you will need to clear your cache, remove the entry from `~/Library/Cookies/HSTS.plist` (or delete that file entirely) and restart Safari. Alternately you can set the `:expires` option on `force_ssl` to `0` which should expired the entry to turn off HSTS. More information on the options for HSTS are available at [Plug.SSL](https://hexdocs.pm/plug/Plug.SSL.html).
```
markmark206: “inadvertently”? “res**e**t cache”?

guides/docs/endpoint.md
```diff
@@ -150,6 +152,16 @@ config :my_app, MyApp.Endpoint,
   force_ssl: [rewrite_on: [:x_forwarded_proto], host: nil]
 ```
 
+### HSTS
+
+HSTS or "strict-transport-security" is a mechanism that allows a website to declare itself as only accessable via a secure connection (HTTPS). It was introduced to prevent man-in-the-middle attacks that strip SSL/TLS. It causes web browers to redirect from HTTP to HTTPS and refuse to connect unless the connection uses SSL/TLS.
```
markmark206: “accessable” -> “accessible”

idyll:
VSCode needs grammar check apparently. :) 

Thanks @markmark206 and @Gazler

Gazler:
Thanks <3

PR #2796, opened 2018-03-09
Update README's travis badge to point to master

sabondano:
Not 100% if this tiny edit makes sense but I noticed that the badge in the README was "failing" while the master branch on Travis was green. This change makes the travis badge in the README point directly to the master branch.


josevalim:

PR #2798, opened 2018-03-09
Add a .formatter.exs to Phoenix itself so we can use import_deps (#2605)

michallepicki:
as discussed in #2605


.formatter.exs
```diff
@@ -0,0 +1,69 @@
...
+      head: 3,
+      head: 4,
+      match: 4,
+      match: 5,
+      options: 3,
+      options: 4,
+      patch: 3,
+      patch: 4,
+      pipe_through: 1,
+      plug: 1,
```
fertapric: Some of these functions belong to Plug, and they are currently exported by its [`.formatter.exs`](https://github.com/elixir-plug/plug/blob/master/.formatter.exs#L3-L21). So, I'm not sure if they should be exported here.

Since `:import_deps` cannot be used inside`:export`,  I would say the final users should be responsible of importing Plug and Phoenix formatters `import_deps: [:phoenix, :plug]`.

fertapric: The `import_deps` configuration could be included in the templates of the Phoenix installer :)

.formatter.exs
```diff
@@ -0,0 +1,69 @@
+[
+  export: [
+    locals_without_parens: [
```
josevalim: Can you put this in a variable and then reuse it? Something like this:

```elixir
locals_without_parens = [...]

[
  locals_without_parens: locals_without_parens,
  export: [locals_without_parens: locals_without_parens]
]
```

.formatter.exs
```diff
@@ -0,0 +1,69 @@
...
+      action_fallback: 1,
+
+      # Phoenix.Endpoint
+      socket: 2,
+
+      # Phoenix.Socket
+      channel: 2,
+      channel: 3,
+      transport: 2,
+      transport: 3,
```
josevalim: The transport macro is going away on v1.4 so we can already remove them. :)

josevalim:
Oh wait, why did you remove the router ones? They are not from plug! Phoenix router has no relation with plug router. They are two separate entities. :)

fertapric:
Sorry 😞 

I thought those ones were coming from Plug and I led @michallepicki in the wrong direction.

josevalim:
No problem! :D thank you for your comments!

PR #2801, opened 2018-03-09
Generate a static_push function in the RouteHelpers

Gazler:
The `static_push/2` function is a drop in replacement for `static_path/2` with
the addition of performing a server push if the adapter allows. This function
only supports using a `Plug.Conn` struct - attempting to use it with an endpoint
will result in an error.

### DIFF
--- a/lib/phoenix/router/helpers.ex
+++ b/lib/phoenix/router/helpers.ex
@@ -153,7 +153,7 @@ defmodule Phoenix.Router.Helpers do
       end
 
       @doc """
-      Perform a server push and return the path to a static asset given its file path.
+      Performs a server push and returns the path to a static asset given its file path.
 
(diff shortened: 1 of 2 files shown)


Gazler:
This depends on https://github.com/elixir-plug/plug/pull/678#pullrequestreview-102651352 - so we should wait for a new version of plug.

lib/phoenix/router/helpers.ex
```diff
@@ -152,6 +152,18 @@ defmodule Phoenix.Router.Helpers do
         endpoint.static_path(path)
       end
 
+      @doc """
+      Generates path to a static asset given its file path.
```
josevalim: First paragraph of docs should always be a short summary. :)

PR #2802, opened 2018-03-09
Expand examples slightly

nathanl:
A tiny follow-on to https://github.com/phoenixframework/phoenix/pull/2765

I wanted to mention "messaging apps", thinking of something like WhatsApp. And I thought of the map example, too.

The goal of this section is to get people excited, so a variety of examples is helpful, I think.


nathanl:
@Gazler 👀

Gazler:
Thanks <3

PR #2804, opened 2018-03-09
Add instructions for custom domains on Heroku deploys

coderberry:
It has been very difficult for me to find solutions to simply using custom domains with Heroku and Phoenix. I have added the small change that is required to make this happen in the Heroku section of the docs.

I hope it helps!


petermm:
fwiw
host: nil
might also be relevant if your app is serving multiple domains, and/or the simpler config will work better in various cases.. eg your dns is root forwarding etc. etc.

PR #2805, opened 2018-03-10
Use static_push by default for the JS and CSS

Gazler:
static_push is not used on the image by default yo show that not
everything needs to use static_push.


installer/templates/phx_web/templates/layout/app.html.eex
```diff
@@ -25,6 +26,5 @@
       <p class="alert alert-danger" role="alert"><%%= get_flash(@conn, :error) %></p>
       <%%= render @view_module, @view_template, assigns %>
     </main>
-    <script type="text/javascript" src="<%%= Routes.static_path(@conn, "/js/app.js") %>"></script>
```
webdeb: Why move it up? I though that its best-practice to load js when the page itself finished loading. Does it have something to do with HTTP2? just asking..

webdeb: k, just learned something new.
https://www.w3schools.com/tags/att_script_defer.asp
> When present, it specifies that the script is executed when the page has finished parsing.

chrismccord:
❤️❤️❤️🐥🔥

PR #2808, opened 2018-03-14
Add link to Ecto guide

oskarrough:
Now that the guide for Ecto isn't upcoming any longer - but actually here - I thought it would be helpful with a link for readers.

### DIFF
--- a/guides/docs/adding_pages.md
+++ b/guides/docs/adding_pages.md
@@ -55,7 +55,7 @@ All of our application's static assets like js, css, and image files live in `as
 │   └── vendor
 ```
 
-There are also non web-related files we should know about. Our application file (which starts our Elixir application and its supervision tree) is at `lib/hello/application.ex`. We also have our Ecto Repo in `lib/hello/repo.ex` for interacting with the database. We'll touch on Ecto in an upcoming guide.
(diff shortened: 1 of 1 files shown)


guides/docs/adding_pages.md
```diff
@@ -55,7 +55,7 @@ All of our application's static assets like js, css, and image files live in `as
 │   └── vendor
 ```
 
-There are also non web-related files we should know about. Our application file (which starts our Elixir application and its supervision tree) is at `lib/hello/application.ex`. We also have our Ecto Repo in `lib/hello/repo.ex` for interacting with the database. We'll touch on Ecto in an upcoming guide.
+There are also non web-related files we should know about. Our application file (which starts our Elixir application and its supervision tree) is at `lib/hello/application.ex`. We also have our Ecto Repo in `lib/hello/repo.ex` for interacting with the database. You can learn more in the [guide for Ecto](https://hexdocs.pm/phoenix/ecto.html).
```
Gazler: Thanks! Could you use a local reference instead so that this works when building locally?

[Ecto guide](ecto.html)

oskarrough: Sure. Should it be `/phoenix/ecto.html`?

Gazler: @oskarrough I think it should just be `ecto.html` - I didn't realise GitHub generated the link in my above comment. I have wrapped it in a code block so it should display correctly now.

Gazler:
Thanks <3

Issue #2809, opened 2018-03-16
CSS not optimize with webpack


obsidienne:
### Expected behavior
In production, CSS should be minized.

### Actual behavior
CSS is not minized in production. 

Following this [medium article](https://medium.com/webpack/webpack-4-mode-and-optimization-5423a6bc597a), optimization is done using uglifyJS.

I think the following code is missing to optimize CSS: https://github.com/webpack-contrib/css-loader#minimize

Thanks

josevalim:
@obsidienne I assume you are using Phoenix master?

obsidienne:
@josevalim Yes, I have copied the configuration from https://github.com/phoenixframework/phoenix/tree/master/installer/templates/phx_assets

PR #2810, opened 2018-03-16
Optimize extracted css

chenxsan:
Should fix https://github.com/phoenixframework/phoenix/issues/2809

### DIFF
--- a/installer/templates/phx_assets/webpack/webpack.config.js
+++ b/installer/templates/phx_assets/webpack/webpack.config.js
@@ -21,7 +21,14 @@ module.exports = {
         test: /\.css$/,
         use: ExtractTextPlugin.extract({
           fallback: 'style-loader',
-          use: 'css-loader'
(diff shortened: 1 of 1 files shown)


chenxsan:
Btw, there's only one `webpack.config.js` file now, I would propose adding two of them, one for `developement`, one for `production`. Since there's no need to extract css when we are developing, it could [speed up our webpack build performance](https://webpack.js.org/guides/build-performance/#avoid-production-specific-tooling).

josevalim:
We use CSS changes in our live reloader. Without it we would always refresh
the whole page.
-- 


josevalim:
Don’t we have another way to say we only want to minimize in production
without having separate files?
-- 


josevalim:
Thanks @chenxsan! @Gazler provided a PR though that is conditional so I merged that instead because that's closer to what you want. :heart:

PR #2813, opened 2018-03-19
Use double quotes for consistency

leveloid:
For consistency, we should chose one and switch all .js files over to single or double quotes. 

webpack.config.js currently uses single quotes, and the rest (app.js, socket.js) use double. Same with app.css - let me know which you prefer!

### DIFF
--- a/installer/templates/phx_assets/webpack/app.js
+++ b/installer/templates/phx_assets/webpack/app.js
@@ -1,7 +1,7 @@
 // We need to import the CSS so that webpack will load it.
 // The ExtractTextPlugin is used to separate it out into
 // its own CSS file.
-import css from '../css/app.css';
+import css from "../css/app.css";
 
 // webpack automatically bundles all modules in your
 // entry points. Those entry points can be configured


josevalim:
Thanks! I believe we prefer double quotes because that’s consistent with
Elixir itself. :)
-- 


leveloid:
@josevalim I did some digging and found that the [Airbnb JavaScript Style Guide](https://github.com/airbnb/javascript#strings) and [Google CSS Style Guide](https://google.github.io/styleguide/htmlcssguide.html#CSS_Quotation_Marks) (both currently widely accepted) both recommend using single quotes.  LMK if that changes anything! :)

webdeb:
@leveloid this are internal conventions by Google and Airbnb. Widely accepted, because people all the time just say: "Oookey, Google"

There is actually no reason for the one or other approach, excerpt you have a String like "He's" or 'This is my Book "My cool Book"!'

Personally I think double qoutes for JS are better, because it also the standard for JSON. So its simpler to move your data around.

PR #2817, opened 2018-03-22
Prevent overiding existing http configuration parameters that are passed to Cowboy

shiroyasha:
I wanted to increase the maximum allowed size for a HTTP header by passing this option to Cowboy. However, I've noticed that the configuration I've set in my config files is not passed to the Cowboy adapter.

Example configuration:

``` elixir
config :my_app, MyApp.Enpoint,
  http: [protocol_options: [max_header_value_length: 8192]]
```

This option was overriden in the generated endpoint, when we called `Keyword.put(config, :http, [:inet6, port: port])`.

The new code checks if the `http` configuration is present and appends the port option.

### DIFF
--- a/installer/templates/phx_web/endpoint.ex
+++ b/installer/templates/phx_web/endpoint.ex
@@ -49,7 +49,14 @@ defmodule <%= endpoint_module %> do
   def init(_key, config) do
     if config[:load_from_system_env] do
       port = System.get_env("PORT") || raise "expected the PORT environment variable to be set"
-      {:ok, Keyword.put(config, :http, [:inet6, port: port])}
+
+      http_config = if Keyword.get(config, :http) do
+        config |> Keyword.get(:http) |> Keyword.put(:port, port)
+      else
+        [:inet6, port: port]
+      end
+
+      {:ok, Keyword.put(config, :http, http_config)}
     else
       {:ok, config}
     end


installer/templates/phx_web/endpoint.ex
```diff
@@ -49,7 +49,14 @@ defmodule <%= endpoint_module %> do
   def init(_key, config) do
     if config[:load_from_system_env] do
       port = System.get_env("PORT") || raise "expected the PORT environment variable to be set"
-      {:ok, Keyword.put(config, :http, [:inet6, port: port])}
```
josevalim: A simpler way to write this would be:

    http = [:inet6, port: port] ++ Keyword.get(config, :http, [])
    {:ok, Keyword.put(config, :http, http)}

shiroyasha: @josevalim  It seems that a default value of `false` is set for the `:http` key here: https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/endpoint/supervisor.ex#L122.

This prevents me to use `Keyword.get(config, :http, [])` because it will return `false` instead of the expected `[]`.

josevalim:
Thanks @shiroyasha but we will be moving away from the `init` configuration in Phoenix v1.4, so this will be no longer an issue!

PR #2819, opened 2018-03-23
Use @impl in application templates

fertapric:
Since https://github.com/phoenixframework/phoenix/commit/49def42a9220198527807176f84ba92d131daa72, Phoenix requires Elixir v1.5 for new apps, so it should be safe now to use `@impl`.


josevalim:
I would go with `@impl true`. I think we can also add this to Ecto.Repo and Phoenix.Endpoint inits. :)

@chrismccord what do you think?

PR #2822, opened 2018-03-29
Update application.ex templates with default docs

fertapric:
The [default documentation of the `*.Application` module for single apps](https://github.com/phoenixframework/phoenix/blob/master/installer/templates/phx_single/lib/app_name/application.ex#L2-L4) matches the template in the new [Library Guidelines](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/pages/Library%20Guidelines.md#avoid-spawning-unsupervised-processes), but not the ones for umbrella apps.


josevalim:

PR #2830, opened 2018-04-05
Update docs about async tests

devonestes:
The section warning about async tests for tests involving a database
was a little confusing to people since it's pretty out dated. This
language was introduce almost [3 years ago](https://github.com/phoenixframework/phoenix/commit/2b456abcea782684861f466d932b451b37279945),
back when async tests touching the database _were_ dangerous. Now that
this is no longer the case thanks to Ecto 2.0, this should be updated
in the guides.


Gazler:
Thanks <3

PR #2831, opened 2018-04-05
Replace extract-text-webpack-plugin with mini-css-extract-plugin

### LINKED ISSUE #2818: Webpack 4 +  ExtractTextPlugin issues
As per the comments [here](https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/749#issuecomment-374549467) it seems like efforts to support webpack 4 with the ExtractTextWebpackPlugin have come to a halt, while there are various issues with the existing beta version. Might be nice to have a look at alternatives until 1.4 is released.

Gazler:
According to the TL;DR of the comment below, mini-css-extract-plugin is
replacing extract-text-webpack-plugin which will be deprecated. Since
this does not have a built in optimizer for CSS, a separate plugin is
used. However overriding the optimizer means that UglifyJS must be
included manually (previously this was a transitive dependency of
webpack which was used automatically in production mode)

https://github.com/webpack-contrib/extract-text-webpack-plugin/issues/749#issuecomment-374549467

This closes #2818

### DIFF
--- a/installer/templates/phx_assets/webpack/package.json
+++ b/installer/templates/phx_assets/webpack/package.json
@@ -15,9 +15,11 @@
     "babel-preset-env": "^1.6.1",
     "copy-webpack-plugin": "^4.5.0",
     "css-loader": "^0.28.10",
-    "extract-text-webpack-plugin": "^4.0.0-beta.0",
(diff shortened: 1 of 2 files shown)


chrismccord:
❤️❤️❤️🐥🔥

Issue #2827, opened 2018-04-01
update Heroku deployment guide with webpack


happysalada:
The current deployment process to heroku won't work with the [current phoenix static assets buildpack](https://github.com/gjaldon/heroku-buildpack-phoenix-static) out of the box. It needs a bit of customization. I was wondering if there is a plan to update the buildpack or if the docs should be updated.

Bascially it's a matter of adding a phoenix_static_buildpack.config file with the following
```
compile="compile"
```

and having a `compile` file (just like a procfile, it's just a file named compile) with the following
```
yarn deploy
cd ..
mix phx.digest
```

If the docs need to be updated, I have time to make a PR.

chrismccord:
It should be updated as part of the 1.4 release. Thanks!

happysalada:
this was closed by #2834

PR #2834, opened 2018-04-07
update heroku deployment guide for webpack

happysalada:
# What
update the heroku deployment guide for webpack
# Comments
- by default the buildpack doesn't use node's latest LTS so it's running node 6.9, I thought it would be a nice improvement to use the LTS version (even if that version changes)
- It seems phoenix wants to use npm by default so I kept the compile file with npm.

refs #2827

### DIFF
--- a/guides/deployment/heroku.md
+++ b/guides/deployment/heroku.md
@@ -106,21 +106,24 @@ Buildpack added. Next release on mysterious-meadow-6277 will use:
 Run `git push heroku master` to create a new release using these buildpacks.
 ```
 
-next you'll need to add a config file named `phoenix_static_buildpack.config`
-with the following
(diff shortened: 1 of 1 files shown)


Gazler:
I like the idea, but I don't think we should specify exact versions in the guide - we will have to update every time a new node version is released. Instead, we should simply link to the configuration section of the buildpack with a note.

Something like:

 > This phoenix static buildpack pack can be configured to change the node version and compile options. Please refer to this [configuration section](https://github.com/gjaldon/heroku-buildpack-phoenix-static#configuration) for details.

happysalada:
@Gazler updated the guide accordingly.
I didn't squash my commits to make the PR more readable. Let me know if you need me to.

happysalada:
@Gazler just rebased off latest master. Let me know if you want any other changes.

Gazler:
Thanks <3

PR #2836, opened 2018-04-09
static-push -> static-path

NatTuck:
It works better this way. Unless there's supposed to be a static_push function that's currently missing.


chrismccord:
`static_push` exists on master and will perform an H2 push if using cowboy 2

PR #2842, opened 2018-04-12
Unify assertion for context test case template.

pwightman:
The `create` and `delete` tests were matching on the struct inside the {:ok, %Thing{} = thing} tuple, whereas the `update` test was matching in a separate assertion. They now all match inside the `{:ok, %Thing{} = thing}` assertion.

This screenshot shows the difference as it exists on `master` currently:

screen shot 2018-04-12 at 09 53 43

This was a simple/logical enough change that I figured I'd propose the solution instead of bringing it up on the mailing list, feel free to reject if not pertinent.


josevalim:

PR #2855, opened 2018-04-20
Update .formatter.exs

princemaple:
BTW the addition of `.formatter.exs` hasn't made into any releases yet.


fertapric:
According to [this comment](https://github.com/phoenixframework/phoenix/pull/2798#discussion_r173411687), the `transport` macro is going away on v1.4 and that's why it's not included.

princemaple:
Ah OK, good to know, thanks!

Issue #2466, opened 2017-08-31
invoking `:init.stop` takes 5 seconds if at least one websocket client is connected


sasa1977:
### Environment

* Elixir version (elixir -v): 1.5.0
* Phoenix version (mix deps): 1.3.0

### Expected behavior

Invoking `:init.stop` should properly terminate idle socket process as soon as possible.

### Actual behavior

If at least one client is connected over websocket, invoking `:init.stop` takes 5 seconds until it kills the socket process brutally:

```
mix phx.new test --no-ecto
cd test
iex -S mix phx.server

# open a browser and navigate to localhost:4000

iex> :init.stop
# the system stops after 5 seconds
```

If no websocket client is connected, the system stops almost immediately.

I didn't test the behaviour with long polling.

### More details

I believe the issue originates [here](https://github.com/phoenixframework/phoenix/blob/b98e8c8a33adb7521118c9bce2d6b93989075b5a/lib/phoenix/transports/websocket.ex#L108). The socket process is trapping exits, so by default it won't terminate when it receives a shutdown signal from its parent.

When handling the `:EXIT` message, [the socket process only reacts to an exit of a child](https://github.com/phoenixframework/phoenix/blob/b98e8c8a33adb7521118c9bce2d6b93989075b5a/lib/phoenix/transports/websocket.ex#L136-L137). Hence the exit sent from the parent is ignored, and the process does nothing. It will ultimately be terminated after the timeout happens in the parent, but we're waiting for 5 seconds for no real purpose.

At surface, the solution would be to detect if the `:EXIT` process originated from the parent, and if so, issue the same exit signal to children (channel processes), wait for them to terminate, or kill them brutally if they refuse to do so (e.g. they are trapping exits).

One problem is that it seems there's no information in the socket process about the parent. I believe this is due to how cowboy (or more precisely ranch) starts socket processes. This could be solved with some trickery, e.g. if the cowboy handler used `Process.info(self(), [:links])` during init phase to determine the parent (the only process which is linked). It's a hack, but it should do the job.

One thing which would be really nice is to allow channels to send some final info before they terminate. I think this might turn out to be quite tricky so maybe that should be tackled separately.

chrismccord:
Related to https://github.com/phoenixframework/phoenix/issues/2432 , in that if we properly start and supervise  channel processes, then this issue is mostly fixed, I think.

sasa1977:
[...]

PR #2861, opened 2018-04-23
Define a explicit and generic socket<->transport contract

josevalim:
This PR is still work in progress but I am submitting it early for discussion.

The main goal of this PR is to define a contract between transports and sockets. This brings two very interesting features:

1. Developers can use Phoenix.Socket as a generic communication layer, no longer coupled to channels. For example, you can implement your own callbacks such as a simple ping/pong communication layer and Phoenix will make sure it works for both websockets AND long polling.

2. Implementing a transport is much much much easier as it only needs to communicate with the socket and is no longer required to worry about channels.

Basically, the whole channels management is only one possible implementation of the socket<->transport contract. You can replace this implementation altogether by not using Phoenix.Socket and defining your own callbacks instead.

/cc @sasa1977 

### TODO

 * [ ] Non-channel based integration tests
 * [x] Deprecate transport macro in sockets
 * [x] Phoenix.Socket and adapter docs
 * [x] Update guides (transport)
 * [ ] CHANGELOG (transport/3, Phoenix.Socket.Transport, serializers, :handler)


josevalim:
Btw, the PR has an increase in the lines of code but that's because of deprecation warnings and extra documentation. Overall it should less LOC since we removed duplication in the serializers and existing transports.

lib/phoenix/channel/server.ex
```diff
@@ -183,13 +218,14 @@ defmodule Phoenix.Channel.Server do
 
   @doc false
   def init({socket, auth_payload, parent, ref}) do
+    _ = Process.monitor(socket.transport_pid)
```
Gazler: Is the match here deliberate?

lib/phoenix/transports/long_poll.ex
```diff
@@ -261,18 +230,21 @@ defmodule Phoenix.Transports.LongPoll do
...
   end
 
-  defp code_reload(conn, opts, endpoint) do
-    reload? = Keyword.get(opts, :code_reloader, endpoint.config(:code_reloader))
-    if reload?, do: Phoenix.CodeReloader.reload!(endpoint)
+  defp status_token_messages_json(conn, token, messages) do
+    status = conn.status || 200
+    messages = [?[, Enum.intersperse(messages, ?,), ?]]
+    token = Phoenix.json_library().encode_to_iodata!(token)
+    send_json(conn, "{\"status\":#{status},\"token\":#{token},\"messages\":#{messages}}")
```
Gazler: Is there any reason not to use a map here and `Phoenix.json_library().encode` the whole thing?

mix.exs
```diff
@@ -178,9 +178,7 @@ defmodule Phoenix.MixProject do
   end
 
   defp aliases do
-    [
-      "docs": ["docs", &generate_js_docs/1]
-    ]
+    ["docs": ["docs", &generate_js_docs/1]]
```
Gazler: Was this change intentional?

lib/phoenix/socket/transport.ex
```diff
@@ -118,17 +137,104 @@ defmodule Phoenix.Socket.Transport do
...
+    * "heartbeat" events in the "phoenix" topic - should just emit
+      an OK reply
+    * "phx_join" on any topic - should join the topic
+    * "phx_leave" on any topic - should leave the topic
+
+  """
+  @callback handle_in({message :: term, opts :: keyword}, state) ::
+              {:ok, state}
+              | {:reply, status :: atom, {opcode :: atom, message :: term}, state}
+              | {:stop, reason :: term, state}
```
sasa1977: Would it also be possible to support hibernation here and in `handle_info`?

lib/phoenix/channel/server.ex
```diff
@@ -10,55 +10,91 @@ defmodule Phoenix.Channel.Server do
...
+          private: channel.__socket__(:private)
+      }
+
+    instrument = %{params: payload, socket: socket}
+
+    Phoenix.Endpoint.instrument socket, :phoenix_channel_join, instrument, fn ->
       ref = make_ref()
 
-      case GenServer.start_link(__MODULE__, {socket, auth_payload, self(), ref}) do
+      case Supervisor.start_child(socket.handler, [socket, payload, self(), ref]) do
```
sasa1977: Do I get this right that this PR also changes the process structure to introduce a dedicated supervisor for each socket connection?

lib/phoenix/channel/server.ex
```diff
@@ -10,55 +10,91 @@ defmodule Phoenix.Channel.Server do
...
-      timeout ->
-        Process.exit(pid, :kill)
-        receive do
-          {:DOWN, ^ref, _, _, _} -> :ok
-        end
+    # otherwise the link in the transport will trigger it to crash.
+    pids_and_refs =
+      for pid <- pids do
+        ref = Process.monitor(pid)
+        GenServer.cast(pid, :close)
```
pdawczak: Interesting - why `cast` instead of `stop`?

lib/phoenix/socket.ex
```diff
@@ -122,50 +123,285 @@ defmodule Phoenix.Socket do
...
+    :erlang.garbage_collect(self())
+    {:ok, state}
+  end
+
+  def __info__(_, state) do
+    {:ok, state}
+  end
+
+  def __terminate__(_reason, {%{channels_inverse: channels_inverse}, _socket}) do
+    Phoenix.Channel.Server.close(Map.keys(channels_inverse))
```
michalmuskala: Wouldn't it make sense to exit the supervisor for the channels here instead of manually closing the connections?

lib/phoenix/socket/transport.ex
josevalim: It would be possible - I don't plan to tackle it in this PR though.

lib/phoenix/channel/server.ex
josevalim: Yes!

lib/phoenix/channel/server.ex
josevalim: `stop` would crash if the process is already dead or dies meanwhile.

lib/phoenix/socket.ex
josevalim: The channel supervisor is shared across all sockets.

josevalim: We could change it so each socket has its own supervisor but that would imply even more sockets per connection. The opposite concern though is that the supervisor we added may become a bottleneck. Unlikely though.

mix.exs
josevalim: I don't think it was. I will check.

lib/phoenix/transports/long_poll.ex
josevalim: If we do that, we need a special formatter only for long polling. I want to share them, which means they have to return already encoded data.

lib/phoenix/channel/server.ex
josevalim: Yes, we usually do it to say "i know this returns something meaningful, but i am sure i don't need it".

lib/phoenix/channel/server.ex
sasa1977: Cool! So I guess that this might also fix #2466 then.

lib/phoenix/channel/server.ex
sasa1977: Based on the usage below, the timeout semantics seems weird. If I read this right, it could happen that if you had say three channels which refuse to stop, and a timeout is say 5 seconds, this whole termination code might run for 15 seconds.

josevalim: Good catch! The current code is better than the previous because we at least close them all at the same time but this should be fixed and use a strategy similar to supervisors.

pdawczak: Thank you @josevalim - I certainly need to dig that one in a little bit more detail then!

lib/phoenix/transports/long_poll.ex
```diff
@@ -261,18 +190,24 @@ defmodule Phoenix.Transports.LongPoll do
...
+    token = Phoenix.json_library().encode_to_iodata!(token)
+    data = [{"status", Integer.to_string(status)}, {"token", token}, {"messages", messages}]
+    send_json(conn, data)
   end
 
-  defp code_reload(conn, opts, endpoint) do
-    reload? = Keyword.get(opts, :code_reloader, endpoint.config(:code_reloader))
-    if reload?, do: Phoenix.CodeReloader.reload!(endpoint)
+  defp send_json(conn, json) do
+    data = for {k, v} <- json, do: [?", k, ?", ?:, v]
```
michalmuskala: Are you sure the data does not contain anything that has to be escaped in JSON? Why not use the encoder?

josevalim: Yes, I am because all of the calls happen on this module.

We cannot use the encoder because the data that comes from the socket is already encoded. Otherwise we would need to encode it twice.

josevalim:

PR #2863, opened 2018-04-27
Fix installer archive installation instructions to use prod build

michallepicki:
Following current instructions from README.md caused two builds: first command built the installer with prod env, and second command was building again with dev env, and only after that installing the installer. I believe the installer should be installed after compiling with prod env?


josevalim:

PR #2866, opened 2018-04-28
improve consistency regarding code formatting (function parentheses)

michallepicki:
This is mostly about adjusting guides and docs for https://github.com/phoenixframework/phoenix/issues/2605 since the formatter won't be able to fix these but I also added / removed parentheses in code in many places. I did not touch:

1. `Kernel.send` - it seems that `send` is not in the formatter's [interal locals_without_parens list](https://github.com/elixir-lang/elixir/blob/master/lib/elixir/lib/code/formatter.ex) so it's written with parens in Elixir codebase since it's being formatted but it's without parens in docs and getting started guide so I was not sure what should be the way
2. `Logger.info`, `warn` and `error` - similarily
3. `Phoenix.HTML`'s functions in eex snippets - seems like these should stay without parens?

### DIFF
--- a/guides/adding_pages.md
+++ b/guides/adding_pages.md
@@ -152,7 +152,7 @@ defmodule HelloWeb.HelloController do
   use HelloWeb, :controller
 
   def index(conn, _params) do
-    render conn, "index.html"
+    render(conn, "index.html")
   end
 end
 ```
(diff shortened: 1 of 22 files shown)


guides/contexts.md
```diff
@@ -228,7 +228,7 @@ defmodule Hello.Accounts.User do
     field :name, :string
     field :username, :string
 
-    timestamps()
+    timestamps
```
josevalim: We definitely need to keep parens here as Elixir would emit a warning otherwise. You can remove parens for calls without args.

michallepicki: I believe I followed ecto's docs with that so it's probably another docs inconsistency

josevalim: The ecto docs are likely just out of date. This code would emit a warning from v1.4.

fertapric: Should `timestamps: 0` be removed from Ecto’s `.formatter.exs` https://github.com/elixir-ecto/ecto/blob/master/.formatter.exs#L9 then?

fertapric: If that’s the case, maybe the formatter should emit a warning when a `<function>: 0` is included in the list of `locals_without_parens` :)

josevalim: It should because it has no effect.

michalmuskala:
I think push and broadcast should have parens. I'm convinced the no-parens option should be in general reserved to some kinds of DSLs.

chrismccord:
Thanks! Can you amend the PR to only apply the parens in cases of guides, docs, and project templates? – no phoenix code format changes. We will decide/apply format changes to phoenix code itself at a later time. Otherwise looks great!

ps I agree with @michalmuskala that parens on push/broadcast make sense as they are not a DSL.

michallepicki:
@chrismccord I updated the PR

chrismccord:
❤️❤️❤️🐥🔥

PR #2873, opened 2018-05-04
Show summary of changes in v1.4 on the default index page

Gazler:
I thought this might be a good idea for the main things that have been added/changed that people may not know about.

This may just be a maintenance burden/more for people to delete when they generate a project though, so am happy for any feedback.

Here's a screenshot:

phoenix-changes

### DIFF
--- a/installer/templates/phx_web/templates/page/index.html.eex
+++ b/installer/templates/phx_web/templates/page/index.html.eex
@@ -5,6 +5,49 @@
 
 <section class="row">
   <article class="column">
+    <h3>New in v1.4</h3>
+    <ul>
+      <li>
+        <a href="https://github.com/phoenixframework/phoenix/pull/2734">
+          Jason is now the default JSON library instead of Poison
+        </a>
+      </li>
+      <li>
+        <a href="https://github.com/phoenixframework/phoenix/pull/2734">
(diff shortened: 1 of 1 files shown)


installer/templates/phx_web/templates/page/index.html.eex
```diff
@@ -3,6 +3,49 @@
...
+          <li>
+            <a href="https://github.com/phoenixframework/phoenix/pull/2801">
+              <code>Router.Helpers.static_push/2</code> function for HTTP/2 server push
+          </li>
+          </li>
+        </ul>
+      </li>
+      <li>
+        <a href="https://github.com/phoenixframework/phoenix/pull/2779">
+          Webpack 4 has been replaced by brunch
```
chrismcg: Other way round right? Brunch has been replaced with Webpack 4

Gazler: Definitely the other way round! Thanks.

michalmuskala:
I wonder how useful that would be for completely new users. Maybe instead of a full list it would be better to provide a link to some upgrade guide or the change log?

josevalim:
@michalmuskala I like this idea. Maybe a simple link under "Resources".

Gazler:
Replaced by https://github.com/phoenixframework/phoenix/pull/2881

PR #2881, opened 2018-05-13
Add changelog link to new template

Gazler:
The phoenix version string that was previously used in the dependency
string has been moved to a module attribute so it can be easily reused.
The `Version` module is used so that both "1.4.0" and "1.4" formats can
easily be generated. The latter format is used when linking to the
changelog in the template.

A test has also been added to ensure the correct version is specified in
mix.exs based on the version of the installer.

### DIFF
--- a/installer/lib/phx_new/generator.ex
+++ b/installer/lib/phx_new/generator.ex
@@ -5,6 +5,8 @@ defmodule Phx.New.Generator do
 
   @phoenix Path.expand("../..", __DIR__)
 
+  @phoenix_version Version.parse!("1.3.0")
+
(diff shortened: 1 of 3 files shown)


Gazler:
@chrismccord It's already there:

 > Update phoenix_dep in installer/lib/phoenix_new.ex and **installer/lib/phx_new/generator.ex** to "~> version to be released"

Will fix conflicts and push.

chrismccord:
❤️❤️❤️🐥🔥

chrismccord:
@Gazler looks like this has caused some failures on CI – can you take a look? <3

PR #2883, opened 2018-05-15
Optimization: Lookup table for channels in JavaScript Socket 

urmastalimaa:
The JavaScript client supports joining multiple channels using a single
Socket connection. The channels are held in an array and whenever a
message is received on the Socket, all channels in the array are
filtered over to look for matches. This means that receiving messages
using multiplexing has O(n) time complexity where n is the number of
channels.

A straightforward optimization is to use a lookup table with the topics
as keys, as the topic of a channel cannot change and every received
message contains the topic.

Although the server will close the older channel with the exact same
topic if there are duplicate channels, one Socket instance on the client
side can reference multiple channels with the same topic before the
close message is received. Thus the lookup table values must be arrays
of channels, which will have a single element the majority of the time.

Although accessing all the channels through `socket.channels` is not
directly documented, the channel array was kept to not accidentally
break clients.

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -712,6 +712,7 @@ const Serializer = {
 export class Socket {
   constructor(endPoint, opts = {}){
     this.stateChangeCallbacks = {open: [], close: [], error: [], message: []}
+    this.channelsByTopic      = {}
     this.channels             = []
     this.sendBuffer           = []
     this.ref                  = 0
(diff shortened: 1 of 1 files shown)


chrismccord:
Thanks! Unless you are you running into performance issues are have measurements to show the existing implementation is lacking I would prefer to keep the code as is and only optimize as necessary.

PR #2885, opened 2018-05-16
Update plug dep to use hex

mobileoverlord:
It looks like this was added as a git override because phoenix master depended on plug ~> 1.5 and it wasn't pushed to hex yet. I was unsure if this was still needed.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -44,7 +44,7 @@ defmodule Phoenix.MixProject do
   defp deps do
     [
       {:cowboy, "~> 1.0 or ~> 2.3", optional: true},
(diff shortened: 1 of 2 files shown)


mobileoverlord:
haha, never mind, there is definitely still stuff in master 😲

Gazler:
For the curious, this was changed to 1.5, then changed back again for some stacktrace changes.

https://github.com/phoenixframework/phoenix/commit/6d4cdfe7c57bad398e764b1b77970e885a6b69e1#diff-6023be6004fce4718dad3dafb576d258

https://github.com/elixir-plug/plug/commit/7e7164576fe181a3aaf6afd629199edf45a0a591

PR #2887, opened 2018-05-17
Add :reloadable_apps endpoint configuration

hubertlepicki:
Whenever I am working in a non-umbrella project that consists of Phoenix application, and a few dependencies, or I am working on a library and have it temporarily available on disk, I am missing the functionality of Phoenix code reloader picking up changes in those projects. I *think* it was possible to configure it before with specifying paths to additional to `web` folders (I have been adding `lib/` this way for example too), but that option is no longer around to specify which folders I want to reload and which not.

This pull request adds support for `:reloadable_apps` Endpoint configuration variable. Let's say I have a project (non-umbrella) consisting of `:ui` and `:core` applications. Currently, Phoenix code reloader will only pick up the changes in :ui.

With my pull request, the default behavior will remain the same. But you can, however specify:

    config :ui, UI.Endpoint,
      reloadable_apps: [:core, :ui],
      ...

This configuration will make both apps code reload. In similar manner you can make any dependency, also for example a library you are working on at the moment that's stored on your disk somewhere - reloadable.

When specified, this configuration option takes precedence over default behavior (in umbrella project or otherwise). So you can, for example, request that reloader compiles only a sub-set of your umbrella project - that might come in handy for big ones.

I am looking for feedback first if you think it's useful. I think if yes, I'll do some little clean up and update documentation before this is ready to being merged.

Let me know what you think about that feature. For me it'd be super useful as I do not use umbrella projects, but maybe that's not the general consensus.

### DIFF
--- a/lib/phoenix/code_reloader.ex
+++ b/lib/phoenix/code_reloader.ex
@@ -14,15 +14,24 @@ defmodule Phoenix.CodeReloader do
 
   @doc """
   Reloads code for the current Mix project by invoking the
-  `:reloadable_compilers`.
+  `:reloadable_compilers` on the list of `:reloadable_apps`.
 
   This is configured in your application environment like:
 
       config :your_app, YourApp.Endpoint,
-        reloadable_compilers: [:gettext, :phoenix, :elixir]
+        reloadable_compilers: [:gettext, :phoenix, :elixir],
+        reloadable_apps: [:ui, :backend]
 
   Keep in mind `:reloadable_compilers` must be a subset of the
   `:compilers` specified in `project/0` in your `mix.exs`.
+
+  The `:reloadable_apps` defaults to `nil`. In such case
+  default behaviour is to reload current project if it
+  consists of single app, or all applications within umbrella
+  project. You can set `:reloadable_apps` to subset of default
+  applications to reload only some of them, empty list - to
+  effectively disable code reloader, or include external
+  applications from library dependencies.
   """
   @spec reload!(module) :: :ok | {:error, binary()}
   defdelegate reload!(endpoint), to: Phoenix.CodeReloader.Server
--- a/lib/phoenix/code_reloader/server.ex
+++ b/lib/phoenix/code_reloader/server.ex
@@ -155,6 +155,8 @@ defmodule Phoenix.CodeReloader.Server do
         mix_compile_unless_stale_config(compilers)
       end
     end
+    
+    :ok
   end
 
   defp mix_compile_project(nil, _, _), do: :ok
@@ -162,6 +164,7 @@ defmodule Phoenix.CodeReloader.Server do
     if app in apps_to_reload do
       mix_compile_unless_stale_config(compilers)
     end
+
     :ok
   end
 
--- a/lib/phoenix/code_reloader/server.ex
+++ b/lib/phoenix/code_reloader/server.ex
@@ -47,13 +47,14 @@ defmodule Phoenix.CodeReloader.Server do
 
   def handle_call({:reload!, endpoint}, from, state) do
     compilers = endpoint.config(:reloadable_compilers)
+    reloadable_apps = endpoint.config(:reloadable_apps) || default_reloadable_apps()
     backup = load_backup(endpoint)
     froms  = all_waiting([from], endpoint)
 
     {res, out} =
       proxy_io(fn ->
         try do
-          mix_compile(Code.ensure_loaded(Mix.Task), compilers)
+          mix_compile(Code.ensure_loaded(Mix.Task), compilers, reloadable_apps)
         catch
           :exit, {:shutdown, 1} ->
             :error
@@ -74,6 +75,14 @@ defmodule Phoenix.CodeReloader.Server do
 
     Enum.each(froms, &GenServer.reply(&1, reply))
     {:noreply, state}
+  end
+
+  defp default_reloadable_apps() do
+    if Mix.Project.umbrella? do
+      Enum.map(Mix.Dep.Umbrella.cached, &(&1.app))
+    else
+      [Mix.Project.config()[:app]]
+    end
   end
 
   def handle_info(_, state) do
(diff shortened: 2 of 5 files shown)


lib/phoenix/code_reloader/server.ex
```diff
@@ -141,6 +142,24 @@ defmodule Phoenix.CodeReloader.Server do
...
+        Enum.member?(apps_to_compile, dep.app)
+      end)
+
+    Enum.each(deps, fn dep ->
+      Mix.Dep.in_dependency(dep, fn _ ->
+        mix_compile_unless_stale_config(compilers)
+      end)
+    end)
+
+    if Enum.member?(apps_to_compile, Mix.Project.get().project[:app]) do
```
josevalim: `Mix.Project.config[:app] in apps_to_compile`.

lib/phoenix/code_reloader/server.ex
```diff
@@ -141,6 +142,24 @@ defmodule Phoenix.CodeReloader.Server do
       :ok
     end
   end
+  defp mix_compile({:module, Mix.Task}, compilers, apps_to_compile) do
+    deps =
+      Enum.filter(Mix.Dep.cached(), fn dep ->
+        Enum.member?(apps_to_compile, dep.app)
```
josevalim: `dep.app in apps_to_compile`?

lib/phoenix/code_reloader/server.ex
```diff
@@ -47,13 +47,14 @@ defmodule Phoenix.CodeReloader.Server do
 
   def handle_call({:reload!, endpoint}, from, state) do
     compilers = endpoint.config(:reloadable_compilers)
+    reloadable_apps = endpoint.config(:reloadable_apps)
```
josevalim: Can we generalize the logic here? Something like this:

    endpoint.config(:reloadable_apps) || default_reloadable_apps()

The `default_reloadable_apps()` will return all umbrella children name (for umbrella case) or a list with only the current app in regular cases. This way we don't need conditional branches later on when compiling. And setting the list to an empty list effectively disables reloading.

hubertlepicki: yep this is a good point

josevalim: For now it is fine if we don't. If it becomes an issue we can discuss it.

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -184,6 +184,7 @@ defmodule Phoenix.Endpoint.Supervisor do
      http: false,
      https: false,
      reloadable_compilers: [:gettext, :phoenix, :elixir],
+     reloadable_apps: [],
```
josevalim: According to the changes above we default to nil.

josevalim:
Btw, we need docs in `Phoenix.Endpoint` about this. Please search for the other reloadable configs. :)

hubertlepicki:
I have updated the pull request with documentation, and also provided slightly more generic solution.

The solution would have been clearer if `Mix.Deps.cached` behaved the same in umbrella and non-umbrella projects. When in umbrella, it includes all the umbrella applications and their dependencies. When outside umbrella - it does not contain the main project application. So there is an extra clause where I first iterate over dependencies and recompile them, and need to check at the end if the main project app has been recompiled & recompile if it wasn't - but was requested either by user or by default.

I have tested this over umbrella and non-umbrella projects, but obviously more testing would be awesome if someone wants to try it.

josevalim:
> The solution would have been clearer if Mix.Deps.cached behaved the same in umbrella and non-umbrella projects. 

They behave the same. In the umbrella case, all umbrella children are dependencies. :) The parent project (i.e. the one in the root) is never included. We just don't care about it in the umbrella case (because it has no code).

lib/phoenix/code_reloader/server.ex
```diff
@@ -129,25 +138,37 @@ defmodule Phoenix.CodeReloader.Server do
...
-        end)
-      end
-    else
-      mix_compile_unless_stale_config(compilers)
-      :ok
-    end
+  defp mix_compile({:module, Mix.Task}, compilers, apps_to_reload) do
+    reloaded = mix_compile_deps(Mix.Dep.cached, apps_to_reload, compilers)
+
+    mix_compile_project(Mix.Project.get.project[:app], apps_to_reload, reloaded, compilers)
```
josevalim: `Mix.Project.config()[:app]`. This returns the cached config.

lib/phoenix/code_reloader/server.ex
```diff
@@ -76,6 +77,14 @@ defmodule Phoenix.CodeReloader.Server do
     {:noreply, state}
   end
 
+  defp default_reloadable_apps() do
+    if Mix.Project.umbrella? do
+      Enum.map(Mix.Dep.Umbrella.cached, &(&1.app))
+    else
+      [Mix.Project.get().project[:app]]
```
josevalim: `Mix.Project.config()[:app]`

lib/phoenix/code_reloader/server.ex
```diff
@@ -129,25 +138,37 @@ defmodule Phoenix.CodeReloader.Server do
...
+    deps
+    |> Enum.filter(&(&1.app in apps_to_reload))
+    |> Enum.map(fn dep ->
+      Mix.Dep.in_dependency dep, fn _ ->
+        mix_compile_unless_stale_config(compilers)
+      end
+    end)
+  end
+
+  defp mix_compile_project(nil, _, _, _), do: :ok
```
josevalim: I think we can remove this clause because I don't think it can ever be nil. Plus `nil` is never supposed to be in `to_reload` or in `reloaded`.

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -184,6 +184,7 @@ defmodule Phoenix.Endpoint.Supervisor do
      http: false,
      https: false,
      reloadable_compilers: [:gettext, :phoenix, :elixir],
+     reloadable_apps: nil,
```
josevalim: Keep alphabetical order please.

lib/phoenix/code_reloader.ex
```diff
@@ -14,15 +14,24 @@ defmodule Phoenix.CodeReloader do
...
 
       config :your_app, YourApp.Endpoint,
-        reloadable_compilers: [:gettext, :phoenix, :elixir]
+        reloadable_compilers: [:gettext, :phoenix, :elixir],
+        reloadable_apps: [:ui, :backend]
 
   Keep in mind `:reloadable_compilers` must be a subset of the
   `:compilers` specified in `project/0` in your `mix.exs`.
+
+  The `:reloadable_apps` defaults to `nil`. In such case
```
josevalim: Can we add a link to this function and to the `Phoenix.CodeReloader` when describing the `code_reloader` option in `lib/phoenix/endpoint.ex`? Otherwise I think this is quite hard to reach.

lib/phoenix/code_reloader/server.ex
hubertlepicki: we can but then we need to wrap this function call in if clause. `Mix.Project.get.project[:app]` is nil in umbrella project, so it can be nil.

hubertlepicki: and `Mix.Project.config()[:app]` behaves the same

lib/phoenix/code_reloader.ex
hubertlepicki: Makes sense. Added it here: https://github.com/phoenixframework/phoenix/pull/2887/files#diff-2e48d48e1c6c3b085dab25613ade742bR77

lib/phoenix/code_reloader/server.ex
```diff
@@ -129,25 +138,37 @@ defmodule Phoenix.CodeReloader.Server do
...
           "in such environments"
   end
 
+  defp mix_compile_deps(deps, apps_to_reload, compilers) do
+    deps
+    |> Enum.filter(&(&1.app in apps_to_reload))
+    |> Enum.map(fn dep ->
+      Mix.Dep.in_dependency dep, fn _ ->
+        mix_compile_unless_stale_config(compilers)
+      end
```
josevalim: Are you sure this `map` is returning application names? Maybe we should explicitly return the app name? Also, I think this pipeline would be clearer as a for comprehension.

hubertlepicki: This is a very good point, it is not working as expected.

In fact this check for already compiled apps does nothing and I can clean it up. In umbrella project there is no "project app", and in non-umbrella it is there but not in deps. It's not going to happen that we need to check if the main app was already compiled.

hubertlepicki:
I did not notify, but I think it is all done now.

josevalim:

stevedomin:
Thanks for adding this @hubertlepicki, it's going to make my developer life a lot better!

PR #2889, opened 2018-05-18
Provide access to X-Headers and Client-SSL Certificate to Socket.connect/2

mobileoverlord:
This PR depends on https://github.com/elixir-plug/plug/pull/715 for providing client ssl certificate access with the cowboy2 adapter to work.

I am posting this PR as a WIP to track progress towards supporting these features. We should kick around what is needed or better ways we could pass these options forward.

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -859,15 +859,15 @@ defmodule Phoenix.Endpoint do
       endpoint configuration
 
     * `connect_info` - a list of keys that represent data to be copied from
-      the transport to be made available in the user socket connect/3 callback.
+      the transport to be made available in the user socket `connect/3` callback.
       
-      Valid Keys
+      The valid keys are:
 
-        `:peer_data` - the result of Plug.Conn.get_peer_data.
-        `:x_headers` - all request headers that have an `X-` prefix.
-        `:uri` - a %URI{} with information from the conn.
+        * `:peer_data` - the result of `Plug.Conn.get_peer_data/1`.
+        * `:x_headers` - all request headers that have an "x-" prefix.
+        * `:uri` - a `%URI{}` with information from the conn.
 
-      Example
+      For example:
 
         connect_info: [:peer_data, :x_headers, :uri]
 
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -858,6 +858,19 @@ defmodule Phoenix.Endpoint do
     * `:code_reloader` - enable or disable the code reloader. Defaults to your
       endpoint configuration
 
+    * `connect_info` - a list of keys that represent data to be copied from
+      the transport to be made available in the user socket connect/3 callback.
+      
+      Valid Keys
+
+        `:peer_data` - the result of Plug.Conn.get_peer_data.
+        `:x_headers` - all request headers that have an `X-` prefix.
+        `:uri` - a %URI{} with information from the conn.
+
+      Example
+
+        connect_info: [:peer_data, :x_headers, :uri]
+
   ## Longpoll configuration
 
     * `:window_ms` - how long the client can wait for new messages
@@ -880,6 +893,19 @@ defmodule Phoenix.Endpoint do
 
     * `:code_reloader` - enable or disable the code reloader. Defaults to your
       endpoint configuration
+    
+    * `connect_info` - a list of keys that represent data to be copied from
+      the transport to be made available in the user socket connect/3 callback.
+      
+      Valid Keys
+
+        `:peer_data` - the result of Plug.Conn.get_peer_data.
+        `:x_headers` - all request headers that have an `X-` prefix.
+        `:uri` - a %URI{} with information from the conn.
+
+      Example
+
+        connect_info: [:peer_data, :x_headers, :uri]
 
   """
   defmacro socket(path, module, opts \\ []) do
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -806,23 +806,25 @@ defmodule Phoenix.Endpoint do
   @doc """
   Defines a websocket/longpoll mount-point for a socket.
 
-  Note: the `:websocket` and `:longpoll` options only have an
-  effect if the socket given as argument has no `transport`
-  declarations in it.
+  Note: for backwards compatibility purposes, the `:websocket`
+  and `:longpoll` options only have an effect if the socket
+  given as argument has no `transport` declarations in it.
 
   ## Options
 
     * `:websocket` - controls the websocket configuration.
       Defaults to `true`. May be false or a keyword list
-      of options. see "WebSocket configuration" for more info.
+      of options. See "Shared configuration" and
+      "WebSocket configuration" for the whole list.
 
     * `:longpoll` - controls the longpoll configuration.
       Defaults to `false`. May be true or a keyword list
-      of options. see "Longpoll configuration" for more info.
+      of options. See "Shared configuration" and
+      "Longpoll configuration" for the whole list.
 
     * `:shutdown` - the maximum shutdown time of each channel
-      when the endpoint is shutting down. Applies only to channel-
-      based sockets.
+      when the endpoint is shutting down. Applies only to
+      channel-based sockets.
 
   ## Examples
 
@@ -832,16 +834,16 @@ defmodule Phoenix.Endpoint do
         longpoll: true,
         websocket: [compress: true]
 
-  ## Websocket configuration
-
-    * `:timeout` - the timeout for keeping websocket connections
-      open after it last received data, defaults to 60_000ms
-
-    * `:transport_log` - if the transport layer itself should log and,
-      if so, the level
+  ## Shared configuration
+  
+  The configuration below can be given to both `:websocket` and
+  `:longpoll` keys:
 
     * `:serializer` - a list of serializers for messages. See
       `Phoenix.Socket` for more information.
+
+    * `:transport_log` - if the transport layer itself should log and,
+      if so, the level
 
     * `:check_origin` - if we should check the origin of requests when the
       origin header is present. It defaults to true and, in such cases,
@@ -871,7 +873,16 @@ defmodule Phoenix.Endpoint do
 
         connect_info: [:peer_data, :x_headers, :uri]
 
+  ## Websocket configuration
+
+  The following configuration applies only to `:websocket`.
+
+    * `:timeout` - the timeout for keeping websocket connections
+      open after it last received data, defaults to 60_000ms
+
   ## Longpoll configuration
+
+  The following configuration applies only to `:longpoll`.
 
     * `:window_ms` - how long the client can wait for new messages
       in its poll request
@@ -881,31 +892,6 @@ defmodule Phoenix.Endpoint do
 
     * `:crypto` - options for verifying and signing the token, accepted
       by `Phoenix.Token`. By default tokens are valid for 2 weeks
-
-    * `:transport_log` - if the transport layer itself should log and, if so,
-      the level
-
-    * `:check_origin` - if we should check the origin of requests when the
-      origin header is present. It defaults to true and, in such cases,
-      it will check against the host value in `YourApp.Endpoint.config(:url)[:host]`.
-      It may be set to `false` (not recommended) or to a list of explicitly
-      allowed origins.
-
-    * `:code_reloader` - enable or disable the code reloader. Defaults to your
-      endpoint configuration
-    
-    * `connect_info` - a list of keys that represent data to be copied from
-      the transport to be made available in the user socket connect/3 callback.
-      
-      Valid Keys
-
-        `:peer_data` - the result of Plug.Conn.get_peer_data.
-        `:x_headers` - all request headers that have an `X-` prefix.
-        `:uri` - a %URI{} with information from the conn.
-
-      Example
-
-        connect_info: [:peer_data, :x_headers, :uri]
 
   """
   defmacro socket(path, module, opts \\ []) do
--- a/lib/phoenix/socket.ex
+++ b/lib/phoenix/socket.ex
@@ -140,6 +140,7 @@ defmodule Phoenix.Socket do
   performing token verification on connect.
   """
   @callback connect(params :: map, Socket.t) :: {:ok, Socket.t} | :error
+  @callback connect(params :: map, Socket.t, connect_info :: map) :: {:ok, Socket.t} | :error
 
   @doc ~S"""
   Identifies the socket connection.
@@ -156,6 +157,8 @@ defmodule Phoenix.Socket do
   Returning `nil` makes this socket anonymous.
   """
   @callback id(Socket.t) :: String.t | nil
+
+  @optional_callbacks connect: 2, connect: 3
 
   defmodule InvalidMessageError do
     @moduledoc """
@@ -445,13 +448,14 @@ defmodule Phoenix.Socket do
       endpoint: endpoint,
       options: options,
       transport: transport,
-      params: params
+      params: params,
+      connect_info: connect_info
     } = map
 
     vsn = params["vsn"] || "1.0.0"
 
     case negotiate_serializer(Keyword.fetch!(options, :serializer), vsn) do
-      {:ok, serializer} -> user_connect(handler, endpoint, transport, serializer, params)
+      {:ok, serializer} -> user_connect(handler, endpoint, transport, serializer, params, connect_info)
       :error -> :error
     end
   end
@@ -534,7 +538,7 @@ defmodule Phoenix.Socket do
     end
   end
 
-  defp user_connect(handler, endpoint, transport, serializer, params) do
+  defp user_connect(handler, endpoint, transport, serializer, params, connect_info) do
     # The information in the Phoenix.Socket goes to userland and channels.
     socket = %Socket{
       handler: handler,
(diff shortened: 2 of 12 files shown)


mobileoverlord:
@josevalim I think this is ready for a review.

@chrismccord and I were discussing the cleanest way to expose some of the transport information to the socket. We talked through the idea of stuffing it into the params, but creating a mixed atoms / strings key map seemed a little dirty. We also discussed changing `UserSocket.connect/2` to `connect/3`. Instead, we landed on extracting the metadata into the socket struct as the cleanest approach.

This PR will allow the long polling and websocket transports to extract `x-headers` and `peer_data` into the socket meta.

This requires the developer to opt in by adding extra transport options to `socket/3`

```
socket "/ws", UserSocket,
      websocket: [
        check_origin: ["//example.com"],
        timeout: 200,
        extract: [:x_headers, :peer_data]
      ]
```

From there, the headers and peer data will be available in `UserSocket.connect/2`

```
def connect(_params,%{meta: %{peer_data: _peer_data, x_headers: _x_headers}} = socket) do
  {:ok, socket}
end
```

This is also implemented for LongPolling

lib/phoenix/socket.ex
```diff
@@ -65,6 +65,7 @@ defmodule Phoenix.Socket do
     * `topic` - The string topic, for example `"room:123"`
     * `transport` - An identifier for the transport, used for logging
     * `transport_pid` - The pid of the socket's transport process
+    * `meta` - Additional socket metadata ectracted from the transport
```
fertapric: Typo s/ectracted/extracted/

lib/phoenix/transports/long_poll.ex
```diff
@@ -99,8 +99,8 @@ defmodule Phoenix.Transports.LongPoll do
       "phx:lp:"
       <> Base.encode64(:crypto.strong_rand_bytes(16))
       <> (System.system_time(:millisecond) |> Integer.to_string)
-
-    arg = {endpoint, handler, opts, conn.params, priv_topic}
+    meta = Transport.extract(conn, Keyword.get(opts, :extract, []))
```
fertapric: What do you think about using `:meta` instead of `:extract` here?

```elixir
socket "/ws", UserSocket,
      websocket: [
        check_origin: ["//example.com"],
        timeout: 200,
        meta: [:x_headers, :peer_data]
      ]
```

Also, the function on transport could be renamed to `Transport.metadata(conn, keys)`, `Transport.get_metadata(conn, keys)`, or `Transport.extract_metadata(conn, keys)`.

On top of that, what do you think about using `:metadata` instead of `:meta`?

lib/phoenix/socket/transport.ex
```diff
@@ -388,6 +388,33 @@ defmodule Phoenix.Socket.Transport do
     end
   end
 
+  def extract(conn, opts) do
```
fertapric: What do you think about including a short documentation? 

Something like:

```
Extracts metadata from the transport.

## Options

    * `:peer_data` - the request peer data (if one is present)
    * `:x_headers` - custom proprietary headers, denoted by the `X-` prefix

```

lib/phoenix/transports/long_poll.ex
mobileoverlord: > What do you think about using :meta instead of :extract here?

We chose the word `extract` to try to better illustrate that this data it being extracted from the transport into the socket.

> Also, the function on transport could be renamed to Transport.metadata(conn, keys), Transport.get_metadata(conn, keys), or Transport.extract_metadata(conn, keys).

I like `Transport.extract_metadata(conn, keys)`. I was feeling a little reluctant when I named it just extract :)

> On top of that, what do you think about using :metadata instead of :meta?

`:meta` was selected for consistency in that it is used in other places like in Presence.

lib/phoenix/socket/transport.ex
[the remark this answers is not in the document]
mobileoverlord: There were strong feelings against providing access to the conn directly to prevent things like cookies and other private data to be lifted off during this phase. After a lot of conversation it was determined that we can solve mostly all problems with access to peer data and x-headers.

lib/phoenix/transports/long_poll.ex
fertapric: > We chose the word `extract` to try to better illustrate that this data it being extracted from the transport into the socket.

I see. What do you think then about `:extract_meta` for the websocket configuration, `:meta` for the property and `Transport.extract_meta/2`?

lib/phoenix/socket/transport.ex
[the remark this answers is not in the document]
rmoorman: @mobileoverlord, fair enough. I recall some resistance against access to things like cookies too.

One use case (if I am not mistaken) is not solved by just handing over `x-*` headers and peer information. For multitenant scenario's, I could imagine one will find the host information (schema, host, port basically) useful too.
(At least I think I would in order to lookup host based information without something in front like nginx that sets x-forwarded-host, port and proto; even if you do use a reverse proxy in production, I think in development it is quite common to not have one in front)

lib/phoenix/socket.ex
```diff
@@ -178,7 +179,8 @@ defmodule Phoenix.Socket do
             serializer: nil,
             topic: nil,
             transport: nil,
-            transport_pid: nil
+            transport_pid: nil,
+            meta: nil
```
chrismccord: `meta: %{}`

lib/phoenix/socket/transport.ex
chrismccord: I would `@doc false` it in this case since it's only here to be shared across the WS and LongPoll transport, not used by others. I also think naming it `extract_metadata` is the way to go since we're down below user land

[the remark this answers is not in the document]
chrismccord: For the above, I think we could also support the `:uri` option to extract the above into into a %URI{}

lib/phoenix/socket/transport.ex
```diff
@@ -388,6 +388,33 @@ defmodule Phoenix.Socket.Transport do
...
+  end
+
+  defp extract_x_headers(meta, _conn, false), do: meta
+  defp extract_x_headers(meta, conn, true) do
+    Map.put(meta, :x_headers, fetch_x_headers(conn))
+  end
+
+  defp fetch_x_headers(conn) do
+    Enum.reduce(conn.req_headers, %{}, fn({header, value}, x_headers) -> 
+      header = String.downcase(header)
```
chrismccord: headers are already downcased if I'm not mistaken, so we can skip this bit

test/phoenix/test/channel_test.exs
```diff
@@ -150,6 +150,7 @@ defmodule Phoenix.Test.ChannelTest do
     channel "foo:*", Channel, assigns: %{user_socket_assigns: true}
 
     def connect(params, socket) do
+
```
chrismccord: ✂️

mobileoverlord:
@chrismccord I rebased and force pushed the updates you requested. I have also added `:uri` as an option.

josevalim:
Note that browsers do not send fragments to the server. The userinfo is a convenience field and can always be extracted from the host.

lib/phoenix/socket.ex
```diff
@@ -178,7 +179,8 @@ defmodule Phoenix.Socket do
             serializer: nil,
             topic: nil,
             transport: nil,
-            transport_pid: nil
+            transport_pid: nil,
+            meta: %{}
```
josevalim: I don't want to make the Socket "fatter". Can we please not add this field to the socket? :)

lib/phoenix/socket/transport.ex
```diff
@@ -388,6 +388,50 @@ defmodule Phoenix.Socket.Transport do
     end
   end
 
+  @doc false
+  def extract_metadata(conn, keys) do
+    %{}
+    |> extract_peer_data(conn, Enum.member?(keys, :peer_data))
+    |> extract_x_headers(conn, Enum.member?(keys, :x_headers))
+    |> extract_uri(conn, Enum.member?(keys, :uri))
```
josevalim: This is not very efficient. It would be better to call a `Enum.reduce/3` on the keys. It would then also be a good opportunity to raise on any unknown keys.

lib/phoenix/socket/transport.ex
```diff
@@ -388,6 +388,50 @@ defmodule Phoenix.Socket.Transport do
...
+    Enum.reduce(conn.req_headers, %{}, fn({header, value}, x_headers) -> 
+      if String.starts_with?(header, "x-") do
+        Map.put(x_headers, header, value)
+      else
+        x_headers
+      end
+    end)
+  end
+
+  defp fetch_uri(conn) do
```
josevalim: Let's pattern match on the `conn` fields instead of doing multiple accesses.

lib/phoenix/test/channel_test.ex
```diff
@@ -256,22 +256,29 @@ defmodule Phoenix.ChannelTest do
   the result of the handler's `connect/2` callback.
   """
   defmacro connect(handler, params) do
+    quote do
+      connect(unquote(handler), unquote(params), %{})
```
josevalim: We should not have a macro that calls another macro.  In this case `connect/2` and `connect/3` should have a private function they call that emits the quoted expression. :)

test/phoenix/integration/websocket_channels_test.exs
```diff
@@ -142,6 +161,45 @@ defmodule Phoenix.Integration.WebSocketChannelsTest do
...
 
+      test "transport x_headers are extracted to the socket metadata" do
+        extra_headers = [{"x-application", "Phoenix"}]
+        {:ok, sock} = WebsocketClient.start_link(self(), @vsn_path, @serializer, extra_headers)
+        WebsocketClient.join(sock, "room:lobby1", %{})
+
+        assert_receive %Message{event: "joined",
+                                payload: %{"meta" => 
+                                  %{"x_headers" => 
+                                    %{"x-application" => "Phoenix"}}}}
```
josevalim: Found the lisper.

lib/phoenix/socket.ex
josevalim: I would prefer if we passed this as an argument to `connect` and let the user extract what they need. There is no reason to keep all of this information around. Although I am not sure what you and Chris discussed. In my my mind, it is ok to call `connect/3` with the meta if it is available, otherwise fallback to `connect/2`.

mobileoverlord: I believe that the discussion was on the fence between the two choices. If I understand your proposal, we would call connect/3 if the function has been exported.

josevalim: > If I understand your proposal, we would call connect/3 if the function has been exported.

Yes. @chrismccord, thoughts?

Other than this everything in this PR is perfect. I like the three options: :uri, :x_headers and :peer_data :+1:.

lib/phoenix/test/channel_test.ex
```diff
@@ -256,22 +256,29 @@ defmodule Phoenix.ChannelTest do
...
   @doc false
-  def __connect__(endpoint, handler, params) do
+  def __connect__(endpoint, handler, params, meta) do
     map = %{
       endpoint: endpoint,
       transport: :channel_test,
       options: [serializer: [{NoopSerializer, "~> 1.0.0"}]],
-      params: __stringify__(params)
+      params: __stringify__(params),
+      meta: meta
```
mobileoverlord: @josevalim Do I need to `__stringify__(meta)` here?

josevalim:
Stringify should not be needed since the keys are known upfront. They
should all be atoms.
-- 


mobileoverlord:
@chrismccord @josevalim This is ready to review again.

lib/phoenix/socket/transport.ex
```diff
@@ -388,6 +388,45 @@ defmodule Phoenix.Socket.Transport do
     end
   end
 
+  @doc false
```
josevalim: Why is it `@doc false`? :):

lib/phoenix/socket/transport.ex
```diff
@@ -388,6 +388,45 @@ defmodule Phoenix.Socket.Transport do
...
+          raise ArgumentError, "connection info keys are expected to be one of [:peer_data, :x_headers, :uri], got: #{inspect(key)}"
+      end
+    end
+  end
+
+  defp fetch_x_headers(conn) do
+    for {header, _} = pair <- conn.req_headers,
+        String.starts_with?(header, "x-"),
+        do: pair,
+        into: %{}
```
josevalim: I moved this to a comprehension but, now that i think about it, since header names may be duplicated, maybe we should return a list here? I.e. remove the `into: %{}` clause?

mobileoverlord: yeah, you're right

lib/phoenix/socket/transport.ex
mobileoverlord: its not because its private its because its undocumented :trollface:

mobileoverlord: I can add a blurb

josevalim:
@mobileoverlord :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

Only a couple things left:

  * We need to document the :connect_info option in Phoenix.Endpoint.socket
  * We need to change the UserSocket generator to generate connect/3
  * We need to document Transport.connect_info
  * We should probably convert x_headers to a list to support duplicated headers

mobileoverlord:
@josevalim ready for review

lib/phoenix/socket/transport.ex
```diff
@@ -388,6 +388,57 @@ defmodule Phoenix.Socket.Transport do
     end
   end
 
+  @doc """
+  Extracts connection information from the conn and return a map.
```
fertapric: Minor nitpick: backticks in “conn” or connection?

lib/phoenix/socket/transport.ex
```diff
@@ -388,6 +388,57 @@ defmodule Phoenix.Socket.Transport do
...
+
+  Keys are retrieved from the optional transport option `:connect_info`.
+  This functionality is transport specific. Please refer to your transports's
```
fertapric: I think it should be “transport’s”

lib/phoenix/socket/transport.ex
```diff
@@ -388,6 +388,57 @@ defmodule Phoenix.Socket.Transport do
...
+          {:peer_data, Plug.Conn.get_peer_data(conn)}
+
+        :x_headers ->
+          {:x_headers, fetch_x_headers(conn)}
+
+        :uri ->
+          {:uri, fetch_uri(conn)}
+
+        _ ->
+          raise ArgumentError, "connection info keys are expected to be one of [:peer_data, :x_headers, :uri], got: #{inspect(key)}"
```
fertapric: Maybe it’s better to use here “:connect_info keys”, due to the key and function name?

And maybe “on of :peer_data, :x_headers, or :uri”?

lib/phoenix/socket/transport.ex
josevalim: I believe it is correct. Or it would also be `transports'`.

lib/phoenix/socket/transport.ex
```diff
@@ -388,6 +388,57 @@ defmodule Phoenix.Socket.Transport do
...
+  documentation for more information.
+
+  The supported keys are:
+
+    * `:peer_data` - the result of `Plug.Conn.get_peer_data/1`.
+    * `:x_headers` - a list of all request headers that have an "x-" prefix.
+    * `:uri` - a `%URI{}` derived from the conn.
+
+  """
+  def connect_info(conn, opts) do
```
fertapric: Should the `connect_info` funcion receive only the `:connect_info` options instead of the whole `opts`? In other words, maybe `keys = Keyword.get(opts, :connect_info, opts)` should be a responsibility of the caller?

josevalim: Good catch. I believe this is also how other functions here behave.

lib/phoenix/socket/transport.ex
fertapric: Ah! I thought that was singular. I think it should be then `transports’`. Luckily, we have some native English speakers reviewing this PR 🙂

lib/phoenix/endpoint.ex
```diff
@@ -858,8 +860,30 @@ defmodule Phoenix.Endpoint do
...
+        * `:x_headers` - all request headers that have an "x-" prefix.
+        * `:uri` - a `%URI{}` with information from the conn.
+
+      For example:
+
+        connect_info: [:peer_data, :x_headers, :uri]
+
+  ## Websocket configuration
+
+  The following configuration applies only to `:websocket`.
```
fertapric: Maybe end this (and the next one) using a semicolon?

I can do this in a different PR: some items in the lists end with dot but others do not :)

fertapric: I can definitely address all those formatting issues in a different PR. There are also “inconsistencies” in Transport :)

fertapric:
I would also test other scenarios:

- connection information is not included by default
- only selected keys are the ones included

mobileoverlord:
@josevalim Are there any remaining changes that you would like me to address? Would you like me to rebase this down to a single commit?

mobileoverlord:
@josevalim @chrismccord I updated this with the change from https://github.com/phoenixframework/phoenix/pull/2889#discussion_r209421001

josevalim:

PR #2894, opened 2018-05-21
Link to OWASP article on session fixation

KronicDeth:
# Changelog
## Enhancements
*  [Context] "session fixation attack" is used without explanation, so link to OWASP article on it as OWASP is a good authority for explaining web security topics.


chrismccord:
❤️❤️❤️🐥🔥

PR #2895, opened 2018-05-21
Consistently link to localhost URLs in Contexts guide

KronicDeth:
# Changelog
## Enhancements
The Contexts Guide was inconsistent in how it styled localhost URLs: sometimes they were plain text, sometimes code, and sometimes links. Since most of the prose is directing the reader to visit the URLs, it seems best to standardize on linking to localhost URLs in all locations in the guide.


chrismccord:
❤️❤️❤️🐥🔥

PR #2898, opened 2018-05-26
Renames Router.Helpers to Router.Routes

nysertxs:


josevalim:
Thanks @nysertxs! but not this issue is with milestone phoenix 2.0, as it is a breaking change, and we don't plan to start working on it soon. For this reason, we can't merge it. Apologies!

PR #2900, opened 2018-05-29
Show a better router error message when a non-Enumerable is passed as params

Gazler:
Currently calling a path such as:

  page_page(conn, :show, :id, "extra-params")

Will raise a `Protocol.UndefinedError` as "extra-params" is not enumerable. A
custom error message is now raised in this scenario to prompt the user what to
do next.

### DIFF
--- a/lib/phoenix/router/helpers.ex
+++ b/lib/phoenix/router/helpers.ex
@@ -194,14 +194,14 @@ defmodule Phoenix.Router.Helpers do
 
       defp segments!(segments, query, reserved, {helper, opts, var_length}) do
         arity = var_length + 3
-        call_vars = if var_length > 0, do: Enum.map(1..(var_length), &(":arg#{&1}")), else: []
+        call_vars = if var_length > 0, do: Enum.map(1..(var_length), &("arg#{&1}")), else: []
 
         raise ArgumentError, """
         #{__MODULE__}.#{helper}_path/#{arity} called with invalid params. The last argument to \
         this function should be a keyword list or a map.
         For example:
 
-          Router.#{helper}_path(conn, :#{opts}, #{Enum.join(call_vars ++ [""], ", ")}page: 5, per_page: 10)
+          Router.#{helper}_path(#{Enum.join(["conn", ":#{opts}" | call_vars], ", ")}, page: 5, per_page: 10)
 
         It is possible you have called this function without defining the proper \
         number of path segments in your router.
--- a/lib/phoenix/router/helpers.ex
+++ b/lib/phoenix/router/helpers.ex
@@ -176,24 +176,37 @@ defmodule Phoenix.Router.Helpers do
       defp to_param(true), do: "true"
       defp to_param(data), do: Phoenix.Param.to_param(data)
 
-      defp segments(segments, [], _reserved) do
-        {:ok, segments}
+      defp segments!(segments, [], _reserved, _opts) do
+        segments
       end
 
-      defp segments(segments, query, reserved) when is_list(query) or is_map(query) do
+      defp segments!(segments, query, reserved, _opts) when is_list(query) or is_map(query) do
         dict = for {k, v} <- query,
                not ((k = to_string(k)) in reserved),
                do: {k, v}
 
 
         case Conn.Query.encode dict, &to_param/1 do
-          "" -> {:ok, segments}
-          o  -> {:ok, segments <> "?" <> o}
+          "" -> segments
+          o  -> segments <> "?" <> o
         end
       end
 
-      defp segments(segments, query, reserved) do
-        {:error, :invalid_query}
+      defp segments!(segments, query, reserved, opts) do
+        helper = Keyword.get(opts, :helper)
+        var_length = Keyword.get(opts, :var_length)
+        opts = Keyword.get(opts, :opts)
+        arity = var_length + 3
+        call_vars = if var_length > 0, do: Enum.map(1..(var_length), &(":arg#{&1}")), else: []
+
+        raise ArgumentError, "#{__MODULE__}.#{helper}_path/#{arity} " <>
+          "called with invalid params. The last argument to this function should " <>
+          "be a keyword list or a map.\n" <>
+          "For example:\n\n" <>
+          "  Router.#{helper}_path(conn, :#{opts}, " <>
+          "#{Enum.join(call_vars ++ [""], ", ")}page: 5, per_page: 10)\n\n" <>
+          "It is possible you have called this function without defining the proper " <>
+          "number of path segments in your router."
       end
     end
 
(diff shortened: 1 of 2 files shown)


lib/phoenix/router/helpers.ex
```diff
@@ -223,7 +241,11 @@ defmodule Phoenix.Router.Helpers do
       end
 
       def unquote(:"#{helper}_path")(conn_or_endpoint, unquote(opts), unquote_splicing(vars), params) do
-        path(conn_or_endpoint, segments(unquote(segs), params, unquote(bins)))
+        path(conn_or_endpoint, segments!(unquote(segs), params, unquote(bins), [
+          opts: unquote(opts),
+          helper: unquote(helper),
+          var_length: length(unquote(vars))
+        ]))
```
josevalim: Sorry for nitpicking but I would rather just pass them all as arguments. Or a tuple of three elements. Build a list of things that are always present is probably the slowest way to go about this. :)

lib/phoenix/router/helpers.ex
```diff
@@ -176,20 +176,38 @@ defmodule Phoenix.Router.Helpers do
...
       end
+
+      defp segments!(segments, query, reserved, opts) do
+        helper = Keyword.get(opts, :helper)
+        var_length = Keyword.get(opts, :var_length)
+        opts = Keyword.get(opts, :opts)
+        arity = var_length + 3
+        call_vars = if var_length > 0, do: Enum.map(1..(var_length), &(":arg#{&1}")), else: []
+
+        raise ArgumentError, "#{__MODULE__}.#{helper}_path/#{arity} " <>
```
josevalim: Let's use a heredoc? It will make the message clearer. :)

lib/phoenix/router/helpers.ex
```diff
@@ -176,20 +176,37 @@ defmodule Phoenix.Router.Helpers do
...
+      defp segments!(segments, query, reserved, {helper, opts, var_length}) do
+        arity = var_length + 3
+        call_vars = if var_length > 0, do: Enum.map(1..(var_length), &(":arg#{&1}")), else: []
+
+        raise ArgumentError, """
+        #{__MODULE__}.#{helper}_path/#{arity} called with invalid params. The last argument to \
+        this function should be a keyword list or a map.
+        For example:
+
+          Router.#{helper}_path(conn, :#{opts}, #{Enum.join(call_vars ++ [""], ", ")}page: 5, per_page: 10)
```
josevalim: If `call_vars` is empty, won't we have two commas? Maybe you want to use intersperse instead?

```iex
iex(2)> "foo, bar, #{Enum.intersperse ~w(arg1 arg2 arg3), ", "}, done"
"foo, bar, arg1, arg2, arg3, done"
```

lib/phoenix/router/helpers.ex
```diff
@@ -176,20 +176,37 @@ defmodule Phoenix.Router.Helpers do
...
+
         case Conn.Query.encode dict, &to_param/1 do
           "" -> segments
           o  -> segments <> "?" <> o
         end
       end
+
+      defp segments!(segments, query, reserved, {helper, opts, var_length}) do
+        arity = var_length + 3
+        call_vars = if var_length > 0, do: Enum.map(1..(var_length), &(":arg#{&1}")), else: []
```
josevalim: Those things are not literal atoms, are they? So maybe it should be `arg#{&1}` instead of `:arg#{&1}`?

test/phoenix/router/helpers_test.exs
```diff
@@ -136,6 +137,35 @@ defmodule Phoenix.Router.HelpersTest do
...
 
+  test "url helper shows an error if an id is accidentally passed" do
+    error_message = """
+    Elixir.Phoenix.Router.HelpersTest.Router.Helpers.bottom_path/5 called with invalid params. The last argument to this function should be a keyword list or a map.
+    For example:
+
+      Router.bottom_path(conn, :bottom, :arg1, :arg2, page: 5, per_page: 10)
+
+    It is possible you have called this function without defining the proper number of path segments in your router.
+    """
```
josevalim: Let's use a regex and match only on the printed helper:

```
error_suggestion = ~r/Router.bottom_path\(conn, :bottom, :arg1, :arg2, page: 5, per_page: 10\)/
```

This way if we change minor details in the message we don't break the test.

lib/phoenix/router/helpers.ex
Gazler: Intersperse in this case results in two commas. That's why I append an empty element - going to change it to a prepend though..

```elixir
"foo, bar, #{Enum.intersperse ~w(), ", "}, done"              
"foo, bar, , done"
```

josevalim: @Gazler I am silly. The interperse would be the same as join. I meant to say a map:

```iex
iex(3)> "foo, bar, #{Enum.map ~w(arg1 arg2 arg3), &[&1 | ", "]}done"
"foo, bar, arg1, arg2, arg3, done"
iex(4)> "foo, bar, #{Enum.map ~w(), &[&1 | ", "]}done"
"foo, bar, done"
```

But that is probably not better or worse than what we have.

lib/phoenix/router/helpers.ex
```diff
@@ -176,20 +176,37 @@ defmodule Phoenix.Router.Helpers do
...
+      defp segments!(segments, query, reserved, {helper, opts, var_length}) do
+        arity = var_length + 3
+        call_vars = if var_length > 0, do: Enum.map(1..(var_length), &("arg#{&1}")), else: []
+
+        raise ArgumentError, """
+        #{__MODULE__}.#{helper}_path/#{arity} called with invalid params. The last argument to \
+        this function should be a keyword list or a map.
+        For example:
+
+          Router.#{helper}_path(#{Enum.join(["conn", ":#{opts}" | call_vars], ", ")}, page: 5, per_page: 10)
```
josevalim: Yeah, this is the best way to go about it! :+1:

lib/phoenix/router/helpers.ex
```diff
@@ -223,7 +240,8 @@ defmodule Phoenix.Router.Helpers do
       end
 
       def unquote(:"#{helper}_path")(conn_or_endpoint, unquote(opts), unquote_splicing(vars), params) do
-        path(conn_or_endpoint, segments(unquote(segs), params, unquote(bins)))
+        path(conn_or_endpoint, segments!(unquote(segs), params, unquote(bins),
+              {unquote(helper), unquote(opts), length(unquote(vars))}))
```
josevalim: Hrm, I think we should rather do `unquote(length(vars))` so the work happens at compile time... but now I think I know an approach we can do to keep the variable names without doing the work at runtime time. What if you unquoted this: `unquote(Enum.map(vars, &Macro.to_string/1))`?

This way you get the variable names, and not the arguments, which will improve the error message but still be short compile-time wise.

josevalim: I am assuming those `vars` are rather decent though and not just generated var1/var2 stuff.

michalmuskala:
Looking at this, I wonder if rather this shouldn't be done using the `blame` callback for exceptions similar to how the new FunctionClauseError works. We could then do some even fancier things like walking the AST.

PR #2909, opened 2018-06-13
Add security header to avoid window control attacks - Closes #2906

### LINKED ISSUE #2906: Updated generated configuration to include headers and discussion on Spectre mitigation
At WWDC Apple gave a talk about security and safari. Specifically what is needed to defend against attacks like Spectre: https://developer.apple.com/videos/play/wwdc2018/207/

There are some additional headers they are supporting like `Cross-Origin-Window-Policy` which is needed to prevent an attacker from loading your site in a frame within their site which would allow the attacker to exploit Spectre. 

I think it's likely worthwhile to have this set by default:
`Cross-Origin-Window-Policy: deny` 
but we probably need documentation around it and discussion of the consequences.

It may also be worthwhile mentioning the techniques for guarding against an attacker compromising a CDN or replace hosted javascript files. But it's not clear to me where the best place to do this would be.

scudelletti:
Hello guys,

This PR adds a new security header that avoid window control attacks as described on issue #2906.

As mentioned in the issue/[presentation](https://developer.apple.com/videos/play/wwdc2018/207/) this new header is already [released on Safari 12](https://developer.apple.com/safari/whats-new/).

PS: This is my first contribution to an Elixir project so please let me know if I need to do something more. The `level:starter` tag was helpful. 👍

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -998,6 +998,8 @@ defmodule Phoenix.Controller do
         context of the application (like critical domain cookies)
       * x-permitted-cross-domain-policies - set to none to restrict
         Adobe Flash Player’s access to data
+      * cross-origin-window-policy - set to deny to avoid window
+        control attacks
 
   A custom headers map may also be given to be merged with defaults.
   """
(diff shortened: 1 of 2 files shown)


chrismccord:
❤️❤️❤️🐥🔥

PR #2912, opened 2018-06-15
Add create_file opts to copy_from

acrogenesis:
This adds to the optional argument `opts` to `copy_from` which can be passed on to ` Mix.Generator.create_file\3`

### DIFF
--- a/lib/mix/phoenix.ex
+++ b/lib/mix/phoenix.ex
@@ -25,8 +25,12 @@ defmodule Mix.Phoenix do
 
   Files are evaluated against EEx according to
   the given binding.
+
+  ## Options
+    * `:force` - forces installation without a shell prompt.
+
   """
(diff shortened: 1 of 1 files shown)


josevalim:
Thanks! All of this is private AP thoughI. So if we are not using this in
Phoenix, is there a reason why we should expose it?
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

acrogenesis:
We are making some custom generators which need to be run autonomously and we need to pass on `%{force: true}` so the generation can run without asking to overwrite the file Y/n.

josevalim:
Yes, but if you are building a new generator you should not rely on those functions. They are private. The whole module is private. And they can be removed and break your generators at any time. :)

PR #2917, opened 2018-06-20
Remove redundant [Y/n] in shell prompt in context generator

michallepicki:
Mix.Shell.IO's `yes?` function already [adds a `[Yn]` hint](https://github.com/elixir-lang/elixir/blob/v1.6.6/lib/mix/lib/mix/shell/io.ex#L56) to the prompt. Without this change we currently have a `Would you like proceed? [Y/n] [Yn] ` prompt.


chrismccord:
❤️❤️❤️🐥🔥

PR #2919, opened 2018-06-20
Make path example consistent with generators

mmartinson:
I think the path in the example probably points to an older generated directory structure.

### DIFF
--- a/installer/templates/phx_assets/webpack/app.js
+++ b/installer/templates/phx_assets/webpack/app.js
@@ -13,7 +13,5 @@ import css from "../css/app.css"
 
 // Import local files
 //
-// Local files can be imported directly using relative
-// paths "./socket" or full ones "web/static/js/socket".
-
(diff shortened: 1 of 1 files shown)


installer/templates/phx_assets/webpack/app.js
```diff
@@ -14,6 +14,6 @@ import css from "../css/app.css"
 // Import local files
 //
 // Local files can be imported directly using relative
-// paths "./socket" or full ones "web/static/js/socket".
+// paths "./socket" or full ones "priv/static/js/socket".
```
chrismccord: Good catch! Please simply make it:

```
// Local files can be imported directly using relative paths, for example:
// import socket from "./socket"
```

chrismccord:
ping @mmartinson :)

mmartinson:
Ooops! Thanks for the reminder.

chrismccord:
❤️❤️❤️🐥🔥

PR #2920, opened 2018-06-20
Allow running phoenix endpoint on unix domain socket

mbuhot:
Hello,

This PR adds support for running a phoenix endpoint with the Cowboy adapters on a local unix socket.
This configuration could be useful when reverse-proxying cowboy using NGINX or other web server.

The only change required to phoenix is to handle the `{:local, path}` tuple result from `:ranch:get_addr` when producing the startup message.

To enable a unix socket, the endpoint is configured with `port` 0 and a `{:local, path}` tuple for `ip` ([erlang docs reference](http://erlang.org/doc/man/inet.html#type-local_address)):

```elixir
# config.exs
config :my_app, MyAppWeb.Endpoint,
  http: [port: 0, ip: {:local, "/path/to/my_app.sock"}],
```

Additionally, the application should remove any existing file at startup, otherwise it will fail to bind:

```elixir
  # application.ex
  def start(_type, _args) do
    File.rm("/path/to/my_app.sock")
    children = [
      MyApp.Repo,
      MyApp.Endpoint
    ]
    opts = [strategy: :one_for_one, name: MyApp.Supervisor]
    Supervisor.start_link(children, opts)
  end
```

Note Erlang/OTP 21.0 is required to have a working `Plug.Static` with unix sockets. In earlier versions the implementation of `:file.sendfile` would fail. (https://bugs.erlang.org/browse/ERL-596)

### DIFF
--- a/lib/phoenix/endpoint/cowboy2_adapter.ex
+++ b/lib/phoenix/endpoint/cowboy2_adapter.ex
@@ -69,13 +69,18 @@ defmodule Phoenix.Endpoint.Cowboy2Adapter do
     "Running #{inspect endpoint} with #{server} at #{uri(scheme, ref)}"
   end
   defp uri(scheme, ref) do
-    {addr, port} = :ranch.get_addr(ref)
-    host =
-      case to_string(:inet.ntoa(addr)) do
-        "0.0.0.0" -> "localhost"
-        host -> host
-      end
+    case :ranch.get_addr(ref) do
+      {:local, unix_path} ->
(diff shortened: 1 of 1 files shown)


josevalim:
@mbuhot could you please submit a pull request to plug as well adding docs to Plug.Adapters.Cowboy2 with information on how to use it with unix sockets? Thank you!

chrismccord:
❤️❤️❤️🐥🔥

PR #2929, opened 2018-06-24
Add a reference to "Introduction to Mix"

smorin:
Add a reference to "Introduction to Mix" to make it easier for people new to Mix find it's documentation.


Gazler:
Thanks <3

PR #2941, opened 2018-06-29
Add Mix task phx.gen.cert; closes #2896

### LINKED ISSUE #2896: Provide task to generate local ssl certificate
There's a lot going on right now that is going to mean we need either a script or instructions to get SSL setup locally with SAN support.

First, HTTP/2 (H2) requires TLS. Even in development mode. 

Google Chrome will throw an H2 protocol error if you use a known bad cipher.

Firefox 48 and Chrome 58 both require SAN to trust a cert. (Unless it's a wildcard.) See https://www.chromestatus.com/feature/4981025180483584

This is actually extra important because without SAN server push (PUSH_PROMISE) will not work in Chrome.

So I don't think there's much of a choice. I think we need to instruct people to generate certs that include SAN for development.

Unfortunately, for openssl to use SAN you need to use a config file as there is no command line option to give it subject alt names.

If we generate an ssl config file like such (lets say this is app-name.cnf):

``` sh
[req]
default_bits = 4096
prompt = no
default_md = sha256
x509_extensions = v3_req
distinguished_name = dn

[dn]
C = EX
ST = Elixir
L = Phoenix
O = App
emailAddress = email@phx.sh
CN = app-name.phx.sh

[v3_req]
subjectAltName = @alt_names

[alt_names]
DNS.1 = app-name.phx.sh
DNS.2 = *.phx.sh
```

Then we can run the following to generate a cert/key pair with OpenSSL:

```sh
openssl req -new -x509 -newkey rsa:2048 -sha256 -nodes -keyout app-name.key -days 3560 -out app-name.crt -config app-name.cnf

```

voltone:


voltone:
Hmm, working with dates in a way that works across Elixir and OTP versions is... a challenge.

josevalim:
Btw, we need to update the instructions in config/dev.exs. Should we do
this now or in another PR?
-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

PR #2944, opened 2018-07-01
add a docker deployment guide

happysalada:
# Why
This PR is based on the fact that I personally struggled a lot to move from Heroku to a docker deployment anywhere. Especially
- environment variables. 
It's true that there are many ways to operate, however, I wish there was a simple way to start before people could explore other ways if they wanted. For me, that simple way was with REPLACE_OS_VARS=true (I understand that this is opinionated)
- a base docker file
There seem to be so many different files out there, I couldn't really figure out what was good and why at first. Having a simple starter would have helped a lot.

Finally, everything with Phoenix was just so good, the only problem I ever had was to deploy to anything else than Heroku.

I understand that this is a seed and might need many revisions, any feedback is more than welcome.

### DIFF
--- a/guides/deployment/docker.md
+++ b/guides/deployment/docker.md
@@ -285,6 +285,44 @@ for example
 if you need to take the server down
 `docker-compose down`
 
+### Graceful system restarts
+
+Prerequisite: docker needs to restart with systemctl, you can verify this with `systemctl | grep docker.service` you should something like `loaded active and running`
+
+in case your machine gets rebooted without your knowledge or in case your app consumes so much memory that it brings docker down, you can restart automatically your app with systemctl.
+Systemctl is a linux primitive to handle starting and stopping applications.
+
+on an ubutun system
+
+- add a .service file
+  `vi /etc/systemd/system/docker-compose-app.service`
+  add the following contents
+
+  ```
+  [Unit]
+  Description=Docker Compose Application Service
+  Requires=docker.service
+  After=docker.service
+
+  [Service]
+  WorkingDirectory=/etc/opt/my_app_name/
+  ExecStart=/usr/local/bin/docker-compose up
+  ExecStop=/usr/local/bin/docker-compose down
+  TimeoutStartSec=0
+  Restart=on-failure
+  StartLimitIntervalSec=60
+  StartLimitBurst=3
+
+  [Install]
+  WantedBy=multi-user.target
+  ```
+
+- enable the service
+  `sudo systemctl enable docker-compose-app`
+
+- start the service
+  `sudo systemctl start docker-compose-app`
+
 ### Releasing the next version
 
 Now that you have a first version working, let's release the next one.
@@ -302,3 +340,7 @@ If you make changes to your code and/or add migrations for example, you simply n
 
 - run the migration if any
   `docker-compose run web /opt/app/bin/start_server migrate`
+
+```
+
+```
--- a/guides/deployment/docker.md
+++ b/guides/deployment/docker.md
@@ -1,0 +1,304 @@
+# Introduction to Deployments with docker
+
+First things first, we need a running application to deploy it. If you don't yet have one, head to
+[Up and Running Guide](up_and_running.html).
+
+Docker enables you to deploy to the platform of your choice AWS, Google cloud, Digital ocean...
+
+In this guide we will cover a simple deployement to a single machine. There are two steps for that.
+
+- Building an image
+  Docker defines images and container. An image is a set of definitions for an environement (installed libraries...). A container is a running instance of an image. One of the intricacy of building an image is handling environement variables.
+
+- Running that image on a container on the server
+  Once the image is built, we will want to compose it with other images (a postgres image for our db) when running on production. We will cover in this part the steps to deployment on the actual server.
+
+## Building our production image
+
+The recommended way to deploy elixir applications is with a release. To build our release we will use [Distillery](https://github.com/bitwalker/distillery). We will first add the necessary to build a release, then handle environment variables, then finally build our release inside our image.
+
+### Building a release
+
+The distillery [doc](https://hexdocs.pm/distillery/getting-started.html) is good and short, read and follow the instructions. Here are the steps you should have taken so far
+
+- add distillery dependency to your mix.exs file (below defp deps do)
+  `{:distillery, "~> MAJ.MIN", runtime: false}`
+  (replace MAJ and MIN with the latest major and minor version of the library)
+
+- run the `mix release.init` task
+
+- make modifications for the migrations. Detailed [here](https://hexdocs.pm/distillery/running-migrations.html#content) (add rel/commands/migrate.sh and modify rel.config.exs)
+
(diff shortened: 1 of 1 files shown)


guides/deployment/docker.md
```diff
@@ -0,0 +1,304 @@
+# Introduction to Deployments with docker
```
fertapric: Looking at https://www.docker.com/, it seems it should be “Docker”

guides/deployment/docker.md
```diff
@@ -0,0 +1,304 @@
...
+
+First things first, we need a running application to deploy it. If you don't yet have one, head to
+[Up and Running Guide](up_and_running.html).
+
+Docker enables you to deploy to the platform of your choice AWS, Google cloud, Digital ocean...
```
fertapric: I would add a semicolon after choice “...the platform of your choice: AWS, Google Cloud, Digital Ocean...” 

Note also the capital letters on the names.

guides/deployment/docker.md
```diff
@@ -0,0 +1,304 @@
...
+
+In this guide we will cover a simple deployement to a single machine. There are two steps for that.
```
fertapric: s/deployement/deployment/

fertapric: I would also add a semicolon here “There are two steps for that:”

fertapric:
As @happysalada mentions in the PR description, this needs a few revisions before being published. Because this documentation establishes a “recommended way” to deploy Phoenix applicarions using Docker, I’ll wait until one of the core members comment on this.

happysalada:
@fertapric thanks for taking the time to review ❤️ 
I'll wait until I get more feedback and bundle the changes together

guides/deployment/docker.md
happysalada: when you say semicolon, I think you mean colon :-)

guides/deployment/docker.md
```diff
@@ -0,0 +1,304 @@
...
+once the message that the database is ready appears, you can Ctrl-c to exit the container
+
+Now you can run your setup with
+`docker-compose up`
+
+go visit `localhost` on your browser and verify that your application is running.You can then exit the running containers with Ctrl-c.
+
+### Running your images on the server
+
+Prerequisite: you need to have docker and docker-compose installed on your server. (digital ocean has these one-click droplets for example, that are vps you can setup with one click that already have docker and docker-compose)
```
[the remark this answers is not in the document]
happysalada: By no means I am a devops, but isn't kubernetes or docker swarm for when you want to deploy with several servers?
could you detail a bit the downsides of running docker-compose? Then perhaps we can just list downsides (if they are not that harmful) and add later a section on how to use Kubernetes or Docker swarm?
There is a good google cloud tutorial on [how to deploy phoenix to kubernetes](https://cloud.google.com/community/tutorials/elixir-phoenix-on-kubernetes-google-container-engine), this is actually where I found the best dockerfile. However, I found that kubernetes adds another layer of complexity.
- You are now deploying to several machines, and that leads to further questions, are they connected in erlang terms, well no.
- Also depending on where you deploy, the configuration will be different (on AWS the database ressource is such and such, with such and such settings). I didn't want to start advocating a particular way, because I thought that it would create more opinionated discussions.

so how about that.
- we leave that section but rename it to `deploying to a single machine`. Add the potential problems that you might run into with docker-compose.
- Then perhaps later somebody can add a section with `deploying to several machines` or `scaling your application` and detail deployement with kubernetes.

I also setup continuous deployement with CircleCI (which have a great free plan btw), but I thought that it was too much promoting a particular service. If no one is against it, I can add my CD setup next.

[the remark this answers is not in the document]
marpaia: The downsides of "deploying" software with docker-compose on a single-host is that there is no resilient process restarting when processes crash, no way to reason about high availability (due to one server), no integrated way to ensure the disk with your database is in a safe, persistent location, etc.

I am by no means the arbiter of what goes in the docs, but I love the idea of perhaps enumerating the potential problems that you might run into with docker-compose at the end. And perhaps suggest some alternatives (like that GCP article) for further information?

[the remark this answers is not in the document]
happysalada: I added system restart handling with systemctl on 
0a24397
could you detail a bit what you mean by `no way to reason about high availability`
so how about the following 

Potential downsides of deploying on a single machine:
- There is no integrated way to ensure that the disk with your database is in a safe, persistent location. (to mitigate that, you can backup your server. On Digital Ocean you can enable backup of your droplet when creating it)

if you want to add something else, and you have a bit of time to write the sentence, I am more than happy to add it.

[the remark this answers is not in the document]
ellispritchard: May I suggest using a multi-stage Docker file, so that the final docker image is as small as possible? After all if we are using a release, we don't need an erlang/elixir install on the box.

For an example: https://github.com/Financial-Times/docker-elixir-build#a-multi-stage-minimal-image-build

Gazler:
First off, let me commend you on a well written guide!

I think the contents here are useful, and would make an excellent blog post, however I am not sure that the contents belong in the core guides.

Docker is a complicated technology, with many different ways of doing this. For example, this guide builds the release and runs it in the same container, however it is also possible to build the release separately and only include the release in the container and run that.

As also mentioned, you're using docker compose, but there are many container orchestration tools available that could be used instead. This guide also covers pushing to docker hub, which is an excellent suggestion, but again, there are many alternatives available. As new versions of Docker come out, this guide will need to be updated to reflect that.

When we merge a guide into the repository, we (as a core team) accept responsibility for the contents of the guides in terms of updates, maintenance and validating correctness. This is a time consuming task when it comes to the code we have already written, and adding in external dependencies such as Docker increase this maintenance burden.

Thanks for your understanding.

---

I think perhaps what we need is a "third party deployment" page in the guides which links to resources such as this. A way that we can say "We cannot guarantee the techniques in these guides work/still work". This gives us the benefit of resources like this being discoverable, without additional maintenance burden.

webdeb:
> I think perhaps what we need is a "third party deployment" page in the guides which links to resources such as this. A way that we can say "We cannot guarantee the techniques in these guides work/still work". This gives us the benefit of resources like this being discoverable, without additional maintenance burden.

Even this alternative should questioned, people are already asking about `node` installations in this repository, if docker guides will be merged, it will be the same. I think a "useful third-party resources" page with a table of links to blogs and projects is a better choice, instead of a complete guide.

happysalada:
@Gazler thanks for taking the time to craft a nice response. I perfectly understand, no worries.

happysalada:
for those interested, I made a [phoenix unofficial guide repo](https://github.com/happysalada/phoenix-the-unoffcial-guide) including this, and CI-CD integration.
I'm hoping to add server instrumentation next week.
The idea is to provide good documentation for things that can't be documented officially.
@ellispritchard I incorporated what you did in the FT repo with the app_version, that was super helpful, thanks!

beardedeagle:
Potentially this is something that could live in the git wiki as user recipes, a la absinthe style

PR #2948, opened 2018-07-06
Add origin check for transports using an MFA tuple

rmoorman:
As briefly discussed on IRC a few days ago (with @josevalim), this PR introduces the possibility for transports to receive an MFA tuple as configuration option for their origin check.

If there are things wrong, missing or to be improved, please let me know.


chrismccord:
❤️❤️❤️🐥🔥

rmoorman:
You are very welcome!

rmoorman:
@chrismccord, you mean when running the phoenix socket javascript on an HTML page with a `file://` url? And I agree that receiving the an %URI struct would be cleaner. Would it be possible to not only pass the host but the whole origin though?

PR #2953, opened 2018-07-09
Instrument error rendering

### LINKED ISSUE #2865: Instrument error rendering
This can also be a good hook for error reporting.

chrismccord:
Closes #2865


josevalim:
@chrismccord i pushed a commit to avoid duplication around send_resp and that introduces some varibles to avoid large function calls. Please let me know what you think!

josevalim:
@chrismccord oh, I also changed the name to `__prepare_conn_and_render__` since you have already called `instrument_render`. I don't think we need to say one is with and the other is without. As long as one is clear that it has instrumentation, we can assume the other does not.

josevalim:
Actually, I just realized that `send_resp/3` stores things in the connection anyway, so the `__put_render__` that you earlier proposed is the best way to go. That's the name I picked.

PR #2956, opened 2018-07-11
Socket endPoint constructor fix

tim2CF:
```
this.endPoint  = `${endPoint}/${TRANSPORTS.websocket}`
```
implementation formally is not correct. If initial `endPoint` contains for example query-string parameters, final URL will be invalid after this concatination. 

In this PR I provided a more generic solution.


chrismccord:
Query string params can be passed via the `params` option. If you feel docs should be updated to better reflect the behavior, please send a PR. Thank you!

PR #2957, opened 2018-07-12
changed guides mentioning *_path and moduledoc in lib/phoenix/view.ex

mnussbaumer:
I think I changed it in all relevant points. I also changed the notes/explanations, let me know if the language is fine and if anything else is missing/needs change. I tried rgrep for _path outside the guides, but it's a never ending list of hits for unrelated stuff, I don't think I have missed something but it could happen

### DIFF
--- a/guides/phoenix_mix_tasks.md
+++ b/guides/phoenix_mix_tasks.md
@@ -231,11 +231,8 @@ Important: If we don't do this, we will see the following warnings in our logs, 
 $ mix phx.server
 Compiling 17 files (.ex)
 
-== Compilation error in file lib/hello_web/controllers/post_controller.ex ==
-** (CompileError) lib/hello_web/controllers/post_controller.ex:22: undefined function HelloWeb.Router.Helpers.post_path/3
-    (stdlib) lists.erl:1338: :lists.foreach/2
-    (stdlib) erl_eval.erl:670: :erl_eval.do_apply/6
-    (elixir) lib/kernel/parallel_compiler.ex:121: anonymous fn/4 in Kernel.ParallelCompiler.spawn_compilers/1
+warning: function HelloWeb.Router.Helpers.post_path/3 is undefined or private
+  lib/hello_web/controllers/post_controller.ex:22: 
 ```
 
 If we don't want to create a context or schema for our resource we can use the `--no-context` flag. Note that this still requires a context module name as a parameter.
@@ -260,17 +257,14 @@ Add the resource to your browser scope in lib/hello_web/router.ex:
     resources "/posts", PostController
 ```
 
-Important: If we don't do this, our application won't compile, and we'll get an error.
+Important: If we don't do this, we'll get the following warning in our logs and the application will error when attempting to load the page:
 
 ```console
 $ mix phx.server
 Compiling 15 files (.ex)
 
-== Compilation error in file lib/hello_web/views/post_view.ex ==
-** (CompileError) lib/hello_web/templates/post/edit.html.eex:3: undefined function HelloWeb.Router.Helpers.post_path/3
-    (stdlib) lists.erl:1338: :lists.foreach/2
-    (stdlib) erl_eval.erl:670: :erl_eval.do_apply/6
-    (elixir) lib/kernel/parallel_compiler.ex:121: anonymous fn/4 in Kernel.ParallelCompiler.spawn_compilers/1
+warning: function HelloWeb.Router.Helpers.post_path/3 is undefined or private
+  lib/hello_web/templates/post/edit.html.eex:3
 ```
 
 Similarly - if we want a context created without a schema for our resource we can use the `--no-schema` flag.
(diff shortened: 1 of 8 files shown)


guides/phoenix_mix_tasks.md
```diff
@@ -232,7 +232,7 @@ $ mix phx.server
 Compiling 17 files (.ex)
 
 == Compilation error in file lib/hello_web/controllers/post_controller.ex ==
-** (CompileError) lib/hello_web/controllers/post_controller.ex:22: undefined function post_path/3
+** (CompileError) lib/hello_web/controllers/post_controller.ex:22: undefined function Routes.post_path/3
```
Gazler: Forgot to mention this before, but  these CompileErrors will now be 

 > warning:  function HelloWeb.Router.Helpers.post_path/3 is undefined or private
 >    lib/hello_web/controllers/post_controller.ex:22

Gazler: We should also update the text on line 228 to say:

 Important: If we don't do this, we will see the following warnings in our logs, and our application will error when trying to execute the function.

guides/routing.md
```diff
@@ -250,10 +250,10 @@ iex> HelloWeb.Router.Helpers.page_path(HelloWeb.Endpoint, :index)
...
 
 ```html
-<a href="<%= page_path(@conn, :index) %>">To the Welcome Page!</a>
+<a href="<%= Routes.page_path(@conn, :index) %>">To the Welcome Page!</a>
 ```
 
-The `page_path` function is imported into our template with `use HelloWeb, :view`. Please see the [View Guide](views.html) for more information.
+The reason we can use `Routes.page_path` instead of the full `HelloWeb.Router.Helpers.page_path` name is because `HelloWeb.Router.Helpers` is aliased as `Routes` by default in the `view/0` definition (`lib/hello_web.ex`) and made available to our templates through `use HelloWeb, :view`. We can, of course, use `HelloWeb.Router.Helpers.page_path(@conn, :index)` instead, but the convention is to use the aliased version for conciseness (note that the alias is only set automatically for use in views, controllers and templates - outside these you need either the full name, or to alias it yourself inside the module definition: `alias HelloWeb.Router.Helpers, as: Routes`). Please see the [View Guide](views.html) for more information.
```
Gazler: Great explanation!

guides/routing.md
```diff
@@ -344,6 +344,15 @@ iex> HelloWeb.Router.Helpers.user_post_path(Endpoint, :index, 42, active: true)
 "/users/42/posts?active=true"
 ```
 
+If we had aliased the `Helpers` module as before (it is only automatically aliased for views, templates and controllers, in this case, since we're inside `iex` we need to do it ourselves), we could instead do:
```
Gazler: I like the additional reinforcement of why `Routes` works elsewhere here.

guides/views.md
```diff
@@ -94,7 +95,9 @@ Now we can reload the page and view source to see what we have.
 <a href="/">Link back to this page</a>
 ```
 
-Great, `page_path/2` evaluated to `/` as we would expect, and we didn't need to qualify it with `Phoenix.View`.
+Great, `Routes.page_path/2` evaluated to `/` as we would expect, we just had to use the alias set in `Phoenix.View`.
+
+If you happen to need access to the path helpers outside views, controllers or templates, you can either call them by the full qualified name, `HelloWeb.Router.Helpers.*_path()` or alias it yourself in the calling module, by defining `alias HelloWeb.Router.Helpers, as: Routes` in the module you want to use, and then calling `Routes.*_path()`, where `Hello*` is the name of your actual application.
```
Gazler: Let's change this to use a concrete path, as `*_path` could be confusing. I'd use `page_path`.

I think we can also get rid of `, where `Hello*` is the name of your actual application.` because we are using the `HelloWeb` application throughout.

Gazler:
There's also the following reference which needs updating:

`./channels.md:404:<script src="<%= static_path(@conn, "/js/app.js") %>"></script>`

guides/views.md
mnussbaumer: Agree, it actually doesn't make much sense as we're speaking of specific cases.

guides/phoenix_mix_tasks.md
mnussbaumer: Will do, totally forgot that the alias would show missing funs on the aliased module .

mnussbaumer:
@Gazler will do these changes later today, I also noticed in the module doc I left a <% end %> statement that I will remove too

lib/phoenix/view.ex
```diff
@@ -10,18 +10,26 @@ defmodule Phoenix.View do
 
   ## Examples
```
mnussbaumer: @Gazler I also changed `lib/web`* etc, to `lib/your_app_web`* etc. It seems to make more sense since we use snippets of supposedly `YourAppWeb`?

Lmk

Gazler:
Thanks <3

PR #2961, opened 2018-07-13
Update docs urls

brandoncc:
As a new user, I started my first app and found that the docs links on the default landing page led to 404s. This PR updates all URLs which started with "phoenixframework.org/docs" to the URL which I felt best replaced it. There may be other URLs which should also be updated, but I am not familiar with the codebase yet, so I am not aware of any others. If there are, I would be happy to update them as well if I am pointed in the right direction.

### DIFF
--- a/guides/channels.md
+++ b/guides/channels.md
@@ -181,7 +181,7 @@ Phoenix ships with a JavaScript client that is available when generating a new P
   - [dn-phoenix](https://github.com/jfis/dn-phoenix)
 
 ## Tying it all together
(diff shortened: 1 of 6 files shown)


chrismccord:
Thank you! We switched the phoenix site to GitHub pages, and it looks like we were relying on S3 redirects for the old guides which is something we should have fixed long ago. Given that folks with the <= 1.3 generators will get bad links, we should probably try to get our redirects back // @Gazler @jeregrine this doable on GH pages?

PR #2962, opened 2018-07-14
Fix incorrect `:halt` key name in the guides

brandoncc:
While the function is named `Plug.Conn.halt/1`, the key that is updated is `:halted`. I believe this is a typo in the guides.


chrismccord:
❤️❤️❤️🐥🔥

PR #2963, opened 2018-07-15
Add load_from_system_env option as true in heroku deployment guide

rohanpujaris:
When we copy prod.exs example from [heroku deploy guide](https://hexdocs.pm/phoenix/heroku.html#making-our-project-ready-for-heroku), load_from_system_env config is missing and it leads to following error while deploying in heroku logs

```
Error R10 (Boot timeout) -> Web process failed to bind to $PORT within 60 seconds of launch
```


rohanpujaris:
@josevalim Oops :) Closing it.

PR #2966, opened 2018-07-16
Adjust headings level to form proper document outline

kennethlarsen:
I noticed that all the generated HTML contained nonconsecutive heading level which is a problem for accessibility (screen readers needs a proper document outline to understand the content of the page). As a bonus, it provides better SEO as well.

By changing the levels they are now living up to the WCAG standards.

This means that the headings now look a bit different since the styling used for the headings are a bit different. In my opinion, this makes it more readable.  I've added screen shots for you to see.
screen shot 2018-07-16 at 19 00 23
screen shot 2018-07-16 at 19 02 02
screen shot 2018-07-16 at 19 02 20
screen shot 2018-07-16 at 19 02 28


chrismccord:
❤️❤️❤️🐥🔥

PR #2972, opened 2018-07-18
Fix context alias generation

rbino:
If the app name was `Prefixed.App`, the generated tests had
```
alias Prefixed.App.Accounts
```
but then called
```
App.Accounts.create_user
```
The same would happen for contexts with names like ```Management.Accounts```.

This fixes that.

### DIFF
--- a/lib/mix/phoenix/context.ex
+++ b/lib/mix/phoenix/context.ex
@@ -25,7 +25,7 @@ defmodule Mix.Phoenix.Context do
     ctx_app   = opts[:context_app] || Mix.Phoenix.context_app()
     base      = Module.concat([Mix.Phoenix.context_base(ctx_app)])
     module    = Module.concat(base, context_name)
-    alias     = module |> Module.split() |> tl() |> Module.concat()
+    alias     = Module.concat([module |> Module.split() |> List.last()])
     basedir   = Phoenix.Naming.underscore(context_name)
     basename  = Path.basename(basedir)
     dir       = Mix.Phoenix.context_lib_path(ctx_app, basedir)
--- a/test/mix/tasks/phx.gen.context_test.exs
+++ b/test/mix/tasks/phx.gen.context_test.exs
@@ -47,7 +47,7 @@ defmodule Mix.Tasks.Phx.Gen.ContextTest do
       context = Context.new("Site.Blog", schema, [])
 
       assert %Context{
-        alias: Site.Blog,
+        alias: Blog,
         base_module: Phoenix,
         basename: "blog",
         module: Phoenix.Site.Blog,


chrismccord:
So this was by design since we only wanted to drop the "MyApp" (first) prefix. You are of course free to alias down to the last namespace, but it's not always desired if you have namespaces that share the same alias, for example `Management.Accounts` and `Employee.Accounts` may be clearer to remain with the prefix vs naked `Accounts` since it's more obvious which is which. Thanks!

rbino:
@chrismccord the problem is that as it is `mix phx.gen.context` generates broken tests in two different cases.

1- If your app has a nested namespace
```
mix phx.new namespaced_test1 --module Namespaced.Test1 --no-brunch --no-html
cd namespaced_test1
mix phx.gen.context Accounts User users name:string
mix test
```
outputs
```
...
  1) test users change_user/1 returns a user changeset (Namespaced.Test1.AccountsTest)
     test/namespaced_test1/accounts/accounts_test.exs:60
     ** (UndefinedFunctionError) function Test1.Accounts.create_user/1 is undefined (module Test1.Accounts is not available)
...
```
this case alone can be fixed either with the proposed PR or with
```diff
diff --git a/lib/mix/phoenix/context.ex b/lib/mix/phoenix/context.ex
index d0bae4c..5705931 100644
--- a/lib/mix/phoenix/context.ex
+++ b/lib/mix/phoenix/context.ex
@@ -24,8 +24,9 @@ defmodule Mix.Phoenix.Context do
   def new(context_name, %Schema{} = schema, opts) do
     ctx_app   = opts[:context_app] || Mix.Phoenix.context_app()
     base      = Module.concat([Mix.Phoenix.context_base(ctx_app)])
+    base_tok  = Module.split()
     module    = Module.concat(base, context_name)
-    alias     = module |> Module.split() |> tl() |> Module.concat()
+    alias     = module |> Module.split() |> Kernel.(--) base_tok |> Module.concat()
     basedir   = Phoenix.Naming.underscore(context_name)
     basename  = Path.basename(basedir)
     dir       = Mix.Phoenix.context_lib_path(ctx_app, basedir)
```
but this latter fix doesn't fix case number 2.

2 - If you try to generate a nested context like the example you gave above
```
mix phx.new test1 --no-brunch --no-html
cd test1
mix phx.gen.context Management.Accounts User users name:string
mix test
```
outputs
```
...
  1) test users update_user/2 with invalid data returns error changeset (Test1.Management.AccountsTest)
     test/test1/management/accounts/accounts_test.exs:48
     ** (UndefinedFunctionError) function Management.Accounts.create_user/1 is undefined (module Management.Accounts is not available)
...
```
This is because the alias in the template is always generated from `context.module`, so it always aliases the full name down to the last token (so the tests should use `Accounts` instead of `Management.Accounts`).

My PR fixed this emitting `Accounts` instead of `Management.Accounts` in the alias use, since I thought that the type of account was reasonably desumed from which test we're in (the context alias is used only in its tests, so it doesn't end up in places where there could be other Accounts aliases).

If you want to be able to use `Management.Accounts` as alias in the test then one solution could be 
```diff
diff --git a/priv/templates/phx.gen.context/context_test.exs b/priv/templates/phx.gen.context/context_test.exs
index 8958ec4..766383b 100644
--- a/priv/templates/phx.gen.context/context_test.exs
+++ b/priv/templates/phx.gen.context/context_test.exs
@@ -1,5 +1,5 @@
 defmodule <%= inspect context.module %>Test do
   use <%= inspect context.base_module %>.DataCase
 
-  alias <%= inspect context.module %>
+  alias <%= inspect context.module %>, as: <%= inspect context.alias %>
 end
```
but it would be a bit redundant for non-namespaced contexts.

Let me know what you think and if you have alternative solutions I can submit.

chrismccord:
I see the issue now. You are right the most straight forward approach is the regular alias . Thanks!

PR #2974, opened 2018-07-18
Don't rely on Ecto.Changeset when `--no-schema` is passed.

### LINKED ISSUE #2968: Generated FallbackController blindly relies on Ecto
### Environment

* Elixir version (elixir -v): Elixir 1.6.5
* Phoenix version (mix deps): phoenix 1.3.3
* Operating system: Linux

### Expected behavior
When using `--no-ecto` with `mix phx.new`, other generators shouldn't write code that depends on Ecto.

### Actual behavior
Calling `mix phx.gen.json`, even with `--no-schema`, writes the following in FallbackController:
```elixir
  def call(conn, {:error, %Ecto.Changeset{} = changeset}) do
    conn
    |> put_status(:unprocessable_entity)
    |> render(NoectotestWeb.ChangesetView, "error.json", changeset: changeset)
  end
```
Now it doesn't compile because Ecto isn't a dep of the app.

Ecto's very useful, so most users will be using it. But I think it's reasonable for the generators to check that Ecto is available before trying to use it, especially if `--no-schema` is being used.

IslandUsurper:
To fix #2968.

Since FallbackController is usually not edited for a while after being generated, we shouldn't write an automatic compiler error into it.

### DIFF
--- a/priv/templates/phx.gen.json/fallback_controller.ex
+++ b/priv/templates/phx.gen.json/fallback_controller.ex
@@ -6,14 +6,14 @@ defmodule <%= inspect context.web_module %>.FallbackController do
   """
   use <%= inspect context.web_module %>, :controller
 
-  def call(conn, {:error, %Ecto.Changeset{} = changeset}) do
+  <%= if schema.generate? do %>def call(conn, {:error, %Ecto.Changeset{} = changeset}) do
     conn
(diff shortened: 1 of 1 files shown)


chrismccord:
❤️❤️❤️🐥🔥

PR #2982, opened 2018-07-22
Minor grammar fix for router.ex doc

omh:
"put less pressure in the Elixir compiler" -> "put less pressure on the Elixir compiler"


chrismccord:
❤️❤️❤️🐥🔥

PR #2984, opened 2018-07-26
Use “static_url” helper instead of “static_path” in templates

remiprev:
I don’t think there is a valid reason to use `static_path` instead of `static_url` in default templates 😄 

`Phoenix.Endpoint` exposes (and documents) the `:static_url` runtime configuration but it’s not used by `static_path` when the user configures it.

If the `static_url` helper is used but `:static_url` is not configured, the fallback makes sure assets are still served from the application URL (eg. `localhost`).

### DIFF
--- a/guides/channels.md
+++ b/guides/channels.md
@@ -401,7 +401,7 @@ Next we need to pass this token to JavaScript. We can do so inside a script tag 
 
 ```html
 <script>window.userToken = "<%= assigns[:user_token] %>";</script>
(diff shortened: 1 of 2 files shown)


josevalim:
Agreed! :+1:

chrismccord:
❤️❤️❤️🐥🔥

50kudos:
I have one reason because I come here to see why `static_path` has been changed to `static_url`

Path is domain independent for built-in assets! So this won't work with generated domain names.
So I just change back to `static_path`.

Just a note. Thanks!

PR #2990, opened 2018-07-31
give a warning when a plug is forwarded more than once

tyrchen:
For some plugs, like Absinthe, it uses init opts to take data so that the plug can behave differently (e.g. for different graphql schema). Thus below cases are valid by using plug, but cannot compile by using phoenix since it complains plug has already been forwarded to.

Use case:

```
  scope "/api" do
    pipe_through(:graphql)

    forward("/btc", Absinthe.Plug, schema: Bitcoin)
    forward("/eth", Absinthe.Plug, schema: Ethereum)
  end
```

### DIFF
--- a/lib/phoenix/router/route.ex
+++ b/lib/phoenix/router/route.ex
@@ -4,6 +4,7 @@ defmodule Phoenix.Router.Route do
   # as it contains internal routing information.
   @moduledoc false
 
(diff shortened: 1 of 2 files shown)


tyrchen:
It could be worked around by defining new plugs.

PR #2994, opened 2018-08-03
Copy peer data when recycling the conn

mobileoverlord:
This requires code that is in Plug master right now. When the connection is recycled the peer information is reset back to default values and the ssl_cert is lost. This makes it really difficult to test interactions with an api that is protected using client SSL


josevalim:
Plug v1.6.2 is out, i hvae updated this PR accordingly.

josevalim:

PR #3004, opened 2018-08-12
Format documentation of mix tasks

fertapric:


lib/mix/tasks/phx.gen.channel.ex
```diff
@@ -55,6 +56,7 @@ defmodule Mix.Tasks.Phx.Gen.Channel do
     mix phx.gen.channel expects just the module name:
 
         mix phx.gen.channel Room
+
```
fertapric: I had other PRs upcoming that include a new line on error messages in case there is indented content at the end. Is that OK?

lib/mix/tasks/phx.digest.clean.ex
```diff
@@ -21,14 +21,14 @@ defmodule Mix.Tasks.Phx.Digest.Clean do
   ## Options
 
     * `-o, --output` - indicates the path to your compiled
-      assets directory. Defaults to `priv/static`.
+      assets directory. Defaults to `priv/static`
```
fertapric: The documentation had a mix of list items ending with and without a dot. I decided to follow Elixir's style and remove dots at the end :)

PR #3005, opened 2018-08-12
Format doc of Mix.Phoenix

fertapric:


lib/mix/phoenix.ex
```diff
@@ -307,6 +310,7 @@ defmodule Mix.Phoenix do
         Existing deps:
 
             #{inspect Map.keys(deps)}
+
```
fertapric: Depends on https://github.com/phoenixframework/phoenix/pull/3004#discussion_r209455704

Gazler:
Thanks <3

PR #3007, opened 2018-08-12
Downcase error starts

fertapric:
Looking at other errors in the codebase, it seems errors should not start with capital letters.

In addition, I think error messages should not contain backticks.

As a side note, I changed three error messages and all the tests are still green 😅 Probably there is nothing wrong with the test suite, but I found it "suspicious".


chrismccord:
❤️❤️❤️🐥🔥

PR #3009, opened 2018-08-12
Update doc and errors of Phoenix.Controller

fertapric:


lib/phoenix/controller.ex
```diff
@@ -1217,7 +1218,7 @@ defmodule Phoenix.Controller do
               inspect(header) <> " with extensions: " <> inspect(exts)
             end)}
 
-      To accept custom formats, register them under the `:mime` library
+      To accept custom formats, register them under the :mime library
       in your config/config.exs file:
 
           config :mime, :types, %{
```
fertapric: There is a line below that could be changed from:

```
And then run `mix deps.clean --build mime` to force it to be recompiled.
```

to 

```
And then run: 

    mix deps.clean --build mime 

to force it to be recompiled.
```

to avoid the backticks. Thoughts?

fertapric: Similarly, there is another example:

https://github.com/phoenixframework/phoenix/blob/bacfa423d91bc09d214a551345a022dd4faea9af/lib/phoenix/controller.ex#L681-L682

that could also be converted to:

```elixir
raise """
cannot render template #{inspect template} because conn.params["_format"] is not set.

Please set:

    plug :accepts, ~w(html json ...)

in your pipeline.
"""
```

jeregrine: Lets work on those as new pr's.

PR #3010, opened 2018-08-12
Update doc of Phoenix.Digester

fertapric:


lib/phoenix/digester.ex
```diff
@@ -276,20 +274,23 @@ defmodule Phoenix.Digester do
...
 
   @doc """
-  Delete compiled/compressed asset files that are no longer in use based on
-  specified criteria.
+  Deletes compiled/compressed asset files that are no longer in use based on
+  the specified criteria.
+
+  ## Arguments
 
-    * `output_path` - The path where the compiled/compressed files will be saved
```
jeregrine: the format changes here are fine but I think output_path was fine and descriptive enough to keep.

fertapric: I thought `output_path` was a bit misleading for a "clean" operation, because that nomenclature is normally used to indicate the path where the results should be written to. 

In fact, the previous description was: "The path where the compiled/compressed files will be saved".

Other option was `assets_path`, but it seemed redundant in this context :)

PR #3015, opened 2018-08-12
Update errors of Phoenix.Controller.Pipeline

fertapric:


lib/phoenix/controller/pipeline.ex
```diff
@@ -53,15 +53,15 @@ defmodule Phoenix.Controller.Pipeline do
     cond do
       fallback == nil ->
         raise """
-        action_fallback can only be called when using Phoenix.Controller.
+        action_fallback/1 can only be called when using Phoenix.Controller.
         Add `use Phoenix.Controller` to #{inspect module}
```
fertapric: This might be changed to: 

```elixir
raise """
action_fallback/1 can only be called when using Phoenix.Controller.

Add:

    use Phoenix.Controller

to #{inspect module}
"""

to remove the backtick. Thoughts?

chrismccord:
I don't think the arity is necessary in this case. Thanks!

PR #3018, opened 2018-08-12
Update doc and errors of Phoenix.ConnTest

fertapric:


lib/phoenix/test/conn_test.ex
```diff
@@ -116,7 +116,7 @@ defmodule Phoenix.ConnTest do
   end
 
   @doc """
-  Deprecated version of conn/0. Use build_conn/0 instead
+  Deprecated version of `conn/0`. Use `build_conn/0` instead.
   """
   @spec conn() :: Conn.t
```
fertapric: I haven't found any examples in the codebase of the new `@doc` metadata. But I guess these two functions could be good candidates :)

fertapric: Or I can include `@doc deprecated: "<hint>"` in a different PR if you think that's a good idea :)

chrismccord:
❤️❤️❤️🐥🔥

PR #3021, opened 2018-08-15
Update doc link to distillery Phoenix guide.

Doerge:


Gazler:
Good spot, thanks. Maybe we should use https://hexdocs.pm/distillery/2.0.0/guides/phoenix_walkthrough.html so that the version is locked, hopefully preventing 404 issues in the future?

Doerge:
Good point. On the flipside that has the risk of getting outdated, and it's harder to spot. I.e. the link won't be dead, you will just end up on an outdated version only indicated by a version number somewhere on the page. Not sure which poison is the best?

Gazler:
Thanks <3

PR #3023, opened 2018-08-16
Enable autocomplete for generated functions in IEx

slashmili:
When a function has `@doc false`, users won't get autocomplete on those
functions and for new users this could be confusing.

more on that : https://github.com/elixir-lang/elixir/blob/master/lib/elixir/pages/Writing%20Documentation.md#hiding-internal-modules-and-functions

### DIFF
--- a/priv/templates/phx.gen.embedded/embedded_schema.ex
+++ b/priv/templates/phx.gen.embedded/embedded_schema.ex
@@ -7,7 +7,9 @@ defmodule <%= inspect schema.module %> do
 <%= for {k, v} <- schema.types do %>    field <%= inspect k %>, <%= inspect v %><%= schema.defaults[k] %>
 <% end %>  end
 
-  @doc false
+  @doc """
+  Builds a changeset based on the `struct` and `params`.
+  """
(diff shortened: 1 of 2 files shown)


priv/templates/phx.gen.embedded/embedded_schema.ex
```diff
@@ -7,7 +7,9 @@ defmodule <%= inspect schema.module %> do
 <%= for {k, v} <- schema.types do %>    field <%= inspect k %>, <%= inspect v %><%= schema.defaults[k] %>
 <% end %>  end
 
-  @doc false
+  @doc """
+  Builds a changeset based on the `struct` and `params`.
```
fertapric: I think this should be ```Builds a changeset based on the `<%= schema.singular %>` and `attrs`.```

Neither `struct` nor `params` are passed as arguments to the `changeset/2` function :)

chrismccord:
They are doc false'd to signal it is a private API for use *within* the context, but not outside of it. So I feel the signal we're trying to send outweighs the lack of autocomplete. Thanks!

PR #3024, opened 2018-08-16
Show a better error when trying use plug forward in router

### LINKED ISSUE #2995: Raise when Plug.Router.forward/2 API is used instead of Phoenix.Router.
Currently, Plug.Router has a [forward/2](https://hexdocs.pm/plug/Plug.Router.html#forward/2) macro that looks like:

    forward "/foo", to: SomePlug, init_opts: [foo: :bar]

Phoenix also has a `forward` macro, with the following API:

    forward "/foo", SomePlug, foo: :bar

A user could easily be confused (for example. when reading documentation for a library based on Plug) and try to use the Plug arguments in the Phoenix Router.

This currently fails with:
```

== Compilation error in file web/router.ex ==
** (ArgumentError) cannot convert the given list to a string.

To be converted to a string, a list must contain only:

  * strings
  * integers representing Unicode codepoints
  * or a list containing one of these three elements

Please check the given list or call inspect/1 to get the list representation, got:

[to: SomePlug, init_opts: [foo: :bar]]

    (elixir) lib/list.ex:821: List.to_string/1
    (phoenix) lib/phoenix/naming.ex:21: Phoenix.Naming.resource_name/2
    (phoenix) lib/phoenix/router/scope.ex:25: Phoenix.Router.Scope.route/8
    web/router.ex:56: (module)
    (stdlib) erl_eval.erl:677: :erl_eval.do_apply/6
    (elixir) lib/kernel/parallel_compiler.ex:198: anonymous fn/4 in Kernel.ParallelCompiler.spawn_workers/6
```

Instead, we should detect if the forward macro that a user is attempting to use the Plug style arguments and raise with the changes required.

Gazler:
Previously, passing a non-atom argument to the forward macro:

    forward "/admin", to: AdminController

Would result in the following error message;

    == Compilation error in file web/router.ex ==
    ** (ArgumentError) cannot convert the given list to a string.

    To be converted to a string, a list must contain only:

      * strings
      * integers representing Unicode codepoints
      * or a list containing one of these three elements

    Please check the given list or call inspect/1 to get the list
    representation, got:

    [to: SomePlug, init_opts: [foo: :bar]]

Now the following error message is displayed:

    forward expects a module as the second argument, [to: AdminController]
    given.

This closes #2995

### DIFF
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -776,8 +776,7 @@ defmodule Phoenix.Router do
 
     quote unquote: true, bind_quoted: [path: path, plug: plug] do
       plug = Scope.expand_alias(__MODULE__, plug)
-      path_segments = Route.forward_path_segments(path, plug, @phoenix_forwards)
-      @phoenix_forwards Map.put(@phoenix_forwards, plug, path_segments)
+      Scope.register_forwards(__MODULE__, path, plug)
       unquote(add_route(:forward, :*, path, plug, plug_opts, router_opts))
     end
   end
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -775,8 +775,7 @@ defmodule Phoenix.Router do
     router_opts = Keyword.put(router_opts, :as, nil)
 
     quote unquote: true, bind_quoted: [path: path, plug: plug] do
-      plug = Scope.expand_alias(__MODULE__, plug)
-      Scope.register_forwards(__MODULE__, path, plug)
+      plug = Scope.register_forwards(__MODULE__, path, plug)
       unquote(add_route(:forward, :*, path, plug, plug_opts, router_opts))
     end
   end
--- a/test/phoenix/router/forward_test.exs
+++ b/test/phoenix/router/forward_test.exs
@@ -142,4 +142,16 @@ defmodule Phoenix.Router.ForwardTest do
     assert conn.resp_body == "health"
     assert conn.private[ApiRouter] == {[], %{Phoenix.Test.HealthController => []}}
   end
+
+  test "forwards raises if using the plug to arguments" do
+    error_message = ~r/expects a module/
+    assert_raise(ArgumentError, error_message, fn ->
+      defmodule BrokenRouter do
+        use Phoenix.Router
+        scope "/" do
+          forward "/health", to: HealthController
+        end
+      end
+    end)
+  end
 end
--- a/lib/phoenix/router/scope.ex
+++ b/lib/phoenix/router/scope.ex
@@ -119,7 +119,7 @@ defmodule Phoenix.Router.Scope do
   @doc """
   Expands alias within scoped block.
   """
-  def expand_alias(module, alias) when is_atom(alias) do
+  def expand_alias(module, alias) do
     if inside_scope?(module) do
       module
       |> get_stack()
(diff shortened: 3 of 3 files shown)


test/phoenix/router/forward_test.exs
```diff
@@ -142,4 +142,18 @@ defmodule Phoenix.Router.ForwardTest do
     assert conn.resp_body == "health"
     assert conn.private[ApiRouter] == {[], %{Phoenix.Test.HealthController => []}}
   end
+
+  test "forwards raises if using the plug to arguments" do
+    error_message = ~r/expects a module/
+    assert_raise(RuntimeError, error_message, fn ->
+      Code.eval_string """
```
josevalim: You don't need the eval string. You can just define the module directly.

lib/phoenix/router.ex
```diff
@@ -781,4 +782,8 @@ defmodule Phoenix.Router do
       unquote(add_route(:forward, :*, path, plug, plug_opts, router_opts))
     end
   end
+
+  defmacro forward(_, args, _, _) do
+    raise "forward expects a module as the second argument, #{Macro.to_string(args)} given"
```
josevalim: Since this is an argument error, maybe it should be `raise ArgumentError`?

lib/phoenix/router.ex
```diff
@@ -771,7 +771,8 @@ defmodule Phoenix.Router do
       end
 
   """
-  defmacro forward(path, plug, plug_opts \\ [], router_opts \\ []) do
+  defmacro forward(path, plug, plug_opts \\ [], router_opts \\ [])
+  defmacro forward(path, {_, _, [mod]} = plug, plug_opts, router_opts) when is_atom(mod) do
```
josevalim: Unfortunately this means we are doing the work at compilation time. Which would break cases like this:

```elixir
mod = Mod
forward "/foo", mod
```

We want to do this at runtime. So probably in one of those two places here: https://github.com/phoenixframework/phoenix/blob/4d7198c370818ea3eca2b33b9a4559aa60178e6f/lib/phoenix/router.ex#L779-L780

lib/phoenix/router.ex
```diff
@@ -775,10 +775,16 @@ defmodule Phoenix.Router do
     router_opts = Keyword.put(router_opts, :as, nil)
 
     quote unquote: true, bind_quoted: [path: path, plug: plug] do
-      plug = Scope.expand_alias(__MODULE__, plug)
-      path_segments = Route.forward_path_segments(path, plug, @phoenix_forwards)
-      @phoenix_forwards Map.put(@phoenix_forwards, plug, path_segments)
-      unquote(add_route(:forward, :*, path, plug, plug_opts, router_opts))
+      case plug do
+        mod when is_atom(mod) ->
```
josevalim: Theoretically we want to do as little work as possible inside the quoted expression. That's because this code will be compiled for every invocation and executed over and over again. So my suggestion for now is, can we put this code inside `Scope.expand_alias`?

Alternatively, we could move the whole mangling of `@phoenix_forwards` to a private function and raise there. It would also be the most performant approach too.

victorolinasc:
Sorry to hook into this issue but there is currently a difference in the way both forwards work. I'll open an issue/pr if it makes sense, but Phoenix retains the path when forwarding and Plug does not. On Plug's documentation we have:

> Forwards requests to another Plug. The path_info of the forwarded connection will exclude the portion of the path specified in the call to forward. If the path contains any parameters, those will be available in the target Plug in conn.params and conn.path_params.

I've stumbled upon this on work and then tried to fallback to Plug's forward but couldn't make it work.

If this gets merged, I think there should be an option for Phoenix forward to work just like Plugs (even though it may not be the default option).

chrismccord:
❤️❤️❤️🐥🔥

PR #3033, opened 2018-08-21
Add a .formatter.exs to generated phx.new projects

aptinio:
and make sure the code generated by default is formatted

PR #3035, opened 2018-08-23
Add .formatter.exs for umbrella projects and update templates

mikeover:
Added the formatter file, created an umbrella project using it, ran `mix format --check-formatted`, and formatted these 2 files to make it pass.

### DIFF
--- a/installer/templates/phx_umbrella/apps/app_name/lib/app_name/application.ex
+++ b/installer/templates/phx_umbrella/apps/app_name/lib/app_name/application.ex
@@ -6,9 +6,13 @@ defmodule <%= app_module %>.Application do
   use Application
 
   def start(_type, _args) do<%= if ecto do %>
-    Supervisor.start_link([
-      <%= app_module %>.Repo,
-    ], strategy: :one_for_one, name: <%= app_module %>.Supervisor)<% else %>
+    Supervisor.start_link(
+      [
+        <%= app_module %>.Repo
+      ],
+      strategy: :one_for_one,
+      name: <%= app_module %>.Supervisor
+    )<% else %>
     Supervisor.start_link([], strategy: :one_for_one, name: <%= app_module %>.Supervisor)<% end %>
   end
 end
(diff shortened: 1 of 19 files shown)


josevalim:
Thanks @mikeover! For the umbrella, we likely want to have a generic file at the umbrella root, [like the one generated by Elixir](https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/tasks/new.ex#L261-L267), and then in the phoenix app have one similar to the one you defined here, in the ecto app another one that includes ecto and so on.

mikeover:
@josevalim Thanks for the lightning fast reviews, just pushed a new commit adding a formatter for `web` within umbrella projects and fixing a bunch of file formats within the umbrella templates. Note that I did not update any templates from `phx_web` or `phx_test` as those are being addressed in #3033. But the rest I tested created an umbrella app both with and without ecto and all of the umbrella-specific templates pass the formatting check.

installer/templates/phx_umbrella/apps/app_name/lib/app_name/application.ex
```diff
@@ -6,9 +6,13 @@ defmodule <%= app_module %>.Application do
   use Application
 
   def start(_type, _args) do<%= if ecto do %>
-    Supervisor.start_link([
-      <%= app_module %>.Repo,
-    ], strategy: :one_for_one, name: <%= app_module %>.Supervisor)<% else %>
```
josevalim: What if we do this:

```
children = [

]

Supervisor.start_link(children, strategy: :one_for_one, name: ...
```

?

In case Ecto is not available, we can do this for simplicity:

```elixir
children = [
  # MyApp.Worker
]

Supervisor.start_link(children, ....)
```

josevalim:
One last comment and we should be good to go!

mikeover:
Is da27bba what you had in mind @josevalim?

mikeover:
Thanks again @josevalim, hopefully the last round 😄 4ccd1eb

chrismccord:
❤️❤️❤️🐥🔥

PR #3037, opened 2018-08-23
chore(formatting): Ensure that code is formatted properly when genera…

tomciopp:
…ting new html and json resources

Handles fourth point in #2605. LMKWYT


priv/templates/phx.gen.embedded/embedded_schema.ex
```diff
@@ -4,7 +4,7 @@ defmodule <%= inspect schema.module %> do
   alias <%= inspect schema.module %>
 
   embedded_schema do
-<%= for {k, v} <- schema.types do %>    field <%= inspect k %>, <%= inspect v %><%= schema.defaults[k] %>
+<%= for {k, v} <- schema.types do %>    field(<%= inspect k %>, <%= inspect v %><%= schema.defaults[k] %>)
```
chrismccord: no parens on `field` as import_deps ecto will exclude it

tomciopp:
I have not verified that `mix phx.gen.context` was formatted. I will make the suggested change and ensure that context is also formatted.

tomciopp:
Everything should be good now.

tomciopp:
@chrismccord Ping!

PR #3044, opened 2018-09-03
Fixes wrong assoc's index name if not specify `id` with the CLI.

linjunpop:
Running with the CLI:

```bash
$ mix phx.gen.html Sales Agent agents token invited_by:references:agents
```

Will generate:

```elixir
defmodule YDL.Repo.Migrations.CreateAgents do
  use Ecto.Migration

  def change do
    create table(:agents, primary_key: false) do
      add(:id, :binary_id, primary_key: true)
      add(:token, :string)
      add(:invited_by_id, references(:agents, on_delete: :nothing, type: :binary_id))

      timestamps()
    end

    create(index(:agents, [:invited_by]))
  end
end
```

Which the index should generate with a `_id` suffix.

```elixir
defmodule YDL.Repo.Migrations.CreateAgents do
  use Ecto.Migration

  def change do
    create table(:agents, primary_key: false) do
      add :id, :binary_id, primary_key: true
      add :token, :string
      add :invited_by_id, references(:agents, on_delete: :nothing, type: :binary_id)

      timestamps()
    end

    create index(:agents, [:invited_by_id])
  end
end
```

### DIFF
--- a/lib/mix/phoenix/schema.ex
+++ b/lib/mix/phoenix/schema.ex
@@ -289,7 +289,12 @@ defmodule Mix.Phoenix.Schema do
 
   defp indexes(table, assocs, uniques) do
     uniques = Enum.map(uniques, fn key -> {key, true} end)
-    assocs = Enum.map(assocs, fn {_, key, _, _} -> {key, false} end)
+    assocs =
+      assocs
+      |> Enum.map(fn {_, key, _, _} ->
+        key = if(String.ends_with?(inspect(key), "_id"), do: key, else: "#{key}_id")
+        {key, false}
+      end)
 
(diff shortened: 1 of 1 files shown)


chrismccord:
The `references` switch requires the explicit fkey name – we don't append the _id for such cases. If you search the tracker you can find previous discussions on rationale. Thanks!

linjunpop:
@chrismccord Do you mean the references should be explicated written as `invited_by_id:references:agents`, and `invited_by:references:agents` should not be acceptable?

linjunpop:
Oh, for the `invited_by:references:agents`, I mean it should not be append with the `_id` to the field.

PR #3045, opened 2018-09-03
Fixes the `references` switch should requires an explict fkey name.

linjunpop:
Running with the CLI:

```bash
$ mix phx.gen.html Sales Agent agents invited_by:references:agents
```

Will generate:

```elixir
 add(:invited_by_id, references(:agents, on_delete: :nothing, type: :binary_id))
```

But the expected result should be the following as the discussion here: https://github.com/phoenixframework/phoenix/pull/3044#issuecomment-418120799

```elixir
add :invited_by, references(:agents, on_delete: :nothing, type: :binary_id)
```


chrismccord:
❤️❤️❤️🐥🔥

PR #3047, opened 2018-09-04
Update urls to be links for reader convenience

choncou:
This is an update to make the final urls listed in the tutorial be actual links the reader can click through on.

Assuming that the reader was actually following the tutorial, this adds a tiny bit of convenience.


chrismccord:
❤️❤️❤️🐥🔥

PR #3049, opened 2018-09-04
update guides

### LINKED ISSUE #2589: Testing guides overhaul
After following the testing guides, it seems like they could be improved

- [ ] each guide uses its own (sometimes implicitly, due to conflicts with previously generated code) `phx.new`, but maybe this isn't necessary.
- [ ] `testing.md`
  - [ ] "Generating More Files" looks like it's supposed to be used in `testing_schemas` or something, but uses a different context name
- [ ] `testing_schemas.md`
  - [ ] "Generating an HTML Resource" mentions manually doing ecto tasks, which are handled automatically
- [ ] `testing_controllers.md`
  - [ ] "Show when the user is not found"  doesn't mention or use `action_fallback`
  - [ ] is it actually useful to say "there's this generator that you could use that does this for you, but don't use that"? maybe this could be rewritten to build functionality on top of generated code, like previous guides?
  - [ ] it seems slightly forced to start by testing a list response two items, then deleting most of the code with more efficient code later; maybe this could be restructured to introduce test setup macros first, then followed by actual tests?
    - maybe the test macro stuff could be put in the `introduction`.

#2493 makes it sound like there might have been some other plan that builds off of `context`

mitchellhenke:
maybe closes #2589

it isn't an overhaul by any means, but corrects a few things mentioned in the issue and some other inconsistencies I found when running through the guide

### DIFF
--- a/guides/phoenix_mix_tasks.md
+++ b/guides/phoenix_mix_tasks.md
@@ -139,9 +139,8 @@ We can also see that files related to the application as a whole - eg. files in 
 
 ```console
 * creating task_tester/lib/hello.ex
-* creating task_tester/lib/hello/endpoint.ex
+* creating task_tester/lib/hello_web/endpoint.ex
 * creating task_tester/lib/hello/repo.ex
-* creating task_tester/test/hello_test.exs
 ```
 
 If we only want to change the qualifying prefix for module names, we can do that with the `--module` flag. It's important to note that the value of the `--module` must look like a valid module name with proper capitalization. The task will throw an error if it doesn't.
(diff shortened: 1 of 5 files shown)


chrismccord:
@mitchellhenke this looks good to go. Can you resolve the issues and re-push? Thanks! <3

chrismccord:
@mitchellhenke nevermind, I took care of it. Thank you! ❤️❤️❤️🐥🔥

mitchellhenke:
@chrismccord thanks for taking care of it! :)

PR #3055, opened 2018-09-12
Mark phoenix.js npm package as side effect free

lukaszsamson:
Rationale:
When importing anything from phoenix.js all the module ends up bundled. Marking the package as side effect free enables build optimizers to do tree shaking and scope hoisting and only the used code lands in bundle.
https://github.com/webpack/webpack/tree/master/examples/side-effects


chrismccord:
❤️❤️❤️🐥🔥

Issue #3052, opened 2018-09-05
Update guides/errors.md


mitchellhenke:
https://github.com/phoenixframework/phoenix/blob/master/guides/errors.md describes the generated error view that used to look like:

```elixir
defmodule Hello.ErrorView do
  use Hello.Web, :view

  def render("404.html", _assigns) do
    "Page not found"
  end

  def render("500.html", _assigns) do
    "Server internal error"
  end

  # In case no render clause matches or no
  # template is found, let's render it as 500
  def template_not_found(_template, assigns) do
    render("500.html", assigns)
  end
end
```

but it should be updated to reflect the new generated error view that looks like:

```elixir
defmodule HelloWeb.ErrorView do
  use HelloWeb, :view

  # If you want to customize a particular status code
  # for a certain format, you may uncomment below.
  # def render("500.html", _assigns) do
  #   "Internal Server Error"
  # end

  # By default, Phoenix returns the status message from
  # the template name. For example, "404.html" becomes
  # "Not Found".
  def template_not_found(template, _assigns) do
    Phoenix.Controller.status_message_from_template(template)
  end
end
```

PR #3061, opened 2018-09-17
Guides: Change to default error view template, use brunch in endpoint code

hlhr:
Fix Issue #3052 

* Remove sentence which doesn't make sense with the new error view
* Change Template to the new template in phoenix 1.3
* Change dev.exs code to brunch (which is the default as far as i know)

### DIFF
--- a/guides/errors.md
+++ b/guides/errors.md
@@ -2,7 +2,7 @@
 
 Phoenix provides an `ErrorView`, `lib/hello_web/views/error_view.ex`, to render errors in our applications. The full module name will include the name of our application, as in `Hello.ErrorView`.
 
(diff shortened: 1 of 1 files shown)


guides/errors.md
```diff
@@ -36,7 +35,7 @@ config :hello, HelloWeb.Endpoint,
   debug_errors: false,
   code_reloader: true,
   cache_static_lookup: false,
-  watchers: [node: ["node_modules/webpack/bin/webpack.js", "--mode", "development", "--watch-stdin",
+  watchers: [node: ["node_modules/brunch/bin/brunch", "watch", "--stdin",
                     cd: Path.expand("../assets", __DIR__)]]
```
chrismccord: Keep the webpack config, as 1.4 uses webpack :)

PR #3062, opened 2018-09-17
Only use css-loader syntax when generate with Webpack

50kudos:
This fixes "no route found for GET /css/phoenix.css" from phoenix
installed with --no-webpack option. Use `:append` from the generator
instead of css-loader for static template generation.

This is the improvement of https://github.com/phoenixframework/phoenix/pull/2779 so it could go with version 1.4.


50kudos:
Sounds good. Let me rebase master and change it.

chrismccord:
Too late. Already took care of it :)

Thank you! ❤️❤️❤️🐥🔥

50kudos:
Thank you. I was just on the way solving conflict on my machine. You are just faster :)

PR #3065, opened 2018-09-17
phoenix.js memory + CPU improvements

gfodor:
We are experimenting with using Phoenix Channels for game networking. Our application is in WebVR so is sensitive to GC pauses, so this PR applies some changes to phoenix.js to reduce memory allocations and save some CPU time, particularly on the receive path.

Summary of changes:

- Avoid string template interpolation for log messages if logger has not been set on socket

- Skip unnecessary walking of the `CHANNEL_LIFECYCLE_EVENTS` array when checking channel membership on receive

- Unroll functional array operations into `for` loops to remove memory allocations for closures, copying, etc.

- Skip additional memory allocations + destructuring in `push` if logging is disabled, and eliminate the need to create an additional closure (with bindings) in `push` when connected.

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -382,17 +382,17 @@ export class Channel {
     })
     this.onClose( () => {
       this.rejoinTimer.reset()
-      this.socket.log("channel", `close ${this.topic} ${this.joinRef()}`)
+      if (this.socket.hasLogger()) this.socket.log("channel", `close ${this.topic} ${this.joinRef()}`)
       this.state = CHANNEL_STATES.closed
       this.socket.remove(this)
     })
     this.onError( reason => { if(this.isLeaving() || this.isClosed()){ return }
-      this.socket.log("channel", `error ${this.topic}`, reason)
+      if (this.socket.hasLogger()) this.socket.log("channel", `error ${this.topic}`, reason)
       this.state = CHANNEL_STATES.errored
       this.rejoinTimer.scheduleTimeout()
     })
     this.joinPush.receive("timeout", () => { if(!this.isJoining()){ return }
-      this.socket.log("channel", `timeout ${this.topic} (${this.joinRef()})`, this.joinPush.timeout)
+      if (this.socket.hasLogger()) this.socket.log("channel", `timeout ${this.topic} (${this.joinRef()})`, this.joinPush.timeout)
       let leavePush = new Push(this, CHANNEL_EVENTS.leave, closure({}), this.timeout)
       leavePush.send()
       this.state = CHANNEL_STATES.errored
@@ -523,7 +523,7 @@ export class Channel {
   leave(timeout = this.timeout){
     this.state = CHANNEL_STATES.leaving
     let onClose = () => {
-      this.socket.log("channel", `leave ${this.topic}`)
+      if (this.socket.hasLogger()) this.socket.log("channel", `leave ${this.topic}`)
       this.trigger(CHANNEL_EVENTS.close, "leave")
     }
     let leavePush = new Push(this, CHANNEL_EVENTS.leave, closure({}), timeout)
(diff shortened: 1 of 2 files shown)


assets/js/phoenix.js
```diff
@@ -911,19 +923,16 @@ export class Socket {
   /**
    * @param {Object} data
    */
-  push(data){
-    let {topic, event, payload, ref, join_ref} = data
-    let callback = () => {
-      this.encode(data, result => {
-        this.conn.send(result)
-      })
+  push(data, reliable = true){
```
mquander: I think the `reliable` parameter is left from your earlier version and isn't used anymore.

chrismccord:
This is great! Do you have numbers pre and post-optimizations that you can share for your app? The logging changes would appear to help, but I'd love to see a breakdown of profiling:

- logging changes only
- filter/map + map/forEach changes only

The map/filter/forEach becomes more obfuscated, so I would only do it if we're yielding measurable improvements. If you can provide some insight it would be super appreciated. Thank you!

gfodor:
Sure thing! Will get that together.

gfodor:
Well this is embarassing!

These changes were a slice of another branch of phoenix.js I have that introduced a change on the send path, and all of the memory gains seemed to be coming from that change -- I never broke these out and measured them separately. So thanks for keeping me honest! :)

I tested in Chrome -- Firefox doesn't seem to have an allocations profiler, just heap snapshotting -- but Chrome certainly seems to be very clever with avoiding memory allocations around string templates and the other things I did here to avoid closure allocations. As best I can tell, the patch here has minimal actual effect, so sorry for the distraction.

As far as the other change goes that I mentioned, now that I've measured things separately, the big allocator is on the send side, and is all the machinery to deal with reliable delivery and timeouts. If you skip all that, sending 1000 empty messages goes from allocation 500-600kb in Chrome on my machine to 5-10kb. You can fast-path it yourself by calling `socket.push` directly (which is a 'fire-and-forget' API basically.) It may be worth adding a new `pushUnreliable` API that skips these allocations -- for now we can just do it in our application code.

To actually eliminate the memory churn and maintain the reliability guarantees I think the timeout reply code would need to be special cased, and the Push objects would need to be memory pooled.

I'll close this PR and I'll open another if we loop around and fix up those allocations. Also, if you think adding a fire-and-forget API is useful, I can cut that one too.

gfodor:
OK, now I am not so sure. I believe the Chrome allocations profiler is discarding objects at the end of the profile which are not rooted (and immediately discards them if it can.) This explains why I see the timeout allocations but not the string concatenations etc. I will try to find a way to actually measure the effect of these changes properly.

gfodor:
OK! I found a way to measure the actual allocations, not the pinned memory, I think, by taking a performance timeline in Chrome of the JS heap size and sending 100 messages right after a GC and looking for the memory spike. I guess very few people care about measuring short lived JS garbage! Here is the data (3 trials each, on the receiver side)

```
100 messages -- base line = 38.113 - 37.933 = 180kb, 38.034 - 37.852 = 180kb, 38.017 - 37.838 = 179kb
100 messages -- no logging optimization = 38.007 - 37.914 = 93kb, 37.952 - 37.860 = 92kb, 37.966 - 37.874 = 92kb
100 messages -- full optimization 37.944 - 37.862 = 82kb, 37.955 - 37.879 = 76kb, 37.964 - 37.893 = 71kb
```

So both sets of changes look like significant wins, in fact the non-logging change looks like it has a more profound effect. Also, the point mentioned above re: the savings to be still had on the send side are valid too.

chrismccord:
❤️❤️❤️🐥🔥

I would prefer to hold off on major api additions for the send side since we don't have any data on it being a problem in practice, and the alternative would be a much poorer API. Thanks!!!

PR #3066, opened 2018-09-22
Default Layout View

virayatta:
I'm not sure why it's must called `LayoutView`. But I think it's just cool, if we can change the default layout view.


chrismccord:
The default is `LayoutView`, but you can set the layout today to whatever you want with a plug:

    plug :put_layout, {WhateverView, "whatever.html"}

No opts required :)

PR #3068, opened 2018-09-27
Raise if the pubsub server name is not configured at compile time

### LINKED ISSUE #2992: PubSub config required at compile time
If an application has the following in `config.exs`:

```elixir
# Configures the endpoint
config :config_test, ConfigTestWeb.Endpoint,
  render_errors: [view: ConfigTestWeb.ErrorView, accepts: ~w(html json)],
  pubsub: [name: ConfigTest.PubSub, adapter: Phoenix.PubSub.PG2, pool_size: 1]
```

Then the supervision tree looks like this (correctly using the PubSub name):

config-exs

However, if the same config us in the `MyApp.Endoint.init/2` function:

```elixir
  def init(_, config) do
    {:ok, Keyword.merge(config, default_config())}
  end

  defp default_config() do
    [
      pubsub: [name: ConfigTest.PubSub, adapter: Phoenix.PubSub.PG2, pool_size: 1]
    ]
  end
```

 then it looks like:

config-init

In this case, a generic name of `Elixir.Supervisor` is used (this is as a result of `Module.concat(nil, "Supervisor")` from https://github.com/phoenixframework/phoenix_pubsub/blob/daedc139726f38b6de2a84ca8d9d1eeb4f5b6307/lib/phoenix/pubsub/pg2.ex#L50)

Gazler:
If the pubsub server is configured at runtime (e.g. in the init callback) then
everything will appear to work as normal, but the pubsub server will have the
name of `Elixir.Supervisor` due to having nil as a name.

This closes #2992

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -507,7 +507,8 @@ defmodule Phoenix.Endpoint do
       server
     else
       raise ArgumentError, """
-      no :pubsub server configured, please setup :pubsub in your config.
+      no :pubsub server was configured at compile time, please setup :pubsub
+      in your config.
 
       By default this looks like:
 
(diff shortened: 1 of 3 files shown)


chrismccord:
❤️❤️❤️🐥🔥

PR #3069, opened 2018-09-28
Log and instrument UserSocket connect

### LINKED ISSUE #3046: Log and instrument UserSocket.connect

chrismccord:
Closes #3046

### DIFF
--- a/lib/phoenix/socket.ex
+++ b/lib/phoenix/socket.ex
@@ -68,6 +68,14 @@ defmodule Phoenix.Socket do
     * `:transport_pid` - The pid of the socket's transport process
     * `:serializer` - The serializer for socket messages
 
+  ## Logging
+
+  Logging for socket connections is set via the `:log` option, for example:
+
+      use Phoenix.Socket, log: :debug
+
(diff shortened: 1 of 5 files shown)


lib/phoenix/socket.ex
```diff
@@ -520,6 +546,10 @@ defmodule Phoenix.Socket do
     :ok
   end
 
+  @doc false
+  def __log__(Phoenix.LiveReloader.Socket, _opts), do: false
```
josevalim: We shouldn't hardcode this here. Instead we should set log: false in the live reloader lib and ship a new version.

chrismccord: Agree, which we will do, but folks who don't `mix deps.update phoenix_live_reload` will get spammy logs. It's not ideal to match on this here, but I think we do something similar elsewhere for ergonomics. Thoughts?

josevalim: Then they can update their `phoenix_live_reload`. :)

chrismccord: Fair :D

PR #3071, opened 2018-09-28
Allow runtime configuration of PubSub

Gazler:
Previously PubSub was required at compile time. This commit allows configuration
at runtime by looking up the PubSub module for the endpoint from the config
instead of compiling it into the module. The compile time errors have been move
to runtime and will trigger when attempting to start the pubsub children.


chrismccord:
❤️❤️❤️🐥🔥

PR #3074, opened 2018-09-30
Presence guide's example in socket.js

juanpabloaj:
the autogenerated code socket.js is most similar to guide's exampleif we compared with app.js


chrismccord:
Thanks! The goal in this case is to show all the code necessary, without diving into the socket.js, so I think the current guide is preferable. Thank you!

PR #3075, opened 2018-10-01
Fix incorrect assert in websocket channel test

edennis:
Stumbled upon this when reading some code. This `assert` obviously doesn't assert much. 😄


chrismccord:
bah, good catch! ❤️❤️❤️🐥🔥

PR #3078, opened 2018-10-03
Fix warning about cowboy_websocket behaviour

Gazler:
When using Cowboy 1.x, the `cowboy_websocket` module is available, but
not a behaviour. This commit ensures that the module defines a
behaviour.

 > warning: module :cowboy_websocket is not a behaviour
 > (in module Phoenix.Endpoint.Cowboy2Handler)
 >   lib/phoenix/endpoint/cowboy2_handler.ex:1


chrismccord:
❤️❤️❤️🐥🔥

PR #3080, opened 2018-10-04
Fix fetching the wrong version of cowboy when deps update

Gazler:
Previously, if the version of a dependency which depends on cowboy (such
as plug) is updated beyond the version specified in the lock file, it
will also update cowboy. This will cause the tests to run against Cowboy
2 even when COWBOY_VERSION is set to 1.0.

This commit raises on the test helper when the version specified in
COWBOY_VERSION is "1.x" and the version of Cowboy in version 2.


chrismccord:
❤️❤️❤️🐥🔥

PR #3082, opened 2018-10-09
Disallows many variable mutations in phoenixjs

jekku:
## What does this PR do?

- It changes a lot of `let` references to `const` in the examples and in the code itself

## What is the rationale behind this?

- It will be much easier to track bugs inside the code because it encourages or disallows reference changes to variables

## What benefits can we get from this?

- Function variables are guaranteed to be more isolated within their own scopes.

## What can be better?

- Loops can be re-referenced to use functional stuff such as `filter` and `map`. But that's for another PR

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -218,7 +218,7 @@ const TRANSPORTS = {
 
 // wraps value in closure or returns closure
 const closure = (value) => {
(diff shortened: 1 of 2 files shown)


chrismccord:
Thanks! Please see other PRs around this issue. The tldr is I prefer the current approach, but I appreciate your willingness to help make Phoenix better. Thank you!

Issue #3083, opened 2018-10-09
Outdated pendingHeartbeatRef causes unnecessary reconnect


mpoeter:
### Environment

* Phoenix version: 1.3.4

Steps to reproduce:
* Load a webpage that creates a Phoenix Socket connection (define a logger function to see what is happening)
* Put the machine to sleep and wait until the server closes the connection (default 1 minute)
* Wake up the machine
* The connection is immediately reestablished, but 30s later the connection is closed and another one is opened.

The reason for this is that when the machine is woken up the connection is gone. But before the `onClose` event is fired, a new heartbeat is pushed on the (dead) connection, thereby setting `pendingHeartbeatRef`. However, the `pendingHeartbeatRef` is not cleared when the connection is reestablished. Obviously we never get an answer for the heartbeat on the already closed connection, so the next time the heartbeat interval is triggered, we see that `pendingHeartbeatRef` is still set. As a result, the connection is closed and reestablished once more.

The fix should be quite simple: set `pendingHeartbeatRef` to `null` in `onConnOpen`.

hlhr:
I can take a look at this but your solution sounds about right. Have to do some more testing and send in a PR later.

chrismccord:
Fixed by #3084

PR #3084, opened 2018-10-09
Reset pendingHeartbeatRef on connection open

hlhr:
Fix for Issue #3083 

Reset the reference to pending heartbeat to prevent unnecessary reconnect.

It's possible to reproduce the behaviour (see #3083 for details) with two separate machines or a docker container as the server.


hlhr:
I can't reproduce this error on my machine, tried it with the version elixir 1.6.6 but all tests run successful. Any suggestions?

chrismccord:
❤️❤️❤️🐥🔥

Issue #3085, opened 2018-10-09
Ensure phx.gen.cert is umbrella-aware


chrismccord:
https://elixirforum.com/t/phoenix-1-4-0-rc-0-released/17182/11?u=chrismccord

voltone:
Actually, I guess the Mix task needs to know which child app to generate a certificate for, since the umbrella doesn't have its own `priv` directory. It does mean that if you have two Phoenix child apps, e.g. one web UI and one API, they are each going to need their own cert. For dev mode that's probably acceptable, and for production the certificate is probably going to be injected through some other means anyway.

voltone:
In the above PR I fail the task at the umbrella level, like other `gen` tasks.

I still don't know how to ensure that the message displayed to the user references the correct Endpoint module. I wrote a function that finds all modules implementing the Endpoint behaviour, but that requires a `Mix.Task.run("compile", [])`, and besides it seems a bit... unconventional.

chrismccord:
Thanks! I took care of the error message by exposing a private function we had on the Phoenix generator side. Thank you!!

PR #3089, opened 2018-10-10
 Block phx.gen.cert task at umbrella level 

voltone:
Addressing #3085

One remaining issue: the printed instructions assume the Endpoint lives under the `#{base}Web` namespace, which may be typically correct for standalone apps, but not for child apps under an umbrella. I looked at some other Mix tasks, but I didn't see an easy way to determine the correct namespace.

### DIFF
--- a/lib/mix/tasks/phx.gen.cert.ex
+++ b/lib/mix/tasks/phx.gen.cert.ex
@@ -49,7 +49,7 @@ defmodule Mix.Tasks.Phx.Gen.Cert do
   @doc false
   def run(all_args) do
     if Mix.Project.umbrella?() do
-      Mix.raise("mix phx.gen.channel can only be run inside an application directory")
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/phx.gen.cert.ex
```diff
@@ -48,6 +48,10 @@ defmodule Mix.Tasks.Phx.Gen.Cert do
 
   @doc false
   def run(all_args) do
+    if Mix.Project.umbrella?() do
+      Mix.raise("mix phx.gen.channel can only be run inside an application directory")
```
Gazler: s/channel/cert?

voltone: oops...

chrismccord:
❤️❤️❤️🐥🔥

PR #3091, opened 2018-10-11
format dotfiles

beardedeagle:
- Capture dotfiles for formatting in phx_single


josevalim:
Please update your Elixir version. Earlier `mix format` versions indeed ignored dot files.

PR #3095, opened 2018-10-13
Remove hardcoded 1.3.0 reference in mix task docs

gshaw:
Phoenix Specific Mix Tasks guide has a reference to generating a 1.3.0 application with `phx.new` task.

This change removes the specific version so the docs stay accurate.

screen shot 2018-10-13 at 11 47 21

https://hexdocs.pm/phoenix/1.4.0-rc.1/phoenix_mix_tasks.html#phoenix-specific-mix-tasks

### DIFF
--- a/guides/phoenix_mix_tasks.md
+++ b/guides/phoenix_mix_tasks.md
@@ -7,11 +7,12 @@ There are currently a number of built-in Phoenix-specific and ecto-specific mix 
 ## Phoenix Specific Mix Tasks
 
 ```console
-
 ➜ mix help | grep -i phx
 mix local.phx          # Updates the Phoenix project generator locally
+mix phx                # Prints Phoenix help information
 mix phx.digest         # Digests and compresses static files
 mix phx.digest.clean   # Removes old versions of static assets.
+mix phx.gen.cert       # Generates a self-signed certificate for HTTPS testing
 mix phx.gen.channel    # Generates a Phoenix channel
 mix phx.gen.context    # Generates a context with functions around an Ecto schema
 mix phx.gen.embedded   # Generates an embedded Ecto schema file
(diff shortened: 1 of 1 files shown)


Gazler:
Given that the output is generated from the command `mix help | grep -i phx`, technically the output of https://github.com/phoenixframework/phoenix/blob/2a10f66e14b130dcac23a91c1906c64b32b208af/installer/lib/mix/tasks/phx.new.ex#L90 should be used (which would be `v1.4.0-rc.0` at the time of writing.)

However I think a better solution is to remove the `mix help | grep -i phx` from the start of the section so we don't have to keep changing the version number.

chrismccord:
Agree with @Gazler . @gshaw Can you remove the line? Thanks!

chrismccord:
Ping @gshaw :)

gshaw:
@chrismccord, sorry, I missed the GitHub notification but still confused about what should be removed.  We could change `phx.new.ex` to not emit a version number in the help.  I.e.,

```elixir
@shortdoc "Creates a new Phoenix v#{@version} application"
```
to
```elixir
@shortdoc "Creates a new Phoenix application"
```

or remove the line 11 `➜ mix help | grep -I phx` in `phoenix_mix_tasks.md` to not show it.

Let me know which change you want and I'll push it (without waiting 10 days).

Gazler:
I prefer having the version in the mix task as i quite often change between versions of the archive, so it is an easy way to see it, so I'd vote for removing line 11.

This is most relevant when people are trying out release candidates so they can quickly see the version of the archive they have installed.

gshaw:
Initially I thought the build was red because the output of the mix command was used in a doctest.  That doesn't appear to be the case.

I don't think it's a problem that the output shown in the text doesn't show the version number. There is value in showing the line that generated the output, if for no other reason next major version update can run the command again to get the next tasks or changed descriptions. E.g., 1.4 added two new tasks.

If people feel the line 11 (the line that shows how the output was generated should be removed I'll push that).

Gazler:
Thanks <3

Good spot on the new tasks.

PR #3101, opened 2018-10-15
Router Update: Allow a map for a path with multiple identifiers

ericsullivan:
I created a non-standard route: 
```
get "/month_projection_records/:year/:month/roles/:role_id", MonthRoleProjectionRecordController, :show, as: :month_role_projection_record
```

It's basically a page from a report (that is not stored in a database), so the multiple identifiers act as a composite primary key. `mix phx.routes` said it was defined:

```elixir
month_role_projection_record_path  GET     /month_projection_records/:year/:month/roles/:role_id  MonthRoleProjectionRecordController :show
```

But my initial usage failed:
```
<%= link "Show", to: month_role_projection_record_path(@conn, :show, month_role_projection_record) %>
```

with the exception:
```
(CompileError)  undefined function month_role_projection_record_path/3
```

I thought it might have been related to the Phoenix.Param protocol, but a simple map also failed:
```
<%= link "Show", to: month_role_projection_record_path(@conn, :show, %{year: 1, month: 1, role_id: 1}) %>
```

A colleague pointed out that the defined paths had a different arity, and this worked:
```
<%= link "Show", to: month_role_projection_record_path(@conn, :show, month_role_projection_record.year, month_role_projection_record.month, month_role_projection_record.role_id) %>
```

That seemed counter-intuitive, and I didn't like that I'd have to refer to the route to figure out the argument order.

I modified the router to create a function that pattern matches on a map with all the required identifiers. I wasn't happy with my use of `unquote_splicing`, but my metaprogramming skills are poor. In fact, this was basically a repackaging of a [stack overflow solution](https://stackoverflow.com/questions/48413292/how-to-dynamically-generate-a-phoenix-controller-function-with-a-map-argument) by José Valim. 

I'm happy to add a few more tests, but I wanted to get feedback in case this isn't a worthwhile feature. Thank You

### DIFF
--- a/lib/phoenix/router/helpers.ex
+++ b/lib/phoenix/router/helpers.ex
@@ -215,8 +215,25 @@ defmodule Phoenix.Router.Helpers do
     {bins, vars} = :lists.unzip(exprs.binding)
     segs = expand_segments(exprs.path)
 
-    # We are using @anno to avoid warnings in case a path has already been defined.
-    quote @anno do
+    q1 = if Enum.count(bins) > 0 do
+      match = Enum.into(bins, [], fn(k) -> {String.to_atom(k), Macro.var(:"#{k}", nil)} end)
+      pattern = {:%{}, [], match}
+
+      quote do
+        def unquote(:"#{helper}_path")(conn_or_endpoint, unquote(opts), vars) when is_map(vars) do
+          unquote(:"#{helper}_path")(conn_or_endpoint, unquote(opts), vars, [])
+        end
+
+        def unquote(:"#{helper}_path")(conn_or_endpoint, unquote(opts), unquote(pattern), params)
+            when is_list(params) or is_map(params) do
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @ericsullivan!

Thanks for your PR! Although I like the idea, I have to say I am a :-1: on this feature for two reasons: today the number of functions we generate for each route is already confusing in terms of which arity to call. Also, for large routers, compilation time is high (we have recently worked on bringing it down to 5s, and such a change will increase the amount of work for every route defined.

In hindsight, maybe it would have been better to always use a map/keyword list. But for now, I think the best is for users to explicitly define those map-based helpers in the cases they need it.

chrismccord:
I agree with José, since we allow the query params enum as the last argument, having a map precede it would be very confusing. If I remember correctly, this was a reason we did not go that route initially for the named params. Your own function is the way to go. Thanks for the PR!

ericsullivan:
Those points all make sense. Thank you both for the code review.

PR #3102, opened 2018-10-16
Use FunctionClauseError for Phoenix.ActionClauseError

### LINKED ISSUE #2657: Phoenix.ActionClauseError vs new FunctionClauseError
When no action clause matches the request parameters phoenix will generate a `Phoenix.ActionClauseError`. This was fine in Elixir < 1.5, where the generic `FunctionClauseError` wasn't very readable and the extra information Phoenix provided in the error message were quite useful.

The situation changed with the Elixir 1.5 release and the `Exception.blame` mechanism that is wonderful - especially the part where it lists the function clauses and analyses them to say why they didn't match the argument.

I know that the `Phoenix.ActionClauseError` also serves another role - it allows serving a 400 response instead of the 500 one - which is correct since the endpoint received unexpected data.

It would be wonderful, if `Phoenix.ActionClauseError` featured a similar message to what is given by `Exception.blame` for `FunctionClauseError`s. I think it would make lives of developers much easier, I'm not sure, though, how feasible that is.

Gazler:
Using `FunctionClauseError` results in showing the available clauses
when displaying the error message. This also renders nicely in the plug
exceptions page.

Instead of using `plug_status` in the exception, the `Plug.Exception`
protocol is explicitly defined for the `ActionClauseError` to allow easy
transitioning between `ActionClauseError` and `FunctionClauseError`.`

This closes #2657

### DIFF
--- a/lib/phoenix/controller/pipeline.ex
+++ b/lib/phoenix/controller/pipeline.ex
@@ -133,8 +133,8 @@ defmodule Phoenix.Controller.Pipeline do
 
   @doc false
   def __catch__(%Plug.Conn{}, :function_clause, controller, action,
-                [{controller, action, [%Plug.Conn{} = conn | _], _loc} | _] = stack) do
-    args = [controller: controller, action: action, params: conn.params]
+      [{controller, action, [%Plug.Conn{} | _] = action_args, _loc} | _] = stack) do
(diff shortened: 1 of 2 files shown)


chrismccord:
😍😍😍🐥🔥

Issue #3103, opened 2018-10-17
Post-install instructions assume project is created in the current directory


bariscicek:
### Environment

* Elixir version (elixir -v): Elixir 1.4.2
* Phoenix version (mix deps):1.3.4 (phoenix) aaa1b55e
* NodeJS version (node -v):v8.11.1
* NPM version (npm -v):5.6.0
* Operating system: Linux 4.10.13-200.fc25.x86_64

### Expected behavior
This is pretty trivial thing though wanted to report as a small improvement. When mix is used with phx.new with a relative directory, the instructions comes omiting the parent diretories. 
Command like:
`mix phx.new Codes/hello`
Tells in instructions:
```
We are all set! Go into your application by running:

    $ cd hello
```
Although trivial, this is misleading. 
### Actual behavior
Post-install instructions should consider parent directories as well. (ie.:
```
We are all set! Go into your application by running:

    $ cd ./Codes/hello
```
).

josevalim:
Closing in favor of #3131.

PR #3104, opened 2018-10-17
Remember the cwd path to correctly report relative_path post install

theodowling:
This will fix issue https://github.com/phoenixframework/phoenix/issues/3103

It seems like `relative_app_path/1` wasn't taking into account the original working path - as it has already changed directory to the `project.project_path` at that stage.

Feel free to ignore if this approach is not ideal.

### DIFF
--- a/installer/lib/mix/tasks/phx.new.ex
+++ b/installer/lib/mix/tasks/phx.new.ex
@@ -136,6 +136,8 @@ defmodule Mix.Tasks.Phx.New do
   defp prompt_to_install_deps(%Project{} = project, generator) do
     install? = Mix.shell.yes?("\nFetch and install dependencies?")
 
+    {:ok, cwd_path} = :file.get_cwd()
+    
     maybe_cd(project.project_path, fn ->
       mix_pending =
         install_mix(install?)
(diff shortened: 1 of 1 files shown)


dalhorinek:
Isn't actually the problem with the 
```
Path.basename(path)
```

that strips the part of the path?

The `:file.get_cwd()` path is same as `project.project_path`, so then it matches the 

```
case Path.relative_to(path, cwd_path) do
  ^path -> Path.basename(path)
  ...
```

```
iex(1)> cwd_path = "/Work/Elixir/Code/hello"
"/Work/Elixir"
iex(2)> path = "/Work/Elixir/Code/hello"
"./Code/hello"
iex(3)> Path.relative_to(path, cwd_path)
"./Code/hello"
iex(4)> Path.basename(Path.relative_to(path, cwd_path))
"hello"
```

So I think the main problem is the basename and there is probably some reason for it, but I can't find out what it could be.

josevalim:
Closing in favor of #3131.

josevalim:
Thanks for the PR in any case!  :heart: :green_heart: :blue_heart: :yellow_heart: :purple_heart:

PR #3106, opened 2018-10-19
Escape Plug opts in routes

fertapric:
The Plug options passed when building the route are not escaped while building the `:dispatch` expression in `Phoenix.Router.Route.exprs/1`.

That produces `CompileError`s when building and unquoting the matches in `Phoenix.Router.__before_compile__/1`.

This is the [offending commit](https://github.com/phoenixframework/phoenix/commit/29a1f01e6b0134534cf617efbfc5d54b54c9cd55#diff-4b61711a8ead48b802b5b21df56705a4L99).

This bug affects v1.4.0-rc.1.

### DIFF
--- a/lib/phoenix/router/route.ex
+++ b/lib/phoenix/router/route.ex
@@ -121,13 +121,16 @@ defmodule Phoenix.Router.Route do
     {_params, fwd_segments} = build_path_match(route.path)
 
     quote do
-      {Phoenix.Router.Route, {unquote(fwd_segments), unquote(route.plug), unquote(route.opts)}}
+      {
+        Phoenix.Router.Route,
+        {unquote(fwd_segments), unquote(route.plug), unquote(Macro.escape(route.opts))}
+      }
     end
   end
 
   defp build_dispatch(%Route{} = route) do
     quote do
-      {unquote(route.plug), unquote(route.opts)}
+      {unquote(route.plug), unquote(Macro.escape(route.opts))}
     end
   end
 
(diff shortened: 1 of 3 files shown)


lib/phoenix/router/route.ex
```diff
@@ -121,13 +121,16 @@ defmodule Phoenix.Router.Route do
...
+        Phoenix.Router.Route,
+        {unquote(fwd_segments), unquote(route.plug), unquote(Macro.escape(route.opts))}
+      }
     end
   end
 
   defp build_dispatch(%Route{} = route) do
     quote do
-      {unquote(route.plug), unquote(route.opts)}
+      {unquote(route.plug), unquote(Macro.escape(route.opts))}
```
fertapric: I think this should always be an atom, because If you don't pass an atom, you will get a `** (CompileError) .../router.ex: invalid quoted expression:` when building the helpers. The error message is not very friendly, so I think we could improve it (in a different PR) with one of these options:

1. Provide a custom `ArgumentError` in the route macros (`match`, `get`, `post`, etc) is `plug_opts` is not an atom
2. Include a guard `is_atom(plug_opts)` in the route macros (`match`, `get`, `post`, etc) so it returns a `FunctionClauseError`.

I personally prefer the option 1, but I understand that it might be an edge case and the option 2 is enough.

fertapric: Forget what I said, that wouldn't work for cases like:

```elixir
action = :index
get "/hello", HelloController, action
```

The check can be done in `Phoenix.Router.Route` though.

chrismccord:
❤️❤️❤️🐥🔥

PR #3111, opened 2018-10-20
The mix task for phx.gen.cert should also propose `cipher_suite: :str…

LostKobrakai:
…ing` like the comments in dev.exs


chrismccord:
Thanks for this! Can you verify which older browsers and OS's we'd be leaving behind by using this as default?

chrismccord:
Oh, I see folks are having to use `:strong` for it to work on recent versions of chrome. That's very unexpected. Let me do some more digging. Thanks!

chrismccord:
Ah, you are right! I didn't know we were already recommending it.

chrismccord:
❤️❤️❤️🐥🔥

PR #3112, opened 2018-10-20
Log the configured URL instead of the raw ip

### LINKED ISSUE #3105: The Cowboy adapter should log out the host in the config instead of localhost or 0.0.0.0 when a host is provided
When you run `mix phx.server` the Phoenix Cowboy adapter logs out the endpoints along with the URL that you can access them on.

Right now the `host` value is based on the IP address. I think it should probably be based on the `host`passed into the endpoint config if there's one present in the url.

```
...
  url: [scheme: "https", host: "server.local", port: 4040],
...
``` 

### Expected behavior

```
[info] Running MyAppWeb.Endpoint with Cowboy2 using https://server.local:4040
```

### Actual behavior

```
[info] Running MyAppWeb.Endpoint with Cowboy2 using https://0.0.0.0:4040
```

This issue is caused because we're not grabbing the host out of the config here:
https://github.com/phoenixframework/phoenix/blob/4c53ace051e9180b82265310b097ce770b20724b/lib/phoenix/endpoint/cowboy2_adapter.ex#L69

sionide21:
Fixes #3105

### DIFF
--- a/lib/phoenix/endpoint/cowboy2_adapter.ex
+++ b/lib/phoenix/endpoint/cowboy2_adapter.ex
@@ -66,22 +66,16 @@ defmodule Phoenix.Endpoint.Cowboy2Adapter do
   @doc false
   def info(scheme, endpoint, ref) do
     server = "cowboy #{Application.spec(:cowboy)[:vsn]}"
-    "Running #{inspect endpoint} with #{server} at #{uri(scheme, ref)}"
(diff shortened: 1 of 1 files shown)


sionide21:
I can't make that spec fail locally

chrismccord:
❤️❤️❤️🐥🔥

PR #3118, opened 2018-10-22
Remove duplicated assert

fertapric:


chrismccord:
I guess we wanted to be extra sure :D ❤️❤️❤️🐥🔥

Issue #3123, opened 2018-10-25
Compiled phoenix.js does not produce strictly es5 output


chrismccord:
https://elixirforum.com/t/phoenix-1-4-rc-2-with-brunch-phoenix-js-is-not-transpiled

It looks like our switch to webpack for building phoenix.js internally produced less strict output than we assumed. We should turn whatever knobs in babel or webpack to make phoenix.js compatible with everything as before, by making it produce strictly es5 output.

hubertlepicki:
I *think* this is the knob you are looking for: https://babeljs.io/docs/en/7.0.0//babel-preset-env#forcealltransforms

...but turning it on had absolutely no effect for me.

davydog187:
I will take a look. Its possible the version is wrong or a misconfigured browserslist

hubertlepicki:
@davydog187 good catch with the wrong order, I did not think about that.

davydog187:
@hubertlepicki I've spent way too many hours of my life configuring the webpack config at my job.

hubertlepicki:
@davydog187 @chrismccord I just tested master with this patch on my app I am updating, and I can confirm that this sorts out the upgrade issues to 1.4, while leaving brunch intact for me. This is good, I think this can be closed.

PR #3126, opened 2018-10-25
Fix Phoenix.js ES5 output, upgrade webpack

davydog187:
Addresses #3123 

Order of webpack loaders was reversed


Gazler:
My hero. Thanks <3

PR #3129, opened 2018-10-28
Remove extra newline in generated context test cases

jeffkreeftmeijer:
Currently, context tests generated by running the html generator (`mix phx.gen.context Accounts User users name:string`), produce a test with an extra line with some trailing whitespace (illustrated with dots here):

        test "update_user/2 with valid data updates the user" do
          user = user_fixture()
          assert {:ok, %User{} = user} = Accounts.update_user(user, @update_attrs)

    ······
          assert user.name == "some updated name"
        end

This patch removes both newlines from the template, resulting in a generated test like this:

        test "update_user/2 with valid data updates the user" do
          user = user_fixture()
          assert {:ok, %User{} = user} = Accounts.update_user(user, @update_attrs)
          assert user.name == "some updated name"
        end

### DIFF
--- a/priv/templates/phx.gen.context/test_cases.exs
+++ b/priv/templates/phx.gen.context/test_cases.exs
@@ -36,9 +36,7 @@
 
     test "update_<%= schema.singular %>/2 with valid data updates the <%= schema.singular %>" do
       <%= schema.singular %> = <%= schema.singular %>_fixture()
-      assert {:ok, %<%= inspect schema.alias %>{} = <%= schema.singular %>} = <%= inspect context.alias %>.update_<%= schema.singular %>(<%= schema.singular %>, @update_attrs)
-
-      <%= for {field, value} <- schema.params.update do %>
+      assert {:ok, %<%= inspect schema.alias %>{} = <%= schema.singular %>} = <%= inspect context.alias %>.update_<%= schema.singular %>(<%= schema.singular %>, @update_attrs)<%= for {field, value} <- schema.params.update do %>
       assert <%= schema.singular %>.<%= field %> == <%= Mix.Phoenix.Schema.value(schema, field, value) %><% end %>
     end
 


priv/templates/phx.gen.context/test_cases.exs
```diff
@@ -38,8 +38,7 @@
       <%= schema.singular %> = <%= schema.singular %>_fixture()
       assert {:ok, %<%= inspect schema.alias %>{} = <%= schema.singular %>} = <%= inspect context.alias %>.update_<%= schema.singular %>(<%= schema.singular %>, @update_attrs)
 
-      <%= for {field, value} <- schema.params.update do %>
-      assert <%= schema.singular %>.<%= field %> == <%= Mix.Phoenix.Schema.value(schema, field, value) %><% end %>
+      <%= for {field, value} <- schema.params.update do %>assert <%= schema.singular %>.<%= field %> == <%= Mix.Phoenix.Schema.value(schema, field, value) %><% end %>
```
fertapric: If there is more than one field, this will print all of them in the same line `assert user.name == "some updated name"assert user.foo == "foo"`.

I think this should work:

```elixir
<%= schema.singular %> = <%= schema.singular %>_fixture()
assert {:ok, %<%= inspect schema.alias %>{} = <%= schema.singular %>} = <%= inspect context.alias %>.update_<%= schema.singular %>(<%= schema.singular %>, @update_attrs)
<%= for {field, value} <- schema.params.update do %>
assert <%= schema.singular %>.<%= field %> == <%= Mix.Phoenix.Schema.value(schema, field, value) %><% end %>
```

jeffkreeftmeijer: Thanks @fertapric, you’re right. Templates always make for the best puzzles. Your suggestion removes the first newline, and produces this:

        test "update_user/2 with valid data updates the user" do
          user = user_fixture()
          assert {:ok, %User{} = user} = Accounts.update_user(user, @update_attrs)
    ······
          assert user.email == "some updated email"
          assert user.name == "some updated name"
        end

That's one step closer, but we still have the trailing whitespace where. Since the other tests don't have a newline before the asserts, another option is to remove it altogether. I've updated the patch to produce this:

        test "update_user/2 with valid data updates the user" do
          user = user_fixture()
          assert {:ok, %User{} = user} = Accounts.update_user(user, @update_attrs)
          assert user.email == "some updated email"
          assert user.name == "some updated name"
        end

chrismccord:
❤️❤️❤️🐥🔥

PR #3135, opened 2018-10-31
Avoid .formatter.exs files with conflicting inputs

fertapric:
[The documentation of `mix format`](https://hexdocs.pm/mix/Mix.Tasks.Format.html#content) states that the parent `.formatter.exs` must not specify files, in its `:inputs` configuration, that are inside the subdirectories listed in the `:subdirectories` option. If this happens, the behaviour of which formatter configuration will be picked is unspecified.

[Upcoming Elixir versions will print an error message if there are conflicting files.](https://github.com/elixir-lang/elixir/pull/8323)

I haven't included the `priv` dir for non-Ecto projects because the installer does not put any Elixir files there, but I'll include it if you think it makes sense.


josevalim:

Issue #3139, opened 2018-11-04
Channel reconnect timer on socket open?


sb8244:
### Environment

* Phoenix.js version: master

### Expected behavior

When a socket opens, the channels would immediately attempt to connect in order to provide maximum availability of message processing.

### Actual behavior

When a socket opens, the channels do not immediately attempt to connect. The channels follow the same reconnection algorithm provided by the socket.

### Ideas

I see two solutions that could solve this problem. The first is to reset all channel reconnect timers onOpen of the socket. The second is to provide a custom reconnection algorithm for channels. I don't see channels as hitting the server when they attempt to connect, so it should be fairly cheap to attempt versus the socket.

chrismccord:
> The first is to reset all channel reconnect timers onOpen of the socket

This sounds like a good plan. Can you send a PR? Thanks!

sb8244:
Can do!

sb8244:
@chrismccord The change involved a bit more (at least for what I'm seeing) in order to prevent trying to rejoin when not in the rejoin flow.

PR #3141, opened 2018-11-04
Allow channels to immediately reconnect when the socket does

sb8244:
The timer state must be known in order to know if the channel was attempting to rejoin or not. It is
not correct to start a channel rejoin timer if it was not previously running. This involved changing the timer logic up a bit to clear out `this.timer` when it's done processing

Related to #3139


chrismccord:
❤️❤️❤️🐥🔥

GrantJamesPowell:
Nice!

PR #3144, opened 2018-11-07
Remove archives repo mention from guides

michallepicki:
I noticed that in the [Release 1.4.0 commit](https://github.com/phoenixframework/phoenix/commit/03243c69a20ff4b6ea757fc216a0569efb18827a#diff-781a53e648f3df8d16a08ec083b04bf4L11) Chris removed notes on pushing the installer to the archives repo, and the 1.4.0 installer is not published there, so the guides should no longer reference this repo.


chrismccord:
Thanks, and good catch! Since the hex archive install builds the archive on user's machines, it should be much more error proof, so we can simply remove the note about the alternative method. Can you kill the "build it ourselves note"? Thanks!

michallepicki:
@chrismccord I updated the PR

PR #3146, opened 2018-11-08
Include Phoenix version patch for changelong link

AdamT:
When generating a Phoenix 1.4.0 project, the link to the 1.4.0 changelong is:
https://github.com/phoenixframework/phoenix/blob/v1.4/CHANGELOG.md

This link is missing the patch. It should be:
https://github.com/phoenixframework/phoenix/blob/v1.4.0/CHANGELOG.md


wojtekmach:
https://github.com/phoenixframework/phoenix/blob/v1.4/CHANGELOG.md is preferred as it's gonna be the same for all 1.4.x maintenance releases; all we need to do is cut a v1.4 branch :)

PR #3153, opened 2018-11-09
Use `phx-hero` class instead of `jumbotron`

blucell:
Starting at phoenix 1.4, the class `jumbotron` does no longer exist, and `phx-hero` is used instead.


Gazler:
Thanks <3

PR #3162, opened 2018-11-14
fix ecto not loaded view error occur

warmestcolor:
when ecto not preload, the view can not render this situation. So use pattern match to identify which modoule(Ecto.NotLoaded or the view modoule) can be rendered.


michalmuskala:
I don't think that's correct. The fact that we don't know what the value of the association is, is extremely different from it being empty. Encoding a not loaded association as an empty object is just incorrect in most situations. It might be fine to add a pattern match in there, though, and crash if that happens with a slightly better error message instead of failing at the `foo.id` line. 

As an added bonus, it could also potentially allow the compiler to produce slightly better code since it knows the pattern matched variable contains a map.

warmestcolor:
add the pattern match in view render, and remove the final render for ecto association to expose error.

chrismccord:
I don't think this is the correct level to catch this. The struct match is fine, but in this case I don't it's a necessary match. Thanks!

PR #3166, opened 2018-11-14
Allow custom paths to be provided to :websocket and :longpoll

Gazler:
Previously, it was not possible to use a socket with a path that differed from
the key name. For example `:websocket` would always end with `/websocket` and
`:longpoll` would always end with `/longpoll`.

This commit allows a `:path` to be passed into the keyword options, the path
will be used instead of the default if available.

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -777,16 +777,18 @@ defmodule Phoenix.Endpoint do
 
     paths =
       if websocket do
-        triplet = {:websocket, socket, socket_config(websocket, Phoenix.Transports.WebSocket)}
-        [{socket_path(path, :websocket), triplet} | paths]
+        config = socket_config(websocket, Phoenix.Transports.WebSocket)
+        triplet = {:websocket, socket, config}
+        [{socket_path(path, config), triplet} | paths]
       else
         paths
       end
 
     paths =
       if longpoll do
-        plug_init = {endpoint, socket, socket_config(longpoll, Phoenix.Transports.LongPoll)}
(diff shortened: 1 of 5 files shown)


lib/phoenix/endpoint.ex
```diff
@@ -778,24 +778,30 @@ defmodule Phoenix.Endpoint do
...
+        [{socket_path(path, "websocket", websocket), triplet} | paths]
       else
         paths
       end
 
     paths =
       if longpoll do
         plug_init = {endpoint, socket, socket_config(longpoll, Phoenix.Transports.LongPoll)}
-        [{socket_path(path, :longpoll), {:plug, Phoenix.Transports.LongPoll, plug_init}} | paths]
+        [{socket_path(path, "longpoll", longpoll), {:plug, Phoenix.Transports.LongPoll, plug_init}} | paths]
```
josevalim: Instead of passing `longpoll` as third argument, I would store `socket_config(longpoll, Phoenix.Transports.LongPoll)` in a variable and pass it as argument instead. This way we don't need to handle non-lists inside `socket_path`. Or, even better, I would put `[path: "longpoll"]` as a default inside `Phoenix.Transports.LongPoll.default_config`, this way you don't need to provide a default here too and you can just do `Keyword.fetch!(config, :path)` inside `socket_path`.  Thoughts?

josevalim:
There will be a new patch release soon but for now you can use Phoenix
directly from the v1.4 branch as a hit dependency.
-- 


feld:
This does not appear to be merged into the v1.4 branch yet, so that wouldn't help

Gazler:
This is in the 1.4 branch now.

PR #3169, opened 2018-11-16
Allow variables to be used when defining socket paths

Gazler:
This commit allows sockets (websocket/longpoll) to use variables in the path.
For example, a socket can be defined with:

    socket "/ws/:user_id", MyApp.UserSocket,
      websocket: [path: "/project/:project_id"]

The params will be available in the connect map:

    def connect(%{"user_id" => user_id, "project_id" => project_id}), do: ...

The `__handler__/2` function now hands off the a `do_handler` which is
responsible for modifying the passed `conn` to add the `path_params` and
`params`. `__handler__/2` is now called with the entire `conn` instead of just
the `path_info` and expects a `conn` to be available in the returned tuple.

As this is only supported in Cowboy 2, the `test_helper.exs` has been
updated to ignore `:cowboy2` tests when testing with `plug_cowboy` 1.x.

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -730,9 +730,11 @@ defmodule Phoenix.Endpoint do
 
     dispatches =
       for {path, socket, socket_opts} <- sockets,
-          {path, return} <- socket_paths(module, path, socket, socket_opts) do
+          {path, type, conn_ast, socket, opts} <- socket_paths(module, path, socket, socket_opts) do
         quote do
-          def __handler__(unquote(path), _opts), do: unquote(Macro.escape(return))
+          defp do_handler(unquote(path), conn, _opts) do
+            {unquote(type), unquote(conn_ast), unquote(socket), unquote(Macro.escape(opts))}
+          end
         end
       end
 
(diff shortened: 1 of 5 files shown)


lib/phoenix/endpoint.ex
```diff
@@ -837,6 +860,17 @@ defmodule Phoenix.Endpoint do
...
 
+  ## Path params
+
+  It is possible to include variables in the path, these will be
+  available in the `params` that are passed to the socket.
+
+      socket "/ws/:user_id", MyApp.UserSocket,
+        websocket: [path: "/project/:project_id"]
+
+
```
josevalim: Extra new line.

lib/phoenix/endpoint.ex
```diff
@@ -762,11 +764,13 @@ defmodule Phoenix.Endpoint do
...
-      def __handler__(path, opts)
-      unquote(dispatches)
-      def __handler__(_, opts), do: {:plug, __MODULE__, opts}
+      def __handler__(%{path_info: path} = conn, opts), do: do_handler(path, conn, opts)
 
       unquote(instrumentation)
+
+      unquote(dispatches)
+      defp do_handler(_path, conn, opts), do: {:plug, conn, __MODULE__, opts}
+
```
josevalim: Extra new line.

PR #3178, opened 2018-11-21
Make explicit the status code in case of redirect

DianaOlympos:
As there are multiple code with slightly different redirect behaviour nowadays (302, 303 and 307 at least), this is just making explicit the behaviour of redirect for status code.

I am not sure of the wording. Would be happy to change it if you think it need more work.


chrismccord:
❤️❤️❤️🐥🔥

PR #3180, opened 2018-11-22
Clean css class from bootstrap

chenxsan:
Now that Phoenix doesn't use bootstrap anymore, I think the `help-block` from bootstrap 3 is no longer needed here.


michallepicki:
I'm not so sure, this css class was moved over from bootstrap to [phoenix.css](https://github.com/phoenixframework/phoenix/blob/d448bb06dd539174cb7a58e9e70952080023fa0a/installer/templates/phx_assets/phoenix.css#L58
)

chenxsan:
Oops, I didn't notice that.

PR #3181, opened 2018-11-23
Introduce `put_static_url` for more dynamic static url generation

rmoorman:
Similar to the `Phoenix.Controller.put_router_url/1` introduced in 1.4, this PR introduces a `Phoenix.Controller.put_static_url/1` function that allows to set the information used to generate the static url within the route helpers more dynamically.

My use-case for this is to enable one phoenix installation to serve multiple domains from one endpoint while having a way to set the correct url to use for static file serving seperately for all of them in cases when I have to use absolute urls. Basically what `put_router_url/1` does but then for static urls.

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -837,8 +837,9 @@ defmodule Phoenix.Controller do
   @doc """
   Puts the URL or `%URI{}` to be used for the static url generation.
 
-  This function overrides the default URL generation pulled
-  from the `%Plug.Conn{}`'s endpoint configuration.
+  Using this function on a `%Plug.Conn{}` struct tells `static_url/2` to use
+  the given information for URL generation instead of the the `%Plug.Conn{}`'s
+  endpoint configuration (much like `put_router_url/2` but for static URLs).
   """
   def put_static_url(conn, %URI{} = uri) do
     put_private(conn, :phoenix_static_url, uri)
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -835,6 +835,19 @@ defmodule Phoenix.Controller do
   end
 
   @doc """
+  Puts the URL or `%URI{}` to be used for the static url generation.
+
+  This function overrides the default URL generation pulled
+  from the `%Plug.Conn{}`'s endpoint configuration.
+  """
+  def put_static_url(conn, %URI{} = uri) do
+    put_private(conn, :phoenix_static_url, uri)
+  end
+  def put_static_url(conn, url) when is_binary(url) do
+    put_private(conn, :phoenix_static_url, url)
+  end
+
+  @doc """
   Puts the format in the connection.
 
   See `get_format/1` for retrieval.
--- a/lib/phoenix/router/helpers.ex
+++ b/lib/phoenix/router/helpers.ex
@@ -157,12 +157,14 @@ defmodule Phoenix.Router.Helpers do
       """
       def static_url(%Conn{private: private} = conn, path) do
         case private do
-          %{phoenix_static_url: %URI{} = uri} ->
-            %URI{uri | path: static_path(conn, path)}
-            |> URI.to_string()
+          %{phoenix_static_url: %URI{path: nil} = uri} ->
+            URI.to_string(%URI{uri | path: path})
+
+          %{phoenix_static_url: %URI{path: static_path} = uri} ->
+            URI.to_string(%URI{uri | path: static_path <> path})
 
           %{phoenix_static_url: url} when is_binary(url) ->
-            url <> static_path(conn, path)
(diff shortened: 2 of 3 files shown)


lib/phoenix/router/helpers.ex
```diff
@@ -156,7 +156,17 @@ defmodule Phoenix.Router.Helpers do
       Generates url to a static asset given its file path.
       """
       def static_url(%Conn{private: private} = conn, path) do
-        static_url(private.phoenix_endpoint, path)
+        case private do
+          %{phoenix_static_url: %URI{} = uri} ->
+            %URI{uri | path: static_path(conn, path)}
```
rmoorman: @chrismccord when I would like to introduce this change, would you rather have the return value of `static_path/2` override the path from the given override `URI` struct or would you prefer that the path in the struct is prepended to the return value of `static_path/2`?

josevalim: I think we just ignore the path in `put_router_url`, right? Anyway, I would make the behaviour the same. :)

lib/phoenix/controller.ex
```diff
@@ -834,6 +834,19 @@ defmodule Phoenix.Controller do
     put_private(conn, :phoenix_router_url, url)
   end
 
+  @doc """
+  Puts the URL or `%URI{}` to be used for the static url generation.
+
+  This function overrides the default URL generation pulled
+  from the `%Plug.Conn{}`'s endpoint configuration.
```
josevalim: Let's mention in the docs what you said in the issues description: that this is the same as `put_router_url` but it is for static urls. :)

josevalim:
Sorry for the delay @rmoorman! I have added two comments, can you please take a look at them? Thanks!

rmoorman:
@josevalim no problem. I am glad that there is a pair of eyes that had a look at it after all!
I am on it!

josevalim:
@rmoorman cool! Please ping me once you address the changes and I will take another look. :)

lib/phoenix/router/helpers.ex
rmoorman: @josevalim I just gave this another thought and I think that discarding the path information when using an `%URI{}` struct might not be an optimal thing to do.

One example when this might not be optimal is when one would like to use an S3 bucket url in the form of `http(s)://s3.amazonaws.com/<bucket>/<object>` (or some other url with a path) to host the static files. While passing a string as an override to `put_static_url/2`, this string could of course also contain a path, but in order to have the same possibility available in case one passes a struct I guess appending the file path to the path inside the uri struct could be a sensible thing to do.

I could imagine that things are different for the routing mechanisms of phoenix though, but why is the path nillified within the `Phoenix.Router.Helpers.url/2` function when an `%URI{}` struct is passed when the endpoint itself can be configured to actually use a path?

josevalim: Good points. Let's keep the path as well then.

rmoorman: Thank you for replying so quickly :-) I did and am currently revisiting the inline documentation.

I am still wondering if the `put_router_url` function behaves correctly when an URI struct is given? As seen [here](https://github.com/phoenixframework/phoenix/blob/master/test/phoenix/router/helpers_test.exs#L415), the test uses a path but it is not present within the final url being generated.

lib/phoenix/controller.ex
rmoorman: @josevalim I just updated the PR to include a reference to the `put_router_url` function (also applied some rewording to hopefully describe a bit clearer what the function does).

rmoorman:
@josevalim I updated the PR to reflect our stance on the discussed points so far.

I am still wondering if the `put_router_url` function behaves correctly when an URI struct is given? As seen [here](https://github.com/phoenixframework/phoenix/blob/master/test/phoenix/router/helpers_test.exs#L415), the test uses a path but it is not present within the final url being generated.

Anyway, please let me know if there is anything else I could improve!

josevalim:
Can you  please take a look at changing `put_router_url` to also consider the path?  If it is relatively straight-forward, maybe we should go ahead and we do it? In the worst scenario, we will learn why it is a bad idea. :) Thankss a lot  :heart:

rmoorman:
@josevalim sure! I will give it a shot but would like to put the work into another PR to keep this one focused.

josevalim:
Perfect. I will re-review this accordingly over the weekend and merge it. Thank you!

rmoorman:
@josevalim Thank you very much in advance!

One further remark/question regarding the to be reviewed changes. The code that updated the path currently calls the static_path helper. Would that be a correct thing to do or would that be unnecessary (maybe it does have some implications regarding manifests)?

josevalim:

PR #3184, opened 2018-11-26
Fix code block format

chenxsan:
There's a formatting problem https://hexdocs.pm/phoenix/1.4.0/Phoenix.ConnTest.html#module-controller-testing

image


Gazler:
Thanks <3

PR #3186, opened 2018-11-26
fix minor typo in contexts guide (ensure_author_exits -> ensure_author_exists)

delta1:
Hi there, I spotted a minor typo so figured I would submit a PR to fix it :) 

Thanks for all the great work


Gazler:
Thanks <3

PR #3195, opened 2018-12-04
Allow a presence to be created that doesn’t broadcast diffs

sb8244:
This is useful when the presence is maintained for server side reasons (knowing who is connected) and does not need sent to channels.

This comes from the work in PushEx to track down the source of scaling problems. In that case, the diff isn't needed as the clients are not sent presence diffs.


sb8244:
Closed after discussion

PR #3196, opened 2018-12-04
Removes extra line break on phx.gen.schema template

nysertxs:
Removes extra line break that makes the formatter update newly generated schema file


josevalim:

PR #3197, opened 2018-12-04
Automatically hibernate channels after some time

josevalim:


lib/phoenix/channel.ex
```diff
@@ -248,6 +247,16 @@ defmodule Phoenix.Channel do
         {:noreply, socket}
       end
 
+  ## Hibernation
+
+  From Erlang/OTP 20, channels automatically hibernate to save memory
+  after 15_000 milliseconds of inactivity. This can be customized by
```
sb8244: "customized by on" is a bit confusing to me

michalmuskala:
Why is this option on `use`? Isn't there a way to pass it as a runtime option? It seems weird to me we're promoting a purely runtime option into a compile-time constant.

josevalim:
We can't unfortunately, unless we tell folks to override start_link.

It is similar to the `log` options. Logging happens at runtime, but because the channel spawning and joining happens inside the framework before we invoke any user code, then it needs to be there at compile time.

josevalim:

PR #3198, opened 2018-12-05
Document the use of `Phoenix.Tracker` instead of `Phoenix.Presence`

sb8244:
This PR adds some documentation to clear up some confusion I had. My use case was around using `Phoenix.Presence` but not broadcasting diff events (handle_diff -> no op). Phoenix.Tracker is a better fit than Presence in this situation, but I was not aware of it from the docs.

### DIFF
--- a/guides/presence.md
+++ b/guides/presence.md
@@ -4,8 +4,6 @@ Phoenix Presence is a feature which allows you to register process information o
 
 Phoenix Presence is special for a number of reasons. It has no single point of failure, no single source of truth, relies entirely on the standard library with no operational dependencies and self heals. This is all handled with a conflict-free replicated data type (CRDT) protocol.
 
-Presence is designed for situations where topic subscription information should be broadcast to connected channels. It is possible to use `Phoenix.Tracker` directly in other use cases, such as wanting to manually dispatch or skip change events.
-
 To get started with Presence we'll first need to generate a presence module. We can do this with the `mix phx.gen.presence` task:
 
 ```console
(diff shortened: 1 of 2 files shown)


guides/presence.md
```diff
@@ -4,6 +4,8 @@ Phoenix Presence is a feature which allows you to register process information o
 
 Phoenix Presence is special for a number of reasons. It has no single point of failure, no single source of truth, relies entirely on the standard library with no operational dependencies and self heals. This is all handled with a conflict-free replicated data type (CRDT) protocol.
 
+Presence is designed for situations where topic subscription information should be broadcast to connected channels. It is possible to use `Phoenix.Tracker` directly in other use cases, such as wanting to manually dispatch or skip change events.
```
josevalim: I believe the guides are used as entry-point, so I am not sure having this quote at the very beginning will be helpful, because users will likely be unaware of what this all means or even confused by it. My suggestion is to keep this information exclusively in the docs and expand it in there a bit.

lib/phoenix/presence.ex
```diff
@@ -5,9 +5,15 @@ defmodule Phoenix.Presence do
...
-  tracker process to handle presence information.
+  module defines a supervisor and a module that implements the
+  `Phoenix.Tracker` behaviour that uses `Phoenix.PubSub` to
+  broadcast presence updates.
+
+  In case you want to use only a subset of the functionality
+  provided by `Phoenix.Presence`, such as tracking processes
+  but without broadcasting updates, we recommend you to look
+  at the `Phoenix.Tracker` functionality from the `phoenix_pubsub`
+  project.
```
josevalim: I have detailed more the information you wanted to add, let me know what you think.

guides/presence.md
sb8244: Yes that is fair. I was bouncing that around in my head

lib/phoenix/presence.ex
sb8244: Great edits @josevalim. I went ahead and removed it from the guide. 

I did tweak the wording around a little bit on one phrase "recommend you to look". Some of my co-workers and I have been discussing the use of "to" verb lately and I've realized that removing it from my writing has helped with clarity.

josevalim:

PR #3201, opened 2018-12-06
Fix default phoenix.css template's SVG not rendering in Firefox

monting:
URL encode default phoenix.css template's `<select>` element down-arrow SVG's `#` into `%23`.

Firefox wasn't rendering the arrow.

screen shot 2018-12-05 at 9 48 16 pm

FF on the left, Chrome on the right


michallepicki:
AFAIK Phoenix's distribution of Milligram already includes a few necessary changes, but I think issues like this one should be fixed upstream in [milligram repository](https://github.com/milligram/milligram)

monting:
@michallepicki I agree. I had checked milligram's repo, and it is indeed fixed in the same way:
https://github.com/milligram/milligram/blob/master/dist/milligram.css#L250

Not sure how long it would take for the Phoenix to get to upgrading milligram, and go through all the necessary changes of doing so. I think this patch would be useful to fix something broken, in the meantime.

chrismccord:
❤️❤️❤️🐥🔥

PR #3204, opened 2018-12-08
Fix DataCase.errors_on/1 when using changeset with array field

kubasub:
## Overview
`DataCase.errors_on/1` raises an error when it receives a changeset with an array field which was `cast`-ed with an invalid type.

## Details
When validating the type of an array field fails, `cast/4` it will put an error on the changeset, such as:
```elixir
[my_list_field: {"is invalid", [type: {:array, :string}, validation: :cast]}]
```

While processing this error, `errors_on/1` will try to `to_string/1` the `{:array, :string}` tuple, which produces the error:
```
** (Protocol.UndefinedError) protocol String.Chars not implemented for {:array, :string}. This protocol is implemented for: Atom, BitString, Date, DateTime, Decimal, Float, Integer, List, NaiveDateTime, Postgrex.Copy, Postgrex.Query, Time, URI, Version, Version.Requirement
```

## Solution
`to_string/1` is called on error metadata as part of processing template error messages (eg. error message `"must be equal to %{number}"` and metadata `[number: 19]`). Thus, **the implemented solution is to skip trying to interpolate error message embeds when the value to be embedded is a tuple.**

## Test Case
I didn't write a test for this, as there aren't any tests for installer templates -- nor do I know where I would add them. I tested locally with a setup like:
```elixir
defmodule MyAppWeb.Drzewo do
  use Ecto.Schema
  import Ecto.Changeset

  embedded_schema do
    field :kolory, {:array, :string}
  end

  def changeset(attrs) do
    cast(%Drzewo{}, attrs, [:kolory])
  end
end

defmodule MyAppWeb.DrzewoTest do
  use MyApp.DataCase

  alias MyAppWeb.Drzewo

  test "errors_on raises an error when passed an invalid type for the array field" do
    drzewo_params = %{kolory: 1}

    changeset = Drzewo.changeset(drzewo_params)
    changeset.valid?

    errors_on(changeset)
  end
end
```

### DIFF
--- a/installer/templates/phx_ecto/data_case.ex
+++ b/installer/templates/phx_ecto/data_case.ex
@@ -45,9 +45,13 @@ defmodule <%= app_module %>.DataCase do
   """
   def errors_on(changeset) do
     Ecto.Changeset.traverse_errors(changeset, fn {message, opts} ->
-      Enum.reduce(opts, message, fn {key, value}, acc ->
-        String.replace(acc, "%{#{key}}", to_string(value))
-      end)
+      Enum.reduce(opts, message, &replace_message_embed/2)
     end)
+  end
+
+  defp replace_message_embed({_key, value}, message) when is_tuple(value), do: message
+
+  defp replace_message_embed({key, value}, message) do
+    String.replace(message, "%{#{key}}", to_string(value))
   end
 end
--- a/installer/templates/phx_ecto/data_case.ex
+++ b/installer/templates/phx_ecto/data_case.ex
@@ -45,13 +45,9 @@ defmodule <%= app_module %>.DataCase do
   """
   def errors_on(changeset) do
     Ecto.Changeset.traverse_errors(changeset, fn {message, opts} ->
-      Enum.reduce(opts, message, &replace_message_embed/2)
+      Regex.replace(~r"%{(\w+)}", message, fn _, key ->
+        Map.get(opts, String.to_atom(key), key) |> to_string()
+      end)
     end)
-  end
-
-  defp replace_message_embed({_key, value}, message) when is_tuple(value), do: message
-
-  defp replace_message_embed({key, value}, message) do
-    String.replace(message, "%{#{key}}", to_string(value))
   end
 end


aaronrenner:
I have also seen this error with a field set up to `{:map, :string}`.

My approach was to rescue the `Protocol.UndefinedError`:

```elixir
 def errors_on(changeset) do
    Changeset.traverse_errors(changeset, fn {message, opts} ->
      Enum.reduce(opts, message, fn {key, value}, acc ->
        try do
          String.replace(acc, "%{#{key}}", to_string(value))
        rescue Protocol.UndefinedError ->
          acc
        end
      end)
    end)
  end
```

Definitely not as pretty as the code in the generator, but it does fix the issue. I wish there was a way to see if a term conformed to the `String.Chars` protocol without rescuing.

wojtekmach:
I think the underlying problem is we're trying to interpolate into the error message key/values that were never meant to be there. Maybe instead of reducing over all error `opts`, we should iterate over `Keyword.take(opts, [:count, ...])`?

Alternatively, we could only go over keys that are meant to be interpolate, perhaps with something like this?

```elixir
def f(msg, opts) do
  opts = Enum.into(opts, %{}, fn {key, value} -> {Atom.to_string(key), value} end)
  Regex.replace(~r"%{(\w+)}", msg, fn _, key -> opts |> Map.get(key, key) |> to_string() end)
end
```

```elixir
iex> msg = "should be %{count} character(s)"
iex> opts = [count: 2, validation: :cast, custom: [:foo, :bar]]
iex> f(msg, opts)
"should be 2 character(s)"
```

Thoughts?

All that being said, I think the generated `errors_on` function is meant to be just a starting point to be tailored to project's custom validations (and custom error opts). For example, if some schema has a array field, how are errors for that schema's changeset delivered to the client? There's probably an Ecto.Changeset.traverse_errors call somewhere in production code to make that happen and perhaps the best way would be to have `errors_on` re-use that logic? So maybe it's only a matter of documentation after all?

installer/templates/phx_ecto/data_case.ex
```diff
@@ -45,8 +45,8 @@ defmodule <%= app_module %>.DataCase do
   """
   def errors_on(changeset) do
     Ecto.Changeset.traverse_errors(changeset, fn {message, opts} ->
-      Enum.reduce(opts, message, fn {key, value}, acc ->
-        String.replace(acc, "%{#{key}}", to_string(value))
+      Regex.replace(~r"%{(\w+)}", message, fn _, key ->
+        Map.get(opts, String.to_atom(key), key) |> to_string()
```
kubasub: I think the use of `String.to_atom/1` here is acceptable since they string is a "trusted" value.

If we're not comfortable with this, I can first transform the opts using the logic in @wojtekmach 's PR comment so that we don't need to use `String.to_atom/1`. I personally just liked the conciseness of the code.

wojtekmach: For a test helper function String.to_atom is imo acceptable 👍

chrismccord:
> All that being said, I think the generated `errors_on` function is meant to be just a starting point to be tailored to project's custom validations (and custom error opts). For example, if some schema has a array field, how are errors for that schema's changeset delivered to the client? There's probably an Ecto.Changeset.traverse_errors call somewhere in production code to make that happen and perhaps the best way would be to have `errors_on` re-use that logic? So maybe it's only a matter of documentation after all?

@wojtekmach worded this perfectly as there is only so much we can do and `errors_on` is generated inside the user's app exactly as a starting point. Map/array handling will be app specific so more user-land code will be required on a per app basis. @kubasub If you feel a doc PR could be helpful in some way around this, please let us know. Thanks!

kubasub:
@chrismccord could I bother you to take one more quick look?

I _completely_ agree with you and Wojtek that `errors_on/1` is a starting point.

My main concern is that the current approach leads to a simple test case failing. We _shouldn't_ add to `errors_on/1`, since it's a starting point. The proposal reworks what is already there to cover more built-in ecto validations. Nothing that's application specific.

My other concern is that `errors_on/1` isn't doing what we actually want it to do.

* The existing logic _loops over a changeset error's `opts`_ and tries to find a match in the template string.
* The proposed logic _loops over the template string_ and tries to find a matching `opt`.

The latter, to me, seems what we want to happen: "Replace the keys in the template string with values from `opts`".

Thanks for taking a look and, of course, for building 🐥🔥.

chrismccord:
❤️❤️❤️🐥🔥

PR #3209, opened 2018-12-11
Don't use json encoder for map fields

### LINKED ISSUE #3208: `mix phx.gen.json` generates incorrect asserts for map fields
### Environment

* Elixir version (elixir -v): 1.7.4
* Phoenix version (mix deps): 1.4.0
* Operating system: MacOS

### Expected behavior
`mix phx.gen.json Accounts User users params:map` should generate following asserts in controller tests (map field asserts with empty map):
```elixir
assert %{
         "id" => id,
         "name" => "some name",
         "params" => %{}
```

### Actual behavior
map field asserts with empty tuple now:
```elixir
assert %{
         "id" => id,
         "name" => "some name",
         "params" => {}
```

van-mronov:
Fixes #3208.

### DIFF
--- a/priv/templates/phx.gen.json/controller_test.exs
+++ b/priv/templates/phx.gen.json/controller_test.exs
@@ -37,7 +37,7 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
 
       assert %{
                "id" => id<%= for {key, val} <- schema.params.create do %>,
-               "<%= key %>" => <%= Phoenix.json_library().encode!(val) %><% end %>
+               "<%= key %>" => <%= if val == %{}, do: "%{}", else: Phoenix.json_library().encode!(val) %><% end %>
              } = json_response(conn, 200)["data"]
     end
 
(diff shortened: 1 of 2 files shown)


priv/templates/phx.gen.json/controller_test.exs
```diff
@@ -37,7 +37,7 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
 
       assert %{
                "id" => id<%= for {key, val} <- schema.params.create do %>,
-               "<%= key %>" => <%= Phoenix.json_library().encode!(val) %><% end %>
+               "<%= key %>" => <%= if val == %{}, do: "%{}", else: Phoenix.json_library().encode!(val) %><% end %>
```
AndrewDryga:
```suggestion
               "<%= key %>" => <%= if val == %{}, do: "{}", else: Phoenix.json_library().encode!(val) %><% end %>
```

Probably I'm missing something, but empty map in json is `{}`, not `%{}` (which is Elixir map).

wojtekmach: The patch is fine, it's about generating _Elixir_ code, for `key = "foo"` and `val = %{}` it would generate:

```elixir
assert
  # ...
  "foo" => %{}
```

which is correct.

Since it's about generating Elixir map, another way would be to do:

```elixir
val |> Phoenix.json_library().encode!() |> Phoenix.json_library().decode!() |> inspect()
```

which is probably a bit too much though :)

van-mronov: @AndrewDryga as @wojtekmach mentioned it's about generating asserts with `json_response` which decodes json body to elixir structures.

@wojtekmach yeah your code is common. Keep PR as is or update?)

van-mronov: So, guys, what do think about this PR?

PR #3212, opened 2018-12-15
Converts non-mutated js vars from let to const

joshnuss:
Hi :wave: 

I noticed that the JavaScript generated for the channel code declared many variable as `let` when `const` would be a bit clearer.

Figured while I did that, might as well fix it everywhere.

This PR converts all the .js code that used `let` but didn't mutate into `const`.

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -8,7 +8,7 @@
  * Connect to the server using the `Socket` class:
  *
  * ```javascript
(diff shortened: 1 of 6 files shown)


joshnuss:
If you'd like, I'd be happy to open a PR to add a `.eslintrc` and a check in CI

chrismccord:
Please see other similar PR and discussions around let vs const. Thanks!

PR #3214, opened 2018-12-16
Update heroku.md

bredikhin:
Fix web socket timeout setting to match the latest version of Phoenix.

### DIFF
--- a/guides/deployment/heroku.md
+++ b/guides/deployment/heroku.md
@@ -200,15 +200,14 @@ config :hello, Hello.Repo,
 Finally, we need to decrease the timeout for the websocket transport in `lib/hello_web/endpoint.ex`:
 
 ```elixir
-defmodule HelloWeb.UserSocket do
-  use Phoenix.Socket
+defmodule HelloWeb.Endpoint do
+  use Phoenix.Endpoint, otp_app: :hello
 
-  ...
-
-  ## Transports
   socket "/socket", HelloWeb.UserSocket,
     websocket: [timeout: 45_000],
     longpoll: false
+
+  ...
 end
 ```
 
@@ -271,7 +270,7 @@ Let's commit all our changes:
 ```
 $ git add config/prod.exs
 $ git add Procfile
-$ git add lib/hello_web/channels/user_socket.ex
+$ git add lib/hello_web/endpoint.ex
 $ git commit -m "Use production config from Heroku ENV variables and decrease socket timeout"
 ```
 


guides/deployment/heroku.md
```diff
@@ -144,6 +144,7 @@ Then, we'll add the production database configuration to `config/prod.exs`:
 ```elixir
 # Configure your database
 config :hello, Hello.Repo,
+  adapter: Ecto.Adapters.Postgres,
```
Gazler: I don't think the adapter is required in the config going forward, it should be specified in the `MyApp.Repo` module instead.

https://github.com/phoenixframework/phoenix/blob/473a3cb40aae1ba36c70dfc4841ace2d3aab7471/installer/templates/phx_ecto/repo.ex#L4

guides/deployment/heroku.md
```diff
@@ -192,23 +193,25 @@ config :logger, level: :info
...
   url: System.get_env("DATABASE_URL"),
   pool_size: String.to_integer(System.get_env("POOL_SIZE") || "10"),
   ssl: true
 ```
 
-Finally, we need to decrease the timeout for the websocket transport in `lib/hello_web/endpoint.ex`:
+Finally, we need to decrease the timeout for the websocket transport in `web/channels/user_socket.ex`:
```
Gazler: I think this path was correct before?

guides/deployment/heroku.md
```diff
@@ -271,7 +274,7 @@ Let's commit all our changes:
 ```
 $ git add config/prod.exs
 $ git add Procfile
-$ git add lib/hello_web/channels/user_socket.ex
+$ git add web/channels/user_socket.ex
```
Gazler: I think this path was correct before?

guides/deployment/heroku.md
```diff
@@ -192,23 +193,25 @@ config :logger, level: :info
...
   pool_size: String.to_integer(System.get_env("POOL_SIZE") || "10"),
   ssl: true
 ```
 
-Finally, we need to decrease the timeout for the websocket transport in `lib/hello_web/endpoint.ex`:
+Finally, we need to decrease the timeout for the websocket transport in `web/channels/user_socket.ex`:
 
 ```elixir
-defmodule HelloWeb.UserSocket do
+defmodule Hello.UserSocket do
```
Gazler: I think this was correct as `HelloWeb.Socket.`

bredikhin:
My bad, was using the wrong version of the example app. The PR is updated as requested, thanks!

chrismccord:
The current PR uses the deprecated `transport` macro. The existing file has the correct `socket` macro and transport options, but it shows it under a `UserSocket` module instead of the user's Endpoint module. @bredikhin can you update the PR to rename the `HelloWeb.UserSocket` to `HelloWeb.Endpoint`, w/ `use Phoenix.Endpoint`? Thanks!

bredikhin:
Pushed the requested update. Also followed the whole guide again and noticed it never mentions `git add` for `compile` and `phoenix_static_buildpack.config` (and if you don't, it just tries to use `brunch`). LMK if I should add a fix for this one or if it can be "safely assumed" (probably shouldn't be).

chrismccord:
❤️❤️❤️🐥🔥 Please send the Buildpack stuff in a new PR. Thanks!!!

PR #3217, opened 2018-12-21
Enabling recycle of custom headers

AlexGascon:
Adding a new argument to specify which headers will be persisted after recycling the connection. By default we keep the existing behavior, so only `accept` and `authorization` are persisted.


chrismccord:
❤️❤️❤️🐥🔥

PR #3225, opened 2019-01-01
Improved phx.routes umbrella error message

efexen:
Hey 👋 & Happy New Year 🎆 

I Ran the `phx.routes` mix task the other day on an umbrella project and I was greeted by the following error message

```
** (Mix) umbrella applications require an explicit router to be given to phx.routes
```

It's not clear how one should specify the router, a file path or a module name so I added an example to make it clear. The new error message as follows:

```
** (Mix) umbrella applications require an explicit router to be given to phx.routes, for example:

  $ mix phx.routes MyAppWeb.Router
```

This is well documented but thought this might save someone a bit of confusion.

We could also iterate through the apps and see if any of them have defined a router module with the default naming conventions and list them to make the error even more useful, something like this:

```
** (Mix) umbrella applications require an explicit router to be given to phx.routes, for example:

  $ mix phx.routes MyAppWeb.Router

Found these routers:

  MyAppWeb.Router
  MyAppApi.Router
```

This PR just includes the improved error message without the router discovery part. I'd be happy to work on that if you think it would make a nice addition.

Hope this is all ok 👍 

Thanks,
Ville

### DIFF
--- a/lib/mix/tasks/phx.routes.ex
+++ b/lib/mix/tasks/phx.routes.ex
@@ -38,14 +38,20 @@ defmodule Mix.Tasks.Phx.Routes do
 
   defp router(nil, base) do
     if Mix.Project.umbrella?() do
-      Mix.raise "umbrella applications require an explicit router to be given to phx.routes"
+      Mix.raise """
+      umbrella applications require an explicit router to be given to phx.routes, for example:
+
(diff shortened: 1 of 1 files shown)


chrismccord:
❤️❤️❤️🐥🔥

PR #3226, opened 2019-01-03
Explicitly include the layout option when generating render_errors

Gazler:
This will hint to the users that a layout can be provided for error
pages.


chrismccord:
❤️❤️❤️🐥🔥

PR #3227, opened 2019-01-03
Update the doc to use socket/3 as socket/2 is deprecated

sb8244:


lib/phoenix/test/channel_test.ex
```diff
@@ -17,7 +17,7 @@ defmodule Phoenix.ChannelTest do
   `subscribe_and_join/4` topics and channels:
 
       {:ok, _, socket} =
-        socket("user:id", %{some_assigns: 1})
+        socket(UserSocket, "user:id", %{some_assigns: 1})
```
sb8244: It looks `UserSocket` is referred to in the other setup in this doc

chrismccord:
❤️❤️❤️🐥🔥

PR #3229, opened 2019-01-03
Fix the Elixir minimum version on installation guide to 1.5

victor-am:
Starting from version 1.4, Phoenix generates the Elixir child spec (making Elixir 1.5 the minimum working version for new Phoenix 1.4 apps, which is the case for the installation guide).

More info on: [v1.4.0 changelog](https://github.com/phoenixframework/phoenix/blob/v1.4/CHANGELOG.md#phxnew-installer).


chrismccord:
❤️❤️❤️🐥🔥

PR #3233, opened 2019-01-07
Use Elixir v1.5's streamlined child specs in docs

### LINKED ISSUE #3231: Update Endpoint docs with new supervisor syntax
The [Endpoint module docs](https://hexdocs.pm/phoenix/Phoenix.Endpoint.html) and [Endpoint guide](https://hexdocs.pm/phoenix/endpoint.html) point out that
>  By default, the Endpoint is added to the supervision tree in lib/hello/application.ex as a supervised process

> supervisor(HelloWeb.Endpoint, [])

while the phx generator [doesn't do that anymore](https://github.com/phoenixframework/phoenix/blob/master/installer/templates/phx_single/lib/app_name/application.ex#L14) and instead adds just the endpoint module name (AFAIK it's thanks to the new supervisor syntax introduced in elixir 1.5).

michallepicki:
Closes #3231

### DIFF
--- a/guides/endpoint.md
+++ b/guides/endpoint.md
@@ -10,7 +10,7 @@ defmodule Hello.Application do
     #...
 
     children = [
-      supervisor(HelloWeb.Endpoint, []),
+      HelloWeb.Endpoint
     ]
 
     opts = [strategy: :one_for_one, name: Hello.Supervisor]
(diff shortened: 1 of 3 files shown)


Gazler:
Thanks <3

PR #3241, opened 2019-01-12
Update heroku.md

bredikhin:
We need to make sure static assets buildpack config files are added to version control and pushed to Heroku, otherwise the default config will be used (which tries to run `brunch`, etc.). Also, some mentions of it are removed from the deployment logs to minimize confusion.


edwinthinks:
Good catch. I didn't notice this issue because I had did a `git add .` but this makes sense! 👍 LGTM :)?

PR #3242, opened 2019-01-12
Initializes Git in Phx.New task

tiagonbotelho:
Based on the Rails framework, when `rails new app` is issued it gets created with git already initialized.

This PR attempts to replicate that behaviour.

I've also made the option of skipping the initialization of `git` entirely by passing `--no-git` as an argument

### DIFF
--- a/installer/lib/mix/tasks/phx.new.ex
+++ b/installer/lib/mix/tasks/phx.new.ex
@@ -92,7 +92,7 @@ defmodule Mix.Tasks.Phx.New do
   @switches [dev: :boolean, webpack: :boolean, ecto: :boolean,
              app: :string, module: :string, web_module: :string,
              database: :string, binary_id: :boolean, html: :boolean,
-             umbrella: :boolean]
+             umbrella: :boolean, git: :boolean]
 
   def run([version]) when version in ~w(-v --version) do
     Mix.shell.info "Phoenix v#{@version}"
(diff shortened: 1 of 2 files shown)


rmoorman:
Hello @tiagonbotelho , I am using git for all of my projects but often the generated folder does not end up being the repository root. For users that have a src directory for example or prefer to have an as clean as possible repository root in general, this would mean that they have to get rid of the .git directory right?

tiagonbotelho:
> often the generated folder does not end up being the repository root
Is this true for an Phoenix project? Could you give me an example please 😊 I believe `phx.new` is responsible for creating  the root of a Phoenix project

You can always use the `--no-git` flag and the `.git` folder will not get generated meaning that you are free to configure git in whatever structure you may prefer 😄 

Let me know if this answers your question @rmoorman 🙇

rmoorman:
@tiagonbotelho but wouldn't it be better to make it opt-in instead of opt-out? So explicitly passing `--git-init` for example in order to initialize a git repository?

mhinz:
Let me play the devil's advocat: Did Rails had a good reason to introduce that?

Adding multiple lines to Phoenix to save oneself from typing `git init` doesn't seem like good tradeoff. I think it's easy to add such code and more flags, but is it really worth the added complexity?

tiagonbotelho:
@mhinz your question is 100% valid thank you 🙇 I tend to lean on the convention over configuration quite a lot, so for me it makes sense that the Phoenix framework provides this straight out the box.

@rmoorman makes a good point, we could make this behaviour opt-in instead of opt-out. But that would mean that initializing a git repository in a newly-created phoenix folder no longer is the convention, which makes me question the validity of this PR.

For me the flow is clear. Whenever I do `mix phx.new` I will almost 100% of the times run `git init` afterwards. Saving the extra step might seem lazy but it is a good addition in terms of UX imho 😊

mpranjic:
I have more negative feelings about this than positive.
For me, most of the time phoenix is not a repo root.
I don't think it's worth adding such option, given it's easy to do it myself (git init... done). Convention is great sometimes (eg. generating .gitignore file brings a lot of value ) but this isn't, and I don't think having to write --no-git is a good UX :smile:

chrismccord:
We generate a gitignore for easy git integration once you are ready for your initial commit, but I don't think we need to go as far as initializing a git repo. Thanks!

PR #3246, opened 2019-01-17
Add default argument for change_* in controller and context's template

m31271n:
Why?

A Context is responsible for abstracting away all backend details, so
schema related things should be placed in it.

If the schema related things is leaked to controller, the controller
is not schema-agnostic anymore.

Before this commit:
```
// file post_controller.ex
alias XX.Posts
alias XX.Posts.Post
changeset = Posts.change_post(%Post{})
```

After this commit:
```
// file post_controller.ex
alias XX.Posts
changeset = Posts.change_post()
```

Schema related things is hidden under context. Awesome! right?

### DIFF
--- a/priv/templates/phx.gen.context/access_no_schema.ex
+++ b/priv/templates/phx.gen.context/access_no_schema.ex
@@ -84,6 +84,6 @@
       %Todo{...}
 
   """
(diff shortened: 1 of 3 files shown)


chrismccord:
In this case `%Post{}`, and the default schema generated code, is a public datastructure, so the caller providing the datastructure is okay. If you want to hide this from the caller, you are free to make the desired changes if that fits your app. Thanks!

PR #3250, opened 2019-01-18
Show socket paths on mix phx.routes. Closes #3237

### LINKED ISSUE #3237: Show socket paths on `mix phx.routes`
Currently only the paths defined in the router are displayed. It'd be nice to also display the paths for sockets and which modules the sockets route to.

chrismccord:
@Gazler let me know your thoughts on the formatting in relation to the router output. Thanks!

### DIFF
--- a/test/phoenix/router/console_formatter_test.exs
+++ b/test/phoenix/router/console_formatter_test.exs
@@ -57,7 +57,38 @@ defmodule Phoenix.Router.ConsoleFormatterTest do
     """
   end
 
-  defp draw(router) do
-    ConsoleFormatter.format(router)
+  describe "endpoint sockets" do
+
+    def __sockets__, do: []
+
+    defmodule FormatterEndpoint do
+      use Phoenix.Endpoint, otp_app: :phoenix
+
+      socket "/socket", TestSocket, websocket: true
+    end
+
+    test "format with sockets" do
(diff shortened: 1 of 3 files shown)


test/phoenix/router/console_formatter_test.exs
```diff
@@ -57,7 +57,40 @@ defmodule Phoenix.Router.ConsoleFormatterTest do
...
+      socket "/socket", TestSocket, websocket: true
+    end
+
+    test "format with sockets" do
+      assert draw(RouterTestSingleRoutes, FormatterEndpoint) == """
+              page_path  GET     /        Phoenix.PageController :index
+      upload_image_path  POST    /images  Phoenix.ImageController :upload
+      remove_image_path  DELETE  /images  Phoenix.ImageController :delete
+
+      Phoenix.Router.ConsoleFormatterTest.FormatterEndpoint sockets:
```
chrismcg: This line would make it harder for tools to parse the route output. E.g an editor that wants to autocomplete or show routes. Maybe the method column could be SOCKET or WS instead?

Gazler: I don't have any objections to adding `WS` for websockets. Long polling actually uses both `GET` and `POST`, maybe we could include both routes?

PR #3258, opened 2019-01-25
Fix typespec of socket.id

linges:
The typespec of Phoenix.Socket lists `id` as `nil`. It should be `String.t` if I see that correctly.


chrismccord:
it defaults to nil, so can you update it to `String.t() | nil`? Thanks!

linges:
Done!

chrismccord:
❤️❤️❤️🐥🔥

PR #3262, opened 2019-01-27
Add Elixir Slack link in the help column of the initial default page

pietrofxq:
Considering that the Elixir community on Slack has now more than 20k members, it's one of the best places to find help. Adding it to Phoenix initial default page should bring even more members to the community.


chrismccord:
❤️❤️❤️🐥🔥

PR #3264, opened 2019-01-29
Allow the app configuration to specify a custom migration module for the generated migration code.

xpdf:
If the current app configuration specifies a custom migration module the generated migration code will use that rather than the default
  Ecto.Migration
     config :ecto_sql, migration_module: MyApplication.CustomMigrationModule

cf.  https://github.com/elixir-ecto/ecto_sql/pull/76

I think I have all the bases covered, if not please advise and I will correct.

Thank you!

### DIFF
--- a/lib/mix/phoenix/schema.ex
+++ b/lib/mix/phoenix/schema.ex
@@ -77,7 +77,6 @@ defmodule Mix.Phoenix.Schema do
     web_path = web_namespace && Phoenix.Naming.underscore(web_namespace)
     embedded? = Keyword.get(opts, :embedded, false)
     generate? = Keyword.get(opts, :schema, true)
-    migration_module = migration_module()
 
     singular =
       module
@@ -125,7 +124,7 @@ defmodule Mix.Phoenix.Schema do
       sample_id: sample_id(opts),
       context_app: ctx_app,
       generate?: generate?,
-      migration_module: migration_module}
+      migration_module: migration_module()}
   end
 
   @doc """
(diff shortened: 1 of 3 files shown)


xpdf:
Hi,

21 days - I do not want to lose this change - what do I need to do to get this merged into master.  I will eat lima beans if I have to.

Thank you.

lib/mix/phoenix/schema.ex
```diff
@@ -343,4 +346,19 @@ defmodule Mix.Phoenix.Schema do
     |> String.trim_leading("_")
     |> String.replace("/", "_")
   end
+
+@doc """
```
josevalim: We shouldn't include docs for private functions.

Also, can please align the definition below with [the one in Ecto.SQL here](https://github.com/elixir-ecto/ecto_sql/blob/d3fc4a8cc03ef601c7158d3cbdb2b477e22e7719/lib/mix/tasks/ecto.gen.migration.ex#L103-L108)?

xpdf: Corrected, thank you.

lib/mix/phoenix/schema.ex
```diff
@@ -76,6 +77,7 @@ defmodule Mix.Phoenix.Schema do
     web_path = web_namespace && Phoenix.Naming.underscore(web_namespace)
     embedded? = Keyword.get(opts, :embedded, false)
     generate? = Keyword.get(opts, :schema, true)
+    migration_module = migration_module()
```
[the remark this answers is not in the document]
xpdf: Corrected, thank you.

test/mix/tasks/phx.gen.schema_test.exs
```diff
@@ -268,6 +268,22 @@ defmodule Mix.Tasks.Phx.Gen.SchemaTest do
         end
       end
     end
+
+    in_tmp_project "uses defaults from generators configuration (binary_id) with Migration Module set", fn ->
```
fertapric: Would it make sense to move this to a different test? This doesn't seem to be related to the `:generators` configuration 🙂 


```elixir
test "generates migrations with a custom migration module", config do
  in_tmp_project config.test, fn ->
    try do
      Application.put_env(:ecto_sql, :migration_module, MyCustomApp.MigrationModule)

      Gen.Schema.run(~w(Blog.Post posts))        

      assert [migration] = Path.wildcard("priv/repo/migrations/*_create_posts.exs")

      assert_file migration, fn file ->
        assert file =~ "use MyCustomApp.MigrationModule"
        assert file =~ "create table(:posts) do"
      end
    after
      Application.delete_env(:ecto_sql, :migration_module)
    end
  end
end
```

I've also moved the cleanup of the Application env to an `after` clause; that way, if `Gem.Schema.run/1` fails, other tests don't get affected.

xpdf: I have updated the tests to reflect your points.  Thank you!

chrismccord:
❤️❤️❤️🐥🔥

PR #3265, opened 2019-01-31
Phoenix.Token.verify/4 will always return true when provided a max_age lower or equal to 0

tiagonbotelho:
Because of the comparison being done in ms if assertions with 0
can sometimes lead to unexpected results

Example:

```
token = Phoenix.Token.sign(Endpoint, @salt, user_id)

assert {:error, :expired} = Phoenix.Token.verify(Endpoint, @salt, token, max_age: 0)
```

Even though this test should always pass, given that the expiration of the token is right now, it can sometimes fail giving it a transient behaviour.

### DIFF
--- a/lib/phoenix/token.ex
+++ b/lib/phoenix/token.ex
@@ -233,6 +233,8 @@ defmodule Phoenix.Token do
     false
   end
 
+  defp expired?(signed, max_age_secs) when max_age_secs <= 0, do: true
+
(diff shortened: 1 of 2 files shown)


tiagonbotelho:
Any news on this? 😊

chrismccord:
❤️❤️❤️🐥🔥

PR #3267, opened 2019-02-05
Inspect body in ConnTest.response/2

qcam:
I would like to propose to print body in `ConnTest.response/2` in `inspect` format.

I think this would be useful and would avoid the test case from crashing when the response body is not printable.


ggVGc:
This seems like a very obvious improvement to me.

chrismccord:
❤️❤️❤️🐥🔥

PR #3269, opened 2019-02-09
update documentation to reflect on function deprecations

1334:
the documentation on `phoenix/lib/phoenix/test/conn_test.ex` was using the deprecated method `conn/0` instead of `build_conn/0`, also in one example was giving the wrong arity


Gazler:
Thanks <3

PR #3270, opened 2019-02-12
Reword sentence in Controllers guide

lasfter:
The semi-colon and "however" are unnecessary, since "If we don't find what we're looking for" and "we might find what we're looking for in Plug" don't disagree.


Gazler:
Thanks <3

PR #3277, opened 2019-02-13
update guide view snippet to agree with latest phx.new

joaofnds:
Just a minor difference that I've noticed while reading the docs, but I think it's worth updating it nonetheless


Gazler:
Thanks <3

PR #3280, opened 2019-02-16
Add elixir_buildpack.config to Heroku documentation

mcelaney:
Added elixir_buildpack.config to the Heroku deployment documentation to prevent new users from getting tripped up by the Elixir 1.5.0 default on Heroku

This is the [forum post](https://elixirforum.com/t/ecto-migration-fails-on-heroku/18112/4) where I figured out the issue when I ran in to it while following the docs most recently.


Gazler:
Thanks <3

PR #3286, opened 2019-02-20
Update heroku.md include erlang_version instructions

petermm:
encourage setting both Elixir and Erlang version in elixir_buildpack.config

additionally I noticed information on using config_vars_to_export which is deprecated https://github.com/HashNuke/heroku-buildpack-elixir#migrating-from-previous-build-pack - and thus removed that from the guide.


Gazler:
Thanks <3

PR #3287, opened 2019-02-20
Add typescript definition to phoenix

princemaple:
So typescript users like me are happy...

FYI there is https://www.npmjs.com/package/@types/phoenix, but it's outdated and a pain to PR.

Feel free to close if it's out of scope / not of interest.


princemaple:
Well looks like npm search index cache is outdated. https://www.npmjs.com/package/@types/phoenix was updated to 1.4 a few days ago.

PR #3289, opened 2019-02-22
Store template names in module attribute

LostKobrakai:
I'm working on an extension of phoenix views to allow for automatically generated component libraries and I need the compiled templates at compile time to further work with them like storing their content for display or differenciating other files in the folder from templates.

### DIFF
--- a/lib/phoenix/template.ex
+++ b/lib/phoenix/template.ex
@@ -164,6 +164,8 @@ defmodule Phoenix.Template do
     codes = Enum.map(pairs, &elem(&1, 1))
 
     quote do
+      @phoenix_templates unquote(names)
+      
       unquote(codes)
 
       # Catch-all clause for template rendering.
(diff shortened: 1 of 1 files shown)


chrismccord:
Can you describe more about what you are doing with the information? The attribute is private so this alone won't enable a proper extension point. Also note that at runtime you have the info you need in the assigns so I need to hear more about your greater approach to say where the best extension point lies. Thanks!

LostKobrakai:
One thing is compiling the template contents into the module, so the raw template can be shown in e.g. a web UI. I don't see how accessing the information at runtime would allow me to do that.

chrismccord:
The compiled function is the raw template contents, no? ie calling `render("my_template.html", %{...})` is the compiled contents that you call to splay the web UI. I'm not following what information is missing that you need at a lower level?

LostKobrakai:
I not only want to show the template results, but also the html template with the eex tags in it. So the source file of the template.

chrismccord:
Ah I see. Note that we injection `my_view.__templates__()` which gives  you what you need, I believe?https://github.com/phoenixframework/phoenix/blob//lib/phoenix/template.ex#L182-L188

LostKobrakai:
Can I call that before the module is fully compiled? At runtime the source files might no longer be available.

LostKobrakai:
I just noticed I can simply use `Phoenix.Template.find_all` and the root I know about 🤐

PR #3290, opened 2019-02-24
Update project tagline to be more clear

GRardB:
This is a super nitpick, but [my friend](https://github.com/jalcine) dared me to open a PR, so here it is.

The project tagline currently reads:

> A productive web framework that does not compromise speed and maintainability.

While this may technically be true, it translates to the following logic:

```
(doesCompromiseSpeed && doesCompromiseMaintainability) === false
```

The issue with this is that since only one of the conditions needs to be `false`, it could be interpreted along the lines of:

> Other frameworks compromise both speed and maintainability, but Phoenix only compromises one of them.

Thus, I think it'd be better to change the "and" to an "or," so that it's clear that Phoenix does not compromise either speed _or_ maintainability. In other words:

```
(doesCompromiseSpeed || doesCompromiseMaintainability) === false
```

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,6 +1,6 @@
 ![phoenix logo](https://raw.githubusercontent.com/phoenixframework/phoenix/master/priv/static/phoenix.png)
 > ### Productive. Reliable. Fast.
-> A productive web framework that does not compromise speed and maintainability.
+> A productive web framework that does not compromise speed or maintainability.
 
 [![Build Status](https://api.travis-ci.org/phoenixframework/phoenix.svg?branch=master)](https://travis-ci.org/phoenixframework/phoenix)
 [![Inline docs](http://inch-ci.org/github/phoenixframework/phoenix.svg)](http://inch-ci.org/github/phoenixframework/phoenix)
--- a/guides/views.md
+++ b/guides/views.md
@@ -75,7 +75,7 @@ Let's open up the `lib/hello_web/templates/page/index.html.eex` and locate this 
 ```html
 <section class="phx-hero">
   <h1><%= gettext "Welcome to %{name}!", name: "Phoenix" %></h1>
-  <p>A productive web framework that<br/>does not compromise speed and maintainability.</p>
+  <p>A productive web framework that<br/>does not compromise speed or maintainability.</p>
 </section>
 ```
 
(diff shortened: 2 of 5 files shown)


achempion:
> While this may technically be true, it translates to the following logic

Could you please elaborate why it wouldn't be

```elixir
(doesNotCompromiseSpeed && doesNotCompromiseMaintainability) === true
```

GRardB:
Sure, yeah! While I don't have specific resources to point out the exact rules of English around this, perhaps this will drive the point home.

**Explanation**

>A productive web framework that does not compromise speed _**and**_ maintainability.

Because of the way that English gets parsed, this ends up being:

>A productive web framework that does **not** (compromise speed _**and**_ maintainability).

Which translates to `!(doesCompromiseSpeed && doesCompromiseMaintainability)`.

If we wanted to say that neither speed nor maintainability are compromised, you would need to say something along the lines of the following:

>A productive web framework that does not compromise speed _**or**_ maintainability.
>A productive web framework that does not compromise speed, **_nor does it compromise_** maintainability.
>A productive web framework that does not compromise speed **_and also does not compromise_** maintainability.

That last one is a bit awkward, but I was trying to figure out how to bend the sentence so that we could use "and."

I think the easiest way to understand all of this is that the first example _definitely_ means that Phoenix doesn't compromise either speed or maintainability. We know that "and" and "or" have different meanings, so switching from "and" to "or" _must_ change the interpretation of the sentence. This is almost like a proof by contradiction 😁 If that doesn't help, perhaps the next example will.

**Thinking in terms of a question**

Pretend that you're in a courtroom being questioned by a lawyer, and they ask the following:

>Did you go to the store _and_ the park on March 1st?

However, you only went to the store. The way this gets interpreted is `didGoToStore && didGoToPark`. If the latter is `false`, then you would respond with:

>No[, I only went to the store].

If the lawyer used "or" instead of "and," you would respond with:

>Yes[, I went to the store].

Similarly, pretend the lawyer asks:

>Does the Phoenix framework compromise speed _and_ maintainability?

The following answers would work:

>No[, it only compromises speed].
>No[, it only compromises maintainability].
>No[, it does not compromise speed _or_ maintainability].
>Yes

Answering "Yes" would mean that both are compromised.

If the question used "or" instead of "and," then you would have the following options:

>Yes[, it compromises both speed and maintainability]
>Yes[, it compromises speed]
>Yes[, it compromises maintainability]
>No

Answering "No" would mean that neither are compromised.

Sorry for the long comment! If that didn't help, I can try to find something more formal around English rules.

mpranjic:
You are overthinking it :)

GRardB:
I wouldn't say I am. For a native English speaker, the sentence doesn't convey the intended message correctly. Not only that, but we don't often structure sentences in the form: `[Something] [negated verb] [noun] [and] [noun]`, and it's because of the ambiguity.

This isn't something that a native speaker needs to think much about; it sounds awkward immediately upon hearing/reading it, especially since it's so uncommon. In fact, a lot of people – when hearing a statement structured like the current tagline – will jokingly respond with: "So, does it compromise _one_ of them?"

chrismccord:
❤️❤️❤️🐥🔥

PR #3293, opened 2019-03-01
Prevent duplicate channel instantiation and joins

### LINKED ISSUE #3171: phoenix.js should close channel when there is a duplicate join without waiting for the server
Otherwise we can have race conditions in the client as two channels are running at the same time.

dballona:
Fixes #3171

### DIFF
--- a/assets/test/channel_test.js
+++ b/assets/test/channel_test.js
@@ -109,11 +109,29 @@ describe("join", () => {
     channel.join()
     assert.equal(socket.channels.length, 1)
 
-    // Move it back to not joined state to simulate a race condition
+    // Move joinedOnce back to false to simulate race condition
     channel.joinedOnce = false;
-    channel.join()
+    assert.throws(() => channel.join(), /^Error: tried to join multiple times/)
+  })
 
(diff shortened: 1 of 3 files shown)


assets/test/channel_test.js
```diff
@@ -105,6 +105,17 @@ describe("join", () => {
     assert.throws(() => channel.join(), /^Error: tried to join multiple times/)
   })
 
+  it("prevents duplicate join when race condition", () => {
+    channel.join()
```
achempion: shall we check for `assert.equal(socket.channels.length, 0)`?

dballona: sure! I'll add a separate test case for this assertion.

dballona: I just realized this spec is actually not impacting this implementation at all as a `channel` is only pushed to the socket `channels` on initialization... `channel.join` has zero impact on that : )

Also, I realized we're not handling the case for duplicate join properly – working on it!

dballona:
I just realized the multiple join implementation I did doesn't make any sense. 
Will close the PR again until I get a reasonable solution... sorry about this.

PR #3295, opened 2019-03-02
Prevent duplicate channel instantiation in phoenix.js

### LINKED ISSUE #3171: phoenix.js should close channel when there is a duplicate join without waiting for the server

dballona:
Fixes #3171 

This PR covers the following scenarios:

1 - When using `socket.channel("topic")` twice, returning the channel itself, gracefully, as [mentioned here](https://github.com/phoenixframework/phoenix/issues/3171#issuecomment-445841298).

2 - When trying to instantiate a channel within the same socket for the same topic, (e.g. `new Channel`), raising an error to prevent duplicate binding and blocking a duplicate channel on client-side as [requested here](https://github.com/phoenixframework/phoenix/issues/3171#issuecomment-445843240)

In this case, a duplicate join is no longer possible as `channel.joinedOnce` prevents a channel from being joined twice, while topic "2" covers cases where two different channels with the same topic are instantiated.

Hope this helps! :-)

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -362,7 +362,7 @@ class Push {
  */
 export class Channel {
   constructor(topic, params, socket) {
-    if (socket.channels.find((c) => c.topic === topic)) {
+    if (socket.channels.find(c => c.topic === topic)) {
       throw new Error(`tried to instantiate the same channel twice. Try using a different topic`)
     }
 
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -362,6 +362,10 @@ class Push {
  */
 export class Channel {
   constructor(topic, params, socket) {
+    if (socket.channels.find((c) => c.topic === topic)) {
+      throw new Error(`tried to instantiate the same channel twice. Try using a different topic`)
+    }
+
     this.state       = CHANNEL_STATES.closed
     this.topic       = topic
     this.params      = closure(params || {})
@@ -920,8 +924,13 @@ export class Socket {
    * @returns {Channel}
    */
   channel(topic, chanParams = {}){
-    let chan = new Channel(topic, chanParams, this)
-    this.channels.push(chan)
+    let chan = this.channels.find(c => c.topic === topic);
+
+    if(!chan){
+      chan = new Channel(topic, chanParams, this)
+      this.channels.push(chan)
+    }
+
     return chan
   }
 
--- a/assets/test/channel_test.js
+++ b/assets/test/channel_test.js
@@ -13,7 +13,7 @@ const defaultTimeout = 10000
 
 describe("constructor", () => {
   beforeEach(() => {
-    socket = { timeout: 1234 }
+    socket = new Socket("/socket", { timeout: 1234 })
   })
 
   it("sets defaults", () => {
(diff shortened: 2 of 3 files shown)


assets/test/socket_test.js
```diff
@@ -391,6 +391,19 @@ describe("channel", () => {
...
+  it("doesn't add duplicated channel if already exists", () => {
+    assert.equal(socket.channels.length, 0)
+
+    channel = socket.channel("topic", {one: "two"})
+    assert.equal(socket.channels.length, 1)
+
+    channel = socket.channel("topic", {two: "three"})
+    assert.equal(socket.channels.length, 1)
+
+     const [foundChannel] = socket.channels
```
achempion: looks like a redundant space

dballona: Fixed @achempion :-)
Thanks for the review!

assets/test/channel_test.js
```diff
@@ -49,11 +49,15 @@ describe("constructor", () => {
     assert.equal(joinPush.timeout, 1234)
   })
 
+  it("throws if instantiating the same channel twice", () => {
+    socket.channel("topic", {one: "two"})
```
achempion: Can't figure out why `socket.channel("topic", {one: "two"})` is throwing the error while 
```js
channel = socket.channel("topic", {one: "two"})
assert.equal(socket.channels.length, 1)

channel = socket.channel("topic", {two: "three"})
assert.equal(socket.channels.length, 1)
````

is not

File: assets/js/phoenix.js [hunk 3]
```diff
@@ -362,6 +362,10 @@ class Push {
  */
 export class Channel {
   constructor(topic, params, socket) {
+    if (socket.channels.find((c) => c.topic === topic)) {
```
achempion: inconsistent formatting with
- `find((c) => c.topic === topic))`
- `find(c => c.topic === topic)`

assets/test/channel_test.js
dballona: `socket.channel(topic)` pushes a new channel into `socket.channels` for the first time it's called and then if called twice, will just pick up the channel with that specific topic and return. We just return the current channel to prevent throwing an exception which we can avoid throwing.

Although if for some reason, either a race condition or a manual instantiation of a new channel happens (through `new Channel()` for instance), we need a safeguard to prevent that channel from being created as it would lead to double-bindings, double-joins, etc.

We could just raise an error if a channel is instantiated twice for all scenarios but I feel like returning the channel gracefully for the `socket.channel(topic)` use-case – which is probably 99% of the use cases – is more user-friendly and less prone to issues.

assets/js/phoenix.js
dballona: Fixed!

assets/test/channel_test.js
dballona: Let me know your thoughts @achempion : )
Also, I sent a proposal to standardize this on the Phoenix mailing list.

achempion: > We could just raise an error if a channel is instantiated twice for all scenarios but I feel like returning the channel gracefully ...

```js
it("throws if instantiating the same channel twice", () => {
```
Is it the case where we should return the old channel instead of throwing an error? 
Because of instantiating the same channel twice.

achempion: Looks like the exception will raise only with some sort of race conditions, maybe we can mention that in test description to be more precise. Still can't understand how we can simulate race conditions with test environment :)

dballona: > Looks like the exception will raise only with some sort of race conditions
That or if the person manually instantiates a new instance of channel through `new Channel` – which I am not even sure if it's applicable use, honestly : )

> Is it the case where we should return the old channel instead of throwing an error?
I think the error, in this case, is more idiomatic as it prevents misuse (e.g.: raising an error when the person is instantiating the same channel twice?)

Summarizing what I tried to accomplish:

1) if you're using `socket.channel('topic')` you will always get a channel being returned. If you call it once it will return a `new Channel()` and for the second time it will pick it up given a certain topic.

2) if for some reason a channel with the same topic is instantiated twice, it will raise an error. This is to prevent any possible race condition on the front end while instantiating the new channel.

3) the `join` race condition is already sorted by `joinedOnce` flag.

dballona:
Hi all, are there any adjustments needed on this one? Let me know if I can help or if something is off!

PR #3298, opened 2019-03-03
resources/4 accepts alias option but not listed in the doc

Joe-noh:
I found that `resources/4` accepts `:alias` option and [tested here](https://github.com/phoenixframework/phoenix/blob/2c586b7d45e0d7fa5cb1b640dc2b624b04afbff9/test/phoenix/router/resources_test.exs#L48) but not mentioned in the document. Please tell me if I'm missing something.


chrismccord:
Thanks! Unfortunately this is not a supported feature. If you think the docs lead you astray let me know.

PR #3299, opened 2019-03-04
Fix documentation generated in prod.exs

andrewmurph:
When running `phx.new` with the `--umbrella` flag, the comments generated in the web app's `/config/prod.exs` have a small bug in that they refer to the `:app_name` instead of `:web_app_name`.

E.g. 
`awesome_umbrella/apps/awesome_web/config/prod.exs` has:
`config :awesome_web, AwesomeWeb.Endpoint,`
yet the comments in that file have 3 references to
`#     config :awesome, AwesomeWeb.Endpoint,`

For non-umbrella projects, the output is correct as `app_name` and `web_app_name` have the same value, however I still think it's more correct to use `web_app_name`.

Hope this is helpful! 😃

### DIFF
--- a/installer/templates/phx_single/config/prod.exs
+++ b/installer/templates/phx_single/config/prod.exs
@@ -22,7 +22,7 @@ config :logger, level: :info
 # To get SSL working, you will need to add the `https` key
 # to the previous section and set your `:url` port to 443:
 #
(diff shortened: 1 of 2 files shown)


chrismccord:
❤️❤️❤️🐥🔥

PR #3300, opened 2019-03-04
Instrument controller calls with Telemetry

arkgil:
This is a first attempt at integrating Telemetry with Phoenix. There are two events emitted, before and after the controller is called, to make it consistent with the existing instrumenters API. The `stop` event carries the total time it took to execute the pipeline. Both events include the same metadata, controller name, action, and conn. 

I have a few questions I hope more experienced contributors will help me answer 😄 

**Endpoint:** is it safe to assume that endpoint is always present in the controller's conn? I thought that it might be quite useful to include it in the event metadata.

**Implementation:** I'm not sure if it's fine to keep it as is, or maybe it would be cleaner to implement it as an instrumenter? 🤔 

**Time measurement name:** I used generic "time", although I'm not sure I like it. Other candidates I thought about are "latency", "response_time", or "processing_time", although none of them is technically correct because more code is invoked before the pipeline is called.

Tell me what you think about it! 🙂 

#### TODO

- [ ] docs

/cc @bryannaegele @binaryseed

### DIFF
--- a/lib/phoenix/controller/pipeline.ex
+++ b/lib/phoenix/controller/pipeline.ex
@@ -24,7 +24,7 @@ defmodule Phoenix.Controller.Pipeline do
                       |> Map.put(:phoenix_action, action))
 
         start = System.monotonic_time()
-        :telemetry.execute([:phoenix, :controller, :call, :start], %{}, %{
+        :telemetry.execute([:phoenix, :controller, :call, :start], %{time: System.system_time()}, %{
           conn: conn,
           controller: __MODULE__,
           action: action
--- a/lib/phoenix/controller/pipeline.ex
+++ b/lib/phoenix/controller/pipeline.ex
@@ -35,7 +35,7 @@ defmodule Phoenix.Controller.Pipeline do
         end
         stop = System.monotonic_time()
         :telemetry.execute([:phoenix, :controller, :call, :stop], %{
-          time: stop - start
+          duration: stop - start
         }, %{
           conn: conn,
           controller: __MODULE__,
--- a/lib/phoenix/controller/pipeline.ex
+++ b/lib/phoenix/controller/pipeline.ex
@@ -23,10 +23,25 @@ defmodule Phoenix.Controller.Pipeline do
                  &(&1 |> Map.put(:phoenix_controller, __MODULE__)
                       |> Map.put(:phoenix_action, action))
 
-        Phoenix.Endpoint.instrument conn, :phoenix_controller_call,
+        start = System.monotonic_time()
+        :telemetry.execute([:phoenix, :controller, :call, :start], %{}, %{
+          conn: conn,
+          controller: __MODULE__,
+          action: action
+        })
+        conn = Phoenix.Endpoint.instrument conn, :phoenix_controller_call,
           %{conn: conn, log_level: @phoenix_log_level}, fn ->
           phoenix_controller_pipeline(conn, action)
         end
+        stop = System.monotonic_time()
+        :telemetry.execute([:phoenix, :controller, :call, :stop], %{
+          time: stop - start
+        }, %{
+          conn: conn,
+          controller: __MODULE__,
+          action: action
+        })
+        conn
       end
 
       @doc false
--- a/mix.exs
+++ b/mix.exs
@@ -49,6 +49,7 @@ defmodule Phoenix.MixProject do
     [
       {:plug, "~> 1.7"},
       {:phoenix_pubsub, "~> 1.1"},
+      {:telemetry, "~> 0.4.0"},
 
       # Optional deps
       {:plug_cowboy, "~> 1.0 or ~> 2.0", optional: true},
--- a/mix.lock
+++ b/mix.lock
@@ -17,5 +17,6 @@
   "plug_crypto": {:hex, :plug_crypto, "1.0.0", "18e49317d3fa343f24620ed22795ec29d4a5e602d52d1513ccea0b07d8ea7d4d", [:mix], [], "hexpm"},
   "poison": {:hex, :poison, "3.1.0", "d9eb636610e096f86f25d9a46f35a9facac35609a7591b3be3326e99a0484665", [:mix], []},
   "ranch": {:hex, :ranch, "1.6.2", "6db93c78f411ee033dbb18ba8234c5574883acb9a75af0fb90a9b82ea46afa00", [:rebar3], [], "hexpm"},
+  "telemetry": {:hex, :telemetry, "0.4.0", "8339bee3fa8b91cb84d14c2935f8ecf399ccd87301ad6da6b71c09553834b2ab", [:rebar3], [], "hexpm"},
   "websocket_client": {:git, "https://github.com/jeremyong/websocket_client.git", "9a6f65d05ebf2725d62fb19262b21f1805a59fbf", []},
 }
--- a/test/phoenix/controller/pipeline_test.exs
+++ b/test/phoenix/controller/pipeline_test.exs
@@ -202,6 +202,28 @@ defmodule Phoenix.Controller.PipelineTest do
     end
   end
 
+  test "emits telemetry event before and after controller call" do
+    :telemetry.attach(:start, [:phoenix, :controller, :call, :start], fn _, measurements, metadata, _ ->
+      send(self(), {:event, :start, measurements, metadata})
+    end, nil)
+    :telemetry.attach(:stop, [:phoenix, :controller, :call, :stop], fn _, measurements, metadata, _ ->
+      send(self(), {:event, :stop, measurements, metadata})
+    end, nil)
+
+    MyController.call(stack_conn(), :show)
+
+    assert_receive {:event, :start, measurements, metadata}
+    assert measurements == %{}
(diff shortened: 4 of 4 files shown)


arkgil:
This behaviour could also be optional (either opt-in or opt-out) - this way the `:telemetry` dependency could be marked as optional as well.

chrismccord:
This is great! The `endpoint` can assumed to always be there once you're in the controller stack, so no worries there. The big thing we need to figure out is how we want to integrate/deprecate the phoenix instrumentation in favor of telemetry. I think we could push the telemetry hooks down into the instrumentation for backwards compatibility, but let me touch base with @josevalim on how we want to go about it long term since telemetry should obviate our existing instrumentation. Thanks!

lib/phoenix/controller/pipeline.ex
```diff
@@ -23,10 +23,25 @@ defmodule Phoenix.Controller.Pipeline do
                  &(&1 |> Map.put(:phoenix_controller, __MODULE__)
                       |> Map.put(:phoenix_action, action))
 
-        Phoenix.Endpoint.instrument conn, :phoenix_controller_call,
+        start = System.monotonic_time()
+        :telemetry.execute([:phoenix, :controller, :call, :start], %{}, %{
```
binaryseed: It'd be helpful to have an explicit `start_time` (non-monotonic) in this measurement

lib/phoenix/controller/pipeline.ex
```diff
@@ -23,10 +23,25 @@ defmodule Phoenix.Controller.Pipeline do
...
+          controller: __MODULE__,
+          action: action
+        })
+        conn = Phoenix.Endpoint.instrument conn, :phoenix_controller_call,
           %{conn: conn, log_level: @phoenix_log_level}, fn ->
           phoenix_controller_pipeline(conn, action)
         end
+        stop = System.monotonic_time()
+        :telemetry.execute([:phoenix, :controller, :call, :stop], %{
+          time: stop - start
```
binaryseed: `duration` makes sense to me as the name of this measurement..

lib/phoenix/controller/pipeline.ex
```diff
@@ -23,10 +23,25 @@ defmodule Phoenix.Controller.Pipeline do
...
           phoenix_controller_pipeline(conn, action)
         end
+        stop = System.monotonic_time()
+        :telemetry.execute([:phoenix, :controller, :call, :stop], %{
+          duration: stop - start
+        }, %{
+          conn: conn,
+          controller: __MODULE__,
+          action: action
+        })
```
bryannaegele: I'd like to see the monotonic start + stop times in the meta or measurements.

bryannaegele:
@GregMefford - your input is welcome

bryannaegele:
>The big thing we need to figure out is how we want to integrate/deprecate the phoenix instrumentation in favor of telemetry.

I'd like to see that happen. What all events could we add that currently aren't exposed by the instrumentation hooks?

lib/phoenix/controller/pipeline.ex
```diff
@@ -23,10 +23,25 @@ defmodule Phoenix.Controller.Pipeline do
...
+        :telemetry.execute([:phoenix, :controller, :call, :start], %{}, %{
+          conn: conn,
+          controller: __MODULE__,
+          action: action
+        })
+        conn = Phoenix.Endpoint.instrument conn, :phoenix_controller_call,
           %{conn: conn, log_level: @phoenix_log_level}, fn ->
           phoenix_controller_pipeline(conn, action)
         end
+        stop = System.monotonic_time()
```
GregMefford: Maybe these should specify the unit so that we can more clearly enforce the unit that the `duration` will have.

```suggestion
        stop = System.monotonic_time(:microsecond)
```

GregMefford:
Thanks for the heads-up @bryannaegele!

I think some of my comments in [the matching PR for Plug](https://github.com/elixir-plug/plug/pull/824#issuecomment-469949756) apply here as well, so check that out if you're following along at home. 🍿

This looks good to me and I think it'll work with Spandex pretty similarly to how it currently works using the `instrumenters` API. I agree with @chrismccord that it might make sense to put this stuff down inside the intrumentation code (rather than wrapping around it), so that ultimately the Telemetry stuff will live on and the legacy instrumenters can be removed in a future version.

lib/phoenix/controller/pipeline.ex
arkgil: Maybe the unit should be configurable? I thought that native unit makes the most sense, since anyone can easily convert it using the `:erlang.convert_time_unit`, but it is not that good from the perspective of docs, examples etc.

arkgil: + telling the users to *always* convert the time to a desired unit is not that great from usability perspective :D

lib/phoenix/controller/pipeline.ex
arkgil: I have nothing against it, but I'm interested in what is the use case 🧐

lib/phoenix/controller/pipeline.ex
tsloughter: But will Phoenix end users deal with this conversion or the telemetry attachments that they include for like publishing to prometheus?

I would think that it would mostly be the authors of latter and definitely go with `native`. If that isn't expected then yea, native may be annoying/confusing to a user of Phoenix.

arkgil: I'd say that it's the users of Phoenix. In case of Prometheus it's quite simple, because it supports only one time unit, from what I remember. But if you're sending e.g. to StatsD, you may choose whatever unit you want.

arkgil: Let's go with native for now, and we can reiterate later.

arkgil:
>  I think we could push the telemetry hooks down into the instrumentation for backwards compatibility, but let me touch base with @josevalim on how we want to go about it long

@chrismccord did you agree on the approach here? I think that and the docs are the only things left to tackle in this PR. By the way, what would be the best place to put the docs for events?

lib/phoenix/controller/pipeline.ex
```diff
@@ -23,10 +23,25 @@ defmodule Phoenix.Controller.Pipeline do
...
+        :telemetry.execute([:phoenix, :controller, :call, :stop], %{
```
josevalim: @binaryseed would you use the controller call stop for anything? This time is most likely not relevant since we already measure the whole plug pipeline.

lib/phoenix/controller/pipeline.ex
josevalim: Agreed on native.

josevalim:
@chrismccord we will go with the approach as in this PR. 👍

chrismccord:
@arkgil sorry for the delay here as I've been wrangling things on the LiveView side :) 
I will sync with José about getting this in place. Thanks!

lib/phoenix/controller/pipeline.ex
binaryseed: As long as I can get access to the `conn` from the plug `:stop` event (to get info like `conn.status`), I don't think this is strictly necessary

Will the Phoenix pipeline always contain `Plug.Telemetry`? What will it's event prefix be?

josevalim: @binaryseed my plan is to plug it with `[:phoenix, :endpoint]` and that will always give you when the response is set.

tsloughter: @josevalim now I can't tell if I need the `stop` or not, hehe. What does "plug it with [:phoenix, :endpoint]" mean?

If spans are to be created in an event that is triggered at the start of a controller then there needs to be an event to finish the span when it finishes.

josevalim: @tsloughter we will emit events for the whole plug pipeline. That would be one span. We will also emit an event for the controller, but we were planning to not create a stop. But if you want a span, then we need the stop. I guess we will include the stop because it at leasts helps differenciate between events happening in controllers compared to other parts of the pipeline.

With that said, this PR is then good to be merged as is. :+1:

tsloughter: Ah. Yea, the issue with a whole plug pipline start/stop is not being able to use the route as the name of the span, since that isn't matched yet, right?

josevalim: Correct. The router comes quite late in the pipeline.

The other idea is to dispatch this event in the router as soon as the route is known. The difference though is that any plug in the router will then also be included as part of the measurements. This can be positive, as business logic starts to be invoked already in the router in many cases.

@chrismccord, any preferences?

tsloughter: Ok. It sounds like it will be similar to what we already do with `opencensus_plug` and `opencensus_phoenix`, but better since the plug part won't have to be manually put at the beginning of the sequence.

Basically a user enabling just the phoenix controller trace gets a span named for the route only covering the controller. But a user could also enable the trace to start from the initial plug handling, making the controller span a child of that initial span. This would allow for spans to be created as children in any of the plugs but the user has to either be ok with the path being used for the span name or defining a function that converts it to the span name.

josevalim:
We will move this to the router when we support telemetry. PR incoming soon.

PR #3302, opened 2019-03-05
Added null reference check on resetHeartbeat() on phoenix.js

KSubedi:
While running phoenix.js on react-native, we were getting a few of these in production:

```
#4 TypeError: null is not an object (evaluating 'this.conn.skipHeartbeat')
```

Adding a null reference check before evaluating that fixed the issue.


chrismccord:
❤️❤️❤️🐥🔥

PR #3307, opened 2019-03-08
phoenix.js: Document that passing params to connect() is deprecated

arjan:
From the [documentation](https://hexdocs.pm/phoenix/js/index.html#socketconnect) it was not clear that passing params into `Socket.connect()` is deprecated, only in runtime this will show on the JS console.


chrismccord:
❤️❤️❤️🐥🔥

PR #3315, opened 2019-03-13
Update css-loader to ^2.1.1

ricardoruwer:
I was having a problem in my Phoenix project, that the `@import`'s of my CSS file was not following the given order, so updating `css-loader` to the latest version just fixed it.


chrismccord:
❤️❤️❤️🐥🔥

PR #3318, opened 2019-03-16
reads slightly more natural

homanchou:
Minor thing.  I was expecting "as" while reading this sentence.  But maybe that's just me.


homanchou:
meh... nevermind I think it reads fine the second time through

PR #3324, opened 2019-03-20
utilize [name] in webpack.config.js

petermm:
spent hours in webpack land as I wanted to output a separate css file.. :/

using [name] is step 1. - that enables one to have multiple entries and thus outputs..

I propose a doc PR later perhaps something like this
```
  entry: {
    'app': ['./js/app.js'].concat(glob.sync('./vendor/**/*.js'))
    // above will output app.js and app.css(if any css is imported in the js files)
    // in the appropriate priv/static folders (js/css)
    //
    // add entry if you want to build more separate js/css files eg:
    //
    // 'app': ['./js/app.js'].concat(glob.sync('./vendor/**/*.js')),
    // 'example': ['./js/other.js']
    //
    // which additionally will output example.js and example.css(if any css is imported in the js files)
  },

```
though I'm uncertain of how verbose/detailed/specific we want to be..

### DIFF
--- a/installer/templates/phx_assets/webpack/webpack.config.js
+++ b/installer/templates/phx_assets/webpack/webpack.config.js
@@ -13,10 +13,10 @@ module.exports = (env, options) => ({
     ]
   },
   entry: {
-    './js/app.js': ['./js/app.js'].concat(glob.sync('./vendor/**/*.js'))
+    'app': ['./js/app.js'].concat(glob.sync('./vendor/**/*.js'))
   },
   output: {
-    filename: 'app.js',
+    filename: '[name].js',
     path: path.resolve(__dirname, '../priv/static/js')
   },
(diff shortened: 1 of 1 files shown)


Gazler:
This is documented in the webpack docs https://webpack.js.org/plugins/mini-css-extract-plugin/#extracting-css-based-on-entry

Given that webpack is a complicated bundler, we try to optimize for the most simple route, given that the config file is generated, users are welcome to modify their webpack config. Perhaps we could point people to the webpack docs so that they know where to look.

chrismccord:
As Gary said, we include the simplest defaults out of the box, and the rest is up to the user relying on webpack resources. If you think a commented out link in the webpack.config.js to the webpack docs would help, please send a doc PR. thanks!

PR #3325, opened 2019-03-21
Add mix npm.install task to phx_new generators and documentation.

TheFirstAvenger:
mix npm.install calls npm install from the assets directory via the --prefix arg.
Umbrella version commented out with example of running in multiple child apps.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -12,3 +12,5 @@
 /installer/deps
 
 erl_crash.dump
+
+*.ez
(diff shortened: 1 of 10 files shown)


chrismccord:
Thanks Mike! After thinking about this, I don't think the indirection is worth it for most users, in the context of folks coming into the platform. Our coupling with npm is very minimal today, and `"cd assets && npm install" is a very minor operation, so I don't think it's a beg enough win in the generators. Of course, folks are free to include the task in their project if it's a quality of life improvement. Thanks!

PR #3333, opened 2019-03-27
Updated with a more simple approach

cgorshing:
I felt like this doc could be simplified a bit for those who possibly have a more simplistic or conventional setup for their application and pushing to Heroku. So I've made some changes, open for discussion of course. Overall, I felt like having the buildpacks explain themselves might be better so we don't have to worry about decay in these docs as the buildpacks change.

Explained to the reader using the latest published version of the buildpack and if needed to use more nightly/edge, then consult the docs.

In some example output there was a line I removed .... "Run `git push heroku master` to create a new release using these buildpacks." When adding the heroku-buildpack-phoenix-static buildpack as I feel like it can cause confusion to beginners. Upon seeing this, someone might think this is a valid command to run, but will error instead at this point.

Removed the git push heroku master ... even though it shows in the buildpacks output, if you run it here, then it will error. Maybe even highlight to not run that command.

I've also removed all the references of setting your application name in the url host field. I've tried plugs and haven't found a case when it is generated bad. The paths are relative paths.

### DIFF
--- a/guides/deployment/heroku.md
+++ b/guides/deployment/heroku.md
@@ -75,17 +75,17 @@ Now that we have the Toolbelt installed, let's create the Heroku application. In
 
 > Note: the first time we use a Heroku command, it may prompt us to log in. If this happens, just enter the email and password you specified during signup.
 
+We are going to use latest published buildpack, you can also get an edge version of this buildpack. See it's [README](https://github.com/HashNuke/heroku-buildpack-elixir) for more instructions on how to do this.
 ```console
-$ heroku create --buildpack "https://github.com/HashNuke/heroku-buildpack-elixir.git"
-Creating mysterious-meadow-6277... done, stack is cedar-14
-Buildpack set. Next release on mysterious-meadow-6277 will use https://github.com/HashNuke/heroku-buildpack-elixir.git.
+$ heroku create --buildpack hashnuke/elixir
+Creating app... done, ⬢ mysterious-meadow-6277
+Setting buildpack to hashnuke/elixir... done
 https://mysterious-meadow-6277.herokuapp.com/ | https://git.heroku.com/mysterious-meadow-6277.git
-Git remote heroku added
 ```
 
 > Note: the name of the Heroku application is the random string after "Creating" in the output above (mysterious-meadow-6277). This will be unique, so expect to see a different name from "mysterious-meadow-6277".
 
-The `--buildpack` option we are passing allows us to specify the [Elixir buildpack](https://github.com/HashNuke/heroku-buildpack-elixir) we want Heroku to use. A [buildpack](https://devcenter.heroku.com/articles/buildpacks) is a convenient way of packaging framework and/or runtime support. In our case it's installing Erlang, Elixir, fetching our application dependencies, and so on, before we run it.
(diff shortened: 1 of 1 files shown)


josevalim:
@cgorshing is there a way to figure out the latest version of the buildpack on Heroku? IIRC the reason why we are using master is because there are some fixes in there that have not been published yet.

cgorshing:
> is there a way to figure out the latest version of the buildpack on Heroku?

I have not been able to find a definitive answer to that yet as `$ heroku buildpacks:info hashnuke/elixir` does not list out any version information nor does https://elements.heroku.com/buildpacks/hashnuke/heroku-buildpack-elixir

However, on 2018-08-13 the buildpack's README was updated to use the Heroku Registry. The history of Phoenix's heroku doc has that line in question added 2018-04-17 (initial commit).

I have no doubts HEAD of that buildpack will contain updates, but for the Phoenix Heroku guide, I think it should work for a simple case, to get people going. Let them touch their toe in the water before jumping all in :)

...... all of that assumes there isn't a show-stopper bug ... which I haven't yet encounter (done this with 3 projects now).

josevalim:
Oh, the issue I had in mind was about the old build pack using an outdated Elixir, but since now we recommend people to add the elixir_versions file, we should be good. :+1:

guides/deployment/heroku.md
```diff
@@ -171,17 +145,21 @@ Now, let's tell Phoenix to use our Heroku URL and enforce we only use the SSL ve
 url: [host: "example.com", port: 80],
 ```
 
-... and replace it with this (don't forget to replace `mysterious-meadow-6277` with your application name):
+... and replace it with this:
 
 ```elixir
-url: [scheme: "https", host: "mysterious-meadow-6277.herokuapp.com", port: 443],
+url: [scheme: "https", host: "", port: 443],
```
josevalim: Can the host be actually empty? I think you have to fill it in.

guides/deployment/heroku.md
```diff
@@ -230,11 +208,7 @@ end
 
 This ensures that any idle connections are closed by Phoenix before they reach Heroku's 55-second timeout window.
 
-Lastly, we'll need to create a [Procfile](https://devcenter.heroku.com/articles/procfile) (a text file called "Procfile" in the root of our project’s folder) with the following line:
-
-```
-web: MIX_ENV=prod elixir --sname server -S mix phx.server
-```
+As is common with Heroku's buildpacks, you can create your own "Procfile", but for our case we will use the default one provided by the [heroku-buildpack-phoenix-static buildpack](https://github.com/gjaldon/heroku-buildpack-phoenix-static).
```
josevalim: We are adding two build packs and both have a default procfile. How does Heroku chooses which one to keep?

guides/deployment/heroku.md
cgorshing: Ah yes - my fault. I only checked with Routes.page_path and not page_url. Correction coming ....

guides/deployment/heroku.md
cgorshing: I added [hopefully] a clarifying sentence for this scenario. See the commit below.

josevalim:

PR #3334, opened 2019-03-28
Incorporate some of LiveView steps into Phoenix

josevalim:
Although this means Phoenix knows a bit about LiveView,
it is justified as LiveView is shaping to became an important
asset in the Elixir community and it will help with its getting
started experience.


josevalim:

PR #3335, opened 2019-03-28
Colorize webpack output

markevich:
Hi!

This is a simple addition that enables colorized output for `Webpack` by default.

Screenshot `after` change for the newly generated project.

image

I am sorry but I’m not quite familiar with the process of working with Phoenix framework versions.
Could you help me and point files/locations where I should write a message in the release notes and how can I add a message to the upgrade guide for existing Phoenix apps.

Thank you!

### DIFF
--- a/guides/errors.md
+++ b/guides/errors.md
@@ -35,7 +35,7 @@ config :hello, HelloWeb.Endpoint,
   debug_errors: false,
   code_reloader: true,
   cache_static_lookup: false,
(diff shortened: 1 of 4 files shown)


markevich:
Why I do it:

When working with a complex front end in Phoenix, it is very difficult to notice that something went wrong when assembling assets. 
The highlight provides some visual assistance in detecting such problems.

With colorized output, you will definitely notice the "red" lines and realize that something went wrong.

image

josevalim:
I think we had the --color on but it messes everything up on Windows. :( So the recommendation is to add `--color` by hand if you prefer it.

Issue #3340, opened 2019-04-01
Phoenix js client doesn't retry connection when the connection has been explicitly closed before


markmeeus:
### Environment

* Phoenix version (mix deps): 1.4.3
* NodeJS version (node -v): 8.9.3
* NPM version (npm -v):5.8.0
* Operating system: MacOs

We are using Phoenix Channels in a webapp which is also wrapped in a native Android and iOS App using WebViews.
On Android, websockets stay alive in the webview even if the app is backgrounded and we need to close the sockets when this happens, reconnecting when the app is foregrounded again.
Not closing the websockets means consuming bandwidth and battery.

What we basically need is to 'pause' and 'resume' the websockets, where resuming means reconnecting and rejoining all open channels.

Up until 1.4.3, we could call the connect function on the socket and rejoinUntillConnected on every channel in the socket.
Our code looks like this.

```javascript 
  sockets.forEach(({socket}) => {
    const connectionState = socket.connectionState();
    if(connectionState !== 'open' && connectionState !== 'connecting'){
      socket.connect();
      socket.channels.forEach(c => c.rejoinUntilConnected());
    }
  });
```
As of 1.4.3 there are 2 issues with this code:
1. => Connect only tries to connect 2 times. If the connection fails (no wifi/4G for instance) it stops trying to reconnect.
2 => rejoinUntilConnected is removed, rejoin seems to work, although I'm not sure if it would have been rejoined if the socket reconnected after a while. (Currently obscured by point 1.)

I can't seem to find a way to get it working again in 1.4.3 without adding a lot of socket state management in our code which would make us even more dependent on the inner workings of phoenix code.

Is there a way to achieve the behaviour?
If not, I'd be happy to look into 'pause/resume' functionality on the socket class.

chrismccord:
`rejoinUntillConnected` was a private API. Instead, you should be able to hook into your foreground and background events, then call `socket.disconnect()` and `socket.reconnect()` ? That should be sufficient to reconnect and rejoin all channels. Is there a reason that doesn't work for you? Thanks!

markmeeus:
Hi Chris, 

Thank you for the quick response!

Simply calling ```socket.connect()``` re-establishes the websocket connection, but it looks like the channels aren't joined again. Calling ```channel.rejoin()``` does the trick.
However, calling ```socket.connect()``` while the network is offline doesn't enter the retry loop.
(1.4.2 socket does enter the retry loop)

[...]

PR #3342, opened 2019-04-03
Fix reconnect retry after explicit disconnect #3340

markmeeus:
The ```closeWasClean``` variable was to to false in the Socket's constructor and set to true in ```disconnect```. When the next call to ```connect``` fails, it still considers the close a 'clean close'.
This PR fixes this by setting the ```closeWasClean``` in the ```connect``` call.


markmeeus:
travis complains about the compiled asset not being correct. Not sure how to fix this.

chrismccord:
❤️❤️❤️🐥🔥

Issue #3343, opened 2019-04-03
Use send_download without encode_www_form


AllanKlaus:
It's is not a bug, but I think it's a feature. I was creating a project that enable users to download files with a large variety of names, that user define. When the final user download a a file with the name like `my custom report.xlsx` when download execute it return a file like `my+custom+report.xlsx` and user need to edit it to download with the correct name. 

Instead of use `send_download(conn, kind, opts)` I used in my project the code below

```
conn
|> put_resp_content_type(content_type, "utf-8")
|> put_resp_header("content-disposition", ~s[attachment; filename="#{filename}"])
|> send_file(conn.status || 200, report, 0, :all)
```

### Expected behavior
When download the file `my custom report.xlsx` the file name that appear to download is `my custom report.xlsx`.

### Actual behavior
When download the file `my custom report.xlsx` the file name that appear to download is `my+custom+report.xlsx`

I created the PR to solve it  #3344

josevalim:
Closing n favor of the OR.

PR #3344, opened 2019-04-03
Enable to use send_download without encode

AllanKlaus:
- Motivation
I was creating a system the enable users to download files with a
variety of names the put spaces and other characters. When downloading
this kind of file names like `my custom report.xlsx` was returning
`my+custom+report.xlsx`. Esthetically it is ugly to users and they need to edit it to download with the correct nam

- Solution
I created the possibility to send option `encode`, the default value of
the option is nil, and it only avoid the encode_www_form if it was
passed as `false`

- Issue: #3343

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -884,7 +884,7 @@ defmodule Phoenix.Controller do
       Defaults to none
     * `:offset` - the bytes to offset when reading. Defaults to `0`
     * `:length` - the total bytes to read. Defaults to `:all`
-    * `:encode` - encode filename using `URI.encode_www_form/1`.
+    * `:encode` - encodes the filename using `URI.encode_www_form/1`.
       Defaults to `true`. When `false`, disables encoding. If you
       disable encoding, you need to guarantee there are no special
       characters in the filename, such as quotes, newlines, etc.
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -884,7 +884,9 @@ defmodule Phoenix.Controller do
       Defaults to none
     * `:offset` - the bytes to offset when reading. Defaults to `0`
     * `:length` - the total bytes to read. Defaults to `:all`
-    * `:encode` - encode filename using URI.encode_www_form. Defaults to `nil`
+    * `:encode` - encode filename using URI.encode_www_form.
+      Defaults to `nil`. Any value different of `false` will apply
+      the encode
 
   ## Examples
 
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -885,8 +885,10 @@ defmodule Phoenix.Controller do
     * `:offset` - the bytes to offset when reading. Defaults to `0`
     * `:length` - the total bytes to read. Defaults to `:all`
     * `:encode` - encode filename using `URI.encode_www_form/1`.
-      Defaults to `true`. When `false`, disables encoding. Disable
-      encoding may cause security implications.
+      Defaults to `true`. When `false`, disables encoding. If you
+      disable encoding, you need to guarantee there are no special
+      characters in the filename, such as quotes, newlines, etc.
+      Otherwise you can expose your application to security attacks
 
   ## Examples
 
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -884,9 +884,9 @@ defmodule Phoenix.Controller do
       Defaults to none
     * `:offset` - the bytes to offset when reading. Defaults to `0`
     * `:length` - the total bytes to read. Defaults to `:all`
-    * `:encode` - encode filename using URI.encode_www_form.
-      Defaults to `nil`. Any value different of `false` will apply
-      the encode
+    * `:encode` - encode filename using `URI.encode_www_form/1`.
+      Defaults to `true`. When `false`, disables encoding. Disable
+      encoding may cause security implications.
 
   ## Examples
 
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -934,7 +934,7 @@ defmodule Phoenix.Controller do
   end
 
   defp encode_filename(filename, encode) when encode == false, do: filename
-  defp encode_filename(filename, encode), do: URI.encode_www_form(filename)
+  defp encode_filename(filename, _encode), do: URI.encode_www_form(filename)
 
   defp ajax?(conn) do
     case get_req_header(conn, "x-requested-with") do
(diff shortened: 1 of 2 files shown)


michalmuskala:
Wouldn't that be a potential [request splitting][1] vulnerability if the filename is not encoded and potentially controlled by the user?

[1]: https://www.owasp.org/index.php/HTTP_Response_Splitting

AllanKlaus:
> Wouldn't that be a potential [request splitting](https://www.owasp.org/index.php/HTTP_Response_Splitting) vulnerability if the filename is not encoded and potentially controlled by the user?

I don't think so.

I tryed to do the attack inside my code like this

```
my_report = "./my_report.xlsx"
filename = my_report <> "\"\r\n%0d%0aConnection:%20Keep-Alive%0d%0aContent-Length:%200%0d%0a%0d%0aHTTP/1.0%20200%20OK%0d%0aContent-Type:%20text/html%0a%0aContent-Length:%2020%0d%0a%0d%0a<html>YOU_ARE_DONE!</html>"

    conn
    |> put_resp_content_type(content_type, "utf-8")
    |> put_resp_header("content-disposition", ~s[attachment; filename="#{filename}"])
    |> send_file(conn.status || 200, my_report, 0, :all)
```

I tryed using `\r\n` and using`%0d%0a`. Tryed to close the quote of file name using a escaped quote and nothing interfered on the request.

josevalim:
Oh, I commented in the wrong place and I missed this discussion.

Even if we are safe request splitting in this case (likely because cowboy rejects \r\n), I think it would be unsafe for Phoenix to assume something else is catching his back. Maybe in the future we are using another  server or Cowboy stops checking it, and now Phoenix has a large vulnerability.

So as mentioned in my comment above, I would first test if `URI.encode` is a better choice. In case it isn't, we can provide a verbose way to disable encoding, but it should most certainly be enabled  out of the box.

lib/phoenix/controller.ex
```diff
@@ -925,13 +926,16 @@ defmodule Phoenix.Controller do
...
-    encoded_filename = URI.encode_www_form(filename)
+    encoded_filename = encode_filename(filename, opts[:encode])
     warn_if_ajax(conn)
     conn
     |> put_resp_content_type(content_type, opts[:charset])
     |> put_resp_header("content-disposition", ~s[attachment; filename="#{encoded_filename}"])
   end
 
+  defp encode_filename(filename, encode) when encode == false, do: filename
+  defp encode_filename(filename, encode), do: URI.encode_www_form(filename)
```
webdeb: better to explicitly mark unused variables
`encode` -> `_encode`

AllanKlaus:
> Oh, I commented in the wrong place and I missed this discussion.
> 
> Even if we are safe request splitting in this case (likely because cowboy rejects \r\n), I think it would be unsafe for Phoenix to assume something else is catching his back. Maybe in the future we are using another server or Cowboy stops checking it, and now Phoenix has a large vulnerability.
> 
> So as mentioned in my comment above, I would first test if `URI.encode` is a better choice. In case it isn't, we can provide a verbose way to disable encoding, but it should most certainly be enabled out of the box.

@josevalim when the name of the file is **my report.xlsx** using `URI.encode` will return **my%20report.xlxs**. So it's discarted to use `URI.encode`.

To don't change the current behavior the default value to encode is `nil`, in this case it will always encode, only when `encode` is `false` it will don't apply the encode. Thinking in the cowboy i agree we need to try another approach. What do you mean in verbose way?

lib/phoenix/controller.ex
AllanKlaus: nice. I changed it.

lib/phoenix/controller.ex
```diff
@@ -884,6 +884,9 @@ defmodule Phoenix.Controller do
       Defaults to none
     * `:offset` - the bytes to offset when reading. Defaults to `0`
     * `:length` - the total bytes to read. Defaults to `:all`
+    * `:encode` - encode filename using URI.encode_www_form.
+      Defaults to `nil`. Any value different of `false` will apply
```
josevalim: I see what you mean. So I think the `nil` as default is confusing. The default should be `true`, we should always encode if `true`, unless set to `false`. We should also say that disabling  it may have security implications.

lib/phoenix/controller.ex
```diff
@@ -925,13 +928,16 @@ defmodule Phoenix.Controller do
 
   defp prepare_send_download(conn, filename, opts) do
     content_type = opts[:content_type] || MIME.from_path(filename)
-    encoded_filename = URI.encode_www_form(filename)
+    encoded_filename = encode_filename(filename, opts[:encode])
```
josevalim:
```suggestion
    encoded_filename = encode_filename(filename, Keyword.get(opts, :encode, true))
```

lib/phoenix/controller.ex
```diff
@@ -925,13 +928,16 @@ defmodule Phoenix.Controller do
...
     warn_if_ajax(conn)
     conn
     |> put_resp_content_type(content_type, opts[:charset])
     |> put_resp_header("content-disposition", ~s[attachment; filename="#{encoded_filename}"])
   end
 
+  defp encode_filename(filename, encode) when encode == false, do: filename
```
josevalim:
```suggestion
  defp encode_filename(filename, false), do: filename
```

lib/phoenix/controller.ex
```diff
@@ -925,13 +928,16 @@ defmodule Phoenix.Controller do
...
+  defp encode_filename(filename, _encode), do: URI.encode_www_form(filename)
```
josevalim:
```suggestion
  defp encode_filename(filename, true), do: URI.encode_www_form(filename)
```

lib/phoenix/controller.ex
```diff
@@ -884,6 +884,9 @@ defmodule Phoenix.Controller do
       Defaults to none
     * `:offset` - the bytes to offset when reading. Defaults to `0`
     * `:length` - the total bytes to read. Defaults to `:all`
+    * `:encode` - encode filename using URI.encode_www_form.
```
josevalim:
```suggestion
    * `:encode` - encode filename using `URI.encode_www_form/1`.
```

lib/phoenix/controller.ex
josevalim:
```suggestion
      Defaults to `true`. When `false`, disables encoding.
```

AllanKlaus:
@josevalim I applied the changes you suggested. I'm new using elixir, I learned a lot with this pr, thanks.

lib/phoenix/controller.ex
```diff
@@ -884,6 +884,11 @@ defmodule Phoenix.Controller do
       Defaults to none
     * `:offset` - the bytes to offset when reading. Defaults to `0`
     * `:length` - the total bytes to read. Defaults to `:all`
+    * `:encode` - encode filename using `URI.encode_www_form/1`.
```
fertapric: Nitpick:

```suggestion
    * `:encode` - encodes the filename using `URI.encode_www_form/1`.
```

AllanKlaus: changed.

josevalim:

PR #3345, opened 2019-04-03
Don't close an already closed connection in beforeunload

dmarkow:
The latest Phoenix.js update (1.4.3) added updates to "optimize reconnects when browser restores from back/forward cache by listening for window beforeunload."

However, if the socket had already been closed before unloading via `socket.disconnect()`, `this.conn` won't exist when the window unloads, throwing a `Cannot read property 'close' of null` error whenever the page is unloaded.

Checking that `this.conn` actually exists before flagging `unloaded` and marking an abnormal close fixes this. Was unsure how to add a test for the `beforeunload` event...


chrismccord:
❤️❤️❤️🐥🔥

inaniyants:
Just in order to simplify search of this problem:
while testing in tools like cypress the error is 
"Uncaught TypeError: Cannot read property 'close' of undefined"

Since in some cases cypress unloads page earlier, then there is a socket connection.

PR #3347, opened 2019-04-04
Fix put_secure_browser_headers/2 docs

sysashi:
Previously the list of headers in documentation was indented with 4 spaces which resulted in incorrect rendering.


josevalim:

PR #3352, opened 2019-04-05
Raise error when route name uses `static` in the prefix

### LINKED ISSUE #3346: Routing requests to a `StaticController` causes a compiler warning
### Environment

* Phoenix version (mix deps): 1.4.3
* Operating system: macOS 10.14.1

### Expected behavior

I'd expect to be able to route requests to a controller called `StaticController` without issues.

### Actual behavior

I get a compiler warning that looks like a route collision:
```elixir
warning: this clause cannot match because a previous clause at line 1 always matches
  lib/radio_web/router.ex:1
```

Repro steps:
- Generate a new Phoenix app: `mix phx.new radio --no-ecto --no-html --no-webpack`
- Add a line to the router that points a route to a `StaticController`, i.e. `get "/test", RadioWeb.StaticController, :test`
- Compile the app and you should see the warning

Other notes:
- Module nesting has no effect; any routing to `StaticController` will cause this
- Removing `plug Plug.Static` from `RadioWeb.Endpoint` also has no effect

I hope I've come to the right place with this. Of course it's an easy one to work around; I'm just wildly curious to learn what's happening. The closest I've come is discovering that `Phoenix.Endpoint` defines `static_path/1` and `static_url/0`, which seems like it _could_ conflict with route helpers...

Thanks for reading, and for the amazing work you all do!

alexgaribay:
Closes #3346

### DIFF
--- a/lib/phoenix/router/scope.ex
+++ b/lib/phoenix/router/scope.ex
@@ -27,6 +27,10 @@ defmodule Phoenix.Router.Scope do
     as      = Keyword.get(opts, :as, Phoenix.Naming.resource_name(plug, "Controller"))
     alias?  = Keyword.get(opts, :alias, true)
 
+    if to_string(as) == "static"  do
+      raise ArgumentError, "`static` is a reserved route prefix generated from #{inspect plug} or `:as` option"
+    end
+
     {path, host, alias, as, pipes, private, assigns} =
       join(module, path, plug, alias?, as, private, assigns)
     Phoenix.Router.Route.build(line, kind, verb, path, host, alias, plug_opts, as, pipes, private, assigns)
--- a/test/phoenix/router/scope_test.exs
+++ b/test/phoenix/router/scope_test.exs
@@ -205,4 +205,24 @@ defmodule Phoenix.Router.ScopedRoutingTest do
     assert conn.status == 200
     assert conn.resp_body == "api v1 users show"
   end
+
+  test "raises for reserved prefixes" do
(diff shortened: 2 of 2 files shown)


lib/phoenix/router/scope.ex
```diff
@@ -27,6 +27,10 @@ defmodule Phoenix.Router.Scope do
     as      = Keyword.get(opts, :as, Phoenix.Naming.resource_name(plug, "Controller"))
     alias?  = Keyword.get(opts, :alias, true)
 
+    if to_string(as) == "static"  do
+      raise ArgumentError, "`static` is a reserved prefix. Supply a different prefix to the route using the `:as` option."
```
chrismccord: Lets also show the controller/plug name they tried to use here. It will avoid confusion if folks aren't familiar with the named helper generation and they only passed `AppWeb.StaticController`.

test/phoenix/router/scope_test.exs
```diff
@@ -205,4 +205,29 @@ defmodule Phoenix.Router.ScopedRoutingTest do
     assert conn.status == 200
     assert conn.resp_body == "api v1 users show"
   end
+
+  test "raises for reserved prefixes" do
+    assert_raise ArgumentError, ~r/`static` is a reserved prefix/, fn ->
+      defmodule StaticController do
```
chrismccord: We don't need to create a real controller here as it's not required a compile time

alexgaribay:
Updated 👍.

josevalim:

PR #3353, opened 2019-04-05
Take paths into account for `%URI{}`'s passed to `put_router_url/2`

rmoorman:
Following #3181 (and the discussion with @josevalim), this PR is an attempt to re-use the path component from `%URI{}` structs set using the `put_router_url/2` helper function. Currently the path is nillified right before route urls are being generated even if a path is present inside the struct.

### DIFF
--- a/lib/phoenix/router/helpers.ex
+++ b/lib/phoenix/router/helpers.ex
@@ -27,7 +27,7 @@ defmodule Phoenix.Router.Helpers do
   end
 
   def url(_router, %URI{} = uri) do
-    URI.to_string(%URI{uri | path: nil})
+    URI.to_string(uri)
   end
 
   def url(_router, endpoint) when is_atom(endpoint) do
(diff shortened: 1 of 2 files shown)


rmoorman:
@josevalim , currently there is one failing test regarding a script name where the path is present twice within the resulting URL.

rmoorman:
@josevalim I just amended this PR with changes that attempt to correct the failing test by removing the path information within the generated route url helper functions (otherwise both, the url function and the route path helper function will generate a string with the path included)

Seems that all tests are green again. Only further remark I currently have is that I don't have an idea if forwarding is still working correctly as the path and connection's script path might have some interconnectedness I currently cannot fully grasp (due to limited hands-on experience with the forwarding functionality).

Please let me know if there is anything that would have to be changed/improved!

josevalim:
Thanks @rmoorman!

The issue is that we have two entry points: when the URI is explicitly given and another for put_router_url. I have decided to decouple those and stop skipping the path only on the put_router_url, that should fix it and be backwards compatible. A fix here: cd1ce956a294bf70a0882b20c0706a18ae570502

Thank you for leading  and investigating this!

rmoorman:
@josevalim thank you for your time and fixing this! It was my pleasure to have a look at the issue.

PR #3356, opened 2019-04-07
Update ex_doc

rodrigues:


rodrigues:
EDIT: oh, it seems elixir requirements for ex_doc are too high, given phoenix supports older versions. Closing…

wojtekmach:
I believe the patch is fine, both ExDoc v0.20.1 and v0.19.3 (that Phoenix master currently uses) require Elixir v1.7. Since dependency is defined as `{:ex_doc, ..., only: :docs}` it's OK if it uses different version than what Phoenix is targeting.

rodrigues:
ok, reopening then @wojtekmach 👍

josevalim:

PR #3359, opened 2019-04-09
update heroku.md

petermm:
make elixir_buildpack.config less "optional" eg. "can be done" vs "is done"


josevalim:

PR #3362, opened 2019-04-10
Update Heroku guide to help non-HTML apps

joladev:
Used the heroku guide recently and noticed references to files that were probably created in an earlier version of the guide, but that don't exist anymore, so removed them (Procfile and compile).

Also, if you're following the guide but skipping the static buildpack because you don't have HTML, you'll end up with a hard to understand error (Heroku stops your app because it doesn't attach to the assigned port). This is because the elixir buildpack uses `mix run --no-halt` and the configuration in the guide doesn't include `server: true`. This is fine if you're using the static buildpack, since it replaces the default Procfile with `mix phx.server` which adds it automatically.

Adding `server: true` just works for everyone. I'm not sure if I should add an explanation of why, or if that would just make it noisy.

### DIFF
--- a/guides/deployment/heroku.md
+++ b/guides/deployment/heroku.md
@@ -106,7 +106,7 @@ erlang_version=21.2.5
 
 We need to compile static assets for a successful Phoenix deployment. The [Phoenix static buildpack](https://github.com/gjaldon/heroku-buildpack-phoenix-static) can take care of that for us, so let's add it now.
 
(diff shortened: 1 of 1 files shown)


josevalim:
Good find!

My concern with adding `server: true` is that any command that you run that boots the app will now try to attach to the port. Having an explicit Procfile would address this too. So maybe we should suggest that, if you don't add the phoenix_static buildpack, you need to add a Procfile?

joladev:
You're right! Updated the text to instead suggest a Procfile if you're not using the static buildpack

josevalim:

Issue #1786, opened 2016-06-29
Provide send_attachment/3


justinbkay:
There doesn't seem to be a mechanism to send data to the browser as an attachment or inline as a file download.  The send_file function doesn't seem to allow the file to be sent outside the rendering of the layout.  The file renders as the pdf source in the page as opposed to sending the file down to the browser to be saved or opened.

aphillipo:
You just need to set the mime type correctly. 

https://hexdocs.pm/plug/Plug.Conn.html#put_resp_content_type/3

application/octet-stream is the right content-type for it to download in *most* browsers.

justinbkay:
That doesn't change anything for me.  It still renders the PDF source in the webpage:

![screen shot 2016-06-29 at 11 21 41 am](https://cloud.githubusercontent.com/assets/4094/16461827/b3619cce-3deb-11e6-9df5-5c21abfa078c.png)

put_resp_content_type(conn, "application/octet-stream", "utf-8")
send_file(conn, 200, attachment.path)

Gazler:
You need to bind the new `conn`

```elixir
conn = put_resp_content_type(conn, "application/octet-stream", "utf-8")
send_file(conn, 200, attachment.path)
```

Or:

```elixir
conn
|> put_resp_content_type("application/octet-stream", "utf-8")
|> send_file(200, attachment.path)
```

justinbkay:
Unfortunately that doesn't make a difference.  It does make it to the conn object:

{"content-type", "application/pdf; charset=utf-8"}],

josevalim:
You should also set content-disposition to attachment. We had plans to
include a function that handles some of it but I don't remember the
outcome. @chrismccord?

On Wednesday, June 29, 2016, Justin <notifications@github.com> wrote:

> Unfortunately that doesn't make a difference. It does make it to the conn
> object:
>
> {"content-type", "application/pdf; charset=utf-8"}],
>
> —
> You are receiving this because you are subscribed to this thread.
> Reply to this email directly, view it on GitHub
> <https://github.com/phoenixframework/phoenix/issues/1786#issuecomment-229429195>,
> or mute the thread
> <https://github.com/notifications/unsubscribe/AAAlbnFuKs0Qv6PMR6wO_h36mhhUEVAHks5qQqvqgaJpZM4JBUYd>
> .
>


-- 


*José Valim*
www.plataformatec.com.br
Skype: jv.ptec
Founder and Director of R&D

[...]

PR #3364, opened 2019-04-11
Support `:disposition_type` option in send_download/3

mnieborak:
Based on issue discussion:
https://github.com/phoenixframework/phoenix/issues/1786#issuecomment-481631844

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -894,6 +894,11 @@ defmodule Phoenix.Controller do
     * `:content_type` - the content type of the file or binary
       sent as download. It is automatically inferred from the
       filename extension
+    * `:disposition` - specifies dispositon type
+      (`:attachment` or `:inline`). If `:attachment` was used,
+      user will be prompted to save the file. If `:inline` was used,
+      the browser will attempt to open the file.
+      Defaults to `:attachment`.
     * `:charset` - the charset of the file, such as "utf-8".
       Defaults to none
     * `:offset` - the bytes to offset when reading. Defaults to `0`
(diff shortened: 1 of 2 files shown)


lib/phoenix/controller.ex
```diff
@@ -894,6 +894,11 @@ defmodule Phoenix.Controller do
     * `:content_type` - the content type of the file or binary
       sent as download. It is automatically inferred from the
       filename extension
+    * `:disposition_type` - specifies dispositon type
+      (`"attachment"` or `"inline"`). If `"attachment"` was used,
+      user will be prompted to save the file. If `"inline"` was used,
+      the browser will attempt to open the file.
+      Defaults to `"attachment"`.
```
josevalim: Let's make it an atom and let's assert it is either the atoms `:attachment` or `:inline`. That's because generally allowing strings can lead to vulnerability.

Also, I would say `:disposition` is good enough for the option name, no need for the `_type` suffix. Thoughts?

mnieborak: Yes, I was thinking that the value of this option should be somehow constrained. Any suggestions how to do it?
And yes, changing option name to just `:disposition` makes sense :)

josevalim: > Yes, I was thinking that the value of this option should be somehow constrained. Any suggestions how to do it?

Something like this:

```elixir
case Keyword.get(options, :disposition, :attachment) do
  :attachment -> "attachment"
  :inline -> "inline"
  other -> raise ArgumentError, "expected :disposition to be :attachment or :inline, got: #{inspect(other)}"
end
```

mnieborak: Thanks @josevalim
I'll do it later today.

hubertlepicki: The name `:disposition_type` was chosen to reflect the naming of this item in RFC describing Content-disposition header. https://tools.ietf.org/html/rfc6266#section-4.2l

RFC does not constrain possible values either, just says that anything other than "inline" or "attachment" should be handled as "attachment".

But I am okay with all the suggested changes, just explaining why the original parameter anme and choice of string was made.

mnieborak: @josevalim I've made changes you suggested.

josevalim:

9mm:
So is this not in phoenix release yet? I have tried adding disposition: :inline and it still downloads the file

Issue #3321, opened 2019-03-19
Pain points when instrumenting for distributed tracing


tsloughter:
I've been working on OpenCensus support for Phoenix projects at work and wanted to open an issue about the difficulties experienced to hopefully find solutions for future Phoenix releases to have easier instrumentation. 

* Wrapping request handling in a span:

There are at least 3 ways to insert span creation into the process of routing a request. A plug can be placed in the pipeline and use `register_before_send` for finishing the span. The `call` function can be overridden as Spandex does, https://github.com/spandex-project/spandex_phoenix/blob/master/lib/spandex_phoenix.ex#L151. Or a Phoenix Instrumenter can be used and create a span on the `start` event and finish on the `stop` event.

All of these have trade offs, two of which apply to all of them and are included below as overall issues. 

I'll note also that I am not a fan of span creation in event handlers like with an Instrumenter or Telemetry event. Doing so requires a guarantee that events are handled in the same process so that the pdict can be utilized to carry the span context. An interceptor that can wrap the actual handling of a request (but still have access to the route match) in a `try/catch` would be the optimal solution in my opinion but understand that might not be an option.

* Route name for span name:

Example: a request route `get "/user/:name"` should be able to result in an initial span with the name `"/user/:name"`.

Spandex relies on a hack to reverse engineer the route, https://github.com/spandex-project/spandex_phoenix/blob/master/lib/spandex_phoenix.ex#L255. And even with this hack it is not possible to name the initial span with it and they name that span always the same, `"request"`.

* Setting status after controller completes:

Currently the `stop` event of instrumenter does not receive the `conn` which makes it impossible to get the response status for marking a span's status. This may be fixed by https://github.com/phoenixframework/phoenix/pull/3300/files#diff-10541d7f803c04ad1cdbba70a70c0a55R40.

josevalim:
> There are at least 3 ways to insert span creation into the process of routing a request. A plug can be placed in the pipeline and use register_before_send for finishing the span. 

This is the approach we will take on v1.5 with Plug.Telemetry. I am not a fan of `try/catch` because it doesn't provide any meaningful extra guarantee compared to the Plug approach. You can still miss an event because of links. Plus Phoenix already checkpoints the connection in many `try/catch` and that helps guarantee that `register_before_send` is invoked.

I understand that telemetry may not be the best solution for tracing out of the box but we also want to unify how Phoenix emit events and telemetry is the best option for now.

> Route name for span name

I agree we should include this information.

[...]

PR #3366, opened 2019-04-12
Add a way to retrieve the request route

groyoh:
This PR adds an easy way in Phoenix to retrieve the path of the route against which a request was matched. Previously, if you for instance defined a `get "/users/:id"` route and you received a `GET /users/1` request, you'd have no easy solution to retrieve that the route was `/users/:id`. You’d only have access to `conn.request_path` which would be something like `/users/1` in such case.

With this PR, you can call `Phoenix.Router.match_path(conn)` to retrieve the route's path.

Accessing the route can be really useful in the following cases:

* `logging` - logging your requests by adding a route metadata in order to filter quite easily by route.
* `metrics` - tagging your request metrics with the route so that you could for example check the performance of your app per route.
* `rate_limiting` - creating a route-based rate limiting.

This feature is similar to a feature implemented in Plug (https://github.com/elixir-plug/plug/pull/630).

### DIFF
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -283,6 +283,14 @@ defmodule Phoenix.Router do
     end
   end
 
+  @doc """
+  Returns the path of the route that the request was matched to.
+  """
+  @spec match_path(Plug.Conn.t()) :: String.t() | nil
+  def match_path(%Plug.Conn{} = conn) do
+    Map.get(conn.private, :phoenix_route)
+  end
+
   defp match_dispatch() do
     quote location: :keep do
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @groyoh! This looks great!

We are planning to add similar information when we tackle #3321. We will add it explicitly as part of the telemetry event as we want to avoid attaching too much metadata to the connection. So we will go with a similar solution to yours but slightly different. More information soon, thank you!

groyoh:
Oh that's some great news! Thanks for the heads up. And please let me know if I can help in any way with that.

PR #3368, opened 2019-04-13
Fix value encoding in json generator controller test update method

### LINKED ISSUE #3367: Generated test for schema including map has a typo
### Environment

* Phoenix version (mix deps): 1.4.2
* NodeJS version (node -v): N/A
* NPM version (npm -v): N/A
* Operating system: macOS 10.14.4

### Expected behavior
```
mix phx.gen.json Blog Post posts data:map
```
Should generate working tests

### Actual behavior
The generated controller tests includes a typo in when representing an empty Map: `{}` instead of `%{}` in the `update` action.

```elixir
      assert %{
               "id" => id,
               "data" => {}
```

This causes the test to fail. Adding the `%` fixes the test.

The issue appears to stem from [here](https://github.com/phoenixframework/phoenix/blob/v1.4.3/priv/templates/phx.gen.json/controller_test.exs#L61), but I'm not sure what it should be instead.

timwis:
Fixes #3367

### DIFF
--- a/priv/templates/phx.gen.json/controller_test.exs
+++ b/priv/templates/phx.gen.json/controller_test.exs
@@ -58,7 +58,7 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
 
       assert %{
                "id" => id<%= for {key, val} <- schema.params.update do %>,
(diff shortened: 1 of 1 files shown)


josevalim:

PR #3370, opened 2019-04-13
Move context test to mirror placement in lib

jarsen:
When `phx.gen` creates a new context, it creates the context and puts it as a sibling to a directory by the same name containing the generated schema: `mix phx.gen.html Accounts User ...` generates both `lib/my_app/accounts.ex` and `lib/my_app/accounts/user.ex`.

The generated tests do not follow the same structure—the tests are both generated under `test/my_app/accounts/`.

The inconsistency for the Context between `lib` and `test` make it difficult to configure tools such as [projectionist.vim](https://github.com/tpope/vim-projectionist) which may rely on a mirrored structure to intuit which test maps to which file.

This PR attempts to move the context's test up a directory so that both `lib` and `test` are in sync.

### DIFF
--- a/guides/contexts.md
+++ b/guides/contexts.md
@@ -61,8 +61,8 @@ username:string:unique
 * creating priv/repo/migrations/20170629175236_create_users.exs
 * creating lib/hello/accounts.ex
 * injecting lib/hello/accounts.ex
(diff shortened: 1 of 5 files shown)


josevalim:

PR #3371, opened 2019-04-13
Refactor flash methods from Phoenix.Controller into Phoenix.Controller.Flash

aforward:
Based on a conversation here:

https://groups.google.com/forum/#!searchin/phoenix-core/liveview%7Csort:date/phoenix-core/61kVmWkF6QI/70BFY0zFCAAJ

The proposed step forward for getting `--live` (and eventually `--no-live`) integrated into phoenix directly.

```
1. phx.new --live flag (it will be opt-in on v1.5 while we figure things out but probably enabled by default on v1.6)
2. phx.new.live generators
3. Bring the functionality in LiveView.Flash to Phoenix so we can get rid of one plug
```

It was suggested to start with #3

Once released, then I can help remove that code from phoenix_live_view and update the documentation to remove the need for the second flash plug.

@josevalim if I have mis-understood, let me know.

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -1273,24 +1273,7 @@ defmodule Phoenix.Controller do
   @doc """
   Fetches the flash storage.
   """
-  def fetch_flash(conn, _opts \\ []) do
-    session_flash = get_session(conn, "phoenix_flash")
-    conn = persist_flash(conn, session_flash || %{})
-
-    register_before_send conn, fn conn ->
-      flash = conn.private.phoenix_flash
-      flash_size = map_size(flash)
-
-      cond do
-        is_nil(session_flash) and flash_size == 0 ->
-          conn
-        flash_size > 0 and conn.status in 300..308 ->
-          put_session(conn, "phoenix_flash", flash)
-        true ->
-          delete_session(conn, "phoenix_flash")
-      end
-    end
-  end
+  defdelegate fetch_flash(conn, opts \\ []), to: Phoenix.Controller.Flash
 
   @doc """
   Persists a value in flash.
@@ -1304,9 +1287,7 @@ defmodule Phoenix.Controller do
       "Welcome Back!"
 
   """
-  def put_flash(conn, key, message) do
-    persist_flash(conn, Map.put(get_flash(conn), flash_key(key), message))
-  end
+  defdelegate put_flash(conn, key, message), to: Phoenix.Controller.Flash
 
   @doc """
   Returns a map of previously set flash messages or an empty map.
@@ -1321,10 +1302,7 @@ defmodule Phoenix.Controller do
       %{"info" => "Welcome Back!"}
 
   """
-  def get_flash(conn) do
-    Map.get(conn.private, :phoenix_flash) ||
-      raise ArgumentError, message: "flash not fetched, call fetch_flash/2"
-  end
+  defdelegate get_flash(conn), to: Phoenix.Controller.Flash
 
   @doc """
   Returns a message from flash by `key`.
@@ -1336,9 +1314,7 @@ defmodule Phoenix.Controller do
       "Welcome Back!"
 
   """
-  def get_flash(conn, key) do
-    get_flash(conn)[flash_key(key)]
-  end
+  defdelegate get_flash(conn, key), to: Phoenix.Controller.Flash
 
   @doc """
   Generates a status message from the template name.
@@ -1364,16 +1340,7 @@ defmodule Phoenix.Controller do
   @doc """
   Clears all flash messages.
   """
-  def clear_flash(conn) do
-    persist_flash(conn, %{})
-  end
-
-  defp flash_key(binary) when is_binary(binary), do: binary
-  defp flash_key(atom) when is_atom(atom), do: Atom.to_string(atom)
-
-  defp persist_flash(conn, value) do
-    put_private(conn, :phoenix_flash, value)
-  end
+  defdelegate clear_flash(conn), to: Phoenix.Controller.Flash
 
   @doc """
   Returns the current request path with its default query parameters:
--- a/lib/phoenix/controller/flash.ex
+++ b/lib/phoenix/controller/flash.ex
@@ -12,8 +12,7 @@ defmodule Phoenix.Controller.Flash do
     conn = persist_flash(conn, found_flash || %{})
 
     register_before_send conn, fn conn ->
-      flash = conn.private[@session_atom]
-        || raise KeyError, key: @session_atom, term: conn.private
+      flash = Map.fetch!(conn.private, @session_atom)
       flash_size = map_size(flash)
 
       cond do
@@ -40,7 +39,7 @@ defmodule Phoenix.Controller.Flash do
   @doc false
   def get_flash(conn) do
     Map.get(conn.private, @session_atom) ||
-      raise ArgumentError, message: "flash not fetched, call fetch_flash/2"
+      raise ArgumentError, "flash not fetched, call fetch_flash/2"
   end
 
   def get_flash(conn, key) do
--- a/lib/phoenix/controller/flash.ex
+++ b/lib/phoenix/controller/flash.ex
@@ -1,0 +1,57 @@
+defmodule Phoenix.Controller.Flash do
+  import Plug.Conn
+
+  @moduledoc false
+
+  @session_key "phoenix_flash"
+  @session_atom :phoenix_flash
+
+  @doc false
+  def fetch_flash(conn, _opts \\ []) do
+    found_flash = get_session(conn, @session_key)
+    conn = persist_flash(conn, found_flash || %{})
+
+    register_before_send conn, fn conn ->
+      flash = conn.private[@session_atom]
+        || raise KeyError, key: @session_atom, term: conn.private
+      flash_size = map_size(flash)
+
+      cond do
+        is_nil(found_flash) and flash_size == 0 ->
+          conn
+        flash_size > 0 and conn.status in 300..308 ->
+          put_session(conn, @session_key, flash)
+        true ->
+          delete_session(conn, @session_key)
+      end
+    end
+  end
+
+  @doc false
+  def clear_flash(conn) do
+    persist_flash(conn, %{})
+  end
+
+  @doc false
+  def put_flash(conn, key, message) do
+    persist_flash(conn, Map.put(get_flash(conn), flash_key(key), message))
+  end
+
+  @doc false
+  def get_flash(conn) do
+    Map.get(conn.private, @session_atom) ||
+      raise ArgumentError, message: "flash not fetched, call fetch_flash/2"
+  end
+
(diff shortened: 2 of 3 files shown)


aforward:
In the examples,

I kept the code as-is and observed the flash working as expected

Screen Shot 2019-04-13 at 5 47 29 PM


Then I tried the fetch_flash as-is

```
    plug :fetch_flash
    # plug Phoenix.LiveView.Flash
```

The flash did not work (as expected)

Screen Shot 2019-04-13 at 5 46 36 PM

I am now testing this branch to get it working "as-is", which it isn't yet.  I will be updating the PR as I troubleshoot the behaviour for the merged code.

josevalim:
Hi @aforward!

What do you think if we break this PR in two? Can you please first send a PR that moves all of the Flash functions to a Flash module?

Then another PR can add the new feature, this would make the code easier to review. :) Thank you!

aforward:
There we go... `__phoenix_flash___` != `__phoenix_flash__`

aforward:
Yes @josevalim, will do

lib/phoenix/controller/flash.ex
```diff
@@ -0,0 +1,97 @@
...
+  end
+  defp computed_salt(salt_base), do: salt_base <> "flash"
+
+  @doc false
+  def sign_token(endpoint_mod, salt_base, %{} = flash) do
+    salt = computed_salt(salt_base)
+    Phoenix.Token.sign(endpoint_mod, salt, Phoenix.json_library().encode!(flash))
+  end
+
+  defp configured_signing_salt!(endpoint) when is_atom(endpoint) do
```
aforward: these two helper methods (`configured_signing_salt!/1` and `random_signing_salt`) were taken from `Phoenix.LiveView.View`

lib/phoenix/controller/flash.ex
```diff
@@ -0,0 +1,97 @@
+defmodule Phoenix.Controller.Flash do
```
aforward: Mostly now verbatim from `Phoenix.LiveView.Flash`.

My intention however, is that all flash capabilities move from the controller into this helper and it incorporates the session retrieval of the flash message.  The `fetch_flash` and other related methods will be then `defdelegate`s into here.

test/phoenix/controller/flash_test.exs
```diff
@@ -122,4 +123,37 @@ defmodule Phoenix.Controller.FlashTest do
...
+    assert get_flash(conn, :notice) == "hi"
+    assert get_flash(conn, "notice") == "hi"
+  end
+
+  test "fetch based on provided cookie" do
+    flash = Flash.sign_token(FlashEndpoint, "othercookie_salt456", %{info: "hello"})
+    conn = conn(flash_cookie: flash) |> Flash.call(signing_salt: "othercookie_salt456")
+    assert get_flash(conn, :info) == "hello"
+    assert get_flash(conn, "info") == "hello"
+  end
```
aforward: Two tests to show that if you have the same salt, then you can use the Flash plug to get your answer.

aforward:
Tested locally against the phoenix live view project

```
$ git diff lib/ mix.exs 
diff --git a/lib/demo_web/router.ex b/lib/demo_web/router.ex
index db8f7db..14e91e3 100644
--- a/lib/demo_web/router.ex
+++ b/lib/demo_web/router.ex
@@ -5,7 +5,8 @@ defmodule DemoWeb.Router do
   pipeline :browser do
     plug :accepts, ["html"]
     plug :fetch_session
-    plug Phoenix.LiveView.Flash
+    plug :fetch_flash
+    plug Phoenix.Controller.Flash
     plug :protect_from_forgery
     plug :put_secure_browser_headers
     plug :put_layout, {DemoWeb.LayoutView, :app}
diff --git a/mix.exs b/mix.exs
index c4d4fe8..8ee6873 100644
--- a/mix.exs
+++ b/mix.exs
@@ -33,7 +33,8 @@ defmodule Demo.Mixfile do
   # Type `mix help deps` for examples and options.
   defp deps do
     [
-      {:phoenix, github: "phoenixframework/phoenix", branch: "v1.4", override: true},
+      # {:phoenix, github: "aforward-oss/phoenix", branch: "f/liveview_flash_plug", override: true},
+      {:phoenix, path: "../../oss/phoenix", override: true},
       {:phoenix_live_view, github: "phoenixframework/phoenix_live_view"},
       {:plug_cowboy, "~> 2.0"},
       {:plug, "~> 1.7"},
```

And the flash messages are persisted as expected

Screen Shot 2019-04-14 at 8 49 42 AM

If turned off `# plug Phoenix.Controller.Flash` then the flash no longer works (as expected).

@josevalim let me know if this aligns with your thoughts on a first step.

lib/phoenix/controller/flash.ex
josevalim: Thanks @aforward!

What if we start in the opposite direction: first we create `Phoenix.Controller.Flash` without any of the liveview code and then we add the liveview code to it?

Also, regarding the signing salt, it would be nice if we could use the session signing salt. But we need to check if it is exposed in any way. I will investigate.

aforward: hehe, ok.

that's how I had structured the original commits.

josevalim: Oh, i forgot to check the commits individually! Sorry for the confusion then!

lib/phoenix/controller/flash.ex
```diff
@@ -0,0 +1,80 @@
...
+  import Plug.Conn
+
+  @moduledoc """
```
josevalim: We probably want to `@moduledoc false` this. :)

lib/phoenix/controller/flash.ex
aforward: @josevalim I have reverted to the original PR, but rolled back the last commit that incorporated LiveView functionality.

And not having a separate salt would be great, let me know.

lib/phoenix/controller.ex
```diff
@@ -1364,16 +1340,9 @@ defmodule Phoenix.Controller do
...
-  end
-
-  defp flash_key(binary) when is_binary(binary), do: binary
-  defp flash_key(atom) when is_atom(atom), do: Atom.to_string(atom)
+  defdelegate clear_flash(conn), to: Phoenix.Controller.Flash
 
-  defp persist_flash(conn, value) do
-    put_private(conn, :phoenix_flash, value)
-  end
+  defdelegate flash_key(binary_or_atom), to: Phoenix.Controller.Flash
```
josevalim: Why do we need to delegate the `flash_key` one? Could we just remove it from here?

lib/phoenix/controller/flash.ex
```diff
@@ -0,0 +1,80 @@
...
+  @session_key "phoenix_flash"
+  @session_atom :phoenix_flash
+
+  @doc """
+  Fetches the flash from server, or a signed message from the client
+
+  We will look for a flash message in the `session_key`, and if none
+  found then we will look in client cookie (under the name
+  `__session_key___`)
+  """
```
josevalim: We don't need to copy the docs, as everything else is a private API.

lib/phoenix/controller/flash.ex
aforward: Sidebar: I really ❤️ git, these back and forth have been a breeze with `git rebase -i HEAD~X`... I wanted to keep this PR conversation so the only unfortunate part is the branch name having to stay unchanged.

lib/phoenix/controller/flash.ex
```diff
@@ -0,0 +1,80 @@
...
+  @doc """
+  Clears all flash messages.
+  """
+  def clear_flash(conn) do
+    persist_flash(conn, %{})
+  end
+
+  @doc false
+  def flash_key(binary) when is_binary(binary), do: binary
+  def flash_key(atom) when is_atom(atom), do: Atom.to_string(atom)
```
josevalim: Hopefully this can be `defp` once we stop delegating.

lib/phoenix/controller/flash.ex
```diff
@@ -0,0 +1,80 @@
...
+  We will look for a flash message in the `session_key`, and if none
+  found then we will look in client cookie (under the name
+  `__session_key___`)
+  """
+  def fetch_flash(conn, _opts \\ []) do
+    found_flash = get_session(conn, @session_key)
+    conn = persist_flash(conn, found_flash || %{})
+
+    register_before_send conn, fn conn ->
+      flash = conn.private[@session_atom]
```
aforward: @josevalim the original code had `conn.private.phoenix_flash`; I changed to use a constant up above.  This will no longer throw an exception if not found; not sure if that tweak is desired or not.

josevalim: Undesired. I would prefer to raise as it would mean something definitely went wrong. :)

aforward: I added `|| raise KeyError, key: @session_atom, term: conn.private` so that we could keep the constants

aforward:
I have done making changes based on all comments above.  Let me know if there is more work to be done / comments to be made.

lib/phoenix/controller/flash.ex
aforward: I tested locally with

```
iex(1)> atom = :c
:c
iex(2)> map = %{b: 1}
%{b: 1}
iex(3)> map[atom] || raise KeyError, key: atom, term: map
** (KeyError) key :c not found in: %{b: 1}
```

josevalim: Then it is best to use `Map.fetch!(conn.private, @session_atom)`, same result, more concise.

josevalim:
So I have an idea on how to make the signing stuff work! We would make it work like CSRF tokens.

We will add three new functions:

  * `load_state` - loads the flash state
  * `dump_state` - dumps the flash state
  * `sign_flash` - signs the given flash

All of those work on a signing salt. The signing salt will be automatically generated whenever `dump_state` or `sign_flash` are  called and stored in the process dictionary. Then in `register_before_send` we will store the  value in pdict in the session. If there is already something in the session and it is exactly the same as pdict, then we don't need to do anything.

On every request, we will see if there is said signing salt in the session and, if so, we put it into pdict. If there is a signing salt, we will also check if there is a cookie. If there is, we go ahead, decode it, merge into the flash, and erase the cookie.

But let's implement all of this as a separate issue.

Then we will have to work on the LiveView integration, which can discuss once this done. :)

lib/phoenix/controller/flash.ex
aforward: It feels like you know this language inside and out 😜

josevalim:

josevalim:
Hi @aforward, just a heads up. We are planning to move the shared logic flash up to the session, so if you are working on finishing this PR, please halt a bit until we are sure about next steps.

aforward:
Thanks.  I am not working on this ATM, I have a separate experimental PR to add - live(view) flag on phx.new, which I will update once this work is finished (as I believe the second salt config will be removed)

Issue #2491, opened 2017-09-12
SRI (Sub resource Integrity) using cache_manifest.json


foeken:
Not a bug, but a discussion if this is something you would merge into Phoenix (and if so: how). 

Let's start with a small intro. SRI (or Sub resource Integrity: https://developer.mozilla.org/en-US/docs/Web/Security/Subresource_Integrity) is a new browser feature intended to make your website a bit safer especially if you use a CDN to serve static content. The idea is that you include a hash for the original script or css inside your script or style tag. The browser will compare the hash with the served file (from the CDN) and reject if there is a mismatch (ie if the CDN has served a tampered file).

The tag should look something like this (I used a script tag, but it also supports style tags):

```html
<script 
  src="/js/app-ca6fe8ab7cb2e1c4a46d511a2d4e877c.js?vsn=d" 
  integrity="sha512-ssOjV24ABrEBpcRVCvdippen4VFEKdhfQEvY1D66K4Wcr/9eTuoS0r8iBitvQymLQ4aYzdW2p4+KGtrd5e61Gg==">
</script>
```

This got me thinking. As Phoenix already supports the `cache_manifest.json` file for digested lookups, this could also be used to calculate the `sha512` hash. In that case we could create a method similar to `static_path/1`, say: `integrity_hash/1` which could be injected into the html.

Now, I just started looking at the `Digester` and it's a breeze to add the hash there. We simply add the following key to the `digests` key.

```elixir
integrity: "sha512-#{Base.encode64(:crypto.hash(:sha512, content))}"
```

But here is my first question: The method `cache_static_manifest/1` inside `supervisor.ex` states on line 333:

```elixir
# TODO v1.4: No longer support old manifests
manifest["latest"] || manifest
```

Is this to say the old style is the `digests` key? If so, adding the integrity key to the old-style would probably not be a good idea.

I just wanted to double-check and confirm or deny this, and get some input on what would be the best approach and if this would be of interest in Phoenix core before I got started.

chrismccord:
I would be 👍 for this change, and it would finally justify the existence of the `script_tag` helper function in the `phoenix_html` project since now we can do this by default. If this is something you would like to take on, please send a PR. With respect to the old style, we only need to support the current one.

foeken:
I will pick it up this weekend. Just as a final check: is it safe to add the integrity hash into the `digests` key of the manifest or is that the 'old' way?

chrismccord:
[...]

PR #3372, opened 2019-04-14
feat(SRI): Add Routes.integrity/1 for subresource integrity validatio…

tomciopp:
…n of static assets

@chrismccord @josevalim Based off of the discussion in #2491, I decided to take a crack at adding a method of handling sub-resource integrity for static assets. This still needs testing but let me know what you think.

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -696,13 +696,16 @@ defmodule Phoenix.Endpoint do
       def static_path(path) do
         Phoenix.Config.cache(__MODULE__, :__phoenix_static__,
                              &Phoenix.Endpoint.Supervisor.static_path/1) <>
-        Phoenix.Config.cache(__MODULE__, {:__phoenix_static__, path},
-                             &Phoenix.Endpoint.Supervisor.static_path(&1, path))
+        elem(static_lookup(path), 0)
       end
 
-      def integrity(path) do
-        Phoenix.Config.cache(__MODULE__, {:__phoenix_integrity__, path},
-                             &Phoenix.Endpoint.Supervisor.integrity(&1, path))
+      def static_integrity(path) do
+        elem(static_lookup(path), 1)
+      end
+
+      def static_lookup(path) do
+        Phoenix.Config.cache(__MODULE__, {:__phoenix_static__, path},
+                             &Phoenix.Endpoint.Supervisor.static_lookup(&1, path))
       end
     end
   end
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -699,6 +699,10 @@ defmodule Phoenix.Endpoint do
         elem(static_lookup(path), 0)
       end
 
+      @doc """
+      Generates a base64-encoded cryptographic hash (sha512) to a static file
+      in `priv/static`. Meant to be used for Subresource Integrity with CDNs.
+      """
       def static_integrity(path) do
         elem(static_lookup(path), 1)
       end
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -707,6 +707,10 @@ defmodule Phoenix.Endpoint do
         elem(static_lookup(path), 1)
       end
 
+      @doc """
+      Returns a two item tuple with the first item being the static_path
+      and the second item being the static_integrity
+      """
       def static_lookup(path) do
         Phoenix.Config.cache(__MODULE__, {:__phoenix_static__, path},
                              &Phoenix.Endpoint.Supervisor.static_lookup(&1, path))
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -699,6 +699,11 @@ defmodule Phoenix.Endpoint do
         Phoenix.Config.cache(__MODULE__, {:__phoenix_static__, path},
                              &Phoenix.Endpoint.Supervisor.static_path(&1, path))
       end
+
+      def integrity(path) do
+        Phoenix.Config.cache(__MODULE__, {:__phoenix_integrity__, path},
+                             &Phoenix.Endpoint.Supervisor.integrity(&1, path))
+      end
     end
   end
 
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -227,7 +227,7 @@ defmodule Phoenix.Endpoint do
   that are automatically defined in your endpoint:
 
     * for handling paths and URLs: `c:struct_url/0`, `c:url/0`, `c:path/1`,
-      `c:static_url/0`, and `c:static_path/1`
+      `c:static_url/0`,`c:static_path/1`, and `c:static_integrity/1`
     * for handling channel subscriptions: `c:subscribe/2` and `c:unsubscribe/1`
     * for broadcasting to channels: `c:broadcast/3`, `c:broadcast!/3`,
       `c:broadcast_from/4`, and `c:broadcast_from!/4`
@@ -424,6 +424,11 @@ defmodule Phoenix.Endpoint do
   Generates a route to a static file in `priv/static`.
   """
   @callback static_path(path :: String.t) :: String.t
+
+  @doc """
+  Generates an integrity hash to a static file in `priv/static`.
+  """
+  @callback static_integrity(path :: String.t) :: String.t
 
   # Channels
 
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -430,6 +430,11 @@ defmodule Phoenix.Endpoint do
   """
   @callback static_integrity(path :: String.t) :: String.t | nil
 
+  @doc """
+  Generates a two item tuple containing the `static_path` and `static_integrity`
+  """
+  @callback static_lookup(path :: String.t) :: {String.t, String.t} | {String.t, nil}
+
   # Channels
 
   @doc """
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -428,7 +428,7 @@ defmodule Phoenix.Endpoint do
   @doc """
   Generates an integrity hash to a static file in `priv/static`.
   """
-  @callback static_integrity(path :: String.t) :: String.t
+  @callback static_integrity(path :: String.t) :: String.t | nil
 
   # Channels
 
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -431,7 +431,7 @@ defmodule Phoenix.Endpoint do
   @callback static_integrity(path :: String.t) :: String.t | nil
 
   @doc """
-  Generates a two item tuple containing the `static_path` and `static_integrity`
+  Generates a two item tuple containing the `static_path` and `static_integrity`.
   """
   @callback static_lookup(path :: String.t) :: {String.t, String.t} | {String.t, nil}
 
--- a/lib/phoenix/endpoint/supervisor.ex
+++ b/lib/phoenix/endpoint/supervisor.ex
@@ -329,13 +329,6 @@ defmodule Phoenix.Endpoint.Supervisor do
   """
   @invalid_local_url_chars ["\\"]
 
-  def static_path(endpoint, path) do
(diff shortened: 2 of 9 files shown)


josevalim:
Hi @tomciopp! Great work, I like this a lot!

I have some suggestions:

1. Instead of having two keys, let's cache both the path and the integrity key under the same place. This way, getting the `static_path` will be something like `static_lookup(path) |> elem(0)` if it is found and `static_lookup(path) |> elem(1)` for the integrity. This will simplify the code and reduce the cache size.

2. Let's call this `static_integrity` for now. I am wondering which of those two APIs are better: a) a function that returns both the path and the integrity or b) a function that returns only the integrity.

Given that, if you want to use the integrity you always need both, I am inclined for the latter. This also means we can simplify 1) as static_path can also be computed from the integrity.

tomciopp:
I agree with caching under the same tag and changing the api name to `static_integrity`. I will update the code at some point later today with those changes.

Right now I only needed to return the integrity hash based on our codebase. However considering the fact that most developers have not heard of SRI and that it offers benefits without any real downside I think we should find a way to enable it by default for everyone. If we go this route, I think it would be smart to have two functions. One that returns just the hash, and one that returns that both so we can avoid duplicate calls when building `Phoenix.HTML.Tag` functions.

I can add `script_tag` and `stylesheet_tag` functions to `phoenix_html` to take advantage of this feature after this PR gets merged.

tomciopp:
@josevalim I've updated the code based on your suggestions. 

I have a few questions based on decisions made. I'll mark up my own code so you can take a look at it.

tomciopp:
@chrismccord @josevalim Is there anything I can do to speed this along? Still awaiting feedback for the most recent changes.

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -327,19 +329,26 @@ defmodule Phoenix.Endpoint.Supervisor do
...
 
-  def static_path(_endpoint, "//" <> _ = path) do
+  def static_path(endpoint, path) do
+    case static_lookup(endpoint, path) do
+      {:nocache, {path, _}} -> {:nocache, path}
+      other -> other
+    end
+  end
+
+  def static_lookup(_endpoint, "//" <> _ = path) do
```
josevalim: This should be the first clause, otherwise it won't ever match.

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -381,25 +390,38 @@ defmodule Phoenix.Endpoint.Supervisor do
...
   end
 
+  defp static_cache(digests, value) do
+    {
+      "/#{value}?vsn=d",
+      static_integrity(get_in(digests, [value, "sha512"]))
+    }
+  end
+
+  defp static_integrity(nil), do: ""
```
josevalim: It would be better to return nil in these cases, no?

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -381,25 +390,38 @@ defmodule Phoenix.Endpoint.Supervisor do
   end
 
   defp warmup_static(endpoint) do
-    for {key, value} <- cache_static_manifest(endpoint) do
+    warmup_static(endpoint, cache_static_manifest(endpoint))
+    endpoint.static_path("/")
+  end
+
+  defp warmup_static(endpoint, %{"latest" => latest, "digests" => digests}) do
+    Enum.each(Map.keys(latest), fn key ->
```
josevalim:
```suggestion
    Enum.each(latest, fn {key, _} ->
```

To avoid traversing the keys twice.

lib/phoenix/endpoint.ex
```diff
@@ -696,8 +696,16 @@ defmodule Phoenix.Endpoint do
...
                              &Phoenix.Endpoint.Supervisor.static_path/1) <>
+        elem(static_lookup(path), 0)
+      end
+
+      def static_integrity(path) do
+        elem(static_lookup(path), 1)
+      end
+
+      def static_lookup(path) do
         Phoenix.Config.cache(__MODULE__, {:__phoenix_static__, path},
```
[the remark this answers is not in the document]
josevalim: Yeah, we don't have to worry about it, it is totally private API.

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -381,25 +390,38 @@ defmodule Phoenix.Endpoint.Supervisor do
   end
 
   defp warmup_static(endpoint) do
-    for {key, value} <- cache_static_manifest(endpoint) do
+    warmup_static(endpoint, cache_static_manifest(endpoint))
+    endpoint.static_path("/")
```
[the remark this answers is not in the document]
josevalim: This is to warm up the `:__phoenix_static__` cache: https://github.com/phoenixframework/phoenix/pull/3372/files#diff-2e48d48e1c6c3b085dab25613ade742bR697

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -327,19 +329,26 @@ defmodule Phoenix.Endpoint.Supervisor do
   """
   @invalid_local_url_chars ["\\"]
 
-  def static_path(_endpoint, "//" <> _ = path) do
+  def static_path(endpoint, path) do
```
josevalim: Is this function ever called now? What called `static_path/2` before is now calling `static_lookup/2`.

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -327,19 +329,26 @@ defmodule Phoenix.Endpoint.Supervisor do
...
     raise_invalid_path(path)
   end
 
-  def static_path(_endpoint, "/" <> _ = path) do
+  def static_lookup(_endpoint, "/" <> _ = path) do
     if String.contains?(path, @invalid_local_url_chars) do
       raise ArgumentError, "unsafe characters detected for path #{inspect path}"
     else
-      {:nocache, path}
+      {:nocache, {path, ""}}
```
josevalim: If we don't have integrity, I would prefer to return `nil`. I left a similar comment below. :)

lib/phoenix/router/helpers.ex
```diff
@@ -227,6 +227,14 @@ defmodule Phoenix.Router.Helpers do
         endpoint.static_url <> endpoint.static_path(path)
       end
 
+      @doc """
+      Generates an integrity hash to a static asset given its file path.
+      """
+
```
josevalim: No new lines between docs and function.

lib/phoenix/router/helpers.ex
```diff
@@ -227,6 +227,14 @@ defmodule Phoenix.Router.Helpers do
...
+      def static_integrity(%Conn{private: %{endpoint: endpoint}}, path) do
```
josevalim: We need to repeat all the three clauses found in `static_path`.

lib/phoenix/endpoint.ex
```diff
@@ -696,8 +696,16 @@ defmodule Phoenix.Endpoint do
...
         Phoenix.Config.cache(__MODULE__, :__phoenix_static__,
                              &Phoenix.Endpoint.Supervisor.static_path/1) <>
+        elem(static_lookup(path), 0)
+      end
+
+      def static_integrity(path) do
+        elem(static_lookup(path), 1)
+      end
+
+      def static_lookup(path) do
```
josevalim: I think this one is private.

josevalim: Do we want to keep this one public or only private for now?

lib/phoenix/endpoint.ex
```diff
@@ -696,8 +696,16 @@ defmodule Phoenix.Endpoint do
       def static_path(path) do
         Phoenix.Config.cache(__MODULE__, :__phoenix_static__,
                              &Phoenix.Endpoint.Supervisor.static_path/1) <>
+        elem(static_lookup(path), 0)
+      end
+
+      def static_integrity(path) do
```
josevalim: We should document this one and the next one (if we decide to keep it public).

lib/phoenix/endpoint/supervisor.ex
tomciopp: According to the spec, it should either be an empty string or a valid metadata. I'm unsure how eex handles nil values for attributes, so I went with the option I knew would work. _I think it should work_ if given a `nil` but am not positive.

https://www.w3.org/TR/SRI/#the-integrity-attribute

lib/phoenix/endpoint/supervisor.ex
tomciopp: Will fix in next commit.

lib/phoenix/endpoint/supervisor.ex
tomciopp: I believe it is called in some tests, and since it is public I didn't want to remove it without at least giving a deprecation notice beforehand.

lib/phoenix/router/helpers.ex
tomciopp: will fix in next commit

lib/phoenix/router/helpers.ex
tomciopp: ok will fix

lib/phoenix/endpoint.ex
tomciopp: I think it should be public. My plan after this was to use this function to create `script_tag` and `stylesheet_tag` functions in `phoenix_html` that auto include the integrity attribute. No sense in looking up the cache twice.

lib/phoenix/endpoint.ex
tomciopp: Ok will add docs.

lib/phoenix/endpoint/supervisor.ex
josevalim: It is not public. The whole module is doc false. We should update the docs to call `static_lookup` instead. :)

lib/phoenix/endpoint/supervisor.ex
josevalim: It should work with nil, yes. Nil is converted into an empty string by EEx. Even if it wasn't, then users could convert it. It is best for us to return the information in the best format possible instead of trying to foresee how it is used by the view.

lib/phoenix/endpoint.ex
josevalim: Perfect.

josevalim: Perfect. Let's go ahead and document it too then!

lib/phoenix/endpoint/supervisor.ex
tomciopp: Ok, then it should clearly return `nil`. Will update.

lib/phoenix/endpoint/supervisor.ex
tomciopp: Will update

lib/phoenix/endpoint.ex
tomciopp: Will update

tomciopp:
@josevalim Ok, everything from the review looks settled. Do you want me to squash the commits?

josevalim:
Hi @tomciopp! The code is excellent, we are almost there. We only need to:

1. Write more docs. Search for `static_path` in the endpoint docs and you will see we also need to mention it in the module docs and add a callback as well

2. Write tests. We need unit tests for digests, for the endpoint and for the router

Thanks!

tomciopp:
@josevalim Sounds good, I will take care of this later today

tomciopp:
@josevalim Done, should we squash into a single commit?

josevalim:
No need to squash, we can do so when merging!

lib/phoenix/endpoint.ex
```diff
@@ -425,6 +425,11 @@ defmodule Phoenix.Endpoint do
   """
   @callback static_path(path :: String.t) :: String.t
 
+  @doc """
+  Generates an integrity hash to a static file in `priv/static`.
+  """
+  @callback static_integrity(path :: String.t) :: String.t
```
josevalim: It can also return `nil`. We should also document the static_lookup callback.

tomciopp: Done

tomciopp:
🎉 👍

josevalim:

PR #3374, opened 2019-04-15
Document phx.server alternative

axelson:
Document that setting `server: true` in your endpoint configuration will allow you to start your Phoenix Endpoint server without using the `mix phx.server` task. Document this in a place that a user will likely find it, in the mix task that they need to use (by default) to start their server.


josevalim:

PR #3375, opened 2019-04-16
Document compress option

sb8244:
This adds documentation for the compress option. There is already an example for it which shows it can be used. I tested it all still works well.


josevalim:

PR #3376, opened 2019-04-16
Provide a `--live` flag to generate a LiveView enabled phoenix project

aforward:
THIS IS A PROOF OF CONCEPT WITH SEVERAL OPEN QUESTIONS.
Other refactors might need to happen first, and then 
will be rebased into this work.  I wanted to get the ball rolling,
so here goes.

By running something like

```
mix phx.new hello_phoenix --no-ecto --live
```

Then, when you run the app

```
mix phx.server
```

You get an example of a LiveView generated page

Screen Shot 2019-04-16 at 8 17 55 AM

This is a follow up to:

https://github.com/phoenixframework/phoenix/pull/3371


installer/templates/phx_umbrella/apps/app_name_web/config/config.exs
```diff
@@ -15,7 +15,10 @@ config :<%= web_app_name %>, <%= endpoint_module %>,
   url: [host: "localhost"],
   secret_key_base: "<%= secret_key_base %>",
   render_errors: [view: <%= web_namespace %>.ErrorView, accepts: ~w(<%= if html do %>html <% end %>json), layout: false],
-  pubsub: [name: <%= web_namespace %>.PubSub, adapter: Phoenix.PubSub.PG2]
+  pubsub: [name: <%= web_namespace %>.PubSub, adapter: Phoenix.PubSub.PG2]<%= if live do %>,
+  live_view: [
+    signing_salt: "<%= live_view_signing_salt %>"
+  ]<% end %>
```
aforward: There was a suggestion to remove the separate live_view signing_salt.  That will require more coordination between the two projects (and clarification that it's OK to proceed with that route).  For now, I have the code generated based [on the instructions](https://github.com/phoenixframework/phoenix_live_view)

installer/templates/phx_live/controllers/page_controller.ex
```diff
@@ -0,0 +1,7 @@
+defmodule <%= web_namespace %>.PageController do
+  use <%= web_namespace %>, :controller
+
+  def index(conn, _params) do
+    Phoenix.LiveView.Controller.live_render(conn, <%= web_namespace %>.PageLiveView, session: %{})
```
aforward: To simplify this to just be `live_render(...)`, we need to also import the controller helper function

```
  def controller do
    quote do
      ...
      # Import convenience functions for LiveView rendering
      import Phoenix.LiveView.Controller, only: [live_render: 3]

       ...
    end
  end
```

I didn't see this explicitly in the instructions, wasn't sure if that was on purpose or not.

installer/templates/phx_single/config/config.exs
```diff
@@ -17,7 +17,10 @@ config :<%= app_name %>, <%= endpoint_module %>,
   url: [host: "localhost"],
   secret_key_base: "<%= secret_key_base %>",
   render_errors: [view: <%= web_namespace %>.ErrorView, accepts: ~w(<%= if html do %>html <% end %>json), layout: false],
-  pubsub: [name: <%= app_module %>.PubSub, adapter: Phoenix.PubSub.PG2]
+  pubsub: [name: <%= app_module %>.PubSub, adapter: Phoenix.PubSub.PG2]<%= if live do %>,
+  live_view: [
+    signing_salt: "<%= live_view_signing_salt %>"
+  ]<% end %>
```
aforward: See comment here
https://github.com/phoenixframework/phoenix/pull/3376/files#r275770071

The live_view_signing_salt might completely go away.

installer/lib/phx_new/single.ex
```diff
@@ -67,6 +67,17 @@ defmodule Phx.New.Single do
...
+    {:eex, "phx_live/controllers/page_controller.ex",        :project, "lib/:lib_web_name/controllers/page_controller.ex"},
+    {:eex, "phx_web/templates/layout/app.html.eex",          :project, "lib/:lib_web_name/templates/layout/app.html.eex"},
+    {:eex, "phx_web/templates/page/index.html.eex",          :project, "lib/:lib_web_name/templates/page/index.html.eex"},
+    {:eex, "phx_web/views/layout_view.ex",                   :project, "lib/:lib_web_name/views/layout_view.ex"},
+    {:eex, "phx_live/live/page_live_view.ex",                :project, "lib/:lib_web_name/live/page_live_view.ex"},
+    {:eex, "phx_test/controllers/page_controller_test.exs",  :project, "test/:lib_web_name/controllers/page_controller_test.exs"},
+    {:eex, "phx_test/views/layout_view_test.exs",            :project, "test/:lib_web_name/views/layout_view_test.exs"},
+    {:eex, "phx_test/live/page_live_view_test.exs",         :project, "test/:lib_web_name/live/page_live_view_test.exs"},
+  ]
+
```
aforward: There is definitely overlap with `template :html`.  I wasn't sure how best to manage conflicts with the same name for the _sample_ liveview where the `phx_live/controllers/page_controller.ex` is different.  I also wasn't sure if we wanted to manage all cases

| live? | html? | Description |
| --- | --- | --- |
| F | F | No HTML / Views at all |
| F | T | Yes HTML, no LiveView |
| T | F | Only show a LiveView example |
| T | T |  Show both the LiveView AND the RegularView |

installer/templates/phx_assets/webpack/app.js
```diff
@@ -15,3 +15,8 @@ import css from "../css/app.css"
 //
 // Local files can be imported directly using relative paths, for example:
 // import socket from "./socket"
+
+<%= if live do %>// Enabled LiveView Socket
+import LiveSocket from "phoenix_live_view"
+let liveSocket = new LiveSocket("/live")
+liveSocket.connect()<% end %>
```
aforward: `--live` was explicitly asked for, so I left this code directly commented out. When moving towards `--live` as the default, we possibly might want this shown, but commented out.

installer/templates/phx_live/controllers/page_controller.ex
josevalim: We should import it, yes!

installer/lib/phx_new/single.ex
josevalim: Live implies HTML. If Live is given, we won't generate a HTML. So the fourth row in your table is not necessary.

josevalim: Alternatively, we don't show a Live example here and leave it for later (in a possible phx.gen.live generator). Especially because if we have a Live example, we don't need a controller.

aforward: I have been playing around with generated examples, and I am going to look at having live just add a snippet within the HTML so that both can co-exist as it does provide a nice view of live and non-live code.

I’ll look separately at the generator, which might make more sense in the long run as “—live” might become the default

installer/templates/phx_live/controllers/page_controller.ex
```diff
@@ -0,0 +1,7 @@
+defmodule <%= web_namespace %>.PageController do
+  use <%= web_namespace %>, :controller
+
+  def index(conn, _params) do
+    render(conn, "index.html")
```
aforward: Moved away from 100% liveview, and instead did an embedded example within the `index.html.eex`

installer/templates/phx_live/templates/page/index.html.eex
```diff
@@ -0,0 +1,35 @@
+<%%= live_render(@conn, <%= web_namespace %>.PageLiveView) %>
```
aforward: Moved from full on `leex` to an embedded version within the index page.

chazwatkins:
Is there anything pending on this being merged other than the conflicts?  I'd love to use the `--live` flag to avoid the tedious ceremony of setting up `live_view`.

aforward:
I'd be willing to clean up the merge conflicts if there was some indication that it might get merged.  Even if it doesn't that OK, but at least knowing _it might_ would be enough for me to repoke at this.  The PR is based on a very early version of LiveCode so a lot might have changed since then.

josevalim:
This is something we would like to get in but we don’t know how nor when at the moment. Even the installation steps changed considerably and they may continue to change, so I think it is best if more work is not put into this for now. We may also want to try different ideas for the home page but we are not sure about it either.

In any case, thanks for all the work @aforward.

aforward:
holding pattern continues :-)

chrismccord:
We'll continue this in #3680 . Thanks!

PR #3380, opened 2019-04-19
Modernize config/prod.secret.exs

josevalim:
We have historically stayed away from environment variables
because it was incompatible with releases but now with
Distillery 2 and `mix release` in Elixir v1.9 this is a
solved problem, so we can push towards those best practices
for new apps.


chrismccord:
❤️❤️❤️🐥🔥

PR #3382, opened 2019-04-24
Fix @spec issue with `Template.template_not_found`

TheFirstAvenger:
Phoenix.Template.template_not_found is designed to be optionally overridden by the user, but the @spec specifies a return type of `no_return`. This PR adds the possible expected return types.


TheFirstAvenger:
Closing as this change trades one warning when specifying your own `template_not_found spec` for another warning when you don't.

Issue #1955, opened 2016-10-02
Provide --no-gettext option for phoenix.new


johnnyshields:
IMHO I18n is overkill for many smaller applications and it would be nice to be able to exclude it when generating a new app.

In addition, folks coming from Rails like myself might prefer the Rails/Linguist-style "key-value" I18n; in my case I have 10,000+ translation keys and a translation workflow based around that structure.

josevalim:
Once the work on the new generators are done we can look into it. It should be relatively straight-forward.

chrismccord:
Noted in the generator PR

tvsijin:
Any updates?

PR #3387, opened 2019-04-26
Add `--no-gettext` flag to `phx.new` task

manelli:
Usage example: `mix phx.new phx_blog --no-gettext`

When a Phoenix app is created with `--no-gettext`:
- gettext files are not generated (`gettext.ex` and `priv/gettext/*`)
- gettext dependency is not required in `mix.exs`
- gettext related code is not generated

Resolves: https://github.com/phoenixframework/phoenix/issues/1955


installer/templates/phx_web/views/error_helpers.ex
```diff
@@ -1,19 +1,19 @@
...
+
   use Phoenix.HTML
 
   @doc """
   Generates tag for inlined form input errors.
   """
   def error_tag(form, field) do
     Enum.map(Keyword.get_values(form.errors, field), fn error ->
-      content_tag(:span, translate_error(error), class: "help-block")
+      content_tag(:span, <%= error_message %>, class: "help-block")
```
michalmuskala: I think in case there's no gettext there needs to be an alternative implementation of error message interpolation - the code in this PR will actually crash when executed, trying to put the error tuple inside of a `span` tag.

josevalim: Ah, yes, good catch. We should probably keep the `translate_error` function but implement it differently. It would be something like this:

    def translate_error({msg, opts}) do
      Enum.reduce(opts, msg, fn {key, value}, acc ->
        String.replace(acc, "%{#{key}", to_string(value))
      end)
    end

manelli: Added the alternative implementation here: https://github.com/phoenixframework/phoenix/pull/3387/commits/982d86dd4befecad3e2daeb412ffa9188cccf15e

josevalim:
So [this bug](https://github.com/elixir-ecto/ecto/issues/2991) actually blocks this PR as the generated code as an alternative to Gettext will fail for some validations.

manelli:
Please let me know if there's anything I can do for this PR to be merged.
Thank you very much!

installer/templates/phx_web/templates/page/index.html.eex
```diff
@@ -1,5 +1,5 @@
 <section class="phx-hero">
-  <h1><%%= gettext "Welcome to %{name}!", name: "Phoenix" %></h1>
```
chrismccord: I'd prefer to leave this since it's an example for new users to see. Can you update?

chrismccord:
Just one small change and this is good to go. Thanks!

installer/templates/phx_web/templates/page/index.html.eex
manelli: Fixed on: https://github.com/phoenixframework/phoenix/pull/3387/commits/124cb48efa294f0fbf0b9832e65256458d7419c0

josevalim:

PR #3388, opened 2019-04-27
Update ErrorView Guide

iv-mexx:
The guide does not match with the default `ErrorView` of a new Project anymore.

I tried to update the guide to match the current generated ErrorView. Since I'm just learning Elixir and Phoenix, I hope I understood everything correctly.

 I'm not sure if its necessary to mention implementing the `def render("404.html", _assigns)` function because

a) It is deleted in the next step right away and does not provide much value over the the default implementation
b) `ErrorView` already contains an example of this possibility which, in my opinion, is very self explanatory

```
  # If you want to customize a particular status code
  # for a certain format, you may uncomment below.
  # def render("500.html", _assigns) do
  #   "Internal Server Error"
  # end
```

### DIFF
--- a/guides/views.md
+++ b/guides/views.md
@@ -182,18 +182,23 @@ This is where the view module and its template from the controller are rendered 
 
 ## The ErrorView
 
-Phoenix has a view called the `ErrorView` which lives in `lib/hello_web/views/error_view.ex`. The purpose of the `ErrorView` is to handle two of the most common errors - `404 not found` and `500 internal error` - in a general way, from one centralized location. Let's see what it looks like.
+Phoenix has a view called the `ErrorView` which lives in `lib/hello_web/views/error_view.ex`. The purpose of the `ErrorView` is to handle errors in a general way, from one centralized location. Let's see what it looks like.
 
 ```elixir
 defmodule HelloWeb.ErrorView do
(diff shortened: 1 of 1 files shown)


guides/views.md
```diff
@@ -212,19 +217,35 @@ config :hello, HelloWeb.Endpoint,
...
 
 The first question is, where does that error string come from? The answer is right in the `ErrorView`.
 
+```elixir
+def template_not_found(template, _assigns) do
+  Phoenix.Controller.status_message_from_template(template)
+end
+```
+
+Great, so we have this `template_not_foun/2` function that takes a template and an `assigns` map, which we ignore. Where is this function being called from? The answer is the `render/5` function defined in the `Phoenix.Endpoint.RenderErrors` module. The whole purpose of this module is to catch errors and render them with a view, in our case, the `HelloWeb.ErrorView`.
```
Gazler: `template_not_foun` is missing a "d".

Gazler:
Thanks! One small change then I think we can merge this in.

guides/views.md
iv-mexx: Done! 👍

Gazler:
Thanks <3

PR #3389, opened 2019-04-28
Add \q command to exit psql after creating user

iv-mexx:
Just a small addition where I got stuck for a second while following the guide.

The second sample also lists the `\q` command to exit the `psql` terminal, but if the reader follows along  from the beginning (and is unfamiliar with the Postgres Terminal at this point) he might get stuck before getting there.


chrismccord:
❤️❤️❤️🐥🔥

PR #3392, opened 2019-05-02
Fix `Phoenix.Controller.current_url`

radditude:
There seems to be a small bug in the latest release - the `current_url` helper throws an undefined function error because it's trying to call `ExampleWeb.Router.url\1` instead of `ExampleWeb.Router.Helpers.url\1`.

I've proposed a fix here using `Module.concat`. I've also updated the tests to use mock `Endpoint` and `Router` modules that more closely follow the structure of those found in Phoenix projects in order to prevent a recurrence.

_Edit_: forgot to include steps to repro the original bug:

- Generate a new empty app using v1.4.4
- Add a line to `ExampleWeb.PageController.index`:
```
    IO.inspect(current_url(conn))
```
- Visit the page and see the error

Thanks for reading!

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -1438,7 +1438,7 @@ defmodule Phoenix.Controller do
   See `current_url/2` to override the default parameters.
   """
   def current_url(%Plug.Conn{} = conn) do
-    router_module(conn).url(conn) <> current_path(conn)
+    Module.concat(router_module(conn), "Helpers").url(conn) <> current_path(conn)
   end
 
   @doc ~S"""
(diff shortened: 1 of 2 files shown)


chrismccord:
Thanks for the report! We shouldn't need the Module.concat. Looks like it was introduced here https://github.com/phoenixframework/phoenix/commit/cd1ce956a294bf70a0882b20c0706a18ae570502

@josevalim can you take a look?

chrismccord:
Perhaps we meant to call `endpoint_module` instead of `router_module`. In that case, that should be the only change we need.

josevalim:
It is not the endpoint module. I will have to partially revert that commit.

PR #3400, opened 2019-05-06
Listen on 127.0.0.1 by default

GriffinMB:
Listening on all interfaces by default may inadvertently overexpose a webservice to a network. 

I haven't seen previous issues about this, and since this is a small change, I figured I would start discussion with a PR.

### DIFF
--- a/installer/templates/phx_single/config/prod.secret.exs
+++ b/installer/templates/phx_single/config/prod.secret.exs
@@ -13,5 +13,5 @@ secret_key_base =
     """
 
 config :<%= app_name %>, <%= endpoint_module %>,
-  http: [:inet6, port: String.to_integer(System.get_env("PORT") || "4000")],
+  http: [:inet6, ip: {127, 0, 0, 1}, port: String.to_integer(System.get_env("PORT") || "4000")],
   secret_key_base: secret_key_base


josevalim:
Yeah, I wouldn't do this on dev because it has shown to have issues on some systems (I think Windows?). I would be fine with trying it for prod though.

GriffinMB:
Ah, interesting. Didn't realize Windows doesn't implement a loop back interface by default. I'll update the PR to be prod-only, which sounds fine to me.

josevalim:
I don't want to say it doesn't implement it, it is more of a "issues have been reported in the past", haha. :D

GriffinMB:
So, not that this is of any urgency or anything, but is there anything in particular that is holding this PR back from being merged?

josevalim:
It looks good to me but I think we need a second pair of eyes on it just in case. :)

installer/templates/phx_single/config/prod.secret.exs
```diff
@@ -13,5 +13,5 @@ secret_key_base =
     """
 
 config :<%= app_name %>, <%= endpoint_module %>,
-  http: [:inet6, port: String.to_integer(System.get_env("PORT") || "4000")],
+  http: [:inet6, ip: {127, 0, 0, 1}, port: String.to_integer(System.get_env("PORT") || "4000")],
```
jonnystoten: This fails because when the `:inet6` flag is given we need to pass an IPv6 address for `:ip`. From some basic testing on my local machine it looks like we can specify the IPv6 loopback address and it will work with both IPv6 and IPv4.
```suggestion
  http: [:inet6, ip: {0, 0, 0, 0, 0, 0, 0, 1}, port: String.to_integer(System.get_env("PORT") || "4000")],
```

achempion:
If we are going to listen on `127.0.0.1` would it fail by default if the server is running inside container? I can remember we need to listen on lo interface in order to get a traffic from outside.

installer/templates/phx_single/config/prod.secret.exs
chulkilee: > we can specify the IPv6 loopback address and it will work with both IPv6 and IPv4.

Actually.. only `localhost` works. If it's bound to ipv6 loopback (`::1`) then ipv4 loopback (`127.0.0.1`) wouldn't work.

```sh
curl -v 127.0.0.1:4000
*   Trying 127.0.0.1...
* TCP_NODELAY set
* Connection failed
* connect to 127.0.0.1 port 4000 failed: Connection refused
* Failed to connect to 127.0.0.1 port 4000: Connection refused
* Closing connection 0
curl: (7) Failed to connect to 127.0.0.1 port 4000: Connection refused
```

```sh
curl -v localhost:4000
*   Trying ::1...
* TCP_NODELAY set
* Connected to localhost (::1) port 4000 (#0)
> GET / HTTP/1.1
> Host: localhost:4000
> User-Agent: curl/7.64.1
> Accept: */*
>
< HTTP/1.1 404 Not Found
(...omitted...)
```

chulkilee:
> If we are going to listen on `127.0.0.1` would it fail by default if the server is running inside container? I can remember we need to listen on lo interface in order to get a traffic from outside.

Listening only on loopback interface on PROD would only work if proxy server is running in the same machine and use the local port for upstream connection.

For most cases, with or without container, it requires listening on "public" interface - either all interface or specific interface (e.g. specific ip address).

I agree that we should encourage to permit specific interface - but using localhost seems bad default for me.

nathanl:
I like making this setting explicit, but I don't think this should be the default configuration.

The best IP to listen on in production depends on the deployment setup:

1. If you're using a reverse proxy (like Nginx) in front of Phoenix, **and** if that reverse proxy is running on the **same host** as Phoenix, Phoenix should listen on `127.0.0.1` to prevent requests from coming to it directly.
2. If you're not using a reverse proxy (using Cowboy only), or if the reverse proxy is running on a different machine, or if Phoenix is running inside a container or VM, Phoenix should listen on `0.0.0.0`; **otherwise it will be completely inaccessible**.

Option 2 is the default behavior for good reason and I think we should leave it as such. But it would be good to let people know about Option 1.

My suggestion is:

- We **explicitly** configure `ip: {0, 0, 0, 0}` rather than leaving it implicit, to surface that option
- We add comments or link to documentation to explain the above

josevalim:
Good call! I think making it explicit is a good compromise. :+1:

josevalim:
The explicit approach ash been pushed in anogther PR.

chulkilee:
For the record, #3993 is the PR mentioned above.

Issue #3394, opened 2019-05-02
SRI Hash mismatch 


tomciopp:
### Environment

* Elixir version (elixir -v): 1.8.1
* Phoenix version (mix deps): 1.4.4
* NodeJS version (node -v): 12.1.0
* NPM version (npm -v): 6.9.0
* Operating system: OSX

### Expected behavior
Routes.static_integrity hash should be valid when added as an attribute on a static asset. 

### Actual behavior
In certain conditions the hash will mismatch causing the asset to not load.

### Background

This was a recent feature added by me, and much to my chagrin I ran into an error once I attempted to run this on our site. I spent yesterday evening tracking down what went wrong and am posting the bug here in case anyone runs into it while attempting to add SRI to their site. 

### The issue

Within the digester code, we add the sha512 as a key during the `build_digest/1` step. However there is a subtle mistake that I made. The `file.content` that is being `sha512`'d  is the raw content, not the digested content. Phoenix will replace the contents of certain attributes so that they reference the same digested urls within the content of the digested file. Since the content is different, obviously the hashes will not match and the integrity attribute will be wrong.

### Fixing

I'm going to clone the digester in my project and see if I can add the `digested_content` to that map in the same step that we are adding the sha. I believe there should be a way to rearrange the steps so that we don't see a slowdown and we can get the correct output. I'm going to take this afternoon to see if I can fix it and will report back here if I fail or with a PR if I succeed. (Success in this case means a verified working version that is accessible over the internet)
If you have any alternative ideas that you think are better options let me know.

josevalim:
It is fine to keep the digested content around too as long as we don't persist it to the manifest. :+1: Thanks for looking into this.

funtikas25:
hello pleas help what the hash is hier ???????
[18:21:27] [WARNING] there has been a problem while writing to the session file ('OperationalError: attempt to write a readonly database')
[18:21:30] [INFO] retrieved: 07b1e0439ba077bc7bdfbbf2341b1d1f:dc
[18:21:30] [DEBUG] performed 249 queries in 380.19 seconds
[18:21:30] [INFO] retrieving the length of query output
[18:21:30] [INFO] retrieved: 21
[18:23:18] [DEBUG] performed 10 queries in 107.48 seconds
[18:23:18] [DEBUG] starting 8 threads
[18:24:14] [INFO] retrieved: o_i________________ 2/21 (9%)

PR #3404, opened 2019-05-07
bugfix(SRI): Ensure that we are hashing digested file content

tomciopp:
This PR fixes the issues I brought up in #3394. I've tested this fix in production on our project and it does work as originally intended. It is slightly inefficient since we are digesting the file twice, we could  improve that but I will leave for comment.

### DIFF
--- a/lib/phoenix/digester.ex
+++ b/lib/phoenix/digester.ex
@@ -8,7 +8,7 @@ defmodule Phoenix.Digester do
   }
 
   defp now() do
-    :calendar.datetime_to_gregorian_seconds(:calendar.universal_time)
+    :calendar.datetime_to_gregorian_seconds(:calendar.universal_time())
   end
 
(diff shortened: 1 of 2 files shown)


josevalim:
Unfortunately this is going to digest the contents twice. Can you please digest it only once and pass it around accordingly? Thank you.

josevalim:
Also, please include a test, so we avoid doing the same mistake in the future. Thank you!

tomciopp:
@josevalim You're right, I've updated the PR to reflect the requested changes.

tomciopp:
I also ran the formatter so there might be some extra changes that are reflected.

josevalim:
Thank you! But please do not format and change code the next time around, as it makes it really hard to review PRs, backport the code, etc.

josevalim:

PR #3406, opened 2019-05-08
Fix secrets related info in Deployment docs

phil-r:
I guess it is a typo, as it is generally bad idea to commit secrets to vcs and `config/prod.secret.exs` itself mentions it:

```
# ... You can also
# hardcode secrets, although such is generally not
# recommended and you have to remember to add this
# file to your .gitignore
```


chrismccord:
❤️❤️❤️🐥🔥

PR #3407, opened 2019-05-09
Context.function_count/1 accounts for both def and defdelegate

aesmail:
This PR will allow `Context.function_count/1` to count both `def` and `defdelegate` functions.

Currently the `Context.function_count/1` function returns the number of `def` functions in the Context file.
I usually change the context file to contain only/mostly `defdelegate` functions which are delegated to modules under `context_folder/api/module_name.ex`. The prompt I get in the terminal about the number of functions I have in the Context file is always inaccurate due to this issue. It's a minor issue and might not be popular, but I'd thought I'd create this PR to scratch my own itch since it was just a one-liner.


josevalim:

PR #3409, opened 2019-05-09
Don't ignore plug params and pass them to init

### LINKED ISSUE #3393: Dialyzer error when passing a plug to get/4
### Environment

* Phoenix version (mix deps): 1.4.4

### Expected behavior
When passing a plug to get/4 no dialyzer errors should be reported.

### Actual behavior
Dialyzer blames that the plug init fun is called with `[]` instead it should be called with a valid kw list (which is correct). Something in the `Router.get` macro expansion is driving dialyzer mad, since the code works ok.

Here's the log output:

> lib/foo_web/router.ex:1:no_return
> Function __checks__/0 has no local return.
> ________________________________________________________________________________
> lib/foo_web/router.ex:21:call
> The function call will not succeed.
> 
> FooWeb.Plugs.Test.init([])
> 
> will never return since it differs in arguments with
> positions 1st from the success typing arguments:
> 
> ([{:someparam, _}, ...])
> ________________________________________________________________________________
> done (warnings were emitted)

Minimal repo here: https://github.com/xadhoom/phx_dializer_issue just check how the `Test` plug is used in the router

Side issue: `plug_opts` in `get/4` has no typing, so is expected that any term may be passed to the custom plug, while really only some terms are valid, for example passing a map as plug_opts fails macro expansion.

xadhoom:
This should fix #3393

### DIFF
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -324,9 +324,9 @@ defmodule Phoenix.Router do
     {matches, _} = Enum.map_reduce(routes_with_exprs, %{}, &build_match/2)
 
     checks =
-      for {%{line: line}, %{dispatch: {plug, _}}} <- routes_with_exprs, into: %{} do
+      for {%{line: line}, %{dispatch: {plug, params}}} <- routes_with_exprs, into: %{} do
         quote line: line do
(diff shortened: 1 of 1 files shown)


josevalim:

PR #3411, opened 2019-05-11
Improve `change_<schema>` when generating a context.

tgxworld:
By default the schema that is generated defines a `changeset` that accepts a `attr` argument. It seems really strange to me that the `change_` function in the context defaults to always passing `%{}` to `changeset` by default.


josevalim:
That's on purpose because the whole goal is to set up an empty changeset. For the operations that actually change it, then there are other operations. Of course, you can change it in your app whenever you want. Thanks for the PR!

PR #3412, opened 2019-05-11
Replace Map.get with Keyword.get inside generated DataCase errors_on function

rosswilson:
The generated `errors_on` function inside DataCase uses `Map.get`, but `opts` is a keyword list.

Right now, the `errors_on` function from `master` causes an error, like:

```
(BadMapError) expected a map, got: [count: 3, validation: :length, kind: :min, type: :string]
```

This is related to the recently merged #3204 PR.


chrismccord:
❤️❤️❤️🐥🔥

PR #3417, opened 2019-05-15
Update Heroku buildpack docs

hubertlepicki:
Heroku deployment guide had two minor errors:
- URL of the generated Heroku app was wrong. It just had the subdomain part, missing the hostname. This results in application redirecting to wrong URL if user follows the guide to the word, and nothing works as a result.
- Procfile said add like "mix phx.server" and again, if user does not know anything about Procfiles and follows the guide, this results in application failing to start. One needs to add line "web: mix phx.server"
- minor formatting change


Gazler:
Thanks <3

PR #3419, opened 2019-05-17
Add route_info lookup

chrismccord:
@josevalim any input on how we should partially make `%Phoenix.Router.Route{}` public? Still needs docs about what is and isn't private, but not sure what's best at the moduledoc/or router_info function doc level. Would appreciate your thoughts


josevalim:
I think all of the current fields are safe to expose, so it is probably a matter of making the module public and hiding the documentation for all other functions (or perhaps move those functions to another module, so we don't end-up `@doc false` everything).

PR #3434, opened 2019-06-06
Fixed template example field name

Aethelflaed:
Minor problem I found while trying to doctest a context.

A little side-question: from what I see, doctesting a context is not supported out-of-the-box, so should I bother doing it or not?


josevalim:
@Aethelflaed we generally don't doctest functions with side-effects because it relies on setting something external, like a database, that may be too complex for the simplicity and style of doctests.

PR #3440, opened 2019-06-08
Fix Multi.run usage in contexts guide

### LINKED ISSUE #3438: Multi example in the Contexts doc might be outdated
### Environment

* Elixir version (elixir -v): Elixir 1.8.2 (compiled with Erlang/OTP 20)
* Phoenix version (mix deps): locked at 1.4.6 (phoenix)
* NodeJS version (node -v):
* NPM version (npm -v): 
* Operating system:

Code in the documentation for Contexts under [Strategies for cross-context workflows](https://github.com/phoenixframework/phoenix/blob/master/guides/contexts.md#strategies-for-cross-context-workflows) possibly outdated.

Was getting a failing test case at the latter part of the doc, the one with the Multi implementation.

Error:
** (FunctionClauseError) no function clause matching in Ecto.Multi.run/3

I checked the Multi.run doc and it says
> receives the repo as the first argument, and the changes so far as the second argument

I updated the run code to `Multi.run(:user, fn _, _ -> Accounts.create_user(params) end)`
 and it works.

### Expected behavior
```
def register_account(attrs) do
  Multi.new()
  |> Multi.run(:account, fn _, _ ->
    IO.puts "do something here"
  end)
end
```
Screenshot 2019-06-09 at 12 49 50 AM

### Actual behavior

```
def register_account(attrs) do
  Multi.new()
  |> Multi.run(:account, fn _ ->
    IO.puts "do something here"
  end)
end
```
Screenshot 2019-06-09 at 12 41 41 AM

Thank you!

wojtekmach:
Closes #3438

### DIFF
--- a/guides/contexts.md
+++ b/guides/contexts.md
@@ -1096,8 +1096,8 @@ defmodule Hello.UserRegistration do
 
   def register_user(params) do
     Multi.new()
-    |> Multi.run(:user, fn _ -> Accounts.create_user(params) end)
-    |> Multi.run(:author, fn %{user: user} ->
+    |> Multi.run(:user, fn _repo, _ -> Accounts.create_user(params) end)
+    |> Multi.run(:author, fn _repo, %{user: user} ->
       {:ok, CMS.ensure_author_exists(user)}
     end)
     |> Repo.transaction()


Gazler:
Thanks <3

PR #3444, opened 2019-06-12
Support Websocket subprotocols

Nitrino:
In the WebSockets standard, there is a feature to pass supported subprotocols.

To do this, the client passes the list of supported subprotocols in `Sec-WebSocket-Protocol` header of the upgrade request. eq: `Sec-Websocket-Protocol: v12.stomp, mqtt`.

The server selects a supported protocol and returns it in the same header `Sec-Websocket-Protocol: mqtt`, then the handshake is successful.

If the server does not return the header, then we get a handshake error, like that:
```javascript
WebSocket connection to 'ws://localhost:4000/socket/websocket' failed:
Error during WebSocket handshake: Sent non-empty 'Sec-WebSocket-Protocol' 
header but no response was received`
```
More details in rfc : https://tools.ietf.org/html/rfc6455#section-1.9

Many protocols require the explicit indication of the sub-protocol when implementing over WebSockets.
For example: 
 - SIP over WebSockets requires `sip` subprotocol: https://tools.ietf.org/html/rfc7118#section-4.1
- SOAP over WebSockets requires `sip` subprotocol: https://winprotocoldoc.blob.core.windows.net/productionwindowsarchives/MS-SWSB/%5bMS-SWSB%5d.pdf

Cowboy allows us to modify the `init` callback to implement this logic.
Example: https://ninenines.eu/docs/en/cowboy/2.1/guide/ws_handlers

In Phoenix, this can be implemented using [сustom dispatch options](https://hexdocs.pm/phoenix/Phoenix.Endpoint.CowboyAdapter.html#content).
```elixir 
  http: [
    dispatch: [
      {:_, [
        {"/socket/websocket", MyAppWeb.Endpoint.Cowboy2CustomHandler, {MyAppWeb.Endpoint, []}},
        {:_, Phoenix.Endpoint.Cowboy2Handler, {MyAppWeb.Endpoint, []}}
      ]}
    ]
  ],
```

But this is a tricky and complicated way requiring the support own cowboy handler.

This PR adds simple WebSockets subprotocol support by passing `subprotocol` option to `Phoenix.Endpoint.socket/3`

Example: 
```elixir
  socket "/socket", MyAppWeb.UserSocket,
    websocket: [subprotocol: "sip"],
    longpoll: false
```

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -836,8 +836,12 @@ defmodule Phoenix.Endpoint do
     * `:compress` - whether to enable per message compresssion on
       all data frames, defaults to false
 
-    * `:subprotocol` - supported websocket subprotocol.
-      Used for handshake `Sec-WebSocket-Protocol` response header
+    * `:subprotocols` - a list of supported websocket subprotocols.
+      Used for handshake `Sec-WebSocket-Protocol` response header, defaults to nil.
+
+      For example:
+
+          subprotocols: ["sip", "mqtt"]
 
   ## Longpoll configuration
 
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -836,6 +836,9 @@ defmodule Phoenix.Endpoint do
     * `:compress` - whether to enable per message compresssion on
       all data frames, defaults to false
 
+    * `:subprotocol` - supported websocket subprotocol.
+      Used for handshake `Sec-WebSocket-Protocol` response header
+
   ## Longpoll configuration
 
   The following configuration applies only to `:longpoll`:
--- a/lib/phoenix/socket/transport.ex
+++ b/lib/phoenix/socket/transport.ex
@@ -354,6 +354,49 @@ defmodule Phoenix.Socket.Transport do
   end
 
   @doc """
+  Checks the Websocket subprotocol request header against the allowed subprotocol.
+
+  Should be called by transports before connecting when appropriate.
+  If the sec-websocket-protocol header matches the allowed subprotocol,
+  it will put sec-websocket-protocol response header and return the given connection.
+
+  Otherwise a 403 Forbidden response will be sent and the connection halted.
+  It is a noop if the connection has been halted.
+  """
+  def check_origin(conn, subprotocol)
+
+  def check_subprotocol(%Plug.Conn{halted: true} = conn, _subprotocol), do: conn
+
+  def check_subprotocol(conn, subprotocol) do
+    import Plug.Conn
+    subprotocols = get_req_header(conn, "sec-websocket-protocol")
+
+    case Enum.member?(subprotocols, subprotocol) do
+      true ->
+        put_resp_header(conn, "sec-websocket-protocol", subprotocol)
+
+      false ->
+        Logger.error """
+        Could not check Websocket subprotocol for Phoenix.Socket transport.
+
+        Subprotocols of the request: #{inspect(subprotocols)}
+        Expected subprotocol: #{inspect(subprotocol)}
+
+        This happens when you are attempting a socket connection to
+        a different subprotocol than the one configured in your endpoint.
+        To fix this issue, you may either:
+
+          1. update websocket: [subprotocol: ...] to your actual subprotocol
+             in your endpoint socket configuration.
+
+          2. check the correctness of the `sec-websocket-protocol` request header
+             sends from the client.
+        """
+        conn |> resp(:forbidden, "") |> halt()
+    end
+  end
+
+  @doc """
   Extracts connection information from `conn` and returns a map.
 
   Keys are retrieved from the optional transport option `:connect_info`.
--- a/lib/phoenix/socket/transport.ex
+++ b/lib/phoenix/socket/transport.ex
@@ -374,8 +374,8 @@ defmodule Phoenix.Socket.Transport do
       [] ->
         conn
 
-      subprotocols_header ->
-        request_subprotocols = subprotocols_header |> hd |> Plug.Conn.Utils.list()
+      [subprotocols_header | _] ->
+        request_subprotocols = subprotocols_header |> Plug.Conn.Utils.list()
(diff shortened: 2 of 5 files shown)


lib/phoenix/endpoint/cowboy2_handler.ex
```diff
@@ -72,6 +75,20 @@ defmodule Phoenix.Endpoint.Cowboy2Handler do
...
+  defp maybe_subprotocol(req, nil), do: req
+
+  defp maybe_subprotocol(req, subprotocol) do
+    subprocols = :cowboy_req.parse_header("sec-websocket-protocol", req)
+
+    case Enum.member?(subprocols, subprotocol) do
+      true ->
+        :cowboy_req.set_resp_header("sec-websocket-protocol", subprocols, req)
+      false ->
+        raise "Cowboy2 adapter expected #{inspect(subprotocol)} websoket subprotocol " <>
```
sb8244: heads up typo in WebSocket (missing c)

Nitrino: Oh, thanks

josevalim:
Looks great @Nitrino ! However, I think we should handle this inside `Phoenix.Transports.WebSocket` using the Plug API. Also, it is probably best to log error and return :forbidden rather than raising, since the process crash can be quite unreadable. Like we do here: https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/socket/transport.ex#L327

Nitrino:
@josevalim I have moved handling Websocket subprotocol inside `Phoenix.Transports.WebSocket`.
 
I also noticed was added [`Phoenix.Endpoint.Cowboy2Handler.copy_resp_headers/2`](https://github.com/phoenixframework/phoenix/commit/4bf53ecaae5a9057ea57c248964490dfdee312af#diff-0b0978dc41c81162d7eb10536f08fd76R120) function for copying the response headers from the `conn` to the cowboy `req`.  But it seems still it is necessary to add a status code copy from the `conn`, because now all unsuccessful answers (e.g. forbidden) are returned with the 204 code.
I can do this in separate PR.

josevalim:
Thanks @Nitrino! Sorry for the delay here. I think the configuration that we allow should be a list of subprotocols. And then we will match our list against the user supplied list, returning the first one. Can you please amend it? I will also drop some comments.

lib/phoenix/endpoint.ex
```diff
@@ -835,6 +835,9 @@ defmodule Phoenix.Endpoint do
     * `:compress` - whether to enable per message compresssion on
       all data frames, defaults to false
 
+    * `:subprotocol` - supported websocket subprotocol.
+      Used for handshake `Sec-WebSocket-Protocol` response header
```
josevalim: What is the default value? What is the type of this option? And as I said, let's go with `:subprotocols`.

lib/phoenix/socket/transport.ex
```diff
@@ -353,6 +353,51 @@ defmodule Phoenix.Socket.Transport do
...
+  It is a noop if the connection has been halted.
+  """
+  def check_origin(conn, subprotocol)
+
+  def check_subprotocol(%Plug.Conn{halted: true} = conn, _subprotocol), do: conn
+  def check_subprotocol(conn, nil), do: conn
+
+  def check_subprotocol(conn, subprotocol) do
+    import Plug.Conn
+    subprotocols = get_req_header(conn, "sec-websocket-protocol")
```
josevalim: Shouldn't we split this on comma and remove spaces? You can use this function for help: https://hexdocs.pm/plug/Plug.Conn.Utils.html#list/1

lib/phoenix/socket/transport.ex
```diff
@@ -353,6 +353,51 @@ defmodule Phoenix.Socket.Transport do
...
+
+        This happens when you are attempting a socket connection to
+        a different subprotocol than the one configured in your endpoint.
+        To fix this issue, you may either:
+
+          1. update websocket: [subprotocol: ...] to your actual subprotocol
+             in your endpoint socket configuration.
+
+          2. check the correctness of the `sec-websocket-protocol` request header
+             sends from the client.
```
josevalim:
```suggestion
             sent from the client.
```

lib/phoenix/socket/transport.ex
```diff
@@ -353,6 +353,51 @@ defmodule Phoenix.Socket.Transport do
...
+
+  Should be called by transports before connecting when appropriate.
+  If the sec-websocket-protocol header matches the allowed subprotocol,
+  it will put sec-websocket-protocol response header and return the given connection.
+  If no sec-websocket-protocol header was sent it will return the given connection.
+
+  Otherwise a 403 Forbidden response will be sent and the connection halted.
+  It is a noop if the connection has been halted.
+  """
+  def check_origin(conn, subprotocol)
```
josevalim:
```suggestion
  def check_subprotocol(conn, subprotocol)
```

josevalim:
Also, we can please add some tests? And great work on the error message!

Nitrino:
Hi @josevalim.
I kind of fixed all the comments.
May I need somewhere else to add integration tests, but I do not find a place where they can be easily added.

josevalim:
Integration tests could be added here: https://github.com/phoenixframework/phoenix/blob/master/test/phoenix/integration/websocket_socket_test.exs#L81

You can see the origin tests and other tests that set headers for an example.

lib/phoenix/socket/transport.ex
```diff
@@ -353,6 +353,75 @@ defmodule Phoenix.Socket.Transport do
...
+
+  def check_subprotocols(%Plug.Conn{halted: true} = conn, _subprotocols), do: conn
+  def check_subprotocols(conn, nil), do: conn
+
+  def check_subprotocols(conn, subprotocols) when is_list(subprotocols) do
+    import Plug.Conn
+    request_subprotocols =
+      conn
+      |> get_req_header("sec-websocket-protocol")
+      |> List.first
```
josevalim: The list can be empty, which means this would return nil. If get_req_header returns an empty list, then it mean we shoudn't do any subverification.

josevalim:

PR #3447, opened 2019-06-16
Changes to init callback description to make it easier to read

axelson:
Specifically it makes it more obvious that exactly `:supervisor` is passed instead of potentially some other atom. Use "the" instead of "a" because there can only be one instance of an atom in the system.


josevalim:

axelson:
Thanks!

PR #3449, opened 2019-06-19
Note that PubSub actually is useful for application development

nathanl:


josevalim:

PR #3451, opened 2019-06-23
Make port configurable from terminal in dev

jamilabreu:


josevalim:
You can change it as necessary in your projects, I don't think this is commonly needed to make it part of all configs. Thanks!

PR #3452, opened 2019-06-25
Use hex in mix local.phx

adamu:
Following from [this Stack Overflow question](https://stackoverflow.com/questions/56743592/why-does-running-local-phx-downgrade-the-phoenix-archive) and a [closed issue](https://github.com/phoenixframework/archives/issues/2) on the archives repository, it is not clear from the `mix local.phx` docs that the command is no longer supported, and effectively downgrades to 1.3.2.

I deprecated the task, added a warning, and made the version that will be installed explicit, so that it's visible from the task output.

```
$ mix local.phx
WARNING: mix local.phx is no longer supported, use: mix archive.install hex phx_new
Are you sure you want to install "https://github.com/phoenixframework/archives/raw/master/phx_new-1.3.4.ez"? [Yn]
```

### DIFF
--- a/installer/lib/mix/tasks/local.phx.ex
+++ b/installer/lib/mix/tasks/local.phx.ex
@@ -1,7 +1,6 @@
 defmodule Mix.Tasks.Local.Phx do
   use Mix.Task
 
-  @url "https://github.com/phoenixframework/archives/raw/master/phx_new.ez"
   @shortdoc "Updates the Phoenix project generator locally"
 
   @moduledoc """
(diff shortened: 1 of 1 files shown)


installer/lib/mix/tasks/local.phx.ex
```diff
@@ -1,17 +1,21 @@
...
   """
+  @deprecated ~s'Use "mix archive.install hex phx_new" instead'
   def run(args) do
-    Mix.Task.run "archive.install", [@url | args]
+    IO.puts(
+      :stderr,
+      "WARNING: mix local.phx is no longer supported, use: mix archive.install hex phx_new"
+    )
+
+    Mix.Task.run("archive.install", [@url | args])
```
josevalim: Hi @adamu! Thanks for the PR and great job answering on Stack Overflow. Instead of deprecating it, I think we should simply update it to use Hex. Something like this:

```suggestion
    Mix.Task.run("archive.install", ["hex", "phx_new" | args])
```

Can you please remove the warning and update it?

adamu:
@josevalim that's a much friendlier solution. Done.

One problem it wont solve is that people will still end up downgrading to phx_new v1.3.4 unless they call the `archive.install hex phx_new` version manually at least once, to get the `local.phx` version that contains this fix. So I added a note to the docs explaining that.

josevalim:

josevalim:
I ported it to v1.4 already, so I will update the note. :)

PR #3453, opened 2019-06-25
Update phx.local docs to indicate deprecated status

vaer-k:
Running `mix phx.local` is not supported after version 1.3. https://github.com/phoenixframework/archives/issues/2


vaer-k:
Closing because an earlier PR covers the issue. #3452

PR #3455, opened 2019-06-29
Remove comment in Dockerfile

maartenvanvliet:
The first line should be executed by Docker as it sets elixir 1.9.0 as the base image


Gazler:
Thanks <3

PR #3456, opened 2019-06-29
Only install required dependencies in Dockerfile

maartenvanvliet:
The assets compilation script now requires `npm` instead of `yarn` so we need to install it first.
Removed `python` and `git` because they are not needed in most deployments.
`build-base` is preserved because `gcc` is often needed.

### DIFF
--- a/guides/deployment/releases.md
+++ b/guides/deployment/releases.md
@@ -154,7 +154,7 @@ Here is an example Docker file to run at the root of your application covering a
 # FROM elixir:1.9.0-alpine as build
 
 # install build dependencies
-RUN apk add --update git build-base nodejs yarn python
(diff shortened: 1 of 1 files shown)


guides/deployment/releases.md
```diff
@@ -154,7 +154,7 @@ Here is an example Docker file to run at the root of your application covering a
 FROM elixir:1.9.0-alpine as build
 
 # install build dependencies
-RUN apk add --update git build-base npm yarn python
+RUN apk add --update build-base npm
```
Gazler: Can we keep git in case anyone is using git dependencies? Then this is good to merge.

maartenvanvliet: Sure!

josevalim:

PR #3457, opened 2019-06-30
Fix Dockerfile build dependency

mainlymortal:
The current nodejs dependency does not include npm for some reason. Following the current instructions will cause a failure. There is no need to keep nodejs as it is a dependency of npm.


Gazler:
Thanks <3

PR #3474, opened 2019-07-11
Correct the pathname referred to in the documentation.

MichaelDimmitt:
The guide for releases suggest a code change to uncomment the following line in  config/prod.secret.exs:
```elixir
config :my_app, MyApp.Endpoint, server: true
```

However, it is actually in config/prod.exs.

elixir version: 1.9.0
phx version: 1.4.9

Overall the documentation worked really well and I was able to run the release. 👍

## Question:
Should the edit be made in config/prod.exs or should that line show up in config/prod/secret.exs when phoenix scaffolds a new app?

I ask this question because further down the documentation [#Runtime configuration](https://github.com/MichaelDimmitt/phoenix/blob/master/guides/deployment/releases.md#runtime-configuration) suggests that you should copy the config/prod/secret.exs into config/releases.exs


MichaelDimmitt:
@maartenvanvliet , @scudelletti, @mainlymortal 

This is before the Runtime Configuration section where you:
Rename config/prod.secret.exs to config/releases.exs

Additionally, is there no feedback on this pr because a test is failing.
Any guidance would be appreciated. Thanks. 🙂

josevalim:
The current guide suggestions are correct, it has been moved to config/prod.secret.exs in more recent releases. :)

PR #3476, opened 2019-07-12
Fix deployment link on Releases Guide

scudelletti:
In the [Releases Guide](https://hexdocs.pm/phoenix/releases.html#releases-assemble) the link for `deployment guide` is returning 404.
This PR aims to fix that with the [correct link](https://hexdocs.pm/phoenix/deployment.html).

Thank you very much for maintaining Phoenix! :tada: :heart:

### DIFF
--- a/guides/deployment/releases.md
+++ b/guides/deployment/releases.md
@@ -19,7 +19,7 @@ $ elixir -v
 
 If you are not familiar with Elixir releases yet, we recommend you to read [Elixir's excellent docs](https://hexdocs.pm/mix/Mix.Tasks.Release.html) before continuing.
 
-Once that is done, you can assemble a release by going through all of the steps in our general [deployment guide](/deployment.html) with `mix release` at the end. Let's recap.
+Once that is done, you can assemble a release by going through all of the steps in our general [deployment guide](https://hexdocs.pm/phoenix/deployment.html) with `mix release` at the end. Let's recap.
 
 First set the environment variables:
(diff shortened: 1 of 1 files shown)


guides/deployment/releases.md
```diff
@@ -19,7 +19,7 @@ $ elixir -v
 
 If you are not familiar with Elixir releases yet, we recommend you to read [Elixir's excellent docs](https://hexdocs.pm/mix/Mix.Tasks.Release.html) before continuing.
 
-Once that is done, you can assemble a release by going through all of the steps in our general [deployment guide](/deployment.html) with `mix release` at the end. Let's recap.
+Once that is done, you can assemble a release by going through all of the steps in our general [deployment guide](https://hexdocs.pm/phoenix/deployment.html) with `mix release` at the end. Let's recap.
```
Gazler: Good spot, the issue here is actually the `/` in `/deployment.html`

Could we change it to this instead?

`general [deployment guide](deployment.html) `

This ensures it will work outside of hexdocs too.

scudelletti: Done on fd47add I also changed the `Up and Running guide` to follow the same approach.

With the relative path they won't work in the github repo though as you can see it [here](https://github.com/scudelletti/phoenix/blob/ds-fix-release-link/guides/deployment/releases.md#releases-assemble).

Gazler:
Thanks <3

PR #3477, opened 2019-07-12
Improve docker keyword in Dockerfile

scudelletti:
This PR aims to use the `AS` keyword pattern as describe in the [docker documentation](https://docs.docker.com/develop/develop-images/multistage-build/#name-your-build-stages) not only in the app stage but also in the build stage.

Thank you for the time and for maintaining this awesome library. :confetti_ball: :tada:


Gazler:
Thanks <3

PR #3479, opened 2019-07-13
Update travis file to OTP 22 and Elixir v1.9

eksperimental:


josevalim:
Thanks @eksperimental! Btw, we would love a PR that removes the `excludes` and to explicit `includes` as we do in other Elixir projects.

eksperimental:
Yes,I also noticed that after I updated the PR. I will work in it then I have some free time

PR #3481, opened 2019-07-15
remove x-headers to headers

Douvi:
## What was done
Socket transport send back all header not only headers starting with `x-`

## Why was it done
Why restrict the header list to name starting by `x-`. You may need at some point some read header name which does not start by `x-`.


chrismccord:
Please see other issues discussing this. Thanks!

PR #3485, opened 2019-07-17
Fix links

tabeth:
Links are broken here.


Gazler:
These links are `.html` to ensure they work on hexdocs with the built docs, they won't work when browsing source docs GitHub.

tabeth:
Ah, good to know!

PR #3486, opened 2019-07-17
Broken link

tabeth:
Fix broken link.


Gazler:
These links are `.html` to ensure they work on hexdocs with the built docs, they won't work when browsing source docs GitHub.

PR #3495, opened 2019-07-25
Improve :check_origin docs

jayjun:
`:check_origin` documentation is duplicated between endpoint configuration and `socket/3`. Moreover, these settings are not the same thing.

This pull request moves all options to `socket/3` and explains that the endpoint configuration sets the default option.


josevalim:

PR #3490, opened 2019-07-22
replace Plug.Logger with Plug.Telemetry in routing guide

ignatiusreza:
this seems like a leftover from 4d61eacd83d61899a895167715d8754c5c8187be ?

PR #3497, opened 2019-07-26
Fix Cowboy WebSocket error handling

jayjun:
`conn` is wrongly nested in pattern match, causing exceptions when passed to `Cowboy2Handler.copy_resp_headers/2`.

Fixes failing tests in  #3474, #3490, #3491 and #3495.


josevalim:

PR #3498, opened 2019-07-26
Build sourcemaps by default in dev

iloveitaly:
Thanks for all of the amazing work on Pheonix! Great framework.

I can't see any reason why we wouldn't want to enable source map generation by default. What do you think?


Gazler:
Good idea, thanks. <3

PR #3499, opened 2019-07-27
Format note on hot code reloading like other notes

rawtaz:
This makes the note stand out a bit, which I feel is justified for this piece of information.


Gazler:
Thanks <3

PR #3500, opened 2019-07-28
Fix trivial ui issue in generated index.html.eex template

thojanssens:
The links "Show", "Edit" and "Delete" are sticked together. Wrap them with span to add a little space in between.

Note that the same is done in the "show" template:
https://github.com/phoenixframework/phoenix/blob/master/priv/templates/phx.gen.html/show.html.eex#L12


chrismccord:
❤️❤️❤️🐥🔥

PR #3507, opened 2019-08-01
Replace _build/dev with _build/prod in releases guide

wojtekmach:
In all cases we assemble the release with `MIX_ENV=prod`!


chrismccord:
❤️❤️❤️🐥🔥

Issue #3509, opened 2019-08-05
Syncing presences is slow in Phoenix JS due to clone()


indrekj:
State is cloned in two places:
syncState: https://github.com/phoenixframework/phoenix/blob/master/assets/js/phoenix.js#L1282
syncDiff: https://github.com/phoenixframework/phoenix/blob/master/assets/js/phoenix.js#L1323

We have a presence channels with couple of thousand of users. There are users joining and leaving quite often.

Here's a 10 second performance test in chrome:
![Screen Shot 2019-08-05 at 11 06 17](https://user-images.githubusercontent.com/3438/62448746-25826200-b771-11e9-8e12-aa1a20d523f0.png)
![Screen Shot 2019-08-05 at 11 06 49](https://user-images.githubusercontent.com/3438/62448750-274c2580-b771-11e9-85d8-5857ac523e14.png)

`value` links to the `clone` function.

With changing clone to no-op:

![Screen Shot 2019-08-05 at 11 10 07](https://user-images.githubusercontent.com/3438/62448929-9a559c00-b771-11e9-8e1b-3f1d2221a5e5.png)

The difference is huge in our case.

Do you think this could be changed? If not, what do you think about introducing a config variable like `allowStateMutations`?

@chrismccord I think you introduced clone in dffe053. The commit message tells about what but not much about why. What do you think?

dvic:
In case we need to keep clone around: is a deep clone needed? If not,  [1] suggests using Object.assign to do a shallow copy is much faster:
![image](https://user-images.githubusercontent.com/1214337/99526570-b8153780-299b-11eb-8871-6e5718ea13b4.png)

Alternatively, we could copy the lodash implementation to also improve the performance in case of a deep clone.

[1] https://frontbackend.com/javascript/what-is-the-the-fastest-way-to-deep-clone-an-object-in-javascript

Issue #3475, opened 2019-07-12
Presence onJoin callback includes old metas in newPresence var


indrekj:
### Environment

* Phoenix JS library v1.4.9

### Issue

```
presence.onJoin((id, current, newPresence) => {
});
```
**Given**: I have a user with one tab (one meta)
**When**: I update the user (from phoenix presence in elixir)
**Then**: I expect `newPresence` to include **one** meta with the updated data
**ACTUAL**: `newPresence` has **two** metas, one old and one updated

If I force onJoin to next event loop like this:

```
presence.onJoin((id, current, newPresence) => {
  setTimeout(() => {
    // now do my thing
  });
});
```

then `newPresence` includes only one meta which is the updated meta, which I actually expect.

### Digging in the code

As I understand then here https://github.com/phoenixframework/phoenix/blob/master/assets/js/phoenix.js#L1329 `state[key]` is set to `newPresence`. Then `state[key].metas.unshift(...curMetas)` adds current metas to the state[key]. This however also mutates the `newPresence` because the var was not cloned. Then `onJoin(key, currentPresence, newPresence)` is called where `newPresence` now has the old metas as well.

setTimeout works because this forces https://github.com/phoenixframework/phoenix/blob/master/assets/js/phoenix.js#L1337 to iterate over leaves which mutates newPresence to be correct again.

It seems to be that this is not a correct behavior. If it is, can you explain why?

The issue does not exist with `onSync` callback because that's called after both joins and leaves are processed. We however don't want to use `onSync` because we want unchanged presences to remain referentially equivalent which `onSync` does not provide as it clones the entire list. Our online user list has thousands of users and we want to leverage referential equivalence and also don't want to map over the whole list each time for fast react based dom updates.

indrekj:
Also, onJoin and onLeave both call a function with 3 arguments: key, currentPresence, newPresence OR leftPresence. The currentPresence however seems to be different for both of them:

onJoin currentPresence value is the same value as before the update.
onLeave currentPresence value is with join metas added and also leave metas removed.

So in onJoin case we get the state before the changes and in onLeave case we get the state after all the changes.

I'm ok submitting a PR but I'm not really sure what is the correct behavior here.

PR #3510, opened 2019-08-05
Optimize syncDiff and syncState

indrekj:
Syncing presences was very slow due to using clone(). For channels that have couple of thousand of users where users are joining/leaving and we also have updates, clone() used almost all the CPU.

onJoin and onLeave callbacks were merged into one onChange callback. This is because previous onJoin and onLeave callbacks were very confusing and this way we don't have to use the the clone function. Previously onJoin and onLeave both were called with 3 arguments: key, currentPresence and newPresence/leftPresence. The currentPresence however was the same value as before the update for onJoin, but updated presence (joins/leaves processed) for onLeave. So in onJoin case we get the state before the changes and in onLeave case we get the state after all the changes.

Also the old onJoin newPresence variables included metas that were in leaves payload which made it hard to detect a join vs an update.

New onChange callback includes the previous presence (state before the updates) and new presence (state after the updates) which makes it much easier to process changes and see what has happened.

onJoin and onLeave are still in the API because of backwards compatibility but they're marked as deprecated.

This addresses both #3509 and #3475

Site with 3K users, before:
Screen Shot 2019-08-05 at 18 29 39

and after:
Screen Shot 2019-08-05 at 18 26 20

@josevalim @chrismccord opinions?

**UPDATE 2020/02/06**: We were able to get to over 22K users in one channel in production env with this code. Just to clarify, we render only 100 most important users, others are only synced but not shown.

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -154,29 +154,26 @@
  *
  * ### Handling individual presence join and leave events
  *
- * The `presence.onJoin` and `presence.onLeave` callbacks can be used to
- * react to individual presences joining and leaving the app. For example:
+ * The `presence.onChange` callback can be used to react to individual
+ * presences joining and leaving the app. onChange can only be used when
+ * deprecated onJoin and onLeave functions are not used.
+ * For example:
  *
  * ```javascript
  * let presence = new Presence(channel)
  *
  * // detect if user has joined for the 1st time or from another tab/device
- * presence.onJoin((id, current, newPres) => {
- *   if(!current){
- *     console.log("user has entered for the first time", newPres)
+ * presence.onChange((id, oldPresence, newPresence) => {
+ *   if(!oldPresence){
+ *     console.log("user has entered for the first time", newPresence)
+ *   } else if (newPresence.metas.length === 0){
+ *     console.log("user has left from all devices", newPresence)
  *   } else {
- *     console.log("user additional presence", newPres)
+ *     console.log("old presence", oldPresence");
+ *     console.log("new presence", newPresence");
  *   }
  * })
  *
- * // detect if user has left from all tabs/devices, or is still present
- * presence.onLeave((id, current, leftPres) => {
- *   if(current.metas.length === 0){
- *     console.log("user has left from all devices", leftPres)
- *   } else {
- *     console.log("user left from a device", leftPres)
- *   }
- * })
  * // receive presence data from server
  * presence.onSync(() => {
  *   displayUsers(presence.list())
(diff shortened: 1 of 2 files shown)


chrismccord:
Nice work! Can you update to avoid the API change and fix the conflicts? Thanks!

indrekj:
> Nice work! Can you update to avoid the API change and fix the conflicts? Thanks!

@chrismccord I updated the commit. I was able to avoid breaking changes. There's more code though. Old code however can be removed when there's a new major version upgrade. I marked onJoin and onLeave as deprecated.

chrismccord:
for now, let's avoid introducing a new API and avoid the deprecations. Thanks!

indrekj:
Ok. I don't think I can make the current api faster without breaking changes unless the current behaviour is not correct and I can change that. I described in #3475 why we cannot use the old API.

We've used my version now in production now for few weeks. The old API at least for us was not usable due to issues in #3475 and also because of the performance problems.

chrismccord:
Thanks for waiting on this! I tried it out, and the current APIs appear broken on this PR. When an additional presence joins, it is always treated as joined for the first time, so something about the PR breaks existing APIs. Also trying out the new onChange API, how would you easily detect that a presence left? It seems I'd have to compare the new/old myself in onChange to figure out who left, vs the onLeave api which provides. me the left presence. I'd love to land the performance improvements, but we need to resolve a couple things first. It also appears that we could change the existing API to be mutable, and it would allow us to be backwards compatible, functionality wise? Thanks!

indrekj:
> Thanks for waiting on this! I tried it out, and the current APIs appear broken on this PR. When an additional presence joins, it is always treated as joined for the first time, so something about the PR breaks existing APIs.

Hm, interesting. Will look into it.

> Also trying out the new onChange API, how would you easily detect that a presence left?

We do it like this:
```
presence.onChange((if, _oldPresence, newPresence) => {
  if (newPresence.metas.length > 0) {
    // add or update user
  } else {
    // remove user
  }
});
```

Basically newPresence always shows the full new state about a user. If `newPresence.metas` is empty then the user left. If there are metas then the user joined or was updated. `_oldPresence` has information about the previous state though we haven't actually needed to use it. 

There's also an example in the docs in the same PR.

> It also appears that we could change the existing API to be mutable, and it would allow us to be backwards compatible, functionality wise?

I tried but I got confused about the expected behaviour. See #3475. If you can define what is the desired behaviour for onJoin and onLeave, then I might be able to look into it.

indrekj:
It's been a year and somehow I still haven't managed to get to this :disappointed: . I'll close it and if I ever get back to it, I'll re-open or create a new PR. If somebody else wants to take it over then please feel free to do it.

PR #3512, opened 2019-08-06
Camelize --web option to generators

jayjun:
It’s not clear if `--web` to generators takes a module or directory name. Both

```bash
$ mix phx.gen.html Sales User users --web Sales
```
```bash
$ mix phx.gen.html Sales User users --web sales
```

feel equally valid. However, passing `sales` generates module names like

```elixir
MyAppWeb.:"Elixir.sales.User"Controller
```

which won’t compile.

Instead of raising on bad input, accept both forms by normalising to camel case first.


aawilson:
The travis failures seem to be cowlib flaking:
```
Failed to fetch record for 'hexpm/cowlib' from registry (using cache instead)

:timeout

** (Mix) Unknown package cowlib in lockfile

The command "mix deps.get" failed and exited with 1 during .
```

chrismccord:
❤️❤️❤️🐥🔥

PR #3513, opened 2019-08-07
Add `--auto-install` flag to phx.new generator

minton:
Usage example: `mix phx.new phx_blog --auto-install`

This optional flag causes it to fetch and install dependencies without prompting the user.

---

I nearly always want to install the mix and npm dependencies for a newly generated Phoenix app. This flag allows the user to indicate they would like the dependencies installed automatically. This allows you to generate Phoenix apps "unattended". For example,

```
mix phx.new phx_blog --auto-install ; cd phx_blog ; mix ecto.create ; mix phx.server
```

### DIFF
--- a/installer/lib/mix/tasks/phx.new.ex
+++ b/installer/lib/mix/tasks/phx.new.ex
@@ -43,6 +43,8 @@ defmodule Mix.Tasks.Phx.New do
 
     * `--verbose` - use verbose output
 
+    * `--auto-install` - automatically fetch and install dependencies without prompting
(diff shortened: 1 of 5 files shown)


chrismccord:
Does `yes | mix phx.new phx_blog` work for you? Thanks!

minton:
Yes, [yes](https://en.wikipedia.org/wiki/Yes_(Unix)) works great and is much simpler. Somehow I had never come across that command before. Thanks!

PR #3514, opened 2019-08-08
Fix live reload RE pattern for live|views directory in phx.new templates

dsdshcym:
After I generated a new project with `phx.new 1.4.9`, I found that live reloading was not working when I changed a file under `live` directory. This PR is a fix for this issue.

- `{live,views}` doesn't seem to be the right way to define alternative matches in PCRE


lexmag:
Already fixed in master via 7c3979869f1f53f9e846d7e46354fe67ab98c5da.

dsdshcym:
@lexmag thanks for the notice!

PR #3516, opened 2019-08-08
Allow custom websocket transports instead of forcing Phoenix.Transpor…

byronpc:
On previous releases (1.3), it was allowed to define custom transports in the config as defined by

```
{"/socket", Phoenix.Endpoint.CowboyWebSocket, {App.Custom.WebSocket, {App.Endpoint, App.UserSocket, [...]}}}
```

Even if `App.Custom.Websocket` is defined, it's currently being forced to use `Phoenix.Transports.Websocket` unlike previous builds prior 1.4

### DIFF
--- a/lib/phoenix/endpoint/cowboy_websocket.ex
+++ b/lib/phoenix/endpoint/cowboy_websocket.ex
@@ -13,7 +13,7 @@ defmodule Phoenix.Endpoint.CowboyWebSocket do
     conn = @connection.conn(req, transport)
 
     try do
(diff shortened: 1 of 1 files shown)


Gazler:
Thanks <3! We'll merge this for Cowboy 1 for backwards compatibility, however it is worth noting that Cowboy 1 will eventually be removed from Phoenix and this technique will no longer work.

If you are defining a non-channels websocket, we now recommend using https://github.com/phoenixframework/phoenix/blob/212f8e0c576bb611d14f6e7466f2c553489e0897/lib/phoenix/socket/transport.ex#L52 instead.

PR #3518, opened 2019-08-08
Add mix/tasks folder for generated apps

iloveitaly:
I imagine most folks using phoenix are going to use mix tasks. Setting up this folder structure removes the decision/research as to where to throw the task files.

As an aside, is`mix phx.gen.task` omitted intentionally? Or would a PR with a mix task generator be a welcome addition?


chrismccord:
I think the existing structure strikes a nice balance. Folks will necessarily have to do some digging on how to write tasks unless we generate an example task for them, but I think that's overkill. We also have folks that already complain about the number of existing files/dirs so I think we're at a good balance currently; therefor I'd keep things as is. If you think this would be a nice addition to the guides, please send a guide PR. Thanks!

iloveitaly:
Makes sense, @chrismccord! Appreciate the thoughtful reply.

PR #3519, opened 2019-08-09
Explain routing in router docs

jayjun:
This pull request adds a “Routing” section to `Phoenix.Router` docs.

While Rails is a popular framework, the routing convention mimicked by Phoenix should be documented for first-time developers and those from other stacks (where regex is far more common). Features like globbing also don’t behave like Rails or Unix counterparts.

### DIFF
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -28,16 +28,71 @@ defmodule Phoenix.Router do
         get "/pages/:page", PageController, :show
       end
 
-  The `get/3` macro above accepts a request of format `"/pages/VALUE"` and
-  dispatches it to the show action in the `PageController`.
-
-  Routes can also match glob-like patterns, routing any path with a common
-  base to the same controller. For example:
-
-      get "/dynamic*anything", DynamicController, :show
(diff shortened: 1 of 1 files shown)


chrismccord:
❤️❤️❤️🐥🔥

lib/phoenix/router.ex
```diff
@@ -28,17 +28,72 @@ defmodule Phoenix.Router do
...
+
+      get "/pages", PageController, :index
+
+  matches a `GET` request to `/pages` and dispatches it to the `index` action in
+  `PageController`.
+
+      get "/pages/:page", PageController, :show
+
+  matches `/pages/hello` and dispatches to the `show` action with
+  `%{"path" => "hello"}` in `params`.
```
jarroput: @jayjun Should it not match `%{"page" => "hello"}` in `params`, as `:page` is the param name?

Gazler: @jarroput Good spot, feel free to open a PR.

jayjun: Good catch @jarroput! Fix in #3526.

PR #3521, opened 2019-08-12
Update Heroku deployment docs to include container based approach

ashkan18:
# Problem
Heroku's setup using buildpacks can get little tricky at times and is not very flexible. For example adding `ImageMagick` to our Heroku setup can get really complicated.

# Alternative Solution
Heroku also [offers `container` stack](https://devcenter.heroku.com/articles/build-docker-images-heroku-yml) which can be used to deploy apps using docker containers. In this way, we need to add a `Dockerfile` to define our image and whats needed to be built, then use `heroku.yml` to tell heroku how to run our image and what needs to be passed to it (think of it as `docker-compose`)

# Change
This PR updates the docs to also describe how we can build a container based app on heroku and set it up with phoenix.

### DIFF
--- a/guides/deployment/heroku.md
+++ b/guides/deployment/heroku.md
@@ -154,40 +154,9 @@ build:
 ```
 
 #### Setup Your app with Dockerfile
-Add `Dockerfile` to your root folder, in this example we used [`bitwalker/alpine-elixir-phoenix:latest`](https://hub.docker.com/r/bitwalker/alpine-elixir-phoenix) as our base image. Here is a sample basic `Dockerfile`:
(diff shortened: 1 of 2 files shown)


Gazler:
Thanks for the PR!

I wonder if we should refer to the existing container documentation in the releases page https://github.com/phoenixframework/phoenix/blob/master/guides/deployment/releases.md#containers for the actual Dockerfile. I think we want to start guiding people towards using releases VS mix phx.server for when we require Elixir 1.9 for Elixir.

ashkan18:
oh yes! missed that one! makes total sense! updating it now! Thanks for reviewing!

Gazler:
Thanks <3

PR #3533, opened 2019-08-23
be able to customise VSN number to allow for custom socket serializers

shdblowers:
## Description of Change

Be able to customise the VSN number used for websockets via Socket constructor options.

## Use Case

Using phoenix channels with a custom serializer.

## Context

Phoenix channels allow us to write custom [Socket Serializers](https://hexdocs.pm/phoenix/Phoenix.Socket.Serializer.html#content). There also already exists a mechanism for choosing a serializer based off of the `vsn` param in the websocket request: https://github.com/phoenixframework/phoenix/blob/v1.4.3/lib/phoenix/socket.ex#L518-L526

The problem is that currently there's no way to specify the `vsn` param the javascript client should use and it always defaults to whatever version - it always chooses its default one.

This means that if one does specify the encode and decode functions (as per https://hexdocs.pm/phoenix/1.4.9/js/index.html#socket ) it's impossible to negotiate the correct serializer with the server, unless there's a separate endpoint for it or you write some custom code in the `Phoenix.Socket` `init` callback.


shdblowers:
fixed that failing test 😅

chrismccord:
❤️❤️❤️🐥🔥

PR #3534, opened 2019-08-24
Fixed default config for new projects so that live_reload works on views again

nshafer:
A while ago the default `dev.exs` config was changed to begin incorporating Phoenix Live View, but a shell style glob `{live,views}` was added instead of a regex `(live|views)`, thus breaking the live reload of views in a new project.

### DIFF
--- a/installer/templates/phx_single/config/dev.exs
+++ b/installer/templates/phx_single/config/dev.exs
@@ -51,7 +51,7 @@ config :<%= app_name %>, <%= endpoint_module %>,
     patterns: [
       ~r"priv/static/.*(js|css|png|jpeg|jpg|gif|svg)$",
       ~r"priv/gettext/.*(po)$",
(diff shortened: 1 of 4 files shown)


chrismccord:
❤️❤️❤️🐥🔥

SylvanSign:
Howdy. I just encountered this issue with a new project I generated using Phoenix 1.4.11 (the latest release as I'm writing this message). I thought I was going crazy when the live reloading wasn't working as advertised in the docs.

Looks like this has been broken in all release versions since 1.4.3 (Mar 2019). Is it possible to prioritize getting this fix into a release soon? Live reload is a wonderful feature when it's working, but it's a bit confusing that the generator produces a broken pattern, especially for a newbie coming in and trying to get a first impression of Phoenix (and potentially LiveView).

Thanks for all the hard work!

PR #3537, opened 2019-08-26
Use 0.1.0 for the default first version in templates

lexmag:


lexmag:
@chrismccord oh, sorry, I have already sent this PR before. Just found the change in my backlog. 🙈

PR #3538, opened 2019-08-26
Inline assigns to map transformation in Phoenix.View

lexmag:
This makes heavily nested rendering ~8% faster.


chrismccord:
❤️❤️❤️🐥🔥

PR #3539, opened 2019-08-26
Optimize resource assigning in Phoenix.View

lexmag:
When profiling a heavy loaded Phoenix application that in some cases calls Phoenix.View.assign_resource/3 many times this change makes the function be 25% faster.

The downside is that it changes `as: nil` behaviour. I suppose it is not a documented behaviour and we better get a speedup.

### DIFF
--- a/lib/phoenix/view.ex
+++ b/lib/phoenix/view.ex
@@ -388,7 +388,12 @@ defmodule Phoenix.View do
   defp to_map(assigns) when is_list(assigns), do: :maps.from_list(assigns)
 
   defp assign_resource(assigns, view, resource) do
-    as = Map.get(assigns, :as) || view.__resource__
+    as =
+      case assigns do
+        %{as: as} -> as
+        _ -> view.__resource__
+      end
+
     Map.put(assigns, as, resource)
   end
 


michalmuskala:
For the `render_many` case, it would probably make sense to hoist the `as` extraction out of of the loop and only do assignment inside, since the assigns don't change between all the resources. This should improve the runtime significantly in that case.
There are probably even more optimisations like that possible for `render_many` - right now it computes layouts & other things like that on every pass as well.

lexmag:
@michalmuskala exactly, thanks! The `render_many` optimization is incoming—still measuring.
Any more places for you can think of?

michalmuskala:
Nothing immediately obvious. 

Is it for JSON endpoint or HTML? For squeezing some additional performance from JSON endpoints if you're using Jason, you could consider leveraging [`Jason.Helpers.json_map`](https://hexdocs.pm/jason/Jason.Helpers.html#json_map/1) that does some work at compile-time to avoid runtime encoding.

lexmag:
It is for MessagePack. :bowtie: 
Oh, `Jason.Helpers.json_map/1` is a smart trick, I'll see if I can apply similar approach in our current setup.

chrismccord:
❤️❤️❤️🐥🔥

PR #3544, opened 2019-09-02
Return iodata instead binary in Socket.Serializer behaviour

Nitrino:
Fix of the mismatch between the types described in the `Phoenix.Socket.Serializer` behaviour and implemented in serializers [`Phoenix.Socket.V2.JSONSerializer`](https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/socket/serializers/v2_json_serializer.ex#L21) and [`Phoenix.Socket.V1.JSONSerializer`](https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/socket/serializers/v1_json_serializer.ex#L36).


chrismccord:
❤️❤️❤️🐥🔥

PR #3548, opened 2019-09-06
websocket, add protocol header option to connect_info

danj3:
Background: sec-websocket-protocol is a header that can be set by the second arg to Websocket

This patch adds:
- an option (transport_protocol) to phoenix.js Socket that becomes arg2 to Websocket
- Websocket will set sec-websocket-protocol header when provided
- socket/transport.ex: validate_config will now accept :protocol
- socket/transport.ex: connect_info will include :protocol taken from req_headers when configured

The JS option is not relevant to LongPoll, though there is no coded defense that prohibits it.

### DIFF
--- a/lib/phoenix/socket/transport.ex
+++ b/lib/phoenix/socket/transport.ex
@@ -219,7 +219,7 @@ defmodule Phoenix.Socket.Transport do
 
     connect_info =
       Enum.map(connect_info, fn
-        key when key in [:peer_data, :uri, :x_headers] ->
+        key when key in [:peer_data, :uri, :x_headers, :protocol ] ->
           key
 
         {:session, session} ->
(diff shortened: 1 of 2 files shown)


lib/phoenix/socket/transport.ex
```diff
@@ -379,6 +379,9 @@ defmodule Phoenix.Socket.Transport do
         :uri ->
           {:uri, fetch_uri(conn)}
 
+        :protocol ->
+          { :protocol, Plug.Conn.get_req_header( conn, "sec-websocket-protocol" ) }
```
josevalim:
```suggestion
          {:protocol, Plug.Conn.get_req_header(conn, "sec-websocket-protocol")}
```

josevalim: My suggestion is to generalize this as `sec_headers`, so we can easily retrieve all security related headers instead of adding them one by one.

chrismccord:
@danj3 can you implement the suggestion José mentioned to generalize this as a `:sec_headers` option? Thanks!

lib/phoenix/socket/transport.ex
jeregrine: The websocket javascript api *only* allows you set the 'sec-websocket-protocol' during the handshake. https://html.spec.whatwg.org/multipage/web-sockets.html#network so in this case it would only make sense to do the one. There are other header values defined in the RFC https://tools.ietf.org/html/rfc6455#section-4.3 but not in the spec. 

Unless you have something else in mind @josevalim

PR #3550, opened 2019-09-08
Allow disabling logger via application configuration

hauleth:
Elixir's `Logger` is very constraining library and some people (like me)
prefer to use different approaches (like `logger` or `lager`) which
(among other things) allow for much more detailed log levels and
structured logging. If someone use such logging library then
`Phoenix.Logger` can generate needless noise in logs. This change allows
to disable this module by such users.

### DIFF
--- a/lib/phoenix.ex
+++ b/lib/phoenix.ex
@@ -44,7 +44,9 @@ defmodule Phoenix do
       :erlang.system_flag(:backtrace_depth, stacktrace_depth)
     end
 
-    Phoenix.Logger.install()
+    if Application.fetch_env!(:phoenix, :logger) do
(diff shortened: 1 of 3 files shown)


lib/phoenix.ex
```diff
@@ -44,7 +44,9 @@ defmodule Phoenix do
       :erlang.system_flag(:backtrace_depth, stacktrace_depth)
     end
 
-    Phoenix.Logger.install()
+    if Application.get_env(:phoenix, :logger, true) do
```
josevalim: The default value should be set in the `mix.exs` file and we should use `fetch_env!` here.

josevalim:

PR #3551, opened 2019-09-10
Do not overwrite `:http` configuration in `prod.secret.exs`

### LINKED ISSUE #3549: specifying that the http listener should use ipv6 should not blow away the rest of the http config
### Environment

* Elixir version (elixir -v): `Elixir 1.8.2 (compiled with Erlang/OTP 22)`
* Phoenix version (mix deps): `locked at 1.4.9 (phoenix) 746d098e`
* NodeJS version (node -v): not installed
* NPM version (npm -v): not installed
* Operating system: Alpine Linux 3.10

### Expected behavior

in prod.secret.exs:

```
config :yourapp, YourApp.Endpoint,
  http: [:inet6, port: 4000]
```

should not blow away the rest of `YourApp.Endpoint[:http]` config.

### Actual behavior

Setting inet6 blows away the rest of the endpoint config including dispatch table.

minibikini:
Currently, default (generated by the installer) `prod.secret.exs` overwrites existing `:http` config (#3549).

Default `prod.secret.exs`:

```elixir
config :demo_app, DemoAppWeb.Endpoint,
  http: [:inet6, port: String.to_integer(System.get_env("PORT") || "4000")],
  secret_key_base: secret_key_base
```

The Config deep merges keyword lists only and `[:inet6, port: ...]` is a regular list so Config overwrites `:http` value. This behavior could be inconvenient (and maybe even confusing) if you have SSL or custom dispatch configuration in `config.exs` or `prod.exs`.

After some digging in Phoenix, Plug.Cowboy, Cowboy, and Ranch the only solution I've managed to find is to set `:inet6` inside `endpoint_config[:http][:transport_options][:socket_opts]` instead of `endpoint_config[:http]`.

Fixes #3549.

### DIFF
--- a/installer/templates/phx_single/config/prod.exs
+++ b/installer/templates/phx_single/config/prod.exs
@@ -25,11 +25,11 @@ config :logger, level: :info
 #       ...
 #       url: [host: "example.com", port: 443],
 #       https: [
-#         :inet6,
 #         port: 443,
 #         cipher_suite: :strong,
 #         keyfile: System.get_env("SOME_APP_SSL_KEY_PATH"),
-#         certfile: System.get_env("SOME_APP_SSL_CERT_PATH")
+#         certfile: System.get_env("SOME_APP_SSL_CERT_PATH"),
+#         transport_options: [socket_opts: [:inet6]]
 #       ]
 #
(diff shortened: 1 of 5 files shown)


josevalim:

PR #3553, opened 2019-09-13
stateChangeCallbacks filter on channel remove

### LINKED ISSUE #3517: Memory leak in stateChangeCallbacks
### Environment

* Phoenix JS library v1.4.9

### Steps to reproduce

- Go to https://codesandbox.io/s/jolly-curie-furb6 
- Click Add Channel button few times and then remove channel
- Open devtools and check console logs

### Expected behavior

When removing a channel through `socket.remove(channel)`, socket should properly clean up any lifecycle events added by the removed channel.

### Actual behavior

```
channels count 1
Callbacks count
- open: 1
- close: 0
- error: 1
- message: 0

channels count 2
Callbacks count

- open: 2
- close: 0
- error: 2
- message: 0

channels count 1
Callbacks count

- open: 2
- close: 0
- error: 2
- message: 0

channels count 0
Callbacks count

- open: 2
- close: 0
- error: 2
- message: 0

```

As you can see the number of callbacks in open and close arrays doesn't decrease after channel is removed. This happens because in line [#L379](https://github.com/phoenixframework/phoenix/blob/master/assets/js/phoenix.js#L380) and [#L379](https://github.com/phoenixframework/phoenix/blob/master/assets/js/phoenix.js#L380) these 2 callbacks are added to the socket using anonymous function but they're not  cleaned up when removing a channel.

snewcomer:
Screen Shot 2019-09-13 at 10 32 48 PM

close #3517

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -376,11 +376,11 @@ export class Channel {
     this.rejoinTimer = new Timer(() => {
       if(this.socket.isConnected()){ this.rejoin() }
     }, this.socket.rejoinAfterMs)
-    this.socket.onError(() => this.rejoinTimer.reset())
+    this.socket.onError(() => this.rejoinTimer.reset(), topic)
     this.socket.onOpen(() => {
       this.rejoinTimer.reset()
       if(this.isErrored()){ this.rejoin() }
-    })
+    }, topic)
     this.joinPush.receive("ok", () => {
       this.state = CHANNEL_STATES.joined
       this.rejoinTimer.reset()
@@ -859,14 +859,15 @@ export class Socket {
    * @example socket.onOpen(function(){ console.info("the socket was opened") })
    *
    * @param {Function} callback
+   * @param {Function} callback
    */
-  onOpen(callback){ this.stateChangeCallbacks.open.push(callback) }
+  onOpen(callback, ref = this.ref){ this.stateChangeCallbacks.open.push({ [ref]: callback }) }
 
   /**
    * Registers callbacks for connection close events
    * @param {Function} callback
    */
-  onClose(callback){ this.stateChangeCallbacks.close.push(callback) }
+  onClose(callback, ref = this.ref){ this.stateChangeCallbacks.close.push({ [ref]: callback }) }
 
   /**
    * Registers callbacks for connection error events
@@ -875,13 +876,13 @@ export class Socket {
    *
    * @param {Function} callback
    */
-  onError(callback){ this.stateChangeCallbacks.error.push(callback) }
+  onError(callback, ref = this.ref){ this.stateChangeCallbacks.error.push({ [ref]: callback }) }
 
   /**
    * Registers callbacks for connection message events
    * @param {Function} callback
    */
-  onMessage(callback){ this.stateChangeCallbacks.message.push(callback) }
+  onMessage(callback, ref = this.ref){ this.stateChangeCallbacks.message.push({ [ref]: callback }) }
 
   /**
    * @private
@@ -893,7 +894,7 @@ export class Socket {
     this.flushSendBuffer()
     this.reconnectTimer.reset()
     this.resetHeartbeat()
-    this.stateChangeCallbacks.open.forEach( callback => callback() )
+    this.stateChangeCallbacks.open.forEach(entry => entry[Object.keys(entry)[0]]() )
   }
 
   /**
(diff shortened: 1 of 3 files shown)


assets/js/phoenix.js
```diff
@@ -893,7 +894,7 @@ export class Socket {
     this.flushSendBuffer()
     this.reconnectTimer.reset()
     this.resetHeartbeat()
-    this.stateChangeCallbacks.open.forEach( callback => callback() )
+    this.stateChangeCallbacks.open.forEach(entry => entry[Object.keys(entry)[0]]() )
```
snewcomer: ugh this doesn't look friendly.  We could go with a different data structure.  Right now this is an array of objects.

We could go with a "tuple" `open = [[topic, callback], [topic, callback]]`

assets/js/phoenix.js
```diff
@@ -859,14 +859,15 @@ export class Socket {
    * @example socket.onOpen(function(){ console.info("the socket was opened") })
    *
    * @param {Function} callback
+   * @param {Function} callback
    */
-  onOpen(callback){ this.stateChangeCallbacks.open.push(callback) }
+  onOpen(callback, ref = this.ref){ this.stateChangeCallbacks.open.push({ [ref]: callback }) }
```
snewcomer: This default would be for non internal calls to this method since it isn't a required API argument

assets/js/phoenix.js
chrismccord: the tuple works, or we can do an object: `{ref, callback}`, your call

assets/js/phoenix.js
```diff
@@ -875,13 +876,13 @@ export class Socket {
    *
    * @param {Function} callback
    */
-  onError(callback){ this.stateChangeCallbacks.error.push(callback) }
+  onError(callback, ref = this.ref){ this.stateChangeCallbacks.error.push({ [ref]: callback }) }
```
chrismccord: we need to return the ref to the caller, and introduce the `sockert.off` function. We also need to generate a new unique ref if one is not passed, ie:

```suggestion
  onError(callback, ref = this.makeRef()){ 
    this.stateChangeCallbacks.error.push({ref, callback})
    return ref
  }

```

assets/js/phoenix.js
```diff
@@ -967,6 +968,11 @@ export class Socket {
    */
   remove(channel){
     this.channels = this.channels.filter(c => c.joinRef() !== channel.joinRef())
+
+    this.stateChangeCallbacks.open = this.stateChangeCallbacks.open.filter(entry => {
+      const topic = Object.keys(entry)[0]
+      return topic !== channel.topic
+    })
```
chrismccord: instead of stateChangeCallbacks filter, we can simply call `off` before our existing channels filter:

```javascript
  remove(channel){
    this.off(channel.topic)
    this.channels = this.channels.filter(c => c.joinRef() !== channel.joinRef())
  }
```

assets/js/phoenix.js
```diff
@@ -376,11 +376,11 @@ export class Channel {
     this.rejoinTimer = new Timer(() => {
       if(this.socket.isConnected()){ this.rejoin() }
     }, this.socket.rejoinAfterMs)
-    this.socket.onError(() => this.rejoinTimer.reset())
+    this.socket.onError(() => this.rejoinTimer.reset(), topic)
```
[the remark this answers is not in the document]
chrismccord: to avoid clashes, let's keep the on[..] callbacks accepting a single function argument, and we will always generate and return the unique reference. This will ensure no reference clashes, and `off` will accept a list of refs to drop. This will require the channel to track its state change refs, but that will be pretty simple:

```diff
class Channel {
  constructor(...){
    ...
+   this.stateChangeRefs = []
  }
  ...
-  this.socket.onOpen(...)
+  this.stateChangeRefs.push(this.socket.onOpen(...))
-  this.socket.onClose(...)
+  this.stateChangeRefs.push(this.socket.oClose(...))

File: assets/js/phoenix.js (hunk 2 above)
Comment (chrismccord) [reply assets/js/phoenix.js]: let's kill the ref argument and always generate one with `this.makeRef()`, then well return it to the caller to optionally use for `off/1`

File: assets/js/phoenix.js [hunk 6]
```diff
@@ -966,9 +984,22 @@ export class Socket {
...
+  /**
+   * stateChangeCallbacks is an array of "tuples" [[ref, callback], [...]]
+   * and can be cleaned up when a channel goes offline
+   * @param {Channel}
+   */
+  off(channel) {
+    let refs = channel.stateChangeRefs
+    this.stateChangeCallbacks.open = this.stateChangeCallbacks.open.filter(([ref]) => {
+      return !refs.includes(ref)
+    })
```
snewcomer: I'm a follow up (or current) addition to this PR is cleaning up other containers for callbacks like `error`...

PR #3556, opened 2019-09-18
Merge arguments from supervisor into endpoint configuration

hauleth:
This allows passing additional configuration options in supervisor spec, ex.:

```elixir
children = [
  {MyApp.Endpoint, url: [path: "/api"]}
]
```

Will be the same as setting:

```elixir
config :my_app, MyApp.Endpoint, url: [path: "/api"]
```

---

Configuration still overwrite values set in supervisor, so:

```elixir
children = [
  {MyApp.Endpoint, url: [path: "/foo"]}
]
```

and

```elixir
config :my_app, MyApp.Endpoint, url: [path: "/bar"]
```

Will result in:

```elixir
assert "/bar" == MyApp.Endpoint.config(:url)[:path]
```

### DIFF
--- a/lib/phoenix/config.ex
+++ b/lib/phoenix/config.ex
@@ -70,13 +70,17 @@ defmodule Phoenix.Config do
   Useful to read a particular value at compilation time.
   """
   def from_env(otp_app, module, defaults) do
-    otp_app
-    |> fetch_config(module)
-    |> merge(defaults)
+    config = fetch_config(otp_app, module)
+
+    merge(defaults, config)
   end
 
-  @doc false
-  def merge(a, b), do: Keyword.merge(b, a, &merger/3)
+  @doc """
+  Take 2 keyword lists and merge them recursively.
+
+  Used to merge configuration values into defaults.
+  """
+  def merge(a, b), do: Keyword.merge(a, b, &merger/3)
 
   defp fetch_config(otp_app, module) do
     case Application.fetch_env(otp_app, module) do
--- a/lib/phoenix/config.ex
+++ b/lib/phoenix/config.ex
@@ -70,8 +70,13 @@ defmodule Phoenix.Config do
   Useful to read a particular value at compilation time.
   """
   def from_env(otp_app, module, defaults) do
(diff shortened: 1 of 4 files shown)


lib/phoenix/config.ex
```diff
@@ -70,9 +70,14 @@ defmodule Phoenix.Config do
   Useful to read a particular value at compilation time.
   """
   def from_env(otp_app, module, defaults) do
-    Keyword.merge(defaults, fetch_config(otp_app, module), &merger/3)
+    otp_app
+    |> fetch_config(module)
+    |> merge(defaults)
   end
 
+  @doc false
```
josevalim: Please document it accordingly. :) Even if used only internally.

lib/phoenix/config.ex
```diff
@@ -70,9 +70,14 @@ defmodule Phoenix.Config do
   Useful to read a particular value at compilation time.
   """
   def from_env(otp_app, module, defaults) do
-    Keyword.merge(defaults, fetch_config(otp_app, module), &merger/3)
+    otp_app
+    |> fetch_config(module)
+    |> merge(defaults)
```
josevalim: Reversing the order of merge but calling it merge is potentially very confusing. Or we keep the order and swap arguments here (non-pipeable) or we call it something like `reverse_merge`. I prefer the former.

hauleth: Maybe name it `override` or `into_defaults`.

lib/phoenix/config.ex
```diff
@@ -72,11 +72,16 @@ defmodule Phoenix.Config do
...
     |> fetch_config(module)
-    |> merge(defaults)
+    |> override(defaults)
   end
 
-  @doc false
-  def merge(a, b), do: Keyword.merge(b, a, &merger/3)
+  @doc """
+  Take 2 keyword lists and override values from first one with values from
+  the second one recursively.
```
josevalim: You are describing merge here, and not overrides. I think this is a strong sign we should just go with merge please? Less chance of confusion. :)

josevalim:

PR #3560, opened 2019-09-25
Update Dockerfile releases.md

pera:
The rel directory may not exist, and there was no previous reference to mix release.init either. Since COPY requires the existence of src, it would stop the execution of docker build in the common case of not having a rel dir.


josevalim:
Fixed in another PR, thank you! :heart:

PR #3566, opened 2019-10-05
added interactive installer

mithereal:
i created a simple interactive installer because i keep forgetting the flags


josevalim:
Thanks @mithereal but we would prefer to keep the installer simple. If you forget the flags, you can always delete the old directory and start again (as I usually do). :)

Issue #3572, opened 2019-10-11
Removing listener might make the next in line listener not being called


ostap0207:
* Phoenix JS version: 1.4.9

In the following example, the second listener will never be called when the "event" comes.

```javascript
var ref = channel.on("event", () => {
 channel.off("event", ref);
});

channel.on("event", () => {
 // never called
});
```

This is happening as the listeners being processed one by one in a `for` loop, which does not take to account a case when the listener is removed during the processing and before the triggering of the next listener.
https://github.com/phoenixframework/phoenix/blob/master/assets/js/phoenix.js#L606

The code example above is quite often used in a different way with RxJS.

```javascript
Rx.Observable.fromEvent(channel, "event")
  .take(1)
  .subscrbe(() => {});
```

This will only take one event and remove the listener before other listeners have been triggered.

### Expected behavior
All listeners are called even if one of the listeners is removed during processing

### Actual behavior
The next in line listener will not be called if one of the listeners is removed during processing

snewcomer:
@ostap0207 Interesting observation!  I think what you are advocating for is the bound array under process is not mutated?  Something like `Array.from` or `slice()` handle these needs?

Also, do you happen to have a use case for triggering all of the listeners, even if the array was further processed during one of the iterations?  I also see being able to stop listeners down the chain as beneficial (but have no specific reason why ;) ).  Lmk your thoughts!

ostap0207:
> Something like Array.from or slice() on the bound array before processing handle these needs?

Yes. I had a look at socket.io and they use `slice()`

https://github.com/component/emitter/blob/6bd7817e8a444cb16e8abdf7dd2d7f04d5ca3dc8/index.js#L143

I think only those listeners that are available when the event comes should be called, and any modifications to the array should not affect that. Often the listeners could be in different parts of the application and I think their logic should be independent of other listeners for the same event.
Otherwise, listeners' logic might get coupled and depend on the order of invocation, which sounds prone to errors.

snewcomer:
Great point on `socket.io` 💯 In principle, I agree with you (just having a hard time coming up with a logical situation).  The only thing that I am unsure of is in your example, you provided

```js
var ref = channel.on("event", () => {
 channel.off("event", ref);
});
```

It seems that the intention was to prevent further calls to some event.  

In any case, that seems like undesirable and potentially error prone behaviour, so a PR would be appreciated it you had time!

Emittery does this...
[...]

PR #3576, opened 2019-10-14
Ensure all event listeners are called

ostap0207:
Fixes: https://github.com/phoenixframework/phoenix/issues/3572

If we have 2 different listeners and one of them removes the listener
in its body then the second listener is never called:

```javascript
var ref = channel.on("event", () => {
 channel.off("event", ref);
});

channel.on("event", () => {
 // never called
});
```

This is because the listeners array is used by reference and all the
changes to it are visible before all the listeners are called. Thus, if
one of the listeners is removed then since `i` has not change the next
in line listener is skipped.

To fix this copy the array of listeners before calling them. This
approach is used in many libraries from what I saw ([1], [2]).

[1]: https://github.com/component/emitter/blob/6bd7817e8a444cb16e8abdf7dd2d7f04d5ca3dc8/index.js#L143
[2]: https://github.com/sindresorhus/emittery/blob/3334cff4daf83bfcf7054d35edc51957e3f70f8d/index.js#L220

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -603,9 +603,10 @@ export class Channel {
     let handledPayload = this.onMessage(event, payload, ref, joinRef)
     if(payload && !handledPayload){ throw new Error("channel onMessage callbacks must return the payload, modified or unmodified") }
 
-    for (let i = 0; i < this.bindings.length; i++) {
-      const bind = this.bindings[i]
-      if(bind.event !== event){ continue }
+    const eventBindings = this.bindings.filter(bind => bind.event === event)
+
+    for (let i = 0; i < eventBindings.length; i++) {
+      const bind = eventBindings[i]
       bind.callback(handledPayload, ref, joinRef || this.joinRef())
     }
   }
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -603,7 +603,7 @@ export class Channel {
     let handledPayload = this.onMessage(event, payload, ref, joinRef)
     if(payload && !handledPayload){ throw new Error("channel onMessage callbacks must return the payload, modified or unmodified") }
 
-    const eventBindings = this.bindings.filter(bind => bind.event === event)
+    let eventBindings = this.bindings.filter(bind => bind.event === event)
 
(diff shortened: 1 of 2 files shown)


ostap0207:
I would like to point out that although this fixes a bug this could potentially break things for people who were relying on a fact that one listener can remove one or all of the following listeners during the processing of the event.

The other approach would be to still clone the array, but double-check that the listener is still present, similarly to https://github.com/sindresorhus/emittery/blob/3334cff4daf83bfcf7054d35edc51957e3f70f8d/index.js#L226.

However, this approach requires a lookup for each listener, which in theory could performance issues depending on how many listeners each event has.

assets/js/phoenix.js
```diff
@@ -603,9 +603,10 @@ export class Channel {
     let handledPayload = this.onMessage(event, payload, ref, joinRef)
     if(payload && !handledPayload){ throw new Error("channel onMessage callbacks must return the payload, modified or unmodified") }
 
-    for (let i = 0; i < this.bindings.length; i++) {
-      const bind = this.bindings[i]
-      if(bind.event !== event){ continue }
+    const eventBindings = this.bindings.filter(bind => bind.event === event)
```
chrismccord:
```suggestion
    let eventBindings = this.bindings.filter(bind => bind.event === event)
```

assets/js/phoenix.js
```diff
@@ -603,9 +603,10 @@ export class Channel {
     let handledPayload = this.onMessage(event, payload, ref, joinRef)
     if(payload && !handledPayload){ throw new Error("channel onMessage callbacks must return the payload, modified or unmodified") }
 
-    for (let i = 0; i < this.bindings.length; i++) {
-      const bind = this.bindings[i]
-      if(bind.event !== event){ continue }
+    let eventBindings = this.bindings.filter(bind => bind.event === event)
+
+    for (let i = 0; i < eventBindings.length; i++) {
+      const bind = eventBindings[i]
```
chrismccord:
```suggestion
      let bind = eventBindings[i]
```

chrismccord:
❤️❤️❤️🐥🔥

PR #3582, opened 2019-10-18
Allow using `put_new_layout` and `put_new_view` in controller pipeline

hauleth:
Earlier the layout and view were first things done within the pipeline, this caused problems when someone (like me) wanted to set the default template on their own (I am using custom naming scheme in form of `MyApp.Controllers.Foo` and view is named `MyApp.Views.Foo`). As Phoenix.Controller.Pipeline isn't public in any way I couldn't omit `use Phoenix.Controller` and I was forced to use `put_layout` and `put_view` which have nasty behaviour that it always overwrite the layout and view, this meant that I couldn't longer define layout/view higher in the Plug pipeline. This change put the layout/view definition as the last thing in the pipeline, which allows to define them within controller pipeline using `put_new_layout` and `put_new_view` functions.

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -167,6 +167,13 @@ defmodule Phoenix.Controller do
 
       use Phoenix.Controller.Pipeline, opts
 
+      @before_compile Phoenix.Controller
+    end
+  end
+
+  @doc false
+  defmacro __before_compile__(_env) do
+    quote do
       plug :put_new_layout, {Phoenix.Controller.__layout__(__MODULE__, opts), :app}
       plug :put_new_view, Phoenix.Controller.__view__(__MODULE__)
     end


josevalim:
Can we have a test about the new behaviour to avoid regressions in the future please? thanks!

hauleth:
After second thoughts it can be breaking change if someone did:

```
defmodule MyApp.FooController do
  use Phoenix.Controller

  plug :foo, []

  defp foo(conn, _opts) do
    view = view_model(conn)

    # earlier `view` would contain view module generated by the `Phoenix.Controller`
    # but now it will fail(?) or return `nil` as view is not yet specified

    conn
  end
end
```

So now I am not so keen for this change, but I would like to hear your opinion. For now I have workaround to my problem and I define custom `action/2` callback that will call `put_new_layout/2` and `put_new_view/2` before firing whole pipeline.

Question is whether this is acceptable breaking change or not. I cannot find any guarantees for the ordering of the plugs in the controller, but that can be surprising for some users if they relied on such order.

josevalim:
That's  a very good point. Now that you say it, I *think* we actually did it after and we moved it to before exactly because you would have no visibility into what would be rendered in the pipeline.

chrismccord:
I see the issue with the existing setup, but given this is the first time this problem has been raised, I'd prefer to wait to see if this is a larger issue. Your own plugs are still free to `put_new_layout`, but it will require the user to set their own layout upstream if you wish to read the value. I may revisit this in the future. Thanks!

PR #3584, opened 2019-10-19
Add option to install Postgres.app

kyle-nweeia:
I received a `DBConnection.ConnectionError` after following the instructions in Installation, the Up and Running Guide, and running `mix ecto.create`. I resolved the problem by installing Postgres.app and starting the Postgres server before running Ecto or Phoenix. If the error is not a bug, this solution could be presented as an option for Mac users to get up and running.

### DIFF
--- a/guides/introduction/installation.md
+++ b/guides/introduction/installation.md
@@ -90,6 +90,8 @@ When we work with Ecto schemas in these guides, we will use PostgreSQL and the P
 
 Postgrex is a direct Phoenix dependency, and it will be automatically installed along with the rest of our dependencies as we start building our app.
 
+Mac users can optionally install [Postgres.app](https://postgresapp.com) to automatically start PostgreSQL.
+
 ### inotify-tools (for linux users)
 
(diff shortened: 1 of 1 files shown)


guides/introduction/installation.md
```diff
@@ -90,6 +90,8 @@ When we work with Ecto schemas in these guides, we will use PostgreSQL and the P
 
 Postgrex is a direct Phoenix dependency, and it will be automatically installed along with the rest of our dependencies as we start building our app.
 
+Mac users can optionally install [Postgres.app](https://postgresapp.com) to automatically start PostgreSQL.
```
mhinz: Isn't this redundant if a link to the installation guides (which also mention Postgres.app) is given just a few lines above? Yes, installing alone won't be enough, Postgres needs to be started as well. This could be explicitly stated.

Maybe improving the wording of the paragraph above is the better solution here. For the sake of keeping the guide short and as platform-agnostic as possible. :)

kyle-nweeia: You're right, it's not necessary to add OS-specific instructions in order to add clarity. I'll make a different edit in another PR.

kelvinst:
I think this task can be closed ask @kyle-nweeia already sent a better one: #3585

PR #3592, opened 2019-10-21
Checks force_ssl has not changed from compile time to runtime

### LINKED ISSUE #3586: `force_ssl` doesn't work in releases.exs
We're trying to enforce ssl and following the [docs](https://hexdocs.pm/phoenix/endpoint.html#force-ssl), we added the `force_ssl` config, but we are also using elixir releases, so we (arbitrarily) put it in our releases.exs file instead of our prod.exs file and it is silently ignored.

I think this happens because `Plug.SSL` is injected by a macro seen [here](https://github.com/phoenixframework/phoenix/blob/029098e38ca830c35e832b44e58a0cd641fe24f1/lib/phoenix/endpoint.ex#L453). Since releases.exs is for runtime configs, the `force_ssl` configuration isn't present during build time when the macros are evaluated and `Plug.SSL` is never injected.

I'm not sure if "fixing" this is the right way to go or if we should just update the documentation to mention that the config must not be in a runtime config such as releases.exs. 

I'm happy to help make the change if you let me know how you'd like to proceed. Thanks again for a great library!

* Elixir version (elixir -v): 1.9.2
* Phoenix version (mix deps): 1.4.10

jesseshieh:
Resolves #3586

### DIFF
--- a/lib/phoenix/endpoint/supervisor.ex
+++ b/lib/phoenix/endpoint/supervisor.ex
@@ -11,44 +11,56 @@ defmodule Phoenix.Endpoint.Supervisor do
   Starts the endpoint supervision tree.
   """
   def start_link(otp_app, mod, opts \\ []) do
-    with {:ok, _} <- check_force_ssl(otp_app, mod),
-         {:ok, _} = ok <- Supervisor.start_link(__MODULE__, {otp_app, mod, opts}, name: mod) do
-      warmup(mod)
-      log_access_info(otp_app, mod)
-      ok
-    else
+    case Supervisor.start_link(__MODULE__, {otp_app, mod, opts}, name: mod) do
+      {:ok, _} = ok ->
+        warmup(mod)
+        log_access_info(otp_app, mod)
+        ok
+
       {:error, _} = error ->
         error
     end
   end
 
-  defp check_force_ssl(otp_app, mod) do
-    compile_config = mod.__compile_config__() |> Keyword.get(:force_ssl)
+  defp check_compile_configs(mod, runtime_configs) do
+    compile_configs = mod.__compile_config__()
 
-    runtime_config =
-      config(otp_app, mod)
-      |> Keyword.take(Phoenix.Endpoint.Supervisor.compile_config_keys())
-      |> Keyword.get(:force_ssl)
+    bad_keys =
+      Phoenix.Endpoint.Supervisor.compile_config_keys()
+      |> Enum.map(fn key ->
+        compile_config = Keyword.get(compile_configs, key)
+        runtime_config = Keyword.get(runtime_configs, key)
 
-    if compile_config != runtime_config do
-      require Logger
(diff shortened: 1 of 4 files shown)


jesseshieh:
@josevalim no rush, but just letting you know this is here whenever you get a chance. Happy to make changes, just lemme know.

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -11,17 +11,47 @@ defmodule Phoenix.Endpoint.Supervisor do
...
       {:error, _} = error ->
         error
     end
   end
 
+  defp check_force_ssl(otp_app, mod) do
+    compile_config = mod.__compile_config__() |> Keyword.get(:force_ssl)
+
+    runtime_config =
+      config(otp_app, mod)
```
josevalim: I don't think we should build the config list again, as we already do on init. My suggestion is to move this logic to inside the `init` function.

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -11,17 +11,47 @@ defmodule Phoenix.Endpoint.Supervisor do
...
+      log_access_info(otp_app, mod)
+      ok
+    else
       {:error, _} = error ->
         error
     end
   end
 
+  defp check_force_ssl(otp_app, mod) do
+    compile_config = mod.__compile_config__() |> Keyword.get(:force_ssl)
```
josevalim: Instead of validating only `force_ssl`, we can simply validate all keys in `compile_config`. That was the goal with making it generic. In a nutshell, you can traverse this keyword list, comparing it with the values in runtime_config.

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -433,4 +463,8 @@ defmodule Phoenix.Endpoint.Supervisor do
       Logger.info("Access #{inspect(endpoint)} at #{endpoint.url()}")
     end
   end
+
```
josevalim: Please add a tiny documentation sayings its purpose.

josevalim:
@jesseshieh glad you pinged me as I don't receive notifications. :) It looks really good, just some ccomments and we should be good to go!

lib/phoenix/endpoint/supervisor.ex
jesseshieh: Oh good catch. Moved.

lib/phoenix/endpoint/supervisor.ex
jesseshieh: Oh I see, sorry about that. I wasn't sure if the error message would be generic enough for future keys, but I think it should be alright. Updated.

lib/phoenix/endpoint/supervisor.ex
jesseshieh: Added.

josevalim:

PR #3593, opened 2019-10-25
Delete struct duplicate field 'route'

qrede:


Gazler:
Good spot. Have opened a PR to Elixir to detect this https://github.com/elixir-lang/elixir/pull/9452 to detect this in future.

Issue #3579, opened 2019-10-16
Need ability to exclude XXX.Router.Helpers from documentation, any advise how to do it?


ArturTTN:
Need ability to exclude XXX.Router.Helpers from documentation, any advise how to do it?

_Originally posted by @ArturTTN in https://github.com/phoenixframework/phoenix/issues/2580#issuecomment-542730567_

sushant12:
Why would you want to do it?

josevalim:
It is currently not possible but if someone wants to submit a PR that allows an option on `use Phoenix.Router` that disables it, it would be welcome.

chrismccord:
I don't consider the documentation problematic, but as José said, a PR is welcome. Thanks!

PR #3596, opened 2019-10-26
Adding option to disable `Router.Helpers` docs to `use Phoenix.Router`

kelvinst:
You said a PR for this was welcome on the issue #3579, so here it is. 

I am still trying to figure out how to test it, since I can't use `Code.fetch_docs/1` on in memory modules. Tried to do the same it's done on [docs tests on elixir repo](https://github.com/kelvinst/elixir/blob/642a794ace79f8098187034f75455bf95de1e7fc/lib/elixir/test/elixir/kernel/docs_test.exs#L99-L121), but the `write_beam` function just writes the beam for the given module, and if I do that for the router, it will save the beam for it, but not for the helper that's defined altogheter.

Any idea? Maybe there is a way to get the helper from it to save its beam, IDK.

### DIFF
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -246,16 +246,18 @@ defmodule Phoenix.Router do
   @http_methods [:get, :post, :put, :patch, :delete, :options, :connect, :trace, :head]
 
   @doc false
-  defmacro __using__(_) do
+  defmacro __using__(opts) do
     quote do
-      unquote(prelude())
+      unquote(prelude(opts))
       unquote(defs())
       unquote(match_dispatch())
     end
   end
 
-  defp prelude() do
+  defp prelude(opts) do
     quote do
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks @kelvinst! Can you please call the option `helpers_moduledoc`? That's because we only really need to set the moduledoc to false, the other docs can stay as is. then it looks good to go!

josevalim:
Ping!

chrismccord:
ping @kelvinst :)

chrismccord:
Another ping for @kelvinst :)

kelvinst:
Done! Thanks so much for reminding me.

josevalim:

anatoliip33:
hello, how to use helpers_moduledoc? do you have examples ?

anatoliip33:
found `use Phoenix.{Router}, helpers_moduledoc: false` , thanks)

PR #3599, opened 2019-10-30
Switch to https for phoenixframework.org website

spk:
Hi, this just switch to https on docs, installation templates etc for phoenixframework.org, cheers! Laurent


Gazler:
Good idea.

Thanks <3

PR #3604, opened 2019-11-06
Fix 404 on "Using MySQL" link

H12:
Using the `.html` file extension here led to a 404 when trying to access the "Using MySQL" section of the readme.

### DIFF
--- a/guides/ecto.md
+++ b/guides/ecto.md
@@ -11,7 +11,7 @@ Newly generated Phoenix projects include Ecto with the PostgreSQL adapter by def
 
 For a thorough, general guide for Ecto, check out the [Ecto getting started guide](https://hexdocs.pm/ecto/getting-started.html). For an overview of all Ecto specific mix tasks for Phoenix, see the [mix tasks guide](phoenix_mix_tasks.html#ecto-specific-mix-tasks).
 
-This guide assumes that we have generated our new application with Ecto integration and that we will be using PostgreSQL. For instructions on switching to MySQL, please see the [Using MySQL](ecto.md#using-mysql) section.
+This guide assumes that we have generated our new application with Ecto integration and that we will be using PostgreSQL. For instructions on switching to MySQL, please see the [Using MySQL](#using-mysql) section.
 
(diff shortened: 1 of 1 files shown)


guides/ecto.md
```diff
@@ -11,7 +11,7 @@ Newly generated Phoenix projects include Ecto with the PostgreSQL adapter by def
 
 For a thorough, general guide for Ecto, check out the [Ecto getting started guide](https://hexdocs.pm/ecto/getting-started.html). For an overview of all Ecto specific mix tasks for Phoenix, see the [mix tasks guide](phoenix_mix_tasks.html#ecto-specific-mix-tasks).
 
-This guide assumes that we have generated our new application with Ecto integration and that we will be using PostgreSQL. For instructions on switching to MySQL, please see the [Using MySQL](ecto.html#using-mysql) section.
+This guide assumes that we have generated our new application with Ecto integration and that we will be using PostgreSQL. For instructions on switching to MySQL, please see the [Using MySQL](ecto.md#using-mysql) section.
```
wojtekmach: it needs to stay as .html because otherwise it will 404 when rendered: https://hexdocs.pm/phoenix/ecto.html#using-mysql

H12: Huh interesting. I'll go ahead and close this PR then.

Thanks for the quick response, & sorry for the PR bloat.

wojtekmach: Oh, I just realised that because we're already on this page if we change it like this:

```markdown
[Using MySQL](#using-mysql)
```

it will work in both cases :D

H12: Oh! In that case, I'll go ahead and re-open and make that change!

H12: If you would prefer this to be one commit instead of two I'll be able to squash in about an hour when I'm at computer configured for my H12 GitHub account (I'm currently on a non-personal machine, and limited to making changes through the browser >.<).

wojtekmach: maintainers have an option to squash when merging so no need to worry about it!

josevalim:

PR #3609, opened 2019-11-11
Bump phoenix_html.js to v2.13.3

kodnin:
When generating an app with the `--no-webpack` option, delete links don't work with _v2.10.0_ of the mentioned file.


josevalim:

kodnin:
Thanks @josevalim!

PR #3612, opened 2019-11-12
Revert "Switch to https for phoenixframework.org website (#3599)"

lau:
I followed the link in the README file to the website and got an error. Try it https://www.phoenixframework.org/

This reverts commit e37c47d887363f4c0d902a0acac8cd3015cf396f.

The reason is that the website currently does not work when visiting
https://www.phoenixframework.org/ due to the SSL certificate.

This revert could be reverted after the issue with the website is fixed. Also see https://github.com/phoenixframework/phoenix_site/issues/47


Gazler:
Good spot. Could we use https://phoenixframework.org/ instead - I'll look into getting https://www.phoenixframework.org working with GH pages.

lau:
Looks like it's fixed now. Cool :)

PR #3616, opened 2019-11-21
Fix error handling when joining a channel using LongPoll

### LINKED ISSUE #3590: Phoenix.js channel.join error handling with LongPoll transport is broken
### Environment

* Elixir version (elixir -v): Elixir 1.9.1 / Erlang/OTP 22
* Phoenix version (mix deps): 1.5.0-dev (ba55f7e)
* NodeJS version (node -v): 10.16.3 (not important)
* NPM version (npm -v): 6.11.3 (not important)

### Issue

With WebSocket (OK behaviour):
**Given**: I have a channel that returns `{:error, :channel_disabled}` on join
**When**: User joins the channel with **WebSocket** transport
**Then** : `channel.join(...).receive('error')` is triggered where we can handle the error and decide if we want to reconnect or not.

With LongPoll:
**Given**: I have a channel that returns `{:error, :channel_disabled}` on join
**When**: User joins the channel with **LongPoll** transport
**Then** : `channel.join(...).receive('error')` is never triggered (**NOT OK**). Instead the whole transport reconnect to the PubSub. In our case we return an error again and so it basically starts reconnecting in a loop.

I did some investigation and it seems that if channel join fails then LongPoll Transport receives non 200 status in `send()` function and that invokes closeAndRetry on the transport. https://github.com/phoenixframework/phoenix/blob/master/assets/js/phoenix.js#L1166-L1167 To be exact then the status is 401 whenever an error is returned from a join. I think that's because of this line: https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/transports/long_poll.ex#L90

### Summary

Channel returning an error in `join` (server side) terminates longpoll connection (client side) instead of invoking `error` listener on the channel join. This affects only LongPoll. WebSocket is fine.

indrekj:
This fixes #3590

With **WebSocket** (OK behaviour):
**Given**: I have a channel that returns `{:error, :channel_disabled}` on join
**When**: User joins the channel with WebSocket transport
**Then**: `channel.join(...).receive('error')` is triggered where we can handle the error and decide if we want to reconnect or not.

With **LongPoll** (NOT OK behaviour):
**Given**: I have a channel that returns `{:error, :channel_disabled}` on join
**When**: User joins the channel with LongPoll transport
**Then**: `channel.join(...).receive('error')` is never triggered (NOT OK). Instead the whole transport reconnects to the PubSub. In our case we return an error again and so it basically starts reconnection loop.

This happened because `LongPoll` returned 401 whenever channel returned
something that was not `{:ok, socket}`. phoenix.js [ajax request][1] has:
```
      if(!resp || resp.status !== 200){
        this.onerror(resp && resp.status)
        this.closeAndRetry()
      }
```
This basically means that when it got 401 then it called
`closeAndRetry` on the whole transport.

`LongPoll` now returns 200 even if the join failed. This way we don't
trigger closeAndRetry. The actual error message is still received by
polling. `phoenix.js` now successfully triggers the join error callback
like with WebSocket.

[1]: https://github.com/phoenixframework/phoenix/blob/bf395cbeafae0f212e8c96c99c173d0bdb278f58/assets/js/phoenix.js#L1164-L1169

### DIFF
--- a/lib/phoenix/transports/long_poll.ex
+++ b/lib/phoenix/transports/long_poll.ex
@@ -87,7 +87,7 @@ defmodule Phoenix.Transports.LongPoll do
 
     receive do
       {:ok, ^ref} -> :ok
-      {:error, ^ref} -> :unauthorized
+      {:error, ^ref} -> :ok
     after
       opts[:window_ms] -> :request_timeout
     end
--- a/test/phoenix/integration/long_poll_channels_test.exs
+++ b/test/phoenix/integration/long_poll_channels_test.exs
@@ -335,7 +335,7 @@ defmodule Phoenix.Integration.LongPollChannelsTest do
         topic: "room:lobby"
       }
 
-      # Publish unauthorized event
+      # Publish event to an unjoined room
       capture_log fn ->
         Phoenix.PubSub.subscribe(__MODULE__, "room:private-room")
         resp = poll :post, "/ws", @vsn, session, %{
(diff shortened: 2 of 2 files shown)


chrismccord:
❤️❤️❤️🐥🔥

PR #3618, opened 2019-11-22
Load cache manifest from specified application

robinvdvleuten:
Within my current Umbrella project, I have multiple applications sharing the same assets. That's why I've created an "empty" application only containing the assets and the cache manifest. Configuring the correct app was easy for `Plug.Static` through the `:from` option. But the `:cache_static_manifest` option could only use a directory within the app itself. 

This PR adds the option to define the application from where the path to the cache manifest must be resolved in a similar fashion as the `Plug.Static` option. Also a normal file path still works so no breaking backwards compatibility.


robinvdvleuten:
Did someone find the time to take a peek at this PR and give some feedback, that would be nice :)

josevalim:

PR #3627, opened 2019-11-30
Added ability to filter out endpoint logs

akoutmos:
The purpose of this change is to introduce the ability to filter out logs for endpoints (configured by the user). This is useful for reducing the noise in logs when using tools like Prometheus and exposing a `/metrics` endpoint that needs to be scraped every 5 seconds or if using Kubernetes and K8s needs to hit a `/health` endpoint.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,4 +1,5 @@
 /_build
+/cover
 /deps
 /doc
 /tmp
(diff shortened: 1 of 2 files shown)


akoutmos:
I extracted out this functionality to my own library which acts as a pass through to Plug.Telemetry when the request path is not in the configured reject list https://github.com/akoutmos/telemetry_filter

PR #3635, opened 2019-12-12
fix broken link

mauricionr:
with `.html` we got a 404 page from github, with `.md` works as expected.

### DIFF
--- a/guides/presence.md
+++ b/guides/presence.md
@@ -39,7 +39,7 @@ children = [
 ]
 ```
 
-Next we will create a channel that Presence can communicate on. For this example we will create a `RoomChannel` ([see the channels guide for more details on this](channels.html)):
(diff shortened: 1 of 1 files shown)


guides/presence.md
```diff
@@ -39,7 +39,7 @@ children = [
 ]
 ```
 
-Next we will create a channel that Presence can communicate on. For this example we will create a `RoomChannel` ([see the channels guide for more details on this](channels.html)):
+Next we will create a channel that Presence can communicate on. For this example we will create a `RoomChannel` ([see the channels guide for more details on this](channels.md)):
```
wojtekmach: unfortunately if we change this to `.md`, when this page is rendered here: https://hexdocs.pm/phoenix/presence.html it will 404 then so I'm afraid we can't fix this here.

PR #3637, opened 2019-12-13
Moving paragraph in ecto documentation

bhicks:
Moving the paragraph in the ecto documentation to describe the code block immediately above it. The current placement is below a different code block, which initially confused me. I feel that the new placement reads better.


chrismccord:
❤️❤️❤️🐥🔥

PR #3643, opened 2019-12-22
Add some missing commands to heroku guide

glenwinters:
This updates the Heroku guide to add a few missing `git add` commands.

### DIFF
--- a/guides/deployment/heroku.md
+++ b/guides/deployment/heroku.md
@@ -218,6 +218,8 @@ Let's commit all our changes:
 
 ```console
 $ git add config/prod.exs
+$ git add config/prod.secret.exs
+$ git add Procfile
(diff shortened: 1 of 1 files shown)


Gazler:
I don't think we want the `prod.secret.exs` file to be committed to the repo? Should we instead recommend moving the configuration into `prod.exs` and using the environment variables there?

glenwinters:
@Gazler That sounds reasonable to me.

chrismccord:
ping @glenwinters can you make the requested changes? Thanks!

josevalim:
On Phoenix master we actually recommend including prod.secret.exs because they should all be in environment variables. There is no gitignore for it either. So while I think this change would be ok, it is also not necessary, as it is included. Thanks!

PR #3644, opened 2020-01-02
Add async params support

### LINKED ISSUE #3515: phoenix.js socket.connect params doesn't accept async functions (Promises)
### Environment

* Phoenix version (mix deps): 1.4.9
* NodeJS version (node -v): 11.3.0
* NPM version (npm -v): 6.4.1
* Operating system: OS X Majave 10.14.6

### Expected behavior

The following code should work with an object, a function or an async function (Promise). No params are passed on the connection when an async function (Promise) is passed.

The following are OK.
```
new Socket('url', params: { ... })
```
```
new Socket('url', params: () => { ... })
```
This one does *NOT* work.
```
new Socket('url', params: aysnc () => { ... })
```

### Actual behavior

'params' is wrapped in a closure. The closure function doesn't await functions that are async / promises.

xfumihiro:
fixes #3515

2 functions are touched in order to support async params:

`endPointURL() -> async endPointUrl()`
not sure if this will break any libraries using this client
in my case I'm using [@absinthe/socket](absinthe-graphql/absinthe-socket) that seems not calling endPointURL directly
(ps. documentation should point out which methods are public so we don't accidentally break things)

`connect(params) -> connect(params, callback)`
the additional callback is needed for testing
should be fine if neglected since the conn hooks are async anyway

test added:
```js
it("supports async params", async () => {
      const params = async () => ({ one: "two" })
      socket = new Socket("/socket", {params})
      const url = await socket.endPointURL()
      assert.equal(url, "ws://example.com/socket/websocket?one=two&vsn=2.0.0")
    })
```

cc/ @brweber2

### DIFF
--- a/assets/.babelrc
+++ b/assets/.babelrc
@@ -1,5 +1,4 @@
 {
-    "presets": [
-        "@babel/preset-env"
-    ]
+  "presets": ["@babel/preset-env"],
+  "plugins": ["@babel/transform-runtime"]
 }
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -795,9 +795,10 @@ export class Socket {
    *
    * @returns {string}
    */
-  endPointURL(){
+  async endPointURL() {
+    const params = await this.params()
     let uri = Ajax.appendParams(
-      Ajax.appendParams(this.endPoint, this.params()), {vsn: this.vsn})
+      Ajax.appendParams(this.endPoint, params), {vsn: this.vsn})
     if(uri.charAt(0) !== "/"){ return uri }
     if(uri.charAt(1) === "/"){ return `${this.protocol()}:${uri}` }
 
@@ -826,20 +827,26 @@ export class Socket {
    * Passing params to connect is deprecated; pass them in the Socket constructor instead:
    * `new Socket("/socket", {params: {user_id: userToken}})`.
    */
-  connect(params){
+  connect(params, callback){
     if(params){
       console && console.log("passing params to connect is deprecated. Instead pass :params to the Socket constructor")
       this.params = closure(params)
     }
-    if(this.conn){ return }
+    if(this.conn){ 
+      callback && callback()
+      return 
+    }
     this.closeWasClean = false
-    this.conn = new this.transport(this.endPointURL())
-    this.conn.binaryType = this.binaryType
-    this.conn.timeout    = this.longpollerTimeout
-    this.conn.onopen     = () => this.onConnOpen()
-    this.conn.onerror    = error => this.onConnError(error)
-    this.conn.onmessage  = event => this.onConnMessage(event)
-    this.conn.onclose    = event => this.onConnClose(event)
+    this.endPointURL().then(url => {
+      this.conn = new this.transport(url)
+      this.conn.binaryType = this.binaryType
+      this.conn.timeout    = this.longpollerTimeout
+      this.conn.onopen     = () => this.onConnOpen()
+      this.conn.onerror    = error => this.onConnError(error)
+      this.conn.onmessage  = event => this.onConnMessage(event)
+      this.conn.onclose    = event => this.onConnClose(event)
+      callback && callback()
+    }).catch(error => console.log(error))
   }
 
   /**
--- a/assets/package-lock.json
+++ b/assets/package-lock.json
@@ -10,7 +10,7 @@
       "integrity": "sha512-OfC2uemaknXr87bdLUkWog7nYuliM9Ij5HUcajsVcMCpQrcLmtxRbVFTIqmcSkSeYRBFBRxs2FiUqFJDLdiebA==",
       "dev": true,
       "requires": {
-        "@babel/highlight": "7.0.0"
+        "@babel/highlight": "^7.0.0"
       }
     },
     "@babel/core": {
@@ -19,20 +19,20 @@
       "integrity": "sha512-IFeSSnjXdhDaoysIlev//UzHZbdEmm7D0EIH2qtse9xK7mXEZQpYjs2P00XlP1qYsYvid79p+Zgg6tz1mp6iVw==",
       "dev": true,
       "requires": {
-        "@babel/code-frame": "7.0.0",
-        "@babel/generator": "7.1.2",
-        "@babel/helpers": "7.1.2",
-        "@babel/parser": "7.1.2",
-        "@babel/template": "7.1.2",
-        "@babel/traverse": "7.1.0",
-        "@babel/types": "7.1.2",
-        "convert-source-map": "1.6.0",
-        "debug": "3.2.5",
-        "json5": "0.5.1",
-        "lodash": "4.17.11",
-        "resolve": "1.8.1",
-        "semver": "5.5.1",
-        "source-map": "0.5.7"
+        "@babel/code-frame": "^7.0.0",
+        "@babel/generator": "^7.1.2",
+        "@babel/helpers": "^7.1.2",
+        "@babel/parser": "^7.1.2",
+        "@babel/template": "^7.1.2",
+        "@babel/traverse": "^7.1.0",
+        "@babel/types": "^7.1.2",
+        "convert-source-map": "^1.1.0",
+        "debug": "^3.1.0",
+        "json5": "^0.5.0",
+        "lodash": "^4.17.10",
+        "resolve": "^1.3.2",
+        "semver": "^5.4.1",
+        "source-map": "^0.5.0"
       },
       "dependencies": {
         "debug": {
@@ -41,7 +41,7 @@
           "integrity": "sha512-D61LaDQPQkxJ5AUM2mbSJRbPkNs/TmdmOeLAi1hgDkpDfIfetSrjmWhccwtuResSwMbACjx/xXQofvM9CE/aeg==",
           "dev": true,
           "requires": {
-            "ms": "2.1.1"
+            "ms": "^2.1.1"
           }
         },
         "ms": {
@@ -58,11 +58,11 @@
       "integrity": "sha512-70A9HWLS/1RHk3Ck8tNHKxOoKQuSKocYgwDN85Pyl/RBduss6AKxUR7RIZ/lzduQMSYfWEM4DDBu6A+XGbkFig==",
       "dev": true,
       "requires": {
-        "@babel/types": "7.1.2",
-        "jsesc": "2.5.1",
-        "lodash": "4.17.11",
-        "source-map": "0.5.7",
-        "trim-right": "1.0.1"
+        "@babel/types": "^7.1.2",
+        "jsesc": "^2.5.1",
+        "lodash": "^4.17.10",
+        "source-map": "^0.5.0",
+        "trim-right": "^1.0.1"
       },
       "dependencies": {
         "jsesc": {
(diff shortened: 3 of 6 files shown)


assets/.babelrc
```diff
@@ -1,5 +1,4 @@
 {
-    "presets": [
-        "@babel/preset-env"
-    ]
+  "presets": ["@babel/preset-env"],
+  "plugins": ["@babel/transform-runtime"]
```
snewcomer: happen to know if/by how much this increases the `js` build output?

assets/js/phoenix.js
```diff
@@ -826,20 +827,26 @@ export class Socket {
...
+    }
     this.closeWasClean = false
-    this.conn = new this.transport(this.endPointURL())
-    this.conn.binaryType = this.binaryType
-    this.conn.timeout    = this.longpollerTimeout
-    this.conn.onopen     = () => this.onConnOpen()
-    this.conn.onerror    = error => this.onConnError(error)
-    this.conn.onmessage  = event => this.onConnMessage(event)
-    this.conn.onclose    = event => this.onConnClose(event)
+    this.endPointURL().then(url => {
```
snewcomer: Since a few ppl have this need, I think we should figure something out here.  Does anybody have an example app we can use to hone in on this solution/alternative solutions?

assets/.babelrc
xfumihiro: 22k -> 29k

assets/js/phoenix.js
xfumihiro: I've published packages to test the patch

```json
    "@xfumihiro/absinthe-socket": "0.2.1-rc1",
    "@xfumihiro/absinthe-socket-apollo-link": "0.2.1-rc2",
    "@xfumihiro/phoenix": "1.4.0-rc1",
```
```js
createAbsintheSocketLink(
      AbsintheSocket.create(
        new PhoenixSocket(WS_URL, {
          reconnect: true,
          params: async () => {
            const accessToken = await getAccessToken()
            const refreshToken = await getRefreshToken()
            const token = accessToken || refreshToken
            return {
                jwt: token,
            }
          },
    }),
  ),
)
```

assets/.babelrc
xfumihiro: hmm..
seems like it should be direct dep after all
https://github.com/atlassian/react-beautiful-dnd/issues/845#issuecomment-426963294

xfumihiro: Screen Shot 2020-01-03 at 5 52 18 PM

phoenix.js 5.62k (gzipped)
babel-runtime 2.83k (gzipped)

assets/js/phoenix.js
snewcomer: So my first instinct is what is preventing a user from awaiting these tokens before stamping out a new `PhoenixSocket` instance?  That is the values are resolved before assigning to `params`... 

Note, I'm not implying this PR is unnecessary or anything.  Just want to understand, from your point of view, what Phoenix should handle and what we should depend on users to implement.

xfumihiro: as discussed in #3515 we need this patch in order for the connect method in PhoenixSocket to get the latest params (ex. tokens) asynchronously. again this isn't necessary for web where we can access **localStorage/cookies** synchronously but is critical for mobile where **AsyncStorage** can only be access asynchronously

snewcomer: Yep the async access makes sense and is solved with async/await.  However, my curiosity is in making createAbsintheSocketLink async and awaiting the access token before calling `new PhoenixSocket`.  It sounds like you are saying it is a timing issue?  Ie asking for the token in createAbsintheSocketLink vs the connect method?

xfumihiro: not sure what you mean but the createAbsintheSocketLink is untouched. we only need the params (in PhoenixSocket constructor) to accept async functions.

chrismccord:
To follow up on the original issue, I still don't see the the need for this. Can't you call your params function first, then connect to the socket with the resolved values?

brweber2:
No, think of refresh tokens in OAuth 2. When the socket is first connected I have one token. But as time goes by the token expires and needs to be replaced with a new token. Therefore a callback function is used for params. The problem for us arises when we used an API (from Amazon, not an obscure one) that has an async function to get a new access token using the refresh token.  phoenix supports callback functions, but they can’t be async right now and thus this issue was opened.

chrismccord:
I suggest you take your current approach, which is to keep a store which wraps the param values. This is not the same thing as global hack. It can async fetch/update the params as needed, for example since it likely knows their TTL if we're talking JWT. Then your params function simply accesses this object, which is in scope, ie:

```javascript
import ParamStore from "./param_store"

let params = new ParamStore(...)
let socket = new Socket("...", params: () => params.get())
params.initRefresh(() => socket.connect() )
params.onRefresh(() => console.log("params refreshed", params.get()))
```
  
Hope this helps! I don't think an impl detail like an async promise should leak inside phoenix, and I don't like the idea of bumping the build size to support when you can unwrap like the above. If this is unworkable for you, let me know. Thanks!

xfumihiro:
@chrismccord I agree that bumping the build size sucks but for modern sites that are already using babel-runtime there should be no increase at all. also what you've suggested is still a hack which adds burden on maintaining and is error prone.

snewcomer:
I really like Chris's idea!  This might be a good utility to document.  

Also there is a very minimal path to support async params and that is always wrap params in `Promise.resolve`.  This creates a Promise that resolves to the value if `params` is a value and creates a Promise if `params` is a Promise.

```js
Promise.resolve(this.liveSocket.params(this.view)).then(params => {
    this.channel = this.liveSocket.channel(`lv:${this.id}`, () => {
      return {
        url: this.href || this.liveSocket.main.href,
        params: params,
        session: this.getSession(),
        static: this.getStatic()
      }
    })
})
```

Let me know what ppl think.  As a complete solution, we would have to handle failure + retry/backoff logic.  However, I think this might be an acceptable add.

lud:
Hi @chrismccord  and @snewcomer 

I have this situation where a socket (at the socket level and not the channel) is autheticated via an external source. I pass a token to the params. This token can expire, so if the socket is disconnected after a while, I need to get a fresh token when the socket tries to reconnect and calls `this.endPointURL`, which in its turn calls the params.

The same scheme can be applied to channels if you use tokens that expire, and if you need to fetch a fresh one before connecting.

But I don't see a need to add a huge bloat of libraries to support async:

Something as basic as this could provide async params to all platforms that already support phoenix.js features:

```js
// Unchanged. It returns a function with 0-length if the value was
// not a funciton.
let closure = (value) => {
  if(typeof value === "function"){
    return value
  } else {
    let closure = function(){ return value }
    return closure
  }
}

  endPointURL(functionThatAcceptsURL){
    const buildUrl = (maybeAsyncParams) => {
      let uri = Ajax.appendParams(
        Ajax.appendParams(this.endPoint, maybeAsyncParams, {vsn: this.vsn})
      if(uri.charAt(0) !== "/"){ return uri }
      if(uri.charAt(1) === "/"){ return `${this.protocol()}:${uri}` }
      functionThatAcceptsURL(`${this.protocol()}://${location.host}${uri}`)
    }

    if (this.params.length) {
      // If this.params accepts an argument, is it the continuation
      // callback. this.params must call buildUrl with the params
      this.params(buildUrl)
    } else {
      // If the function does not accept a callback, then it must
      // simply return the params, which we pass to buildUrl
      buildUrl(this.params())
    }
    // buildUrl will in its turn call functionThatAcceptsURL with the
    // URL.
  }

  connect(params){
    if(params){
      console && console.log("passing params to connect is deprecated. Instead pass :params to the Socket constructor")
      this.params = closure(params)
    }
    if(this.conn){ return }

    const handleURL = (asyncEndpointURL) => {
      if(this.conn){ return }
      this.closeWasClean = false
      this.conn = new this.transport(asyncEndpointURL)
      this.conn.binaryType = this.binaryType
      this.conn.timeout    = this.longpollerTimeout
      this.conn.onopen     = () => this.onConnOpen()
      this.conn.onerror    = error => this.onConnError(error)
      this.conn.onmessage  = event => this.onConnMessage(event)
      this.conn.onclose    = event => this.onConnClose(event)
    }

    this.endPointURL(handleURL)
  }
```

chrismccord:
As mentioned above, you can track this value is a variable store which your params function returns so we have no plans to change things at this time. Thanks!

lud:
Well as you said:

> I don't think an impl detail like an async promise should leak inside phoenix, and I don't like the idea of bumping the build size to support when you can unwrap like the above. If this is unworkable for you, let me know. Thanks!

I was answering that with basic JS callbacks there would be neither leak inside pheonix nor build size bump, unlike @xfumihiro solution. Now if you tell me that you just do not want to support it, it is fine.

I am not sure if this param store is a reference to a known library. But if it is just a simple shared object I feel it requires to be able to fetch the a token before declaring the socket variable, asynchronously then, and to be able to parse said token, or to poll regularly for new token. I think it is limitating, and very different from the current phoenix.js API which has a philosophy of declare everything from the get-go and let things happen.

That beeing said I understand your point, we have our own async wrapper (quite ugly though) to provide async params.

And thanks for the great work by the way :)

mupkoo:
Hello 👋 ,

There are more than a few services that provide authentication as a service and work with tokens that can expire, and most of the time they work with promises so you need to resolve it before you can pass it on.

I feel like the added value of this will out-weight the minor added complexity.

Here is a super simple example.

```javascript
let promiseClosure = (value) => {
  return () => {
    if (typeof value === 'function') {
      value = value();
    }
  
    return Promise.resolve(value);
  }
};

// Here how this is going to affect the code for connecting
endPointURL() {
  return this.params().then((params) => {
    let uri = Ajax.appendParams(
      Ajax.appendParams(this.endPoint, params), 
      {vsn: this.vsn}
    );

    if (uri.charAt(0) !== "/") { return uri }
    if (uri.charAt(1) === "/") { return `${this.protocol()}:${uri}` }

    return `${this.protocol()}://${location.host}${uri}`
  });
}

connect(params) {
  if (params) {
    console && console.log("passing params to connect is deprecated. Instead pass :params to the Socket constructor")
    this.params = promiseClosure(params)
  }

  if(this.conn){ return }

  this.endPointURL().then((url) => {
    this.closeWasClean = false
    this.conn = new this.transport(params)
    this.conn.binaryType = this.binaryType
    this.conn.timeout    = this.longpollerTimeout
    this.conn.onopen     = () => this.onConnOpen()
    this.conn.onerror    = error => this.onConnError(error)
    this.conn.onmessage  = event => this.onConnMessage(event)
    this.conn.onclose    = event => this.onConnClose(event)
  });
}
```

And since `promiseClosure` is working with simple values, a callback or a callback that returns a promise, this means that this is not going to affect how is this being used at all of the below will be valid

```javascript
new Socket('/socket', { params: { token: 'secret' } });
new Socket('/socket', { params: () => ({ token: 'secret' }) });
new Socket('/socket', { params: async () => ({ token: 'secret' }) });
```

chrismccord:
@mupkoo please see my comment here:
https://github.com/phoenixframework/phoenix/issues/3515#issuecomment-628192821

The full solution is likely taking control of the entire connection life-cycle process, rather than using the params callback to handle token refreshing. Thanks!

lud:
Hey @chrismccord , is it a good practice to create a new `socket.channel( … )` with the same topic when the previous channel closed, or was excplicitly left ?

chrismccord:
They client will send a new unique reference for the opened channel so no worries there

PR #3649, opened 2020-01-11
Mention PORT env var in Heroku deployment docs

maltoe:
Hi,

This patch changes `config.prod.exs` in the Heroku deployment guide, to make the endpoint bind to the
port requested by Heroku to prevent the R10 error from happening. As far as I can tell, the error should be happening to everyone currently following the guide strictly (?).

https://devcenter.heroku.com/articles/error-codes#r10-boot-timeout


maltoe:
Test failure seems unrelated?

Gazler:
Thanks <3

PR #3650, opened 2020-01-12
Ignore directory created by elixir language server

dkuku:
It's not strictly phoenix issue but this directory is annoying for me. And I imagine that I'm not the only person using this extension


Gazler:
Thanks for the suggestion. I think something like this should be handled by a global .gitignore.

https://help.github.com/en/github/using-git/ignoring-files#create-a-global-gitignore

PR #3653, opened 2020-01-17
Add elixir install line

inoas:
Either the whole passage should be removed because...
- it is distro specific
- it is explained fully already on the linked elixir install page

... or it should contain the full example like on the elixir install page


Gazler:
I am in agreement with removing the entire section from:

 > People using Debian-based systems may need to explicitly install Erlang to get all the needed packages.

onward.

inoas:
Replaced by https://github.com/phoenixframework/phoenix/pull/3653

PR #3655, opened 2020-01-18
Update package.json to get latest js dependencies

### LINKED ISSUE #3654: Time to upgrade javascript deps for installer?
It has been a while since the last update of dependencies in the `package.json` template for `phx_new`. I just did a new install of `1.4.11` and `npm audit` reports 2 moderate severity vulnerabilities present because the newest versions cannot be installed due to the `package.json` version constraints.

The following packages have `package.json` constraints that prevent updates to their newest versions.

- `copy-webpack-plugin` can be updated from 4.5.0 -> 5.1.1 [changelog](https://github.com/webpack-contrib/copy-webpack-plugin/blob/master/CHANGELOG.md)
- `css-loader` can be updated from 2.1.1 -> 3.4.2 [changelog](https://github.com/webpack-contrib/css-loader/blob/master/CHANGELOG.md)
- `mini-css-extract-plugin` can be updated from 0.4.0 -> 0.9.0 [changelog](https://github.com/webpack-contrib/mini-css-extract-plugin/blob/master/CHANGELOG.md)
- `terser-webpack-plugin` can be updated from 1.1.0 -> 2.3.2 [changelog](https://github.com/webpack-contrib/terser-webpack-plugin/blob/master/CHANGELOG.md)

Additionally `webpack` can be  upgraded from `4.4.0` to `4.41.5`. The changelog is on their [release page](https://github.com/webpack/webpack/releases)

Among other breaking changes in the modules above is a dependency on at least Node.js 8.9.0 (which is actually [no longer supported](https://nodejs.org/en/about/releases/), so I don't think that should be an issue). 

We should of course verify that the breaking changes in the new package versions are not an issue.

skovmand:
This is a PR to update version constraints in `package.json`. Fixes #3654 

As mentioned in #3654 there are breaking changes in the dependencies. We should verify that they are not an issue.

I have tested that `mix phx.server` runs as normal, and that the webpack compilation runs without errors. I haven't done a detailed testing of the compilation of individual files, channels or liveview etc.

### DIFF
--- a/installer/templates/phx_assets/package.json
+++ b/installer/templates/phx_assets/package.json
@@ -13,12 +13,12 @@
     "@babel/core": "^7.0.0",
     "@babel/preset-env": "^7.0.0",
     "babel-loader": "^8.0.0",
-    "copy-webpack-plugin": "^4.5.0",
-    "css-loader": "^2.1.1",
-    "mini-css-extract-plugin": "^0.4.0",
+    "copy-webpack-plugin": "^5.1.1",
+    "css-loader": "^3.4.2",
+    "mini-css-extract-plugin": "^0.9.0",
     "optimize-css-assets-webpack-plugin": "^5.0.1",
-    "terser-webpack-plugin": "^1.1.0",
-    "webpack": "4.4.0",
+    "terser-webpack-plugin": "^2.3.2",
+    "webpack": "4.41.5",
     "webpack-cli": "^3.3.2"
   }
 }


skovmand:
Note: In this PR I have set `webpack` to a fixed version because it was fixed too in the previous template (it was set to `4.4.0`). But is there a reason for it being a fixed version with no opportunity for minor or patch updates? I think it would make sense to at least get patch versions, perhaps also minor releases.

This is current version constraint in this PR:

```
"webpack": "4.41.5"
```

installer/templates/phx_assets/package.json
```diff
@@ -13,12 +13,12 @@
...
-    "mini-css-extract-plugin": "^0.4.0",
+    "copy-webpack-plugin": "^5.1.1",
+    "css-loader": "^3.4.2",
+    "mini-css-extract-plugin": "^0.9.0",
     "optimize-css-assets-webpack-plugin": "^5.0.1",
-    "terser-webpack-plugin": "^1.1.0",
-    "webpack": "4.4.0",
+    "terser-webpack-plugin": "^2.3.2",
+    "webpack": "4.41.5",
     "webpack-cli": "^3.3.2"
```
snewcomer: I think we would want a lock file update before merging as well!

skovmand: I think one of the good things about not having a lock file is that the packages are installed only based on the version constraints in `package.json`, and that would allow for the latest versions if this PR is merged. If we introduced a lock file, wouldn't that specify the exact versions to be installed? And in turn that we would have to update that lock-file far more often?

I'm not sure, I'm still a bit confused about how npm actually uses the lockfile (I usually use `npm ci` to get the exact lockfile dependencies, but I'm unsure how `npm install` handles a lockfile).

skovmand: After doing a bit of reading, I think that adding a `package-lock.json` file and running `npm install` would be the same as pinning all dependencies (and their dependencies) to specific versions. As far as I can see, the `package.json` file [will only be used as reference if it is newer](https://medium.com/coinmonks/everything-you-wanted-to-know-about-package-lock-json-b81911aa8ab8) than `package-lock.json`. There would be both advantages and disadvantages to this, I think.

Just for some stats: By running `npm ls | wc -l` inside a freshly installed Phoenix installation gives me 1377 different dependencies. (Removing deduped entried with `npm ls | grep -v deduped | wc -l` gives me 764). :open_mouth: 

An advantage to using a lockfile would be that we are very sure that a release of `phx_new` has good dependencies at the time of its release.

A disadvantage would be that this lockfile would very quickly specify old versions of packages, and also pretty quickly reference packages with known vulnerabilities. If a user runs `npm install` with an outdated lock-file, the user would get these exact dependencies. So we would have to make a new release of `phx_new` every time any dependency in the dependency tree needs to be updated. 

If there was no lockfile however, `npm install` would resolve the newest packages within the package version constraints, so it would fetch the newest versions possible. Then we would only have to make `phx_new` releases when the constraints no longer allow new packages (which is the current situation).

skovmand: I just found two further references:  [npm-install](https://docs.npmjs.com/cli-commands/install.html) [npm-package-lock.json](https://docs.npmjs.com/configuring-npm/package-lock-json)

chrismccord:
we don't generate a package-lock.json inside new projects so this is good to go. Thanks!
❤️❤️❤️🐥🔥

HusseinMorsy:
@skovmand are there any reasons, why `@babel/core` and `@babel/preset-env` are not updated to the newest version (from `7.0.0` to `7.8.3`) ?

skovmand:
@HusseinMorsy The version constraint `"^7.0.0"` allows `7.8.3` to be installed, so running `npm install` will install `7.8.3` in your example.

HusseinMorsy:
@skovmand thank you very much the helpful explanation. Now I got it.

Issue #2875, opened 2018-05-07
Add support for a cache aware H2 PUSH_PROMISE with CASPer


idyll:
In order to complete H2 support in Phoenix we need to support a mechanism for the server to push or hint resources. There are two mechanisms for this now: Early Hints and Push Promises.

Initially I was hoping to add support for Early Hints but during the process it became apparent that it may be months before *any* of the major web browser support early hints. (This is based on a conversation on twitter with Ilya.)

That means at least initially we need to support Push Promise as well.

I've looked at how Rails has added support for Push Promise (I know they say that they are doing early hints, but that's not 100% accurate. They are using h2o to convert an early hint from puma into a push promise.) We can do the same thing for anyone that wants to proxy through h2o like rails does. H2O is using a technique called CASPer when they store a representation of a list of assets on the client in a cookie.

Their implementation is here:
https://github.com/h2o/h2o/blob/master/lib/http2/casper.c

As discussion of it is here:
https://h2o.examp1e.net/configure/http2_directives.html#http2-casper

TL;DR they are using golomb compressed sets (https://www.imperialviolet.org/2011/04/29/filters.html) aka a bloom filter.

So to support Early Hints and Push Promises we need:

1. a dependency for a Bloom filter. (maybe https://github.com/gmcabrita/bloomex)
2. functions to read and write the cookie.
3. functions to get a list of the assets that should be pushed for a page.
4. a function that takes a list of assets and the cookie and determines which assets to push.
5. a function that connects this all into a push promise.
6. a function that uses the list from 3 to do early hints.

I think we should support them both. This gives the end use maximum flexibility and there's very little extra effort on top of Push Promise needed to support Early Hints.

I am hoping to gather some feedback on this and then I will start working on a PR. 

Plug already has support for Push Promise (but there's going to be a Cowboy fix before it fully works) and there's an open PR for Early Hints support.

Gazler:
@idyll Thanks for opening this proposal.

The solution by h2o is neat. There's another proposal for caching with server push https://datatracker.ietf.org/doc/draft-ietf-httpbis-cache-digest/?include_text=1 which is somewhat related this.

From an API point of view, if we already have `server_push(conn, "foo.css")` then we'd maybe want something like:

```elixir
# Cache is default
server_push(conn, "foo.css")

# Implementation is transparent
server_push(conn, "foo.css", cache: false)

# Instead of
server_push(conn, "foo.css", casper: false)
```

[...]

PR #3656, opened 2020-01-21
Add `defer` to default blueprint script

snewcomer:
`defer` is an easy, well supported attribute we can provide in our blueprints to start loading scripts much earlier.  `defer` basically says, go to network but don't execute until `domInteractive` (once HTML is parsed and loaded).  Currently with the `js` file at the end of the body tag, it goes to network and executes only until after the HTML is parsed and loaded.

97% browser support: 

https://caniuse.com/#feat=script-defer

## Before (last tick at bottom of performance timeline)
Screen Shot 2020-01-20 at 8 02 00 PM

## After (second tick at same time as css is fetched)
Screen Shot 2020-01-20 at 8 00 05 PM

### DIFF
--- a/installer/templates/phx_web/templates/layout/app.html.eex
+++ b/installer/templates/phx_web/templates/layout/app.html.eex
@@ -7,6 +7,7 @@
     <title><%= app_module %> · Phoenix Framework</title>
     <link rel="stylesheet" href="<%%= Routes.static_path(@conn, "/css/app.css") %>"/>
     <%%= csrf_meta_tag() %>
+    <script defer type="text/javascript" src="<%%= Routes.static_path(@conn, "/js/app.js") %>"></script>
   </head>
   <body>
     <header>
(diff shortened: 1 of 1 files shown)


chrismccord:
@Gazler I think we had this at some point, but reverted. I can't think of a reason to avoid at this point, but do you remember any issues?

Gazler:
It was reverted in https://github.com/phoenixframework/phoenix/commit/5bca542e432563c02a55b77c0cf69eed0ee793b9

I Can't remember why we didn't keep the defer.

snewcomer:
I'll take a risk and assume we won't have issues.  However, if ppl are desperate for performance improvements, this is one change they will likely.  So this PR isn't strictly necessary.  Perhaps we can file this a "best practices" for phoenix applications or do we want it as the default?

achempion:
I've checked the history of this change and it seems was reverted because of usage of `static_push` function instead of `static_path`. As I understand `defer` was just a minor optimisation accompanying the PR.

The whole discussion about `static_push` implementation is [here](https://github.com/phoenixframework/phoenix/issues/2875) and [here](https://github.com/phoenixframework/phoenix/pull/2805) was the introduction of such feature.

Here is somewhat related mention on [elixir forum](https://elixirforum.com/t/phoenix-1-4-0-released/17182/114).

I think we should accept `defer` to the framework as it was removed just by the accident.

chrismccord:
❤️❤️❤️🐥🔥

PR #3660, opened 2020-01-24
Fix file path formatting

kimprado:
Adjustment for file path code highlighting in Contexts Guide.

We have:
... `lib/hello_web/views/session_view.ex:`
... `lib/hello_web/templates/session/new.html.eex:`

We wish
... `lib/hello_web/views/session_view.ex`:
... `lib/hello_web/templates/session/new.html.eex`:

### DIFF
--- a/guides/contexts.md
+++ b/guides/contexts.md
@@ -533,7 +533,7 @@ defmodule HelloWeb.SessionView do
 end
 ```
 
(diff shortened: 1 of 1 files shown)


chrismccord:
❤️❤️❤️🐥🔥

PR #3661, opened 2020-01-28
Use non-deprecated socket/3 in guide

maartenvanvliet:
socket/2 is deprecated, see: https://github.com/phoenixframework/phoenix/blob/v1.4.12/lib/phoenix/test/channel_test.ex#L240


Gazler:
Thanks <3

PR #3663, opened 2020-01-30
Remove Debian distro specific help

inoas:
... as it is redundant info found on the elixir install page.
Replaces https://github.com/phoenixframework/phoenix/pull/3653


Gazler:
Thanks <3

PR #3664, opened 2020-02-01
Add option to not define default layout and view

hauleth:
This can be useful for applications that want to control the defaults on their own, for example I use different naming scheme where instead of `MyAppWeb.FooController` each controllers is named as `MyAppWeb.Controllers.Foo` and the views are named `MyApp.Views.Foo`, which is suits more into overall naming scheme after file path.

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -162,9 +162,7 @@ defmodule Phoenix.Controller do
   connection-agnostic and typically invoked from your views.
   """
   defmacro __using__(opts) do
-    {put_default_views, opts} = Keyword.pop(opts, :put_default_views, true)
-
-    quote bind_quoted: [put_default_views: put_default_views, opts: opts] do
+    quote bind_quoted: [opts: opts] do
       import Phoenix.Controller
 
       # TODO v2: No longer automatically import dependencies
@@ -172,7 +170,7 @@ defmodule Phoenix.Controller do
 
       use Phoenix.Controller.Pipeline, opts
 
-      if put_default_views do
+      if opts[:put_default_views] do
         plug :put_new_layout, {Phoenix.Controller.__layout__(__MODULE__, opts), :app}
         plug :put_new_view, Phoenix.Controller.__view__(__MODULE__)
       end
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -41,6 +41,9 @@ defmodule Phoenix.Controller do
     * `:log` - the level to log. When false, disables controller
       logging
 
+    * `:put_default_views` - controls whether the default view
+      and layout should be set or not
+
   ## Connection
 
   A controller by default provides many convenience functions for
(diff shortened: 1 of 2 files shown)


josevalim:
Sorry but I don't understand why we need this change. If you don't use the defaults, you have to set them yourself, so can't you just set them? Phoenix is not going to get in the way... this change does not seem to be necessary.

hauleth:
@josevalim the problem is that I do not have general solution to set my own default without overriding anything set earlier in the upstream. So in other words I cannot do:

```elixir
defmodule MyAppWeb.Controllers.Foo do
  use Phoenix.Controller

  plug :put_new_layout, {MyAppWeb.View.Layout, :app}
  plug :put_new_view, MyAppWeb.View.Foo
  # …
end
```

This is just example, because in my application I am using my own template, so it is more like:

```elixir
defmodule MyAppWeb.Controller do
  defmacro __using__(opts \\ []) do
    layout_mod = layout_mod(__CALLER__.module)
    view_mod = view_mod(__CALLER__.module)
    quote do
      use Phoenix.Controller

      plug :put_new_layout, {unquote(layout_mod), :app}
      plug :put_new_view, unquote(view_mod)
    end
  end

  # …
end
```

And then in my controller it is just:

```elixir
defmodule MyAppWeb.Controllers.Foo do
  use MyAppWeb.Controller

  # …
end
```

Current workaround is to use non-public `Phoenix.Controller.Pipeline` instead of `Phoenix.Controller`, but this has obvious the problem of using non-public API:

```elixir
defmodule MyAppWeb.Controller do
  defmacro __using__(opts) do
    view = view_module(__CALLER__.module)

    quote bind_quoted: [view: view, opts: opts] do
      use Phoenix.Controller.Pipeline, opts

      import Phoenix.Controller
      import Plug.Conn

      plug :put_new_layout, {MyAppWeb.Views.Layout, :app}
      plug :put_new_view, view
    end
  end

  defp view_module(module) do
    module
    |> Module.split()
    |> replace()
  end

  defp replace(parts, ret \\ [])

  defp replace([], ret) do
    raise(
      ArgumentError,
      "Expected controller, got: #{inspect(Module.concat(Enum.reverse(ret)))}"
    )
  end

  defp replace(["Controllers" | rest], ret),
    do: Module.concat(Enum.reverse(ret, [Views | rest]))

  defp replace([hd | tl], ret), do: replace(tl, [hd | ret])
end
```

This is (almost) exact copy from the application that I am working on, so you see that I use a little bit different naming scheme from the "official" Phoenix naming scheme (I just find it much more straightforward and consistent with the rest of modules).

Without that change (or making `Phoenix.Controller.Pipeline` public) there is no "safe" way to define custom default view and layout in the same way the Phoenix does it.

josevalim:
Perfect. Can you please call the option `put_default_views`? Then we can merge it. Thanks!

lib/phoenix/controller.ex
```diff
@@ -159,16 +162,20 @@ defmodule Phoenix.Controller do
   connection-agnostic and typically invoked from your views.
   """
   defmacro __using__(opts) do
-    quote bind_quoted: [opts: opts] do
+    {put_default_views, opts} = Keyword.pop(opts, :put_default_views, true)
```
josevalim: If you do this here, you will force the options to be given at compile time and something like this will break:

    opts = [put_default_views: false]
    use Phoenix.Controller, opts

It is generally a no-no to process options outside of quote. :)

josevalim:

hauleth:
@josevalim a little bit too fast, as now it changes default behaviour to not define the default views (as `opts[:put_default_views]` will return `nil` which is falsey value).

hauleth:
It probably should use `Keyword.get/3` but within quote body instead of in macro body.

josevalim:
Yeah, i noticed before merging. My bad, i pushed a fix already. :)

PR #3665, opened 2020-02-05
Fix rejoining channels after explicit reconnect

### LINKED ISSUE #3458: Explicit reconnect doesn't rejoin channels
* Phoenix version: 1.4.8

### Expected behavior

Explicit reconnect should rejoin the channels.

### Actual behavior

Doing an explicit reconnect on existing open connection using:
```
socket.disconnect()
socket.connect()
```

or
```
socket.reconnectTimer.scheduleTimeout()
```

does not rejoin the channels. Instead we start getting `reason: "unmatched topic"` errors.

We use this to try to switch from LongPoll connection to WebSocket connection.

Basically what we do is (coffeescript):
```
  startAvoidLongPollWorker = (socket) ->
    work = ->
      # Avoid changing transports and interfering with the reconnection logic
      # when we don't have an established connection and the reconnection might
      # already be in progress
      return if socket.connectionState() != 'open'

      # Return if the user is already not using LongPoll
      return if socket.transport != Phoenix.LongPoll

      newTransport = findNewTransport(socket.transport)

      # Return if the new transport is the same
      return if socket.transport == newTransport

      # Change socket transport and schedule a reconnect
      socket.transport = newTransport
      socket.reconnectTimer.scheduleTimeout()

    setInterval(work, AVOID_LONGPOLL_WORKER_INTERVAL_IN_MS)
```

As a workaround we currently did:
```
    socket.onOpen ->
      # Do channel rejoin in next event loop, otherwise it's too soon and
      # doesn't work.
      channels = socket.channels
      setTimeout(->
        channels.forEach((channel) ->
          channel.rejoinTimer.scheduleTimeout()
        )
      )
```

indrekj:
Previously when a socket which had channels open was explicitly reconnected (`socket.disconnect(() => socket.connect())` then the existing channels were not rejoined.

This was because `onConnClose` was not called. This was due to a timing issue in `teardown` function. This function called close on connection and also cleaned up onclose listener on the connection object. The close however happened in the next event loop by which time the onclose listener was already removed.

The channels stayed in `joined` state even though socket was disconnected. On reconnect `joined` channels were not rejoined.

Fixes #3458

### DIFF
--- a/assets/test/socket_test.js
+++ b/assets/test/socket_test.js
@@ -263,19 +263,22 @@ describe("with transports", done =>{
       socket = new Socket("/socket")
     })
 
-    it("removes existing connection", () => {
+    it("removes existing connection", done => {
       socket.connect()
-      socket.disconnect()
-
-      assert.equal(socket.conn, null)
+      socket.disconnect(() => {
+        assert.equal(socket.conn, null)
+        done()
+      })
     })
 
-    it("calls callback", () => {
+    it("calls callback", done => {
       let count = 0
       socket.connect()
-      socket.disconnect(() => count++)
-
-      assert.equal(count, 1)
+      socket.disconnect(() => {
+        count++
+        assert.equal(count, 1)
+        done()
+      })
     })
 
     it("calls connection close callback", () => {
@@ -630,14 +633,13 @@ describe("with transports", done =>{
       assert.ok(spy.calledOnce)
     })
 
-    it('does not schedule reconnectTimer timeout if normal close after explicit disconnect', () => {
+    it('does not schedule reconnectTimer timeout if normal close after explicit disconnect', done => {
       const spy = sinon.spy(socket.reconnectTimer, 'scheduleTimeout')
 
-      const event = { code: 1000 }
-
-      socket.disconnect()
-
-      assert.ok(spy.notCalled)
+      socket.disconnect(() => {
+        assert.ok(spy.notCalled)
+        done()
+      })
     })
 
     it('schedules reconnectTimer timeout if not normal close', () => {
@@ -650,17 +652,19 @@ describe("with transports", done =>{
       assert.ok(spy.calledOnce)
     })
 
-    it('schedules reconnectTimer timeout if connection cannot be made after a previous clean disconnect', () => {
+    it('schedules reconnectTimer timeout if connection cannot be made after a previous clean disconnect', done => {
       const spy = sinon.spy(socket.reconnectTimer, 'scheduleTimeout')
 
-      socket.disconnect();
-      socket.connect();
+      socket.disconnect(() => {
+        socket.connect();
 
-      const event = { code: 1001 }
+        const event = { code: 1001 }
 
-      socket.onConnClose(event)
+        socket.onConnClose(event)
 
-      assert.ok(spy.calledOnce)
+        assert.ok(spy.calledOnce)
+        done()
+      })
     })
 
     it("triggers onClose callback", () => {
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -939,12 +939,20 @@ export class Socket {
   }
 
   teardown(callback, code, reason){
-    if(this.conn){
-      this.conn.onclose = function(){} // noop
+    if(this.conn) {
       if(code){ this.conn.close(code, reason || "") } else { this.conn.close() }
-      this.conn = null
     }
-    callback && callback()
+
+    // Force connection cleanup and callback to the next event loop. This is
+    // necessary for `conn.close()` to finish and trigger all necesary
+    // callbacks.
+    setTimeout(() => {
+      if(this.conn){
+        this.conn.onclose = function(){} // noop
+        this.conn = null
+      }
+      callback && callback()
+    })
   }
 
   onConnClose(event){


snewcomer:
@indrekj 👋 This is great!  Coincidentally I was looking at this the last few days as well.

One question I had queued up was we [setup](https://github.com/phoenixframework/phoenix/blob/ec598c104ca8e910b0dbcb0653cb7dc7ae9f7fc6/assets/js/phoenix.js#L857) `this.conn.onclose` to `onConnClose` but then [nullify](https://github.com/phoenixframework/phoenix/blob/ec598c104ca8e910b0dbcb0653cb7dc7ae9f7fc6/assets/js/phoenix.js#L944) it.   Is it possible that it never is called and the solution would be to ensure `onclose` is called before nullifying it?

indrekj:
I initially thought the same thing and I tried fixing it by swapping https://github.com/phoenixframework/phoenix/blob/ec598c104ca8e910b0dbcb0653cb7dc7ae9f7fc6/assets/js/phoenix.js#L944-L945 these two lines, but that didn't seem to help. Maybe it will work if nullifying would happen in the next event loop, but I did not try that.

achempion:
Does it somehow related to https://github.com/phoenixframework/phoenix/pull/3295?
The PR fixes https://github.com/phoenixframework/phoenix/issues/3171.

assets/test/socket_test.js
```diff
@@ -270,6 +270,29 @@ describe("with transports", done =>{
...
+
+    it("sets channels to error state", () => {
+      socket.connect()
+
+      const channel = socket.channel("topic", {one: "two"})
+      channel.join().trigger("ok", {})
+      assert.equal(channel.state, "joined")
+
+      socket.disconnect()
+      assert.equal(channel.state, "errored")
```
snewcomer: I'm trying to bridge the gap between what might happen in tests and what happens in prod.  Basically, I am assuming `mock-socket` sets the `readyState` to CLOSED and then follows the following steps - [spec](https://html.spec.whatwg.org/multipage/web-sockets.html#feedback-from-the-protocol%3Aconcept-websocket-closed) - which is to fire an `error` event, thus setting the `state === 'errored'` (our code) and firing a `close` event...all happening in a [task](https://html.spec.whatwg.org/multipage/webappapis.html#queue-a-task).  

Example tests asserting the `errored` state when calling `socket.disconnect()` w/ `setTimeout`:
https://github.com/snewcomer/phoenix/commit/fef70317cc41f445d92371475543b126b896e914

So, overall `onConnClose` is called from what I can see.  Just in the next macroTask.

indrekj:
What also works is to change teardown function to something like this:
```
  teardown(callback, code, reason){
    if(this.conn){
      if(code){ this.conn.close(code, reason || "") } else { this.conn.close() }
    }
    setTimeout(function() {
      if (this.conn) {
        this.conn.onclose = function(){} // noop
        this.conn = null
      }
      callback && callback()
    }
  }

```
Basically if conn.close happens immediately and setting onclose to null is in the next event loop then it seems to be working just fine.

assets/test/socket_test.js
snewcomer: Ok just tested a running app an do not see `onConnClose` called.  So in a test context, this test ensures you don't need to `assert` at some point in the future b/c you made the `onConnClose` sync instead of async.

Thinking in my own head, I wonder how we could expand this test given the diff b/c test and prod...

snewcomer:
@indrekj That seems like the fix 👍!  I'm assuming the bug was in the order of operations - `close()` before no-oping and cleaning up.

indrekj:
@snewcomer I updated the PR. I didn't really know how to test this though.

Here's also the demo app for the bug: https://github.com/indrekj/reconnect_bug_demo Just start it up and follow the instructions on the root page.

snewcomer:
@indrekj The demo app is great!  Thank you for that.  I think this is pretty close!

1. It seems (from basic testing) that the event listener `onclose` is not always in the next tick.  My understanding is `onclose` will be called when the `socket.conn.readyState === 3`.  So I think we can write a recursive function to check for this state and then cleanup.

something like the following...

```js
function waitForSocketClosed(conn, callback) {
  if (conn.readyState === SOCKET_STATES.closed) {
    callback();
  } else {
    waitForSocketClosed(socket, callback); // maybe with some sort of incremental backoff?
  }
}

waitForSocketClosed(this.conn, () => {
  if (this.conn) {
    this.conn.onclose = function(){} // noop
    this.conn = null
  }
  callback && callback()
})
```

2. That will still leave our test suites with two rougher paradigms - testing callbacks (as you have correctly done in the test changes) and asynchrony. It seems like the latter is and will cause indeterminate tests since either setTimeout or a recursive function will leak state across multiple tests.  Seems we still need to figure this out.

Lmk your thoughts!

snewcomer:
@indrekj I'm sure you are busy with other things but is this something you would like me to take a shot at?  You have definitely laid the foundation!

indrekj:
Hey, I won't be touching this in the next couple of weeks. Hopefully can get back to it after that. If you want to finish it sooner yourself then I'd appreciate it.

chrismccord:
Discussion moved to #3708 . Thanks!!!

PR #3667, opened 2020-02-06
Replace includes with indexOf in phoenix.js

indrekj:
Sadly we have to support IE11 which does not support [Array.prototype.includes()
](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/includes). The fix itself it quite simple: instead of includes we can use indexOf.


josevalim:
Hi @indrekj! Can you please revert the changes to priv/static and installer? we cannot accept precompiled files in general because that can be a security issue. Thank you!

If you can do that to the other PRs, it would be very appreciated!

indrekj:
Updated

indrekj:
CI is green now

josevalim:

PR #3670, opened 2020-02-15
Remove ActiveXObject for requests for IE5/6

snewcomer:
web standards and our browser support allow us to avoid this code path


assets/js/phoenix.js
```diff
@@ -1196,9 +1196,7 @@ export class Ajax {
       let req = new XDomainRequest() // IE8, IE9
```
snewcomer: We could technically remove this. Current worldwide usage for both combined is at < 1%.

snewcomer:
Getting errors in CI b/c of this...

Screen Shot 2020-03-17 at 9 36 05 PM

PR #3671, opened 2020-02-15
Add built-in mechanism for rendering nested layouts

josevalim:


josevalim:

juliogoldbach:
Hi @josevalim, I see from this PR changes that passing a struct as `assigns` for render functions in lib/phoenix/view.ex is not possible anymore since Map.new just accepts variables with Enumerable protocol. Was this the expected result? I dont see anything related to this in v1.5 changelog.

josevalim:
It was never really expected to receive structs, so this was intentional but I didn’t think structs were actually being given. Probably worth a CHANGELOG entry.

juliogoldbach:
@josevalim I added a PR to update the v1.5 CHANGELOG. Thanks for the fast answer

PR #3676, opened 2020-02-21
Add publicPath to Webpack config

type1fool:
# Problem
Some JS libraries generate chunk files when Webpack is running, and the current config leads to 404 errors when these chunks are referenced in the compiled JS. For example, the [Monaco Editor](https://microsoft.github.io/monaco-editor/) library generates chunks for service workers and languages.

# Solution
By adding setting `publicPath: '/js/'` to the Webpack config's output, these chunks can be successfully imported. With the updated config, the chunked filenames are generated as `[name].js`. The entry key has been changed to `app`, which produces an `app.js` file in the `priv/static/js` directory.

With the Monaco library imported with the new Webpack config, there are now dozens of files in the js directory:

```
# priv/static/js
...
52.js
53.js
54.js
55.js
56.js
57.js
58.js
59.js
6.js
60.js
61.js
62.js
63.js
7.js
8.js
9.js
app.js
css.worker.js
editor.worker.js
html.worker.js
json.worker.js
ts.worker.js
```

### DIFF
--- a/installer/templates/phx_assets/webpack.config.js
+++ b/installer/templates/phx_assets/webpack.config.js
@@ -18,7 +18,8 @@ module.exports = (env, options) => ({
   },
   output: {
     filename: 'app.js',
-    path: path.resolve(__dirname, '../priv/static/js')
+    path: path.resolve(__dirname, '../priv/static/js'),
+    publicPath: "/js/"
   },
   devtool: devMode ? 'source-map' : undefined,
   module: {
--- a/installer/templates/phx_assets/webpack.config.js
+++ b/installer/templates/phx_assets/webpack.config.js
@@ -14,10 +14,10 @@ module.exports = (env, options) => ({
     ]
   },
   entry: {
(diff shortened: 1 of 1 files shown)


chrismccord:
@Gazler I'm not at all familiar with this setting or if it's broadly application for the default project generator. Can you take a look when you have some time? Thanks!

Gazler:
This appears to make no practical difference for a default export, adding 4 characters to the output:

```bash
$ diff app-new.unminified.js app-old.unminified.js 
80c80
< /******/ 	__webpack_require__.p = "/js/";
---
> /******/ 	__webpack_require__.p = "";
```

So I think we should be good to merge it if it helps prevent tripping people up with certain libraries.

type1fool:
thanks, guys! 🤘🏻

chrismccord:
❤️❤️❤️🐥🔥

type1fool:
@chrismccord love fried chicken?

image

chrismccord:
Only that of the long-lived bird that cyclically regenerates to be born again variety :D

PR #3679, opened 2020-02-23
Add root layouts

chrismccord:


lib/phoenix/controller.ex
```diff
@@ -535,6 +535,62 @@ defmodule Phoenix.Controller do
...
+  """
+  @spec put_new_root_layout(Plug.Conn.t, {atom, binary | atom} | false) :: Plug.Conn.t
+  def put_new_root_layout(%Plug.Conn{state: state} = conn, layout)
+      when (is_tuple(layout) and tuple_size(layout) == 2) or layout == false do
+    if state in @unsent do
+      update_in conn.private, &Map.put_new(&1, :phoenix_root_layout, layout)
+    else
+      raise AlreadySentError
+    end
+  end
```
josevalim: I think we don't need this one because we don't have a default operation (like we call put_layout for controllers) that puts the root layout.

josevalim:
Just one minor comment about a function we probably don't need. Otherwise, looks great! :shipit:

PR #3680, opened 2020-02-24
Add phx.new --live 

chrismccord:
This is based on the work @aforward started in #3376 (❤️) . All the setup and tests are in place so it's now a matter of determining what we want to go with naming wise and home page feature wise. I included a Modal component with the app, which is something our live CRUD generators could check for and add, but it would be nice to include it here if we can make it worthwhile. Currently I use the modal to show `hex.outdated` results, but I think we can do better so consider this a starting point to explore from live home page wise.


installer/templates/phx_live/live/home_live.ex
```diff
@@ -0,0 +1,31 @@
...
+  def mount(_params, _session, socket) do
+    {:ok, socket}
+  end
+
+  def handle_params(_params, _uri, socket) do
+    {:noreply, socket}
+  end
+end
+
+defmodule <%= web_namespace %>.HomeLive.DepsCheck do
```
chrismccord: I dumped this component here since it's likely not to survive our discussions. If we want to go this direction I'm fine with an additional file in the `live/` dir for a component, but dumping here for experiments :)

installer/lib/mix/tasks/phx.new.ex
```diff
@@ -40,6 +40,8 @@ defmodule Mix.Tasks.Phx.New do
 
     * `--no-gettext` - do not generate gettext files.
 
+    * `--live` - include Phoenix LiveView.
```
josevalim:
```suggestion
    * `--live` - include Phoenix LiveView to make it easier than ever
      to build interactive, real-time applications
```

Also, please move this option up in the docs.

With love,

The Marketing Department.

installer/lib/phx_new/generator.ex
```diff
@@ -150,12 +151,14 @@ defmodule Phx.New.Generator do
       phoenix_static_path: phoenix_static_path(phoenix_path),
       pubsub_server: pubsub_server,
       secret_key_base: random_string(64),
+      live_view_signing_salt: random_string(64),
```
josevalim: So I had already added this in the past, see `lv_signing_salt` two lines below. :)

installer/templates/phx_live/templates/page/home.html.leex
```diff
@@ -0,0 +1,48 @@
+<%%= if @live_view_action in [:deps] do %>
+  <%%= live_modal @socket, DemoWeb.HomeLive.DepsCheck,
+    id: "deps",
+    redirect_path: Routes.home_path(@socket, :index) %>
+<%% end %>
```
josevalim: I think you can re-use the original index.html.leex as long as you wrap this first line in a `<% if live do %>`. Less duplication this way. WDYT?

installer/templates/phx_single/lib/app_name_web.ex
```diff
@@ -43,8 +47,34 @@ defmodule <%= web_namespace %> do
...
 
+  def live_component do
+    quote do
+      use Phoenix.LiveComponent
+      use Phoenix.HTML
+
+      import <%= web_namespace %>.ErrorHelpers<%= if gettext do %>
+      import <%= web_namespace %>.Gettext<% end %>
+      import <%= web_namespace %>.LiveHelpers
+      alias <%= web_namespace %>.Router.Helpers, as: Routes
```
josevalim: There is a good amount of duplication between view, live_component and live_view. We could do this:

```elixir
def live_component do
  quote do
    use Phoenix.LiveComponent
    unquote(view_helpers())
  end
end

defp view_helpers do
  quote do
    use Phoenix.HTML
    import <%= web_namespace %>.ErrorHelpers<%= if gettext do %>
    import <%= web_namespace %>.Gettext<% end %>
    import <%= web_namespace %>.LiveHelpers
    alias <%= web_namespace %>.Router.Helpers, as: Routes
  end
end
```

But then I don't know if we are trying too hard. Thoughts?

installer/templates/phx_web/endpoint.ex
```diff
@@ -12,7 +12,9 @@ defmodule <%= endpoint_module %> do
 
   socket "/socket", <%= web_namespace %>.UserSocket,
     websocket: true,
-    longpoll: false
+    longpoll: false<%= if live do %>
+
+  socket "/live", Phoenix.LiveView.Socket, websocket: [connect_info: [session: @session_options]]<% end %>
```
josevalim:
```suggestion
  socket "/live", Phoenix.LiveView.Socket,
    websocket: [connect_info: [session: @session_options]]<% end %>
```

installer/templates/phx_web/router.ex
```diff
@@ -1,10 +1,14 @@
 defmodule <%= web_namespace %>.Router do
-  use <%= web_namespace %>, :router<%= if html do %>
+  use <%= web_namespace %>, :router<%= if html do %><%= if live do %>
+
+  import Phoenix.LiveView.Router<% end %>
```
josevalim: We don't need this here. :) We've already added it to `use MyAppWeb, :router`.

installer/lib/phx_new/single.ex
```diff
@@ -155,4 +194,10 @@ defmodule Phx.New.Single do
   def gen_bare(%Project{} = project) do
     copy_from project, __MODULE__, :bare
   end
+
+  def assert_live_switches!(project) do
+    unless Project.html?(project) and Project.webpack?(project) do
+      raise "cannot generate --live project with --no-html or --no-webpack. LiveView requires HTML and webpack"
```
josevalim: It may be worth supporting a version where there is no webpack dependency either. Imagine if we could skip the need for node altogether! But this can be for a later step!

installer/lib/phx_new/web.ex
```diff
@@ -48,6 +48,20 @@ defmodule Phx.New.Web do
     {:eex,  "phx_test/views/page_view_test.exs",              :web, "test/:web_app/views/page_view_test.exs"},
   ]
 
+  template :live, [
+    {:eex, "phx_live/templates/layout/root.html.leex", :web, "lib/:web_app/templates/layout/root.html.leex"},
+    {:eex, "phx_live/templates/layout/app.html.leex",  :web, "lib/:web_app/templates/layout/app.html.leex"},
+    {:eex, "phx_live/templates/layout/live.html.leex", :web, "lib/:web_app/templates/layout/live.html.leex"},
+    {:eex, "phx_live/templates/page/home.html.leex",   :web, "lib/:web_app/templates/page/home.html.leex"},
```
josevalim: I mentioned this in the template itself, but we can probably re-use the same "page/index.html.eex" as long as we wrap the live view specific code in a conditional. This way it will be less things to us maintain. :)

installer/templates/phx_live/live/home_live.ex
josevalim: Right, let's talk about it but I would prefer to keep things as simple as possible and go with more complexity only when people call `mix phx.gen.live`. So for this particular example, I would propose to not use a modal (we can generate it on scaffold like we do with the action fallback), and simply have a div show up when we press the button to check for outdated deps, with no live navigation. If we can have some sort of CSS-based animation for the div showing up, then even better, because we can teach even more things to users.

installer/templates/phx_live/live/home_live.ex
```diff
@@ -0,0 +1,31 @@
...
+
+  def render(assigns) do
+    ~L"""
+    <h2>Results of <code>mix hex.outdated</code></h2>
+    <pre><%%= @deps %></pre>
+    """
+  end
+
+  def update(_assigns, socket) do
+    {deps, _} = System.cmd("mix", ["hex.outdated"])
```
josevalim: We probably want to do something like this:

```elixir
if not MyApp.Endpoint.config(:code_reloading) do
  raise "action disabled when not in development
end

{deps, _} = System.cmd("mix", ["hex.outdated"])
```

Just in case someone deploys the button to prod? 🤔

wojtekmach: I'm sure there will be many other ideas what component to ship with but anyway here's mine: live `IEx.Helpers.runtime_info()`, it can show interactivity (at least uptime changes every 1s) and doesn't require Mix so it's safe to ship to prod.

installer/templates/phx_web/endpoint.ex
chrismccord: This is what I had at first, but it failed `--check-formatted` and that pesky formatter wants it on ea single line :)

installer/templates/phx_web/router.ex
chrismccord: I was thinking I prefer it in the router since you don't often create new routers, and when you do, they will often carry their own concerns, ie API only. I don't have a super strong opinion and the import won't hurt other routes. Thoughts?

installer/lib/phx_new/single.ex
chrismccord: Baby steps. For now webpack is definitely required as otherwise phoenix.js and phoenix_live_view.js would have to be included on the layout, and referenced off the global Phoenix.

installer/lib/phx_new/web.ex
chrismccord: Agree. I wasn't sure what direction we were going to go content wise, but if it's 95% the same, then it makes sense to share

installer/templates/phx_live/live/home_live.ex
chrismccord: agree. I also had concerns with folks deploying to prod :D , but it wasn't likely to stay. I like @wojtekmach idea! I will give it a try

wojtekmach: glad you like it! If you want to be fancy there could be a list of checkboxes:

```
[x] system [x] memory [x] limits [ ] allocators [ ] applications 
```

that is transformed into list of atoms passed to runtime_info/1 :D

installer/templates/phx_web/endpoint.ex
josevalim: Bump the Elixir version. Since V1.7 or V1.8 we respect the user choice in here.

installer/templates/phx_web/router.ex
josevalim: I see your point. Having everything in there feels more consistent thought :/

installer/templates/phx_live/live/home_live.ex
josevalim: It could also be a radio button that shows one of them. One is always enabled. :)

installer/templates/phx_web/router.ex
chrismccord: Agree. app_web.ex it is!

installer/lib/phx_new/single.ex
```diff
@@ -47,6 +47,18 @@ defmodule Phx.New.Single do
     {:eex, "phx_test/views/page_view_test.exs",              :project, "test/:lib_web_name/views/page_view_test.exs"},
   ]
 
+  template :live, [
+    {:eex, "phx_live/templates/layout/root.html.leex",       :project, "lib/:lib_web_name/templates/layout/root.html.leex"},
+    {:eex, "phx_live/templates/layout/app.html.leex",        :project, "lib/:lib_web_name/templates/layout/app.html.leex"},
+    {:eex, "phx_live/templates/layout/live.html.leex",       :project, "lib/:lib_web_name/templates/layout/live.html.leex"},
+    {:eex, "phx_web/templates/page/index.html.eex",          :project, "lib/:lib_web_name/templates/page/home.html.leex"},
```
josevalim: Btw, should we just go ahead and call it "index.html"?

```suggestion
    {:eex, "phx_web/templates/page/index.html.eex",          :project, "lib/:lib_web_name/templates/page/index.html.leex"},
```

installer/templates/phx_web/templates/page/index.html.eex
```diff
@@ -1,7 +1,21 @@
...
   <h1><%%= gettext "Welcome to %{name}!", name: "Phoenix" %></h1><% else %>
   <h1>Welcome to Phoenix!</h1><% end %>
-  <p>A productive web framework that<br/>does not compromise speed or maintainability.</p>
+  <p>A productive web framework that<br/>does not compromise speed or maintainability.</p><%= if live do %>
+
+  <br/>
+  <h2>Live Hex module doc search</h2>
+  <form phx-change="suggest" phx-submit="search">
+    <input type="text" name="q" value="<%%= @query %>" placeholder="Module..." list="results" autocomplete="off"/>
+    <%%= if Enum.empty?(@results) and @query !== "" do %>
```
josevalim:
```suggestion
    <%%= if @results == [] and @query != "" do %>
```

installer/templates/phx_assets/app.css
```diff
@@ -0,0 +1,52 @@
+/* This file is for your main application css. */
+@import "./phoenix.css";<%= if live do %>
+@import "../node_modules/nprogress/nprogress.css";
+
+/* LiveView specific classes for your customizations */
```
josevalim: I think this should go in phoenix.css, everything else we use in scaffold or home pages go there.

PR #3690, opened 2020-03-02
Add trailing_slash option to Phoenix.Router using and match macro

### LINKED ISSUE #3573: Support trailing option in router definitions to generate paths with trailing slashes

mcrumm:
This PR adds a `:trailing_slash` option to Phoenix.Router that can be used to modify the generated helper functions:

1. You can define it globally when using Phoenix.Router:

```elixir
defmodule MyAppWeb.Router do
  use Phoenix.Router, trailing_slash: true
end
```

This will cause all generated `*_path` and `*_url` functions to append a trailing slash to the router paths.

2. You can define it per-route, like so:

```elixir
defmodule MyAppWeb.Router do
  use Phoenix.Router

  scope "/", MyAppWeb do
    get("/posts", PostController, :index)
    get("/users", UserController, :index, trailing_slash: true)
  end
end
```

```elixir
alias MyAppWeb.Router.Helpers, as: Routes

Routes.post_path(MyAppWeb.Endpoint, :index) # => "/posts"
Routes.user_path(MyAppWeb.Endpoint, :index) # => "/users/"
```

**Caveats**

This currently has no impact on calls (through the helpers) to `endpoint.url()`.  I'm not sure if this is a problem, and/or what the best solution might be, to ensure we avoid duplicating slashes somewhere in the path.

Closes #3573

### DIFF
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -554,6 +554,8 @@ defmodule Phoenix.Router do
     * `:assigns` - a map of data to merge into the connection when a route matches
     * `:metadata` - a map of metadata used by the telemetry events and returned by
       `route_info/4`
+    * `:trailing_slash` - a boolean to flag whether or not the helper functions
+      append a trailing slash. Defaults to `false`.
 
   ## Examples
 
@@ -827,6 +829,8 @@ defmodule Phoenix.Router do
     * `:assigns` - a map of data to merge into the connection when a route matches
     * `:log` - the level to log the route dispatching under,
       may be set to false. Defaults to `:debug`
+    * `:trailing_slash` - whether or not the helper functions append a trailing
+      slash. Defaults to `false`.
 
   """
   defmacro scope(options, do: context) do
--- a/lib/phoenix/router/route.ex
+++ b/lib/phoenix/router/route.ex
@@ -21,11 +21,12 @@ defmodule Phoenix.Router.Route do
     * `:assigns` - the route info
     * `:pipe_through` - the pipeline names as a list of atoms
     * `:metadata` - general metadata used on telemetry events and route info
-
+    * `:trailing_slash?` - whether or not the helper functions append a trailing slash
   """
 
   defstruct [:verb, :line, :kind, :path, :host, :plug, :plug_opts,
-             :helper, :private, :pipe_through, :assigns, :metadata]
+             :helper, :private, :pipe_through, :assigns, :metadata,
+             :trailing_slash?]
 
   @type t :: %Route{}
 
@@ -45,15 +46,17 @@ defmodule Phoenix.Router.Route do
   Receives the verb, path, plug, options and helper
   and returns a `Phoenix.Router.Route` struct.
   """
-  @spec build(non_neg_integer, :match | :forward, atom, String.t, String.t | nil, atom, atom, atom | nil, atom, map, map, map) :: t
-  def build(line, kind, verb, path, host, plug, plug_opts, helper, pipe_through, private, assigns, metadata)
+  @spec build(non_neg_integer, :match | :forward, atom, String.t, String.t | nil, atom, atom, atom | nil, atom, map, map, map, boolean) :: t
+  def build(line, kind, verb, path, host, plug, plug_opts, helper, pipe_through, private, assigns, metadata, trailing_slash?)
       when is_atom(verb) and (is_binary(host) or is_nil(host)) and
            is_atom(plug) and (is_binary(helper) or is_nil(helper)) and
            is_list(pipe_through) and is_map(private) and is_map(assigns) and
-           is_map(metadata) and kind in [:match, :forward] do
+           is_map(metadata) and kind in [:match, :forward] and
+           is_boolean(trailing_slash?) do
     %Route{kind: kind, verb: verb, path: path, host: host, private: private,
            plug: plug, plug_opts: plug_opts, helper: helper,
-           pipe_through: pipe_through, assigns: assigns, line: line, metadata: metadata}
+           pipe_through: pipe_through, assigns: assigns, line: line, metadata: metadata,
+           trailing_slash?: trailing_slash?}
   end
 
   @doc """
--- a/lib/phoenix/router/scope.ex
+++ b/lib/phoenix/router/scope.ex
@@ -6,7 +6,7 @@ defmodule Phoenix.Router.Scope do
   @pipes :phoenix_pipeline_scopes
   @top :phoenix_top_scopes
 
-  defstruct path: [], alias: [], as: [], pipes: [], host: nil, private: %{}, assigns: %{}, log: :debug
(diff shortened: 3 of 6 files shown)


lib/phoenix/router/route.ex
```diff
@@ -45,15 +46,17 @@ defmodule Phoenix.Router.Route do
   Receives the verb, path, plug, options and helper
   and returns a `Phoenix.Router.Route` struct.
   """
-  @spec build(non_neg_integer, :match | :forward, atom, String.t, String.t | nil, atom, atom, atom | nil, atom, map, map, map) :: t
-  def build(line, kind, verb, path, host, plug, plug_opts, helper, pipe_through, private, assigns, metadata)
+  @spec build(non_neg_integer, :match | :forward, atom, String.t, String.t | nil, atom, atom, atom | nil, atom, map, map, map, boolean | nil) :: t
+  def build(line, kind, verb, path, host, plug, plug_opts, helper, pipe_through, private, assigns, metadata, trailing_slash \\ nil)
```
chrismccord: I don't think we call this anywhere without the trailing_slash argument now, so can make it non-optional, unless I missed some calls?

josevalim:
@mcrumm I don't think we should support both styles. We either support it route based, so can you do:

```elixir
scope "/", trailing_slash: true do
```

Or the whole module one. Honestly, I would rather roll with the whole module one because the implementation is simpler - but I am not sure if this would satisfy the requirements.

lib/phoenix/router.ex
```diff
@@ -420,8 +421,9 @@ defmodule Phoenix.Router do
     routes_with_exprs = Enum.map(routes, &{&1, Route.exprs(&1)})
 
     helpers_moduledoc = Module.get_attribute(env.module, :helpers_moduledoc)
+    trailing_slash = Module.get_attribute(env.module, :trailing_slash)
 
-    Helpers.define(env, routes_with_exprs, docs: helpers_moduledoc)
+    Helpers.define(env, routes_with_exprs, docs: helpers_moduledoc, trailing_slash: trailing_slash)
```
josevalim: It seems this option is not being used in `Helpers.define`. Once we decide which mode we want to support, then we can decide what to do with this. :)

mcrumm: You're correct, this carried over from an earlier experiment.

mcrumm:
@josevalim I had originally intended only to add the module option, but the tests, specifically where the path is a filename with extension, led me to believe we'd need a route-based override for the module option: https://github.com/phoenixframework/phoenix/blob/b304b97205b3519ac13677d31c7970c6398d6958/test/phoenix/router/helpers_test.exs#L199-L204

josevalim:
I see. So we should probably remove the option given to `use Phoenix.Router`. :)

lib/phoenix/router/route.ex
```diff
@@ -45,15 +46,17 @@ defmodule Phoenix.Router.Route do
...
       when is_atom(verb) and (is_binary(host) or is_nil(host)) and
            is_atom(plug) and (is_binary(helper) or is_nil(helper)) and
            is_list(pipe_through) and is_map(private) and is_map(assigns) and
-           is_map(metadata) and kind in [:match, :forward] do
+           is_map(metadata) and kind in [:match, :forward] and
+           (is_nil(trailing_slash) or is_boolean(trailing_slash)) do
```
josevalim: Can we enforce it to be boolean in the caller (and non-optional, as @chrismccord said).

josevalim:
To be clear, let's remove the option on `use Phoenix.Router` and instead support it being given to `scope`.

lib/phoenix/router/route.ex
mcrumm: Done.

lib/phoenix/router/route.ex
mcrumm: Done.

mcrumm:
Thanks @josevalim!  I'll dive into scope bit a more and make the requisite changes.

mcrumm:
@josevalim @chrismccord I moved the option to the scope macro. The tests still need updates but I wanted to get 👀 on this first and make sure it looks okay before I proceed.

lib/phoenix/router/scope.ex
```diff
@@ -113,7 +114,8 @@ defmodule Phoenix.Router.Scope do
       pipes: top.pipes,
       private: Map.merge(top.private, private),
       assigns: Map.merge(top.assigns, assigns),
-      log: Keyword.get(opts, :log, top.log)
+      log: Keyword.get(opts, :log, top.log),
+      trailing_slash?: Keyword.get(opts, :trailing_slash, top.trailing_slash?) == true
```
mcrumm: I'm not sure if this is correct. Nested scopes don't appear to get the option from their parent scope, so maybe I'm not targeting something correctly for the default?

test/phoenix/router/helpers_test.exs
```diff
@@ -23,41 +23,43 @@ defmodule Phoenix.Router.HelpersTest do
   defmodule Router do
     use Phoenix.Router
 
-    get "/posts/top", PostController, :top, as: :top
-    get "/posts/bottom/:order/:count", PostController, :bottom, as: :bottom
-    get "/posts/:id", PostController, :show
-    get "/posts/file/*file", PostController, :file
-    get "/posts/skip", PostController, :skip, as: nil
+    scope "/", trailing_slash: true do
```
josevalim: I don't think we should add trailing slash  to all of our routes. Choose a select few (or preferably add a couple of new ones for testing).

lib/phoenix/router/scope.ex
josevalim: FWIW it looks correct to me. Are you seeing something that says differently? Do you have a failing test?

mcrumm: Not yet, but I'll add one to be sure.

test/phoenix/router/helpers_test.exs
mcrumm: Agreed. I made this change just to get the tests passing in their current form.  I'm going to revert these changes, as well as the test changes that added the trailing slashes to all the paths.

I'll add a new scope to the test router, and a block of tests to exercise the new option.

mcrumm:
@josevalim @chrismccord Ready for another look.  Lmk if I should expand/shrink the test cases at all. Thanks for all the input!

josevalim:

josevalim:
Shit, I forgot you could just merge it too, sorry!

paulstatezny:
I upgrade to `1.4.15` and I get this dialyzer error. Pretty certain it's related to this PR:

```
lib/dealer_api_web/router.ex:1:pattern_match
The pattern can never match the type.

Pattern:
95, _ = true

Type:
<<_::8, _::size(1)>>, false
```

mcrumm:
@paulstatezny This has been fixed. Use the 1.4 branch for the time being until a new release is published. Thanks!

PR #3694, opened 2020-03-04
Add missing floki package for phx.new --live

HusseinMorsy:
The floki package is missing when generating a new phoenix project with the option --live.

Without this package you will get the following error when running `mix test:

```
...
** (UndefinedFunctionError) function Floki.parse_document/1 is undefined (module Floki is not available)
...
```


josevalim:

PR #3695, opened 2020-03-04
Set correct webpack path for phoenix_live_view in umbrella apps

samgaw:
A quick fix against #3680 relating to the installer generating a `--live` app in an umbrella project.

Currently the Webpack template has a [hard coded reference](https://github.com/phoenixframework/phoenix/pull/3680/files#diff-d1254ba0442b3c34b3011c5e46f2ff17R12) to the live_view assets file path that is specific to a standard Phoenix project structure:
```
"phoenix_live_view": "file:../deps/phoenix_live_view" 
```

This copies how `phoenix_html` is set in the generator by checking whether it's in an umbrella project and inserts the correct file path.


chrismccord:
❤️❤️❤️🐥🔥

PR #3698, opened 2020-03-07
Standardize Endpoint Telemetry events

GregMefford:
This isn't complete yet (needs docs/tests updated) and would be a breaking change as written, but I wanted to open a discussion about whether this is a better way to handle this telemetry than what we're currently doing, and how we might move forward.

## Guiding Principles

* `Plug.Telemetry` doesn't offer a consistent way to ensure that there is always a `:start` and either a `:stop` or `:failure` event. This is important for consistent metrics and distributed tracing integration, as described in [this issue](https://github.com/beam-telemetry/telemetry/issues/57).
* In my opinion, `Phoenix.Endpoint` measurements should account for all the time that the `Endpoint` and any downstream code is executing (which includes sending the response, handling errors, etc.). `Plug.Telemetry` measures the duration between when that Plug is called and just before the response begins to be sent.
* Generating a `Plug.Telemetry` into the user's application code with that is namespaced as `phoenix.endpoint` feels weird to me, because it's in _their_ codebase, but it's documented in Phoenix itself. For example, if I change the prefix in my own `Endpoint` module or delete the plug altogether, it can no longer be easily consumed by other libraries. By comparison, if `phoenix.endpoint.stop` is emitted from Phoenix code as in this proposal, a drop-in library can just know how to consume it without any configuration and, for example, send that data to some vendor's APM solution.
* Users can still use `Plug.Telemetry` if they want to, but it will be clear to them that the event prefixes will be completely up to them and in no way related to Phoenix, since they added them by themselves.

## Problems to be Resolved

* Applications generated with the existing default `Plug.Telemetry` plug would end up having duplicated events unless they removed it from their `Endpoint` module. I would consider that a breaking change. We could obviously solve that by using a different name, but IMO this name is the correct name for this particular event, so I'm curious to hear what others would propose as a solution.
* The measurements and metadata are the same as the existing event, but now it's measuring a different thing, so the durations will seem to increase. This is also a breaking change because people who care about metrics will see a non-trivial "increase in latency" which is really just accounting for all the time spent between the `before_send` hooks fire and the actual end of the request. Again, this could be solved by naming the metric something else, but I'm not sure what else to call it.
* I didn't propose that we have any way to put a different prefix on the metric. We could consider putting the user's `MyAppWeb.Endpoint` module in the metadata to better support multiple Phoenix apps in an umbrella, but wasn't sure that was desired in addition to the `Plug.Conn` struct, and it would be an additive change to the existing metadata for this event.

/cc @bryannaegele @binaryseed @tsloughter

### DIFF
--- a/installer/templates/phx_web/endpoint.ex
+++ b/installer/templates/phx_web/endpoint.ex
@@ -37,7 +37,6 @@ defmodule <%= endpoint_module %> do
   end
 
   plug Plug.RequestId
-  plug Plug.Telemetry, event_prefix: [:phoenix, :endpoint]
 
   plug Plug.Parsers,
     parsers: [:urlencoded, :multipart, :json],
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -592,16 +592,32 @@ defmodule Phoenix.Endpoint do
         conn = put_in conn.script_name, script_name()
         conn = Plug.Conn.put_private(conn, :phoenix_endpoint, __MODULE__)
 
+        start = System.monotonic_time()
+        measures = %{time: System.system_time()}
+        meta = %{conn: conn, options: opts}
+        :telemetry.execute([:phoenix, :endpoint, :start], measures, meta)
         try do
           super(conn, opts)
         rescue
           e in Plug.Conn.WrapperError ->
             %{conn: conn, kind: kind, reason: reason, stack: stack} = e
+            measures = %{duration: System.monotonic_time() - start}
+            meta = %{kind: kind, error: reason, stacktrace: stack}
+            :telemetry.execute([:phoenix, :endpoint, :failure], measures, meta)
             Phoenix.Endpoint.RenderErrors.__catch__(conn, kind, reason, stack, config(:render_errors))
         catch
           kind, reason ->
             stack = __STACKTRACE__
+            measures = %{duration: System.monotonic_time() - start}
+            meta = %{kind: kind, error: reason, stacktrace: stack}
+            :telemetry.execute([:phoenix, :endpoint, :failure], measures, meta)
             Phoenix.Endpoint.RenderErrors.__catch__(conn, kind, reason, stack, config(:render_errors))
+        else
+          conn ->
+            measures = %{duration: System.monotonic_time() - start}
+            meta = %{conn: conn, options: opts}
+            :telemetry.execute([:phoenix, :endpoint, :stop], measures, meta)
+            conn
         end
       end
 


GregMefford:
Another issue I just noticed: should the `:failure` events get emitted _after_ the `Phoenix.Endpoint.RenderErrors.__catch__` call? I think the answer is probably yes, but I'm nervous what happens if something crashes inside there and we're already in a `rescue`.

josevalim:
Thanks @GregMefford!

If we are going to emit those, we need a different name, otherwise it is a backwards incompatible change. Also note that one benefit of Plug.Telemetry event is that it is included after static, since a lot of times we don't want to mix static reports with the other response data - and we often don't want info about static in our logs either. If you do want, you can just move it up.

I think the best course of action is to add telemetry to plug_cowboy, which will truly encompass the whole request. If we consider that, the benefit brought by this extra level of instrumentation is very low.

josevalim:
Actually, once we add it to PlugCowboy, we also need to add it here, as Phoenix uses its own handler.

tsloughter:
Will you have the "name" of the matched route in PlugCowboy?

josevalim:
@tsloughter that is only available for events downstream - when we reach the router. There is a route_dispatch event.

tsloughter:
I wouldn't bother emitting events before that. 

Maybe useful for a small group of people but any generic thing that will plugin to get metrics and spans will want the route and to ignore any event earlier than that.

bryannaegele:
Wouldn't the route would be there in plug_cowboy but it wouldn't be parsed,
just the raw string?

josevalim:
> I wouldn't bother emitting events before that.

Correct. And we do emit events earlier. So my proposal is to emit the following events:

  * [:plug_adapter, :request, :start] - the absolute beginning of the request
  * [:phoenix, :endpoint, :start] - application specific, user controlled
  * [:phoenix, :router_dispatch, :start] - once we reach the router

The last two are already implemented, the proposal is to add the first one (instead of this PR).

tsloughter:
@bryannaegele do you mean the url? That isn't what is needed. Metrics and traces need to use the match route, `/user/:user`, and not the path, `/user/cf16e4c1-d975-45ae-8b67-7c6798e2e4d1`, or you have explosions in dimensions.

bryannaegele:
They're fine in a trace, bad for metrics.

tsloughter:
Na, just as bad in a trace. Especially for some UI's like jaeger. It grinds to a halt if you are using the path instead of the matched route for the span name. We had/have that issue on a Phoenix project at Postmates :)

The recommendation is to use the route https://github.com/open-telemetry/opentelemetry-specification/blob/master/specification/data-http.md#name

bryannaegele:
Good to know.

So maybe post-match route is the best we can do.

GregMefford:
I think the HTTP URL/route doesn't _have_ to be the span name - the name could just be something like `Phoenix.Endpoint` or `PlugCowboy.call` with a span attribute of `http.target: /user/cf16e4c1-d975-45ae-8b67-7c6798e2e4d1`. Then once we get to dispatch time, we could pull out the route name into a different span attribute or a span name or whatever if needed for the back-end you're using.

I think there is value in including trace data from upstream of the parsed request, for context around how we arrived at that point in an overall distributed trace. Those could be modeled as a different service if your visualization tool requires that in order to search, etc.

So I think it makes sense to just close this PR, then. We can get essentially the same information from `PlugCowboy` without making a breaking change. My only issues with that solution are:
1. Functionality is spread across two repos (that's up to the maintainers to choose)
2. I feel like the naming convention isn't as obvious (also up to the maintainers and I can live it with)

PR #3699, opened 2020-03-07
Generate web view_helpers/0 when html not selected

mcrumm:
This fixes a compilation error in the generated {app_name}_web.ex file:

```
== Compilation error in file lib/demo_web.ex ==
** (CompileError) lib/demo_web.ex:40: undefined function view_helpers/0
    (elixir 1.10.2) src/elixir_locals.erl:114: anonymous fn/3 in :elixir_locals.ensure_no_undefined_local/3
    (stdlib 3.9) erl_eval.erl:680: :erl_eval.do_apply/6
    (elixir 1.10.2) lib/kernel/parallel_compiler.ex:304: anonymous fn/4 in Kernel.ParallelCompiler.spawn_workers/7
```

~The view_helpers/0 function is not generated in the web context if "--no-html" is used, so we need to check for html before generating the line to invoke view_helpers().~

Some of the imports/aliases in `view_helpers/0` are used even without html. This PR moves the html conditional into the function so that `view_helpers()` can be invoked in all cases.

## Examples

`mix phx.new demo --no-html --no-gettext`

```elixir
  defp view_helpers do
    quote do
      import DemoWeb.ErrorHelpers
      alias DemoWeb.Router.Helpers, as: Routes
    end
  end
```

`mix phx.new demo --no-html`

```elixir
  defp view_helpers do
    quote do
      import DemoWeb.ErrorHelpers
      import DemoWeb.Gettext
      alias DemoWeb.Router.Helpers, as: Routes
    end
  end
```

`mix phx.new demo`

```elixir
  defp view_helpers do
    quote do
      # Use all HTML functionality (forms, tags, etc)
      use Phoenix.HTML

      import DemoWeb.ErrorHelpers
      import DemoWeb.Gettext
      alias DemoWeb.Router.Helpers, as: Routes
    end
  end
```

`mix phx.new demo --live`

```elixir
  defp view_helpers do
    quote do
      # Use all HTML functionality (forms, tags, etc)
      use Phoenix.HTML

      # Import convenience functions for LiveView rendering
      import Phoenix.LiveView.Helpers

      import DemoWeb.ErrorHelpers
      import DemoWeb.Gettext
      alias DemoWeb.Router.Helpers, as: Routes
    end
  end
```

`mix phx.new demo --umbrella --no-html --no-context`

```elixir
  defp view_helpers do
    quote do
      import DemoWeb.ErrorHelpers
      alias DemoWeb.Router.Helpers, as: Routes
    end
  end
```

...

`mix phx.new demo --umbrella --live`

```elixir
  defp view_helpers do
    quote do
      # Use all HTML functionality (forms, tags, etc)
      use Phoenix.HTML

      # Import convenience functions for LiveView rendering
      import Phoenix.LiveView.Helpers

      import DemoWeb.ErrorHelpers
      import DemoWeb.Gettext
      alias DemoWeb.Router.Helpers, as: Routes
    end
  end
```

### DIFF
--- a/installer/templates/phx_single/lib/app_name_web.ex
+++ b/installer/templates/phx_single/lib/app_name_web.ex
@@ -34,10 +34,10 @@ defmodule <%= web_namespace %> do
         namespace: <%= web_namespace %>
 
       # Import convenience functions from controllers
-      import Phoenix.Controller, only: [get_flash: 1, get_flash: 2, view_module: 1]
+      import Phoenix.Controller, only: [get_flash: 1, get_flash: 2, view_module: 1]<%= if html do %>
 
       # Include shared imports and aliases for views
-      unquote(view_helpers())
+      unquote(view_helpers())<% end %>
     end
   end<%= if live do %>
 
--- a/installer/templates/phx_single/lib/app_name_web.ex
+++ b/installer/templates/phx_single/lib/app_name_web.ex
@@ -34,10 +34,10 @@ defmodule <%= web_namespace %> do
         namespace: <%= web_namespace %>
 
       # Import convenience functions from controllers
-      import Phoenix.Controller, only: [get_flash: 1, get_flash: 2, view_module: 1]<%= if html do %>
+      import Phoenix.Controller, only: [get_flash: 1, get_flash: 2, view_module: 1]
 
       # Include shared imports and aliases for views
-      unquote(view_helpers())<% end %>
+      unquote(view_helpers())
     end
   end<%= if live do %>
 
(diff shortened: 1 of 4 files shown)


josevalim:
Hi @mcrumm. I think the correct fix is actually to always generate view_helpers. That's because in there we have the router alias and the gettext import, that you may want to have even if not using HTML.

mcrumm:
Good catch! I'll make the requisite changes.

mcrumm:
@josevalim Ready for another look

josevalim:
Great! can you change the test that uses `--no-html` to assert  that a `defp view_helpers` is generated? Thanks!

PR #3701, opened 2020-03-08
Add Telemetry generator to installer

mcrumm:
This is still a work-in-progress.  Opening a draft PR to get early feedback.

### DIFF
--- a/installer/templates/phx_web/telemetry.ex
+++ b/installer/templates/phx_web/telemetry.ex
@@ -22,14 +22,8 @@ defmodule <%= web_namespace %>.Telemetry do
       summary("phoenix.endpoint.stop.duration",
         unit: {:native, :millisecond}
       ),
-      summary("phoenix.endpoint.stop.duration",
-        tags: [:method, :request_path],
-        tag_values: &tag_method_and_request_path/1,
-        unit: {:native, :millisecond}
-      ),
       summary("phoenix.router_dispatch.stop.duration",
-        tags: [:controller_action],
-        tag_values: &tag_controller_action/1,
+        tags: [:route],
         unit: {:native, :millisecond}
       ),<%= if ecto do %>
 
@@ -54,19 +48,5 @@ defmodule <%= web_namespace %>.Telemetry do
       # This function must call :telemetry.execute/3 and a metric must be added above.
       # {<%= web_namespace %>, :count_users, []}
     ]
-  end
-
-  # Extracts labels like "GET /"
-  defp tag_method_and_request_path(%{conn: conn}) do
-    Map.take(conn, [:method, :request_path])
-  end
-
-  # Extracts controller#action from route dispatch
-  defp tag_controller_action(%{plug: plug, plug_opts: plug_opts}) when is_atom(plug_opts) do
-    %{controller_action: "#{inspect(plug)}##{plug_opts}"}
-  end
-
-  defp tag_controller_action(%{plug: plug}) do
-    %{controller_action: inspect(plug)}
   end
 end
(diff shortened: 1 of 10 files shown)


installer/templates/phx_web/telemetry.ex
```diff
@@ -0,0 +1,52 @@
...
+  defp metrics do
+    [
+      # Phoenix Metrics
+      summary("phoenix.endpoint.stop.duration",
+        unit: {:native, :millisecond}
+      ),
+      summary("phoenix.router_dispatch.stop.duration",
+        tags: [:route],
+        unit: {:native, :millisecond}
+      ),<%= if ecto do %>
```
josevalim: @mcrumm I changed the metrics we emit here. Doing a tag per request path led to an explosition of data when trying it in prod, so I removed that metric. Doing per controller+action was not helpful for LiveView, so I am doing it per route instead.

mcrumm: Good call!

installer/test/phx_new_test.exs
```diff
@@ -155,6 +155,26 @@ defmodule Mix.Tasks.Phx.NewTest do
...
+        assert file =~ "{:telemetry_poller, measurements: periodic_measurements()"
+        assert file =~ "defp periodic_measurements do"
+        assert file =~ "# {PhxBlogWeb, :count_users, []}"
+        assert file =~ "defp metrics do"
+        assert file =~ "summary(\"phoenix.endpoint.stop.duration\","
+        assert file =~ "summary(\"phoenix.router_dispatch.stop.duration\","
+        assert file =~ "# Database Metrics"
+        assert file =~ "summary(\"phx_blog.repo.query.total_time\","
+        assert file =~ "defp tag_method_and_request_path"
+        assert file =~ "defp tag_controller_action"
```
mcrumm: Reminder to remove these since we removed the functions

installer/test/phx_new_umbrella_test.exs
```diff
@@ -182,6 +182,26 @@ defmodule Mix.Tasks.Phx.New.UmbrellaTest do
...
+        assert file =~ "{:telemetry_poller, measurements: periodic_measurements()"
+        assert file =~ "defp periodic_measurements do"
+        assert file =~ "# {PhxUmbWeb, :count_users, []}"
+        assert file =~ "defp metrics do"
+        assert file =~ "summary(\"phoenix.endpoint.stop.duration\","
+        assert file =~ "summary(\"phoenix.router_dispatch.stop.duration\","
+        assert file =~ "# Database Metrics"
+        assert file =~ "summary(\"phx_umb.repo.query.total_time\","
+        assert file =~ "defp tag_method_and_request_path"
+        assert file =~ "defp tag_controller_action"
```
mcrumm: Reminder to remove these since we removed the functions

mcrumm:
@josevalim two things:

1. the builds are failing because we're not installing Telemetry.Metrics in CI - I'm not sure the best way to fix that.

2. `phx.new.web` will generate each new umbrella app with vm metrics - I started seeing duplication in the events emitted when I added the console reporter to two new web apps in an umbrella.  I suppose this is only a problem if you run the umbrella app itself, but I wasn't sure if this is what should be the default behavior.

josevalim:
> the builds are failing because we're not installing Telemetry.Metrics in CI - I'm not sure the best way to fix that.

I pushed a commit that should fix it.

> phx.new.web will generate each new umbrella app with vm metrics 

Did you add the console reporter twice? Because the vm metrics are emitted by the built-in telemetry poller which is tied to the telemetry poller app, so you shouldn't have dups.

mcrumm:
Yes, there were two console reporters running, one in each of two `_web` apps in an umbrella. So that is the expected behavior :)

josevalim:
@mcrumm CI is good now, the only error is a heisenbug. We can ship this! :)

PR #3702, opened 2020-03-11
Add phx.gen.live

chrismccord:
Only thing left remaining is to flesh out the generated LiveView tests to fully exercise the CRUD functionality


lib/mix/tasks/phx.gen.live.ex
```diff
@@ -0,0 +1,242 @@
+defmodule Mix.Tasks.Phx.Gen.Live do
+  @shortdoc "Generates LiveView, templates, and context for a resource"
+
+  @moduledoc """
```
chrismccord: @josevalim I duped most of the `phx.gen.html` docs here, but I wasn't happy with a cross-ref because this may be the first encounter of a user to any phx task and the xref may cause them to miss our notes. Thoughts?

installer/templates/phx_live/templates/layout/root.html.leex
```diff
@@ -4,7 +4,7 @@
     <meta charset="utf-8"/>
     <meta http-equiv="X-UA-Compatible" content="IE=edge"/>
     <meta name="viewport" content="width=device-width, initial-scale=1.0"/>
-    <%%= live_title_tag assigns[:page_title] || <%= app_module %>, suffix: " · Phoenix Framework" %>
+    <%%= live_title_tag assigns[:page_title] || "<%= app_module %>", suffix: " · Phoenix Framework" %>
```
josevalim: I was thinking that maybe, for simplicity, we just do this:

```suggestion
    <%%= live_title_tag assigns[:page_title] || "<%= app_module %> · Phoenix Framework" %>
```

And let people explore the rest if they want to?

lib/mix/tasks/phx.gen.live.ex
josevalim: :+1:

lib/mix/tasks/phx.gen.live.ex
```diff
@@ -0,0 +1,242 @@
...
+      mix phx.gen.live Accounts User users name:string age:integer
+
+  The first argument is the context module followed by the schema module
+  and its plural name (used as the schema table name).
+
+  The context is an Elixir module that serves as an API boundary for
+  the given resource. A context often holds many related resources.
+  Therefore, if the context already exists, it will be augmented with
+  functions for the given resource.
+
```
josevalim: We should probably add a tiny paragraph about the LiveViews we generate here.

lib/mix/tasks/phx.gen.live.ex
```diff
@@ -0,0 +1,220 @@
...
+            live "/#{schema.plural}/:id", #{inspect(schema.alias)}Live.Show, :show
+            live "/#{schema.plural}/:id/show/edit", #{inspect(schema.alias)}Live.Show, :edit
+          end
+      """
+    else
+      Mix.shell().info """
+
+      Add the live routes to your browser scope in #{Mix.Phoenix.web_path(ctx_app)}/router.ex:
+
+          live "/#{schema.plural}", #{inspect(schema.alias)}Live.Index, :index
```
josevalim: The routes part is the same on both `if/else` branches, should we should probably avoid this duplication. If you want to avoid extra new lines, remember you can always use `\` at the last element in a new line in a string/heredoc to skip a newline.

priv/templates/phx.gen.live/form.ex
```diff
@@ -0,0 +1,56 @@
...
+  use <%= inspect context.web_module %>, :live_component
+
+  alias <%= inspect context.module %>
+
+  def render(assigns) do
+    Phoenix.View.render(<%= inspect context.web_module %>.<%= inspect Module.concat(schema.web_namespace, schema.alias) %>View, "form.html", assigns)
+  end
+
+  def update(%{<%= schema.singular %>: <%= schema.singular %>} = assigns, socket) do
+    changeset = socket.assigns[:changeset] || <%= inspect context.alias %>.change_<%= schema.singular %>(<%= schema.singular %>)
```
josevalim: Could we use `assign_new` here instead? 🤔

josevalim: Or alternatively, just always create a new one, this is not meant to be invoked multiple times anyway.

priv/templates/phx.gen.live/form.ex
```diff
@@ -0,0 +1,56 @@
...
+  def handle_event("save", %{"<%= schema.singular %>" => <%= schema.singular %>_params}, socket) do
+    save_<%= schema.singular %>(socket, socket.assigns.action, <%= schema.singular %>_params)
+  end
+
+  defp save_<%= schema.singular %>(socket, :edit, <%= schema.singular %>_params) do
+    case <%= inspect context.alias %>.update_<%= schema.singular %>(socket.assigns.<%= schema.singular %>, <%= schema.singular %>_params) do
+      {:ok, _<%= schema.singular %>} ->
+        {:noreply,
+         socket
+         |> put_flash(:info, "<%= schema.human_singular %> saved successfully")
```
josevalim:
```suggestion
         |> put_flash(:info, "<%= schema.human_singular %> updated successfully")
```

?

priv/templates/phx.gen.live/form.ex
```diff
@@ -0,0 +1,56 @@
...
+     socket
+     |> assign(assigns)
+     |> assign(changeset: changeset)}
+  end
+
+  def handle_event("validate", %{"<%= schema.singular %>" => <%= schema.singular %>_params}, socket) do
+    changeset =
+      socket.assigns.<%= schema.singular %>
+      |> <%= inspect context.alias %>.change_<%= schema.singular %>(<%= schema.singular %>_params)
+      |> Map.put(:action, :update)
```
josevalim: Why update? Is it an arbitrary choice? Maybe we can use `:validate` then?

priv/templates/phx.gen.live/form.ex
```diff
@@ -0,0 +1,56 @@
...
+
+    {:noreply, assign(socket, changeset: changeset)}
+  end
+
+  def handle_event("save", %{"<%= schema.singular %>" => <%= schema.singular %>_params}, socket) do
+    save_<%= schema.singular %>(socket, socket.assigns.action, <%= schema.singular %>_params)
+  end
+
+  defp save_<%= schema.singular %>(socket, :edit, <%= schema.singular %>_params) do
```
josevalim: It is a pity we have two separate clauses but I think unifying them is non trivial. We would need to add new functions to the context and odds are that those will grow apart anyway.

priv/templates/phx.gen.live/form.html.leex
```diff
@@ -0,0 +1,12 @@
...
+  id: "form-#{@id}",
+  phx_target: "#form-#{@id}",
+  phx_change: "validate",
+  phx_submit: "save" %>
+<%= for {label, input, error} <- inputs, input do %>
+  <%= label %>
+  <%= input %>
+  <%= error %>
+<% end %>
+  <%%= submit "Save", phx_disable_with:  "Saving..." %>
```
josevalim:
```suggestion
  <%%= submit "Save", phx_disable_with: "Saving..." %>
```

priv/templates/phx.gen.live/index.ex
```diff
@@ -0,0 +1,43 @@
+defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web_namespace, schema.alias) %>Live.Index do
+  use <%= inspect context.web_module %>, :live_view
+
+  alias <%= inspect context.module %>
+  alias <%= inspect schema.module %>
+
+  def render(assigns) do
+    Phoenix.View.render(<%= inspect context.web_module %>.<%= inspect Module.concat(schema.web_namespace, schema.alias) %>View, "index.html", assigns)
+  end
```
josevalim: Should we make this a convention? If you don't define a `render` clause, and the module name is `MyApp.FooLive.Index`, we will automatically call `MyApp.FooView, "index.html"`? I think it can positively reduce the amount of boilerplate.

priv/templates/phx.gen.live/index.ex
```diff
@@ -0,0 +1,43 @@
...
+
+  def mount(_params, _session, socket) do
+    {:ok,
+     socket
+     |> assign(page_title: gettext("Listing <%= schema.human_plural %>"))
+     |> assign(<%= schema.singular %>: nil)}
```
josevalim: This line is not necessary as we assign to it on `handle_params`.

priv/templates/phx.gen.live/index.ex
```diff
@@ -0,0 +1,43 @@
...
+
+  def handle_params(params, _, socket) do
+    <%= schema.singular %> =
+      case socket.assigns.live_view_action do
+        :edit -> <%= inspect context.alias %>.get_<%= schema.singular %>!(params["id"])
+        :new -> %<%= inspect schema.alias %>{}
+        :index -> nil
+      end
+
+    {:noreply, fetch(assign(socket, <%= schema.singular %>: <%= schema.singular %>))}
```
josevalim: Calling `fetch` here would reload all entries when we close the modal. Perhaps this is not what we want right now... but I guess users would want this if they add any sorting or pagination. Should we optimize for which scenario?

priv/templates/phx.gen.live/index.ex
```diff
@@ -0,0 +1,43 @@
...
+  end
+
+  def handle_event("edit", %{"id" => id}, socket) do
+    {:noreply, assign(socket, <%= schema.singular %>_id: id)}
+  end
```
josevalim: This is not used afaik?

priv/templates/phx.gen.live/form.ex
```diff
@@ -0,0 +1,56 @@
...
+        {:noreply, assign(socket, changeset: changeset)}
+    end
+  end
+
+  defp save_<%= schema.singular %>(socket, :new, <%= schema.singular %>_params) do
+    case <%= inspect context.alias %>.create_<%= schema.singular %>(<%= schema.singular %>_params) do
+      {:ok, _<%= schema.singular %>} ->
+        {:noreply,
+         socket
+         |> put_flash(:info, "<%= schema.human_singular %> created successfully")
```
josevalim: Maybe we should introduce `put_flash(info: "...")` so we can consistently use keywords syntax? Otherwise we use keyword syntax in one, the atom syntax in other cases, and it can be confusing for newcomers. Sticking with one or the other will most likely kelp.

priv/templates/phx.gen.live/live_test.exs
```diff
@@ -0,0 +1,36 @@
+defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web_namespace, schema.alias) %>LiveTest do
+  use <%= inspect context.web_module %>.ConnCase, async: true
```
josevalim: Should this be two separate live views?

josevalim: Btw, `async: true` here won't work for MySQL since they don't have concurrent tests in the sandbox. We either need to remove this or introduce a flag under `:generators` which controls the default value for async here.

```suggestion
  use <%= inspect context.web_module %>.ConnCase
```

priv/templates/phx.gen.live/index.ex
chrismccord: we necessarily have to call fetch if we want the updated entries or newly added entry to appear. The other option is going the fully optimized route of temporary_assigns + phx-update=append, but we were concerned before that it was too much to start with. Still same thoughts?

priv/templates/phx.gen.live/index.ex
josevalim: Note: we can remove this and colocate templates in latest LV.

priv/templates/phx.gen.live/index.ex
josevalim: My concern is the following:

1. When we open up the model, we do a query, and that is not necessary

2. When we close the model, we reload the data, and that is not necessary either

3. This may lead to usability issues, as people may not expect the underlying list to reload when I cancel the model.

However, I think we can solve 1 and 2 by rewriting it as:

```elixir
      case socket.assigns.live_view_action do
        :edit ->
          {:noreply, assign(socket, post: <%= inspect context.alias %>.get_<%= schema.singular %>!(params["id"]))}

        :new ->
          {:noreply, assign(socket, post: %<%= inspect schema.alias %>{})}

        :index ->
          {:noreply, assign(socket, post: nil) |> fetch()}
      end
```

Maybe that makes it good enough? Thoughts?

priv/templates/phx.gen.live/form.ex
```diff
@@ -0,0 +1,56 @@
...
+  defp save_<%= schema.singular %>(socket, :edit, <%= schema.singular %>_params) do
+    case <%= inspect context.alias %>.update_<%= schema.singular %>(socket.assigns.<%= schema.singular %>, <%= schema.singular %>_params) do
+      {:ok, _<%= schema.singular %>} ->
+        {:noreply,
+         socket
+         |> put_flash(:info, "<%= schema.human_singular %> updated successfully")
+         |> push_redirect(to: socket.assigns.return_to)}
+
+      {:error, %Ecto.Changeset{} = changeset} ->
+        {:noreply, assign(socket, changeset: changeset)}
```
josevalim:
```suggestion
        {:noreply, assign(socket, :changeset, changeset)}
```

priv/templates/phx.gen.live/form.ex
```diff
@@ -0,0 +1,56 @@
...
+  defp title(:new), do: "New <%= schema.human_singular %>"
+  defp title(:edit), do: "Edit <%= schema.human_singular %>"
+
+  def handle_event("validate", %{"<%= schema.singular %>" => <%= schema.singular %>_params}, socket) do
+    changeset =
+      socket.assigns.<%= schema.singular %>
+      |> <%= inspect context.alias %>.change_<%= schema.singular %>(<%= schema.singular %>_params)
+      |> Map.put(:action, :validate)
+
+    {:noreply, assign(socket, changeset: changeset)}
```
josevalim:
```suggestion
    {:noreply, assign(socket, :changeset, changeset)}
```

aptinio:
Ooh, can't wait to try this out! Thanks for all your real hard work @chrismccord, @josevalim, et al. :heart:

PR #3707, opened 2020-03-17
Add `mix setup` to generated apps

wojtekmach:
I'd like to propose that newly generated Phoenix apps have a `mix setup`
task (an alias) that sets up developer's environment: fetching mix
dependencies, preparing the database, installing npm deps, etc. This
way, instead of copy-pasting various commands from the README, there's
just one command to remember: `mix setup`. This also serves a good
starting point for developers to put their project-specific setup steps.

This is inspired by: https://github.com/github/scripts-to-rule-them-all.

I've used this pattern over the years on many projects, here are some OSS examples:

- https://github.com/hexpm/hexpm/blob/e52e883ec5360a674f604fed7b54d6c48c431687/mix.exs#L77
- https://github.com/hexpm/diff/blob/d86d1f2f61ffce9f2d69df5125781935487e3f19/mix.exs#L66
- https://github.com/elixir-lang/ex_doc/blob/v0.21.3/mix.exs#L40

Here's how this would look like:

    defp aliases() do
      [
        setup: ["deps.get", "ecto.setup", "cmd npm install --prefix assets"],
        "ecto.setup": ...,
        ...
      ]
    end

(Originally posted at https://groups.google.com/forum/#!topic/phoenix-core/s-Ws0gbzXIM.)

===

There's one caveat: the running commands are without a tty so there are
no progress bars, colors, etc which might be considered worse experience
than running these commands manually today. For me running just one
command by far outweight that downside.

This is a proof-of-concept, if you'd like to pursue this I'm happy to
finish this up including documenting and testing it!

### DIFF
--- a/installer/templates/phx_single/README.md
+++ b/installer/templates/phx_single/README.md
@@ -2,9 +2,7 @@
 
 To start your Phoenix server:
 
-  * Install dependencies with `mix deps.get`<%= if ecto do %>
-  * Create and migrate your database with `mix ecto.setup`<% end %><%= if webpack do %>
-  * Install Node.js dependencies with `npm install --prefix assets`<% end %>
+  * Setup the project with `mix setup`
   * Start Phoenix endpoint with `mix phx.server`
 
 Now you can visit [`localhost:4000`](http://localhost:4000) from your browser.
--- a/installer/templates/phx_single/mix.exs
+++ b/installer/templates/phx_single/mix.exs
@@ -12,8 +12,8 @@ defmodule <%= app_module %>.MixProject do
       elixir: "~> 1.7",
       elixirc_paths: elixirc_paths(Mix.env()),
       compilers: [:phoenix<%= if gettext do %>, :gettext<% end %>] ++ Mix.compilers(),
-      start_permanent: Mix.env() == :prod,<%= if ecto do %>
-      aliases: aliases(),<% end %>
+      start_permanent: Mix.env() == :prod,
+      aliases: aliases(),
       deps: deps()
     ]
   end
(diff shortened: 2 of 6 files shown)


wojtekmach:
Regarding the caveat about not having tty, an obvious fix is to not use Mix for running the command in the first place and add an `./bin/setup` shell script (like GitHub does.) I've avoided creating an additional file, I think mix.exs is a more obvious place to look for this, but I think this is worth mentioning anyway.

wojtekmach:
I've updated the PR description and the branch, we can simply use `mix cmd` to run arbitrary commands.

samgaw:
Will this not break in an umbrella project as aliases like `ecto.setup` are in a separate Mix project from where ever the Phoenix assets live?

wojtekmach:
I believe it won't: instead of defining `mix setup` in the umbrella root, you'd define it in each umbrella child app (some don't need a setup at all), then assets etc would be relative to where the `mix setup` is defined. But I haven't tried it yet and will make sure it works with an umbrella, good catch.

wojtekmach:
@samgaw it does not work in an umbrella out of the box but for a different reason. Because aliases cannot be specified as `@recursive true`, when we run `mix setup` in umbrella root it tries to find a task or alias in the _root_ and not finding it. A solution is to run in the root `mix cmd mix setup`, this will run `mix setup` in every child app. (You can alias it as: `setup: ["cmd mix setup"]` to run just `mix setup` in the root.)

chrismccord:
@wojtekmach this looks great! Can you add the changes for the `--umbrella` as well? Thanks!

wojtekmach:
updated!

chrismccord:
❤️❤️❤️🐥🔥

Issue #3669, opened 2020-02-13
Channel doesn't know when connection is disconnected and receive errors for all pushes after socket reconnects


myobie:
This concerns the `socket.js` client and not the server. I am using the `.js` file directly in the browser and my callsite js is written inside a `<script type="module">` and what you see below is what I am doing without any bundling or cross-compilation.

### Environment

* Operating system: macOS 10.15.3
* Browser: Safari

### Setup

```js
import { Socket } from './socket.js'

const socket = new Socket('/socket')
socket.connect()

const channel = socket.channel('topic')
channel.onError((data) => console.error('channel error', data))
channel.onClose((data) => console.error('channel close', data))
channel.join()

async function sleep(amount) {
  return new Promise((resolve, _reject) => {
    setTimeout(() => { resolve() }, amount)
  })
}
```

### Expected behavior

```js
(async () => {
  socket.disconnect()
  await sleep(1000)

  // Expect the channel to be in the same errored state it would be in if the remote side closed the connection
  assert(channel.state === "errored")

  // Pushing now will buffer and eventually timeout
  let push1 = channel.push("echo")
  await sleep(11000)
  assert(push1.receivedResp.status === 'timeout')

  // reconnect
  socket.connect()
  await sleep(1000)

  // pushing now will work, as the channel has rejoined
  let push2 = channel.push("echo")
  await sleep(1000)
  assert(push2.receivedResp.status === 'ok')
})()
```

### Actual behavior

```js
(async () => {
  socket.disconnect()
  await sleep(1000)

  // The channel stays joined, even tho if the remote had closed the socket it would end up in an errored state
  assert(channel.state === "joined")

  // Pushing now will buffer and eventually timeout, which is correct because the socket could reconnect
  let push1 = channel.push("echo")
  await sleep(11000)
  assert(push1.receivedResp.status === 'timeout')

  // reconnect
  socket.connect()
  await sleep(1000)

  // pushing now will fail as the remote channel is no longer a process and so there is no topic for this socket to match to
  let push2 = channel.push("echo")
  sleep(1000)
  assert(push2.receivedResp.status === 'error')
  assert(push2.receivedResp.response.reason === 'unmatched topic')
})()
```

chrismccord:
I don't see a `channel.join(...)` in the above code. Can you clarify?

myobie:
@chrismccord sorry, that was a bad copy/pasta on my part, I'll add it right now.

chrismccord:
[...]

PR #3708, opened 2020-03-18
Async teardown on disconnect

snewcomer:
close #3665 #3458 #3669 

There are a couple of things going on to fix a variety of issues and use cases for users:

**1.** `disconnect` now returns a Promise.  Cleanup on the connection is inherently async and non deterministic so this seems more inline with how it should work.

- Documentation on `onclose` && `readyState`: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/onclose. 
- After we call `conn.close()`, a WebSocket's `onclose` handler will be called at some unknown time in the future (next tick, maybe a few ticks in the browser run loop).

**2.** check `bufferedAmount` before closing the connection. This ensures no data is going across the network before closing. This is recommended here under "Closing the connection":

- https://developer.mozilla.org/en-US/docs/Web/API/WebSockets_API/Writing_WebSocket_client_applications


chrismccord:
Thanks Scott! I have admittedly avoided promises in the client so far, initially because of spotty support, but I think polyfills would accommodate everyone if needed? Alternatively, can this patch be changed to use callbacks instead? If there is not a specific need for a promise, I think disconnect accepting a closure is simpler. Thoughts?

chrismccord:
ping @snewcomer – Thoughts on above?

assets/js/phoenix.js
```diff
@@ -940,12 +940,46 @@ export class Socket {
   }
 
   teardown(callback, code, reason){
-    if(this.conn){
-      this.conn.onclose = function(){} // noop
-      if(code){ this.conn.close(code, reason || "") } else { this.conn.close() }
-      this.conn = null
+    if (!this.conn) {
```
indrekj, line 943: is this if needed? I see similar checks in waitForBufferDone and waitForSocketClosed

assets/js/phoenix.js
```diff
@@ -940,12 +940,46 @@ export class Socket {
...
+
+  waitForBufferDone(callback, tries = 1) {
+    if (tries === 5 || !this.conn || (this.conn.bufferedAmount && this.conn.bufferedAmount === 0)) {
+      callback()
+      return
+    }
+
+    setTimeout(() => {
+      this.waitForBufferDone(callback, tries + 1)
+    }, 150 * tries)
```
indrekj, line 971: can you name 150 and 5 above, so they'd be less magic numbers

assets/js/phoenix.js
```diff
@@ -940,12 +940,46 @@ export class Socket {
...
+          this.conn.onclose = function(){} // noop
+          this.conn = null
+        }
+
+        callback && callback()
+      })
+    })
+  }
+
+  waitForBufferDone(callback, tries = 1) {
```
indrekj, line 963: I think this and function below should be marked as private

assets/js/phoenix.js
```diff
@@ -940,12 +940,46 @@ export class Socket {
...
+    if (tries === 5 || !this.conn || (this.conn.bufferedAmount && this.conn.bufferedAmount === 0)) {
```
indrekj, line 964: Is the `this.conn.bufferedAmount &&` guard because of longpoll transport?

assets/js/phoenix.js
snewcomer: Yep we don't want to kick off the process if no `conn`

assets/js/phoenix.js
snewcomer: Private class fields are the future and while I agree I would introduce a `_` or something, I don't want to introduce a new pattern.

assets/js/phoenix.js
snewcomer: Yes

assets/js/phoenix.js
snewcomer: Also a great idea but b/c this is just one large file, creating const's is much less appealing.

assets/js/phoenix.js
indrekj: I mean above functions are annotated with:
```
  /**
   * @private
   */
```
comment.

assets/js/phoenix.js
indrekj: it could be in the same function as an assignment like `const maxTriesCount = 5`. but I guess it's fine as now as well. My only concern is that when changing code and I see timeout with number 150, then I have no idea how was this chosen and if it is safe to change. I usually add some information in the commit message or as a code comment how this was chosen.

chrismccord:
❤️❤️❤️🐥🔥

PR #3709, opened 2020-03-18
Removed dead repo link from docs

m1ome:
https://github.com/jfis/dn-phoenix is a dead link, it should be removed from docs


chrismccord:
❤️❤️❤️🐥🔥

PR #3710, opened 2020-03-18
Include install dependencies step in Release documentation

piisalie:
<3 <3 Ya'll are awesome thank you for everything <3 <3

When walking through the process of building a release for the first time, we've
seen a few people copy these steps out and expect them to run.

On a new server they fail, as `npm install` is not listed.

This causes a "webpack not found" error, which can be confusing (especially to
those without experience with JavaScript and its build process).


chrismccord:
❤️❤️❤️🐥🔥

PR #3713, opened 2020-03-19
Add LiveDashboard to phx.new

chrismccord:


installer/lib/mix/tasks/phx.new.ex
```diff
@@ -12,9 +12,12 @@ defmodule Mix.Tasks.Phx.New do
 
   ## Options
 
-    * `--live` - include Phoenix LiveView to make it easier than ever
+    * `--live` - include Phoenix.LiveView to make it easier than ever
       to build interactive, real-time applications
 
+    * `--dashboard` - include Phoenix.LiveDashboard which provides
+      real-time performance monitoring and debugging tools for developers.
+
```
chrismccord: I guess this is redundant because it is default. Should we say anything about this other than the --no-dashboard below?

josevalim: Yup, remove this one!

installer/templates/phx_single/lib/app_name_web.ex
```diff
@@ -64,7 +64,8 @@ defmodule <%= web_namespace %> do
 
       import Plug.Conn
       import Phoenix.Controller<%= if live do %>
-      import Phoenix.LiveView.Router<% end %>
+      import Phoenix.LiveView.Router<% end %><%= if dashboard do %>
+      import Phoenix.LiveDashboard.Router<% end %>
```
josevalim: For discussion: maybe we should import it in the router itself inside the `if Mix.env == :dev` block?

installer/templates/phx_web/router.ex
```diff
@@ -19,7 +19,21 @@ defmodule <%= web_namespace %>.Router do
...
-  end
+  end<%= if dashboard do %>
+
+  # Enables LiveDashboard only for development
+  #
+  # If you want to use the LiveDashboard in production, you should put
+  # it behind authentication and allow only admins to access it.
+  # If your application does not have an admins-only section yet,
+  # you can use Plug.BasicAuth to set up some basic authentication
+  # as long as you are also using SSL (which you should anyway)
```
josevalim:
```suggestion
  # as long as you are also using SSL (which you should anyway).
```

installer/templates/phx_single/lib/app_name_web.ex
chrismccord: It feels less foot-gun'y to leave it here?

josevalim: Having it together helps people understand where “live_dashboard” comes from though and where to look for more info.

josevalim: But I am happy either way!

installer/templates/phx_umbrella/apps/app_name_web/lib/app_name.ex
```diff
@@ -64,7 +64,8 @@ defmodule <%= web_namespace %> do
 
       import Plug.Conn
       import Phoenix.Controller<%= if live do %>
-      import Phoenix.LiveView.Router<% end %>
+      import Phoenix.LiveView.Router<% end %><%= if dashboard do %>
+      import Phoenix.LiveDashboard.Router<% end %>
```
josevalim:
```suggestion
      import Phoenix.LiveView.Router<% end %>
```

PR #3720, opened 2020-03-23
Update gitignore in umbrealla template.

grandafrato:
Updates a comment in the gitignore of the umbrella template to be up to date with the most recent version of phoenix's directory structure.


Gazler:
Thanks <3

PR #3721, opened 2020-03-23
Adds two options to the generated error_tag helper.

joshchernoff:
The options add a way to specify the content_tag's name and class.

example. 
`<%= error_tag f, :some_field, class: "text-red-400 error", tag: :div %>`


josevalim:
Hi @joshchernoff. In those cases, I would almost ways recommend for you to change the error_tag directly and make it app specific instead of supporting options. We want the generated code to be simple, so if options are necessary, you can add them yourself. Thanks!

PR #3722, opened 2020-03-26
Handle 'Accept: ' header with missing value as '*/*'

OldhamMade:
Requests to phoenix which contain an accept header with an empty value (eg: `Accept: `) cause `Phoenix.NotAcceptableError` to be raised. There seems to be no way to handle this through changes to the `config :mime, :types` setting.

This change allows such headers to be handled as "any" (`*/*`).

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -1272,6 +1272,12 @@ defmodule Phoenix.Controller do
     end
   end
 
+  # If the accept header is empty, we treat it as */* and
+  # we use the first format specified in the accepts list.
+  defp parse_header_accept(conn, [], [], [first|_]) do
+    put_format(conn, first)
+  end
+
   defp parse_header_accept(conn, [], acc, accepted) do
     acc
     |> Enum.sort()
--- a/test/phoenix/controller/controller_test.exs
+++ b/test/phoenix/controller/controller_test.exs
@@ -355,6 +355,16 @@ defmodule Phoenix.Controller.ControllerTest do
       assert conn.params["_format"] == nil
     end
 
+    test "treats empty accept header as any" do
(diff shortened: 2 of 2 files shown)


OldhamMade:
Odd, [travis seems to have passed the tests](https://travis-ci.org/github/phoenixframework/phoenix/builds/667423171) but it hasn't updated this PR.

josevalim:
@OldhamMade when would empty accept headers be sent? Is it even according to the spec?

josevalim:
In any case, thanks for the PR!

OldhamMade:
> @OldhamMade when would empty accept headers be sent?

I have a free-to-use public service which receives a few million requests per day, and a percentage of those come from clients which set an empty `Accept: ` header. I have no control over the clients, so this PR provides a fallback.

> Is it even according to the spec?

I don't think it is! I _think_ it's happening because of default settings not being overridden in client code. A lot of the "empty" requests come from `OkHttp`, maybe there's a quick-start snippet which is being copy/pasted with and empty header and used as-is.

josevalim:
Yes, this is tricky. I would say the best behaviour would be to raise as it does today, especially if it is not supported by the spec.

Have you considered having your own plug that writes the accept header if it is empty?

OldhamMade:
> Have you considered having your own plug that writes the accept header if it is empty?

I did try this, however adding the custom plug after `use Phoenix.Endpoint` in my app's `Endpoint` didn't work. Is there an earlier point where I can introduce a plug? If so, I'll withdraw this PR and create a separate hex package.

josevalim:
We only perform content negotiation in the router, so if you do it anywhere in the endpoint, it should be enough, no?

josevalim:
Closing per discussion above, thanks!

PR #3725, opened 2020-03-29
Fix devMode in webpack.config.js

frankdugan3:
When using `webpack --mode production`, webpack does not set `NODE_ENV` within the build script itself [(see docs)](https://webpack.js.org/guides/production/#specify-the-mode). The result is that `devMode` will always be true in the current `webpack.config.js`, even when running the `deploy` script.

This PR checks the `options.mode` property to set `devMode`. A common alternative is to set `"deploy": "env NODE_ENV=production webpack --mode production"` in `package.json`, but I think this method is cleaner and removes cross-platform concerns about how to set environment variables.

### DIFF
--- a/installer/templates/phx_assets/webpack.config.js
+++ b/installer/templates/phx_assets/webpack.config.js
@@ -4,49 +4,52 @@ const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const TerserPlugin = require('terser-webpack-plugin');
 const OptimizeCSSAssetsPlugin = require('optimize-css-assets-webpack-plugin');
 const CopyWebpackPlugin = require('copy-webpack-plugin');
-const devMode = process.env.NODE_ENV !== 'production'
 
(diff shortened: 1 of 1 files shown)


snewcomer:
@frankdugan3 Are you indicating the current script does not trigger a production build for you?

frankdugan3:
@snewcomer A production build does trigger, but `devMode` is still `true` in a production build with the current script, so anything in `webpack.config.js` that depends on the value of `devMode` will always think it's not a production build.

otijhuis:
I agree that it would be nice not to have `NODE_ENV=production` in the script. The problem right now though is that this still doesn't work for something like a `postcss.config.js`  when you check for NODE_ENV to only enable `purgecss` in production. Most online guides follow that same pattern and it's very confusing when it doesn't work.

sntran:
I also have the same issue as @otijhuis when using `purgecss`. `NODE_ENV` is still not "production".

PR #3734, opened 2020-04-01
Update doc for `get_flash/2`

mathieuprog:
Specify that `nil` is returned if there is no message for `key`


josevalim:

PR #3736, opened 2020-04-02
Add phx.gen.live to list of context-related generators

aaronrenner:
This adds `phx.gen.live` to the list of context-related generators (`phx.gen.html`, `phx.gen.json` and `phx.gen.context`) everywhere they're listed together.


josevalim:

PR #3737, opened 2020-04-02
Allow phx.gen.context's help to be extended

aaronrenner:
This allows generators that build on top of `phx.gen.context` to pass in their own help message while using `phx.gen.context`'s argument validation logic. This is the identical pattern that's being used in `phx.gen.schema` to allow `phx.gen.context` to override the help message.

https://github.com/phoenixframework/phoenix/blob/165ff2d131f4494243440d16fe6558d3787b3972/lib/mix/tasks/phx.gen.schema.ex#L122-L135

https://github.com/phoenixframework/phoenix/blob/165ff2d131f4494243440d16fe6558d3787b3972/lib/mix/tasks/phx.gen.schema.ex#L174-L201

This change will allow `phx.gen.auth` to print its own error messages instead of printing out error messages that mention `phx.gen.html` and friends. This could also be useful to allow `phx.gen.live` to have a custom error messages.

### DIFF
--- a/lib/mix/tasks/phx.gen.context.ex
+++ b/lib/mix/tasks/phx.gen.context.ex
@@ -100,11 +100,11 @@ defmodule Mix.Tasks.Phx.Gen.Context do
   end
 
   @doc false
-  def build(args) do
+  def build(args, help \\ __MODULE__) do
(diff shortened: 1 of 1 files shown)


josevalim:

PR #3706, opened 2020-03-15
Heroku Guide Improvement: Add instructions to configure Node and NPM

mcelaney:
Adds an example of configuring Node and NPM in `phoenix_static_buildpack.config` so that devs pushing new builds know how to overcome the Node v6.9.2 and NPM v3.10.9 defaults in the Phoenix Static buildpack.

PR #3739, opened 2020-04-03
heroku: don't suggest setting both node and npm version

petermm:
per discussion in https://github.com/phoenixframework/phoenix/pull/3706

setting both node and npm - will lead to issues down the line (eg when user wants to use node 12 etc) - it's also complicated figuring out what npm version to actually use etc. etc.

we are still conservative on suggesting node 10 - could be future proofing going 12.16.1.. but that is separate issue..


Gazler:
Thanks <3

PR #3740, opened 2020-04-03
Fix failing test in new --live project

aptinio:
because Routes.live_dashboard_path/2
called from root.html.leex is undefined.


josevalim:
Closing in favor of #3755, thanks! 

PR #3742, opened 2020-04-04
 Add set_allow_snapshot_isolation: :on to config/test.exs for MSSQL

aaronrenner:
This automatically adds the `set_allow_snapshot_isolation: :on` setting to new phoenix apps using MSSQL databases. This setting makes MSSQL compatible with Ecto's SQL sandbox and allows the
user to run their database-backed tests concurrently.

### DIFF
--- a/installer/lib/phx_new/generator.ex
+++ b/installer/lib/phx_new/generator.ex
@@ -221,7 +221,14 @@ defmodule Phx.New.Generator do
   end
 
   defp get_ecto_adapter("mssql", app, module) do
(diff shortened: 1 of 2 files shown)


aaronrenner:
Just found a deadlock issue with mssql: https://github.com/elixir-ecto/ecto_sql/issues/221

We should probably wait until that issue is resolved before considering merging this change.

aaronrenner:
It looks like SQL sandbox isn't quite ready to support transactional tests with MSSQL.

PR #3743, opened 2020-04-05
Revise phx.gen.presence module doc

pseigo:
After #2554, an excerpt from `mix help phx.gen.context` reads:

> A new file will be generated in lib/my_app_web/channels/**presence.ex**, where **my_app_web** is the snake cased version of the module name provided.

But see the actual behaviour:

```elixir
$ mix phx.gen.presence MyPresence 
* creating lib/app_name_web/channels/my_presence.ex 
```

Changed this in the doc and did some cleaning up.


Gazler:
Thanks <3

PR #3746, opened 2020-04-05
token salt can be treated as namespace

achempion:
Improves the documentation about token generation.

First time I read the documentation about `Phoenix.Token` and saw the
`salt`, I become clear to me stat salt should be some kind of
randomly generated string.

After exploring the codebase and [researching the
internet](https://elixirforum.com/t/how-is-phoenix-token-different-from-jwt/2349/5)
I found out my code readability can be improved by replacing randomly generated string with a description of the token namespace.

I think this change will help improve the usability of the `Phoenix.Token` module by hinting that salt can be namespace as well.

I found `user_salt` variable a little bit confusing because a user doesn't have salt and if it has, it'll be some randomly generated string for password hashing which is not the case for `Phoenix.Token`.

I would like to propose the change of argument naming from `salt` to
`namespace` to completely remove thoughts about `salt` as a randomly generated string.

### DIFF
--- a/lib/phoenix/token.ex
+++ b/lib/phoenix/token.ex
@@ -15,8 +15,8 @@ defmodule Phoenix.Token do
   the id from a database. For example:
 
       iex> user_id = 1
-      iex> token = Phoenix.Token.sign(MyApp.Endpoint, "user salt", user_id)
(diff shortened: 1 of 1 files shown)


josevalim:

PR #3750, opened 2020-04-07
Fix implementation (and docs) for logging requests

fertapric:
The previous implementation was:

```elixir
level = metadata[:options][:log] || :info
```

If `log: false` was given to disable the logs (as the documentation specifies), that expression was not disabling the logs, it was setting the log level to the default value `:info`.

In addition, the `:log` option given to `Plug.Telemetry` must not be under an `:options` key, because as the `Plug.Telemetry` documentation specifies: "the metadata of the event is the  whole `Plug.Conn` under the `:conn` key and any leftover options given to the plug under `:options`".


josevalim:

PR #3752, opened 2020-04-08
Fix typo in Live index template

theharq:
This change fixes a typo error in the template generated by LiveView.

I found the issue while testing the new `phx.gen.live` generator.

PR #3753, opened 2020-04-08
Fix phx.gen.live tr#id

leandrocp:
Replace hard coded `post.id` with `schema.singular.id`


leandrocp:
https://github.com/phoenixframework/phoenix/pull/3752

PR #3755, opened 2020-04-09
Fix dashboard installer, closes #3740, closes #3751

### LINKED ISSUE #3751: Compilation error on apps generated with --no-html
### Environment

* Elixir version (elixir -v): 1.10.2
* Phoenix version (mix deps): master@9d7444f8
* NodeJS version (node -v): 10.19.0
* NPM version (npm -v): 6.13.4
* Operating system: Ubuntu 19.10

### Expected behavior

After running `mix phx.new app_with_no_html --no-html`, I should be able to run `mix test` and all of the tests should pass.

### Actual behavior

When I run `mix test` after generating a new project with `--no-html`, I get the following error.

```
$ mix test
Compiling 1 file (.ex)

== Compilation error in file lib/app_with_no_html_web/endpoint.ex ==
** (UndefinedFunctionError) function Phoenix.LiveDashboard.RequestLogger.init/1 is undefined (module Phoenix.LiveDashboard.RequestLogger is not available)
    Phoenix.LiveDashboard.RequestLogger.init([param_key: "request_logger", cookie_key: "request_logger"])
    (plug 1.10.0) lib/plug/builder.ex:304: Plug.Builder.init_module_plug/4
    (plug 1.10.0) lib/plug/builder.ex:288: anonymous fn/5 in Plug.Builder.compile/3
    (elixir 1.10.2) lib/enum.ex:2111: Enum."-reduce/3-lists^foldl/2-0-"/3
    (plug 1.10.0) lib/plug/builder.ex:286: Plug.Builder.compile/3
    (plug 1.10.0) expanding macro: Plug.Builder.__before_compile__/1
    lib/app_with_no_html_web/endpoint.ex:1: AppWithNoHtmlWeb.Endpoint (module)
```

The problem appears to be that the live dashboard code is being injected into `endpoint.ex` when its not injected elsewhere. If I comment out the following lines, in `endpoint.ex` I can get the tests to pass.

```elixir
  socket "/live", Phoenix.LiveView.Socket, websocket: [connect_info: [session: @session_options]]
```
and

```elixir
  plug Phoenix.LiveDashboard.RequestLogger,
    param_key: "request_logger",
    cookie_key: "request_logger"
```

mcrumm:
This corrects a number of issues introduced with the addition of :phoenix_live_dashboard.

- Adds `live_dashboard_path` to the `app.html.eex` dead view
- Conditions `live_dashboard` route on existence so the generated tests pass (#3740 // Thanks @aptinio! )
- Ensures the dashboard works even when `--no-html` is given (#3751 // Thanks @aaronrenner!)

When generated in a project without both html and live, we have to close the gap wrt the session. The router in such cases would look as follows:

```elixir
defmodule MyAppWeb.Router do
  use MyAppWeb, :router

  pipeline :api do
    plug :accepts, ["json"]
  end

  scope "/api", DevAppWeb do
    pipe_through :api
  end

  # Enables LiveDashboard only for development
  #
  # If you want to use the LiveDashboard in production, you should put
  # it behind authentication and allow only admins to access it.
  # If your application does not have an admins-only section yet,
  # you can use Plug.BasicAuth to set up some basic authentication
  # as long as you are also using SSL (which you should anyway).
  if Mix.env() in [:dev, :test] do
    import Phoenix.LiveDashboard.Router

    scope "/" do
      pipe_through [:fetch_session, :protect_from_forgery]
      live_dashboard "/dashboard", metrics: MyAppWeb.Telemetry
    end
  end
end
```

### DIFF
--- a/installer/templates/phx_live/templates/layout/root.html.leex
+++ b/installer/templates/phx_live/templates/layout/root.html.leex
@@ -13,7 +13,7 @@
       <section class="container">
         <nav role="navigation">
           <ul>
-            <li><a href="https://hexdocs.pm/phoenix/overview.html">Get Started</a></li><%= if live and dashboard do %>
+            <li><a href="https://hexdocs.pm/phoenix/overview.html">Get Started</a></li><%= if dashboard do %>
             <li><%%= link "LiveDashboard", to: Routes.live_dashboard_path(@conn, :home) %></li><% end %>
           </ul>
         </nav>
--- a/installer/templates/phx_live/templates/layout/root.html.leex
+++ b/installer/templates/phx_live/templates/layout/root.html.leex
@@ -13,8 +13,8 @@
       <section class="container">
         <nav role="navigation">
           <ul>
-            <li><a href="https://hexdocs.pm/phoenix/overview.html">Get Started</a></li><%= if dashboard do %>
-            <li><%%= link "LiveDashboard", to: Routes.live_dashboard_path(@conn, :home) %></li><% end %>
+            <li><a href="https://hexdocs.pm/phoenix/overview.html">Get Started</a></li><%= if dashboard do %><%%= if function_exported?(Routes, :live_dashboard_path, 2) do %>
+            <li><%%= link "LiveDashboard", to: Routes.live_dashboard_path(@conn, :home) %></li><%% end %><% end %>
           </ul>
         </nav>
         <a href="https://phoenixframework.org/" class="phx-logo">
--- a/installer/templates/phx_live/templates/layout/root.html.leex
+++ b/installer/templates/phx_live/templates/layout/root.html.leex
@@ -13,8 +13,10 @@
       <section class="container">
         <nav role="navigation">
           <ul>
-            <li><a href="https://hexdocs.pm/phoenix/overview.html">Get Started</a></li><%= if dashboard do %><%%= if function_exported?(Routes, :live_dashboard_path, 2) do %>
-            <li><%%= link "LiveDashboard", to: Routes.live_dashboard_path(@conn, :home) %></li><%% end %><% end %>
+            <li><a href="https://hexdocs.pm/phoenix/overview.html">Get Started</a></li>
+            <%= if dashboard do %><%%= if function_exported?(Routes, :live_dashboard_path, 2) do %>
+              <li><%%= link "LiveDashboard", to: Routes.live_dashboard_path(@conn, :home) %></li>
+            <%% end %><% end %>
           </ul>
         </nav>
         <a href="https://phoenixframework.org/" class="phx-logo">
--- a/installer/templates/phx_web/router.ex
+++ b/installer/templates/phx_web/router.ex
@@ -37,7 +37,7 @@ defmodule <%= web_namespace %>.Router do
   # If your application does not have an admins-only section yet,
   # you can use Plug.BasicAuth to set up some basic authentication
   # as long as you are also using SSL (which you should anyway).
-  if Mix.env() == :dev do
+  if Mix.env() in [:dev, :test] do
     import Phoenix.LiveDashboard.Router
 
     scope "/" do<%= if html do %>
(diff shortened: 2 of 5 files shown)


installer/templates/phx_web/router.ex
```diff
@@ -28,21 +37,15 @@ defmodule <%= web_namespace %>.Router do
   # If your application does not have an admins-only section yet,
   # you can use Plug.BasicAuth to set up some basic authentication
   # as long as you are also using SSL (which you should anyway).
-  if Mix.env() == :dev do
-    import Phoenix.LiveDashboard.Router
+  if Mix.env() in [:dev, :test] do
```
josevalim, line 40: Unfortunately, this means if someone deploys a new app with the layout as is, the production version won't work. My suggestion is to keep this dev only and do `<%= if function_exported?(Routes, :live_dashboard_path, 2) do %>` in the layout.

installer/templates/phx_web/router.ex
```diff
@@ -28,21 +37,15 @@ defmodule <%= web_namespace %>.Router do
...
+    import Phoenix.LiveDashboard.Router<%= unless html && live do %>
+
+    defdelegate fetch_live_flash(conn, opts), to: Phoenix.LiveView.Router<% end %>
```
josevalim, line 43: We don't use flash on the dashboard, so we don't need to add this. :)

mcrumm: There's a `put_flash` in there for a node disconnection message, or at least there was yesterday :)

josevalim: @mcrumm good call! Note we don't read from the flash anywhere (no calls to @flash or live_flash), so it is not used anyway. So I think we can remove this here and we find an alternative for the flash message on the dashboard. :)

installer/templates/phx_web/router.ex
mcrumm: Updated.  This leads to the following warning in non-dev environments:

```
warning: DeadAppWeb.Router.Helpers.live_dashboard_path/2 is undefined or private
```

Is this acceptable, or can/should I do more to silence the warning?

installer/templates/phx_live/templates/layout/root.html.leex
```diff
@@ -13,8 +13,8 @@
       <section class="container">
         <nav role="navigation">
           <ul>
-            <li><a href="https://hexdocs.pm/phoenix/overview.html">Get Started</a></li><%= if live and dashboard do %>
-            <li><%%= link "LiveDashboard", to: Routes.live_dashboard_path(@conn, :home) %></li><% end %>
+            <li><a href="https://hexdocs.pm/phoenix/overview.html">Get Started</a></li><%= if dashboard do %><%%= if function_exported?(Routes, :live_dashboard_path, 2) do %>
+            <li><%%= link "LiveDashboard", to: Routes.live_dashboard_path(@conn, :home) %></li><%% end %><% end %>
```
josevalim, line 17: Would this indent better?

```suggestion
            <li><a href="https://hexdocs.pm/phoenix/overview.html">Get Started</a></li>
            <%= if dashboard do %><%%= if function_exported?(Routes, :live_dashboard_path, 2) do %>
              <li><%%= link "LiveDashboard", to: Routes.live_dashboard_path(@conn, :home) %></li>
            <%% end %><% end %>
```

installer/templates/phx_web/router.ex
```diff
@@ -31,18 +40,10 @@ defmodule <%= web_namespace %>.Router do
   if Mix.env() == :dev do
```
josevalim, line 40: Yeah, given the warning, I would go with:

```suggestion
  if Mix.env() in [:dev, :test] do
```

installer/templates/phx_web/router.ex
josevalim: Let's silence the warning in test and we are good. I dropped a suggestion.

installer/templates/phx_live/templates/layout/root.html.leex
mcrumm: Selfishly I want to drop the leading spaces so the rendered HTML is _also_ indented properly, but yes this looks better :) :)

installer/templates/phx_web/router.ex
mcrumm: Done, thanks!

PR #3759, opened 2020-04-14
Remove live view signing salt in non-liveview project

nittin-shankar:
When running `mix phx.new project_name`, I get a  live_view signing salt config in `config.exs`, although it's not a live_view project. This PR conditionally adds this config only for live_view projects.


josevalim:
I am in favor of keeping it around simply because it makes it easier for people to add liveview-related projects in the future, such as the phx live dashboard. IN any case, thanks!

PR #3760, opened 2020-04-14
Fix helper paths for phx.gen.live with --web

mcrumm:
- Ensure scoped routes via `--web` are generated properly.  (e.g. `user_path ==> admin_user_path`)
- Select the form component via the `"form"` tag (so it works with at-myself)
- Fix formatting (albeit there may still be some formatting to fix in the generated app, depending on the length of the path helper)

// @aaronrenner


josevalim:

PR #3761, opened 2020-04-15
Add support to `anchor` param on router helpers

rrrcompagnoni:
## Objective
Add support to `anchor` param on router helpers.

### Details
Recently I had a specific use case for navigation based on hash anchor coming from the URL, I was trying to understand how could I do it by the current support from the URL helpers and I thought that could be useful add this support by default.

I have applied the format tool on those files and I am not sure if it would be appreciated, let me know if I should revert it. Due to it, the review might be a little hard, the proposal changes start on 
`segments` function at `lib/phoenix/router/helpers.ex:350` and are being tested on `url helper with query strings` spec beginning at `test/phoenix/router/helpers_test.exs:114`.

### DIFF
--- a/lib/phoenix/router/helpers.ex
+++ b/lib/phoenix/router/helpers.ex
@@ -6,10 +6,10 @@ defmodule Phoenix.Router.Helpers do
   alias Plug.Conn
 
   @anno (if :erlang.system_info(:otp_release) >= '19' do
-    [generated: true, unquote: false]
-  else
-    [line: -1, unquote: false]
-  end)
+           [generated: true, unquote: false]
+         else
+           [line: -1, unquote: false]
+         end)
 
   @doc """
   Callback invoked by the url generated in each helper module.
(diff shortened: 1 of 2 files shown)


chrismccord:
Unfortunately we cannot rely on an `anchor` key as it would be a backwards incompatible change since `anchor` would otherwise be treated as a query param. Your best bet today is to wrap your anchor generation in your own function. In the future, please leave formatting out. Thank you !

henrik:
Found this PR when trying to figure out why there wasn't something built in for anchors/fragments. Makes sense not to want the ambiguity of an "anchor" (or "fragment"?) key among the params.

The least bad alternatives I can think of would be to either allow an additional keyword list or possibly a positional argument, e.g.

``` elixir
# Option A
Routes.foo_path(@conn, :show, 123, [my_param: "hi"], anchor: "yo")

# Option B
Routes.foo_path(@conn, :show, 123, [my_param: "hi"], "yo")
```

or to name the key in a way that it's highly unlikely that it'd ever be intended as a param:

``` elixir
# Option C
Routes.foo_path(@conn, :show, 123, my_param: "hi", phx_anchor: "yo")

# Option D
Routes.foo_path(@conn, :show, 123, my_param: "hi", "#": "yo")
```

I like option C for how relatively light-weight it feels (don't need to go back and add explicit brackets to a param list, or to add an empty param list if there are no params), but option A is perhaps more elegant in that it completely(?) removes any risk of param/anchor ambiguity or backwards incompatibility.

@chrismccord Are you open to any of these as a PR?

chrismccord:
I think best option today is to make your own function which accepts the arguments. I don't love the available options so a wrapping function is your best bet today :)

PR #3763, opened 2020-04-15
include `module` field in package.json

heavyk:
makes it easy to just `import { Socket } from 'phoenix'` when using rollup.


chrismccord:
Does this have any caveats for folks using webpack or otherwise that we should be aware of? // @snewcomer @Gazler

Gazler:
I don't think so, webpack also supports this. https://webpack.js.org/guides/author-libraries/#final-steps

chrismccord:
❤️❤️❤️🐥🔥

PR #3773, opened 2020-04-19
Fix broken link in Deploying with Heroku guide.

whilefalse:
I noticed the link out to the releases documentation was broken on the Deploying with Heroku guide. I'm presuming it was meant to link to the Phoenix releases docs, which just means changing from an absolute to a relative link.

Hopefully this is ok to merge. If so, do you know how I would go about fixing this guide for old versions of the documentation? I'm presuming they run from a tag, but I obviously can't update the release tag. This link seems to have been introduced in v1.4.12 of the docs, so the following would need updating:

- v1.4.12
- v1.4.13
- v1.4.14
- v1.4.15
- v1.4.16
- v1.5.0-rc.0

I'm happy to do this, I'm just not sure about the process given they are for released versions and the docs live with the code.

Thanks!

### DIFF
--- a/guides/deployment/heroku.md
+++ b/guides/deployment/heroku.md
@@ -352,7 +352,7 @@ build:
 
 ### Setup releases and Dockerfile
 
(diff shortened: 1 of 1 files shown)


josevalim:

PR #3776, opened 2020-04-20
Update default socket serializer to use V2

jjcarstens:
Since v1.4, the [socket documentation states that `Phoenix.Socket.V2.JSONSerializer` is used by default](https://github.com/phoenixframework/phoenix/blob/v1.4/lib/phoenix/socket.ex#L89-L92).

However, that is not the case because when the `vsn` parameter is excluded, it gets set to `"1.0.0"` making the default `V1.JSONSerializer`.

This changes that to be `"2.0.0"` when `vsn` is not explicitly provided in the connect params in order to match documentation and use `V2.JSONSerializer` by default.

### DIFF
--- a/lib/phoenix/socket.ex
+++ b/lib/phoenix/socket.ex
@@ -424,7 +424,7 @@ defmodule Phoenix.Socket do
       connect_info: connect_info
     } = map
 
-    vsn = params["vsn"] || "1.0.0"
+    vsn = params["vsn"] || "2.0.0"
 
     options = Keyword.merge(socket_options, options)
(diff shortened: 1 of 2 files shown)


chrismccord:
The client provides the version, so when we shipped V2, the phoenix.js client included the 2.0.0 VSN param. In this case, we must default to 1.0 for backwards compatibility for existing clients that cannot update. I understand the confusion about "default", but V2 is the default provided you your Phoenix is running the phoenix.js that shipped with it. Thanks!

PR #3781, opened 2020-04-22
Add scss to default applications using webpack

### LINKED ISSUE #3767: Add node-sass and sass-loader to new apps
We removed them because node-sass was slow and prone to failures on Windows but those issues have been fixed in recent versions. [Here are some reference steps](https://blog.isnorcreative.com/2019/02/14/using-sass-with-phoenix-1-4-and-webpack.html) but we don't need style loader (and likely don't need to keep the .css rules).

chrismccord:
Closes #3767


installer/templates/phx_assets/package.json
```diff
@@ -18,6 +18,9 @@
     "babel-loader": "^8.0.0",
     "copy-webpack-plugin": "^5.1.1",
     "css-loader": "^3.4.2",
+    "sass-loader": "^8.0.2",
+    "node-sass": "^4.13.1",
+    "style-loader": "^1.1.4",
```
josevalim, line 23: We don't need style loader?

```suggestion
```

cpjolicoeur:
Are you open to expanding this a bit to support `sass` extensions as well instead of just `scss`?

Maybe something like:

     test: /\.(?:sa|s?c)ss$/,

jalberto:
to be able to use both `sass` and `scss` will be great, as everyone can choose its favorite format.

PR #3784, opened 2020-04-22
Fix guides: plug code typo

miXwui:
`action_name` typo will result in:

```
CompileError
lib/hello_web/controllers/hello_controller.ex:4: cannot find or invoke local action_name/0 inside guard. Only macros can be invoked in a guard and they must be defined before their invocation. Called as: action_name()
```

Should be `action` as defined in https://hexdocs.pm/phoenix/Phoenix.Controller.html#module-guards


josevalim:

PR #3794, opened 2020-04-25
Fix minor typo in phoenix/guides/routing.md

tobischo:
Corrects `rasied` to `raised` in `If no route matches, no pipeline is invoked and a 404 error is raised.` of the pipeline description


Gazler:
Thanks <3

PR #3796, opened 2020-04-25
improve docker example in release guide

otijhuis:
- add python (needed for nodegyp rebuild, will display error otherwise)
- mkdir isn't needed, WORKDIR will create the directory if it doesn't
  exist
- combine commands where possible to reduce the amount of docker layers
- copy package*.json and install packages before doing the deploy so the
  install won't happen every time you modify a js/css file or other
  asset
- use npm ci instead of npm install
- bash not needed, sh/ash is available
- install ncurses-libs, the app wouldn't start otherwise
- do the chown when /app is still empty and add chown to the copy
  command, this saves docker from creating a large layer (about 10Mb)
  just for the permission changes
- add a default cmd
- use apk --no-cache so the cache doesn't get added to the image

I'd actually recommend using docker buildkit to improve performance/caching even more and use more explicit node versions and such but I didn't want to make the example too complicated.

Image size went down almost 20Mb with the changes implemented. You can also change assets now without it needing to reinstall npm packages every time.

### DIFF
--- a/guides/deployment/releases.md
+++ b/guides/deployment/releases.md
@@ -175,10 +175,9 @@ Here is an example Docker file to run at the root of your application covering a
 FROM elixir:1.9.0-alpine AS build
 
 # install build dependencies
-RUN apk add --update build-base npm git
+RUN apk add --no-cache build-base npm git python
 
 # prepare build dir
-RUN mkdir /app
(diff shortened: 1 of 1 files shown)


josevalim:

PR #3798, opened 2020-04-26
Minor improvement in secret generating

qcam:
I think we can omit padding since eventually they would be stripped out :).


josevalim:

PR #3799, opened 2020-04-26
Minor improvements around controller

qcam:
Would it fit to use `Plug.Conn.put_private/3` in these cases?

### DIFF
--- a/lib/phoenix/controller/pipeline.ex
+++ b/lib/phoenix/controller/pipeline.ex
@@ -19,15 +19,10 @@ defmodule Phoenix.Controller.Pipeline do
 
       @doc false
       def call(conn, action) when is_atom(action) do
-        conn =
-          update_in(
-            conn.private,
-            &(&1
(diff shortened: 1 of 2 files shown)


josevalim:
We don't do it to avoid setting the private and connection multiple times. We would need a better API in Plug to allow passing multiple keys at once.

qcam:
Would it make sense use `Plug.Conn.merge_private/2`? https://hexdocs.pm/plug/Plug.Conn.html#merge_private/2

josevalim:
Yes!!!! we need to check which Plug version it was added though and make sure the requirement is correct.

qcam:
I think we introduced that in Plug 1.5.1 (https://github.com/elixir-plug/plug/blob/v1.6/CHANGELOG.md#v151-2018-05-17), while `master`  uses Plug `~> 1.10`. I guess we're good to use `merge_private` then?

josevalim:
Yes!

PR #3800, opened 2020-04-26
Elaborate a bit more the intro to channels

fxn:
Channels are introduced in a way that could make people think broadcasts are automatic.

This patch proposes an intro that makes the programming model a bit more clear in my mind.

### DIFF
--- a/guides/realtime/channels.md
+++ b/guides/realtime/channels.md
@@ -14,8 +14,12 @@ Some possible use cases include:
 - Notifying a browser that a page's CSS or JavaScript has changed (this is handy in development)
 
 Conceptually, Channels are pretty simple.
-Clients connect and subscribe to one or more topics, whether that's `public_chat` or `updates:user1`.
-Any message sent on a topic, whether from the server or from a client, is sent to all clients subscribed to that topic (including the sender, if it's subscribed), like this:
+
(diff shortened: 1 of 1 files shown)


guides/realtime/channels.md
```diff
@@ -14,8 +14,12 @@ Some possible use cases include:
 - Notifying a browser that a page's CSS or JavaScript has changed (this is handy in development)
 
 Conceptually, Channels are pretty simple.
-Clients connect and subscribe to one or more topics, whether that's `public_chat` or `updates:user1`.
-Any message sent on a topic, whether from the server or from a client, is sent to all clients subscribed to that topic (including the sender, if it's subscribed), like this:
+
+First, clients connect to the server using some transport, like WebSocket. Once connected, they join one or more topics. For example, to interact with a public chat room clients may join a topic called `public_chat`, and to receive updates from a product with ID 7, they may need to join a topic called `product_updates:7`.
```
dlederle, line 18: Minor issue: the guide introduces the different transports a few paragraphs down. This sentence would be clearer if we stay focused on the channel abstraction instead of the underlying transport. Or, pull the introduction of transports earlier in the doc.

I like the example overall 👏

fxn: Software documentation is not always strictly ordered like you'd do in a math book. For didactic purposes.

In this case, "transport" is not meant to be `transport` as in the Phoenix concept, it just an informal way to say:

1. That you open a persistent connection.
2. That you can choose how to do it.
3. That WebSocket is your mental model.

We could replace the word "transport" with something more casual, say

> First, clients open a persitent connection to the server, normally a WebSocket.

I took a more risky compromise using "Channel servers" down below.

josevalim:

PR #3801, opened 2020-04-27
bump suggested node_version due to npm bug

petermm:
node 10.19.0 was bitten by this bug "invalid bin entry" - failing on npm install (sic!) https://github.com/npm/cli/issues/613

see also https://elixirforum.com/t/cant-deploy-phoenix-app-to-heroku/31023

💙

does this need to be "backported" to 1.5 branch to appear on site?


Gazler:
Thanks <3

This does need backported, I'll take care of that though.

PR #3802, opened 2020-04-27
Update phoenix_ecto version for umbrella app generator

jgelens:
Otherwise I ended up with a `CheckRepoStatus.init/1 is undefined` after a fresh install.


josevalim:

PR #3809, opened 2020-04-29
Guard ChannelTest functions

aaronrenner:
When using ChannelTest, one of my test cases accidentally passed in `nil` instead of a socket and I got an error from the internals of Phoenix.
```
** (UndefinedFunctionError) function nil.channel_pid/0 is undefined
    nil.channel_pid()
   (phoenix 1.5.0-rc.0) lib/phoenix/test/channel_test.ex:410: Phoenix.ChannelTest.push/3
```

With these changes it now raises a `FunctionClauseError` at the boundary of the public API and lets the user know they called the function incorrectly.

```
** (FunctionClauseError) no function clause matching in Phoenix.ChannelTest.push/3
      (phoenix 1.6.0-dev) lib/phoenix/test/channel_test.ex:407: Phoenix.ChannelTest.push(nil, "event", %{"payload" =>"foo"})
```

### DIFF
--- a/lib/phoenix/test/channel_test.ex
+++ b/lib/phoenix/test/channel_test.ex
@@ -404,7 +404,7 @@ defmodule Phoenix.ChannelTest do
 
   """
   @spec push(Socket.t, String.t, map()) :: reference()
-  def push(socket, event, payload \\ %{}) do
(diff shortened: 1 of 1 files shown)


josevalim:

PR #3815, opened 2020-04-30
Extract user agent from socket connect_info

wojtekmach:
cc @josevalim


wojtekmach:
If we ever have a 3rd header we want to extract we probably should instead have a mechanism to grab arbitrary request headers, e.g.: `connect_info: [req_headers: ["user-agent", ~/^x-/]]` or something.

wojtekmach:
or: `connect_info: [req_headers: & &1 == "user-agent" or String.starts_with?(&1, "x-")]`

josevalim:

PR #3820, opened 2020-05-01
Fix formatting of generated *_live_test.exs

aaronrenner:
When running the formatter after `phx.gen.live`, it corrects the following lines in `test/my_app_web/live/post_live_test.exs`

Screen Shot 2020-05-01 at 2 16 08 PM

This updates the generator so its code will pass the formatter.


josevalim:

PR #3821, opened 2020-05-03
allow custom error response from socket handler

stephanos:
This is a half-done PR to illustrate my point: It'd be great if you could return a custom error from the `connect` function of the WebSocket handler.

Right now when you return `:error` from it, it always returns a 403 status code. I'm not sure why, because it could be possible that you want to return a 400 or 429, for example, in certain situations. And adding a body to explain the error is missing, too, in my opinion.

For example, I want to return a 400 with a detailed error message since my endpoint requires a specific header to be set.

If you're okay with this in general, I can add a proper test and change the docs, too.

### DIFF
--- a/lib/phoenix/socket.ex
+++ b/lib/phoenix/socket.ex
@@ -203,8 +203,8 @@ defmodule Phoenix.Socket do
   See `Phoenix.Token` documentation for examples in
   performing token verification on connect.
   """
-  @callback connect(params :: map, Socket.t) :: {:ok, Socket.t} | {:error, integer, [{binary, binary}], String.t} | :error
-  @callback connect(params :: map, Socket.t, connect_info :: map) :: {:ok, Socket.t} | {:error, integer, [{binary, binary}], String.t} | :error
+  @callback connect(params :: map, Socket.t) :: {:ok, Socket.t} | {:error, term}
+  @callback connect(params :: map, Socket.t, connect_info :: map) :: {:ok, Socket.t} | {:error, term}
 
   @doc ~S"""
   Identifies the socket connection.
@@ -450,12 +450,11 @@ defmodule Phoenix.Socket do
         result
 
       :error ->
-        :error
+        {:error, :undefined}
     end
   end
 
   defp result({:ok, _}), do: :ok
-  defp result(:error), do: :error
   defp result({:error, _}), do: :error
 
   def __init__({state, %{id: id, endpoint: endpoint} = socket}) do
@@ -587,20 +586,20 @@ defmodule Phoenix.Socket do
           invalid ->
             Logger.error "#{inspect handler}.id/1 returned invalid identifier " <>
                            "#{inspect invalid}. Expected nil or a string."
-            :error
+            {:error, :undefined}
         end
 
       {:error, _reason} = err ->
         err
 
       :error ->
-        :error
+        {:error, :undefined}
 
       invalid ->
         connect_arity = if function_exported?(handler, :connect, 3), do: "connect/3", else: "connect/2"
         Logger.error "#{inspect handler}. #{connect_arity} returned invalid value #{inspect invalid}. " <>
                      "Expected {:ok, socket} or :error"
-        :error
+        {:error, :undefined}
     end
   end
 
--- a/lib/phoenix/socket.ex
+++ b/lib/phoenix/socket.ex
@@ -198,7 +198,10 @@ defmodule Phoenix.Socket do
 
       {:ok, assign(socket, :user_id, verified_user_id)}
 
-  To deny connection, return `:error`.
+  To deny connection, return `:error`. An empty 403 response will be
+  returned since browsers do not allow accessing the server response
+  due to security reasons. However, if you want to send a custom error response
+  to non-browser clients, return a {:error, status_code, headers, body} tuple.
 
   See `Phoenix.Token` documentation for examples in
   performing token verification on connect.
--- a/lib/phoenix/socket.ex
+++ b/lib/phoenix/socket.ex
@@ -457,7 +457,7 @@ defmodule Phoenix.Socket do
 
   defp result({:ok, _}), do: :ok
   defp result(:error), do: :error
-  defp result({:error, _, _, _}), do: :error
+  defp result({:error, _}), do: :error
 
   def __init__({state, %{id: id, endpoint: endpoint} = socket}) do
     _ = id && endpoint.subscribe(id, link: true)
@@ -590,6 +590,9 @@ defmodule Phoenix.Socket do
                            "#{inspect invalid}. Expected nil or a string."
             :error
         end
+
+      {:error, _reason} = err ->
+        err
 
       :error ->
         :error
--- a/lib/phoenix/socket.ex
+++ b/lib/phoenix/socket.ex
@@ -198,10 +198,8 @@ defmodule Phoenix.Socket do
 
       {:ok, assign(socket, :user_id, verified_user_id)}
 
-  To deny connection, return `:error`. An empty 403 response will be
-  returned since browsers do not allow accessing the server response
-  due to security reasons. However, if you want to send a custom error response
-  to non-browser clients, return a {:error, status_code, headers, body} tuple.
+  To deny connection, return `:error` or `{:error, reason}`.
+  TBD
 
   See `Phoenix.Token` documentation for examples in
   performing token verification on connect.
@@ -593,16 +591,13 @@ defmodule Phoenix.Socket do
             :error
         end
 
-      {:error, status_code, headers, body} = err when is_number(status_code) and is_list(headers) and is_binary(body) ->
-        err
-
       :error ->
         :error
 
       invalid ->
         connect_arity = if function_exported?(handler, :connect, 3), do: "connect/3", else: "connect/2"
         Logger.error "#{inspect handler}. #{connect_arity} returned invalid value #{inspect invalid}. " <>
-                     "Expected {:ok, socket}, {:error, status_code, headers, body} or :error"
+                     "Expected {:ok, socket} or :error"
         :error
     end
   end
--- a/lib/phoenix/socket.ex
+++ b/lib/phoenix/socket.ex
@@ -203,8 +203,8 @@ defmodule Phoenix.Socket do
   See `Phoenix.Token` documentation for examples in
   performing token verification on connect.
   """
-  @callback connect(params :: map, Socket.t) :: {:ok, Socket.t} | {:error, term}
-  @callback connect(params :: map, Socket.t, connect_info :: map) :: {:ok, Socket.t} | {:error, term}
+  @callback connect(params :: map, Socket.t) :: {:ok, Socket.t} | {:error, term} | :error
+  @callback connect(params :: map, Socket.t, connect_info :: map) :: {:ok, Socket.t} | {:error, term} | :error
 
   @doc ~S"""
   Identifies the socket connection.
--- a/lib/phoenix/socket.ex
+++ b/lib/phoenix/socket.ex
@@ -198,8 +198,7 @@ defmodule Phoenix.Socket do
 
       {:ok, assign(socket, :user_id, verified_user_id)}
 
-  To deny connection, return `:error` or `{:error, reason}`.
-  TBD
+  To deny connection, return `:error`.
 
   See `Phoenix.Token` documentation for examples in
   performing token verification on connect.
--- a/lib/phoenix/socket.ex
+++ b/lib/phoenix/socket.ex
@@ -450,11 +450,12 @@ defmodule Phoenix.Socket do
         result
 
       :error ->
-        {:error, :undefined}
+        :error
     end
   end
 
   defp result({:ok, _}), do: :ok
+  defp result(:error), do: :error
   defp result({:error, _}), do: :error
 
   def __init__({state, %{id: id, endpoint: endpoint} = socket}) do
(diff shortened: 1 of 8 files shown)


josevalim:
How are they connecting to the endpoint? The reason why this has been limited is because we can't read those reasons from the browser anyway... unfortunately. :(

stephanos:
I see. I'm not connecting to it from the browser. I'm connecting from a Python-based application that can read the status code from the server.

I guess it depends on whether Phoenix is supposed to be "web-only" or if it can be used for APIs as well. If it's the former, I agree, no need to add support for this. If it's the latter, I don't see why Phoenix would want to prevent you from doing that. Because right now it's impossible (without writing your own `websocket` transport).

But even so, if the browser can't read it, what's the harm in allowing the user to pick the status code?

lib/phoenix/transports/websocket.ex
```diff
@@ -32,6 +32,7 @@ defmodule Phoenix.Transports.WebSocket do
         case handler.connect(config) do
           {:ok, state} -> {:ok, conn, state}
           :error -> {:error, Plug.Conn.send_resp(conn, 403, "")}
+          {:error, status, body} -> {:error, Plug.Conn.send_resp(conn, status, body)}
```
josevalim, line 35: If we are going to allow all of this, then we probably should do `{:error, status, headers, body}`.

josevalim:
The direction looks good. We need docs and tests besides what was requested above. Thank you!

stephanos:
Cool, I'll flesh it out a bit more!

stephanos:
Okay, I added a test and documentation.

lib/phoenix/transports/websocket.ex
```diff
@@ -31,12 +31,19 @@ defmodule Phoenix.Transports.WebSocket do
 
         case handler.connect(config) do
           {:ok, state} -> {:ok, conn, state}
-          :error -> {:error, Plug.Conn.send_resp(conn, 403, "")}
+          :error -> {:error, resp(conn, 403, [], "")}
+          {:error, status, headers, body} -> {:error, resp(conn, status, headers, body)}
```
josevalim, line 35: I believe you also need to change the long polling transport.

josevalim: Actually, if we remeber that sockets and channels are meant to be transport agnostic, this approach introduces an issue because we are adding HTTP semantics to the socket. So if you have a non-HTTP transport, the HTTP response won't be useless. Maybe the best option is to move this to transport but I have no idea how to do so at the moment.

stephanos: Oh! Just to clarify, is the problem the `headers` bit? Because as you can see, right now it sends `Plug.Conn.send_resp(conn, 400, "")` which is already using the HTTP status code and body. Just want to understand the problem better :)

stephanos: FYI I've added a test for the long polling transport - didn't have to change any code to make it work.

stephanos: Nevermind, I think I understand the problem now. The `Socket` is supposed to be transport agnostic but by passing the error details through it, that's no longer the case. So if we find a way to move the HTTP-specific error-handling _into_ the Websocket module, we'd fix that.

stephanos: After tinkering with it for a bit I see two options:

(a) adding an option
```
websocket: [
    connection_error_handler: CustomConnectionErrorHandler,
```
to provide your own error handling 

(b) allowing users to provide their own `websocket` transport implementation

```
websocket: [
    implementation: MyWebsocketTransport
```

What do you think

josevalim: Yes, I think something like: `websocket: [error_handler: {module, function, args}]` that is called with the `conn` and must return a `conn` can work. WDYT?

stephanos: I'll make that change 👍

stephanos:
I published my current progress. I'm not sure I'm super happy with it.

The problem that I'm seeing is that it's hard to make this a small change: I started applying this error handler to the LongPolling transport, and it seems to break down there since it expects very specific response codes and bodies. Whereas the WebSocket is very generic and sparse. To incorporate the custom error handler in the LongPolling module would require even more changes I think.

Overall, I wonder if it's easier to just allow a custom WebSocket implementation? As far as I can see, right now it's hard-coded in two places:

cowboy2_handler.ex
```
try do
      case endpoint.__handler__(conn, opts) do
        {:websocket, conn, handler, opts} ->
          case Phoenix.Transports.WebSocket.connect(conn, endpoint, handler, opts) do
``` 

endpoint.ex
```
if websocket do
        config = Phoenix.Socket.Transport.load_config(websocket, Phoenix.Transports.WebSocket)
```

I believe both places have access to the endpoint options, so it'd be possible to just provide your own implementation. That would be a fairly minimal change, most likely, and allow all everyone to customize a transport, if so desired.

Seems like a very flexible and easy option to me, what do you think?

lib/phoenix/socket.ex
```diff
@@ -589,6 +590,9 @@ defmodule Phoenix.Socket do
             :error
         end
 
+      {:error, _reason} = err ->
+        err
+
       :error ->
         :error
```
josevalim, line 597: Let's rewrite this to `{:error, :undefined}`.

lib/phoenix/socket.ex
```diff
@@ -456,6 +456,7 @@ defmodule Phoenix.Socket do
 
   defp result({:ok, _}), do: :ok
   defp result(:error), do: :error
```
josevalim, line 458: We should be able to remove this clause if we rewrite `:error` to `{:error, :undefined}`.

lib/phoenix/socket.ex
```diff
@@ -203,8 +203,8 @@ defmodule Phoenix.Socket do
   See `Phoenix.Token` documentation for examples in
   performing token verification on connect.
   """
-  @callback connect(params :: map, Socket.t) :: {:ok, Socket.t} | :error
-  @callback connect(params :: map, Socket.t, connect_info :: map) :: {:ok, Socket.t} | :error
+  @callback connect(params :: map, Socket.t) :: {:ok, Socket.t} | {:error, integer, [{binary, binary}], String.t} | :error
+  @callback connect(params :: map, Socket.t, connect_info :: map) :: {:ok, Socket.t} | {:error, integer, [{binary, binary}], String.t} | :error
```
josevalim, line 207: We should update those two. :)

lib/phoenix/transports/websocket.ex
```diff
@@ -2,10 +2,20 @@ defmodule Phoenix.Transports.WebSocket do
...
+    def handle(conn, _error) do
+      {:error, Plug.Conn.send_resp(conn, 403, "")}
+    end
+  end
+
   def default_config() do
     [
       path: "/websocket",
       serializer: [{V1.JSONSerializer, "~> 1.0.0"}, {V2.JSONSerializer, "~> 2.0.0"}],
+      connection_error_handler: &__MODULE__.ConnectionErrorHandler.handle/2,
```
josevalim, line 18: We can simply call it `:error_handler`. :)

lib/phoenix/transports/websocket.ex
```diff
@@ -2,10 +2,20 @@ defmodule Phoenix.Transports.WebSocket do
   @moduledoc false
   alias Phoenix.Socket.{V1, V2, Transport}
 
+  defmodule ConnectionErrorHandler do
+    def handle(conn, {:error, :wrong_method}) do
```
josevalim, line 6: Let's call the handler only with the `reason` instead of `{:error, reason}`.

lib/phoenix/transports/websocket.ex
josevalim: Also, this should be a MFA, something like `{__MODULE__, :handle, []}` which you are going to invoke like this:

    {m, f, args} = config.error_handler
    apply(m, f, [conn, reason | args])

The reason we cannot use anonymous functions is because the error_handler is defined at the module level and anonymous functions do not "survive" after the module is compiled, unless they are in the `&m.f/arity` form. So it is best to use the tuple approach.

lib/phoenix/socket.ex
stephanos: I'm a little worried about this change since it seems to break backward-compatibility? I don't know if people rely on this returning just `:error`.

lib/phoenix/transports/websocket.ex
stephanos: neat, I didn't know that!

lib/phoenix/socket.ex
josevalim: To be clear, we will do this:

```
:error ->
  {:error, :undefined}
```

So internally, the `error_handler` always has a reason. The socket users themselves can still return `:error`.

lib/phoenix/socket.ex
```diff
@@ -203,8 +203,8 @@ defmodule Phoenix.Socket do
   See `Phoenix.Token` documentation for examples in
   performing token verification on connect.
   """
-  @callback connect(params :: map, Socket.t) :: {:ok, Socket.t} | :error
-  @callback connect(params :: map, Socket.t, connect_info :: map) :: {:ok, Socket.t} | :error
+  @callback connect(params :: map, Socket.t) :: {:ok, Socket.t} | {:error, term}
+  @callback connect(params :: map, Socket.t, connect_info :: map) :: {:ok, Socket.t} | {:error, term}
```
josevalim, line 207:
```suggestion
  @callback connect(params :: map, Socket.t) :: {:ok, Socket.t} | {:error, term} | :error
  @callback connect(params :: map, Socket.t, connect_info :: map) :: {:ok, Socket.t} | {:error, term} | :error
```

lib/phoenix/socket.ex
josevalim: Actually, no. You are correct, this would be a backwards incompatible change for people implementing custom transports. Gah, I am so sorry.

I think we will have to revert the `:error -> {:error, :undefined}` changes.

Here is my (hopefully) last proposal:

1. Allow `{:error, reason}`
2. The custom error handler will only be called if `{:error, reason}` is returned

lib/phoenix/transports/websocket.ex
```diff
@@ -2,10 +2,16 @@ defmodule Phoenix.Transports.WebSocket do
   @moduledoc false
   alias Phoenix.Socket.{V1, V2, Transport}
 
+  defmodule ErrorHandler do
+    def handle(conn, :wrong_method), do: Plug.Conn.send_resp(conn, 400, "")
+    def handle(conn, _reason), do: Plug.Conn.send_resp(conn, 403, "")
+  end
```
josevalim, line 8: No need to define a separate module for it, just define a `error_handler` function directly in WebSocket. :)

josevalim:
Hi @stephanos, thanks for the latest changes. I think we are almost there. Just to make sure nothing was lost, here are the pending changes:

I think we will have to revert the `:error -> {:error, :undefined}` changes (sorry). Here is my (hopefully) last proposal:

* Allow `{:error, reason}` but without rewriting `:error`
* The custom error handler will only be called if `{:error, reason}` is returned

WDYT?

themitigater:
+1
Thanks, Stephanos and Jose.

lib/phoenix/socket.ex
stephanos: Only saw this now. No worries! Made the revert, took me 5 minutes :)

stephanos:
@josevalim that works.

The tradeoff for only calling the error handler for `{:error, reason}`, though, is that you won't be able to customize the existing error behavior (ie the 400/403 status codes right now). For me that's fine, but I wanted to emphasize that. It might seem a little inconsistent?

josevalim:
> The tradeoff for only calling the error handler for {:error, reason}, though, is that you won't be able to customize the existing error behavior (ie the 400/403 status codes right now). For me that's fine, but I wanted to emphasize that. It might seem a little inconsistent?

I think this is fine as long as the only `:error` we may receive is the one returned by the socket. This way, you can full control by simply not returning `:error`. :)

josevalim:
Awesome! I think only two changes left:

1. docs for the `Endpoint.socket/3` that mentions the new option with some examples
2. Match on `{:error, reason}` and provide an error handler for long polling

josevalim:
Wait, we had error handling on long-polling before. Did we remove it because it did not work as we expected? In any case, if we don't add `error_handler` to long poll, we need to make sure it will convert `{:error, _}` to `:error`.

stephanos:
>I think this is fine as long as the only :error we may receive is the one returned by the socket. This way, you can full control by simply not returning :error. :)

Yes, that's true with the only exception being 
```
  def connect(conn, _, _, _) do
    {:error, Plug.Conn.send_resp(conn, 400, "")}
  end
```
which is hard-coded in the Websocket module and you don't have any control over that. But I guess that's fine? Just wanted to point out that caveat.

> docs for the Endpoint.socket/3 that mentions the new option with some examples

Done. Let me know if this is good enough.

> Did we remove it because it did not work as we expected?

Yes, I think at this point I didn't feel comfortable with that since long polling has a certain contract with the browser that I didn't want to break. It's also less likely to be used outside of the browser context, IMHO, than Websockets. I've added a test to ensure that the error tuple works.

josevalim:
> which is hard-coded in the Websocket module and you don't have any control over that. But I guess that's fine? 

I am fine with converting this into a `error_handler` call with the atom of your choice. :) I am also fine with leaving it as is for now.

test/phoenix/integration/long_poll_channels_test.exs
```diff
@@ -88,7 +88,11 @@ defmodule Phoenix.Integration.LongPollChannelsTest do
     channel "room:*", RoomChannel
 
     def connect(%{"reject" => "true"}, _socket) do
-      :error
+      :error_logger_tty_h
```
josevalim, line 91:
```suggestion
      :error
```

josevalim:

stephanos:
@josevalim thank you for your guidance and patience :)

PR #3823, opened 2020-05-05
Fix Cowboy.Drainer link

sb8244:
This is such a great addition to Phoenix defaults!

One side question: Is there a best practice on draining keep-alive connections now? I have done some hacky things in the past like set an app config and then check it in the controller, but I'm wondering if there's a better solution.


chrismccord:
❤️❤️❤️🐥🔥

PR #3824, opened 2020-05-05
fix bufferedAmount check in js

h3xby:
on `socket.disconnect()` `waitForBufferDone` always retries 5 times even with empty buffer because when `bufferedAmount === 0` then `if (this.conn.bufferedAmount)` evaluates to false

if I understood correctly, `!bufferedAmount` will be enough - for longpoll there is no buffering, so nothing to wait; for websocket - check for `0`

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -955,7 +955,7 @@ export class Socket {
   }
 
   waitForBufferDone(callback, tries = 1) {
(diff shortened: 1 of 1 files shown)


snewcomer:
Hi! 👋 Is there a fix to something you saw or just syntax improvements?

h3xby:
аs i write in description, on `socket.disconnect()` `waitForBufferDone` always retries 5 times, even with empty buffer it will wait around 1.5s before closing connection

PR #3825, opened 2020-05-06
Document self() is the channel server in join/3

fxn:
## This patch

The documentation of channels does not reveal the underlying implementation (which is good).

Therefore, if you want to monitor the channel, as recommended in the docs of `terminate/2`, the user needs to be explicitly told which PID to monitor, and when it is a good idea to spawn it.

To realize this is not obvious without being documented, compare with the situation in the socket `connect/3` callback. Similar entry point from a user's perspective, but `self()` there is not the one to monitor if you want to keep track of the socket itself.

## Feedback

Since we are here, let me also comment that as a user the perils of `terminate/2` + trapping exits are not clear from the docs. You wonder. "Why is it discouraged? If you cannot trust `terminate/2`, why does the callback exist in the first place?"

From the point of view of the user, a channel is not an arbitrary/generic GenServer, from the point of view of the user, a channel is a process of some sort managed by Phoenix. Why documenting `terminate/2` to say that you cannot rely on it? If the solution is a monitor, why doesn't Phoenix implement a monitor for me and provide a robust callback?

Furthermore, I have stopped channels in several ways, taking things down, sending `kill` signals, killing sockets, ..., I have not been able to terminate a channel without the callback being called.

Of course, the docs are correct. My point is that since a channel is managed by the framework, as a user I'd appreciate something less vague like "even if you trap exists, if this concrete thing happens, the callback won't be invoked". That way, you can evaluate the risk. Also, some legit use cases for the callback would answer the question "why is it in the behaviour if you cannot trust it?"

Unfortunately, I do not know enough to volunteer a patch, so leaving only a comment.

### DIFF
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -167,11 +167,15 @@ defmodule Phoenix.Channel do
 
   `terminate/2`, however, won't be invoked in case of errors nor in
   case of exits. This is the same behaviour as you find in Elixir
-  abstractions like `GenServer` and others. Typically speaking, if you
-  want to clean something up, it is better to monitor your channel
-  process and do the clean up from another process.  Similar to GenServer,
+  abstractions like `GenServer` and others. Similar to `GenServer`,
   it would also be possible `:trap_exit` to guarantee that `terminate/2`
   is invoked. This practice is not encouraged though.
+
+  Typically speaking, if you want to clean something up, it is better to
+  monitor your channel process and do the clean up from another process.
+  All channel callbacks including `join/3` are called from within the
+  channel process. Therefore, `self()` in any of them returns the PID to
+  be monitored.
 
   ## Exit reasons when stopping a channel
 


lib/phoenix/channel.ex
```diff
@@ -169,9 +169,13 @@ defmodule Phoenix.Channel do
   case of exits. This is the same behaviour as you find in Elixir
   abstractions like `GenServer` and others. Typically speaking, if you
   want to clean something up, it is better to monitor your channel
-  process and do the clean up from another process.  Similar to GenServer,
-  it would also be possible `:trap_exit` to guarantee that `terminate/2`
-  is invoked. This practice is not encouraged though.
+  process and do the clean up from another process. The PID returned
+  by `self()` in `join/3` is the one you want to monitor, you'd
+  normally spawn it in that callback.
```
josevalim, line 174: "you'd normally spawn it in that callback." this line not clear to me. Can you please expand what you mean? :)

fxn: Ah! I mean `join/3`, which was mentioned, maybe word it differently?

fxn: Oh, maybe "it" is also confusing. The intention is to say: "You'd normally spawn the monitor in `join/3`". Let's rewrite it :).

josevalim: Ah, the misunderstanding is rather with what is being spawned. You mean to spawn a process to monitor it and clean it up. You can also have a central process.

fxn: Gotcha. To talk with code, the use case I have in mind is something like this (written on the fly here):

```elixir
def join(...) do
  ...
  monitor(socket)
end

def monitor(socket) do
  channel_pid = self()
  spawn(fn ->
    Process.monitor(channel_pid)
    receive do
      {:DOWN, _ref, :process, _monitored_pid, _reason} ->
        on_leave(socket)
    end
  end)
end
```

Where `on_leave()` is a custom function in which yoou log the leave of the user (whose ID is in the socket), and do something for metrics, telemetry, or Prometheus directly, you know.

So perhaps not suggest a particular approach, and leave that detail to the user? Document only the PID?

Which would be the requivalent of that snippet with a central process BTW?

fxn: I have the same for sockets, but in that case I need a hack to get the transport PID:

```elixir
defoverridable init: 1

@impl true
def init(state) do
  res = {:ok, {_, socket}} = super(state)
  monitor(socket)
  res
end
```

So, in this case, probably that cannot be documented if that hack is using private API.

josevalim: The transport id in the socket is `self()`, no?

The centralized process would be sending a message to a locally registered process with `self()`. Then the process monitors it and call the relevant code when it crashes.

fxn: > The transport id in the socket is self(), no?

Yes. The transport ID is not `self()` in `connect/3` (you know that, just explaining for the archives), so I need to decorate `init/1`. I believe there is no public interface to get it when the connection was established.

Once you are in `init/1`, chosing `self()` or `socket.transport_pid` is a matter of API usage or respecting the public contract (because the docs do not tell you they are the same).

josevalim: Wait, I actually thought that `self()` in `connect/3` was the transport id. :D I guess that `connect/3` is called before we start the transport.

Btw, why do you need to monitor both transport and channel?

fxn: > The centralized process would be sending a message to a locally registered process with self(). Then the process monitors it and call the relevant code when it crashes.

In the approach of my snippet, something I like is that the function the monitor runs is isolated in a process. If it does anything weird, I do not need to be defensive about the monitor crashing. However, my skills are very rough, so perhaps there's also a way as simple as that to make a centralized process to be robust.

josevalim: Yup, your approach is totally fine too!

fxn: Oh, if I spawn when `:DOWN` is received, is the same. And that way the app does not have twice as much processes as joined channels constantly, it only spawns to run the callback, which statistically it is going to be way less simulatenous processes.

fxn: > Wait, I actually thought that self() in connect/3 was the transport id. :D I guess that connect/3 is called before we start the transport.

@chrismccord told me the reason is that he prefers not to spawn a process until `connect/3` has accepted the connection. So Phoenix needs the return value of the callback in order to decide.

> Btw, why do you need to monitor both transport and channel?

You are going to laugh now :D.

In my app, I only need to monitor socket. I want socket-specific things, like logging disconnections by user, or having a gauge in Prometheus of opened sockets.

I do that at the socket level because that is the actual thing I am interested in. One socket may spawn other 27 channels, so what belongs naturally in the socket, stays attached to the socket.

OK, why I use this approach for sockets and do not provide a doc patch? Because it is a hack using private APIs and I would not know what to propose.

So, why do I provide a doc patch for channels if I am not using this? Because I noticed the lack of public contract about `self()`, so I fired the editor to contribute because you know how I am :D.

fxn: Oh! I also want to register for how long a user was connected, which is something `:telemetry` does not provide unless I missed it. I am going to assume that the monitor is fired in a timely manner so that the metric won't be exact to the nanosecond, but won't be 5 seconds off if the server is not under a heavy load (this was a practical advice from @michalmuskala).

You see, socket-specific data.

Then channels may have their own specific data too. During the lifetime of a connection people may join and leave and arbitrary number of channels (including none), so what's specifically about the socket, is naturally attached to socket code and processes.

josevalim: I see. There are some things:

1. We can document what `self()` is, but then we are documenting how Elixir behaves, not Phoenix. `self()` is always the current process. It is a matter of knowing what the current process is. :)

2. Even if we document the channel, you are telling me that using the socket is usually better, so this may not have much purpose. :)

Your `init` code could possibly be written like this:

```elixir
def init(arg) do
  monitor(self())
  super(arg)
end
```

`init/1` is public API, so this is OK-ish - but indeed I am not sure how much we want to encourage this. :)

fxn: > We can document what self() is, but then we are documenting how Elixir behaves, not Phoenix. self() is always the current process. It is a matter of knowing what the current process is. :)

Let me explain how I see it :). What Phoenix needs to tell you is that "the channel process that is going to die when the channel terminates is `self()`". As a user, how does the channel process communicates with your `self()` is something you do not control. That is opaque. You have a callback to do what I want to do, which is `terminate/2`. See? That is public contract. That callback is called in turn by the actual, real GenServer callback, but that is private. Your public interface is `terminate/2` and then you don't need a PID.

But as I said in the feedback, everybody tells you to not use `terminate/2` + traps exits (which by now I am cargo culting), so I need a new contract that guarantees which PID to monitor without revealing the implementation.

Maybe this is all implicit for a seasoned Elixir programmed that perhaps may think something like "there's no other sane/idiomatic way this can be implemented".

But then, looks at what happens with sockets! That is the counterexample. Without looking at the source code, your interface is what the documentation says.

> Even if we document the channel, you are telling me that using the socket is usually better, so this may not have much purpose. :)

No, no! I am saying that in my particular case I am using only a monitor for sockets. But of course there are use cases for channels. Indeed, the first time I saw this approach was in a [SO answer](https://stackoverflow.com/a/33941469/8362448) of @chrismccord for channels.

fxn: Alternatively, Phoenix would provide a callback `terminate_i_mean_for_real/2`, for sockets, and for channels. That would be the perfect and easy public interface.

josevalim: So would this be a better description? "`join` and  `terminate` are both called from within the channel process. Therefore, `self()` in both callbacks point to the same PID." Then we could expand with examples of the spawn monitor. 

From my understanding, your concerns about self() is related to the process life-cycle. So saying they are both part of the same process life-cycle may help build the necessary mental modal?

fxn: That is perfect 😀.

fxn:
@josevalim I pushed another iteration.

Here, I group together the comments about `GenServer` first, and move the ones about monitors to its own paragraph.

Then, the docs say `self()` is the channel server in all callbacks because while I am thinking logging and metrics (which naturally monitor starting at `join/3`), I guess there are cleanup use cases in which you monitor if at some point something happens, you receive a particular message, perform some action, ... So the wording is a bit more open.

What do you think about this one?

josevalim:

PR #3827, opened 2020-05-08
add example for channel.push

heavyk:
it wasn't immediately obvious reading the documentation that I could receive replies from phoenix push events. I added an example

EDIT: wait, it's not working like that now for me.. I must have done something. I'll update when I figure out why.. :)

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -512,7 +512,7 @@ export class Channel {
    * channel.push("event")
    *   .receive("ok", payload => { console.log("phoenix replied:", payload) })
    *   .receive("error", err => { console.log("phoenix errored", err) })
-   *   .receive("timeout", () => { console.log("timed out pushing") })
+   *   .receive("timeout", () => console.log("timed out pushing"))
    * @param {string} event
    * @param {Object} payload
    * @param {number} [timeout]
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -503,6 +503,16 @@ export class Channel {
   canPush(){ return this.socket.isConnected() && this.isJoined() }
 
   /**
+   * Sends a message `event` to phoenix with the payload `payload`.
+   * Phoenix receives this in the `handle_in(event, payload, socket)`
+   * function. if phoenix replies or it times out (default 10000ms),
+   * then optionally the reply can be received.
+   *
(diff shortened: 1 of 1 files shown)


heavyk:
that form of doing things worked for a bit, then it stopped working. I looked into it and it seems that in my code, I'm no longer getting the "ok" event right away, and instead I'm getting a "timeout" event, then the "ok" event. so, besides figuring out why I'm timing out so quickly (it's locahost), for whatever reason after timing out, it appears that no longer will I receive the ok because after timing out, it calls `reset` which calls `cancelRefEvent` which does `this.channel.off(this.refEvent)` which I think is what's preventing me from receiving the "ok" event after a timeout.

gonna take a walk. I'll see if my suspicion is right, and make another PR when I get back, if that is indeed the problem. cheers

assets/js/phoenix.js
```diff
@@ -503,6 +503,16 @@ export class Channel {
...
   /**
+   * Sends a message `event` to phoenix with the payload `payload`.
+   * Phoenix receives this in the `handle_in(event, payload, socket)`
+   * function. if phoenix replies or it times out (default 10000ms),
+   * then optionally the reply can be received.
+   *
+   * @example
+   * channel.push("event")
+   *   .receive("ok", payload => { console.log("phoenix replied:", payload) })
+   *   .receive("error", err => { console.log("phoenix errored", err) })
```
chrismccord, line 514:
```suggestion
   *   .receive("error", err => console.log("phoenix errored", err))
```

assets/js/phoenix.js
```diff
@@ -503,6 +503,16 @@ export class Channel {
...
 
   /**
+   * Sends a message `event` to phoenix with the payload `payload`.
+   * Phoenix receives this in the `handle_in(event, payload, socket)`
+   * function. if phoenix replies or it times out (default 10000ms),
+   * then optionally the reply can be received.
+   *
+   * @example
+   * channel.push("event")
+   *   .receive("ok", payload => { console.log("phoenix replied:", payload) })
```
chrismccord, line 513:
```suggestion
   *   .receive("ok", payload => console.log("phoenix replied:", payload))
```

chrismccord:
❤️❤️❤️🐥🔥

PR #3829, opened 2020-05-08
Add fixtures module to generators

### LINKED ISSUE #3797: Generate a separate module for fixtures on phx.gen.* scaffolds
Today, Phoenix generators include a fixture function into every context and controller/live test file. The issue with this approach is that we often need to share resources between tests, so we end-up forced to move them elsewhere.

My proposal is to generate those fixtures at `test/supoprt/fixtures/CONTEXT_fixtures.ex`. So for example, `mix phx.gen.html Accounts User users` will generate a `MyApp.AccountsFixtures` module which is imported into the tests instead. This is what what is used by `mix phx.gen.auth`.

aaronrenner:
This adds a fixtures module to the output of `phx.gen.context` and uses the module throughout `phx.gen.live`, `phx.gen.html` and phx.gen.json`. The module looks like this:

```elixir
defmodule DemoLiveApp.BlogFixtures do
  @moduledoc """
  This module defines test helpers for creating
  entities via the `DemoLiveApp.Blog` context.
  """

  def post_fixture(attrs \\ %{}) do
    {:ok, post} =
      attrs
      |> Enum.into(%{
        content: "some content",
        rating: 42,
        title: "some title"
      })
      |> DemoLiveApp.Blog.create_post()

    post
  end
end
```

This change originally came about from #3797 and has been discussed further here: https://github.com/aaronrenner/phoenix_generator_proposals_live/pull/1.

Any feedback or suggestions would be appreciated!

Fixes #3797

### DIFF
--- a/lib/mix/phoenix/context.ex
+++ b/lib/mix/phoenix/context.ex
@@ -12,6 +12,7 @@ defmodule Mix.Phoenix.Context do
             basename: nil,
             file: nil,
             test_file: nil,
+            test_fixtures_file: nil,
             dir: nil,
             generate?: true,
             context_app: nil,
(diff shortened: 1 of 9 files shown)


josevalim:

PR #3833, opened 2020-05-11
Fix syntax highlighting for a controller test example in documentation

merwan:
Related documentation section: https://hexdocs.pm/phoenix/testing_controllers.html#the-index-action

- Use inline code instead of double quotes to emphasize path to file
- Fix typo in language identifier for the syntax highlighting


chrismccord:
❤️❤️❤️🐥🔥

PR #3834, opened 2020-05-11
Update Phoenix.Logger docs

qcam:


chrismccord:
This doesn't appear to be something we emit from telemetry. What are you seeing `result` come from? Thanks!

lib/phoenix/logger.ex
```diff
@@ -69,12 +69,12 @@ defmodule Phoenix.Logger do
 
     * `[:phoenix, :channel_joined]` - dispatched at the end of a channel join
       * Measurement: `%{duration: native_time}`
-      * Metadata: `%{params: term, socket: Phoenix.Socket.t}`
+      * Metadata: `%{result: :ok | :error, params: term, socket: Phoenix.Socket.t}`
```
qcam, line 72: I think we emit `:result` at https://github.com/phoenixframework/phoenix/blob/ac35fba991d550494c9b788a9d57bbf5679a3dcf/lib/phoenix/channel/server.ex#L301.

qcam: I was wrong about the other 🙈. It does not seem to emit `:result`.

chrismccord: and I only checked the other 🙈 :D

chrismccord:
❤️❤️❤️🐥🔥

PR #3838, opened 2020-05-11
only do code reloading when env is dev

heavyk:
when using CodeReloader with vscode, the elixir_ls would continue to recompile the files I have `__phoenix_recompile__?` over and over. see https://github.com/elixir-lsp/elixir-ls/issues/239

I figure others who may be using vscode are likely to be experiencing the same problem. since the language server compiles the code with env test, this change will prevent the infinite loop as shown in the above bug report by restricting code reloading only to be in the dev environment.

question: would it be preferable instead if this were configurable or if it were to say, `Mix.env() != :test`?

### DIFF
--- a/lib/mix/tasks/compile.phoenix.ex
+++ b/lib/mix/tasks/compile.phoenix.ex
@@ -10,9 +10,11 @@ defmodule Mix.Tasks.Compile.Phoenix do
   def run(_args) do
     {:ok, _} = Application.ensure_all_started(:phoenix)
 
-    case touch() do
-      [] -> {:noop, []}
-      _  -> {:ok, []}
+    if Mix.env() == :dev do
+      case touch() do
+        [] -> {:noop, []}
+        _  -> {:ok, []}
+      end
     end
   end
 


chrismccord:
The code reloader only touches files that report they need recompiled ( `__phoenix_recompile__?/0` returns true), so this should not be touching files over and over since the __phoenix_recompile__ should not be claiming to need constant recompilation. For example, the only place we use this function is our templates, and they compare the hash of the file contents for wether to return true or false, and they will only return true if the template file contents changes. In your referenced issue, you are hardcoding __phoenix_recompile__ to return true, which will indeed cause a recompilations over and over, but that's not the case in practice. Can you clarify? Thanks!

heavyk:
it does continue to recompile (for whatever reason). open this repo in vscode and look for yourself:
https://github.com/heavyk/elixir-ls-bug-repro

you will notice that without doing anything, it'll recompile 8 files over and over (the ones with `__phoenix_recompile__?, do: true` in them)

heavyk:
oh, so maybe by clarify, you meant my workflow.. as I'm developing channels, I got tired of restarting phoenix every time I changed the channel. I found out that codereloader will reload the files I have that set to true every request. I added the code reloader as in the repo, and I noticed that indeed my channels were recompiled every time that I reloaded the page, but elixir_ls would do a lot of compilation..

---

I have an idea for a fix!! when touching the file, I should read it and generate a hash and save it into ets. so then in the recompile phase, then, it should only touch the file if there is no hash, or if the hash differs. would that be a better fix?

chrismccord:
The hardcoded `__phoenix_recompile__?` always returning true is the problem. The channel modules will be recompiled on refresh in development already so it not necessary. Additionally, phoenix_recompile is about indicating a module for recompilation when it depends on some external file changes, which is not the case here. If you are seeing otherwise, it's possible your elixir_ls extension is compiling in the same dev env and eating your changes. Perhaps it's working but you don't get a browser refresh on changes? If so you can add the channels dir to your live reload config:

```elixir
config :app, AppWeb,
  live_reload: [
    patterns: [
      ~r"priv/static/.*(js|css|png|jpeg|jpg|gif|svg)$",
      ~r"priv/gettext/.*(po)$",<% end %>
      ~r"lib/<%= lib_web_name %>/(live|views|channels)/.*(ex)$",  # <===== channels added
      ~r"lib/<%= lib_web_name %>/templates/.*(eex)$"
    ]
  ]

Closing as this does not appear to be an issue on our side. Thanks!

heavyk:
I don't think I'm using live_reload, because it's a json/websocket only api (no static files served). I looked into trying to figure out how to send a reload signal to the client over channels from looking at the injected code in the plug but it seemed difficult to figure out how to hook into that.

I definitely was not getting the channels to refresh on reload in dev. I tried adding channels to the config there like you showed, and it still does not appear to be working. I think it's because I'm not serving any static files.

chrismccord:
Try to isolate the issue outside of your editor. So quit your editor and use a different one to edit a channel. Issue an HTTP request and you should see your channel file recompiled.

chrismccord:
Ah I see, it's a `because it's a json/websocket only api`, so that is likely it. We don't recompile files for websocket/channel "requests". Only http requests. You could add dev specific code to your channels to issue a recompile, ie `Phoenix.CodeReloader.reload!(AppWeb.Endpoint)`

rmoorman:
@heavyk and @chrismccord , I am currently using [cortex](https://github.com/urbint/cortex) to overcome exactly that "only recompiling for http requests" in the phoenix and create-react-app based SPA I am working on right now. Maybe that could be of some help?

heavyk:
@rmoorman omg, that works beautifully! and it works really great for a lot of other things like automatically running tests and stuff

thank you so much for the suggestion!! ❤️ ❤️ ❤️

PR #3839, opened 2020-05-11
relax restriction on replies/broadcasts needing to be a map

heavyk:
as I'm developing in phoenix channels, I noticed a bit of a difference between apis. instead of adding restrictions, this PR aims to relax them.

in a channel it is possible to write:
```ex
def MyChannel do
  def join("some:room", msg, socket) do
    # ...
    {:ok, [:status, 1,2,"string", %{my: "map"}], socket}
  end
  # ... 
end
```

it's even possible to reply with just a single integer (which I thought was pretty cool), but in `handle_in/3`, I can send (from js) anything I want to, to, eg. `channel.push('counter', 3)` and the second argument is received as an integer, but I cannot reply or broadcast an integer back -- it has to be a map.

this commit relaxes the restriction allowing anything that the serializer is able to serialize as a valid payload.

now, it's possible to return `{:reply, {:ok, 1234}, socket}`

### DIFF
--- a/lib/phoenix/channel/server.ex
+++ b/lib/phoenix/channel/server.ex
@@ -136,7 +136,7 @@ defmodule Phoenix.Channel.Server do
   The message is encoded as `Phoenix.Socket.Broadcast`.
   """
   def broadcast(pubsub_server, topic, event, payload)
-      when is_binary(topic) and is_binary(event) and is_map(payload) do
+      when is_binary(topic) and is_binary(event) do
     broadcast = %Broadcast{
       topic: topic,
       event: event,
@@ -153,7 +153,7 @@ defmodule Phoenix.Channel.Server do
   Raises in case of crashes.
   """
   def broadcast!(pubsub_server, topic, event, payload)
-      when is_binary(topic) and is_binary(event) and is_map(payload) do
+      when is_binary(topic) and is_binary(event) do
     broadcast = %Broadcast{
       topic: topic,
       event: event,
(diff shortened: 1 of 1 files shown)


chrismccord:
We should rather move the other direction and only allow map payloads on join. Can you update the PR? Thanks!

heavyk:
why would you want to restrict it to a map? wrapping a counter or a list in an object seems like extra data to send over the wire. it should be anything that a serializer can handle -- and if the data cannot be sent, `serializer.encode!()` will already fail. it seems like an arbitrary restriction, so I looked to see if there was any instances where the payload was introspected and would fail if it was anything other than a map, but there was no places where it was introspected, so any valid json can be returned

if the serializer fails, then it's an invalid payload. there really shouldn't be any reason to restrict the payload

chrismccord:
It is an arbitrary restriction. If you search the issues you'll find some historical discussions around this, but the tldr; is we would rather be strict to allow external channel clients to have an expected payload contract. For example, just last week I was working with someone building a rust channels client and the shape of the expected payloads is non loose for a typed language. It all can be done, but we prefer to remain strict for ease of coordination.

heavyk:
I do think that the channel payload should be determined on an individual basis by the user (the rust channel implementation in this case).

anyway, as I was looking to expand out other things, I'm realising that to do other stuff I want to do, I have to do my own channel implementation anyway, so I'll just maintain a more liberal socket implementation as well. that also allows me to reduce the payload size (2 bytes!) as well, by using an integer for the ref.

so, whatever... I'm not going to worry about it. you can close this PR in you want.. I was just trying to offer a PR instead of just posting an issue and complaining. phoenix is quite nice and it's great that it's so easily modified.

Issue #3836, opened 2020-05-11
Allow using Brotli compressed assets


hauleth:
### Environment

* Elixir version (elixir -v): `1.10.3`
* Phoenix version (mix deps): any

### Expected behavior

When I create Brotli-compressed file (with extension `.br`) then `mix phx.digest` should also support it and use it in `cache_manifest.json` properly.

### Actual behavior

Brotli compressed files are treated as "regular" files.

josevalim:
A PR is welcome.

josevalim:
To clarify, I don't believe we should apply brotli compression automatically given it is not supported by default in the VM, but we should allow a workflow like this:

    $ npm deploy --prefix assets
    $ mix phx.digest.brotli       # <- this is a 3rd part task, not part of Phoenix
    $ mix phx.digest

If the workflow above does not work, then a PR is welcome. I will close this one for now as we don't plan to tackle it ourselves. Thanks!

hauleth:
I agree that Phoenix shouldn't apply Brotli compression to the assets. To be completely honest - I do not think it should apply GZIP either, it all should be handled by assets bundler. The problem is that it do not respect the Brotli compressed assets. So assuming that we have:

```
foo.css
foo.css.gz
foo.css.bz
```

Already then running `mix phx.digest` will result with:

```
foo.css
foo.css.gz
foo.css.bz
foo-<hash 1>.css
foo-<hash 1>.css.gz
foo.css-<hash 2>.bz
```

So `.bz` is. not treated as "compressed `foo.css`". That was the point of this issue.

I would love to provide PR, but right now I am too busy with other things and this issue was created as a result of the Slack conversation, so I would like it to be open for the time of being until I or someone will pick it up with a PR.

josevalim:
I see, this can hopefully be easily fixed, we just need to list all .br as a possible compression extension. I will reopen this for now but we may close it in the future if there is no interest in tackling this feature.

jayjun:
A third-party Mix task can work if `mix phx.digest` ignores `.br` files (see pull request above) then ran in this order

```shell
$ npm deploy --prefix assets
$ mix phx.digest
$ mix phx.digest.brotli
```

because `foo.css` must be digested into `foo-<hash>.css` before compression.

---

But pre-compressed Brotli assets are harder. Assets bundlers cannot supply pre-compressed _digest_ assets because digest files are not available before `mix phx.digest`.

[...]

PR #3840, opened 2020-05-12
Do not digest Brotli compressed files

jayjun:
`mix phx.digest` should not digest Brotli compressed files, which may come from third-party asset bundlers.

See #3836.


lib/phoenix/digester.ex
```diff
@@ -119,7 +119,7 @@ defmodule Phoenix.Digester do
 
   defp compiled_file?(file_path) do
     Regex.match?(@digested_file_regex, Path.basename(file_path)) ||
-      Path.extname(file_path) == ".gz" ||
+      Path.extname(file_path) in [".gz", ".br"] ||
```
kelvinst, line 122: isn't there any way to make this configurable? I mean, no problem on `[".gz", ".br"]` being the default config value, but I can see how this could be an useful compile time config

jayjun: `Plug.Static` only supports `.gz` and `.br` right now. Do you have other use cases in mind?

kelvinst: Oh, never mind, I though this was just a blacklist of files to not digest, now I read the actual function name and I think it does not make sense to allow people to configure it. Sorry for bothering

josevalim:
Looks great, we just need a test. :)

jayjun:
My bad, tests pushed!

josevalim:

PR #3842, opened 2020-05-12
Ignore Elixir LS cached data

joeapearson:
If you're using the Elixir language service then it caches files in a `.elixir_ls` directory alongside your other project files.  On the assumption that many other users are using Elixir LS, this pull request adds that directory to the template `.gitignore` file by default.

### DIFF
--- a/installer/templates/phx_single/gitignore
+++ b/installer/templates/phx_single/gitignore
@@ -32,3 +32,6 @@ npm-debug.log
 # we ignore priv/static. You may want to comment
 # this depending on your deployment strategy.
 /priv/static/
+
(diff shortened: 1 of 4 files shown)


lukaszsamson:
It might not be needed soon, see https://github.com/elixir-lsp/elixir-ls/pull/232

joeapearson:
Ah fair enough, I hadn't seen that PR.  I agree it'd be a better solution for elixir LS to manage this itself.

chrismccord:
This is something you should add to your global gitignore as we try to avoid things editor/external tooling specific. Thanks!

PR #3845, opened 2020-05-13
Fix a few minor typos in guides

fboyer:
A few minor typos were introduced in a recent commit that I happen to stumble upon.


mcrumm:
Thanks!

PR #3850, opened 2020-05-16
heroku.md: Setup -> set up

henrik:
"setup" is the noun and "set up" is the (phrasal) verb.

E.g. "I will set up the setup."


Gazler:
Thanks <3

PR #3852, opened 2020-05-17
Allow error_tag to have attr overrides

happycollision:
This allows developers to override the "class" attribute and/or add more
tag attributes to the default error_tag.


chrismccord:
Since this is defined in user land, you can and should extend it as needed with these kinds of  features, but it's better to keep it simple out of the box. Thanks!

happycollision:
Totally understandable. I figured maybe matching the behavior of other tag helpers out of the box would be a good thing. But the simplicity is nice indeed.

PR #3853, opened 2020-05-17
Update conn in Cowboy2 Handler

binaryseed:
This PR updates the Cowboy2 Handler so that it updates the conn after `call` - this is done so the telemetry event will get the updated values, for example the response `status`.

This mirrors the existing behavior of `plug_adapter` telemetry events in `plug_cowboy`...

https://github.com/elixir-plug/plug_cowboy/blob/master/lib/plug/cowboy/handler.ex#L40-L45

### DIFF
--- a/lib/phoenix/endpoint/cowboy2_handler.ex
+++ b/lib/phoenix/endpoint/cowboy2_handler.ex
@@ -60,6 +60,7 @@ defmodule Phoenix.Endpoint.Cowboy2Handler do
 
         {:plug, conn, handler, opts} ->
           %{adapter: {@connection, req}} =
+            conn =
             conn
             |> handler.call(opts)
             |> maybe_send(handler)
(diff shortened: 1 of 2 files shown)


lib/phoenix/endpoint/cowboy2_handler.ex
```diff
@@ -60,6 +60,7 @@ defmodule Phoenix.Endpoint.Cowboy2Handler do
 
         {:plug, conn, handler, opts} ->
           %{adapter: {@connection, req}} =
+            conn =
```
Gazler, line 63: Should we indent the subsequent lines?

```
            conn =
              conn
              |> handler.call(opts)
              |> maybe_send(handler)
```

binaryseed: I'll format it however you want, I have the code formatter on in my editor and this is what it does...

binaryseed:
cc @josevalim

josevalim:

PR #3862, opened 2020-05-21
Grammar: "Peace of mind" is not hyphenated

edjroot:


henrik:
This phrasing (with hyphens) is also on https://www.phoenixframework.org/ and in the GitHub repo description.

josevalim:
Thanks @edjroot!

I am trying to understand this better. According to my Google skills, we hyphenate words when we have a compound adjectives before the noun. But I am struggling find the noun in "Peace-of-mind from prototype to production". My *assumption* is that the sentence leaves it up to the reader to know exactly what "peace of mind" applies to. For example, I could read it as:

  * *Phoenix gives you* peace of mind from prototype to production

But I could also read it as:

  * Peace-of-mind *web development* from prototype to production

Although I have no idea if any of these interpretations are valid. Help? :)

henrik:
I read it as "Get peace of mind from prototype to production", and in that case, I believe "peace of mind" would be a noun phrase, and the hyphens shouldn't be there.

(Casting my mind back to uni linguistics classes a decade ago… I'm sure my terminology is off :D)

I agree that if it's seen as adjectival as in the second example above, the hyphens would make good sense. But I feel the "web development" would need to stay in there – that interpretation of the sentence doesn't work (to me anyway) if it's left out.

Compound adjectives could work in other cases without the modified noun being present, e.g. "Long-nosed since birth" or whatever :)

I think the difference is down to "long" actually being an adjective ("Long since birth" would also work) but "peace of mind" only becomes adjectival by modifying another noun. Like "school" in "school bus". You couldn't say "school since birth".

chrismccord:
👍🏻 updated ❤️❤️❤️🐥🔥

PR #3863, opened 2020-05-22
Warn if the host provided for the Endpoint is invalid

leandrocp:
I'm pretty sure the validation may be improved, but this PR is to gather feedback related to https://github.com/phoenixframework/phoenix_live_view/issues/911#issuecomment-632905958

Decided to warn instead of raise to follow the same pattern already in use for validating Endpoint config, but I can change to raise if it's better.

### DIFF
--- a/lib/phoenix/endpoint/supervisor.ex
+++ b/lib/phoenix/endpoint/supervisor.ex
@@ -303,6 +303,10 @@ defmodule Phoenix.Endpoint.Supervisor do
     host   = host_to_binary(url[:host] || "localhost")
     port   = port_to_integer(url[:port] || port)
 
+    if host =~ ~r/\:/ do
+      Logger.warn("host #{inspect(host)} is invalid")
+    end
+
     %URI{scheme: scheme, port: port, host: host}
   end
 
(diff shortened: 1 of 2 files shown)


lib/phoenix/endpoint/supervisor.ex
```diff
@@ -303,6 +302,10 @@ defmodule Phoenix.Endpoint.Supervisor do
     host   = host_to_binary(url[:host] || "localhost")
     port   = port_to_integer(url[:port] || port)
 
+    if host =~ ~r/\:/ do
```
josevalim, line 305:
```suggestion
    if host =~ ":" do
```

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -303,6 +302,10 @@ defmodule Phoenix.Endpoint.Supervisor do
...
+      Logger.warn(":host configuration value #{inspect(host)} for #{inspect(endpoint)} is invalid")
```
josevalim, line 306:
```suggestion
      Logger.warn("url: [host: ...] configuration value #{inspect(host)} for #{inspect(endpoint)} is invalid")
```

josevalim:

PR #3869, opened 2020-05-27
Demonstrate an issue with exception telemetry events

binaryseed:
I've come across a strange issue that happens inside a telemetry handler for the `[:plug_adapter, :call, :exception]` event.

Attempting to `inspect` a value inside the Telemetry handler causes a failure that I can't explain. I don't see the log entry that I would expect from the handler being detached...

I'm not totally sure this has something to do with Phoenix, but this particular event is executed inside of an existing `catch`:

https://github.com/phoenixframework/phoenix/blob/c85a23b2b72c0c29a44ed61e1137a5f79c428fee/lib/phoenix/endpoint/cowboy2_handler.ex#L76-L83

* First commit shows a handler functioning correctly
* Second commit shows an `IO.inspect` causing the rest of the handler fail to execute

@josevalim Any idea what could be happening here?

### DIFF
--- a/test/phoenix/integration/endpoint_test.exs
+++ b/test/phoenix/integration/endpoint_test.exs
@@ -147,7 +147,7 @@ defmodule Phoenix.Integration.EndpointTest do
       :plug_adapter_exception,
       [:plug_adapter, :call, :exception],
       fn event, measurements, metadata, :none ->
-        # IO.inspect {measurements, metadata}
+        IO.inspect {measurements, metadata}
         send(test, {:event, event, measurements, metadata})
       end,
       :none
--- a/test/phoenix/integration/endpoint_test.exs
+++ b/test/phoenix/integration/endpoint_test.exs
@@ -139,6 +139,26 @@ defmodule Phoenix.Integration.EndpointTest do
     end
   end
 
+  test ":exception inspect totally fails" do
+    test = self()
+    {:ok, _} = ProdEndpoint.start_link()
+
+    :telemetry.attach(
+      :plug_adapter_exception,
+      [:plug_adapter, :call, :exception],
+      fn event, measurements, metadata, :none ->
+        # IO.inspect {measurements, metadata}
+        send(test, {:event, event, measurements, metadata})
+      end,
+      :none
+    )
+
+    {:ok, resp} = HTTPClient.request(:get, "http://127.0.0.1:#{@prod}/oops", %{})
+    assert resp.status == 500
+
+    assert_receive {:event, [:plug_adapter, :call, :exception], %{duration: _}, _}
+  end
+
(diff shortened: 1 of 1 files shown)


josevalim:
So what is happening is that, once you return 500, the client closes the connection, so Cowboy kills the request process (the one emitting the telemetry event), without waiting for it to terminate. Trapping exits solve it but I don't think we should always do it. I would say we are emitting the event in the wrong place and we should emit it elsewhere. Maybe we should move it to when the body is sent in the adapter.

josevalim:
To be fair, this is working according to telemetry:span/3 proposal. The event is not guaranteed to be emitted in case of linked exits.

binaryseed:
Ahh, so the `inspect` is a red herring, it's actually just any operation that takes a meaningful chunk of time! I confirmed this by doing a `Process.sleep` instead and the same behavior happens

If this is the case, is it true that both the `:stop` and `:exception` event in Plug and here in Phoenix are at risk of getting terminated at essentially an arbitrary point in time?

Like you said, perhaps the termination events should be moved so that they always happen before cowboy starts shutting down the request process. Where do you think that could happen? I could look into a PR for that to see how that might work out...

binaryseed:
Aside from moving the `execute` call, would it be bad practice / dangerous to call `Process.flag(:trap_exit, true)` inside the handler callback itself? It seems hacky but it does resolve the issue

josevalim:
> Aside from moving the execute call, would it be bad practice / dangerous to call Process.flag(:trap_exit, true) inside the handler callback itself? It seems hacky but it does resolve the issue

The race still exists though. It can terminate before you call the trap_exit. It may be less likely to happen though.

The only guarantee I can think is by calling it on the send_resp adapter callback.

binaryseed:
Would it work to execute a callback inside `Plug.Conn.send_resp`?

https://github.com/elixir-plug/plug/blob/master/lib/plug/conn.ex#L405

That seems like it can replace the `:stop` easily, but not the `:exception`...

josevalim:
Ah, true. Maybe we need to use cowboy stream or something else then. :(

binaryseed:
I'm noticing that cowboy has a handler for metrics, and it takes a callback function:

https://github.com/ninenines/cowboy/blob/master/src/cowboy_metrics_h.erl

It looks like that's all a layer below Plug concepts though, but it does look more reliable since it's wired up inside cowboy itself

josevalim:
Good find! We should definitely consider moving the telemetry events there.

binaryseed:
This isn't a Phoenix issue, so I'm closing this here and will continue the conversation in `plug_cowboy`

binaryseed:
`plug_cowboy` issue: https://github.com/elixir-plug/plug_cowboy/issues/38

PR #3880, opened 2020-06-02
Use entrypoint name for extracted CSS

alexblackie:
Instead of hard-coding `app`, we should use the entrypoint name, which
allows multiple CSS entrypoints to be generated automatically. This is
especially useful when you want to segregate your frontends, such as
both "public" and "admin" sections which need entirely independent CSS
assets.


Gazler:
Thanks <3

PR #3886, opened 2020-06-04
Docs: missed reference to `@view_template`

richardkmichael:


richardkmichael:
I'm not 100% this should change.  The "ie" example code uses `view_template/1`, but perhaps it is supposed to imply the internal use of reserved assign `@view_template` ?

PR #3889, opened 2020-06-05
Update cowboy2_adapter.ex

### LINKED ISSUE #3888: Documentation on Drainer not clear
* Phoenix version 1.5.3

Currently the docs state
>
```
:drainer` - a drainer process that triggers when your application is
      shutting to wait for any on-going request to finish. It accepts all
      options as defined by [`Plug.Cowboy`](https://hexdocs.pm/plug_cowboy/Plug.Cowboy.Drainer.html).
      Defaults to `[]` and can be disabled by setting it to false.
```
Does `[]` mean that Drainer is enabled by default? I checked the implementation but the code is a too convoluted to get a clear answer. Also in my supervision tree I don't see the drainer process.

leandrocp:
Close #3888

Ref https://github.com/phoenixframework/phoenix/blob/33ba200025afcf79f854dc6bc11bd9e27535056c/lib/phoenix/endpoint/cowboy2_adapter.ex#L69-L73

### DIFF
--- a/lib/phoenix/endpoint/cowboy2_adapter.ex
+++ b/lib/phoenix/endpoint/cowboy2_adapter.ex
@@ -15,9 +15,10 @@ defmodule Phoenix.Endpoint.Cowboy2Adapter do
       to `false`
 
     * `:drainer` - a drainer process that triggers when your application is
-      shutting to wait for any on-going request to finish. It accepts all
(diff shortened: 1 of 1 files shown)


bryannaegele:
It might be informative for users to also add a note that `Plug.Cowboy.Drainer` only stops accepting new connections and doesn't actively close long-lived connections, such as sockets.

PR #3894, opened 2020-06-08
Add focus trapping and add aria tags to modal

LostKobrakai:


chrismccord:
we shouldn't ship with a predefined hook in user land. If anything I would instead prefer to ditch the escape key from the generator vs the requiring custom JS in user land. Can you elaborate on the accessibility issues with the escape key closing the modal? Thanks!

LostKobrakai:
There's nothing wrong with the esc key. Those are additional things to improve accessibility in modals besides the already existing use `esc` to close. Focus trapping keeps the focus on elements within the modal, so you can't for example tab to elements below the modal and the other additions are just some basic aria tags, e.g. `aria-modal` hints to assistive tech that the modal should be the current context and not the stuff in the background (like focus trapping does for focus).

> We shouldn't ship with a predefined hook in user land.

That's why I was wondering if it might be a good idea to have e.g. `phx-trap-focus` and handle that part from phoenix_liveview directly.

PR #3897, opened 2020-06-08
Fix typo in transport.ex

pgeraghty:


chrismccord:
In this case, the connect function does not receive state, but it should build and return state, using the map of options, params, etc. Thanks!

PR #3899, opened 2020-06-10
Speed up build times for webpack

Gazler:
On an unmodified default generated project (with --live) using the
Phoenix 1.5.3 installer, the build time is:

    $ webpack --mode development

    Version: webpack 4.41.5
    Time: 1681ms

Changing the source-map mode to `eval-cheap-module-source-map` changes
this time slightly:

    Time: 1397ms

The `hard-source-webpack-plugin` caches intermediate build steps, which
produces significantly faster rebuilds:

    Time: 474ms

Combining these two results in:

    Time: 193ms

This improvement seems to scale well too. On a project including
tailwind, the build time by default is close to 6 seconds, but with the
above config, 212ms.

`eval-cheap-module-source-map` preserves line numbers in the source
maps, but not column information. This seems like a resonable trade off
vs `eval-module-source-map` which is a bit slower.

    Time: 491ms

### DIFF
--- a/installer/templates/phx_assets/package.json
+++ b/installer/templates/phx_assets/package.json
@@ -20,6 +20,7 @@
     "css-loader": "^3.4.2",
     "sass-loader": "^8.0.2",
     "node-sass": "^4.13.1",
+    "hard-source-webpack-plugin": "^0.13.1",
     "mini-css-extract-plugin": "^0.9.0",
     "optimize-css-assets-webpack-plugin": "^5.0.1",
     "terser-webpack-plugin": "^2.3.2",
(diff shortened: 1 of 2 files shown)


Gazler:
I don't think we'll run into any of the issues listed in https://github.com/mzgoddard/hard-source-webpack-plugin#troubleshooting but we could disable this plugin for production mode to be safe.

joshchernoff:
> I don't think we'll run into any of the issues listed in https://github.com/mzgoddard/hard-source-webpack-plugin#troubleshooting but we could disable this plugin for production mode to be safe.
> 
> ```diff
> -    plugins: [
> -      new HardSourceWebpackPlugin(),
> -      new MiniCssExtractPlugin({ filename: '../css/[name].css' }),
> -      new CopyWebpackPlugin([{ from: 'static/', to: '../' }])
> -    ]
> +    plugins: []
> +      .concat(devMode ? [new HardSourceWebpackPlugin()] : [])
> +      .concat([
> +        new MiniCssExtractPlugin({ filename: '../css/app.css' }),
> +        new CopyWebpackPlugin([{ from: 'static/', to: '../' }])
> +      ])
> ```

This is a great reason why to have separate webpack configs per env.

PR #3902, opened 2020-06-11
Fix typo

johnshaughnessy:
The word in the `phx.routes` `mix` task appears to be misspelled. 

Disclaimer: I am learning phoenix/elixir and don't understand this code yet. This might not be the right fix so someone more knowledgeable should check.


josevalim:

PR #3908, opened 2020-06-20
Fix plugs as composition example

ShPakvel:
- Eliminate plug methods dependency from the first example
- Reuse data already found, like message

It makes more clear and fair comparison.
But it still shows nesting complexity in the first example and
much more composable, clear, and reusable way in the second example.


josevalim:

PR #3911, opened 2020-06-20
Fix migrate task description

ShPakvel:
- Remove not existed -v option
- Correct description of --to option


josevalim:

PR #3912, opened 2020-06-21
Reference Gigalixir docs to install the CLI now that we have differen…

jesseshieh:
…t methods of installation for each platform.


chrismccord:
❤️❤️❤️🐥🔥

PR #3913, opened 2020-06-22
Fix bug in presence documentation example

mveytsman:
The example presence code has a bug where it's building a map of users keyed by id, but then using a key that it gets from the presences map to index into it. The users table is going to have integer primary keys, but presences keys are cast to a string (see https://hexdocs.pm/phoenix/Phoenix.Presence.html#c:track/3) 

This means that the code as written will always return `nil` for the user.

I'm not sure if adding `String.to_integer` here is the cleanest approach, please let me know if there's a better a way to do this!


josevalim:

PR #3914, opened 2020-06-23
Move ‘invalid-feedback’ style selector block to non-LV section

DavidOliver:
The 'invalid-feedback' class is used in standard generated forms.


chrismccord:
❤️❤️❤️🐥🔥

PR #3916, opened 2020-06-25
Fix bug in mix phx.gen.json shell instructions

foucist:
(Note: example given is from the phx.gen.json test)

After `mix phx.gen.json Blog Post posts title:string --web Blog` is run, the generated files rely on `Routes.blog_post_path`.

The outputted shell instructions should include the alias `as: :blog` in the routing scope so that `Routes.blog_post_path` works.


josevalim:

PR #3918, opened 2020-06-26
Update package-lock.json to match package.json

iobtl:
Package.json was updated to 1.5.0 in master branch while package-lock.json was still at 1.4.0. This may cause issues with older versions of npm.


josevalim:

PR #3921, opened 2020-06-29
Make code snippet match text; uses render_to_string not render_to_iodata

ali-raheem:
The code snipper makes use of `render_to_string/3` but the preceeding text instead references `render_to_iodata/3`.

This mismatch is confusing.


josevalim:

PR #3924, opened 2020-07-02
Configurable static compressors

mitchellhenke:
Following on the work in #3836 and #3840, I wanted to have a quick go at a more robust feature described in https://github.com/phoenixframework/phoenix/issues/3836#issuecomment-629623791

I added `compress_file?` to the behaviour to be able to not compress some assets with certain compressors if the content is too small (or similar).

I haven't written any docs/tests yet, and the naming/structure could use some updates, but figured I'd start the discussion with this 🙂

### DIFF
--- a/lib/mix/tasks/phx.digest.ex
+++ b/lib/mix/tasks/phx.digest.ex
@@ -37,6 +37,7 @@ defmodule Mix.Tasks.Phx.Digest do
 
   @doc false
   def run(all_args) do
+    Mix.Task.run "compile", all_args
     {opts, args, _} = OptionParser.parse(all_args, switches: [output: :string], aliases: [o: :output])
     input_path = List.first(args) || @default_input_path
     output_path = opts[:output] || input_path
--- a/lib/phoenix/digester.ex
+++ b/lib/phoenix/digester.ex
@@ -119,7 +119,7 @@ defmodule Phoenix.Digester do
 
   defp compiled_file?(file_path) do
     compressors = Application.fetch_env!(:phoenix, :static_compressors)
-    compressed_extensions = Enum.map(compressors, &(&1.file_extension))
+    compressed_extensions = Enum.flat_map(compressors, &(&1.file_extensions))
 
     Regex.match?(@digested_file_regex, Path.basename(file_path)) ||
       Path.extname(file_path) in compressed_extensions ||
@@ -157,14 +157,15 @@ defmodule Phoenix.Digester do
       if compressor.compress_file?(file.absolute_path, file.content, file.digested_content) do
         compressed_digested = compressor.compress(file.digested_content)
         compressed = compressor.compress(file.content)
+        [file_extension | _] = compressor.file_extensions
 
         File.write!(
-          Path.join(path, file.digested_filename <> compressor.file_extension),
+          Path.join(path, file.digested_filename <> file_extension),
           compressed_digested
         )
 
         File.write!(
-          Path.join(path, file.filename <> compressor.file_extension),
+          Path.join(path, file.filename <> file_extension),
           compressed
         )
       end
--- a/lib/phoenix/digester.ex
+++ b/lib/phoenix/digester.ex
@@ -154,16 +154,16 @@ defmodule Phoenix.Digester do
     compressors = Application.fetch_env!(:phoenix, :static_compressors)
 
     Enum.each(compressors, fn(compressor) ->
-      if compressor.compress_file?(file.absolute_path, file.content, file.digested_content) do
-        compressed_digested = compressor.compress(file.digested_content)
-        compressed = compressor.compress(file.content)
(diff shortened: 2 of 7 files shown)


lib/mix/tasks/phx.digest.ex
```diff
@@ -37,6 +37,7 @@ defmodule Mix.Tasks.Phx.Digest do
 
   @doc false
   def run(all_args) do
+    Mix.Task.run "compile", all_args
```
josevalim, line 40: Do we need this line? :)

lib/phoenix/digester.ex
```diff
@@ -118,8 +118,11 @@ defmodule Phoenix.Digester do
   defp manifest_join(path, filename), do: Path.join(path, filename)
 
   defp compiled_file?(file_path) do
+    compressors = Application.fetch_env!(:phoenix, :static_compressors)
+    compressed_extensions = Enum.map(compressors, &(&1.file_extension))
```
josevalim, line 122: It may be worth calling it `file_extensions` and using `flat_map` in case one format can work with mutliple extensions.

lib/mix/tasks/phx.digest.ex
josevalim: Ah, in case the compressor is defined within my project.

lib/phoenix/digester.ex
josevalim: And then we pick the first one when writing the file to disk.

lib/phoenix/digester.ex
```diff
@@ -148,18 +151,24 @@ defmodule Phoenix.Digester do
...
-
-      File.write!(
-        Path.join(path, file.filename <> ".gz"),
-        :zlib.gzip(file.content)
-      )
-    end
+    compressors = Application.fetch_env!(:phoenix, :static_compressors)
+
+    Enum.each(compressors, fn(compressor) ->
+      if compressor.compress_file?(file.absolute_path, file.content, file.digested_content) do
```
josevalim, line 157: Do we need to pass content and digested_content? :)

josevalim:
I like it! :+1:

lib/phoenix/digester.ex
mitchellhenke: done!

lib/phoenix/digester.ex
mitchellhenke: This is something I was less sure about. I took some inspiration from [cowboy's](https://github.com/ninenines/cowboy/blob/master/src/cowboy_compress_h.erl#L119-L126) (non-static) response compression and [Webpack compression plugin](https://webpack.js.org/plugins/compression-webpack-plugin/), where there is a threshold for the size. The webpack plugin also allows specifying a minimum compression ratio.

A quick example is robots.txt, where if you want no crawling, the content is:
```
User-agent: *
Disallow: /
```

Which ends up being 26 bytes, and running `mix phx.digest` generates robots.txt.gz which ends up being larger at 46 bytes.

In terms of supplying both content and digested_content, I was erring on supplying more information, as they potentially are different sizes or end up compressing slightly differently. I also considered passing the whole `file` map, but I was worried about exposing too much of the structure publicly and potentially locking things in.

josevalim: @mitchellhenke really smart! Maybe a better API is then to call `compress_file(filename, contents)` which returns `{:ok, result} | :error`? For example, could I want to encode one but not the other?

josevalim:
Looks great! We just need docs and we can ship it!

josevalim:
And tests! :D

lib/phoenix/digester/gzip.ex
```diff
@@ -0,0 +1,18 @@
+defmodule Phoenix.Digester.Gzip do
+  @moduledoc ~S"""
+  Gzip compressor for Phoenix.Digester
+  """
+  @behaviour Phoenix.Digester.Compressor
+
+  def compress_file(file_path, content) do
+    if Path.extname(file_path) in Application.fetch_env!(:phoenix, :gzippable_exts) do
+      {:ok, :zlib.gzip(content)}
```
josevalim, line 9: Should we also check the file is smaller, like we do for brotli? In any case, that can be done in a separate PR. :)

josevalim:

benregn:
I'm pretty new to Phoenix and I'm not familiar with your release process. Could I ask which release will this be in?

josevalim:
Phoenix v1.6 (current master)!

benregn:
Awesome❗ Got a bit confused when I followed the docs added in [`c7b7afa`](https://github.com/phoenixframework/phoenix/pull/3924/commits/c7b7afaf7a721f9401f044f21547c431319e4db3) but it didn't work and the code I had in `lib` didn't match what I saw on `master` 😅

fschoenfeldt:
Good Job @josevalim !

PR #3925, opened 2020-07-08
Fix broken link to contexts guide

the-undefined:
The contexts guide link and name are pluralized, but this page was linking to a singular version that doesn't exist.

Screenshot 2020-07-08 at 06 33 29

Screenshot 2020-07-08 at 12 13 45

Not reading them in order means I am jumping around quite a bit :)


Gazler:
Thanks <3

PR #3928, opened 2020-07-10
Add additional example of using tag_values in Telemetry guide

aaronrenner:
I'm finding `:tag_values` is really powerful when controlling how metrics show up in LiveDashboard, so I thought it might be worth adding another example. This also serves to promote the new `Phoenix.LiveView` telemetry events.


josevalim:

PR #3929, opened 2020-07-11
Keep example and clarification consistent

sodapopcan:
The example code uses `#{messenger}` while the clarifying explanation uses `#{id}`.  Perhaps this was on purpose—and if so, please carry on—but it made me do a few double takes trying to figure out where `#{id}` was referenced in the example.


josevalim:

PR #3938, opened 2020-07-24
Add :conn to router_dispatch exception events

mcrumm:
Looking for feedback. This would allow introspection of the connection as it existed when the router_dispatch start event was executed, which will be helpful for debugging, especially in applications with more than one router.


josevalim:
Yeah, it should have been there since the beginning. We should look into using the new telemetry:span so we stop making such mistakes. :D I will merge this so we can backport it too!

josevalim:

PR #3940, opened 2020-07-26
Expose trace context headers in websockets and long polling

bryannaegele:
In order to correlate distributed traces to actions within websockets we need to have access to trace context headers in order to extract the context without complicated workarounds. 

The [W3C Trace Context Headers](https://www.w3.org/TR/trace-context-1/) are in a firm place naming-wise to add support in Phoenix for usage by projects such as OpenTelemetry. This PR exposes the two official context headers to the `connect_info` map. This is necessary to provide seamless tracing of websockets for users in the `opentelemetry_phoenix` project.

/cc @josevalim @tsloughter @hauleth

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -768,7 +768,11 @@ defmodule Phoenix.Endpoint do
 
         * `:peer_data` - the result of `Plug.Conn.get_peer_data/1`
 
-        * `:trace_context_headers` - a list of all trace context headers
+        * `:trace_context_headers` - a list of all trace context headers. Supported
+          headers are defined by the [W3C Trace Context Specification](https://www.w3.org/TR/trace-context-1/).
+          These headers are necessary for libraries such as [OpenTelemetry](https://opentelemetry.io/) to extract
+          trace propagation information to know this request is part of a larger trace
+          in progress.
 
         * `:x_headers` - all request headers that have an "x-" prefix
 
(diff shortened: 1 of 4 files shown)


lib/phoenix/endpoint.ex
```diff
@@ -768,6 +768,8 @@ defmodule Phoenix.Endpoint do
 
         * `:peer_data` - the result of `Plug.Conn.get_peer_data/1`
 
+        * `:trace_context_headers` - a list of all trace context headers
```
tsloughter, line 771: Should include examples in the docs.

tsloughter:
Nice +1

lib/phoenix/endpoint.ex
bryannaegele: I figured we'd have examples in https://github.com/opentelemetry-beam/opentelemetry_phoenix and the main lib. Maybe one in the telemetry guide in this repo, too?

tolbrino: A link to the W3C spec as already mentioned in the PR message would be nice here too.

josevalim:
Looks great! We just need more docs and examples. The docs should be self-sufficient even if someone never heard about open telemetry before. :D

bryannaegele:
I added some _very_ preliminary docs. Until I have opentelemetry-beam/opentelemetry_phoenix in a place for release with docs, examples, and patterns laid out there, I'm hesitant to add much more right now. The lib is in-progress and nobody I've spoken to has done tracing with sockets, let alone with Otel.

We need to figure the patterns and best practices figured out first, then we can document further 😆

Sound good?

josevalim:

PR #3946, opened 2020-07-29
use `~` in scss to resolve from node_modules

LostKobrakai:
See https://webpack.js.org/loaders/css-loader/#import for details

### DIFF
--- a/installer/templates/phx_assets/app.scss
+++ b/installer/templates/phx_assets/app.scss
@@ -1,6 +1,6 @@
 /* This file is for your main application CSS. */
 @import "./phoenix.css";<%= if live do %>
-@import "../node_modules/nprogress/nprogress.css";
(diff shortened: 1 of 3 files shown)


createdbypete (approved):

cw789:
I like the notation, but for people not knowing webpack well it's not obvious anymore what's really happening...
And we'll make a file depending on webpack that probably don't need to know anything about webpack.

LostKobrakai:
Even just the ˋ@importˋ working like it does is css-loader and therefore webpack behaviour. But I can see that it‘s less obvious.

cw789:
Point with `@import`. Yes right.

cw789:
Obsolete now as nprogress was replaced with topbar.
Thanks for learning me this neat feature.

PR #3949, opened 2020-07-30
Make instructions for importing socket.js accurate

eksperimental:


installer/templates/phx_assets/app.js
```diff
@@ -9,9 +9,8 @@ import "../css/app.scss"
 //
 // Import deps with the dep name or local files with a relative path, for example:
 //
-//     import {Socket} from "phoenix"
-//     import socket from "./socket"
-//
+// import {Socket} from "phoenix"
+// import socket from "./socket"
```
chrismccord, line 13: we favor the `import {Socket} from "phoenix"` syntax. Can you update the guide only to make it clearer, but leave the import? Thanks!

eksperimental: done

chrismccord:
Sorry for the back and forth, but I was wrong earlier. In this case we are importing `socket.js`, which itself does the `{Socket} from "phoenix"` import. So the current released snippet is correct. Can you update the instructions if you think they could be clearer? Thanks!

eksperimental:
> Sorry for the back and forth, but I was wrong earlier. In this case we are importing `socket.js`, which itself does the `{Socket} from "phoenix"` import. So the current released snippet is correct. Can you update the instructions if you think they could be clearer? Thanks!

Sure. No worries.

Please have a look, I have edited the template as I think it is better to have commented line to be uncommented out of the docs.

installer/templates/phx_assets/app.js
```diff
@@ -33,4 +33,4 @@ liveSocket.connect()
 // >> liveSocket.enableLatencySim(1000)  // enabled for duration of browser session
 // >> liveSocket.disableLatencySim()
 window.liveSocket = liveSocket
-<% end %>
+<% end %>
```
chrismccord, line 36:
```suggestion
<% end %>

```

chrismccord:
Thanks for being patient on this one. In the interest of less generated comments, I'd prefer to the keep the single comment with the socket import. Thanks!

PR #3959, opened 2020-08-07
Remove XSS vulnerability with proper escaping

al2o3cr:
A reader might copy this example without realizing a big difference between `#{}` and `<%= %>` - the former allows arbitrary HTML to be inserted.

Should this be called out more specifically in the paragraph following the example?


chrismccord:
❤️❤️❤️🐥🔥

PR #3962, opened 2020-08-18
Run installer tests in ci

aaronrenner:
It appears that the installer tests accidentally were removed during the migration to GitHub Actions. This adds those tests back into CI.


mcrumm:
Thank you! :heart:

PR #3966, opened 2020-08-22
Correct floki dep

skylerwshaw:
The current mix.exs templates don't specify the required version of the [Floki](https://github.com/philss/floki) dependency. The test suite consumes `Floki.parse_document/1`, which was available as of Floki [0.24.0](https://github.com/philss/floki/blob/master/CHANGELOG.md#0240---2020-01-01).

Since we're currently specifying semver `>=0.0.0`, other dependencies such as my recent addition of `{:phoenix_inline_svg, "~> 1.3"}`, which consume older versions of Floki can result in the application reverting to a dependency prior to the addition of `parse_document/1`.

Related PR to update documentation and bump the version in `phoenix_live_view`: https://github.com/phoenixframework/phoenix_live_view/pull/1128


josevalim:

PR #3967, opened 2020-08-22
Explicitly mention `mix phx.new` in Installation guide

toraritte:
It is indeed mentioned in _Up and Running_, but this is the page I re-read first every time I return to Phoenix after several months with other langs, and always have to click around because I forgot what the exact command is (plus it also changed from `mix phoenix.new` at one point). It might also be prudent to give proper context to newcomers as well.

Thanks for considering!


josevalim:

Issue #3969, opened 2020-08-24
New project compilation fails with Elixir 1.9.4 and OTP 20.3.8.26


aaronrenner:
I'm attempting to generate a new project with Elixir 1.9.4 and OTP 20.3.8.26 (the minimum  Elixir/OTP versions specified in `.github/workflows/ci.yml`) and cowboy is failing to compile.

## Steps to recreate

1. Set your elixir version to 1.9.4 and erlang version to 20.3.8.26
2. Run the following commands to set up a new test app
```shell-session
$ cd installer
$ mix phx.new dev_app
$ cd dev_app
$ mix deps.get
$ mix test
==> connection
Compiling 1 file (.ex)
Generated connection app
==> gettext
Compiling 1 file (.yrl)
Compiling 1 file (.erl)
Compiling 20 files (.ex)
Generated gettext app
===> Compiling ranch
===> Compiling telemetry
==> telemetry_metrics
Compiling 7 files (.ex)
Generated telemetry_metrics app
===> Compiling telemetry_poller
==> decimal
Compiling 1 file (.ex)
Generated decimal app
==> jason
Compiling 8 files (.ex)
Generated jason app
==> db_connection
Compiling 14 files (.ex)
Generated db_connection app
==> ecto
Compiling 55 files (.ex)
Generated ecto app
==> phoenix_pubsub
Compiling 11 files (.ex)
Generated phoenix_pubsub app
===> Compiling cowlib
===> Compiling cowboy
===> Compiling src/cowboy_loop.erl failed
src/cowboy_loop.erl:80: illegal pattern
src/cowboy_loop.erl:81: variable 'Reason' is unbound
src/cowboy_loop.erl:82: variable 'Stacktrace' is unbound

==> dev_app
** (Mix) Could not compile dependency :cowboy, "/Users/aaron/.asdf/installs/elixir/1.9.4/.mix/rebar3 bare compile --paths="/Users/aaron/projects/phoenix/installer/dev_app/_build/test/lib/*/ebin"" command failed. You can recompile this dependency with "mix deps.compile cowboy", update it with "mix deps.update cowboy" or clean it with "mix deps.clean cowboy"
```

## Other notes

According to #3903 it seems like cowboy may require OTP 21.3. Do we need to upgrade Phoenix's minimum OTP version to this?

josevalim:
[...]

PR #3970, opened 2020-08-25
Add integration tests and fix formatting in telemetry.ex

aaronrenner:
This adds very basic integration tests which generate a new phoenix app, check for compilation warnings and ensure the code is properly formatted. I'm planning to expand on this in future PRs to test the generators, add dependency caching and ensure the generated app's test suites pass.

This also fixes two issues:
1. telemetry.ex has some formatting issues that were caught
2. Cowboy won't compile with the current OTP 20.3.8.26 (#3969). Things finally started compiling correctly with OTP 21.2.7

## Other notes

I am running the mix tasks via the command line so the tests can be run with `async: true`. This required me to create a `MixTaskRunner` module to respond to prompts, and is a variant of what's used in phx_gen_auth.

Integration tests are set up to run as a separate job in CI so we can get faster feedback around if the traditional test suite passed.

I do have some work on a local branch that downloads, compiles and caches the mix dependencies, but it was being flakey in CI, so I thought it could use some more work.

Please let me know if this is going in the right direction for integration tests. I also had a goal of keeping this PR small and iterating on it to add more tests and fix issues that come up. Any feedback would be greatly appreciated!


integration_test/support/code_generator_case.exs
```diff
@@ -0,0 +1,84 @@
...
+    quote do
+      setup_all do
+        # Get Mix output sent to the current
+        # process to avoid polluting tests.
+        Mix.shell(Mix.Shell.Process)
+      end
+
+      import unquote(__MODULE__)
+    end
+  end
```
josevalim, line 18:
```suggestion
  using do
    quote do
      import unquote(__MODULE__)
    end
  end
  
  setup_all do
    # Get Mix output sent to the current
    # process to avoid polluting tests.
    Mix.shell(Mix.Shell.Process)
  end
```

integration_test/support/mix_task_runner.exs
```diff
@@ -0,0 +1,170 @@
...
+    prompt_responses =
+      opts
+      |> Keyword.get(:prompt_responses, [])
+      |> List.wrap()
+
+    port =
+      Port.open(
+        {:spawn_executable, mix_path},
+        [:exit_status, :stderr_to_stdout, args: args, env: get_env(opts)] ++ port_options
+      )
```
josevalim, line 79: Can we use `System.cmd` instead of Port.open? 🤔

josevalim: Actually, I see you want to have control over the response... I am honestly a bit torn. Shelling out to start multiple commands are going to make the integration tests very expensive. I would prefer if we could invoke Mix directly from the current instance, without starting a new app, and use Mix.Shell.Process. It makes the tests less of integration tests but I think the benefits will be visible in the long term. It is a difference between commands taking 2ms vs 200ms. We do already have a structure similar to what I described in `phx.new_test` which maybe we can extend to support these use cases?

Thoughts?

aaronrenner: I started with the approach that phoenix currently uses of running the new app inside of the current beam instance, but ran into a few problems.

1. Mix is designed to be a singleton, so running tests with `async: true` will cause unexpected errors. The tests in this suite are about 30 seconds faster on my machine when running `mix test.integration` (async) instead of `mix test.integration --trace`
1. I had difficulty with checks like [this one](https://github.com/aaronrenner/phx_gen_auth/blob/1a8e74deb4eff273dc60b16b64dbee1ad701d91e/lib/mix/tasks/phx.gen.auth.ex#L155-L157) when running in the same beam instance as other tests that do load these libraries.
    ```elixir
    unless Code.ensure_loaded?(Ecto.Adapters.SQL) do
      raise_with_help("mix phx.gen.auth requires ecto_sql", :phx_generator_args)
    end
    ```
1. I believe there's several places where I've seen settings from phoenix itself affect how the generated app runs, and had to keep that in mind or work around it when going through the tests.

Overall, I think we'll run into less issues in the long run if we run these tasks as a separate beam instance. I think we'd get the most performance improvements is to come up with a caching strategy for deps and the compiled deps, such that each dependency only has to be compiled once for the entire test run (instead of once per app). I've made some attempts at this with phx_gen_auth but still haven't come up with an ideal solution.

integration_test/support/code_generator_case.exs
aaronrenner: I think the current implementation doesn't need me to set the `Mix.shell` since we're running separate beam instances. If we decide to go that route, this can be removed.

.github/workflows/ci.yml
```diff
@@ -13,7 +13,7 @@ jobs:
       matrix:
         include:
           - elixir: 1.9.4
-            otp: 20.3.8.26
+            otp: 21.2.7
```
aaronrenner, line 16: Since https://github.com/phoenixframework/phoenix/issues/3969 was closed, I should probably remove this change. However, the integration tests won't pass on this version of OTP, so do you mind if the integration tests for elixir 1.9.4 run on a different version of OTP than the unit tests?

integration_test/support/mix_task_runner.exs
josevalim: Thank you for the reply. I agree with your points then! One last question then: is it possible for us to use `System.cmd` and abandon the prompts? Instead we can use flags such as `--force` to always run particular commands in certain ways. WDYT?

Also note that, once this is in, we should consider removing the `phx.new_test` integration tests that we have today, once we have fully ported them.

.github/workflows/ci.yml
josevalim: It is fine, yes. :)

integration_test/support/mix_task_runner.exs
aaronrenner: At the moment, the need to respond "No" when generating a new project and asked if we want to fetch and install dependencies. Are you thinking we'd add a `mix phx.new --force` option to not install dependencies?

https://github.com/phoenixframework/phoenix/blob/0c045af1a86b24cfa02187abdeeb58da6cd6dcaf/installer/lib/mix/tasks/phx.new.ex#L150-L154

Here are the other places where the generators ask for responses:


_Prompts if the user wants to overwrite the directory_
https://github.com/phoenixframework/phoenix/blob/0c045af1a86b24cfa02187abdeeb58da6cd6dcaf/installer/lib/mix/tasks/phx.new.ex#L346-L350

_Asks the user if they want to proceed when generating into an existing context_
https://github.com/phoenixframework/phoenix/blob/0c045af1a86b24cfa02187abdeeb58da6cd6dcaf/lib/mix/tasks/phx.gen.context.ex#L257-L283

_Prompts the user if they want to overwrite existing files_
https://github.com/phoenixframework/phoenix/blob/0c045af1a86b24cfa02187abdeeb58da6cd6dcaf/lib/mix/phoenix.ex#L318-L342

I do have a test in phx_gen_auth that generates into an existing context and has to respond yes to a prompt: https://github.com/aaronrenner/phx_gen_auth/blob/1a8e74deb4eff273dc60b16b64dbee1ad701d91e/test/mix/tasks/phx_gen_auth/integration_tests/default_app_test.exs#L129. This module is also nice because it will raise if it encounters an unexpected prompt instead of the test just hanging.

I'm completely happy to remove this module if you think there's a better way to do this, but we'll have to come up with a way to bypass these prompts.

josevalim: The first one can be solved with `--skip-dependencies` that we will add. I would solve all other prompts with `--force`. :)

aaronrenner: That simplifies things a lot. :smile:  I'll submit a PR for that.

josevalim: Sounds great! Now that I think about it, I would probably call the flag `--install` and declare it as a boolean, so we get `--no-install` for free. :) Also please directly ping me once you submit the new PR so I take a look at the code, as I don't receive notifications for this repo.

integration_test/support/mix_task_runner.exs
```diff
@@ -0,0 +1,45 @@
+defmodule Phoenix.Integration.MixTaskRunner do
```
josevalim, line 1: It probably doesn't need to be a separate module anymore and we can move this helper function into code_generator_case?

josevalim (commented):
Just one extra comment and we are ready to ship it!

aaronrenner:
Updated! I think it should be ready to go!

josevalim:

josevalim:
Awesome! I think the next step is to finish porting the existingphx.new_test so we can remove some cruft :)

PR #3973, opened 2020-08-25
Fix Phoenix.Router.Route.build/13 spec

michallepicki:
`pipe_through` is a list of atoms, as noted in the [module docs](https://github.com/phoenixframework/phoenix/blob/60b87846f3dde0800f93c7db375c356d6dbdc6b9/lib/phoenix/router/route.ex#L22)

### DIFF
--- a/lib/phoenix/router/route.ex
+++ b/lib/phoenix/router/route.ex
@@ -46,7 +46,7 @@ defmodule Phoenix.Router.Route do
   Receives the verb, path, plug, options and helper
   and returns a `Phoenix.Router.Route` struct.
   """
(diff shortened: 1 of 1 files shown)


auvipy (commented):
you might need to update tests as well?

michallepicki:
Hi @auvipy 
I checked, [RouteTest](https://github.com/phoenixframework/phoenix/blob/a041089f41013fff17e3bed04f4eef24130c2a6c/test/phoenix/router/route_test.exs#L14) is passing and is already correctly passing `pipe_through` as a list of atoms. I think I don't need to do any other changes since my PR only fixes the `@spec`, not the library code.

josevalim:

PR #3978, opened 2020-08-29
use "/" to detect invalid hosts instead of ":"

### LINKED ISSUE #3975: Endpoint url.host warning logic flawed for ipv6 addresses
# Environment

* Elixir version (elixir -v): 1.10.4
* Phoenix version (mix deps): 1.5.4
* NodeJS version (node -v): v14.8.0
* NPM version (npm -v): 6.14.7
* Operating system: macOS

### Expected behavior

I should be able to use ipv6 addresses in urls without the framework warn me.

### Actual behavior

In https://github.com/phoenixframework/phoenix_live_view/issues/911 it is assumed that `:` can't be present in a host and https://github.com/phoenixframework/phoenix/pull/3863 implements the display of a warning when `:` is present.
This makes the framework warn with a config such as
```
config :hello, HelloWeb.Endpoint,
  http: [ip: {0, 0, 0, 0, 0, 0, 0, 1}, port: 4000],
  url: [host: "::1", port: 4000]
```

Even though having `:` is valid in hosts using ipv6 addresses, for instance, here localhost would look like `http://[::1]:4000`.
This SO answer provides references: https://serverfault.com/a/205794

What happens in effect is:
```
mix phx.server
[info] Running HelloWeb.Endpoint with cowboy 2.8.0 at ::1:4000 (http)
[warn] url: [host: ...] configuration value "::1" for HelloWeb.Endpoint is invalid
[warn] url: [host: ...] configuration value "::1" for HelloWeb.Endpoint is invalid
[warn] url: [host: ...] configuration value "::1" for HelloWeb.Endpoint is invalid
[info] Access HelloWeb.Endpoint at http://[::1]:4000
```

robotvert:
`":"` is a valid character if the host is an IPv6 address.

edit: this assumes the http protocol, we don't need to handle `file://`, do we?

Fixes #3975

### DIFF
--- a/lib/phoenix/endpoint/supervisor.ex
+++ b/lib/phoenix/endpoint/supervisor.ex
@@ -302,7 +302,7 @@ defmodule Phoenix.Endpoint.Supervisor do
     host   = host_to_binary(url[:host] || "localhost")
     port   = port_to_integer(url[:port] || port)
 
-    if host =~ "/" do
+    if host =~ ~r"[^:]:\d" do
       Logger.warn("url: [host: ...] configuration value #{inspect(host)} for #{inspect(endpoint)} is invalid")
     end
 
--- a/lib/phoenix/endpoint/supervisor.ex
+++ b/lib/phoenix/endpoint/supervisor.ex
@@ -302,7 +302,7 @@ defmodule Phoenix.Endpoint.Supervisor do
     host   = host_to_binary(url[:host] || "localhost")
     port   = port_to_integer(url[:port] || port)
 
-    if host =~ ":" do
+    if host =~ "/" do
       Logger.warn("url: [host: ...] configuration value #{inspect(host)} for #{inspect(endpoint)} is invalid")
     end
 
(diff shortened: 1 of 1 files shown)


josevalim:
@robotvert unfortunately the issue was using "host:port". Maybe we can have a more complex regex that checks if `:` is followed by a number? Would that rule out ipv6 cases?

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -302,7 +302,7 @@ defmodule Phoenix.Endpoint.Supervisor do
     host   = host_to_binary(url[:host] || "localhost")
     port   = port_to_integer(url[:port] || port)
 
-    if host =~ ":" do
+    if host =~ "/" do
```
josevalim, line 305: WDYT about this then?

```suggestion
    if host =~ ~r"[^:]:\d" do
```

josevalim:
I saw the original issue so I understand better your use case. Fix pushed.

PR #3979, opened 2020-08-29
Don't create compile-time dependencies for action_fallback

michalmuskala:
The plug is not used at compile-time, it's safe not to create the compile-time dependency. This should break a lot of compile dependencies in phoenix applications using `action_fallback` - as a "central" component it tends to accumulate a lot of dependencies itself.


josevalim:

PR #3982, opened 2020-09-01
Checks origins host names case insensitive

sofacoder:
URLs schemes and hosts are case insensitive and should checked that way.

### DIFF
--- a/lib/phoenix/socket/transport.ex
+++ b/lib/phoenix/socket/transport.ex
@@ -614,9 +614,9 @@ defmodule Phoenix.Socket.Transport do
   defp compare_host?(_request_host, nil),
     do: true
   defp compare_host?(request_host, "*." <> allowed_host),
-    do: String.ends_with?(request_host, allowed_host)
+    do: String.ends_with?(String.downcase(request_host), String.downcase(allowed_host))
   defp compare_host?(request_host, allowed_host),
-    do: request_host == allowed_host
+    do: String.downcase(request_host) == String.downcase(allowed_host)
 
   # TODO: Deprecate {:system, env_var} once we require Elixir v1.9+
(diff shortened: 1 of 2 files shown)


chrismccord:
This should already be normalized at the webserver layer so shouldn't be necessary. If you are aware of an http client where this is causing an issue, please let us know. Thanks!

sofacoder:
Well, if I take https://github.com/chrismccord/phoenix_chat_example an send request for example using https://github.com/davidstump/SwiftPhoenixClient using `let socket = Socket("ws://LOCALHOST:4000/socket/websocket")` that will result it phoenix logging this:
```
[error] Could not check origin for Phoenix.Socket transport.

This happens when you are attempting a socket connection to
a different host than the one configured in your config/
files. For example, in development the host is configured
to "localhost" but you may be trying to access it from
"127.0.0.1". To fix this issue, you may either:

  1. update [url: [host: ...]] to your actual host in the
     config file for your current environment (recommended)

  2. pass the :check_origin option when configuring your
     endpoint or when configuring the transport in your
     UserSocket module, explicitly outlining which origins
     are allowed:

        check_origin: ["https://example.com",
                       "//another.com:888", "//other.com"]
```

Of course one could argue that this client already should fix that. But while [URI RFC](https://tools.ietf.org/html/rfc3986) recommends lower case it also states that applications should accept uppercase letters for scheme and host. So IMHO it would be just defensive to handle that issue on the server.
We run into that with a real world application, not saying it would have been avoidable, but that fix would have saved us a few grey hairs without breaking anything in phoenix (Thats a guess, so maybe you see technical issue with it that I don't see).

sofacoder:
@chrismccord You are keep your decision?

PR #3989, opened 2020-09-10
replace remaining Mix.Config usage

dvic:
As per https://github.com/phoenixframework/phoenix/commit/cbfc71c7fcd9c5b95ac5ed63719ac88a4927af85, `use Mix.Config` can be replaced with `import Config`.


josevalim:

PR #3990, opened 2020-09-13
Add --colors option in watchers for colorized Webpack output

altdsoy:
Added the --colors option for the Endpoint config in dev.exs
Having a colorized output is a little prettier.
Changes added for both regular and umbrella app generation.


chrismccord:
We used to ship with this by default but it caused issues, maybe on windows? If you can look thru the issues and track down what the problems were and see if they still exist, I would love to add it back. Thanks!

PR #3991, opened 2020-09-14
Use 127.0.0.1 in dev and test

chulkilee:
When `:ip` is not given, phoenix is listening on all interfaces. This is not desired for dev/test mode, where it's unlikely to expose servers to public interfaces.

Add `:ip` option with ipv4 loopback (`127.0.0.1`) to all generated dev/test configuration.

(kind of related: #3400)

### DIFF
--- a/installer/templates/phx_single/config/dev.exs
+++ b/installer/templates/phx_single/config/dev.exs
@@ -7,7 +7,9 @@ import Config
 # watchers to your application. For example, we use it
 # with webpack to recompile .js and .css sources.
 config :<%= app_name %>, <%= endpoint_module %>,
-  http: [port: 4000],
+  # Binding to loopback ipv4 address prevents access from other machines.
+  # Change to `ip: {0, 0, 0, 0}` to allow access from other machines.
+  http: [ip: {127, 0, 0, 1}, port: 4000],
   debug_errors: true,
   code_reloader: true,
   check_origin: false,
--- a/installer/templates/phx_single/config/test.exs
+++ b/installer/templates/phx_single/config/test.exs
@@ -3,7 +3,7 @@ import Config
 # We don't run a server during test. If one is required,
 # you can enable the server option below.
 config :<%= app_name %>, <%= endpoint_module %>,
-  http: [port: 4002],
+  http: [ip: {127, 0, 0, 1}, port: 4002],
   server: false
 
 # Print only warnings and errors during test
--- a/installer/templates/phx_umbrella/apps/app_name_web/config/dev.exs
+++ b/installer/templates/phx_umbrella/apps/app_name_web/config/dev.exs
@@ -5,7 +5,9 @@
 # watchers to your application. For example, we use it
 # with webpack to recompile .js and .css sources.
 config :<%= web_app_name %>, <%= endpoint_module %>,
-  http: [port: 4000],
+  # Binding to loopback ipv4 address prevents access from other machines.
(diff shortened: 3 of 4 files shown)


nathanl:
I think this is a good idea - I accidentally created a dup in https://github.com/phoenixframework/phoenix/pull/4010 but closed it after I saw this PR.

I think some explanatory comments would be helpful - see my PR and feel free to use what I wrote, or I can PR some comments afterward if this PR is merged.

chulkilee:
@nathanl thanks a lot! I copied some from your PR #4010 . Please feel free to make any suggestion on the help text.

chrismccord:
Looks like #3400 covers this. We like the idea, but need to sanity check a few things around windows iirc. If we go this direction, it will land along with 1.5. Thanks!!

chulkilee:
@chrismccord #3400 *does not* cover this, since it only updates `installer/templates/phx_single/config/prod.secret.exs`  while this PR is only to update  *dev* and *test*.

Can we reopen this ticket for dev/test change?

chulkilee:
I also confirmed it works on Windows with the follow step:

- Downloaded virtualbox vm "MSEdge on Win10 (x64) Stable 1809" from https://developer.microsoft.com/en-us/microsoft-edge/tools/vms/
- Installed Elixir and Erlang with the installer from https://elixir-lang.org/install.html#windows
  - Elixir 1.10.4, Erlang 21.3 (default at this point)
- Added `http: [ip: { 127, 0, 0, 1 }, port: 4000]` to `dev.exs`
- Started the server with `mix phx.server` => it showed `localhost:4000`  instead of `0.0.0.0:4000`
  - Also it didn't ask for the firewall (it happened when I dropped the config to use `0.0.0.0:4000`)
- Opened http://127.0.0.1:4000 from IE => it worked!

installer/templates/phx_single/config/dev.exs
```diff
@@ -7,7 +7,9 @@ import Config
 # watchers to your application. For example, we use it
 # with webpack to recompile .js and .css sources.
 config :<%= app_name %>, <%= endpoint_module %>,
-  http: [port: 4000],
+  # Binding to loopback ipv4 address prevents access from other machines.
+  # Change to `ip: {0, 0, 0, 0}` to allow access from other machines.
```
nathanl, line 11: After "to allow access from other machines", I would add "or from the host machine if Phoenix is running in a VM or container""

nathanl:
@chrismccord Just seconding that this is not the same as #3400 - that's about prod, this is about dev and test.

installer/templates/phx_single/config/dev.exs
chulkilee: Running in vm / container does not necessarily requires binding on public interface - e.g. users may run proxy on the machine - in that case it runs in container/vm but does require that... but I'd like to add that as an "example" (not another case with "or")  :heart:.

Could you help improve the wording?

> Change to `ip: {0, 0, 0, 0}` to allow access from other machines,
> such as accessing the app runnining in a virtual machine or container
> from the host machine if Phoenix is running in a VM or container

Also.. I'm not sure "Phoenix is running" is right statement. Do we have similar comments somewhere? We're running this app, not "Phoenix".

chulkilee:
@chrismccord @josevalim could you reopen and review this PR? As @nathanl said, this PR is different from #3400 - and actually #3993 (with IP address tuple) was merged by @josevalim .

I rebased the branch to fix conflict. If needed I'll create new PR - let me know!

PR #3993, opened 2020-09-14
Use ip option instead of transport socket_opts

chulkilee:
#3551 converted the config from list to keyword by moving the first `inet` / `inet6` option under `transport_options[:socket_opts]`.

All options is down to [ranch_top](https://ninenines.eu/docs/en/ranch/1.7/manual/ranch_tcp/)

This can be simplified by using `:ip` option  actually, since ranch seems to pick up right socket based on the `:ip` option.

Also it could be simpler to use, since if a user want to bind to specific ip address by changing only `:ip` option, not two places to make sure use right `socket opts`.

## Docs

[ranch_tcp (1.7)](https://ninenines.eu/docs/en/ranch/1.7/manual/ranch_tcp/)

> inet
>     Set up the socket for IPv4.
> inet6
>     Set up the socket for IPv6.
> ip
>     Interface to listen on. Listen on all interfaces by default.

See also [ranch (1.7)](https://ninenines.eu/docs/en/ranch/1.7/manual/ranch/)

> Transport options.
>
> socket_opts
>     Socket options given to Transport:listen/1. Please refer to the documentation of the transport module you are using for more details.

## Example

```elixir
[http: [port: 4000]]
[http: [:inet, port: 4000]]
[http: [port: 4000, ip: {0, 0, 0, 0}]]
[http: [port: 4000, transport_options: [socket_opts: [:inet]]]]
# => 0.0.0.0:4000

[http: [ip: {127, 0, 0, 1}, port: 4000]]
# => 127.0.0.1:4000

[http: [:inet6, port: 4000]]
[http: [port: 4000, ip: {0, 0, 0, 0, 0, 0, 0, 0}]]
[http: [port: 4000, transport_options: [socket_opts: [:inet6]]]]
# => :::4000

http: [ip: {0, 0, 0, 0, 0, 0, 0, 1}, port: 4000],
# => ::1:4000
```

### DIFF
--- a/installer/templates/phx_single/config/prod.secret.exs
+++ b/installer/templates/phx_single/config/prod.secret.exs
@@ -13,8 +13,8 @@ secret_key_base =
 
 config :<%= app_name %>, <%= endpoint_module %>,
   http: [
-    port: String.to_integer(System.get_env("PORT") || "4000"),
-    transport_options: [socket_opts: [:inet6]]
+    ip: {0, 0, 0, 0, 0, 0, 0, 0},
+    port: String.to_integer(System.get_env("PORT") || "4000")
   ],
   secret_key_base: secret_key_base
 
--- a/installer/templates/phx_umbrella/apps/app_name_web/config/prod.secret.exs
+++ b/installer/templates/phx_umbrella/apps/app_name_web/config/prod.secret.exs
@@ -7,8 +7,8 @@ secret_key_base =
 
 config :<%= web_app_name %>, <%= endpoint_module %>,
   http: [
-    port: String.to_integer(System.get_env("PORT") || "4000"),
-    transport_options: [socket_opts: [:inet6]]
+    ip: {0, 0, 0, 0, 0, 0, 0, 0},
+    port: String.to_integer(System.get_env("PORT") || "4000")
   ],
   secret_key_base: secret_key_base
 
--- a/installer/templates/phx_single/config/prod.exs
+++ b/installer/templates/phx_single/config/prod.exs
@@ -25,11 +25,11 @@ config :logger, level: :info
 #       ...
 #       url: [host: "example.com", port: 443],
 #       https: [
+#         ip: {0, 0, 0, 0, 0, 0, 0, 0},
 #         port: 443,
 #         cipher_suite: :strong,
 #         keyfile: System.get_env("SOME_APP_SSL_KEY_PATH"),
-#         certfile: System.get_env("SOME_APP_SSL_CERT_PATH"),
-#         transport_options: [socket_opts: [:inet6]]
+#         certfile: System.get_env("SOME_APP_SSL_CERT_PATH")
 #       ]
 #
 # The `cipher_suite` is set to `:strong` to support only the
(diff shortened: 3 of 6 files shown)


josevalim:
I understand it is less places to change but I am afraid `[:inet6]` may be a bit more clearer than a 8 element tuple. Also, people may want to just set inet6 without binding to a particular ip? 🤔

josevalim:
And there is also the question if 0.0.0.0 will work on OSes like Windows or if it will require special permissions.

chulkilee:
> people may want to just set inet6 without binding to a particular ip?
> And there is also the question if 0.0.0.0 will work on OSes like Windows or if it will require special permissions.

`:inet6` is by default binding to all interfaces, so that's same with `ip: {0, 0, 0, 0, 0, 0, 0, 0}` - at least from log output.

Also those ips (non-routable daddress) are pretty de-fecto (and maybe standard?).. so I guess erlang should handle them for different platforms transparently... but I haven't tested. 

>  I am afraid [:inet6] may be a bit more clearer than a 8 element tuple

Although `:inet6` (vs `:inet`) is the standard way to specify address type in Erlang.. this can be one of confusing part for new elixir developer, since it's not documented anywhere in phoenix actually. However ip with 8 elements (or 4 elements) are more easy to guess (or search) since it is under `ip` key.

Also please note that you may have conflicting config in several places - see the following example, where there are three places indicating ipv4 or ipv6.

```
http: [:inet6, ip: { 0, 0, 0, 0 }, transport_options: [socket_opts: [:inet]]]
```

For the first-time phoenix developer, I think it's better to be super clear - instead of making them diving into other library/erlang *too deep*. I don't want to scare them ;)

installer/templates/phx_single/config/prod.secret.exs
```diff
@@ -13,8 +13,8 @@ secret_key_base =
 
 config :<%= app_name %>, <%= endpoint_module %>,
   http: [
-    port: String.to_integer(System.get_env("PORT") || "4000"),
-    transport_options: [socket_opts: [:inet6]]
+    ip: {0, 0, 0, 0, 0, 0, 0, 0},
```
josevalim, line 16:
```suggestion
    # Enable IPv6 and bind on all addresses
    ip: {0, 0, 0, 0, 0, 0, 0, 0},
```

installer/templates/phx_umbrella/apps/app_name_web/config/prod.secret.exs
```diff
@@ -7,8 +7,8 @@ secret_key_base =
 
 config :<%= web_app_name %>, <%= endpoint_module %>,
   http: [
-    port: String.to_integer(System.get_env("PORT") || "4000"),
-    transport_options: [socket_opts: [:inet6]]
+    ip: {0, 0, 0, 0, 0, 0, 0, 0},
```
josevalim, line 10:
```suggestion
    # Enable IPv6 and bind on all addresses
    ip: {0, 0, 0, 0, 0, 0, 0, 0},
```

josevalim:

nathanl:
@chulkilee Thanks for making this explicit. FYI https://github.com/phoenixframework/phoenix/pull/4127 builds on this work.

PR #3998, opened 2020-09-21
Add --install and --no-install options to phx.new

aaronrenner:
These new options allow the user to skip the "Fetch and install dependencies?" prompt when generating new projects. This option will be very helpful when generating new projects programatically like during integration testing (#3970) or for phoenixdiff.org.

I was a little unsure how to document this pair of options, so I indented them under a "Skipping prompts" heading.

Screenshot of new phx.new options

/cc @josevalim


installer/lib/mix/tasks/phx.new.ex
```diff
@@ -49,6 +49,9 @@ defmodule Mix.Tasks.Phx.New do
     * `--binary-id` - use `binary_id` as primary key type
       in Ecto schemas
 
+    * `--no-install` - do not fetch and install dependencies
+      while generating this project
```
josevalim, line 53: Given the default is `--install`, we only document the non-default behaviour. :)

josevalim: Oh, i see though. You are making it so it doesn't even ask. Ignore me, i will add it back.

josevalim:

aaronrenner:
@josevalim Would you mind back-porting this to phoenix 1.5? I'd like to be able to use this in the phx_gen_auth tests without having to run on phoenix master.

aaronrenner:
Thank you! :heart:

PR #4003, opened 2020-09-24
Replace obsolete doc badge by hex badges

sabiwara:
Hi :wave:
It seems that the link for the documentation badge is dead, what do you with of using the regular hex badges?


sabiwara:
Oh sorry I thought it was a dead link but it seems to be working... Closing, apologies!

PR #4005, opened 2020-09-28
Updated Dockerfile in releases.md

savtrip:
Added `deps.get --only $MIX_ENV` to use the environment variable. This ensures mix only pulls the dependencies necessary for the type of mix environment set within the dockerfile.


josevalim:

PR #4007, opened 2020-09-28
fix generators for big argument list >50

sulphur:
Hello, 
I've stumbled on an issue where generated test cases weren't valid because  `IO.inspect` and `Kernel.inspect` by default inspect only 50 elements in maps. Therefore
generating a schema for example :

`mix phx.gen.live NewPerformers Performer performers attr1 ... attr55`

will result in invalid maps in test cases where `@create_attrs` etc. will be truncated by `inspect`
with `...}` at the end
like this:

```%{attr1: "some value", ..., attr50: "some value, ...}```

It is my first look in the Phoenix code base so my fix is rather naive. I think that I might be fixed in a better way like maybe writing a helper or something. But for now, I have no better idea ;) 

Cheers
Piotr

### DIFF
--- a/lib/mix/phoenix.ex
+++ b/lib/mix/phoenix.ex
@@ -351,4 +351,8 @@ defmodule Mix.Phoenix do
       Module.concat(["#{base}Web"])
     end
   end
+
+  def to_text(data) do
+    inspect data, limit: :infinity, printable_limit: :infinity
+  end
 end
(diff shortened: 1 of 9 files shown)


josevalim:
This looks great @sulphur! I wonder if we should encapsulate this in its own function and simply call it from the template. This way, if we need to change other parameters in the future, we have to change only one place. WDYT? Can you also please rebase the PR? Thanks!

sulphur:
Hi, I've extracted to a helper function that i've put into `Mix.Phoenix` module (not sure whether it is a good place) and i've rebased to latest `master` branch.

josevalim:

PR #4008, opened 2020-09-30
Add reference for Model View Controller

Cyber28:
Since this page is meant to serve as an introduction to the framework, it may be worth including further reference on the MVC pattern as many beginners may not be familiar with it.


josevalim:
Thanks @Cyber28! Unfortunately, the (client) MVC referenced in the wikipedia article is not quite the same as server-side MVC. If there is a more appropriate reference, we would love to use it though!

PR #4009, opened 2020-10-01
Add binary serialization support to default serializers

chrismccord:
Initial pass. We need to decide how we want to support the caller push and receive of binary data. For example, we can support `push("foo", <<>>)` or we can require an explicit `push("foo", {:binary, "..."})` since we can't really distinguish between someone pushing a string vs binary and as of now we require maps. Likewise, on the `handle_in` side, either `handle_in("foo", <<>>, socket)` or `handle_in("foo", {:binary, <<>>}, socket)`

### DIFF
--- a/lib/phoenix/socket/serializers/v2_json_serializer.ex
+++ b/lib/phoenix/socket/serializers/v2_json_serializer.ex
@@ -2,15 +2,62 @@ defmodule Phoenix.Socket.V2.JSONSerializer do
   @moduledoc false
   @behaviour Phoenix.Socket.Serializer
 
+  @push 0
+  @reply 1
+  @broadcast 2
+
   alias Phoenix.Socket.{Broadcast, Message, Reply}
 
   @impl true
-  def fastlane!(%Broadcast{} = msg) do
+  def fastlane!(%Broadcast{payload: {:binary, data}} = msg) do
+    topic_size = byte_size!(msg.topic, :topic, 255)
+    event_size = byte_size!(msg.event, :event, 255)
+
+    bin = <<
+      @broadcast::size(8),
+      topic_size::size(8),
+      event_size::size(8),
+      msg.topic::binary-size(topic_size),
+      msg.event::binary-size(event_size),
+      data::binary
+    >>
+
+    {:socket_push, :binary, bin}
+  end
+
+  def fastlane!(%Broadcast{payload: %{}} = msg) do
     data = Phoenix.json_library().encode_to_iodata!([nil, nil, msg.topic, msg.event, msg.payload])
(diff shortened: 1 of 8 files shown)


shoop2019:
Merge

shoop2019:
Cool

lib/phoenix/socket/serializers/v2_json_serializer.ex
```diff
@@ -2,15 +2,58 @@ defmodule Phoenix.Socket.V2.JSONSerializer do
...
+      event_size::size(8),
+      msg.topic::binary-size(topic_size),
+      msg.event::binary-size(event_size),
+      data::binary
+    >>
+
+    {:socket_push, :binary, bin}
+  end
+
   def fastlane!(%Broadcast{} = msg) do
```
josevalim, line 28: we should probably validate that the payload is a `map` and print a nice error message if it isn't (now that we removed the guard from the channel funs).

lib/phoenix/socket/serializers/v2_json_serializer.ex
```diff
@@ -23,13 +66,39 @@ defmodule Phoenix.Socket.V2.JSONSerializer do
     {:socket_push, :text, Phoenix.json_library().encode_to_iodata!(data)}
   end
 
+  def encode!(%Message{payload: {:binary, data}} = msg) do
+    join_ref_size = byte_size(msg.join_ref)
+    topic_size = byte_size(msg.topic)
+    event_size = byte_size(msg.event)
```
josevalim, line 72: Worth raising if we are ever over the byte_size.

josevalim:
Looks great! We should look into adding an integration test too in websocket_channels_test just to make sure the whole experience works.

chrismccord:
Addressed all feedback. I'll push docs next and I believe this will be good to go. Thanks!

lib/phoenix/socket/serializers/v2_json_serializer.ex
```diff
@@ -10,8 +10,8 @@ defmodule Phoenix.Socket.V2.JSONSerializer do
 
   @impl true
   def fastlane!(%Broadcast{payload: {:binary, data}} = msg) do
-    topic_size = byte_size(msg.topic)
-    event_size = byte_size(msg.event)
+    topic_size = byte_size!(msg.topic, :topic, 256)
```
Gazler, line 13: Shouldn't the max here be 255?

chrismccord: `byte_size!` does `< max`, so we are enforcing 255, but I think 255 with `<=` would be clearer?

PR #4010, opened 2020-10-01
Bind to localhost by default in dev

nathanl:
For comparison, binding to localhost by default in development has been Rails' behavior since 2013 - https://guides.rubyonrails.org/4_2_release_notes.html#default-host-for-rails-server

### DIFF
--- a/installer/templates/phx_single/config/prod.exs
+++ b/installer/templates/phx_single/config/prod.exs
@@ -25,12 +25,16 @@ config :logger, level: :info
 #       ...
 #       url: [host: "example.com", port: 443],
 #       https: [
+#         ip: {0, 0, 0, 0},
 #         port: 443,
 #         cipher_suite: :strong,
 #         keyfile: System.get_env("SOME_APP_SSL_KEY_PATH"),
 #         certfile: System.get_env("SOME_APP_SSL_CERT_PATH"),
 #         transport_options: [socket_opts: [:inet6]]
 #       ]
+#
+# If you intend to run Phoenix behind a reverse proxy, consider
+# changing `ip` to `{127, 0, 0, 1}` to prevent direct access.
 #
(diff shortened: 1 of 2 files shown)


installer/templates/phx_single/config/prod.exs
```diff
@@ -26,12 +26,16 @@ config :logger, level: :info
 #       url: [host: "example.com", port: 443],
 #       https: [
 #         port: 443,
+#         ip: {0, 0, 0, 0},
```
nathanl, line 29: I think this is Cowboy's default, but it seems good to be explicit, if nothing else to show that we have the opportunity to specify this.

nathanl:
Oh whoops, I just noticed that https://github.com/phoenixframework/phoenix/pull/3991 already exists. If you merge that, I'd be happy to add comments like these afterward.

installer/templates/phx_single/config/prod.exs
```diff
@@ -25,13 +25,17 @@ config :logger, level: :info
...
 #         port: 443,
 #         cipher_suite: :strong,
 #         keyfile: System.get_env("SOME_APP_SSL_KEY_PATH"),
 #         certfile: System.get_env("SOME_APP_SSL_CERT_PATH"),
 #         transport_options: [socket_opts: [:inet6]]
 #       ]
 #
+# If you intend to run Phoenix behind a reverse proxy, consider
+# changing `ip` to `{127, 0, 0, 1}` to prevent direct access.
+#
```
chulkilee, line 38: FYI this is correct only if phoenix and the reverse proxy is on the same machine.

It's pretty common to have reverse proxy in different machine - then it should bind to public interface.

nathanl: Yes, good clarification.

PR #4012, opened 2020-10-05
Cache compiled deps to improve integration tests speed

aaronrenner:
This adds caching to the integration test suite to improve the test suite speed from ~80 seconds to ~15 seconds when fully cached. Here is the approach this takes:

1. Dependencies are downloaded each time using `mix deps.get`. This generates the project's `mix.lock` file and ensures we get the newest versions of the dependencies each test run.
2. A hash is generated from the `mix.lock` file and used as the cache key when compiling or restoring dependencies. Projects with the same `mix.lock` file will use the same precompiled dependencies. There was an issue where projects with the same `mix.lock` hash would both receive a cache miss and start compiling the same dependencies in parallel. To fix this issue, this PR uses a GenServer per cache key to ensure multiple projects won't compile the same dependencies at the same time.

 ## Other notes

1. Since the build times when everything is cached are significantly better, this also adds caching to the Github build. I added a step to generate a new timestamp to be used in the cache key, which when combined with the `restore-keys` option makes it so each test run has the cache from the previous test run.
2. I added `integration_test/support` to the build path when running with `TEST_SUITE=integration`. I chose this route because it removed a bunch of `Code.require_file/2` calls I'd have to write.

I'd definitely appreciate any feedback on how things can be improved here. I'm thinking once we have decent caching with the integration tests, we can migrate tests out of `phx_new_test.exs` into here and start adding more integration coverage.

/cc @josevalim

### DIFF
--- a/integration_test/support/deps_compiler/cache_server.ex
+++ b/integration_test/support/deps_compiler/cache_server.ex
@@ -1,0 +1,66 @@
+defmodule Phoenix.Integration.DepsCompiler.CacheServer do
+  @moduledoc false
+
+  alias Phoenix.Integration.CodeGeneratorCase
+
+  use GenServer
+
+  def start_link(opts) when is_list(opts) do
+    {start_opts, server_opts} = Keyword.split(opts, [:name])
+
+    hash = Keyword.fetch!(server_opts, :hash)
+    cache_root_path = Keyword.fetch!(server_opts, :cache_root_path)
+
+    GenServer.start_link(__MODULE__, %{hash: hash, cache_root_path: cache_root_path}, start_opts)
+  end
+
+  def compile_or_restore(server, app_root_path, timeout \\ 90_000) do
+    case GenServer.call(server, {:maybe_compile, app_root_path}, timeout) do
+      :compiled ->
+        :ok
+
+      {:cache_hit, cache_path} ->
+        File.cp_r!(
+          cache_path,
+          app_root_path
+        )
+
+        File.cp_r!(
+          Path.join([app_root_path, "_build", "dev"]),
+          Path.join([app_root_path, "_build", "test"])
+        )
+
+        :ok
+    end
+  end
+
+  @impl true
+  def init(state), do: {:ok, state}
+
+  @impl true
+  def handle_call({:maybe_compile, app_root_path}, _from, state) do
+    build_cache_with_hash_path = Path.join([state.cache_root_path, "build", state.hash])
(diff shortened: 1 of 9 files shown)


aaronrenner:
Looks like something is a little flakey during the build. I'll need to investigate more.

integration_test/support/deps_compiler/cache_server.ex
```diff
@@ -0,0 +1,66 @@
...
+      # rebar compiled dependencies. Rebar compiles its dependencies in deps/ and
+      # symlinks them into _build/. This means if we restored the cache to a new project
+      # and left the symlinks to deps in tact, the rebar-based dependencies would have to be
+      # recompiled because their compiled code would be missing.
+      {_output, 0} =
+        System.cmd("cp", [
+          "-rL",
+          Path.join(app_root_path, "_build"),
+          build_cache_with_hash_path
+        ])
```
josevalim, line 61: I am afraid this won't work on Windows?

aaronrenner: Ah, good point. Do you happen to know where rebar compiles code on windows (`deps/` + symlink, or `_build/`)? Is there a better way to work around this issue?

josevalim:
Hi @aaronrenner! To be honest, I am not super happy with calling `mix deps.get` every time because if I go to the past and re-run the tests, they can now fail as they are no longer consistent.

What do you think about something like this:

1. Define a mix.exs file for the integration tests that has *all of the dependencies* that we are going to use across all builds. This mix.exs could be the `mix.exs` of the integration tests themselves. I.e. we could treat the `integration/` folder as a separate application that we run `mix test` inside of.

2. For each test, we simply copy the whole `_build` and `deps` directory to the generated project. Each child project may not use all deps, but that's fine. If some lock is outdated (because it no longer resolves correctly due to an update mix.exs from Phoenix generators), we raise.

This should be even faster, it requires running `mix deps.get` once, and it allows us to version the lock and have reproducible builds.

integration_test/support/deps_compiler/cache_server.ex
josevalim: We do this on Mix: https://github.com/elixir-lang/elixir/blob/master/lib/mix/lib/mix/utils.ex#L424 - but that means you would have to traverse the directories yourself. If rebar is linking the whole `deps/foo`, then this is easy to check and do yourself though.

PR #4013, opened 2020-10-05
Add support for Mix Releases to phx.new

akoutmos:
While this PR is not 100% ready (tests need to be added/updated, template white spacing needs to be cleaned up, and docs need to be updated), I wanted to gauge interest in adding a new CLI flag to `mix phx.new`. This new CLI flag would give the user the option to generate a new Phoenix project that leverages Mix Releases as opposed to the default `prod.secret.exs` strategy.

Sample usage would be something like:
```
$ mix phx.new my_app --mix-release --live --binary-id
```

I find that with most of the Phoenix projects that I personally start, setting up Mix Releases is usually one of the first things I do. I also find this to be the case in a professional work settings as well regardless of deployment target (Docker/VM).

Curious how other people feel about this proposal as well. If the Phoenix team is `:ok` with adding this new flag to the installer, then I'll go ahead and add the aforementioned missing bits.

Thanks!

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -10,5 +10,6 @@
 /installer/tmp
 /installer/doc
 /installer/deps
+/installer/phx_new-*.ez
 
 erl_crash.dump
--- a/installer/lib/mix/tasks/phx.new.ex
+++ b/installer/lib/mix/tasks/phx.new.ex
@@ -33,6 +33,10 @@ defmodule Mix.Tasks.Phx.New do
       Please check the driver docs for more information
       and requirements. Defaults to "postgres".
 
+    * `--mix-release` - leverage Mix Releases for the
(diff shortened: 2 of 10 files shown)


josevalim:
Hi @akoutmos, unfortunately Elixir v1.11 is a strong requirement for this, so we have a single config/runtime.exs that works for both PaaS deployments and releases. I was also thinking that, if we need to generate something, it could be done as `mix phx.gen.release`, as it doesn't need to be part of a new app. For those reasons,  I think we have to wait a bit for this. Thanks for the PR!

akoutmos:
That sounds good @josevalim. I was on the fence as to whether this belonged in core Phoenix as well (as not to bloat the generator with too many options). Appreciate the feedback!

PR #4014, opened 2020-10-07
Improve integration test speed

aaronrenner:
This improves integration test speed by following @josevalim's suggestions in #4012 (thanks José!). Below is an overview of the approach this takes:

1. Creates a new mix project in `integration_test`. This project contains the list of dependencies used by all phoenix projects, so the dependencies can be locked via `integration_test/mix.lock` and old builds can be re-run even if new versions of dependencies have been uploaded to hex.
2. To improve test speed of the generated projects, `mix.lock`, `deps`, and `_build` are copied from the integration test project to every generated project. Since some commands are run with `MIX_ENV=test` and others are run with `MIX_ENV=dev` by default, this initially compiles the dependencies in the test environment and then copies `_build/test` to `_build/dev`.
3. Since the integration test is a separate project now, `CodeGeneratorCase` is now being compiled at compile time instead of being required at runtime as a `.exs` file.

This brings the actual ex_unit test execution time from 80+ seconds to ~11 seconds.


integration_test/test/test_helper.exs
```diff
@@ -0,0 +1,11 @@
+# Copy _build/test to _build/dev so it only has to be compiled once
+File.rm_rf!(Path.expand("../_build/dev", __DIR__))
+
+File.cp_r!(
+  Path.expand("../_build/test", __DIR__),
+  Path.expand("../_build/dev", __DIR__)
+)
+
+
```
josevalim, line 9:
```suggestion
```

integration_test/test/test_helper.exs
```diff
@@ -0,0 +1,11 @@
...
+ExUnit.configure(max_cases: 2, timeout: 180_000)
```
josevalim, line 10: Won't this be too few for a machine with 16 cores? :)

josevalim:
This looks great! :heart: I have added only two very minor comments. :)

integration_test/test/test_helper.exs
aaronrenner: Good catch. That was a remnant of the previous integration test setup. 😅

I just pushed an update.

josevalim:
Amazing work! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

Just something to verify, if we change the lock on phx.new to something that is no longer compatible in our integration lock, is the error message clear enough? :)

aaronrenner:
@josevalim Good question. I'll check it out.

aaronrenner:
@josevalim I generated an app with `--database=mysql` because its dependencies aren't listed in `integration/mix.exs` and got the following error:

```
  1) test newly generated app has no warnings or errors (Phoenix.Integration.CodeGeneration.AppWithMySQLTest)
     test/code_generation/app_with_mysql_test.exs:4
     ** (RuntimeError) mix command failed with exit code: 1
     
     mix do clean, compile --warnings-as-errors
     
     Unchecked dependencies for environment dev:
     * myxql (Hex package)
       the dependency is not available, run "mix deps.get"
     ** (Mix) Can't continue due to errors on dependencies
     
     
     Options
     cd: /home/aaron/projects/phoenix/installer/tmp/F5UTGQoxTZ/new with mysql/phx_blog
     env: []
     
     code: with_installer_tmp("new with mysql", fn tmp_dir ->
     stacktrace:
       (phoenix_integration 0.1.0) test/support/code_generator_case.ex:35: Phoenix.Integration.CodeGeneratorCase.mix_run!/3
       test/code_generation/app_with_mysql_test.exs:8: anonymous fn/1 in Phoenix.Integration.CodeGeneration.AppWithMySQLTest."test newly generated app has no warnings or errors"/1
       (phoenix_integration 0.1.0) test/support/code_generator_case.ex:70: Phoenix.Integration.CodeGeneratorCase.with_installer_tmp/3
       test/code_generation/app_with_mysql_test.exs:5: (test)



Finished in 3.2 seconds
1 test, 1 failure

Randomized with seed 526362

ExUnit Compilation exited abnormally with code 1 at Wed Oct  7 15:38:25
```

Do you think this is enough information, provided the user has looked at https://github.com/phoenixframework/phoenix/tree/master/integration_test#how-tests-are-written? I can see about wrapping this error if you think that would be better.

josevalim:
I think that's perfect. thank you for double checking! 💯

PR #4015, opened 2020-10-07
Test with elixir 1.11.0 and OTP 23.1.1

aaronrenner:
This upgrades ci to test on Elixir 1.11.0 and OTP 23.1.1.


josevalim:
Oops, I merged another PR just before even though this was first, sorry!

PR #4017, opened 2020-10-08
Fix tests generated by --live generator

feliperenan:
Fixes https://github.com/phoenixframework/phoenix/issues/3947

You can read the motivation of this PR in the issue, but in short, in order to fix all tests those are the changes:

* Set string_attr only when the field is a string 631ebe5
* Ensure empty list for fields with list type 4d533ea
* Use date as map on live_test generator 50afac3

### Testing

In case you want to test, those are the commands I've been using:

```
$ mix phx.gen.live Shoppers Shopper shoppers favorites:array:string birth_date:date name:string created_at:naive_datetime time:time created_at_b:utc_datetime
```

After that, run `mix test` and all tests should be green.


feliperenan:
@josevalim this PR is ready for review 🙌

lib/mix/tasks/phx.gen.html.ex
```diff
@@ -176,13 +176,13 @@ defmodule Mix.Tasks.Phx.Gen.Html do
...
         {label(key), ~s(<%= textarea f, #{inspect(key)} %>), error(key)}
       {key, :date} ->
-        {label(key), ~s(<%= date_select f, #{inspect(key)} %>), error(key)}
+        {label(key), ~s(<%= date_input f, #{inspect(key)} %>), error(key)}
       {key, :time} ->
-        {label(key), ~s(<%= time_select f, #{inspect(key)} %>), error(key)}
+        {label(key), ~s(<%= time_input f, #{inspect(key)} %>), error(key)}
       {key, :utc_datetime} ->
-        {label(key), ~s(<%= datetime_select f, #{inspect(key)} %>), error(key)}
+        {label(key), ~s(<%= datetime_local_input f, #{inspect(key)} %>), error(key)}
```
josevalim, line 183: I gave it a try locally and it seems that `datetime_local_input` is not supported in Firefox nor Safari. Nor `datetime_input`. We may have to use another aproach to solve the date/time/datetime issues, possibly keeping them as `datetime_select` and use maps in tests as you proposed as alternative. Sorry for misleading. :(

priv/templates/phx.gen.live/form_component.html.leex
```diff
@@ -10,5 +10,7 @@
   <%= input %>
   <%= error %>
 <% end %>
-  <%%= submit "Save", phx_disable_with: "Saving..." %>
+  <div>
```
feliperenan, line 13: This is not part of this issue but I'll make the generated form a bit more beautiful.

### before
image


### after 
image

Just copied from here: https://github.com/phoenixframework/phoenix/blob/c8883af5582a38496e4b7e45e05d3a4d759a6caa/priv/templates/phx.gen.html/form.html.eex#L12-L14

lib/mix/tasks/phx.gen.html.ex
feliperenan: Damm browsers supports 😢 

ok, I will make the change for using maps on the tests instead 👍

lib/mix/phoenix/schema.ex
```diff
@@ -175,6 +175,35 @@ defmodule Mix.Phoenix.Schema do
     |> Enum.into(%{}, fn {k, t} -> {k, type_to_default(k, t, action)} end)
   end
 
+  def value_as_map(%Date{} = date), do: %{day: date.day, month: date.month, year: date.year}
```
feliperenan, line 178: @josevalim before adding docs & tests, what do you think about this solution? Since this change is required only for live tests, I tried to figure out a way to change these params only here.

feliperenan: My first thought was to use this function:

https://github.com/phoenixframework/phoenix/blob/c8883af5582a38496e4b7e45e05d3a4d759a6caa/lib/mix/phoenix/schema.ex#L203

Then we could have those params created as we wish. But I couldn't figure out how 😢

lib/mix/tasks/phx.gen.html.ex
```diff
@@ -176,13 +176,13 @@ defmodule Mix.Tasks.Phx.Gen.Html do
       {key, :text} ->
         {label(key), ~s(<%= textarea f, #{inspect(key)} %>), error(key)}
       {key, :date} ->
-        {label(key), ~s(<%= date_select f, #{inspect(key)} %>), error(key)}
+        {label(key), ~s(<%= date_select f, #{inspect(key)}, year: [options: 2010..#{Date.utc_today().year}] %>), error(key)}
```
feliperenan, line 179: Since params is using dates from `2010` I had to choose either update there or here or the date select will not have `2010` as an option. The issue updating the params is that tests will break after some years, that is why I'd rather keep this change here.

josevalim: Interesting. I prefer to change the value we generate for the tests then, make sure that it always mirrors Date.utc_today. WDYT?

lib/mix/phoenix/schema.ex
josevalim: I like the current solution to be honest. My only suggestion is to call this function something like `live_form_value`. WDYT?

lib/mix/tasks/phx.gen.html.ex
josevalim: To be more precise, I would change `type_to_default` to return `Date.add(Date.utc_today, -1, :days)` for create and `Date.utc_today()` for update. We need to do similar to naive_datetime and friends too. ALthough for NaiveDateTime we can do -1 hour. Thoughts?

feliperenan: Make more sense haha 👍 

Just updated. For `NaiveDatetime` and `Datetime` I am doing `-100_000` which goes backwards one day, hope that is fine 🙈

lib/mix/phoenix/schema.ex
```diff
@@ -254,11 +256,29 @@ defmodule Mix.Phoenix.Schema do
...
         :time_usec      -> ~T[14:00:00.000000]
         :uuid           -> "7488a646-e31f-11e4-aace-600308960662"
-        :utc_datetime   -> DateTime.utc_now() |> DateTime.add(-100_000) |> DateTime.truncate(:second)
-        :utc_datetime_usec -> DateTime.add(DateTime.utc_now(), -100_000)
-        :naive_datetime -> NaiveDateTime.utc_now() |> NaiveDateTime.add(-100_000) |> NaiveDateTime.truncate(:second)
-        :naive_datetime_usec -> NaiveDateTime.add(NaiveDateTime.utc_now, -100_000)
-        _               -> "some #{key}"
+        :utc_datetime   ->
+          DateTime
+          |> build_datetime()
```
feliperenan, line 261: @josevalim I feel this code is getting too complicated but I couldn't think in another way to build the date here and assert them in the tests. The issue is that we need to force the seconds and microseconds in order to proper assert `datetime` and `naive_datetime`.

Another option would be removing the date from the assertion, but I am afraid we are going to lose test coverage with that. Do you have something better in mind?

ulissesalmeida: Looks like the seconds is a constant:
```
seconds = -24 * 3600
```

Then it means, it doesn't need to be here. 

Then I think it's possible to create something like:

```elixir
# ...
:utc_date_time -> build_utc_date_time()
# ...

def build_utc_date_time(), 
  do: DateTime.utc_now() |> DateTime.add(@seconds) |> DateTime.truncate(:second)

# apply the same pattern for the other data constructors
```

The naming here could be improved, for example, I have no idea what the `seconds` represent here.

I hope it helps 🙏  
Great job

feliperenan: Humm... I updated the code inspired by your version. It sounds better now. Wdyt?

One thing that I don't like is that I will need to do the same in the tests, or I could expose those functions and use there 🤔.

priv/templates/phx.gen.live/live_test.exs
```diff
@@ -4,9 +4,9 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
   import Phoenix.LiveViewTest
   import <%= inspect context.module %>Fixtures
 
-  @create_attrs <%= inspect schema.params.create %>
-  @update_attrs <%= inspect schema.params.update %>
-  @invalid_attrs <%= inspect for {key, _} <- schema.params.create, into: %{}, do: {key, nil} %>
+  @create_attrs <%= inspect for {key, value} <- schema.params.create, into: %{}, do: {key, Mix.Phoenix.Schema.live_form_value(value)} %>
+  @update_attrs <%= inspect for {key, value} <- schema.params.update, into: %{}, do: {key, Mix.Phoenix.Schema.live_form_value(value)} %>
+  @invalid_attrs <%= inspect for {key, value} <- schema.params.create, into: %{}, do: {key, value |> Mix.Phoenix.Schema.live_form_value() |> Mix.Phoenix.Schema.build_invalid_value()} %>
```
josevalim, line 9: Should we rename `build_invalid_value` to `invalid_form_value` for consistency with `live_form_value`?

josevalim:
@feliperenan one last (minor) comment and we can ship it. :)

feliperenan:
@josevalim applied and tests are green now.

josevalim:

udoschneider:
It seems there are still some issues with `utc_datetime`. Generating an app with:

```
$ mix phx.new live_ts_test --live
$ # Follow instructions (e.g. ecto and routes)     
$ mix phx.gen.live Context Entry entries timestamp:utc_datetime
$ mix test
```

creates tests that do not work out of the box. Using 1.5.6 raises errors. Using master with this commit “just” fails.

feliperenan:
humm... in this case test fails because the in `@invalid_attrs` will be generated with something like:

```Elixir
  @invalid_attrs %{timestamp: %{day: 26, hour: 21, minute: 1, month: 10, year: 2020}}
```

Am I right? 🤔 

I am not sure if we can generate an invalid attribute for this case since there is a select for each field and user's couldn't submit an invalid date. Maybe this is something we can live with since we are using `date_select` or is there something we could do here?

josevalim:
You can generate a date like 30/feb of that year, 31st April, etc. :)

feliperenan:
Ok, make sense haha. We also need to change the assert message:

https://github.com/phoenixframework/phoenix/blob/779259bc6e281bba2f1e71e4b642e97281399095/priv/templates/phx.gen.live/live_test.exs#L36

In this case the error message would be `is invalid` instead of `can't be blank`. I will work on this during this week.

PR #4019, opened 2020-10-10
Add Elixir 1.11 to CI

adrianomitre:
This pull request adds Elixir 1.11 to CI and updates Erlang/OTP minor and/or patch versions to their latest.

Its main distinction from https://github.com/phoenixframework/phoenix/pull/4015 is that it mantains conformity to [José Valim advice](https://github.com/dashbitco/broadway/pull/188#issuecomment-649314540):
> I would still advise folks to go with:
> * For each Elixir version, add a run with earliest supported [major] Erlang/OTP
> * For the last Elixir version, also test the latest supported [major] Erlang/OTP

Please refer to the original discussion on https://github.com/dashbitco/broadway/pull/188 for the full rationale.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -45,11 +45,11 @@ jobs:
       fail-fast: false
       matrix:
         include:
-          # Can't run integration tests before OTP 20.2 due to incompatiblity with cowlib.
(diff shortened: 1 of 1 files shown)


josevalim:

PR #4021, opened 2020-10-11
Silence context generation warnings on `--no-context`

### LINKED ISSUE #3994: `mix phx.gen.live --no-context` warns about existing context
### Environment

* Elixir version (elixir -v): Elixir 1.10.3 (compiled with Erlang/OTP 22)
* Phoenix version (mix deps): phoenix 1.5.4

### Expected behavior

When I run `mix phx.gen.live --no-context`, I do not expect a warning about adding to an existing context.

### Actual behavior

The `--no-context` command warns about adding to an existing context.  If you proceed `y`, then it doesn't create the context as expected.  However, the warning is confusing because it implies the generator is going to inject code into your existing context.

See terminal output below.

```terminal
$ mix phx.gen.live --no-context --no-schema --web commish FantasyLeagues FantasyLeague fantasy_leagues fantasy_league_name:string 

You are generating into an existing context.

The Ex338.FantasyLeagues context currently has 9 functions and 5 files in its directory.

  * It's OK to have multiple resources in the same context as long as they are closely related. But if a context grows too large, consider breaking it apart

  * If they are not closely related, another context probably works better

The fact two entities are related in the database does not mean they belong to the same context.

If you are not sure, prefer creating a new context over adding to the existing one.

Would you like to proceed? [Yn] y

* creating lib/ex338_web/live/commish/fantasy_league_live/show.ex
* creating lib/ex338_web/live/commish/fantasy_league_live/index.ex
* creating lib/ex338_web/live/commish/fantasy_league_live/form_component.ex
* creating lib/ex338_web/live/commish/fantasy_league_live/form_component.html.leex
* creating lib/ex338_web/live/commish/fantasy_league_live/index.html.leex
* creating lib/ex338_web/live/commish/fantasy_league_live/show.html.leex
* creating test/ex338_web/live/commish/fantasy_league_live_test.exs
```

landongrindheim:
When a context exists and a user issues `mix phx.gen.live --no-context`, there is a warning about generating into an existing context, even though the user is not using the context. This change prevents that warning from displaying in the shell.

**Note:** The files being output of the generators are not changing, only the warning/prompt.

The tests in this branch, without the corresponding production code applied, output the following:

```
 ** (RuntimeError) no shell process input given for yes?/1
 stacktrace:
   (mix 1.10.4) lib/mix/shell/process.ex:148: Mix.Shell.Process.yes?/1
   (phoenix 1.6.0-dev) lib/mix/tasks/phx.gen.context.ex:279: Mix.Tasks.Phx.Gen.Context.prompt_for_code_injection/1
   (phoenix 1.6.0-dev) lib/mix/tasks/phx.gen.live.ex:98: Mix.Tasks.Phx.Gen.Live.run/1
   test/mix/tasks/phx.gen.live_test.exs:288: anonymous fn/0 in Mix.Tasks.Phx.Gen.LiveTest."test with --no-context does not emit warning when context exists"/1
   (elixir 1.10.4) lib/file.ex:1544: File.cd!/2
   installer/test/mix_helper.exs:36: MixHelper.in_tmp_project/2
   test/mix/tasks/phx.gen.live_test.exs:282: (test)

```

While the reporting issue (#3994) was specifically targeting `mix phx.gen.live`, I found that the issue also applied to `mix phx.gen.json` and `mix.gen.html` as well.

Closes #3994

### DIFF
--- a/lib/mix/tasks/phx.gen.context.ex
+++ b/lib/mix/tasks/phx.gen.context.ex
@@ -254,6 +254,7 @@ defmodule Mix.Tasks.Phx.Gen.Context do
     """
   end
 
+  def prompt_for_code_injection(%Context{generate?: false}), do: :ok
   def prompt_for_code_injection(%Context{} = context) do
     if Context.pre_existing?(context) do
       function_count = Context.function_count(context)
--- a/test/mix/tasks/phx.gen.html_test.exs
+++ b/test/mix/tasks/phx.gen.html_test.exs
@@ -259,6 +259,31 @@ defmodule Mix.Tasks.Phx.Gen.HtmlTest do
     end
   end
 
+  test "with --no-context no warning is emitted when context exists", config do
+    in_tmp_project config.test, fn ->
+      Gen.Html.run(~w(Blog Post posts title:string))
+
+      assert_file "lib/phoenix/blog.ex"
+      assert_file "lib/phoenix/blog/post.ex"
+
+      Gen.Html.run(~w(Blog Comment comments title:string --no-context))
+      refute_received {:mix_shell, :info, ["You are generating into an existing context" <> _]}
+
+      assert_file "test/phoenix_web/controllers/comment_controller_test.exs", fn file ->
+        assert file =~ "defmodule PhoenixWeb.CommentControllerTest"
+      end
+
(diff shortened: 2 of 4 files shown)


josevalim:

PR #4023, opened 2020-10-12
perf(phoenix.js): Browser recommendations suggest using pagehide inst…

tomciopp:
…ead of unload

This should be double checked to make sure everything is kosher.

https://developers.google.com/web/updates/2018/07/page-lifecycle-api#legacy-lifecycle-apis-to-avoid
https://developers.google.com/web/updates/2018/07/page-lifecycle-api#event-unload

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -767,7 +767,8 @@ export class Socket {
       this.decode = this.defaultDecoder
     }
     if(phxWindow && phxWindow.addEventListener){
-      phxWindow.addEventListener("unload", e => {
+      const terminationEvent = "onpagehide" in self ? "pagehide" : "unload";
+      phxWindow.addEventListener(terminationEvent, e => {
         if(this.conn){
           this.unloaded = true
           this.abnormalClose("unloaded")


chrismccord:
Does pagehide fire on unload scenarios, for example when the page is refreshed, a link is clicked, a form is submitted, etc?

tomciopp:
@chrismccord I wasn't sure so I added the following code. 


```
```

tomciopp:
The event should also be highly compatible, so I don't think we will have issues here.

https://developer.mozilla.org/en-US/docs/Web/API/Window/pagehide_event

tomciopp:
@chrismccord Anything else need to be checked here?

tomciopp:
@chrismccord Checking in again

tomciopp:
@chrismccord Bumping...

chrismccord:
No need to bump. If it's here and open, it's on my list :)

tomciopp:
👍

dealloc:
Just want to add that if you run the Chrome lighthouse test we lose some points on "Best Practices"
Screenshot from 2021-04-04 20-33-50

aspala:
Phoenix LiveView also registers an unload event: https://github.com/phoenixframework/phoenix_live_view/blob/master/assets/js/phoenix_live_view.js#L737

See also: https://github.com/phoenixframework/phoenix_live_view/issues/1373

chrismccord:
We may use a different binding based on how other features shake out, but it won't be because of an arbitrary benchmark , especially when such bindings are necessary on some browsers for LV, and have no ill effects on others :) Thanks!

tomciopp:
@chrismccord It's not just a benchmark thing, I think every browser vendor explicitly states to avoid using this event. It's highly likely that if we depend on this event some features will be subtly broken. Any performance benefits are a side benefit. 

https://developer.mozilla.org/en-US/docs/Web/API/Window/unload_event#usage_notes
https://developers.google.com/web/updates/2018/07/page-lifecycle-api#legacy-lifecycle-apis-to-avoid

tomciopp:
Aslo `pagehide` is literally the event that fires before `unload` in the page lifecycle there should be no reason for this to cause any ill effects.

dealloc:
if compatibility is a problem ([caniuse](https://caniuse.com/?search=pagehide)) we could always fall back to `unload`, but use `pagehide` by default

cduruk:
Just as a note — we had previously issues with `unload` not firing on mobile browsers and had to switch to `visibilitychange`. In our use-case, it was a drop-in replacement for every case of `unload`.

mcrumm:
> pagehide is literally the event that fires before unload

I'm afraid it's just not that straightforward. Not all browsers fire the same lifecycle events. Not all browsers fire lifecycle events in the same order. In some browsers the order of events changes depending on the context, for instance when using the back/forward buttons vs. opening a link in a new tab vs. refreshing the page (both button and address bar).

Here are, just as an example, two related Webkit bugs:

* https://bugs.webkit.org/show_bug.cgi?id=151234
* https://bugs.webkit.org/show_bug.cgi?id=151610

cduruk:
@mcrumm — have you folks explored switching to `visibilitychange`? I think `pagehide` and the Lighthouse scores are red-herring here if we are considering moving away from `unload`. My 0.02$ is  `unload` is  actively discouraged in strong terms by both Chromium and Firefox folks, it doesn't work well on iOS Safari at all (in my personal experience) and it's good to explore what we can do here.

tomciopp:
@mcrumm Based on my prior reading and knowledge `pagehide` is strictly superior to `unload`. Every browser fires these events in that order and have been doing so for ~8 years at minimum. (In the above referenced bug, I believe it was also the case that `unload` would not fire so we would be in the same place) 

However after doing more reading[0] I agree w/ @cduruk that `visibilitychange` is likely the correct event to listen to in the long run. Since all the related bugs to these have been patched we should be good going forward. Is there an official policy on what browsers we support? 

[0] https://www.igvita.com/2015/11/20/dont-lose-user-and-app-state-use-page-visibility/

cduruk:
I'll send another PR w/ `visibilitychange` tonight.

aspala:
Hello!

I just wanted to thank you all for your involvement with this issue. I really appreciate the elixir/phoenix community! 

Thank you!

chrismccord:
There is a lot more nuance to this issue and we have many many hours working out a solution with unload for all browsers. We are using unload to avoid the bfcache on certain browsers/versions. In previous versions of chrome for example (may not be the case now), websocket connections would refuse to reconnect when the page was restored from bfcache, and in other browsers we'd be slow to reconnect so we used unload to enter an aggressive reconnect state. I believe the `pagehide` event is really what we'd want, but it's not supported everywhere. The `visibilitychange` isn't correct because we want to remain connected when the tab is not in focus. I'd accept a PR that replaces unload with something cross browser, but it's a lot more complicated and requires testing than just a find replace. Thanks!

cduruk:
> There is a lot more nuance to this issue and we have many many hours working out a solution with unload for all browsers. We are using unload to avoid the bfcache on certain browsers/versions. In previous versions of chrome for example (may not be the case now), websocket connections would refuse to reconnect when the page was restored from bfcache, and in other browsers we'd be slow to reconnect so we used unload to enter an aggressive reconnect state. I believe the `pagehide` event is really what we'd want, but it's not supported everywhere. The `visibilitychange` isn't correct because we want to remain connected when the tab is not in focus. I'd accept a PR that replaces unload with something cross browser, but it's a lot more complicated and requires testing than just a find replace. Thanks!

Appreciate the context on the issue. I'd open the PR as a starting point and would be testing it extensively. We are all professionals here, no one is suggesting a mindless copy-paste, we all love this project.

Does this project / LV have docs on what browsers / market penetration it's targeting against? That'd help make these decisions with a bit more data. I am mainly curious about the state of LV / sockets on iOS because in my experience `unload` simply doesn't fire in most cases there.

chrismccord:
FYI master now uses pagehide to enter a reconnect state https://github.com/phoenixframework/phoenix/commit/27d00def0fd11ee1f8dd0377470635d3cedd8dc8

fschoenfeldt:
thanks @chrismccord !

caioaao:
(sorry in advance for necroposting, but I felt this is a continuation to this thread)

It's been 3 years since this discussion, so maybe it's good to revisit the decision of not using `visibilitychanged`? Chrome 88 [introduced aggressive throttling](https://developer.chrome.com/blog/timer-throttling-in-chrome-88/) for chained timers (which is used by phoenix socket), and we've seen issues with sockets not reconnecting after focusing on a tab that was in the background. In these cases we also didn't see the `pagehide/show` events firing. 

Some more evidence:
- This article, suggested by MDN, advocates for relying on visibilitychanged events: https://www.igvita.com/2015/11/20/dont-lose-user-and-app-state-use-page-visibility/
- Supabase seems to be suffering from the same stuff I am: https://github.com/supabase/realtime-js/issues/121 they opted to moving the healthcheck to a web worker though (funny enough, they mention in that thread that the project was inspired in phoenix-js, and that they got the `pagehide` stuff from here)

PR #4027, opened 2020-10-14
Run tests from generated apps in CI

aaronrenner:
The goal of this PR is to have the integration tester ensure the test suites from generated apps pass. By default, tests that use the database are excluded, but can be included using standard ExUnit tag related arguments like `--include database:postgresql`.

I'm submitting this as a draft because I'd like to get your thoughts on this approach and still have some things to figure out.

* We need to have the databases accessible via local ports on the test runner's container, so we don't have to modify the configs of the generated apps. I'm currently setting up a proxy with `socat` but that doesn't appear to be working for `mssql`. I wonder if `iptables` would work better for this, but haven't been able to figure it out yet.
* I'm still trying to figure out how I should break these tests into different files so tests can run in parallel, but I don't want to have so many test files it gets confusing. I'm thinking I'll start out with one test module per database and break it up later. I know the postgresql module will be huge and the test suite won't achieve the parallelism it could, but maybe it will become clear how to split things up later. Also, the next step in the integration test work is to run the `mix phx.gen.*` generators on a project and ensure these tests pass as well.
* I still need migrate the rest of the tests from 
  * `phx_new_test.exs`
  * `phx_new_web_test.exs`
  * `phx_new_ecto_test.exs`
  * `phx_new_umbrella_test.exs`

@josevalim If you have any thoughts on the direction of this work, I'd greatly appreciate it. Thanks!


integration_test/test/code_generation/app_with_defaults_test.exs
```diff
@@ -1,12 +1,314 @@
...
+    test "has no compilation or formatter warnings" do
+      with_installer_tmp("new with defaults", fn tmp_dir ->
+        {app_root_path, output} = generate_phoenix_app(tmp_dir, "phx_blog")
 
-      assert_no_compilation_warnings(app_root_path)
-      assert_passes_formatter_check(app_root_path)
-    end)
+        assert_file(Path.join(app_root_path, "README.md"))
+
+        assert_file(Path.join(app_root_path, ".formatter.exs"), fn file ->
```
josevalim, line 11: I would honestly not move these assertions here. They are perfect for unit testing. Making sure the code compiles (and is formatted) and so on are a great use for integration tests tho!

aaronrenner: Perfect. I thought they might be faster and more appropriate as unit tests too. 👍I did this because I thought there might have been a desire to remove `phx_new_test.exs`, but I'm thinking it makes sense to leave it.

Do you think we should have the code compilation and formatter checks in separate tests from running a database-backed test suite? It will be slower, but I'm thinking this may still be a good idea, so people can still get some benefit from the test suite if the databases aren't running.

josevalim:
Hi @aaronrenner! This looks exciting and I think the database setup will be necessary to test things like phx.gen.auth. In my opinion, changing the configuration files are fine, but I am fine with either approach.

My only suggestion is to avoid seeing this as a replacement for the unit tests. In particular, I don't think we need to migrate the assertions from the existing generators. The generators test a bunch of scenarios which we will never completely fully explore via the integration tests (I mean, we could but the integration tests would take so long to run that nobody would run them locally and we would get stuck in a very long feedback cycle).

Here are things I would test:

1. Make sure the code compiles and tests pass out of the box for two or three basic scenarios (this is what we already do)
2. Have a test for the development workflow (i.e. we can do a request, code is compiled dynamically, basically [the test here](https://github.com/phoenixframework/phoenix/blob/master/test/mix/tasks/phx.new_test.exs)) for the scenarios in `1.`
3. Make sure the tests for the gen commands pass (for each database)

And that's pretty much it. It is mean to catch obvious errors but not test every possible combinations our generators can emit. Does it make sense? :)

integration_test/test/code_generation/app_with_defaults_test.exs
josevalim: We want to remove `phx_new_test` but the one inside "test/mix/tasks", not the installer ones. :)

> Do you think we should have the code compilation and formatter checks in separate tests from running a database-backed test suite?

I would say so because we want to check the `--no` options and umbrella stuff works too. I would reserve the database tests to check the output of the other complex `mix phx.gen.*` generators most of all. I guess one way to see this is that the database connection is required to test `mix phx.gen.*` more than anything else.

aaronrenner:
@josevalim I just updated these integration tests to run the `mix phx.gen.{html,live,json}` commands. Once #4033 is merged and this is rebased on top, the tests should pass.

In investigated replacing the development workflow test in `test/mix/tasks/phx.new_test.exs`, but I wasn't able to easily shutdown the server after running `mix phx.server` because of the zombie process issue. I think this test would be best left as is.

Please let me know if there are any changes you'd like to see to these tests or more generators you'd like to have tested. Once this looks good, I'll start preparing `phx.gen.auth` to be merged in.

integration_test/test/code_generation/app_with_mssql_adapter_test.exs
```diff
@@ -0,0 +1,158 @@
...
+            "mssql"
+          ])
+
+        assert_no_compilation_warnings(app_root_path)
+        assert_passes_formatter_check(app_root_path)
+      end)
+    end
+
+    @tag database: :mssql
+    test "has a passing test suite" do
```
josevalim, line 19: I think we can skip this one for mssql and mysql. This will by definition be tested when we test the generators later.

integration_test/test/code_generation/app_with_mssql_adapter_test.exs
```diff
@@ -0,0 +1,158 @@
+defmodule Phoenix.Integration.CodeGeneration.AppWithMSSQLAdapterTest do
+  use Phoenix.Integration.CodeGeneratorCase, async: true
+
+  describe "new with defaults" do
+    test "has no compilation or formatter warnings" do
+      with_installer_tmp("app_with_mssql_adapter", fn tmp_dir ->
```
josevalim, line 6: I think we can skip this one for mssql and mysql. This will by definition be tested when we test the generators later.

integration_test/test/code_generation/app_with_mssql_adapter_test.exs
```diff
@@ -0,0 +1,158 @@
...
+          generate_phoenix_app(tmp_dir, "default_mssql_app", ["--database", "mssql"])
+
+        drop_test_database(app_root_path)
+        assert_tests_pass(app_root_path)
+      end)
+    end
+  end
+
+  describe "phx.gen.html" do
+    test "has no compilation or formatter warnings" do
```
josevalim, line 31: I think we can skip this one for mssql and mysql, it is unlikely to change between adapters.

josevalim:
@aaronrenner this looks great! I have added some comments about some tests we can skip for both mysql and mssql.

Something else we could do is to use setup_all in to setup the basic phoenix tests across the same suite.

Something like this:

```elixir
setup_all do
  with_installer_setup "original" do
    # this should compile all code, including deps for dev and test
    generate_app(...)
  end
end
```

And then on each test:

```elixir
test "compiles" do
  with_installer_run "original", "name of the copy" do
    assert_compiles()
  end
end
```

The `with_installer_setup` removes any existing dir and sets a basic one up while the `with_installer_run` copies the compiled one. Mixing this with the existing deps copying should make it overall very efficient.

Btw, when we copy the deps from integration_test itself, do we also copy the `_build` directory?

josevalim:
> In investigated replacing the development workflow test in test/mix/tasks/phx.new_test.exs, but I wasn't able to easily shutdown the server after running mix phx.server because of the zombie process issue. I think this test would be best left as is.

You can do this when starting mix phx.server:

     System.cmd "elixir", ["--no-halt", "-e", "spawn fn -> IO.gets('') && System.halt(0) end", "-S", "mix", "phx.server"]

The idea is to setup a process that waits until `IO.gets('')` returns something (in this case, :eof) to shutdown the VM. You could even add a `--watch-stdin` flag to `mix phx.server` but I think we can skip it for now.

In any case, I think both `mix phx.server` and [the `with_installer_setup` suggestion I gave here](https://github.com/phoenixframework/phoenix/pull/4027) can be done in another PR. Let's get this one merged first. :)

Btw, when we copy the deps from integration_test itself, do we also copy the _build directory? Also, pls let me know when you think this is ready for merge. :) Thank you! :heart:

aaronrenner:
Hi @josevalim! Thanks for the suggestion on how to shut down the phoenix server after the test halts. I'm looking forward to trying it!

We are copying the `_build`, `deps`, and `mix.lock` file from the integration test project into each of the generated apps to save significant build time:

https://github.com/phoenixframework/phoenix/blob/779259bc6e281bba2f1e71e4b642e97281399095/integration_test/test/support/code_generator_case.ex#L10-L26

I do like this approach where we actually generate the app during the test, because then we have the flexibility to generate it with any number of options even in the same test suite. For example, when testing `phx.gen.live`, this PR generates the app with `--live` where the rest of the generator tests don't use this option. But I agree, we can tackle that in a future PR if we want.

The last thing I'm trying to do is to get the 1.9.4/21.3.8.17 integration test to pass. I think when I copy `_build` from the integration test project into the test app, I need to expand the symlinks. I've been trying to put together some code to do that, it looks like I've run out of time this morning. I'll try to get that done in the next evening or two.

Thanks again for all your help on this!

josevalim:
> The last thing I'm trying to do is to get the 1.9.4/21.3.8.17 integration test to pass.

Don't worry about it. There is a PR that soon will remove support for 1.9.4 and 1.10.0 on the new app generator, which means we no longer need to test those.

> I do like this approach where we actually generate the app during the test, because then we have the flexibility to generate it with any number of options even in the same test suite.

In my suggestion you would lose this ability. The goal is to share apps only within the same block. So for example, we don't need to generate the app twice, one for testing formatting and another for testing the DB. But that's a separate convo. :)

josevalim:
I gone ahead and bumped the integration tests to only Elixir v1.11. Which is the same we are doing in PR #4040.

aaronrenner:
Thanks for taking care of the issue with the older version of Elixir! I just fixed a couple more failing tests, so it's good to go from my side!

Please let me know if there's any other changes you'd like to see on this PR.

josevalim:
Fantastic job! Nothing more for this PR, we will tackle the rest later. :)

josevalim:

josevalim:
> Hi @josevalim! Thanks for the suggestion on how to shut down the phoenix server after the test halts. I'm looking forward to trying it!

Btw, I went ahead and removed the tests from `test` right now, in preparation to us adding them to the integration folder. :)

aaronrenner:
Perfect! I'll make that my next PR!

PR #4032, opened 2020-10-20
Add MIX_ENV when running digest

matthewp:
This adds the `MIX_ENV=prod` to the docs when running `mix.digest`.
Omitting this will cause errors if there are dev only deps.


josevalim:

PR #4036, opened 2020-10-22
Fix typo

gandhiShepard:
I think the author intended to say "The" instead of "To".

...much respect


josevalim:

PR #4039, opened 2020-10-24
determine duplicate channels using Array.prototype.filter until IE11 de-supported

### LINKED ISSUE #3965: phoenix.js error in IE11
### Environment

* Phoenix version: 1.5.4

### Expected behavior
Works fine with IE11.

### Actual behavior
IE11 throws an exception: “Object doesn't support property or method 'find'”

Commit that has introduced that exception: https://github.com/phoenixframework/phoenix/commit/8b0d39058ca19a3a55249f6392eb940e22047380

The line is 
```js
let dupChannel = this.channels.find(c => c.topic === topic && (c.isJoined() || c.isJoining()))
```

levistovall:
…herwise channels.find
fixes #3965

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -1243,7 +1243,7 @@ export class Socket {
 
   leaveOpenTopic(topic) {
     let ver = getInternetExplorerVersion()
-    let dupChannelArrowExpr = c => c.topic === topic && (c.isJoined() || c.isJoining()))
+    let dupChannelArrowExpr = c => c.topic === topic && (c.isJoined() || c.isJoining())
     let dupChannel
     if(ver>=11.0 && ver<12.0) {
       dupChannel = Array.prototype.filter(dupChannelArrowExpr)[0]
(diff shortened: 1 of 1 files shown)


ivanovv:
@levistovall why don't we fallback to safe version without checking IE version, and put a comment to rewrite it once we drop support for IE11?

levistovall:
@ivanovv so you'd rather handle the case of any browser with Array.prototype.filter for now and then potentially change it back to this.channel.find when IE11 support is removed? That makes sense to me; I just wanted to make sure I understood you.

ivanovv:
@levistovall yes, exactly that

levistovall:
@ivanovv does it look okay to you now?

ivanovv:
@levistovall it sure does. The problem is – framework authors ignore this PR and the issue it fixes

@josevalim @chrismccord can you chime in, guys? Do we still support IE11?

chrismccord:
We are no longer officially supporting IE11 within phoenix.js source itself. Microsoft has announced end-of-life for the browser for less than a year from now, and the market share is half a percent or less. For folks requiring IE11 support, they can add the necessary polyfills to their own asset builder and achieve IE11 support. Thanks!

Issue #4038, opened 2020-10-24
Update mix phx.new to use Elixir v1.11 new config/runtime.exs


josevalim:
The plan is to do a gradual migration. For now, we will update the generator so that:

1. It requires Elixir v1.11 and later

2. `config/prod.secret.exs` will be renamed to `config/runtime.exs`. We want to run the `config/runtime.exs` contents only in prod, so we will wrap its existing body in a `if config_env() == :prod` check.

3. We will replace the use of `Mix.env()` in all configs by `config_env()`

4. We will update the guides to refer to `config/runtime.exs` first and foremost (although we should add a parens and say it was called `config/prod.secret.exs` in previous releases)

landongrindheim:
I should have left a `closes` keyword on #4040. Is this good to be closed now?

josevalim:
Yes, thank you! :heart:

syfgkjasdkn:
👋

Is there any downside to using runtime config for everything not related to compilation?

I've made a small dummy project to try out this idea where [I moved all config to `config/runtime.exs`](https://github.com/syfgkjasdkn/runtime-config-all-in/pull/1/files) and then [grouped `config/3` by module being configured](https://github.com/syfgkjasdkn/runtime-config-all-in/pull/2/files).

In that setup when I change some config value, the whole project is not being recompiled which seems like a plus.

The only downside I see is that `runtime.exs` file seems a bit unstructured.

josevalim:
The main limitation is that there is no easy way right now to know if something is a compile or a runtime config besides trial and error, so I am wary of blindingly pushing folks to this direction. When the new Application.compile_env! becomes widespread enough, we will at least have some indicators, that’s when we can start using this more.

PR #4040, opened 2020-10-25
Update `mix phx.new` to use Elixir v1.11 new config/runtime.exs

landongrindheim:
I assume this is a naive approach to addressing #4038. I followed the steps laid out within the issue:

> 1.   It requires Elixir v1.11 and later
> 2.   config/prod.secret.exs will be renamed to config/runtime.exs. We want to run the config/runtime.exs contents only in prod, so we will wrap its existing body in a if config_env() == :prod check.
> 3.   We will replace the use of Mix.env() in all configs by config_env()
> 4.    We will update the guides to refer to config/runtime.exs first and foremost (although we should add a parens and say it was called config/prod.secret.exs in previous releases)

Due to the nature of these changes, CI will fail for Elixir versions lower than 1.11.


installer/templates/phx_single/config/prod.exs
```diff
@@ -50,6 +50,6 @@ config :logger, level: :info
 #
 # Check `Plug.SSL` for all available options in `force_ssl`.
 
-# Finally import the config/prod.secret.exs which loads secrets
+# Finally import the config/runtime.exs which loads secrets
 # and configuration from environment variables.
-import_config "prod.secret.exs"
+import_config "runtime.exs"
```
josevalim, line 55: We should not need to import this. As runtime.exs is always read automatically. :)

installer/templates/phx_single/config/runtime.exs
```diff
@@ -0,0 +1,32 @@
+import Config
+# In this file, we load production configuration and secrets
+# from environment variables. You can also hardcode secrets,
+# although such is generally not recommended and you have to
+# remember to add this file to your .gitignore.
```
josevalim, line 5:
```suggestion

# config/runtime.exs is executed for all environments, including
# during releases. It is executed after compilation and before the
# system starts, so it typically used load production configuration
# and secrets from environment variables or elsewhere. Do not define
# any compile-time configuration in here, as it won't be applied.
```

installer/lib/phx_new/generator.ex
```diff
@@ -198,7 +198,7 @@ defmodule Phx.New.Generator do
     config :#{binding[:app_name]}, #{binding[:app_module]}.Repo#{kw_to_config adapter_config[:test]}
     """
 
-    config_inject project_path, "config/prod.secret.exs", """
+    config_inject project_path, "config/runtime.exs", """
```
josevalim, line 201: I am afraid this will introduce it it in the wrong place. You need to introduce inside the "if" block and it needs to be indented properly. :)

josevalim:
thanks @landongrindheim ! I have added some comments. note you will have to update the CI file: integration tests should now run only on v1.11.

You also need to wrap test/mix/tasks/phx.new_test.exs into an if block and only execute it for Elixir v1.11. :)

.github/workflows/ci.yml
```diff
@@ -29,10 +29,34 @@ jobs:
           apk add --no-progress --update git
           mix deps.get
       - run: mix test
+      # - run: MIX_ENV=docs mix inch.report
```
landongrindheim, line 32: I wasn't sure what to do with this, so I left it with the original grouping (`test-elixir`)

test/mix/tasks/phx.new_test.exs
```diff
@@ -1,100 +1,102 @@
...
-end
-
-_ = Kernel.ParallelCompiler.compile Path.wildcard("installer/lib/**/*.ex")
-Code.require_file("installer/test/mix_helper.exs")
-
-# Define a fake live reload socket.
-defmodule Phoenix.LiveReloader.Socket do
-  def child_spec(_) do
-    Supervisor.child_spec({Task, fn -> :ok end}, [])
+if Version.match?(System.version(), ">= 1.11.0") do
```
landongrindheim, line 1: Maybe this could be inverted to `Mix.raise` if the version was smaller than 1.11. It would make the diff cleaner.

```elixir
if Version.match?(System.version(), ">= 1.11.0") do
  Mix.raise(...)
end

# Remove archive to avoid conflicts
with list when is_list(list) <- :code.lib_dir(:phx_new) do
```

.github/workflows/ci.yml
josevalim: Yup, leave it here, remove the one from the installer. :)

.github/workflows/ci.yml
```diff
@@ -29,10 +29,34 @@ jobs:
...
+          - elixir: 1.11.0
+            otp: 23.1.1
+    steps:
+      - uses: actions/checkout@v2.3.1
+      - name: Install Dependencies
+        run: |
+          mix local.rebar --force
+          mix local.hex --force
+          apk add --no-progress --update git
+          mix deps.get
```
josevalim, line 56: Instead of having this as a separate group, we can move the installer test with the integration tests, because they always need to run in the same Elixir versions. :)

installer/lib/phx_new/generator.ex
```diff
@@ -83,6 +83,28 @@ defmodule Phx.New.Generator do
     end
   end
 
+  def prod_only_config_inject(path, file, to_inject) do
+    file = Path.join(path, file)
+
+    contents =
+      case File.read(file) do
```
josevalim, line 90: The file will always exist because we just generated it, so you can use `File.read!` here.

installer/lib/phx_new/generator.ex
```diff
@@ -83,6 +83,28 @@ defmodule Phx.New.Generator do
...
+        """
+      end
+
+    case split_with_self(contents, "if config_env() == :prod do") do
+      [left, middle, right] ->
+        File.write!(file, [left, middle, ?\n, String.trim_trailing(to_inject), ?\n, right])
+
+      :error ->
+        Mix.raise ~s[Could not find "if config_env() == :prod do" in #{inspect(file)}]
+    end
```
josevalim, line 105: similar here, it should always be there!

```suggestion
    [left, middle, right] = split_with_self(contents, "if config_env() == :prod do") do
    File.write!(file, [left, middle, ?\n, String.trim_trailing(to_inject), ?\n, right])
```

test/mix/tasks/phx.new_test.exs
josevalim: If we raise the tests will fail. It is fine to leave this as is for now because we plan to move all of this elsewhere soon!

installer/lib/phx_new/generator.ex
landongrindheim: `Mix.Tasks.Phx.New.Ecto.run/1` depends on the presence of this check in installer/test/phx_new_ecto_test.exs. Is that something that needs to change?

josevalim: Good call. Trust the tests, not me. So this is good to go. :)

installer/lib/phx_new/generator.ex
```diff
@@ -83,6 +83,28 @@ defmodule Phx.New.Generator do
...
+  def prod_only_config_inject(path, file, to_inject) do
+    file = Path.join(path, file)
+
+    contents =
+      case File.read(file) do
+        {:ok, bin} -> bin
+        {:error, _} -> """
+          import Config
+
+          if config_env() == :prod do
```
josevalim, line 95: You need to close the block, otherwise the file will be invalid. :)

```suggestion
          if config_env() == :prod do
          end
```

installer/templates/phx_umbrella/config/extra_config.exs
```diff
@@ -1,3 +1,4 @@
+import Config
```
josevalim, line 1: I think this change is wrong. If we keep this, we will have two `import Config` on the file we are augmenting. What was the reason for this change? :)

installer/lib/phx_new/web.ex
```diff
@@ -9,7 +9,7 @@ defmodule Phx.New.Web do
     {:config, "#{@pre}/config/config.exs",            :project, "config/config.exs"},
     {:config, "#{@pre}/config/dev.exs",               :project, "config/dev.exs"},
     {:config, "#{@pre}/config/prod.exs",              :project, "config/prod.exs"},
-    {:config, "#{@pre}/config/prod.secret.exs",       :project, "config/prod.secret.exs"},
+    {:config, "#{@pre}/config/runtime.exs",           :project, "config/runtime.exs"},
```
josevalim, line 12: I think this should be:

```suggestion
    {:prod_config, "#{@pre}/config/runtime.exs",           :project, "config/runtime.exs"},
```

And then in here:

https://github.com/phoenixframework/phoenix/blob/0048cd7a239982b7b4c6c20de780886b9027dee1/installer/lib/phx_new/generator.ex#L58

You will add a new clause that is exactly the same as `:config`, except it calls your new `prod_only_config_inject` function (great job on it btw). If you don't do this change, `web.ex` will inject contents outside of the `config_env() == :prod` block. :)

installer/templates/phx_umbrella/config/extra_config.exs
landongrindheim: You're 100% correct. This was done without fully realizing the consequences.

My next task was figuring out why there were two `import Config` statements 😁. Thanks for pointing this out!

josevalim:
@landongrindheim I have made three remaining comments, one that you already merged, and once they are complete we can merge this! 🚀

installer/lib/phx_new/web.ex
landongrindheim: @josevalim I've tried a couple of approaches now:

First, using the [current web config/runtime.exs](installer/templates/phx_umbrella/apps/app_name_web/config/runtime.exs), the contents of that file are appended after `import Config`. Any other prod-specific configs are injected using the `prod_only_config_inject` function. This results in the app_root/config/runtime.exs file that looks like what I would expect:
<details>
  <summary>current config/runtime.exs</summary>

```elixir
# In this file, we load production configuration and secrets
# from environment variables. You can also hardcode secrets,
# although such is generally not recommended and you have to
# remember to add this file to your .gitignore.
import Config

if config_env() == :prod do
  database_url =
    System.get_env("DATABASE_URL") ||
      raise """
      environment variable DATABASE_URL is missing.
      For example: ecto://USER:PASS@HOST/DATABASE
      """

  config :dev_app, DevApp.Repo,
    # ssl: true,
    url: database_url,
    pool_size: String.to_integer(System.get_env("POOL_SIZE") || "10")

  secret_key_base =
    System.get_env("SECRET_KEY_BASE") ||
      raise """
      environment variable SECRET_KEY_BASE is missing.
      You can generate one by calling: mix phx.gen.secret
      """

  config :dev_app_web, DevAppWeb.Endpoint,
    http: [
      # Enable IPv6 and bind on all interfaces.
      # Set it to  {0, 0, 0, 0, 0, 0, 0, 1} for local network only access.
      ip: {0, 0, 0, 0, 0, 0, 0, 0},
      port: String.to_integer(System.get_env("PORT") || "4000")
    ],
    secret_key_base: secret_key_base

  # ## Using releases
  #
  # If you are doing OTP releases, you need to instruct Phoenix
  # to start each relevant endpoint:
  #
  #     config :dev_app_web, DevAppWeb.Endpoint, server: true
  #
  # Then you can assemble a release by calling `mix release`.
  # See `mix help release` for more information.
end
```
</details>

This has the downside of being order dependent: The expectation is that the web config will be copied before the DB config (or any other prod-spcific config).

---

I also tried the approach you suggested (using `prod_only_config_inject` to inject the web config into app_root/config/runtime.exs, which contains `if config_env() == :prod do`). This approach was a little smarter about indentation the web config didn't have to have the environment check and didn't require indentation. However, I've had a hard time getting the formatting correct with this approach: `prod_only_config_inject` will insert a newline before and after the injected text. This leads to a config file which isn't formatted correctly (specifically, the newline before `end`):

<details>
  <summary>injected config/runtime.exs</summary>

```elixir
# In this file, we load production configuration and secrets
# from environment variables. You can also hardcode secrets,
# although such is generally not recommended and you have to
# remember to add this file to your .gitignore.
import Config

if config_env() == :prod do
  database_url =
    System.get_env("DATABASE_URL") ||
      raise """
      environment variable DATABASE_URL is missing.
      For example: ecto://USER:PASS@HOST/DATABASE
      """

  config :dev_app, DevApp.Repo,
    # ssl: true,
    url: database_url,
    pool_size: String.to_integer(System.get_env("POOL_SIZE") || "10")

  secret_key_base =
    System.get_env("SECRET_KEY_BASE") ||
      raise """
      environment variable SECRET_KEY_BASE is missing.
      You can generate one by calling: mix phx.gen.secret
      """

  config :dev_app_web, DevAppWeb.Endpoint,
    http: [
      # Enable IPv6 and bind on all interfaces.
      # Set it to  {0, 0, 0, 0, 0, 0, 0, 1} for local network only access.
      ip: {0, 0, 0, 0, 0, 0, 0, 0},
      port: String.to_integer(System.get_env("PORT") || "4000")
    ],
    secret_key_base: secret_key_base

  # ## Using releases
  #
  # If you are doing OTP releases, you need to instruct Phoenix
  # to start each relevant endpoint:
  #
  #     config :dev_app_web, DevAppWeb.Endpoint, server: true
  #
  # Then you can assemble a release by calling `mix release`.
  # See `mix help release` for more information.

end
```
</details>

I'm sure there's an ergonomic way around this, but I haven't hit it yet. This approach has the upside of being order-independent, but I'm not quite there with the implementation. I can push a commit that shows this work if you're interested in seeing it. I can do some commit cleanup once the direction is decided.

josevalim: I think what you can do for now is to go with the second approach and automatically amend the end. If you see this: "\n\nend\n", you can rewrite it to "\nend\n". I think this transformation is always safe. :)

josevalim: Well, there is another approach you can take that actually solves all of this: just inject the contents, without indentation nor anything, and then call `Code.format_string!`. :D It will indent and resolve everything for us. Plus, even better, it will guarantee it is formatted. We can also call it for the regular config_inject too!

landongrindheim: I haven't figured out how to plug in `Code.format_string!` yet. I've tried a few variations of:
```elixir
    case split_with_self(contents, "if config_env() == :prod do") do
      [left, middle, right] ->
        formatted_contents =
          Enum.join([left, middle, ?\n, indent(to_inject, 2), ?\n, right])
          |> Code.format_string!()
        File.write!(file, formatted_contents)
```

Each time I end up with exceptions like:
```
** (SyntaxError) nofile:32:1: unexpected reserved word: end
    (elixir 1.11.1) lib/code.ex:654: Code.format_string!/2
```

I've also tried formatting the file in place:
```elixir
      [left, middle, right] ->
        File.write!(file, [left, middle, ?\n, indent(to_inject, 2), ?\n, right])
        Code.format_file!(file)
```

The result, however, is the same as my non-formatted output (which fails the formatting check).

Any tips for using `Code.format_#{thing}`?

josevalim: If format is failing is because the file is indeed invalid. Can you please print the contents of the file before formatting?

josevalim: Also, you shouldn’t need indent anymore. :)

landongrindheim: This is the output of `IO.inspect` when I shim it into the pipeline:

```elixir
"# In this file, we load production configuration and secrets\n# from environment variables. You can also hardcode secrets,\n# although such is generally not recommended and you have to\n# remember to add this file to your .gitignore.\nimport Config\n\nif config_env() == :prod do10  secret_key_base =\n    System.get_env(\"SECRET_KEY_BASE\") ||\n      raise \"\"\"\n      environment variable SECRET_KEY_BASE is missing.\n      You can generate one by calling: mix phx.gen.secret\n      \"\"\"\n  \n  config :dev_app_web, DevAppWeb.Endpoint,\n    http: [\n      # Enable IPv6 and bind on all interfaces.\n      # Set it to  {0, 0, 0, 0, 0, 0, 0, 1} for local network only access.\n      ip: {0, 0, 0, 0, 0, 0, 0, 0},\n      port: String.to_integer(System.get_env(\"PORT\") || \"4000\")\n    ],\n    secret_key_base: secret_key_base\n  \n  # ## Using releases\n  #\n  # If you are doing OTP releases, you need to instruct Phoenix\n  # to start each relevant endpoint:\n  #\n  #     config :dev_app_web, DevAppWeb.Endpoint, server: true\n  #\n  # Then you can assemble a release by calling `mix release`.\n  # See `mix help release` for more information.10\nend\n"
```

josevalim: You can see a 10 after the “do”. That’s because Enum.join is converting ?\n to 10. Either use “IO.iodata_to_binary” instead of Enum.join or use “\n”. I think iodata_to_binary is better!

landongrindheim: That did the trick. Thank you!

landongrindheim:
Thanks for all your help on this @josevalim!

I'm happy to clean up commit history if that is of value. I'm also happy to add the formatting to `config_inject` as well 😄

installer/templates/phx_single/config/runtime.exs
```diff
@@ -0,0 +1,34 @@
+import Config
+
+# config/runtime.exs is executed for all environments, including
+# during releases. It is executed after compilation and before the
+# system starts, so it typically used load production configuration
+# and secrets from environment variables or elsewhere. Do not define
+# any compile-time configuration in here, as it won't be applied.
```
josevalim, line 7:
```suggestion
# any compile-time configuration in here, as it won't be applied.
# The block below contains prod specific runtime configuration.
```

josevalim:

PR #4041, opened 2020-10-26
Update package.json

Boefjim:
### Change
Added "type": "module"

### Reason
Currently NodeJS will give the following error when trying to import phoenix.js:
`The requested module 'phoenix' is expected to be of type CommonJS, which does not support named exports. CommonJS modules can be imported by importing the default export.`
Indicating that the type of the package is a module will fix this.


chrismccord:
We specify the `module` field on master. Thanks! ❤️❤️❤️🐥🔥

PR #4048, opened 2020-10-29
Pin `id` in generated json controller tests

happycollision:
The assertions as written were matching an unpinned `id` variable, which was only confirming that an `id` existed, not that it was the _correct_ id.


happycollision:
Hmm. There are lots of tests that are flaky for me locally. I think my latest change should properly update the needed test so the pinning is expected in the output.

happycollision:
Yep.

Issue #4020, opened 2020-10-11
Upgrading webpack from 4 to 5


joshchernoff:
### Environment

phoenix 1.5.5
elixir      1.10.3-otp-23 
erlang   23.0.1 
nodejs  12.18.4  
npm      6.14.6
OSX 10.15.6 (19G2021)
Webpack 5
```
% mix phx.server
Compiling 59 files (.ex)
Generated morphic_pro app
[info] [alarm_handler: {:set, {:system_memory_high_watermark, []}}]
[info] Running MorphicProWeb.Endpoint with cowboy 2.8.0 at 0.0.0.0:4001 (https)
[info] Access MorphicProWeb.Endpoint at https://localhost:4001
[webpack-cli]
TypeError: Cannot read property 'tap' of undefined
[error] Task MorphicProWeb.Endpoint started from #PID<0.1043.0> terminating
** (stop) :watcher_command_error
    (phoenix 1.5.5) lib/phoenix/endpoint/watcher.ex:37: Phoenix.Endpoint.Watcher.watch/3
    (elixir 1.10.3) lib/task/supervised.ex:90: Task.Supervised.invoke_mfa/2
    (stdlib 3.13) proc_lib.erl:226: :proc_lib.init_p_do_apply/3
Function: &Phoenix.Endpoint.Watcher.watch/3
    Args: ["node", ["node_modules/webpack/bin/webpack.js", "--config", "webpack.dev.config.js", "--color", "--bail"], [cd: "/Users/joshchernoff/Dev/morphic.pro/assets"]]
```
Looks like watch is broken with the webpack 5 upgrade. 

Also, note `--watch-stdin` is no longer an option

joshchernoff:
Not related to phoenix.

zoidbergwill:
I think this is related to Phoenix, and we should re-open?

Webpack 5 is [out](https://webpack.js.org/blog/2020-10-10-webpack-5-release/)

`--watch-stdin` is gone, which is used in the base here: https://github.com/phoenixframework/phoenix/blob/master/installer/templates/phx_single/config/dev.exs#L19

We wanna bump webpack [here](https://github.com/phoenixframework/phoenix/blob/master/installer/templates/phx_assets/package.json#L27)

markevich:
@joshchernoff Hi! Can you please explain why it's not related to Phoenix?

[...]

PR #4053, opened 2020-11-02
Webpack > Use --watch instead of --watch-stdin

ruudk:
This allows Webpack 5. 

Fixes https://github.com/phoenixframework/phoenix/issues/4020


ruudk:
Maybe #4054 is a better approach?

josevalim:
Closing it in favor of #4054 as it is more comprehensive. Thank you!

PR #4054, opened 2020-11-02
Update to webpack v5

### LINKED ISSUE #4020: Upgrading webpack from 4 to 5

cw789:
**Revision**

- Including v1.x version annotation within package.json
- Again "source-map" instead of "eval-cheap-source-map" for now

Background "source-map":
Terser plugin doesn't work with "eval-cheap-module-source-map".

Suggestion for additional PR (if it will improve build time):
```
const webpack = require('webpack');
...
plugins: [
  new webpack.EvalSourceMapDevToolPlugin({
     exclude: ['./vendor/**/*.js']
   })
]
devtool: false
```

---

**Initial** 

Including update to webpack 5.
Some things are probably opinionated.
- Obvious "fixed" versions within package.json (real version is anyway in package-lock.json)
- No license field in package.json as not defined it must be MIT ...

Edit:
- Replaced `optimize-css-assets-webpack-plugin` with `css-minimizer-webpack-plugin` (Thx @optikfluffel)
- `source-map` instead of  `eval-cheap-source-map`

~Has open Issue (warning in CLI, but seem to work anyway): https://github.com/webpack/webpack-cli/issues/1918~
Concerns discussion: https://elixirforum.com/t/upgrade-to-webpack-5/35010

Resolve #4020

### DIFF
--- a/installer/templates/phx_assets/package.json
+++ b/installer/templates/phx_assets/package.json
@@ -8,21 +8,19 @@
     "phoenix": "file:<%= @phoenix_webpack_path %>"<%= if @html do %>,
     "phoenix_html": "file:<%= @phoenix_html_webpack_path %>"<% end %><%= if @live do %>,
     "phoenix_live_view": "file:<%= @phoenix_live_view_webpack_path %>",
-    "topbar": "^0.1.4"<% end %>
+    "topbar": "^1.x"<% end %>
   },
   "devDependencies": {
-    "@babel/core": "^7.0.0",
-    "@babel/preset-env": "^7.0.0",
-    "babel-loader": "^8.0.0",
-    "copy-webpack-plugin": "^5.1.1",
-    "css-loader": "^3.4.2",
-    "sass-loader": "^8.0.2",
-    "sass": "^1.27.0",
-    "hard-source-webpack-plugin": "^0.13.1",
-    "mini-css-extract-plugin": "^0.9.0",
-    "optimize-css-assets-webpack-plugin": "^5.0.1",
-    "terser-webpack-plugin": "^2.3.2",
-    "webpack": "4.41.5",
-    "webpack-cli": "^3.3.2"
+    "@babel/core": "^7.x",
+    "@babel/preset-env": "^7.x",
+    "babel-loader": "^8.x",
+    "copy-webpack-plugin": "^7.x",
+    "css-loader": "^5.x",
+    "css-minimizer-webpack-plugin": "^1.x",
+    "mini-css-extract-plugin": "^1.x",
+    "sass": "^1.x",
+    "sass-loader": "^10.x",
+    "webpack": "^5.x",
+    "webpack-cli": "^4.x"
   }
 }
--- a/installer/templates/phx_assets/webpack.config.js
+++ b/installer/templates/phx_assets/webpack.config.js
@@ -1,30 +1,21 @@
 const path = require('path');
 const glob = require('glob');
-const HardSourceWebpackPlugin = require('hard-source-webpack-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
-const TerserPlugin = require('terser-webpack-plugin');
-const OptimizeCSSAssetsPlugin = require('optimize-css-assets-webpack-plugin');
+const CssMinimizerPlugin = require('css-minimizer-webpack-plugin');
 const CopyWebpackPlugin = require('copy-webpack-plugin');
 
 module.exports = (env, options) => {
   const devMode = options.mode !== 'production';
 
   return {
-    optimization: {
-      minimizer: [
-        new TerserPlugin({ cache: true, parallel: true, sourceMap: devMode }),
-        new OptimizeCSSAssetsPlugin({})
-      ]
-    },
     entry: {
       'app': glob.sync('./vendor/**/*.js').concat(['./js/app.js'])
     },
     output: {
-      filename: '[name].js',
       path: path.resolve(__dirname, '../priv/static/js'),
+      filename: '[name].js',
       publicPath: '/js/'
     },
-    devtool: devMode ? 'eval-cheap-module-source-map' : undefined,
     module: {
       rules: [
         {
@@ -45,9 +36,19 @@ module.exports = (env, options) => {
       ]
     },
     plugins: [
-      new MiniCssExtractPlugin({ filename: '../css/[name].css' }),
-      new CopyWebpackPlugin([{ from: 'static/', to: '../' }])
-    ]
-    .concat(devMode ? [new HardSourceWebpackPlugin()] : [])
+      new MiniCssExtractPlugin({ filename: '../css/app.css' }),
+      new CopyWebpackPlugin({
+        patterns: [
+          { from: 'static/', to: '../' }
+        ]
+      })
+    ],
+    optimization: {
+      minimizer: [
+        '...',
+        new CssMinimizerPlugin()
+      ]
+    },
+    devtool: devMode ? 'source-map' : undefined
   }
 };
--- a/installer/templates/phx_single/config/dev.exs
+++ b/installer/templates/phx_single/config/dev.exs
@@ -18,7 +18,8 @@ config :<%= @app_name %>, <%= @endpoint_module %>,
       "node_modules/webpack/bin/webpack.js",
       "--mode",
       "development",
-      "--watch-stdin",
+      "--watch",
+      "--watch-options-stdin",
       cd: Path.expand("../assets", __DIR__)
     ]
   ]<% else %>[]<% end %>
--- a/installer/templates/phx_umbrella/apps/app_name_web/config/dev.exs
+++ b/installer/templates/phx_umbrella/apps/app_name_web/config/dev.exs
@@ -16,7 +16,8 @@ config :<%= @web_app_name %>, <%= @endpoint_module %>,
       "node_modules/webpack/bin/webpack.js",
       "--mode",
       "development",
-      "--watch-stdin",
+      "--watch",
+      "--watch-options-stdin",
       cd: Path.expand("../apps/<%= @web_app_name %>/assets", __DIR__)
     ]
   ]<% else %>[]<% end %>
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -149,14 +149,14 @@ defmodule Phoenix.Endpoint do
       You can configure it to whatever build tool or command you want:
 
           [node: ["node_modules/webpack/bin/webpack.js", "--mode", "development",
-              "--watch-stdin"]]
+              "--watch", "--watch-options-stdin"]]
 
       The `:cd` option can be used on a watcher to override the folder from
       which the watcher will run. By default this will be the project's root:
       `File.cwd!()`
 
           [node: ["node_modules/webpack/bin/webpack.js", "--mode", "development",
-              "--watch-stdin", cd: "my_frontend"]]
+              "--watch", "--watch-options-stdin", cd: "my_frontend"]]
 
     * `:live_reload` - configuration for the live reload option.
       Configuration requires a `:patterns` option which should be a list of


installer/templates/phx_single/config/dev.exs
```diff
@@ -16,7 +16,10 @@ config :<%= app_name %>, <%= endpoint_module %>,
       "node_modules/webpack/bin/webpack.js",
       "--mode",
       "development",
-      "--watch-stdin",
+      "--watch-",
```
ruudk, line 19: Is this correct?
```suggestion
      "--watch",
```

cw789: Of course not. Thanks a lot.

optikfluffel:
[`optimize-css-assets-webpack-plugin`](https://github.com/NMFR/optimize-css-assets-webpack-plugin) isn't working with webpack 5 and has to be replaced:

> ⚠️ For webpack v5 or above please use css-minimizer-webpack-plugin instead.

Also: is there a reason you removed the `parallel: true` option from the TerserPlugin config?

cw789:
> [`optimize-css-assets-webpack-plugin`](https://github.com/NMFR/optimize-css-assets-webpack-plugin) isn't working with webpack 5 and has to be replaced:
> 
> > ⚠️ For webpack v5 or above please use css-minimizer-webpack-plugin instead.

Good catch - wasn't aware of that.

> Also: is there a reason you removed the `parallel: true` option from the TerserPlugin config?

Not sure anymore - isn't it default to `true`?

optikfluffel:
@cw789 you're right, [parallel is true by default](https://github.com/webpack-contrib/terser-webpack-plugin/blob/661d1f8cd6aee22dd924365a8d7c62bfd0cc5d90/README.md#parallel) now

jeregrine (approved):

jeregrine:
I have set this up locally and can confirm it woks. And I upgraded a legacy project. @chrismccord @josevalim

installer/templates/phx_single/config/dev.exs
```diff
@@ -16,7 +16,10 @@ config :<%= app_name %>, <%= endpoint_module %>,
       "node_modules/webpack/bin/webpack.js",
       "--mode",
       "development",
-      "--watch-stdin",
+      "--watch",
+      "--watch-options-stdin",
+      "--stats",
+      "minimal",
```
josevalim, line 22:
```suggestion
```

installer/templates/phx_assets/package.json
```diff
@@ -1,30 +1,27 @@
 {
-  "repository": {},
-  "description": " ",
-  "license": "MIT",
+  "private": true,
   "scripts": {
-    "deploy": "webpack --mode production",
-    "watch": "webpack --mode development --watch"
+    "deploy": "webpack --mode production --stats errors-only",
+    "watch": "webpack --mode development --stats minimal --watch"
```
josevalim, line 5:
```suggestion
    "deploy": "webpack --mode production",
    "watch": "webpack --mode development --watch"
```

josevalim:
Thank you @cw789! I removed the stats stuff because this is something we discussed in the past and we should probably have a separate discussion before moving forward. This looks good to be merged to me. By the time Phoenix v1.6 is out, webpack will probably be more stable and some plugins will have caught up.

cw789:
Thanks everyone for reviewing.

I'm totally fine by removing the stats stuff within this PR. Maybe it's not bad to show people what's going on.
Even if I'm more focused in what is Phoenix doing (webpack should just do it's minimal necessary things and work).

By the way I also would like the same version syntax `1.x` within the `mix.exs` - `deps`.
It makes it more obvious which range of versions you will get.
But that of course need to be also part of a different discussion.

installer/templates/phx_assets/package.json
```diff
@@ -10,21 +8,20 @@
     "phoenix": "file:<%= phoenix_webpack_path %>"<%= if html do %>,
     "phoenix_html": "file:<%= phoenix_html_webpack_path %>"<% end %><%= if live do %>,
     "phoenix_live_view": "file:<%= phoenix_live_view_webpack_path %>",
-    "nprogress": "^0.2.0"<% end %>
+    "nprogress": "^0.2"<% end %>
```
createdbypete, line 11: Minor nitpick but something I noticed when you first install something with `npm` it will reorder this to put it above phoenix to respect alphabetical ordering of the dependencies. Perhaps we can move it there now and save future projects from that diff?

installer/templates/phx_assets/package.json
```diff
@@ -10,21 +8,20 @@
...
+    "@babel/core": "^7.x",
+    "@babel/preset-env": "^7.x",
+    "babel-loader": "^8.x",
+    "copy-webpack-plugin": "^6.x",
+    "css-loader": "^5.x",
+    "css-minimizer-webpack-plugin": "^1.x",
+    "mini-css-extract-plugin": "^1.x",
+    "sass": "^1.x",
+    "sass-loader": "^10.x",
+    "terser-webpack-plugin": "^5.x",
```
createdbypete, line 23: Noticed on there [plugin page for Terser](https://webpack.js.org/plugins/terser-webpack-plugin/) it mentions:

> If you are using webpack v5 or above you do not need to install this plugin. Webpack v5 comes with the latest `terser-webpack-plugin` out of the box. 

So I think that means we can remove this line too? 🎉

installer/templates/phx_assets/webpack.config.js
```diff
@@ -45,9 +37,20 @@ module.exports = (env, options) => {
...
-    .concat(devMode ? [new HardSourceWebpackPlugin()] : [])
+      new MiniCssExtractPlugin({ filename: '../css/app.css' }),
+      new CopyWebpackPlugin({
+        patterns: [
+          { from: 'static/', to: '../' }
+        ]
+      })
+    ],
+    optimization: {
+      minimize: true,
```
createdbypete, line 48: I believe this will force the assets to be minimised even in development, this may not be ideal for performance reasons if it's going to recompile on every change. Webpack appears to use the `mode` to automatically set this to `true` when using `--mode production` and `false` with `--mode development`.

Reference: [Mode documentation](https://webpack.js.org/configuration/mode/#mode-development)

installer/templates/phx_assets/package.json
cw789: I agree. Also concerns `npm update` it will always reorder it.

installer/templates/phx_assets/package.json
cw789: Thanks, wasn't aware of that.
It works really as expected without it.

installer/templates/phx_assets/webpack.config.js
cw789: Was lazy here, but was afraid of the same. Totally agree.

createdbypete (approved):

navinpeiris:
Live reloading seems to be broken after these changes. Might be my something in my config, but thought I'll let you know just in case.

installer/templates/phx_assets/webpack.config.js
```diff
@@ -1,21 +1,14 @@
 const path = require('path');
 const glob = require('glob');
-const HardSourceWebpackPlugin = require('hard-source-webpack-plugin');
 const MiniCssExtractPlugin = require('mini-css-extract-plugin');
 const TerserPlugin = require('terser-webpack-plugin');
```
ruudk, line 4: This is still used but no longer required in package.json 🤔

dvic: This can go away, it's included in webpack 5: https://webpack.js.org/configuration/other-options/#cache

ruudk: Thanks, TIL!

ruudk: I think the `terser-webpack-plugin` is a transitive dependency. It's pulled in via the `webpack` dependency. But if we use this directly, we should depend on this ourself. 

My editors warns about this:
Screenshot 2020-11-11 at 17 46 14@2x

rockwood:
> Live reloading seems to be broken after these changes. Might be my something in my config, but thought I'll let you know just in case.

This is an issue for for me as well. Live reload is triggering an event for every file copied into /priv/static. Previously, it would only trigger an event for the changed file. I have quite a few images and each time I change a js file, this is what I see in the logs:

```
[webpack-cli] Compilation starting...
[webpack-cli] Compilation finished
webpack output....
[debug] Live reload: priv/static/js/app.js
[debug] Live reload: priv/static/js/app.js
[debug] Live reload: priv/static/js/app.js
[debug] Live reload: priv/static/images/alert.svg
[debug] Live reload: priv/static/images/alert.svg
[debug] Live reload: priv/static/images/alert.svg
[debug] Live reload: priv/static/images/logo.png
[debug] Live reload: priv/static/images/logo.png
[debug] Live reload: priv/static/images/logo.png
...hundreds more
```

The browser is trying to refresh for each of those events.

Could this be an issue webpack's copy plugin? I assume it shouldn't be copying files that haven't changed.

installer/templates/phx_assets/webpack.config.js
cw789: Remove of `terser-webpack-plugin`:
I had the same thing in doubt. But as `npm` is strange anyways I removed it.
Will undo it.

cw789:
> Could this be an issue webpack's copy plugin? I assume it shouldn't be copying files that haven't changed.

Indeed this doesn't look like I would expect it to be too.
But I'm not aware of how the behavior regarding `live-reload` and `CopyWebpackPlugin` was before.

For now I have no solution.

createdbypete:
> Live reloading seems to be broken after these changes. Might be my something in my config, but thought I'll let you know just in case.

🤔 Interesting, I've not experienced this in my project so far but then I don't have a huge number of things in `/static`. Could this be reproduced in a repo we could take a look at?

vortizhe:
I think webpack-cli 5.x isn't watching stdin any more so SIGINT signal do not close node process.

After phoenix server closed, check `ps aux` for node or webpack processes.

I use a small `watch.js` file to monitor end and SIGINT signals.

```
#! /usr/bin/env node
const process = require('process');
const child_process = require('child_process');

const stdin = process.openStdin();
const child = child_process
  .spawn('node_modules/webpack/bin/webpack.js', 
    ['--config', 'webpack.dev.js'], 
    {detached: true}
  );

const endAll = () => {
  process.kill(-child.pid);
  process.exit(0);
};

child.stderr.on('data', (data) => console.error(data.toString('utf8')));
child.stdout.on('data', (data) => console.log(data.toString('utf8')));

stdin.on('end', endAll);
process.on('SIGINT', endAll);
```

vortizhe:
I think webpack-cli 5.x isn't watching stdin any more so SIGINT signal do not close node process.

After phoenix server closed, check `ps aux` for node or webpack processes.

I use a small `watch.js` file to monitor end and SIGINT signals.

```
#! /usr/bin/env node
const process = require('process');
const child_process = require('child_process');

const stdin = process.openStdin();
const child = child_process
  .spawn('node_modules/webpack/bin/webpack.js', 
    ['--mode', 'development'], 
    {detached: true}
  );

const endAll = () => {
  process.kill(-child.pid);
  process.exit(0);
};

child.stderr.on('data', (data) => console.error(data.toString('utf8')));
child.stdout.on('data', (data) => console.log(data.toString('utf8')));

stdin.on('end', endAll);
process.on('SIGINT', endAll);
```

rockwood:
@vortizhe That's the issue I'm seeing. Thanks!

createdbypete:
https://github.com/webpack/webpack-cli/issues/885 seems related to this issue? This was reported way before v5 release though so I wonder what's changed in v5 to make this noticeable for you folks if they are related.

installer/templates/phx_assets/webpack.config.js
```diff
@@ -45,9 +37,19 @@ module.exports = (env, options) => {
...
+        ]
+      })
+    ],
+    optimization: {
+      minimizer: [
+        new TerserPlugin({}),
+        new CssMinimizerPlugin({})
+      ]
+    },
+    devtool: devMode ? 'eval-cheap-module-source-map' : undefined
```
sn3p, line 53: The [Terser plugin page](https://webpack.js.org/plugins/terser-webpack-plugin/#note-about-source-maps) mentions:

> **Note about source maps**
> Works only with `source-map`, `inline-source-map`, `hidden-source-map` and `nosources-source-map` values for the [devtool](https://webpack.js.org/configuration/devtool/) option.

Can't confirm but just a heads up.

installer/templates/phx_assets/package.json
sn3p: @cw789 did you uninstall and removed it from the plugins list? Of still need to require and keep it in the plugins?
The docs are rather vague.

installer/templates/phx_assets/webpack.config.js
cw789: Good point, which of the options should we pick then, if `eval-cheap-module-source-map` is really not accepted anymore?

installer/templates/phx_assets/package.json
cw789: I re-added it to the `package.json`, as we directly require it in the `webpack.config.js`.
It would also work with not having it in the `package.json` but that kind of seems wrong to me.

installer/templates/phx_assets/webpack.config.js
sn3p: Not my area of expertise, but could look into [SourceMapDevToolPlugin](https://webpack.js.org/plugins/source-map-dev-tool-plugin/) instead.

installer/templates/phx_assets/package.json
sn3p: The docs kinda gave me the impression it is bundled with webpack.
The same was done with [ProvidePlugin](https://webpack.js.org/plugins/provide-plugin/) which can be used as:

```js
const webpack = require("webpack");
new webpack.ProvidePlugin({ ... }),
```

cw789: Ahh. That's a different thing.
So we should remove it from `package.json` and in `webpack.config.js` - `require("webpack")` and use it as `new webpack.TerserPlugin`...

sn3p: No my point was I hoped this would work but it doesn't :(

sn3p: Looking at this the TerserPlugin is added by default:
https://webpack.js.org/configuration/mode/#mode-production (see plugins).

So adding `new TerserPlugin({})` might not be needed.

cw789: Seems to work only if i completely remove the `optimization` part.
But then we don't have minified CSS...

cw789: It'll work if I replace `new TerserPlugin({})` with `'...',`.
But that seems a bit weird to have a string escaped JS spread syntax there.

It's documented here: https://webpack.js.org/plugins/mini-css-extract-plugin/#minimizing-for-production
(but without the `,` at the end it's not working on my side)

And here: https://webpack.js.org/blog/2020-10-10-webpack-5-release/#changes-to-the-configuration

installer/templates/phx_assets/webpack.config.js
cw789: Not my expertise either.

My suggestion would be:
```
const webpack = require('webpack');
...
plugins: [
  new webpack.EvalSourceMapDevToolPlugin({
     exclude: ['./vendor/**/*.js']
   })
]
devtool: false
```

This would create inline source maps...

installer/templates/phx_assets/package.json
aptinio: If we use 
```
minimizer: [`...`, new CssMinimizerPlugin()]
```
We can remove both
```
"terser-webpack-plugin": "^5.x",
```
and
```
const TerserPlugin = require('terser-webpack-plugin');
```
from `package.json` and `webpack.config.js` respectively.

cw789:
So I again invested bit of time in this.
I don't see anymore the problem of livereload beeing triggered for every static assets copied each time by the webpack copy plugin.

I only see livereload beeing triggered twice every time.
For compilation starting & finished.
```
[webpack-cli] watching files for updates...
[webpack-cli] Compilation starting...
[debug] Live reload: priv/static/js/app.js
[debug] Live reload: priv/static/js/app.js
[webpack-cli] Compilation finished
```
But I don't know if this behavior was the same before.
I also don't see this as blocking issue for now?

So from my side, this is ready to be merged.
Thanks all.

installer/templates/phx_assets/webpack.config.js
cw789: @Gazler
I've replaced `eval-cheap-module-source-map` with `source-map` for now again.
Because Terser plugin doesn't work with `eval-cheap-module-source-map`. Just to keep it simple first.

But would be great if we could improve the build time later again.

Gazler: @cw789 I guess we only need Terser in production, right? So couldn't we disable Terser in dev and use cheap source maps for dev?

cw789: I think the point is it would at least need EvalSourceMapDevToolPlugin to have `eval-cheap-module-source-map` available again...

installer/templates/phx_assets/webpack.config.js
```diff
@@ -45,9 +36,19 @@ module.exports = (env, options) => {
...
+      new MiniCssExtractPlugin({ filename: '../css/app.css' }),
+      new CopyWebpackPlugin({
+        patterns: [
+          { from: 'static/', to: '../' }
+        ]
+      })
+    ],
+    optimization: {
+      minimizer: [
+        '...',
```
aptinio, line 48:
```suggestion
        `...`,
```
This should use backticks.

cw789: Why should this use JS template literal?

aptinio: The CssMinimizerWebpackPlugin [documentation](https://webpack.js.org/plugins/css-minimizer-webpack-plugin/) says so, but I haven't actually confirmed if single quotes work too.

dvic: I think they just meant to use markdown syntax there, it has no further meaning as far as I know, the docs on this page just use regular quotes:

https://webpack.js.org/blog/2020-10-10-webpack-5-release/

image

cw789: Ahh ok. That's fine. I saw the documentation too. They also miss the `,` colon there (without it's not working).
So of course it'll work with single quotes too.

There is no need for backticks (JS template literal) - they are used for e.g. `Hello ${name}`.

aptinio: Thanks for clarifying that :relaxed:

cw789:
@navinpeiris & @rockwood

Do you still have issues with livereload by using the latest CopyWebpackPlugin?

cw789:
@vortizhe

> I think webpack-cli 5.x isn't watching stdin any more so SIGINT signal do not close node process. The param `--watch-options-stdin` seems to do nothing.

Can you confirm this behavior with the latest webpack-cli?
I saw they have changed a lot which could resolve this issue.

jlgeering:
`--watch-options-stdin` doesn't do anything for me either, as `webpack` is still running when I quit `mix phx.server`

my current (wip) workaround is a variation of @vortizhe's script, which starts webpack directly from node instead of using the typical `webpack-cli`:

```javascript
const webpack = require('webpack');
const configurator = require('./webpack.config.js');

const config = configurator({}, {
    mode: 'development'
});

const compiler = webpack({
    ...config,
    mode: 'development',
});

console.log("starting webpack --mode development --watch");
const watching = compiler.watch({
    aggregateTimeout: 300,
}, (err, stats) => {
    if (err) {
        console.error(err);
        return;
    }

    console.log("webpack - build completed");

    console.log(stats.toString({
        chunks: false,  // Makes the build much quieter
        colors: true    // Shows colors in the console
    }));
});

process.stdin.on('end', function () {
    process.exit(0);
});
process.stdin.resume();
```

which I start with

```elixir
	watchers: [
		node: [
			"node",
			"watcher.js",
			cd: Path.expand("../assets", __DIR__)
		]
	]

```

it seem to work and it seem to be about the same as the "normal" webpack --watch, and as I get use to it, I start to even get to like it, as I feel like I have more control over the watch options and the logs

not sure this is the "right" thing to do for phx.new (`--watch-options-stdin` most probably isn't either), but it might at least be useful to someone else out there...

dvic:
I can confirm that we are experiencing the same problem (we are not using this branch but the same approach with `--watch-options-stdin`)

dvic:
I'm using this script:

```js
const process = require("process")
const child_process = require("child_process")

const stdin = process.openStdin()

const child = child_process.spawn(
  "node_modules/webpack/bin/webpack.js",
  ["--mode", "development", "--watch"],
  { detached: true },
)

const endAll = () => {
  process.kill(child.pid)
  process.exit(0)
}

child.stderr.on("data", (data) => console.error(data.toString("utf8")))
child.stdout.on("data", (data) => console.log(data.toString("utf8")))

stdin.on("end", endAll)

process.on("SIGINT", endAll)
```

and 

```elixir
  watchers: [
    node: [
      "watch.js",
      cd: Path.expand("../assets", __DIR__)
    ]
  ]
```

And this indeed seems to solve the problem (thanks @vortizhe).

DavidOliver:
As a light-touch, non-JS-config-changing workaround, I'm running the mix phx.server command with this shell script:

```
mix phx.server
kill $(pidof webpack)
```

This murders all processes named webpack after aborting iex, not just the one started by the mix command, but it's good enough for my purposes.

joshchernoff:
A note about deprecations: https://webpack.js.org/migrate/5/#make-sure-your-build-has-no-errors-or-warnings
Not sure if its worth adding `--trace-deprecation ` to the build command.

joshuataylor:
I've always had issues with zombie processes with webpack and iex, I actually have to run it outside of iex in another process.

chrismccord:
Do we have any info on why `watch-stdin` was removed? Having the process exit on EOF is common practice and it won't be great if the only way for us to do this is to ship another js file which watches for end.

mveytsman:
It looks like this is a bug in webpack. 

There's a draft PR fixing it (https://github.com/webpack/webpack-cli/pull/2217). I can confirm that webpack exits on STDIN closing with this PR, which I think will fix our problems.

chrismccord:
Awesome and a relief! Thanks for the digging Max!

aptinio:
A new version of webpack-cli has been release ([4.3.0](https://github.com/webpack/webpack-cli/blob/master/CHANGELOG.md#430-2020-12-25)) and I've confirmed that it now respects the `--watch-options-stdin` option.

josevalim:
Awesome, let's bump the versions and rebase and hopefully this will be soon good to go! :heart: Thanks everyone!

ruudk:
Don't forget to add back the `terser-webpack-plugin` dependency to `packages.json`, see  https://github.com/phoenixframework/phoenix/pull/4054#discussion_r521494617

szTheory:
Upgraded a project to Webpack 5 using these changes (with webpack-cli 4.3.0, `--watch` and `--watch-options-stdin`) and it worked including live reload in dev.

cw789:
> Awesome, let's bump the versions and rebase and hopefully this will be soon good to go! ❤️ Thanks everyone!

Done my work. Updated the description of the PR.
Thanks everybody!

installer/templates/phx_single/config/dev.exs
```diff
@@ -18,7 +18,8 @@ config :<%= @app_name %>, <%= @endpoint_module %>,
       "node_modules/webpack/bin/webpack.js",
       "--mode",
       "development",
-      "--watch-stdin",
+      "--watch",
+      "--watch-options-stdin",
```
dvic, line 22: @cw789 one more thing: shouldn't `templates/phx_umbrella/apps/app_name_web/config/dev.exs` also be updated?

cw789: Probably yes. I wasn't aware of it.
(My first time within the Phoenix repo ;)

Thanks. I'll fix it later.

cw789: Should I note that then there somewhere as comment?
Or am I allowed to?
I think it's not obvious single install vs. umbrella for newbees.
But possibly I missed something.

josevalim: No worries, we can also update it when we merge. Unfortunately we cannot add comments because they would appear in the user generated file. :)

dvic: I also missed it :) I missed also another instance: in `installer/templates/phx_assets/package.json` the `watch` command.

cw789: > `installer/templates/phx_assets/package.json`

True. It's again missing there.
Had it in there earlier - must have overwritten it. 🥴

cw789: @dvic 

> I also missed it :) I missed also another instance: in `installer/templates/phx_assets/package.json` the `watch` command.

I just checked that again, but for `package.json`: `"watch": "webpack --mode development --watch"` there is no need for `--watch-options-stdin`.
This is nowhere in use and could only be called by `npm run watch` where I don't see any need for `--watch-options-stdin`.

cw789:
Hopefully my last changes:
- Missing `--watch-options-stdin` in umbrella installation dev config `templates/phx_umbrella/apps/app_name_web/config/dev.exs`
- Upgrade `copy-webpack-plugin` to `^7.x`
- Upgrade `topbar` to `^1.x`

installer/templates/phx_single/config/dev.exs
dvic: Ah you are totally right! It's indeed only necessary when Phoenix is executing Webpack.

cw789:
Are we going foreward with this (webpack v5)?
Or at least with https://github.com/phoenixframework/phoenix/pull/4146?

Background:
New Phoenix projects created with Node.js v15 seems to be in a "broken" state. Even when it's not Phoenixs fault (JS packages issues), it is probably frustrating for new devs (I also already felt apart this trouble).

If webpack means to much maintenance trouble it's probably the better way to slim down JS things (webpack?, Sass?, ...).
I know this is not the place for discussion. ;)

Snowpack seems promising:
https://www.richardtaylor.dev/articles/replacing-webpack-with-snowpack-in-a-phoenix-application

josevalim:
@cw789 the plan is to go ahead with this but we need to balance between pushing to Webpack 5 too soon and that leading to frustration because the ecosystem did not catch up yet. I assume that, by the time we release a new Phoenix version, this will be safe, otherwise we will go with #4146. :)

If there is anything **broken** right now - not warnings but literally something does not work - then we should do the smallest PR possible to fix it and backport to the v1.5 branch. Otherwise these changes go to v1.6. :)

chrismccord:
@cw789 as José said please let us know what issues you are having on the current setup. I just built a fresh 1.5 app and webpack has no problems.

cw789:
Phoenix 1.5.7 wont work with Node.js v15.
≤ v14 seems to work...

@chrismccord Can you approve you've used Node.js ≤ v14?

I've not reproduced this issue, but seems so from reports in an issue & Elixir forum.

So the smallest PR is mentioning it will only work with Node.js ≤ v14 ...
Or by backporting the replacement of node-sass with sass and a minor upgrade of webpack v4.

chrismccord:
Sidebar, I actually plan on removing node-sass so that is perhaps the fastest way to fix this for all users?

chrismccord:
Note:

```
$ node -v
v15.2.1
```

cw789:
Relatet threads:
- https://elixirforum.com/t/npm-installation-for-fresh-phoenixliveview-project-failing-in-windows-10/37170
- https://github.com/phoenixframework/phoenix/issues/4126

chrismccord:
node-sass wouldn't be removed until 1.6, but the fact other folks might be having issues with webpack/node/et al is not new, so we need to tread carefully. I'm still seeing vast majority of folks have no issues on the current setup. Historically the js build tool ecosystem has been fragile, so moving to this PR is not guaranteed to be a panacea. We may cause as many issues as we solve for people, so we need to be sure that webpack v5 is golden across OS's, node versions, major packages, etc.

To be clear, we will be moving to webpack v5 as soon as makes sense, but I hope this gives insight into the slow roll to get there.

cw789:
Thanks Chris. I understand the rollout tempo & totally agree.
I also know the difficulties, so I would recommend to keep the complexity within the dark side (JS ecosystem ;) always as low as possible...

& all that is not a critic. It's the cause why I came across to Elixir.
Thanks for all!

cduruk:
Just tested this with node v15 and it does indeed work and fixes the current broken state of the [new project](https://github.com/phoenixframework/phoenix/issues/4126#issuecomment-765643594). Thought definitely agree, I'm sure this would break something new for someone soon, so not too bullish either way (as new projects work fine with node v14)

kivikakk:
I've been testing this for the last few days and it's been working okay, but I haven't gone too deep with it yet. (v14 less desirable for me as I'm on darwin/arm64 and afaik there's no native build for v14 there yet, only v15.)

skovmand-wh:
It would be good to change the minimum required Node.js version in the docs, which currently state:

> it's important to note that Phoenix requires version 5.0.0 or greater.

https://github.com/phoenixframework/phoenix/blob/master/guides/introduction/installation.md

lessless:
Thanks a lot to everyone involved. I hope there are simpler solutions coming our way. Maybe snowpack. 

Meanwhile, I had to set `url: false` for `css-loader` https://github.com/webpack-contrib/css-loader#url in order to fix image resolution

```
ERROR in ./css/app.scss
Module build failed (from ./node_modules/mini-css-extract-plugin/dist/loader.js):
ModuleBuildError: Module build failed (from ./node_modules/css-loader/dist/cjs.js):
```

in  `assets/css/app.scss`
```
@import "./components/betting_table";
```

in `assets/css/components/betting_table.scss`

```
background-image: url("/images/main-illustration.svg");
```


and the image itself is located at `assets/static/images/main-illustration.svg`



my webpack config

```
const path = require('path');
const glob = require('glob');
const TerserPlugin = require("terser-webpack-plugin");
const MiniCssExtractPlugin = require('mini-css-extract-plugin');
const CssMinimizerPlugin = require('css-minimizer-webpack-plugin');
const CopyWebpackPlugin = require('copy-webpack-plugin');
const webpack = require('webpack');

module.exports = (env, options) => {
  const devMode = options.mode !== 'production';

  return {
    optimization: {
      minimize: devMode ? undefined : true,
      minimizer: [
        `...`,
        new TerserPlugin({  parallel: true }),
        new CssMinimizerPlugin(),
      ]
    },
    entry: {
      './js/app.js': glob.sync('./vendor/**/*.js').concat(['./js/app.js'])
    },
    output: {
      filename: 'app.js',
      path: path.resolve(__dirname, '../priv/static/js')
    },
    module: {
      rules: [
        {
          test: /\.js$/,
          exclude: /node_modules/,
          use: {
            loader: 'babel-loader'
          }
        },
        {
          test: /\.s?css$/,
          use: [
            {
              loader: MiniCssExtractPlugin.loader,
              options: {
                publicPath: '/',
              },
            },
            {
              loader: 'css-loader',
              options: {url: false},
            },
            'postcss-loader',
            'sass-loader',
          ]
        }
      ]
    },
    plugins: [
      new MiniCssExtractPlugin({ filename: '../css/app.css' }),
      new CopyWebpackPlugin({patterns: [{ from: 'static/', to: '../' }]}),
      new webpack.ProvidePlugin({
        $: "jQuery",
        jQuery: "jQuery",
        Modal: "exports-loader?Modal!bootstrap/js/dist/modal",
        Util: "exports-loader?Util!bootstrap/js/dist/util"
      }),
    ],
    devtool: devMode ? 'source-map' : undefined
  }
};

```

aj-foster:
In case anyone is following along with these changes in their own project — and also as an FYI for the author and maintainers — I ran into an issue where the `glob` package was not available for use during a deploy (`npm run deploy`).

Original issue:

```
$ npm run deploy
> @ deploy /builds/pluralsight/experience/content/lab-tools/report-csp/report-csp/server/assets
> webpack --mode production
[webpack-cli] Failed to load '[redacted]/assets/webpack.config.js' config
[webpack-cli] Error: Cannot find module 'glob'
```

**Note** this only happened after I upgraded `css-minimizer-webpack-plugin` to `2.x`; presumably `glob` was available as a sub-dependency of this package in `1.x`. Either way, it might be good to include `glob` as a dev dependency in `package.json` since we use it in the webpack config.

At the end of the day, the changes I made are:

```diff
-    "css-minimizer-webpack-plugin": "^1.x",
+    "css-minimizer-webpack-plugin": "^2.x",
+    "glob": "^7.1.6",
```

Thank you all for your contributions and thoughtfulness.

installer/templates/phx_assets/webpack.config.js
```diff
@@ -1,30 +1,21 @@
...
     output: {
-      filename: '[name].js',
       path: path.resolve(__dirname, '../priv/static/js'),
+      filename: '[name].js',
       publicPath: '/js/'
     },
-    devtool: devMode ? 'eval-cheap-module-source-map' : undefined,
     module: {
       rules: [
         {
```
acconrad, line 21: Hey so I did my own webpack upgrade and I think you'll need a few other rules to ensure the most compatibility:

```js
        {
          test: /\.svg$/i,
          type: 'asset/inline'
        },
        {
          test: /\.(png|jpg|jpeg|gif)$/i,
          type: 'asset/resource',
        },
        {
          test: /\.(woff|woff2|eot|ttf|otf)$/i,
          type: 'asset/resource',
        },
```

This comes directly from the [Webpack 5 tutorial](https://webpack.js.org/guides/asset-management/#loading-fonts) to ensure fonts and images are loaded successfully as well

installer/templates/phx_assets/webpack.config.js
```diff
@@ -45,9 +36,19 @@ module.exports = (env, options) => {
...
+          { from: 'static/', to: '../' }
+        ]
+      })
+    ],
+    optimization: {
+      minimizer: [
+        '...',
+        new CssMinimizerPlugin()
+      ]
+    },
```
acconrad, line 51: I also think you will get compilation errors if you don't specify `target: 'nodeX.Y',` because the default option for `target` is `web` which will usually cause things like `http` and `url` to blow up upon upgrading. Since these assets are built in a node-like environment targeting Node over web will ensure Webpack successfully compiles.

acconrad (commented):
I did my own Webpack 5 upgrade and the only way to get all compilation errors to disappear was to implement a few additional changes on top of yours

installer/templates/phx_assets/webpack.config.js
cw789: I also noticed once that this is needed if you want to process assets by webpack.
But then I didn't touch any of these anymore.
It was not clear for me which file types people really need to have in here.

installer/templates/phx_assets/webpack.config.js
cw789: That's a thing I for myself would need to explore more in detail. I didn't get the consequences of this now.
For Phoenix deploys the `target` in my opinion should be `web` - not `node`.
But you might be right.

cw789:
Thanks @acconrad for your review.

@all
I stopped ongoing maintenance of this PR past 3 months ago.
Of course in the meanwhile Node & Webpack are moving.
So this is no up-to-date reference anymore for using webpack v5 with Phoenix.

When someone tell me for sure this will get relevance for Phoenix again, of course I could have a fresh look.

installer/templates/phx_assets/webpack.config.js
acconrad: makes sense - i think it's pretty safe to say that in general css, js, images, and fonts are all fair game

installer/templates/phx_assets/webpack.config.js
acconrad: I thought the same too but when I deployed with `web` I was getting errors around `url` and `http` and you'd have to set resolutions that resolve to `false` for both. Since this is a `node` environment for how we build our assets if you change it to `node` everything will resolve and pass.

acconrad:
@cw789 I'm happy to take this over! You can either close this out and I can resume it in my own branch or I can just take this over from this open PR? @josevalim @chrismccord any recommendations?

Yattol:
> Just tested this with node v15 and it does indeed work and fixes the current broken state of the [new project](https://github.com/phoenixframework/phoenix/issues/4126#issuecomment-765643594). Thought definitely agree, I'm sure this would break something new for someone soon, so not too bullish either way (as new projects work fine with node v14)

It breaks also with node 16, which is the default version if you install it via brew, which is quite common for OS X users.
(https://github.com/phoenixframework/phoenix/issues/4126#issuecomment-827072525)

I think it should at least be mentioned in the getting started guide, it's quite frustrating to follow a list of instructions and see they don't work, it doesn't give a good 'first impression' to newcomers.

guidotripaldi:
> I think webpack-cli 5.x isn't watching stdin any more so SIGINT signal do not close node process. The param `--watch-options-stdin` seems to do nothing.

with webpack v.5 we have to give both `--watch` and `--watch-options-stdin` params to its CLI to have it working as expected (i.e. watching for changes and for SIGINT):

```
# config/dev.exs

use Mix.Config

config :myapp_web, MyAppWeb.Endpoint,
  http: [port: 4000],
  debug_errors: true,
  code_reloader: true,
  check_origin: false,
  watchers: [
    node: [
      "node_modules/webpack/bin/webpack.js",
      "--mode",
      "development",
      "--watch",
      "--watch-options-stdin",
     cd: Path.expand("../assets", __DIR__)
    ]
  ]

...
```

acconrad:
@cw789 I've also created a new version at #4310

PR #4055, opened 2020-11-02
Add Ecto.Enum support to generators

TheFirstAvenger:
Add `Ecto.Enum` support to generators. Format:

```
mix phx.gen.schema Blog Post posts body:text status:enum:published:unpublished:deleted
```

Generates html `select` tags with appropriate values in `phx.gen.html` and `phx.gen.live`:

```
<%= select f, :status, Ecto.Enum.values(MyApp.Blog.Post, :status) %>
```


aaronrenner:
This is exciting to see! What do you think about adding an enum field to the `phx.gen.*` integration tests in these places?

https://github.com/phoenixframework/phoenix/blob/b5580e9cd1c78e0460d2df30807f4f99bd8b3051/integration_test/test/code_generation/app_with_defaults_test.exs#L30
https://github.com/phoenixframework/phoenix/blob/b5580e9cd1c78e0460d2df30807f4f99bd8b3051/integration_test/test/code_generation/app_with_defaults_test.exs#L53
https://github.com/phoenixframework/phoenix/blob/b5580e9cd1c78e0460d2df30807f4f99bd8b3051/integration_test/test/code_generation/app_with_defaults_test.exs#L77
https://github.com/phoenixframework/phoenix/blob/b5580e9cd1c78e0460d2df30807f4f99bd8b3051/integration_test/test/code_generation/app_with_defaults_test.exs#L89
https://github.com/phoenixframework/phoenix/blob/b5580e9cd1c78e0460d2df30807f4f99bd8b3051/integration_test/test/code_generation/app_with_defaults_test.exs#L113
https://github.com/phoenixframework/phoenix/blob/b5580e9cd1c78e0460d2df30807f4f99bd8b3051/integration_test/test/code_generation/app_with_defaults_test.exs#L141

The integration tests will catch any formatting issues in the generated code, as well as any accidental failures in the generated test suite.

TheFirstAvenger:
@aaronrenner Thanks! I hadn't seen those before, and adding those tests pointed out the need to increment the Ecto deps in the generated mix.exs to 3.5 😁. Changes pushed.

guides/ecto.md
```diff
@@ -470,7 +470,7 @@ defmodule HelloPhoenix.MixProject do
     [
       {:phoenix, "~> 1.4.0"},
       {:phoenix_ecto, "~> 4.0"},
-      {:ecto_sql, "~> 3.4.4"},
+      {:ecto_sql, "~> 3.5.0"},
```
josevalim, line 473: Let's bump to `~> 3.5` instead.

lib/mix/phoenix/schema.ex
```diff
@@ -164,12 +167,14 @@ defmodule Mix.Phoenix.Schema do
   end
 
   @doc """
-  Generates some sample params based on the parsed attributes.
+  Generates some sample params based on the parsed attributes. LiveView cannot be given an invalid
+  value in a select field so we drop enum params for the invalid_live params.
   """
-  def params(attrs, action \\ :create) when action in [:create, :update] do
+  def params(attrs, action \\ :create) when action in [:create, :update, :invalid, :invalid_live] do
```
josevalim, line 173: Please take a look at 9a019f607. It solved these problems differently. In any case, `invalid_form_value` already sets it to nil, which is enough to make it invalid, no? Otherwise, I suggest we change the generator to include a select with blank, something like `<%= select f, :foo, prompt: "Choose" %>`.

lib/mix/phoenix/schema.ex
```diff
@@ -322,13 +329,36 @@ defmodule Mix.Phoenix.Schema do
...
   defp validate_attr!({_name, type} = attr) when type in @valid_types, do: attr
+  defp validate_attr!({_name, {:enum, vals}} = attr), do: validate_enum_attr!(vals, attr)
   defp validate_attr!({_name, {type, _}} = attr) when type in @valid_types, do: attr
   defp validate_attr!({_, type}) do
     Mix.raise "Unknown type `#{inspect type}` given to generator. " <>
               "The supported types are: #{@valid_types |> Enum.sort() |> Enum.join(", ")}"
   end
 
+  defp raise_enum_missing_vals! do
+    if :erlang.phash2(1, 1) == 0 do
```
josevalim, line 342: Is this to appease Dialyzer? We don't run it here, so we can remove it. :)

lib/mix/phoenix/schema.ex
```diff
@@ -283,6 +289,7 @@ defmodule Mix.Phoenix.Schema do
   defp type_to_default(key, t, :update) do
     case t do
         {:array, _}     -> []
+        {:enum, values} -> values |> Atom.to_string() |> String.split(":") |> tl() |> hd() |> String.to_atom()
```
josevalim, line 292: maybe call `translate_enum_values |> hd()` instead? :)

priv/templates/phx.gen.schema/migration.exs
```diff
@@ -1,10 +1,15 @@
-defmodule <%= inspect schema.repo %>.Migrations.Create<%= Macro.camelize(schema.table) %> do
+<%
+translate = fn
+  {:enum, _} -> :string
+  other -> other
+end
```
josevalim, line 5: Please create an actual function on `Mix.Phoenix.Schema`. :)

priv/templates/phx.gen.schema/schema.ex
```diff
@@ -1,11 +1,16 @@
-defmodule <%= inspect schema.module %> do
+<%
+translate = fn
+  {:enum, opts} -> ~s|Ecto.Enum, values: #{inspect Keyword.get(opts, :values)}|
+  other -> inspect other
+end
```
josevalim, line 5: Please create an actual function on `Mix.Phoenix.Schema`. :)

lib/mix/phoenix/schema.ex
TheFirstAvenger: Yes, it was. I have rewritten it in a clean way that avoids dialyzer warnings without phash2.

josevalim:
Btw @TheFirstAvenger, please ping me when you think this is ready for a last review!

TheFirstAvenger:
@josevalim @aaronrenner After rebasing, I am getting the following error on all 24 of the integration tests run with `mix test --include database`. I verified I am getting it on master now too:

```
 23) test phx.gen.html has a passing test suite (Phoenix.Integration.CodeGeneration.AppWithDefaultsTest)
     test/code_generation/app_with_defaults_test.exs:49
     ** (RuntimeError) mix command failed with exit code: 1

     mix phx.new /Users/binns/git/phoenix/installer/tmp/Csa7xy11T4/app_with_defaults/phx_blog --dev --no-install

     ** (Mix) --dev projects must be generated inside Phoenix directory


     Options
     cd: /Users/binns/git/phoenix/integration_test
     env: []

     code: with_installer_tmp("app_with_defaults", fn tmp_dir ->
     stacktrace:
       (phoenix_integration 0.1.0) test/support/code_generator_case.ex:36: Phoenix.Integration.CodeGeneratorCase.mix_run!/3
       (phoenix_integration 0.1.0) test/support/code_generator_case.ex:17: Phoenix.Integration.CodeGeneratorCase.generate_phoenix_app/3
       test/code_generation/app_with_defaults_test.exs:51: anonymous fn/1 in Phoenix.Integration.CodeGeneration.AppWithDefaultsTest."test phx.gen.html has a passing test suite"/1
       (phoenix_integration 0.1.0) test/support/code_generator_case.ex:108: Phoenix.Integration.CodeGeneratorCase.with_installer_tmp/3
       test/code_generation/app_with_defaults_test.exs:50: (test)
```

josevalim:
@TheFirstAvenger run `mix archive.uninstall phx_new` and they should pass.

lib/mix/phoenix/schema.ex
TheFirstAvenger: @josevalim The reason I went down the path of separating out enums in live tests is that `render_change` throws an argument error if you try to set anything but one of the valid options in the select field (including trying to set nil):

```
       3) test Index saves new post (DefaultMssqlAppWeb.PostLiveTest)
          test/default_mssql_app_web/live/post_live_test.exs:26
          ** (ArgumentError) value for select "post[status]" must be one of ["unpublished", "published", "deleted"], got: nil
          code: |> render_change() =~ "can&apos;t be blank"
          stacktrace:
            (phoenix_live_view 0.14.7) lib/phoenix_live_view/test/live_view_test.ex:876: Phoenix.LiveViewTest.call/2
            test/default_mssql_app_web/live/post_live_test.exs:36: (test)
```

I discussed this with @chrismccord when this first came up, and the best option we came up with was to simply drop enums from the list of invalid values when dealing with the LiveView tests. It functions correctly on dead views in `phx.gen.html`, so there was no reason to drop them there, hence splitting out `:invalid` and `:invalid_live`. I am open to any suggestions on the path forward.

josevalim: What do you think about my suggestion of adding a prompt to the `select`? This way we can continue submitting blank strings (which is the default invalid value today) and we don't have to change the tests.

TheFirstAvenger: @josevalim apologies, I had tried using a prompt prior, and then re-tried it after you suggested it, but forgot to mention that the LiveView `render_change` also raises with a prompt present when given a nil value:

```
       1) test Index saves new post (RainyDayWeb.PostLiveTest)
          apps/rainy_day_web/test/rainy_day_web/live/post_live_test.exs:26
          ** (ArgumentError) value for select "post[status]" must be one of ["unpublished", "published", "deleted"], got: ""
          code: |> render_change() =~ "can&apos;t be blank"
          stacktrace:
            (phoenix_live_view 0.14.7) lib/phoenix_live_view/test/live_view_test.ex:876: Phoenix.LiveViewTest.call/2
            test/rainy_day_web/live/post_live_test.exs:36: (test)
```

josevalim: Hrm, that sounds like a bug in LiveView, let me try to reproduce it.

josevalim:
Ping in case there is something I can help with. :)

TheFirstAvenger:
@josevalim Sorry, been busy the past few weeks. I went back and looked at this, and I think I was using "placeholder" instead of "prompt" when trying it, because it is now working as you suggested with a prompt. I have pushed the cleaner version.

The current issue is that [this line](https://github.com/phoenixframework/phoenix/pull/4055/commits/6fabc333d30e7b1a4515ef06bcbfba00ac01d9a1#diff-2ed329afaff5aa5d07f7ac2a7c8f6e8bb4ff65609152ca4059a65adf2edea972R113) is now generating a line longer than the formatter likes at [this line](https://github.com/phoenixframework/phoenix/blob/master/priv/templates/phx.gen.context/test_cases.exs#L32) in the generated tests, so the formatter check fails. The same thing is happening on [this line](https://github.com/phoenixframework/phoenix/blob/master/priv/templates/phx.gen.live/live_test.exs#L8) in the generated post_live_test.exs file. @chrismccord and I discussed the possibility of updating `Mix.Phoenix.to_text` to be smarter and wrap the text when necessary, however that gets pretty messy as you would need to pass in the current column count left of the function call so it knows how much to start the current line at to know if/how it should wrap. Thoughts on how to approach this?

josevalim:
@TheFirstAvenger Could to_text just generate something multiline by default? If you don't want to go down this rabbit hole, make the field name shorter for now. :)

aaronrenner:
@TheFirstAvenger Here's another place where we generate a multi-line map that comes out properly formatted:

https://github.com/phoenixframework/phoenix/blob/31c75ebe4437e7f7d172a70155f322b85aa4032c/priv/templates/phx.gen.context/fixtures.ex#L5-L7

It's not very pretty, but I wonder if that would solve the problem you're running into.

TheFirstAvenger:
@josevalim If I am thinking about it correctly, if we always multiline, then the formatter check will fail when the line doesn't need to be multiline? Either way I shortened the field and values on the failing test for now and things are passing. All set for a last review.

josevalim:
@TheFirstAvenger the formatter respects newlines in data structures and keeps them.

josevalim:

PR #4061, opened 2020-11-06
Update example Dockerfile and add some notes

nathanl:
- Use an Elixir 1.11 image, built by the hexpm team
- Break apart build steps for better caching (eg, changing
  `config/prod.exs` does not require re-running `mix deps.get`)
- Add notes about configuring a containerized application

### DIFF
--- a/guides/deployment/releases.md
+++ b/guides/deployment/releases.md
@@ -202,6 +202,11 @@ COPY assets/package.json assets/package-lock.json ./assets/
 RUN npm --prefix ./assets ci --progress=false --no-audit --loglevel=error
 
 COPY priv priv
+
+# Note: if your project uses a tool like https://purgecss.com/,
+# which customizes asset compilation based on what it finds in
+# your Elixir templates, you will need to move the asset compilation step
+# down so that `lib` is available.
 COPY assets assets
 # use webpack to compile npm dependencies - https://www.npmjs.com/package/webpack-deploy
 RUN npm run --prefix ./assets deploy
--- a/guides/deployment/releases.md
+++ b/guides/deployment/releases.md
@@ -191,9 +191,11 @@ ENV MIX_ENV=prod
 COPY mix.exs mix.lock ./
 RUN mix deps.get --only $MIX_ENV
 RUN mkdir config
-# if our compile-time config changes, we may need to
-# recompile dependencies
-COPY config/config.exs config/prod.exs config/
+# Dependencies sometimes use compile-time configuration. Copying
+# these compile-time config files before we compile dependencies
+# ensures that any relevant config changes will trigger the dependencies
+# to be re-compiled.
+COPY config/config.exs config/$MIX_ENV.exs config/
 RUN mix deps.compile
 
 # build assets
@@ -244,5 +246,5 @@ At the end, you will have an application in `/app` ready to run as `bin/my_app s
 
 A few points about configuring a containerized application:
 
-- If you run your app in a container, the endpoint needs to listen on a "public" IP address like `0.0.0.0` so that the app can be reached from outside the container. Whether the host should publish the container's ports to its own public IP or to localhost depends on your needs.
-- The more configuration you can provide at runtime (using `config/runtime.exs`), the more reusable your images will be across environments. In particular, secrets like database credentials and your endpoint's `:secret_key_base` should not be compiled into the image, but rather should be provided when creating containers based on that image.
+- If you run your app in a container, the `Endpoint` needs to be configured to listen on a "public" `:ip` address (like `0.0.0.0.0.0.0.0`) so that the app can be reached from outside the container. Whether the host should publish the container's ports to its own public IP or to localhost depends on your needs.
+- The more configuration you can provide at runtime (using `config/runtime.exs`), the more reusable your images will be across environments. In particular, secrets like database credentials and API keys should not be compiled into the image, but rather should be provided when creating containers based on that image. This is why the `Endpoint`'s `:secret_key_base` is configured in `config/runtime.exs` by default.
--- a/guides/deployment/releases.md
+++ b/guides/deployment/releases.md
@@ -172,10 +172,10 @@ Elixir releases work well with container technologies, such as Docker. The idea 
 Here is an example Docker file to run at the root of your application covering all of the steps above:
 
 ```docker
-FROM elixir:1.9.0-alpine AS build
+FROM hexpm/elixir:1.11.2-erlang-23.1.2-alpine-3.12.1 as build
 
 # install build dependencies
-RUN apk add --no-cache build-base npm git python
+RUN apk add --no-cache build-base npm git python3
 
 # prepare build dir
 WORKDIR /app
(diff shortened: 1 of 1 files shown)


nathanl:
I'm not an expert in this area, but these are some things I've learned recently. Critique is welcome. 😄

guides/deployment/releases.md
```diff
@@ -189,26 +189,36 @@ ENV MIX_ENV=prod
...
 # install mix dependencies
 COPY mix.exs mix.lock ./
-COPY config config
-RUN mix do deps.get --only $MIX_ENV, deps.compile
+RUN mix deps.get --only $MIX_ENV
+RUN mkdir config
+# if our compile-time config changes, we may need to
+# recompile dependencies
+COPY config/config.exs config/prod.exs config/
+run mix deps.compile
```
tomciopp, line 197: @nathanl RUN should be capitalized here

guides/deployment/releases.md
```diff
@@ -189,26 +189,36 @@ ENV MIX_ENV=prod
...
 
 COPY priv priv
 COPY assets assets
+# use webpack to compile npm dependencies - https://www.npmjs.com/package/webpack-deploy
 RUN npm run --prefix ./assets deploy
 RUN mix phx.digest
 
-# compile and build release
+# compile and build the release
 COPY lib lib
```
dvic, line 211: The only thing the be aware here is that the build will 'not work' if you you are purging your CSS (i.e., removing the unused CSS classes) because it will miss the source code from `lib/my_web`. However this is something that is already an issue with the current version of the `Dockerfile` but I just wanted to mention it here, maybe we should leave a comment about this in the `Dockerfile`?

nathanl: @dvic Sorry, I don't understand the problem yet. 😊 In an example project, if I delete some lines from `assets/css/somefile.css`, that busts the cache for the `COPY assets assets` step, which re-triggers the npm compilation step and should mean that the everything from there on is re-run. What would be broken in that case?

dvic: No worries. I'll try to elaborate. The problem only occurs when you use something like https://purgecss.com. This plugin scans your source code and keeps track of all the "words" it encounters. It then discards any css classes that are not in this set of seen words (in order to minimize the css). The problem we encountered with the above Docker build is that `COPY lib lib` is being performed *after* `npm run deploy` has run, therefore, a library like `purgecss` won't see the Elixir source code and will miss all the classes used in the `.eex` templates.

nathanl: @dvic Got it, thanks for explaining. I added a comment.

guides/deployment/releases.md
```diff
@@ -189,26 +189,41 @@ ENV MIX_ENV=prod
 
 # install mix dependencies
 COPY mix.exs mix.lock ./
-COPY config config
-RUN mix do deps.get --only $MIX_ENV, deps.compile
+RUN mix deps.get --only $MIX_ENV
+RUN mkdir config
+# if our compile-time config changes, we may need to
+# recompile dependencies
```
josevalim, line 195: Is this something done automatically for us? Or would we need to trigger it manually?

guides/deployment/releases.md
```diff
@@ -189,26 +189,41 @@ ENV MIX_ENV=prod
...
+COPY config/config.exs config/prod.exs config/
```
josevalim, line 196: Should we copy `config/$MIX_ENV.exs` instead? Is that possible?

guides/deployment/releases.md
```diff
@@ -221,7 +236,13 @@ COPY --from=build --chown=nobody:nobody /app/_build/prod/rel/my_app ./
...
-CMD ["bin/my_app", "start"]
+ENTRYPOINT ["bin/my_app"]
+CMD ["start"]
 ```
 
 At the end, you will have an application in `/app` ready to run as `bin/my_app start`.
+
+A few points about configuring a containerized application:
+
+- If you run your app in a container, the endpoint needs to listen on a "public" IP address like `0.0.0.0` so that the app can be reached from outside the container. Whether the host should publish the container's ports to its own public IP or to localhost depends on your needs.
```
josevalim, line 247: If I want to change this, where do I go? What are our defaults? :)

guides/deployment/releases.md
```diff
@@ -221,7 +236,13 @@ COPY --from=build --chown=nobody:nobody /app/_build/prod/rel/my_app ./
...
+- The more configuration you can provide at runtime (using `config/runtime.exs`), the more reusable your images will be across environments. In particular, secrets like database credentials and your endpoint's `:secret_key_base` should not be compiled into the image, but rather should be provided when creating containers based on that image.
```
josevalim, line 248: Let's also mention secret_key_base is part of config/runtime.exs by default. :)

josevalim:
Btw, after this is done, I would like to add a `mix phx.gen.dockerfile`. Is this something you have interest working on @nathanl or can I make it generally available for anyone to pick it up? Thanks :heart:!

guides/deployment/releases.md
```diff
@@ -189,26 +189,41 @@ ENV MIX_ENV=prod
...
+RUN mkdir config
+# if our compile-time config changes, we may need to
+# recompile dependencies
+COPY config/config.exs config/prod.exs config/
+RUN mix deps.compile
 
 # build assets
 COPY assets/package.json assets/package-lock.json ./assets/
+# install all npm dependencies from scratch
 RUN npm --prefix ./assets ci --progress=false --no-audit --loglevel=error
```
dbernheisel, line 202: I don't want to over-complicate an example, but oftentimes the node version needs to be specified for consistent builds, and the easiest way I can control this is by building the frontend assets in a separate container and copying the artifacts over.

For example:

```dockerfile
## BACKEND
FROM hexpm/elixir:1.11.2-erlang-23.1.4-alpine-3.12.1 AS builder
 
# ...

## FRONTEND

FROM node:14.14.0-alpine AS frontend
# PurgeCSS needs to see the Elixir stuff
COPY lib/my_app_web ./lib/my_app_web

# The package.json may include references to JavaScript inside Elixir dependencies. Copy them over:
COPY --from=builder /app/deps/phoenix ./deps/phoenix
COPY --from=builder /app/deps/phoenix_html ./deps/phoenix_html
COPY --from=builder /app/deps/phoenix_live_view ./deps/phoenix_live_view

COPY assets/package.json assets/package-lock.json ./assets/
RUN npm --prefix assets ci --progress=false --no-audit --loglevel=error
COPY assets ./assets
RUN npm --prefix assets run deploy

## APP

FROM builder AS app
COPY --from=frontend /app/priv/static ./priv/static
RUN mix phx.digest

COPY priv/gettext ./priv/gettext
COPY priv/repo ./priv/repo
```

Unfortunately doing it this way means I have to copy each folder under `/priv` separately, unless someone knows a more straightforward way to copy-merge them from the multi-stage builds.

WDYT of including a comment to mention frontend assets being a multi-stage build step if needing to control the nodejs version?

josevalim:
> Btw, after this is done, I would like to add a mix phx.gen.dockerfile.

We are having second thoughts if this is worth compared to the current guides, so let's put this on the backburner for now. :)

guides/deployment/releases.md
nathanl: It's now specified in `config/runtime.exs` in a new app - https://github.com/phoenixframework/phoenix/blob/master/installer/templates/phx_single/config/runtime.exs#L21
I'm updating the comment to be a little more specific.

guides/deployment/releases.md
nathanl: Yep, that works. 😄

guides/deployment/releases.md
nathanl: Clarifying.

guides/deployment/releases.md
nathanl: @alfredbaudisch would you mind making a followup PR after this one is merged? You can tag me on it and I'll try to chime in.

nathanl: @dbernheisel would you mind making a followup PR after this one is merged? You can tag me on it and I'll try to chime in.

nathanl:
@josevalim I pushed a commit incorporating your suggestions - see what you think.

josevalim:

guides/deployment/releases.md
```diff
@@ -172,10 +172,10 @@ Elixir releases work well with container technologies, such as Docker. The idea
 Here is an example Docker file to run at the root of your application covering all of the steps above:
 
 ```docker
-FROM elixir:1.9.0-alpine AS build
+FROM hexpm/elixir:1.11.2-erlang-23.1.2-alpine-3.12.1 as build
 
 # install build dependencies
-RUN apk add --no-cache build-base npm git python
+RUN apk add --no-cache build-base npm git python3
```
joshchernoff, line 178: Just out of curiosity, what depends on python?

mcrumm: Node.js uses gyp (via [node-gyp](https://github.com/nodejs/node-gyp)) for compiling native addons.

joshchernoff: Thank you 🙏

PR #4062, opened 2020-11-07
Add integration test for development workflow

aaronrenner:
This ports the recently removed `mix/tasks/phx.new_test.exs` into the integration test suite so we can ensure the development workflow and code compilation works as expected. Here is the previous test for reference:

https://github.com/phoenixframework/phoenix/blob/8e13ccc2f95911655bb9287d49126cb329daa18b/test/mix/tasks/phx.new_test.exs#L44-L92

Since this is running the test in a separate beam instance, we have to send requests using a real http client instead of directly calling the endpoint in elixir. The code to start `mix phx.server` without leaving a zombie process came from @josevalim's comment here: https://github.com/phoenixframework/phoenix/pull/4027#issuecomment-717788510

### DIFF
--- a/integration_test/test/code_generation/app_with_no_options_test.exs
+++ b/integration_test/test/code_generation/app_with_no_options_test.exs
@@ -1,6 +1,8 @@
 defmodule Phoenix.Integration.CodeGeneration.AppWithNoOptionsTest do
   use Phoenix.Integration.CodeGeneratorCase, async: true
 
+  @epoch {{1970, 1, 1}, {0, 0, 0}}
+
   test "newly generated app has no warnings or errors" do
     with_installer_tmp("app_with_no_options", fn tmp_dir ->
       {app_root_path, _} =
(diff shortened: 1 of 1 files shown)


integration_test/test/code_generation/app_with_no_options_test.exs
```diff
@@ -17,4 +19,78 @@ defmodule Phoenix.Integration.CodeGeneration.AppWithNoOptionsTest do
...
+      assert response.status_code == 200
+      assert response.body =~ "PhxBlog"
+
+      assert File.stat!(Path.join(app_root_path, "lib/phx_blog_web/views/page_view.ex")) > @epoch
+
+      # Ensure /priv static files are copied
+      assert File.exists?(Path.join(app_root_path, "priv/static/js/phoenix.js"))
+
+      assert_passes_formatter_check(app_root_path)
+      assert_tests_pass(app_root_path)
```
josevalim, line 52: We already asserted this in the previous test, so I wouldn't assert again. Actually, I would even go in the opposite direction which is to merge this test with the previous one to cut build times. I would do this:

```elixir
      {app_root_path, _} =
        generate_phoenix_app(tmp_dir, "phx_blog", [
          "--no-html",
          "--no-webpack",
          "--no-ecto",
          "--no-gettext",
          "--no-dashboard"
        ])
      assert_no_compilation_warnings(app_root_path)
      assert_dev_workflow(app_root_path)
      assert_passes_formatter_check(app_root_path)
      assert_tests_pass(app_root_path)
```

WDYT?

josevalim:
@aaronrenner this is fantastic! I have just added a suggestion about merging the tests to cut build time.

I think with this the infrastructure is ready and we can start porting `mix phx.gen.auth`. WDYT?

integration_test/test/code_generation/app_with_no_options_test.exs
aaronrenner: I was originally going to do that, but the dev workflow test can't have `--no-html` because that causes a GET request to `"http://localhost:4000/"` return a 404 because no routes are generated with this flag. I guess the question would be, do we want to assert that the dev workflow gets a 200 response?

If we end up leaving it as a separate test, do you think it still makes sense to extract a `assert_dev_workflow/1` helper?

aaronrenner:
@josevalim I completely agree. I just started moving the integration test suite in phx.gen.auth to use the same helpers we're using here to see if there's any other issues we'll run into during the port.

integration_test/test/code_generation/app_with_no_options_test.exs
josevalim: I think we can remove —no-html to merge them. :)

josevalim: Oh, we need to test without said option. Should we merge this into the test which does consider most options?

aaronrenner: Is this where you're thinking about moving it?

https://github.com/phoenixframework/phoenix/blob/3392efd1821b53275837066f94e8d69f519055f7/integration_test/test/code_generation/app_with_defaults_test.exs#L14-L22

It will have to go in one of the database tests because this app was not generated with `--no-ecto` and I don't think the http request won't succeed if it was started without a database. We don't have any other tests with `--no-ecto` but without `--no-html`. Do you mind if the development workflow test the test suite to be run with `--include database`?

Regarding the `assert_dev_workflow/1` helper function, do you see this being called in multiple tests? It's currently referencing project specific paths, so if I was going to do it in a generic way to be used in multiple tests, I'd probably have to have the code figure out which files need to be modified for the tests. If we're only going to have the dev workflow code in 1 test, I'm wondering if it would be better to leave it inline.

josevalim:
We have explored all options so I think this is the best approach after all. Thanks for the discussion 

PR #4064, opened 2020-11-10
Parse installer templates at compile time

aaronrenner:
This PR changes the installer eex templates to be parsed at compile time instead of runtime. This came about because I noticed a non-trivial amount of time in the phx.gen.auth integration tests were spent on parsing the larger eex templates. This changes phx_new's templates to be generated at runtime instead of compile time and decreases the amount of time the installer tests take from 7 seconds to 5.5 seconds. Integration test times appear to fluctuate more, so unfortunately I haven't seen a noticeable decrease with this change like I had hoped. However, I'm not aware of any drawbacks of having template parsing occur at compile time, so I thought I'd submit this anyways.

/cc @josevalim


installer/lib/phx_new/generator.ex
```diff
@@ -26,9 +26,22 @@ defmodule Phx.New.Generator do
...
+
+          if format in [:config, :prod_config, :eex] do
+            compiled = EEx.compile_file(path)
+
+            quote do
+              @external_resource unquote(path)
+              def render(unquote(name), unquote(source), assigns) do
+                {result, _} = Code.eval_quoted(unquote(Macro.escape(compiled)), [assigns: assigns], [file: unquote(path)])
+                result
+              end
```
josevalim, line 38: You shouldn't need to eval but instead inject the compile time. Something like this:

```suggestion
            compiled = EEx.compile_file(path)

            quote do
              @external_resource unquote(path)
              @file unquote(path)
              def render(unquote(name), unquote(source), assigns), do: unquote(compiled)
```

Have you tried it? :)

josevalim:
Great call! I have added some comments. :)

installer/lib/phx_new/generator.ex
aaronrenner: Thanks for the feedback. That looks much nicer! I also just learned about `@file`. :)

I just tried your suggestion, but I get the following error:

```
== Compilation error in file lib/phx_new/umbrella.ex ==
** (CompileError) templates/phx_umbrella/mix.exs:1: expected "assigns" to expand to an existing variable or be part of a match
    (elixir 1.11.1) expanding macro: Kernel.var!/1
    templates/phx_umbrella/mix.exs:1: Phx.New.Umbrella.render/3
    /home/aaron/projects/phoenix/installer/lib/phx_new/umbrella.ex:1: Phx.New.Generator.__before_compile__/1
```

Do you happen to have an idea of what's wrong?

josevalim: I don't know. I will have to play with it locally. So I will merge this and give it a try. :)

josevalim:

installer/lib/phx_new/generator.ex
josevalim: I got it!

So variables inside a quote have their own context:

```elixir
iex> quote do: assigns
{:assigns, [], Elixir}
```

In this case it is `Elixir` but inside a module it would be module name. `EEx` expects a variable `assigns`, outside of any special context. The solution is simply to wrap it on `var!(assigns)`. I have also added a guard to avoid warnings. :)

josevalim: Btw, now that we are compiling it, the integration tests should be even faster. :)

aaronrenner: Ah, that makes sense. Thank you!

Next, I'm going to update the `phx.gen.*` code to parse their templates at compile time too. I'm trying to limit the amount that `phx.gen.auth`'s test suite slows down the phoenix integration tests.

josevalim: > I'm trying to limit the amount that phx.gen.auth's test suite slows down the phoenix integration tests.

I don't think this will be possible because we allow the templates to be overridden locally, so it is always dynamic, unfortunately.

If times are an issue, we should consider a couple things:

1. Minimize the number of integration tests from gen.auth. I have not checked the current implementation but we will be fine with one test per database adapter?

2. Share the initially generated code, like I have mentioned in the past. For example, we don't need to generate two apps, one to test the scaffold and another for phx.gen.auth. They can probably both be done on top of the same project?

aaronrenner: Unfortunately there are quite a few different permutations of `phx.gen.auth`. We'll need test for all 3 databases + tests for the two additional hashing libraries + binary id support, etc. I spent the evening trying to cache the project's scaffolding but for some reason recompiles were being triggered and slowed the test processes down.

I think my plan will be to set up a branch that merges phx.gen.auth into phoenix. With that branch set up we'll be able to iterate there different ways to make the test suite faster (caching, eliminating tests, etc) and be in a position to merge the generator if we need to release the next version of phoenix.

josevalim: @aaronjensen fwiw, I also wouldn't mind merging the generators first and then working on the integration tests later one by one.

Regarding the integration tests, one option is that, instead of treating those as permutations, we can treat them as combinations. So the tests are:

1. defaults (postgres + default hashing lib)
2. mysql (other hashing lib + binary id)
3. mssql (other hashing lib)

I don't even think we need to test all hashing lib combinations, since we rely on an existing contract, and the integration tests are meant to be smoke, but given we need to test the three databases, we can include them in.

PR #4069, opened 2020-11-17
Fix code style

c4710n:
According to https://github.com/lexmag/elixir-style-guide#modules, the
order should be use/import/alias/require. So, I adjust this piece of
code to follow the rule.


josevalim:
Phoenix doesn't follow any particular style guide besides what is dictated the Elixir code formatted, which we are slowing adopting. Thanks!

c4710n:
Thanks for letting me know. ;)

PR #4072, opened 2020-11-18
Ability to run Github Action code locally

agbell:
This PR wraps the github action build steps in an Earthfile.  This allows for the running of the build steps locally. 

## Running the CI steps locally
If you install earthly `brew install earthly` you can now run the current github action build locally using:
```
> earth -P +all
...
=========================== SUCCESS ===========================
```

You can run individual sections like this:
```
 # all unit tests on all specified elixir and otp versions
> earth +all-test
```
```
# all integration tests on all specified elixir and otp versions
> earth -P all-integration-test 
```
```
# npm test
> earth +npm # npm test
```

You can also run the tests on specific version like this:
```
> earth --build-arg ELIXIR={{someversion}}  --build-arg OTP=${{someversion}} +test
```
The same works for +integration-test and +npm

## Caveats 
The build process is equivalent to the previous version to the best of my abilities.  You can see some runs here:
https://github.com/earthly/phoenix/actions/runs/370395818

NB: I did find the +test is sometimes flaky, but I think this is inherent to the tests and not the Earthfile wrapping.  My knowledge of elixir wasn't strong enough to track that down.

## Performance:
This wrapping seems to add about 30 seconds to the build time. My thinking is that is probably worth it, because you can now run these steps locally as often as you'd like, without need to push your branch. Locally it can be iterated on much faster.  On a dev machine, due to caching between steps, the process can be significantly faster, after the first run.

---
cc @josevalim ( I heard via @vladaionescu that you wanted a way to reproduce build failures more easily.  I'd be happy to walk-through how it works, if that is helpful. It is very possible this could be further optimized with more knowledge of elixir and mix. )


josevalim:
Thanks @agbell, this was super fast and is very welcome! We will discuss between us and follow up.

josevalim:
@aaronrenner I would like to hear your thoughts on this, especially as the integration tests you have been working on would be the biggest beneficial of running this locally. I assume with this we no longer need the docker-compose.yml.

Earthfile
```diff
@@ -0,0 +1,64 @@
...
+    COPY --dir assets config installer lib integration_test priv test ./
+    WORKDIR /src/installer
+    RUN mix deps.get
+
+    WORKDIR /src/integration_test 
+    RUN mix deps.get
+    WITH DOCKER
+        RUN docker run --name postgres -e POSTGRES_PASSWORD=postgres -p 5432:5432 -d postgres && \
+            docker run --name mysql -e MYSQL_ALLOW_EMPTY_PASSWORD=yes -p 3306:3306 -d mysql && \
+            docker run --name mssql -e ACCEPT_EULA=Y -e SA_PASSWORD=some!Password -p 1433:1433 -d mcr.microsoft.com/mssql/server:2019-latest && \
```
josevalim, line 33: Could we run the docker compose file instead? Would that be preferred?

Earthfile
```diff
@@ -0,0 +1,64 @@
...
+    COPY --dir assets config installer lib integration_test priv test ./
+    RUN mix test
+
+all-integration-test:
+    BUILD --build-arg ELIXIR=1.11.1 --build-arg OTP=21.3.8.18 +integration-test
+    BUILD --build-arg ELIXIR=1.11.1 --build-arg OTP=23.1.1 +integration-test
+
+integration-test:
+    FROM +setup
+    COPY --dir assets config installer lib integration_test priv test ./
```
josevalim, line 24: Why do we need `./` at the end? I assume COPY only copy the files in a directory so that copies the files at root?

Earthfile
agbell: Yep, a docker compose would work here. It would look like this:
```
WITH DOCKER --compose docker-compose.yml
       RUN  mix test --include database
    END
```
I can make that change if that is your preferred way to structure things.

Earthfile
agbell: The last one is the destination, so this is copying all those folders into the working directory. I wasn't actually sure if they were all needed to run the integration test. Are there any that are unnecessary?

mcrumm:
This is awesome, thanks @agbell!

I was able to get a successful run locally on Ubuntu with the following:

```sh
earth --allow-privileged +all
```

I did notice however, after a failed run, there were some cascading failures on subsequent runs, generally around duplicate container names:

```sh
+integration-test | docker: Error response from daemon: Conflict. The container name "/postgres" is already in use by container "948ca7897809a1eb348793eaac734d94aa60351f7bbbfcec1011e0b6327e8690". You have to remove (or rename) that container to be able to reuse that name.
```

Running `earth prune` between test runs appears to resolve the issue.

agbell:
Thanks @mcrumm Yes, I forgot to mention the -P or --allow-privileged is needed for starting docker from earthly.

> I did notice however, after a failed run, there were some cascading failures on subsequent runs, generally around duplicate container names:
That shouldn't happen.  Let me see if I can address that.

agbell:
> I did notice however, after a failed run, there were some cascading failures on subsequent runs, generally around duplicate container names

I moved to the existing docker-compose which, I think, should eliminate this problem. Let me know if it happens again @mcrumm .

mcrumm:
Looks good now @agbell, thanks! I was able to get a successful run after a failed run without a prune in between. :)

Gazler:
This is my first time encountering the earth tool. As someone who has spent multiple hours trying to track down weird edge cases on CI, I think this looks awesome. Thanks for opening the PR @agbell 

I did notice that locally for me, the test runner output was kind of all over the place:

earth-run

Is there something that can be done to clean up this output?

agbell:
> Is there something that can be done to clean up this output?

Totally, we have a fix coming that will improve it drastically.  

https://github.com/earthly/earthly/pull/514

I was released today, but not in brew yet.  So give it a couple days and then 
```
brew upgrade earthly
```
and it should be better, though still not perfect.

Earthfile
agbell: I moved it to docker compose

aaronrenner:
@agbell Thanks for posting PR, it looks really interesting! I see a couple places where the Earthfile can be adjusted (like we don't need to install the installer deps in the integration_tests), but I'm excited to play around with it locally so I can make better suggestions. 

Out of curiosity, how long does the first run usually take? I am running on an older linux laptop, but it's been running for about half an hour so far.

aaronrenner:
@agbell After initially running `sudo earth -P +all`, now when I try to run the integration tests by themselves I get a message saying the layer has already been cached.

```shell-session
$ sudo earth -P --build-arg ELIXIR=1.11.1 --build-arg OTP=21.3.8.18 +integration-test
...
   +integration-test | *cached* --> WITH DOCKER RUN mix test --include database
=========================== SUCCESS ===========================
```
How would I make the integration tests execute here?

agbell:
Hi @aaronrenner 
> Out of curiosity, how long does the first run usually take? I am running on an older linux laptop, but it's been running for about half an hour so far.

That seems on the long side, especially on linux where there is not much overhead to docker, I'd love to know where it slows down.  Ideally, It shouldn't be too much slower than doing the equivalent steps outside of earthly.

> How would I make the integration tests execute here?

You can add the --no-cache flag to turn off all caching, or alternatively touch some file in the integration test area.  With caching on earthly assumes that if not files have changed that it is using for a step, that the results can be cached.  It is assuming tests are deterministic basically.   

Are you thinking that integration tests should never cache, I think that should be doable as well. 

I work on Earthly, so this a fun test case for me :) Are you on the elixir slack channel? maybe we can talk it out.

aaronrenner:
Thanks for the suggestion, @agbell. It seems like we may need to adjust the layers in the earth file because if I run
```sh-session
$ touch integration_test/test/code_generation/app_with_defaults_test.exs
$ time sudo earth -P --build-arg ELIXIR=1.11.1 --build-arg OTP=21.3.8.18 +integration-test
```
it takes about 10 mins because it has to re-setup docker in docker and the containers to be used by the test. The full output from my machine is available here: https://gist.github.com/aaronrenner/641ae0c68008b1d1d53930aeed76bddb.

I am on Elixir Slack (aaronrenner) if you want to reach out, but I probably won't be able to get to spend time on this until later in the evenings.

agbell:
@aaronjensen  and I chatted on the weekend.  The speed is looking better now, but Aaron had some further suggestions.  I will attempt to incorporate those and then resubmit.

Earthfile
```diff
@@ -0,0 +1,80 @@
...
+    FROM +test-setup
+    COPY assets assets
+    RUN mix deps.get
+    SAVE ARTIFACT assets
+
+setup-base:
+   ARG ELIXIR=1.11.2
+   ARG OTP=23.1.1
+   FROM hexpm/elixir:$ELIXIR-erlang-$OTP-alpine-3.12.0
+   RUN apk add --no-progress --update git docker docker-compose
```
aaronrenner, line 68: Out of curiosity, why are docker and docker-compose being installed in setup-base? I think they are only needed by the integration tests.

agbell: Just for simplicity's sake I was trying to put all the dependency installing in setup step, but they could be put in top of integration-test instead.

Earthfile
```diff
@@ -0,0 +1,83 @@
+
```
aaronrenner, line 1: I found out this blank line is important when my editor automatically stripped it and I got the following error:

```
% sudo earth -P --build-arg ELIXIR=1.11.1 --build-arg OTP=21.3.8.18 +integration-test

           buildkitd | Found buildkit daemon as docker container (earthly-buildkitd)
line 1:0 no viable alternative at input 'all:'
Error: build target: build main: bkClient.Build: failed to solve: rpc error: code = Unknown desc = syntax error: line 1:0 no viable alternative at input 'all:'
```

It looks like you've already opened an issue on it https://github.com/earthly/earthly/issues/538, but I wanted to call that out in case anyone else is trying this out.

.github/workflows/ci.yml
```diff
@@ -3,60 +3,31 @@ on: [push, pull_request]
...
+    env:
+      GIT_URL_INSTEAD_OF: "https://github.com/=git@github.com:"
+      FORCE_COLOR: 1
+    
     strategy:
       fail-fast: false
       max-parallel: 1
       matrix:
         include:
-          - elixir: 1.9.4
```
agbell, line 16: It seems that there are a couple of tests that are flakey on these versions, when run in earthly, at least in my testing.  Without them, the tests pass consistently.

agbell:
@aaronrenner I have incorporated your suggested changes and also tracked down a couple of issues unrelated issues.  Could you take another look?

aaronrenner (commented):
@agbell Thanks for all of your work on this. It looks like it took some work to ensure the DBs were up before running the tests, but it appears to be worth it because I'm not having issues with the SQL server tests any more!

After playing around with this a bit, here's a list of my thoughts and questions.

1. It's really great to be able to run CI locally and troubleshoot errors instead of having to push it up to Github. I also tried using it during development, but since any code change causes the layer containing the `_build` folder to be invalidated, as well as all of the docker containers needing to be started again, I'm finding it's still faster for me to do my TDD cycles on my local machine instead of with earthy. I personally see running earthy as a final step before I push a PR up to github. When documenting how to use earthly, we should indicate when is the appropriate time to use it vs running tests locally.
2. I think the `--interactive` option is a killer feature (it opens up an interactive console when a command returns a non-zero exit status) and will be really helpful when debugging CI issues. Earthly also has some helpful documentation on [debugging techniques](https://docs.earthly.dev/guides/debugging). Unfortunately, it appears `--interactive` doesn't work yet if there's a failure inside of the `WITH DOCKER` statement. (Not a deal breaker at all, but something worth opening an issue on if it doesn't already exist).
3. My machine is kinda slow so if I run the `+all` target, I getting ex_unit timeouts. I can solve this by running the targets individually. Is there any way to control the max concurrency in earthly?
4. It's nice how easy it is to test a new elixir version. All that needs to be done is to update the build arg: 

    `sudo earth -P --build-arg ELIXIR=1.11.2 --build-arg OTP=21.3.8.18 +integration-test`
5. We don't currently use caching, but if we started doing something like running dialyzer in an integration test, we'd need to figure out how to integrate with github actions. It looks like there are a couple of issues being worked on, but it's not supported yet: https://github.com/earthly/earthly/issues/375#issuecomment-739064483

Earthly is really cool and I appreciate being able to run CI locally. This list is only to summarize my experience to the rest of the phoenix team and maybe it will be helpful to the earthly team too. I'll defer to the rest of the phoenix team on what approach they'd like to take with our CI strategy.

aaronrenner:
One other thing I noticed is this branch's integration tests appear to be 2 minutes slower than those on master. What do you think is contributing to that slowdown?

agbell:
Great feedback @[aaronrenner](https://github.com/aaronrenner)!  This feedback is very helpful for us.  I agree that being able to run the tests on a specific version of Elixir from the command line can be pretty helpful when you need it.

**Speed:**
Regarding integration tests taking a bit longer, that one is super interesting. There is some overhead caused by earthly, but that is not the issue in this case.  The issue is the order of operations I believe.  

In the Earthfile we compile the integration tests with `mix deps.compile`, then start the dockerfiles and wait for them to be up before starting the tests.  This was your suggested workflow and I think it's the correct way to do things.  In master, the compilation happens as part of `mix test --include database` which happens after the docker images are started.  This means if the compilation is faster than the DB startup then the tests will fail (which was an issue I got stuck on for a while )but since the compilation takes a while, this seems to almost never happen.   If compiles got faster it could start breaking though.  

My feeling is the way the earthfile is doing it is 'correct' though a bit slower.  I think that if we moved `mix deps.compile` into the `WITH DOCKER` we could get the best of both worlds, with a slightly more confusing build script.  
You can see in the original earthfile the build was faster, when it mirrored the github actions process:  [Here](https://github.com/phoenixframework/phoenix/runs/1418811508 
)
**Other feedback**
Regarding # 2: The interactive mode, I agree that the interactive feature is super helpful for debugging.  We have just addressed `WITH DOCKER` [issue](https://github.com/earthly/earthly/pull/592) so this can be used in the integration tests as well.

Regarding #5 Caching between runs in Github actions is high on our priority list.  You can see it at the top of our 'under consideration' on the [roadmap](https://github.com/earthly/earthly/projects/1). So its possible we will have in place by the time you need it

Regarding concurrency, there is not a way to manually control that at present.  We may need to just to widen ELIXIR_ASSERT_TIMEOUT to support slower runs.  

If this seems like a valuable PR, or is close to one, awesome!  If it's not the direction phoenix wants to go, then thanks for the feedback.

josevalim:
> In this branch we compile the integration tests with mix deps.compile, then start the dockerfiles and wait for them to be up before starting the tests.

Could it be caused because we are compiling the deps twice? One in dev (the default) and another in test? Or do we run everything with MIX_ENV=test?

agbell:
> Could it be caused because we are compiling the deps twice? One in dev (the default) and another in test? Or do we run everything with MIX_ENV=test?

Oh, do you mean [this](https://github.com/phoenixframework/phoenix/pull/4072/files#diff-2705f6016a64189abd14041e2f3e8dbb829553295d9088e4eb42eccb5e61bf47R52) and [this](https://github.com/phoenixframework/phoenix/pull/4072/files#diff-2705f6016a64189abd14041e2f3e8dbb829553295d9088e4eb42eccb5e61bf47R62) are duplicating some effort?  Or something else? 

I can change it back to copying all files in, starting docker-compose and then running the tests.  That is faster in github actions but gets less help from the cache when running incrementally locally, as the cache is invalidated on any change on what is copied.  If faster GHA time is the priority though, I could see that making sense.

josevalim:
Those lines look good, so please ignore me. :)

aaronrenner:
> This means if the compilation is faster than the DB startup then the tests will fail (which was an issue I got stuck on for a while ) but since the compilation takes a while, this seems to almost never happen.

@agbell That's really interesting. I ran a couple builds with timestamps and it seems like database startup is taking 71 seconds

* [Build that checks for databases to be up](https://github.com/aaronrenner/phoenix/runs/1521708668)
* [Build that skips database checks](https://github.com/aaronrenner/phoenix/runs/1521919365?check_suite_focus=true#step:5:53)

It looks like `mix deps.compile` takes 44 seconds so there's a chance we could still run into issues even if we ran it inside of `WITH DOCKER`. We just have to fill up time before SQL server starts up. It would be awesome if earthly could start up these external services out of band, but I get it isn't designed that way. I think it's probably better to optimize this script for running in CI instead of in local development, since we'll probably be only running this locally when we're running into CI issues. WDYT?

It's also really nice to see the speed is comparable for all of the other operations.

aaronrenner:
@agbell I just wanted to let you know we're going to go ahead with earthly for CI. 🎉 

I think there's a couple things left to do on this PR:

- [ ] Optimize the integration tests to maintain speed in CI rather than speed when iterating in local development.
- [ ] Add documentation about running tests with earthly (I'm still trying to decide where this should go though). We should say this is more for being able to execute and troubleshoot CI locally and actual development workflows will most likely be faster running on the local machine.

Thanks so much for all of the work you've done on this PR and on earthly itself! I'll be happy to continue to help as we see this PR through to completion.

agbell:
Awesome @aaronrenner! I'll work on the integration speed this week a bit.

agbell:
@aaronrenner  I'm made some changes that make the GHA run faster and added some documentation. I added some information to the integration readme.  Should I also add info to the CONTRIBUTING.md?  

There is a lot of variation in the github action run time, so I can quite pin down the difference.  However, the unit tests now are within some amount of seconds from the time in master.  The integration tests are about a minute slower.   That seems to be about the best that I can do.  

 Let me know what you think.

.github/workflows/ci.yml
```diff
@@ -3,60 +3,31 @@ on: [push, pull_request]
 jobs:
   test-elixir:
     runs-on: ubuntu-latest
-    container:
-      image: hexpm/elixir:${{ matrix.elixir }}-erlang-${{ matrix.otp }}-alpine-3.12.0
-      env:
-        ELIXIR_ASSERT_TIMEOUT: 2000
-
+    env:
+      GIT_URL_INSTEAD_OF: "https://github.com/=git@github.com:"
```
aaronrenner, line 7: I was reading the docs and it looks like this option might be obsolete now:

https://github.com/earthly/earthly/blob/main/docs/earthly-command/earthly-command.md#--git-url-instead-of-git-instead-of-obsolete

integration_test/README.md
```diff
@@ -26,6 +26,25 @@ This allows all tests to be run with the following command
...
+
+## Alternative ways to run the integration tests
+
+Phoenix uses Earthly as part of the CI process. It is possible to use earthly to test changes to integration tests locally.  It is also possible to run the CI process locally, which is helpful when debugging CI failures.  
+
+[Installation Instructions](https://docs.earthly.dev/installation)
+
+To run integration tests against all supported Elixir and OTP versions:
+
+    $ earth -P +all-integration-test
```
aaronrenner, line 38: I think the command name has to change once this is upgraded to earthly 0.4.0.

Earthfile
```diff
@@ -0,0 +1,103 @@
+# Earthfile
+# 
+# The following blank line is required until https://github.com/earthly/earthly/issues/538 is resolved
```
aaronrenner, line 3: I think this can be removed now that 0.4.0 is out

aaronrenner (commented):
@agbell Thanks so much for your changes! The docs look great! I don't think we need to worry about adding anything to CONTRIBUTING.md at the moment. I also appreciate your work on the performance improvements and think it's in a good condition to move forward.

Sorry it's taken me a while to get time to look at this, but in the meantime it looks like earthly 0.4.0 has been released. Before we merge, would you mind upgrading this PR to earthly 0.4.0 and rebasing these changes on master? It looks like the command name changed from `earth` to `earthly` and it fixed a few of the issues we've made notes for.

Thanks again for all of your work on this! I'm looking forward to getting this into master.

agbell:
@aaronrenner done! I rebased it over master and force pushed it. Thanks for helping me with this process!

We are now using 0.4.1 which has a fix for interactive mode working inside `WITH DOCKER` blocks and as you've noticed, the command is now `earthly`.  

Also, it seems I was counting wrong before, on the performance numbers.  I wasn't counting the time for github actions to pull the docker images.  For the total build time, it looks like we are either the same or slightly faster than master if you look at the total time! 6:30 for this and close to 7:00 for master, though I think that is within a margin of error.

Earthfile
```diff
@@ -0,0 +1,99 @@
...
+
+test-setup:
+   FROM +setup-base
+   COPY mix.exs .
+   COPY mix.lock .
+   COPY .formatter.exs .
+   COPY package.json .
+   RUN mix local.rebar --force
+   RUN mix local.hex --force
+   RUN mix deps.get
```
aaronrenner, line 99: Trailing newline? 😁

agbell: added

aaronrenner (approved):

josevalim (approved):

PR #4073, opened 2020-11-20
Log the transport URL, not the endpoint URL

garthk:
We were confused by the following logger output after setting our `transport` to `LongPoll`:

```plain
socket: transport: connected to wss://example.com/live/websocket`
```

This PR adds a WebSocket-API-compatible [url] property to `LongPoll`, after which `Socket.onConnOpen` can report `this.conn.url` reflecting the URL over which the connection was made. In case of a custom transport without such a property, it falls back to `Socket.endpointURL`.

[url]: https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/url

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -1024,7 +1024,7 @@ export class Socket {
    * @private
    */
   onConnOpen(){
-    if(this.hasLogger()) this.log("transport", `connected to ${this.endPointURL()}`)
+    if(this.hasLogger()) this.log("transport", `connected to ${this.conn.url || this.endPointURL()}`)
     this.unloaded = false
     this.closeWasClean = false
     this.flushSendBuffer()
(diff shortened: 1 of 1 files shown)


garthk:
Workaround without this PR goes something like this:

```javascript
class MyLongPoll extends LongPoll {
  get url() {
    return this.endpointURL();
  }
};

let liveSocket = new LiveSocket('/live', Socket, {
  params: {
    _csrf_token: getMetaContent('csrf-token'),
    _traceparent: getMetaContent('traceparent'),
  },
  hooks: Hooks,
});

if (flags.has("longpoll_default")) {
  liveSocket.socket.transport = MyLongPoll;
}

liveSocket.socket.logger = (kind, msg, data) => {
  if (kind === 'transport' && typeof msg === 'string' && msg.slice(0, 13) === 'connected to ') {
    msg = `connected to ${liveSocket.socket.conn.url}`;
  }
  debug(`socket: ${kind}: ${sanitise(msg)}`, data);
};
```
… hence my interest in fixing it upstream.

garthk:
Is there a way to re-run the test other than to push again? It failed in `actions/setup-elixir`.

chrismccord:
Sorry this has taken a bit of time. Master js build has deviated significantly from this PR. Please resubmit based off master if you have a chance. Thanks!

PR #4075, opened 2020-11-24
Fix typo in contexts guide

yellowmint:
Guide in router mounts SessionController under `/sessions` path, not `/session`. Article later on also refers to `/sessions`.

> Now if you visit the sign-in page at http://localhost:4000/sessions/new and enter a bad email address, you should be greeted with your flash message.

> Let's sign in at http://localhost:4000/sessions/new


mcrumm:
Thanks!

PR #4077, opened 2020-11-25
Add phx.gen.auth generator

aaronrenner:
This adds the `mix phx.gen.auth` task from phx_gen_auth v0.6.0 into phoenix master. The only additional change is to address a test failure caused by some changes to `current_path/1` on phoenix master (see https://github.com/aaronrenner/phx_gen_auth/issues/96).

This PR is missing the phx.gen.auth overview doc (https://github.com/aaronrenner/phx_gen_auth/blob/master/guides/overview.md). Is that something you'd like as part of this PR and if so, where would be a good place to work this in? We'll also probably want to add this to phoenix's mix tasks guide in a future PR.

Let me if you see anything stylistically that needs to be updated to match the rest of the phoenix codebase or anything else that would be good to change. Thanks!

/cc @josevalim


integration_test/test/code_generation/app_with_mssql_adapter_test.exs
```diff
@@ -83,4 +83,29 @@ defmodule Phoenix.Integration.CodeGeneration.AppWithMSSQLAdapterTest do
...
+        mix_run!(~w(phx.gen.auth Accounts User users --hashing-lib pbkdf2), app_root_path)
+
+        assert_no_compilation_warnings(app_root_path)
+        assert_passes_formatter_check(app_root_path)
+      end)
+    end
+  end
+
+  @tag database: :mssql
+  test "has a passing test suite" do
```
josevalim, line 101: Should this be moved inside the describe block above?

integration_test/test/code_generation/app_with_defaults_test.exs
```diff
@@ -161,4 +161,29 @@ defmodule Phoenix.Integration.CodeGeneration.AppWithDefaultsTest do
...
+        mix_run!(~w(phx.gen.auth Accounts User users), app_root_path)
+
+        assert_no_compilation_warnings(app_root_path)
+        assert_passes_formatter_check(app_root_path)
+      end)
+    end
+  end
+
+  @tag database: :postgresql
+  test "has a passing test suite" do
```
josevalim, line 179: Should this be moved inside the describe block above?

integration_test/test/code_generation/app_with_defaults_test.exs
```diff
@@ -161,4 +161,29 @@ defmodule Phoenix.Integration.CodeGeneration.AppWithDefaultsTest do
       end)
     end
   end
+
+  describe "phx.gen.auth" do
```
josevalim, line 165: What if we also add the name of the hashing lib in the describe? E.g.: phx.gen.auth+bcrypt

lib/mix/tasks/phx.gen.auth.ex
```diff
@@ -0,0 +1,680 @@
...
+  ## Custom table names
+
+  By default, the table name for the migration and schema will be
+  the plural name provided for the resource. To customize this value,
+  a `--table` option may be provided. For example:
+
+      mix phx.gen.auth Accounts User users --table accounts_users
+
+  This will cause the generated tables to be named `"accounts_users"` and `"accounts_users_tokens"`.
+
```
josevalim, line 73:
```suggestion
```

josevalim:
@aaronrenner just some minor comments but otherwise it looks great to me!

Regarding the guides, we need to do three changes which can be done separately:

1. Bring in the guide under a new subsection called "Authentication", that comes right under subsection "Guides". The current guide can now be called "mix phx.gen.auth" and I believe it should be a mixture of the current README and the guide. I.e. it should explain how to use the generator and also describe the security considerations.

2. Update the [mix task guide](https://hexdocs.pm/phoenix/mix_tasks.html) to mention `mix phx.gen.auth` and simply link to the guide above

3. We should probably update the context guides to use another example beyond authentication /cc @chrismccord

integration_test/test/code_generation/app_with_defaults_test.exs
josevalim: The same applies to the test files, btw. :)

aaronrenner: Thanks for catching that. I fixed them all in 02e8988ebff9e022cd326d64c98be695b7ba73bb.

lib/mix/tasks/phx.gen.auth/hasing_library.ex
```diff
@@ -0,0 +1,47 @@
+defmodule Mix.Tasks.Phx.Gen.Auth.HashingLibrary do
```
feliperenan, line 1: I think there is type on the file name 😅 

`%s/hasing/hashing`

aaronrenner: Thanks for catching that @feliperenan! Fixed in https://github.com/phoenixframework/phoenix/pull/4077/commits/ff10072354005220223402eef5b07e0f45704e72.

josevalim (commented):
LGTM! Feel free to merge it at your discretion, if you would rather add the guides now or later, it is your call. :heart:

PR #4081, opened 2020-11-26
Bump ecto requirement to ~> 3.5.0

aaronrenner:
I was trying to add `redact: true` to `phx.gen.auth` and realized we haven't increased our `ecto_sql` dependency to 3.5, yet. This bumps the required ecto_sql version for both the generated projects and the integration tests.

Please let me know if we need to include a higher patch version.


aaronrenner:
Closing in favor of #4055

PR #4087, opened 2020-12-03
Private asset package by default

chulkilee:
I found `assets/package.json` from `phx.new` uses MIT license.

I believe it should be private and unlicensed by default, for following reasons:

- it is unlikely to publish the package to repository, since it's for final assets not for a library
- Elixir part of generated code does not have any license information

From [package.json](https://docs.npmjs.com/cli/v6/configuring-npm/package-json)

> Finally, if you do not wish to grant others the right to use a private or unpublished package under any terms:
>
> { "license": "UNLICENSED" }
>
> Consider also setting "private": true to prevent accidental publication.

---

Update: since `license` doesn't seem to be required, and npm does not emit error when `private` is true and `license` is omitted, updated this PR to omit license instead of using "UNLICENSED".

### DIFF
--- a/installer/templates/phx_assets/package.json
+++ b/installer/templates/phx_assets/package.json
@@ -1,6 +1,4 @@
 {
-  "repository": {},
-  "description": " ",
   "private": true,
   "scripts": {
     "deploy": "webpack --mode production",
--- a/installer/templates/phx_assets/package.json
+++ b/installer/templates/phx_assets/package.json
@@ -1,7 +1,7 @@
 {
   "repository": {},
   "description": " ",
-  "license": "MIT",
+  "private": true,
   "scripts": {
     "deploy": "webpack --mode production",
     "watch": "webpack --mode development --watch"


cw789:
I think it should be `private: true` only.
License is something unique per projects basis.

I've already addressed `private: true` in the following PR: https://github.com/phoenixframework/phoenix/pull/4054

Totally agree with your mentioned reasons.

chulkilee:
The problem is, it is adding `MIT` license to the npm package, while elixir package (mix or LICENSE) does not have any license info. This is inconsistent.

Like `mix new` does not add any license info.. the generator should avoid adding license info, as this is *unique per projects basis* :)

cw789:
Yeah, definitely we should remove the license.
But that doesn't mean it's necessary to add "unlicence" explicitly.

chulkilee:
Oh that's good point.. but I believe the lack of license seems "unlicensed"

https://choosealicense.com/no-permission/

>  Unless you include a license that specifies otherwise, nobody else can copy, distribute, or modify your work without being at risk of take-downs, shake-downs, or litigation.

... or we may put empty string or omit `license` attribute at all. Not sure that's recommended by npm toolings. I saw some command shows warning if license is missing a while ago.

lukaszsamson:
Empty `"license"` in `package.json` results in a warning during `npm i`.

cw789:
> Empty `"license"` in `package.json` results in a warning during `npm i`.

Not when `private: true` is set.

chulkilee:
Updated PR to just drop `license` and add `private`. If #4054 will be merged soon, then let's use that. Otherwise (e.g. more changes needed there) - then probably better to merge this PR first to make one change at a time..

cw789:
What I meant I've already integrated this in PR https://github.com/phoenixframework/phoenix/pull/4054.
With `private: true` also `description` & `repository` won't be necessary anymore (no warning).

josevalim:
I will go ahead and merge this since the convo on #4054 is going (although ideally we will merge both for v1.6).

josevalim:

PR #4088, opened 2020-12-03
Use 127.0.0.1 in dev and test

chulkilee:
Reviving #3991 as the old PR haven't got any attention probably it's always closed.. 😢 

- It does not overlap with other PRs
- I confirmed the change works in Microsoft Windows as well


josevalim:

PR #4089, opened 2020-12-05
Add mix phx.gen.auth to guides

aaronrenner:
This adds a new `mix phx.gen.auth` guide under the Authentication guide sub section.

Screen Shot 2020-12-05 at 12 28 51 PM

This also adds `mix phx.gen.auth` to the Mix Tasks guide.

### DIFF
--- a/guides/authentication/mix_phx_gen_auth.md
+++ b/guides/authentication/mix_phx_gen_auth.md
@@ -28,7 +28,7 @@ And finally, let's start our phoenix server and try it out.
 
 ## Developer Responsibilities
 
-Since phoenix generates this code into your application instead of building these modules into phoenix itself, you now have complete freedom to modify the authentication system so it works best with your use case. The one caveat with using a generated authentication system is it will not be updated after it's been generated. Therefore as improvements are made to the output of `mix phx.gen.auth`, it becomes your responsibility to determine if these changes need to be ported into your application.
+Since phoenix generates this code into your application instead of building these modules into Phoenix itself, you now have complete freedom to modify the authentication system so it works best with your use case. The one caveat with using a generated authentication system is it will not be updated after it's been generated. Therefore as improvements are made to the output of `mix phx.gen.auth`, it becomes your responsibility to determine if these changes need to be ported into your application. Security-related and other important improvements will be explicitly and clearly marked in CHANGELOG and upgrade notes.
 
 ## Generated Code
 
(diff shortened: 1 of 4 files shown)


guides/authentication/mix_phx_gen_auth.md
```diff
@@ -0,0 +1,87 @@
...
+
+    $ mix test
+
+And finally, let's start our phoenix server and try it out.
+
+    $ mix phx.server
+
+## Developer Responsibilities
+
+Since phoenix generates this code into your application instead of building these modules into phoenix itself, you now have complete freedom to modify the authentication system so it works best with your use case. The one caveat with using a generated authentication system is it will not be updated after it's been generated. Therefore as improvements are made to the output of `mix phx.gen.auth`, it becomes your responsibility to determine if these changes need to be ported into your application.
```
josevalim, line 31:
```suggestion
Since phoenix generates this code into your application instead of building these modules into Phoenix itself, you now have complete freedom to modify the authentication system so it works best with your use case. The one caveat with using a generated authentication system is it will not be updated after it's been generated. Therefore as improvements are made to the output of `mix phx.gen.auth`, it becomes your responsibility to determine if these changes need to be ported into your application. Security-related and other important improvements will be explicitly and clearly marked in CHANGELOG and upgrade notes.
```

guides/authentication/mix_phx_gen_auth.md
```diff
@@ -0,0 +1,87 @@
...
+
+The generated code ships with an auth module with a handful of plugs that fetch the current user, require authentication and so on. For instance, in an app named Demo which had `mix phx.gen.auth Accounts User users` run on it, you will find a module named `DemoWeb.UserAuth` with plugs such as:
+
+  * `fetch_current_user` - fetches the current user information if available
+  * `require_authenticated_user` - must be invoked after `fetch_current_user` and requires that a current user exists and is authenticated
+  * `redirect_if_user_is_authenticated` - used for the few pages that must not be available to authenticated users
+
+### Confirmation
+
+The generated functionality ships with an account confirmation mechanism, where users have to confirm their account, typically by email. However, the generated code does not forbid users from using the application if their accounts have not yet been confirmed. You can trivially add this functionality by customizing the plugs generated in the Auth module.
```
josevalim, line 51: Avoid works like trivially because, whenever a user cannot pull it off, they will feel frustrated.

```suggestion
The generated functionality ships with an account confirmation mechanism, where users have to confirm their account, typically by email. However, the generated code does not forbid users from using the application if their accounts have not yet been confirmed. You can add this functionality by customizing the `require_authenticated_user` in the `Auth` module to check for the `confirmed_at` field (and any other property you desire).
```

josevalim (approved):
Looks great, just some very minor suggestions!

PR #4091, opened 2020-12-05
Add default to first param of generated Schema.changeset function

TheFirstAvenger:
This is a very minor tweak that I have made to every schema I have ever generated with mix `mix phx.gen`.
```elixir
  def changeset(user, attrs) do
    user
    |> cast(attrs, [:name])
    |> validate_required([:name])
  end
```
becomes:
```elixir
  def changeset(user \\ %User{}, attrs) do
    user
    |> cast(attrs, [:name])
    |> validate_required([:name])
  end
```
and
```elixir
  def create_user(attrs \\ %{}) do
    %User{}
    |> User.changeset(attrs)
    |> Repo.insert()
  end
```
becomes
```elixir
  def create_user(attrs \\ %{}) do
    attrs
    |> User.changeset()
    |> Repo.insert()
  end
```

### DIFF
--- a/priv/templates/phx.gen.context/schema_access.ex
+++ b/priv/templates/phx.gen.context/schema_access.ex
@@ -43,8 +43,8 @@
 
   """
   def create_<%= schema.singular %>(attrs \\ %{}) do
-    %<%= inspect schema.alias %>{}
-    |> <%= inspect schema.alias %>.changeset(attrs)
+    attrs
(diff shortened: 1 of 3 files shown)


chrismccord:
I believe @josevalim and I went back and forth on this exact change when working on the context generators way back, and we settled on context being the one responsible for the struct and/or default args. Absolutely feel free to make such changing to your apps, but I thing we're set on current generators. Thanks!

PR #4092, opened 2020-12-06
Use redact: true for password & hash in mix phx.gen.auth

TheFirstAvenger:
Use redact: true for password & hash in `mix phx.gen.auth`


aaronrenner (approved):
Just tested this out and and confirmed the password does not show up when inspecting the schema or the changeset. Thanks @TheFirstAvenger!

josevalim:

Issue #4079, opened 2020-11-25
Phoenix.js push with an undefined payload


sb8244:
### Environment

* Elixir version (elixir -v): 1.9.1
* Phoenix version (mix deps): shipped version with Phoenix 1.5.7
* NodeJS version (node -v): 12.16.2
* NPM version (npm -v): 6.14.4
* Operating system: Mac OS

### Expected behavior

`channel.push("example")` sends a push to the server with an empty payload

### Actual behavior

Previously, code like `channel.push("example")` would work correctly and make it up to the server without issue. After the change to add [ArrayBuffer encoder](https://github.com/phoenixframework/phoenix/blame/master/assets/js/phoenix.js#L677), the payload is assumed to be a defined object that responds to `constructor`. So an error like this appears:

![image](https://user-images.githubusercontent.com/1231659/100291189-16698980-2f4b-11eb-9835-1ac32ba2d983.png)

I can see the argument that the payload should have always been an object, so I would understand closing this. However, I wanted to bring it up just in case it matters.

PR #4093, opened 2020-12-06
Default pushed payload to an empty object if not provided

sb8244:
The documentation provides an example like `channel.push("event")`, but that no longer works with an undefined payload object. An empty object should be the equivalent functionality as the previous undefined case.

#4079 

-----

A second option is to update this code: 

```
  encode(msg, callback){
    if(msg.payload.constructor === ArrayBuffer){
```

to 

```
  encode(msg, callback){
    if(msg.payload && msg.payload.constructor === ArrayBuffer){
```

However, I like the idea of not worrying about an undefined msg.payload in the code. But this approach should be completely side effect free, since it would be equivalent to the code that used to work.

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -525,7 +525,7 @@ export class Channel {
    * @param {number} [timeout]
    * @returns {Push}
    */
-  push(event, payload, timeout = this.timeout){
+  push(event, payload = {}, timeout = this.timeout){
     if(!this.joinedOnce){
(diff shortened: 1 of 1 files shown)


chrismccord:
❤️❤️❤️🐥🔥

juampi:
Stumbled upon this issue as well, glad to see it's already fixed on `master` ❤️

Just wondering, when is the next release planned for?

PR #4095, opened 2020-12-07
Docs: MyApp.Endpoint -> MyAppWeb.Endpoint 

### LINKED ISSUE #4094: Doc typos: Broken Endpoint code snippets
### Environment

* Operating system: macOS 10.14.6

### Expected behavior
All [Endpoint examples](https://hexdocs.pm/phoenix/Phoenix.Endpoint.html#module-endpoints) from the Phoenix docs should run when `YourApp` or `MyApp` is replaced with your app name

### Actual behavior
Several Endpoint examples from Phoenix docs do not work as documented, as Endpoint is declared/referenced as `YourApp.Endpoint` in the docs but Phoenix generates endpoints under Web: `YourAppWeb.Endpoint`

njwest:
This fixes #4094 

Ran a find-and-replace for `MyApp.Endpoint` -> `MyAppWeb.Endpoint` and `YourApp.Endpoint` -> `YourAppWeb.Endpoint` in the docs

Phoenix generates the endpoint under AppWeb, not under App; I think the docs should reflect this, as I discovered this issue when attempting to use `broadcast!` functions taken from the docs

### DIFF
--- a/guides/deployment/deployment.md
+++ b/guides/deployment/deployment.md
@@ -60,7 +60,7 @@ Keep in mind that, if you by any chance forget to run the steps above, Phoenix w
 
 ```console
 $ PORT=4001 MIX_ENV=prod mix phx.server
-10:50:18.732 [info] Running MyApp.Endpoint with Cowboy on http://example.com
+10:50:18.732 [info] Running MyAppWeb.Endpoint with Cowboy on http://example.com
 10:50:18.735 [error] Could not find static manifest at "my_app/_build/prod/lib/foo/priv/static/cache_manifest.json". Run "mix phx.digest" after building your static files or remove the configuration from "config/prod.exs".
 ```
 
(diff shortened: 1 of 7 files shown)


aaronrenner:
I'm not sure if it was an oversight or intentional to have this be `MyApp` instead of `MyAppWeb`, but if it was an oversight we might also want to update the following places:

https://github.com/phoenixframework/phoenix/blob/249c286e4e353571853c49c74bab9e10e4170cb8/lib/phoenix/channel.ex#L17
https://github.com/phoenixframework/phoenix/blob/249c286e4e353571853c49c74bab9e10e4170cb8/lib/phoenix/channel.ex#L20
https://github.com/phoenixframework/phoenix/blob/249c286e4e353571853c49c74bab9e10e4170cb8/lib/phoenix/channel.ex#L92
https://github.com/phoenixframework/phoenix/blob/249c286e4e353571853c49c74bab9e10e4170cb8/lib/phoenix/channel.ex#L95

njwest:
My pleasure! Thanks for maintaining such a great framework :P The Phoenix and Elixir teams are the best!

This syntax may have been intentional, but even so, I recommend editing the docs to include `Web`; as-is, the snippets may slow default Phoenix endpoint users down, as was my experience having to debug using `App.Endpoint.Broadcast` rather than `AppWeb.Endpoint.Broadcast` in a newly generated app

If a user has declared their own endpoint outside of the Web module in their Phx project, they're more likely to know to use App.Endpoint rather than AppWeb.Endpoint than users who are starting from generated Phx modules/

I'm inclined to edit the reference in `channels.ex` as well, as that is how the module is referenced [in the Channels testing guide](https://github.com/phoenixframework/phoenix/blob/master/guides/testing/testing_channels.md#L95).

I'll push `channel.ex` edits to the fork in case that is what y'all desire :P

josevalim:

PR #4096, opened 2020-12-07
Update phx.gen.auth confirmation controller template

jrdnull:
See: https://github.com/aaronrenner/phx_gen_auth/pull/102


aaronrenner (approved):
Thanks @jrdnull! :tada:

PR #4098, opened 2020-12-07
show correct bound address when server is started

### LINKED ISSUE #4097: Show custom bound address when server starts
### Environment

* Elixir version (elixir -v): 1.11.2 compiled with Erlang/OTP 23
* Phoenix version (mix deps): 1.5.4
* NodeJS version (node -v): v10.16.0
* NPM version (npm -v): 6.9.0
* Operating system: MacOs Big Sur 11.0.1

### Expected behavior
When running mix phx.server with the following config:
```elixir
http: [port: 4000, ip: {172, 20, 10 , 3}]
```

It is expected that Phoenix output the following to the terminal during server startup:

```bash
[info] Running DevAppWeb.Endpoint with cowboy 2.8.0 at 172.20.10.3:4000 (http)
[info] Access DevAppWeb.Endpoint at http://172.20.10.3:4000 (http)
```

### Actual behavior
Instead it outputs:
```bash
[info] Running DevAppWeb.Endpoint with cowboy 2.8.0 at 172.20.10.3:4000 (http)
[info] Access DevAppWeb.Endpoint at http://0.0.0.0:4000 (http)
```

When a user clicks on the link that is generated the browser can not access the page because the server is not bound to http:0.0.0.0:4000 but rather http:172.20.10.3:4000 as is expected.

AlexJuca:
This PR fixes #4097 so phoenix can show the correct ip when we a custom bound address is configured. @aaronrenner

### DIFF
--- a/lib/phoenix/endpoint/supervisor.ex
+++ b/lib/phoenix/endpoint/supervisor.ex
@@ -433,7 +433,21 @@ defmodule Phoenix.Endpoint.Supervisor do
 
   defp log_access_info(otp_app, endpoint) do
     if Phoenix.Endpoint.server?(otp_app, endpoint) do
-      Logger.info("Access #{inspect(endpoint)} at #{endpoint.url()}")
+      config = config(otp_app, endpoint)
+      for {scheme, _} <- [http: 4000, https: 4040], _ = config[scheme] do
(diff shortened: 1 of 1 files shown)


josevalim:
See discussion on #4097 on why this is not the best option forward. We want to log the url, not the bound ip+port. :) Thanks!

PR #4100, opened 2020-12-10
Add Tailwind to phx.new

dbernheisel:
This adds a `--tailwind` flag to phx.new generator. It's off by default.

Because Tailwind has utility classes that are largely in the HTML, this required me to reorganize a bit how the CSS and HTML templates are organized:

- Existing HTML templates remain as-is, but is now referred to as `:html_sass` or `:sass`
- Move milligram.css out of the phoenix.css and include it as a separate file. 
- Move LiveView-specific CSS out of `app.css` into it's own `phoenix_live.css` file. This leaves `app.css` as only imports of relative files or dependencies.
- Internally, this also sets up the project generator to recognize `sass` as an option; it is set to the opposite of `--tailwind` for now and not accessible via CLI arguments.

---

## Tailwind

- Adds `mix phx.new --tailwind` that is compatible with `--live`. There are new runtime assertions that require `--tailwind` to have HTML and Webpack flags enabled. (cannot have `--no-webpack --tailwind` or `--no-html --tailwind`).
- When not using Tailwind, the `phoenix.css` file is included. When using Tailwind, `milligram.css` and `phoenix.css` are excluded and utility classes are used instead.
- `phoenix_live.css` is always included when `--live` is used, regardless of Tailwind.
- Removes `sass` as the CSS preprocessor and replaces it with `tailwindcss`, `autoprefixer`, `postcss`, `postcss-import`, `postcss-nested`, and `postcss-loader` for Webpack. These PostCSS plugins make the CSS syntax very Sass-like.
- Sets up the project with `app.css` instead of `app.scss` and webpack to match on `*.css` instead of `*.[s]?css`
- The template for handling flash messages is quite different. Before, it would always render the elements even if flash messages weren't present. This is ok for simple styling when the computed height/width was 0, but with Tailwind classes it's harder to do so now it's conditionally rendered if there are flash messages. LMK if you see an opportunity to simplify this.
---

## Screenshots

I made an effort to make the styling stay consistent.

phoenix-new-tailwind-option
phoenix-tailwind-comparison
phoenix-tailwind-live-comparison

## Template Diffs

**LiveView**: 

<details>
<summary>layouts/app.html.eex</summary>

```diff
diff --git a/../my_app_sass/lib/my_app_sass_web/templates/layout/app.html.eex b/lib/my_app_web/templates/layout/app.html.eex
index 09ffdad..bbd1dcf 100644
--- a/../my_app_sass/lib/my_app_sass_web/templates/layout/app.html.eex
+++ b/lib/my_app_web/templates/layout/app.html.eex
@@ -1,5 +1,37 @@
-<main role="main" class="container">
-  <p class="alert alert-info" role="alert"><%= get_flash(@conn, :info) %></p>
-  <p class="alert alert-danger" role="alert"><%= get_flash(@conn, :error) %></p>
+<main role="main" class="max-w-7xl py-0 px-8 my-0 mx-auto w-full relative">
+  <%= if info = get_flash(@conn, :info) do %>
+    <div class="rounded-md mb-4 bg-blue-100 p-4">
+      <div class="flex">
+        <div class="flex-shrink-0">
+          <svg class="h-5 w-5 text-blue-400" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="currentColor" aria-hidden="true">
+            <path fill-rule="evenodd" d="M18 10a8 8 0 11-16 0 8 8 0 0116 0zm-7-4a1 1 0 11-2 0 1 1 0 012 0zM9 9a1 1 0 000 2v3a1 1 0 001 1h1a1 1 0 100-2v-3a1 1 0 00-1-1H9z" clip-rule="evenodd" />
+          </svg>
+        </div>
+        <div class="ml-3 flex-1 md:flex md:justify-between">
+          <p class="text-sm text-blue-700" role="alert">
+            <%= info %>
+          </p>
+        </div>
+      </div>
+    </div>
+  <% end %>
+
+  <%= if error = get_flash(@conn, :error) do %>
+    <div class="rounded-md mb-4 bg-red-100 p-4">
+      <div class="flex">
+        <div class="flex-shrink-0">
+          <svg class="h-5 w-5 text-red-400" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="currentColor" aria-hidden="true">
+            <path fill-rule="evenodd" d="M10 18a8 8 0 100-16 8 8 0 000 16zM8.707 7.293a1 1 0 00-1.414 1.414L8.586 10l-1.293 1.293a1 1 0 101.414 1.414L10 11.414l1.293 1.293a1 1 0 001.414-1.414L11.414 10l1.293-1.293a1 1 0 00-1.414-1.414L10 8.586 8.707 7.293z" clip-rule="evenodd" />
+          </svg>
+        </div>
+        <div class="ml-3 flex-1 md:flex md:justify-between">
+          <p class="text-sm text-red-700" role="alert">
+            <%= error %>
+          </p>
+        </div>
+      </div>
+    </div>
+  <% end %>
+
   <%= @inner_content %>
 </main>
```
</details>

<details>
<summary>layouts/root.html.leex</summary>

```diff
diff --git a/../my_app_sass/lib/my_app_sass_web/templates/layout/root.html.leex b/lib/my_app_web/templates/layout/root.html.leex
index eb27091..dd4ca90 100644
--- a/../my_app_sass/lib/my_app_sass_web/templates/layout/root.html.leex
+++ b/lib/my_app_web/templates/layout/root.html.leex
@@ -5,23 +5,23 @@
     <link phx-track-static rel="stylesheet" href="<%= Routes.static_path(@conn, "/css/app.css") %>"/>
     <script defer phx-track-static type="text/javascript" src="<%= Routes.static_path(@conn, "/js/app.js") %>"></script>
   </head>
-  <body>
-    <header>
-      <section class="container">
+  <body class="m-2">
+    <header class="w-full border-b-2 border-gray-100 mb-6">
+      <section class="max-w-7xl px-6 w-full my-0 mx-auto relative flex flex-col-reverse md:flex-row-reverse justify-between items-center">
         <nav role="navigation">
-          <ul>
-            <li><a href="https://hexdocs.pm/phoenix/overview.html">Get Started</a></li>
+          <ul class="block m-3 mt-0 md:mt-3 text-right whitespace-nowrap">
+            <li><a class="text-blue-500 hover:text-gray-500 block" href="https://hexdocs.pm/phoenix/overview.html">Get Started</a></li>
             <%= if function_exported?(Routes, :live_dashboard_path, 2) do %>
-              <li><%= link "LiveDashboard", to: Routes.live_dashboard_path(@conn, :home) %></li>
+              <li><%= link "LiveDashboard", to: Routes.live_dashboard_path(@conn, :home), class: "text-blue-500 hover:text-gray-500 block" %></li>
             <% end %>
           </ul>
         </nav>
-        <a href="https://phoenixframework.org/" class="phx-logo">
-          <img src="<%= Routes.static_path(@conn, "/images/phoenix.png") %>" alt="Phoenix Framework Logo"/>
+        <a href="https://phoenixframework.org/" class="my-3 md:my-4 mx-3 block">
+          <img class="max-w-lg" src="<%= Routes.static_path(@conn, "/images/phoenix.png") %>" alt="Phoenix Framework Logo"/>
         </a>
       </section>
     </header>
```
</details>

<details>
<summary>layouts/live.html.leex</summary>

```diff
diff --git a/../my_app_sass/lib/my_app_sass_web/templates/layout/live.html.leex b/lib/my_app_web/templates/layout/live.html.leex
index 8dcded5..b077097 100644
--- a/../my_app_sass/lib/my_app_sass_web/templates/layout/live.html.leex
+++ b/lib/my_app_web/templates/layout/live.html.leex
@@ -1,11 +1,37 @@
-<main role="main" class="container">
-  <p class="alert alert-info" role="alert"
-    phx-click="lv:clear-flash"
-    phx-value-key="info"><%= live_flash(@flash, :info) %></p>
+<main role="main" class="max-w-7xl py-0 px-8 my-0 mx-auto w-full relative">
+  <%= if @flash["info"] do %>
+    <div phx-click="lv:clear-flash" phx-value-key="info" class="rounded-md mb-4 bg-blue-100 p-4">
+      <div class="flex">
+        <div class="flex-shrink-0">
+          <svg class="h-5 w-5 text-blue-400" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="currentColor" aria-hidden="true">
+            <path fill-rule="evenodd" d="M18 10a8 8 0 11-16 0 8 8 0 0116 0zm-7-4a1 1 0 11-2 0 1 1 0 012 0zM9 9a1 1 0 000 2v3a1 1 0 001 1h1a1 1 0 100-2v-3a1 1 0 00-1-1H9z" clip-rule="evenodd" />
+          </svg>
+        </div>
+        <div class="ml-3 flex-1 md:flex md:justify-between">
+          <p class="text-sm text-blue-700" role="alert">
+            <%= live_flash(@flash, :info) %>
+          </p>
+        </div>
+      </div>
+    </div>
+  <% end %>
 
-  <p class="alert alert-danger" role="alert"
-    phx-click="lv:clear-flash"
-    phx-value-key="error"><%= live_flash(@flash, :error) %></p>
+  <%= if @flash["error"] do %>
+    <div phx-click="lv:clear-flash" phx-value-key="error" class="rounded-md mb-4 bg-red-100 p-4">
+      <div class="flex">
+        <div class="flex-shrink-0">
+          <svg class="h-5 w-5 text-red-400" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="currentColor" aria-hidden="true">
+            <path fill-rule="evenodd" d="M10 18a8 8 0 100-16 8 8 0 000 16zM8.707 7.293a1 1 0 00-1.414 1.414L8.586 10l-1.293 1.293a1 1 0 101.414 1.414L10 11.414l1.293 1.293a1 1 0 001.414-1.414L11.414 10l1.293-1.293a1 1 0 00-1.414-1.414L10 8.586 8.707 7.293z" clip-rule="evenodd" />
+          </svg>
+        </div>
+        <div class="ml-3 flex-1 md:flex md:justify-between">
+          <p class="text-sm text-red-700" role="alert">
+            <%= live_flash(@flash, :error) %>
+          </p>
+        </div>
+      </div>
+    </div>
+  <% end %>
 
   <%= @inner_content %>
 </main>
```
</details>

<details>
<summary>live/page.html.leex</summary>

```diff
diff --git a/../my_app_sass/lib/my_app_sass_web/live/page_live.html.leex b/lib/my_app_web/live/page_live.html.leex
index 52509c2..69f23ed 100644
--- a/../my_app_sass/lib/my_app_sass_web/live/page_live.html.leex
+++ b/lib/my_app_web/live/page_live.html.leex
@@ -1,47 +1,47 @@
-<section class="phx-hero">
-  <h1><%= gettext "Welcome to %{name}!", name: "Phoenix" %></h1>
-  <p>Peace of mind from prototype to production</p>
+<section class="text-center border-b-2 border-gray-100 bg-gray-100 rounded-lg mb-8 p-12 text-lg">
+  <h1 class="text-4xl font-normal leading-tight tracking-tighter mb-4"><%= gettext "Welcome to %{name}!", name: "Phoenix" %></h1>
+  <p class="mb-4">Peace of mind from prototype to production</p>
 
-  <form phx-change="suggest" phx-submit="search">
-    <input type="text" name="q" value="<%= @query %>" placeholder="Live dependency search" list="results" autocomplete="off"/>
+  <form class="mb-2" phx-change="suggest" phx-submit="search">
+    <input class="mb-4 p-2 shadow-sm focus:ring-indigo-500 focus:border-indigo-500 block w-full sm:text-sm border-gray-300 rounded-md" type="text" name="q" value="<%= @query %>" placeholder="Live dependency search" list="results" autocomplete="off"/>
     <datalist id="results">
       <%= for {app, _vsn} <- @results do %>
         <option value="<%= app %>"><%= app %></option>
       <% end %>
     </datalist>
-    <button type="submit" phx-disable-with="Searching...">Go to Hexdocs</button>
+    <button class="uppercase inline-flex items-center px-6 py-3 border border-transparent text-xs font-bold rounded-md shadow-sm text-white bg-indigo-600 hover:bg-indigo-700 focus:outline-none focus:ring-2 focus:ring-offset-2 focus:ring-indigo-500" type="submit" phx-disable-with="Searching...">Go to Hexdocs</button>
   </form>
 </section>
 
-<section class="row">
-  <article class="column">
-    <h2>Resources</h2>
-    <ul>
+<section class="flex">
+  <article class="flex-1">
+    <h2 class="text-3xl font-medium tracking-tight mb-4">Resources</h2>
+    <ul class="w-full mb-5 list-inside list-disc space-y-2">
       <li>
-        <a href="https://hexdocs.pm/phoenix/overview.html">Guides &amp; Docs</a>
+        <a class="text-blue-500 hover:text-gray-500" href="https://hexdocs.pm/phoenix/overview.html">Guides &amp; Docs</a>
       </li>
       <li>
-        <a href="https://github.com/phoenixframework/phoenix">Source</a>
+        <a class="text-blue-500 hover:text-gray-500" href="https://github.com/phoenixframework/phoenix">Source</a>
       </li>
       <li>
-        <a href="https://github.com/phoenixframework/phoenix/blob/v1.5/CHANGELOG.md">v1.5 Changelog</a>
+        <a class="text-blue-500 hover:text-gray-500" href="https://github.com/phoenixframework/phoenix/blob/v1.5/CHANGELOG.md">v1.5 Changelog</a>
       </li>
     </ul>
   </article>
-  <article class="column">
-    <h2>Help</h2>
-    <ul>
+  <article class="flex-1">
+    <h2 class="text-3xl font-medium tracking-tight mb-4">Help</h2>
+    <ul class="mb-5 list-inside list-disc space-y-2">
       <li>
-        <a href="https://elixirforum.com/c/phoenix-forum">Forum</a>
+        <a class="text-blue-500 hover:text-gray-500" href="https://elixirforum.com/c/phoenix-forum">Forum</a>
       </li>
       <li>
-        <a href="https://webchat.freenode.net/?channels=elixir-lang">#elixir-lang on Freenode IRC</a>
+        <a class="text-blue-500 hover:text-gray-500" href="https://webchat.freenode.net/?channels=elixir-lang">#elixir-lang on Freenode IRC</a>
       </li>
       <li>
-        <a href="https://twitter.com/elixirphoenix">Twitter @elixirphoenix</a>
+        <a class="text-blue-500 hover:text-gray-500" href="https://twitter.com/elixirphoenix">Twitter @elixirphoenix</a>
       </li>
       <li>
-        <a href="https://elixir-slackin.herokuapp.com/">Elixir on Slack</a>
+        <a class="text-blue-500 hover:text-gray-500" href="https://elixir-slackin.herokuapp.com/">Elixir on Slack</a>
       </li>
     </ul>
   </article>
```
</details>

**DeadView**

<details>
<summary>layouts/app.html.eex</summary>

```diff
diff --git a/my_app_dead/lib/my_app_dead_web/templates/layout/app.html.eex b/my_app_dead_tailwind/lib/my_app_dead_tailwind_web/templates/layout/app.html.eex
index 7187115..87d2544 100644
--- a/my_app_dead/lib/my_app_dead_web/templates/layout/app.html.eex
+++ b/my_app_dead_tailwind/lib/my_app_dead_tailwind_web/templates/layout/app.html.eex
@@ -1,32 +1,62 @@
     <link rel="stylesheet" href="<%= Routes.static_path(@conn, "/css/app.css") %>"/>
     <script defer type="text/javascript" src="<%= Routes.static_path(@conn, "/js/app.js") %>"></script>
   </head>
-  <body>
-    <header>
-      <section class="container">
+  <body class="m-2">
+    <header class="w-full border-b-2 border-gray-100 mb-6">
+      <section class="max-w-7xl px-6 w-full my-0 mx-auto relative flex flex-col-reverse md:flex-row-reverse justify-between items-center">
         <nav role="navigation">
-          <ul>
-            <li><a href="https://hexdocs.pm/phoenix/overview.html">Get Started</a></li>
+          <ul class="block m-3 mt-0 md:mt-3 text-right whitespace-nowrap">
+            <li><a class="text-blue-500 hover:text-gray-500 block" href="https://hexdocs.pm/phoenix/overview.html">Get Started</a></li>
             <%= if function_exported?(Routes, :live_dashboard_path, 2) do %>
-              <li><%= link "LiveDashboard", to: Routes.live_dashboard_path(@conn, :home) %></li>
+              <li><%= link "LiveDashboard", to: Routes.live_dashboard_path(@conn, :home), class: "text-blue-500 hover:text-gray-500 block" %></li>
             <% end %>
           </ul>
         </nav>
-        <a href="https://phoenixframework.org/" class="phx-logo">
-          <img src="<%= Routes.static_path(@conn, "/images/phoenix.png") %>" alt="Phoenix Framework Logo"/>
+        <a href="https://phoenixframework.org/" class="my-3 md:my-4 mx-3 block">
+          <img class="max-w-lg" src="<%= Routes.static_path(@conn, "/images/phoenix.png") %>" alt="Phoenix Framework Logo"/>
         </a>
       </section>
     </header>
-    <main role="main" class="container">
-      <p class="alert alert-info" role="alert"><%= get_flash(@conn, :info) %></p>
-      <p class="alert alert-danger" role="alert"><%= get_flash(@conn, :error) %></p>
+    <main role="main" class="max-w-7xl py-0 px-8 my-0 mx-auto w-full relative">
+      <%= if info = get_flash(@conn, :info) do %>
+        <div class="rounded-md mb-4 bg-blue-100 p-4">
+          <div class="flex">
+            <div class="flex-shrink-0">
+              <svg class="h-5 w-5 text-blue-400" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="currentColor" aria-hidden="true">
+                <path fill-rule="evenodd" d="M18 10a8 8 0 11-16 0 8 8 0 0116 0zm-7-4a1 1 0 11-2 0 1 1 0 012 0zM9 9a1 1 0 000 2v3a1 1 0 001 1h1a1 1 0 100-2v-3a1 1 0 00-1-1H9z" clip-rule="evenodd" />
+              </svg>
+            </div>
+            <div class="ml-3 flex-1 md:flex md:justify-between">
+              <p class="text-sm text-blue-700" role="alert">
+                <%= info %>
+              </p>
+            </div>
+          </div>
+        </div>
+      <% end %>
+
+      <%= if error = get_flash(@conn, :error) do %>
+        <div class="rounded-md mb-4 bg-red-100 p-4">
+          <div class="flex">
+            <div class="flex-shrink-0">
+              <svg class="h-5 w-5 text-red-400" xmlns="http://www.w3.org/2000/svg" viewBox="0 0 20 20" fill="currentColor" aria-hidden="true">
+                <path fill-rule="evenodd" d="M10 18a8 8 0 100-16 8 8 0 000 16zM8.707 7.293a1 1 0 00-1.414 1.414L8.586 10l-1.293 1.293a1 1 0 101.414 1.414L10 11.414l1.293 1.293a1 1 0 001.414-1.414L11.414 10l1.293-1.293a1 1 0 00-1.414-1.414L10 8.586 8.707 7.293z" clip-rule="evenodd" />
+              </svg>
+            </div>
+            <div class="ml-3 flex-1 md:flex md:justify-between">
+              <p class="text-sm text-red-700" role="alert">
+                <%= error %>
+              </p>
+            </div>
+          </div>
+        </div>
+      <% end %>
       <%= @inner_content %>
     </main>
   </body>
```
</details>

<details>
<summary>templates/index.html.eex</summary>

```diff
diff --git a/my_app_dead/lib/my_app_dead_web/templates/page/index.html.eex b/my_app_dead_tailwind/lib/my_app_dead_tailwind_web/templates/page/index.html.eex
index dd0a7f8..d87f329 100644
--- a/my_app_dead/lib/my_app_dead_web/templates/page/index.html.eex
+++ b/my_app_dead_tailwind/lib/my_app_dead_tailwind_web/templates/page/index.html.eex
@@ -1,37 +1,37 @@
-<section class="phx-hero">
-  <h1><%= gettext "Welcome to %{name}!", name: "Phoenix" %></h1>
-  <p>Peace of mind from prototype to production</p>
+<section class="text-center border-b-2 border-gray-100 bg-gray-100 rounded-lg mb-8 p-12 text-lg">
+  <h1 class="text-4xl font-normal leading-tight tracking-tighter mb-4"><%= gettext "Welcome to %{name}!", name: "Phoenix" %></h1>
+  <p class="mb-4">Peace of mind from prototype to production</p>
 </section>
 
-<section class="row">
-  <article class="column">
-    <h2>Resources</h2>
-    <ul>
+<section class="flex">
+  <article class="flex-1">
+    <h2 class="text-3xl font-medium tracking-tight mb-4">Resources</h2>
+    <ul class="w-full mb-5 list-inside list-disc space-y-2">
       <li>
-        <a href="https://hexdocs.pm/phoenix/overview.html">Guides &amp; Docs</a>
+        <a class="text-blue-500 hover:text-gray-500" href="https://hexdocs.pm/phoenix/overview.html">Guides &amp; Docs</a>
       </li>
       <li>
-        <a href="https://github.com/phoenixframework/phoenix">Source</a>
+        <a class="text-blue-500 hover:text-gray-500" href="https://github.com/phoenixframework/phoenix">Source</a>
       </li>
       <li>
-        <a href="https://github.com/phoenixframework/phoenix/blob/v1.5/CHANGELOG.md">v1.5 Changelog</a>
+        <a class="text-blue-500 hover:text-gray-500" href="https://github.com/phoenixframework/phoenix/blob/v1.5/CHANGELOG.md">v1.5 Changelog</a>
       </li>
     </ul>
   </article>
-  <article class="column">
-    <h2>Help</h2>
-    <ul>
+  <article class="flex-1">
+    <h2 class="text-3xl font-medium tracking-tight mb-4">Help</h2>
+    <ul class="mb-5 list-inside list-disc space-y-2">
       <li>
-        <a href="https://elixirforum.com/c/phoenix-forum">Forum</a>
+        <a class="text-blue-500 hover:text-gray-500" href="https://elixirforum.com/c/phoenix-forum">Forum</a>
       </li>
       <li>
-        <a href="https://webchat.freenode.net/?channels=elixir-lang">#elixir-lang on Freenode IRC</a>
+        <a class="text-blue-500 hover:text-gray-500" href="https://webchat.freenode.net/?channels=elixir-lang">#elixir-lang on Freenode IRC</a>
       </li>
       <li>
-        <a href="https://twitter.com/elixirphoenix">Twitter @elixirphoenix</a>
+        <a class="text-blue-500 hover:text-gray-500" href="https://twitter.com/elixirphoenix">Twitter @elixirphoenix</a>
       </li>
       <li>
-        <a href="https://elixir-slackin.herokuapp.com/">Elixir on Slack</a>
+        <a class="text-blue-500 hover:text-gray-500" href="https://elixir-slackin.herokuapp.com/">Elixir on Slack</a>
       </li>
     </ul>
   </article>
```
</details>


josevalim:
Hi @dbernheisel,

While we have discussed moving to tailwind, we are not 100% sure yet. If we do move to it though, we wouldn't use it as an option, but rather the default. Your PR is a great first step though, because it helps us assess the changes before we fully commit to it, so thank you!

dbernheisel:
Sure thing! I'll make another commit to make it the default. If it ends up not as a default then we could drop the new commits.

dbernheisel:
Updated to use Tailwind CSS by default without a flag.

- Removes phoenix.css
- Removes milligram.css
- Removes any usage of Sass.

A remaining issue at the moment is if a user generates a project with `--no-webpack` they will no longer get a styled welcome page. To get it to look good again, this would require separating the HTML for non-webpack projects. It generates with an empty `app.css` and a `page.html.eex` with Tailwind classes.

henrik:
Could it make sense to get these steps merged as a separate PR, assuming they're seen as an incremental improvement independent of our Tailwind future? Would also make that future merge a little smaller.

> - Move milligram.css out of the phoenix.css and include it as a separate file.
> - Move LiveView-specific CSS out of app.css into it's own phoenix_live.css file. This leaves app.css as only imports of relative files or dependencies.

chrismccord:
Hold off for now on continued efforts. The Phoenix team has a number of items to complete first before we can explore this direction, so things are subject to change on what tailwind integration could look like.

lessless:
@dbernheisel is there any real value in removing sass?

dbernheisel:
> @dbernheisel is there any real value in removing sass? 

Yes, because Tailwind recommends using PostCSS as the preprocessor to avoid quirks when running both PostCSS and a separate  preprocessor like Sass.
 https://tailwindcss.com/docs/using-with-preprocessors#using-post-css-as-your-preprocessor

vrcca:
> A remaining issue at the moment is if a user generates a project with `--no-webpack` they will no longer get a styled welcome page. To get it to look good again, this would require separating the HTML for non-webpack projects. It generates with an empty `app.css` and a `page.html.eex` with Tailwind classes.


We could default to [using the CDN](https://tailwindcss.com/docs/installation#using-tailwind-via-cdn) instead. Wdyt?

kevinlang:
Thanks for this PR, @dbernheisel!

I used this locally by merging it into an up to date branch in my own fork, creating a new `phx.new` archive from the new code, and using that to generate my project. :)

c4710n:
I have another idea about this.

We can provide a more flexible option, like `--assets-repo`. It specify the git repo which is used for initializing the `assets` dir.

This method has following PROS:
+ less efforts to maintain the assets related generator. (Maintaining generators is a hard work）
+ won't bind  Phoneix to a specific framework, such as TailwindCSS, Bulma, etc. (Frontend related tech is evolving fast, it is not a good idea to bind them to Phoenix)

Then:
+ Phoenix team provides an example `assets` dir which shows the basic interface which Phoenix needs, just like now.
+ Developers who don't like the default setup, implement their own assets repo.

type1fool:
I am excited to see Tailwind moving closer to adoption either as the default CSS framework or behind a flag. Over the past month, I have been using the experimental JIT compiler in a Laravel project, and the Webpack compile time dropped from 10-30 seconds to ~1 second during development. Phoenix's asset pipeline has not suffered as much in my experience, but I have not worked in a full-fledged Phoenix app using Tailwind.

Obviously, this tight feedback loop is useful, and the resulting CSS file is in the 50KB range as opposed to the 11MB file produced without JIT. JIT mode is still experimental in Tailwind 2.1 due to some [known limitations](https://tailwindcss.com/docs/just-in-time-mode#known-limitations), but they've been making fast progress.

Is this on the radar?

https://tailwindcss.com/docs/just-in-time-mode
https://github.com/tailwindlabs/tailwindcss/issues

BartOtten:
Wasn’t Bulma the new kid on the block? Ow no, that was last year.

I see why this PR has been made; as this year Tailwind is the coolest kid in town. However; I think Phoenix should stick to pure CSS without JS (Webpack) dependency. Once you go Tailwind style, there is no coming back from it unless you rewrite your templates as the style is highly coupled with the structure.

Adding support for ‘variants’ of the default Phoenix setup does seem like a better solution to me. Everybody can pick their latest greatest and Phoenix itself could keep a default which does not require JS.

AndrewDryga:
@c4710n isn't this is already possible if people just publish their own generators (as mix archives to Hex) nowadays? Phoenix then would just need to provide a common library that they can use as a dependency.

I like this idea a lot because frameworks should not be too opinionated, front-end world changes too quickly. There should be one safe and simple default, but having a flag for each front-end framework is too much to remember (or to remove as obsolete stuff) after code generation.

c4710n:
@AndrewDryga That is right. 

But in my opinion, maintaining generators is a hard work. You have to:
1. change generator.
2. generate code.
3. test generated code.
4. oops, something wrong, go back to step 1.

 If there is no additional code to execute[^1], it doesn't have to be a generator. It can be a simple Git repo[^2], and it has other benefits:
+ code can be maintained as a real project, no need to repeat above 4 steps.
+ other people can preview the project directly.

And, in this way, **Phoenix doesn't need to provide anything.** It's ok to replace `assets` manually[^3]. But it is better if it can be done like I said above.

[^1]: I think this is true for `assets` dir. In this directory, there's no application related names, no need to make it dynamic.
[^2]: For example, in my own [repo](https://github.com/c4710n/PETAL-starter), I mainly maintain the directory `apps/hello_web/assets` which can be copied to any projects. No need to generate elixir code.
[^3]: No one is always initializing the project. Generally, It will be done only once.

lessless:
Well, 

Adopting a particular CSS framework is a big decision with long-lasting and broad consequences.
On the other side if generator will get interest from the community and will organically become a defacto standard then it might make sense to merge it into Phoenix.

That will also solve the burden of maintenance - other Tailwind users will be happy to contribute to something that solves their pain.

/Yevhenii

foucist:
> We could default to [using the CDN](https://tailwindcss.com/docs/installation#using-tailwind-via-cdn) instead. Wdyt?

It's common to customize/extend tailwindcss and then run PurgeCSS after to end up with a nice 5kb or 10kb css file.  The CDN approach would miss out on that.

> Wasn’t Bulma the new kid on the block? Ow no, that was last year.

Bulma was started beginning of 2016, Tailwind CSS came out a year and a half later in late 2017.  They're both very different frameworks, I would describe Bulma as more like Bootstrap, it provides a fixed look and a variety of components out of the box, but if you want to customize things heavily then you still have to fall back to "raw" css.   

Whereas Tailwind CSS is more of an atomic css framework where you can make your own components by assembling the the utility classes given, meaning there's less need to write "raw" css.  I think it's one of the more accessible & user-friendly front-end approaches for people like me that are bad at css.  It also acts like a design language, all the spacing and font sizing and colors and hues are all setup consistently within the utility classes.

woylie:
I wouldn't like to see Tailwind as the default for Phoenix. The current approach with inlining Milligram and some custom styles is the perfect choice: It gives you some basic configuration and styles to work with and get started, without bringing a baggage of dependencies on board and pushing you into any particular direction. Especially one where you're encouraged leaving standard-based "raw" css in favour of arbitrary framework-specific utility classes. It's easy enough to add any heavy-weight CSS framework to the current default installation. So if you want to add generator support for Tailwind, please make it optional.

dbernheisel:
With the change to move to esbuild 🎉, I don't believe this makes sense as-is anymore.

foucist:
In case anyone else stumbles across this page, check out the new mix task for tailwind https://github.com/phoenixframework/tailwind

PR #4106, opened 2020-12-13
Add redact option to fields in generators

TheFirstAvenger:
`mix phx.gen.html Accounts Superhero superheroes name secret_identity:redact`

When redact is used in a generator, it is added to the generated schema:

```elixir
    field :secret_identity, :string, redact: true
```

~~`text_input` form elements in html/live templates become `password_input`:~~ (Feature Removed)

```elixir
  <%= label f, :secret_identity %>
  <%= password_input f, :secret_identity %>
  <%= error_tag f, :secret_identity %>
```

~~and show/index generated templates in html/live don't list the field:~~(Feature Removed)

```elixir
<ul>

  <li>
    <strong>Name:</strong>
    <%= @superhero.name %>
  </li>

</ul>
```


lib/mix/phoenix/schema.ex
```diff
@@ -158,6 +170,8 @@ defmodule Mix.Phoenix.Schema do
     Enum.map(attrs, fn attr ->
       attr
       |> drop_unique()
+      |> drop_redact()
+      |> drop_unique()
```
josevalim, line 174: This makes me think we should probably refactor the code so we do a pre-processing on all attributes collecting redact/unique. WDYT?

TheFirstAvenger: @josevalim I rewrote the code to pre-process as you suggested. It both collects the lists of uniques/redacts and removes them from the cli_attrs so subsequent calls don't have to all manually ignore them.

lib/mix/tasks/phx.gen.html.ex
```diff
@@ -190,7 +190,7 @@ defmodule Mix.Tasks.Phx.Gen.Html do
       {key, {:enum, _}}  ->
         {label(key), ~s|<%= select f, #{inspect(key)}, Ecto.Enum.values(#{inspect(schema.module)}, #{inspect(key)}), prompt: "Choose a value" %>|, error(key)}
       {key, _}  ->
-        {label(key), ~s(<%= text_input f, #{inspect(key)} %>), error(key)}
+        {label(key), ~s|<%= #{maybe_password("text_input", schema, key)} f, #{inspect(key)} %>|, error(key)}
```
josevalim, line 193: I don't think we should treat all redacts as passwords. Due to GDPR, you may want to redact SSN, emails and even addresses!

priv/templates/phx.gen.html/index.html.eex
```diff
@@ -3,15 +3,15 @@
 <table>
   <thead>
     <tr>
-<%= for {k, _} <- schema.attrs do %>      <th><%= Phoenix.Naming.humanize(Atom.to_string(k)) %></th>
+<%= for {k, _} <- schema.attrs, k not in schema.redacts do %>      <th><%= Phoenix.Naming.humanize(Atom.to_string(k)) %></th>
```
josevalim, line 6: Similar to the GDPR above, I wouldn't necessarily remove redacts from here. I think it is fine to keep them here, it will be clearer in the development experience and they can always remove it later.

lib/mix/tasks/phx.gen.html.ex
TheFirstAvenger: I agree that there will be plenty of real world cases where password fields aren't needed. As this is a generator, I like the idea of giving them the password field by default, and they can swap it back if they don't want it (default secure). If we went the other way, they might not know of or think to use the password field.

lib/mix/phoenix/schema.ex
```diff
@@ -142,13 +144,21 @@ defmodule Mix.Phoenix.Schema do
...
+      prefix1 = byte_size(attr) - 7
+      prefix2 = byte_size(attr) - 14
+      attr_name = attr |> String.split(":") |> hd() |> String.to_atom()
+      case attr do
+        <<attr::size(prefix2)-binary, ":redact:unique">> -> {[attr | attrs], [attr_name | uniques], [attr_name | redacts]}
+        <<attr::size(prefix2)-binary, ":unique:redact">> -> {[attr | attrs], [attr_name | uniques], [attr_name | redacts]}
+        <<attr::size(prefix1)-binary, ":unique">> -> {[attr | attrs], [attr_name | uniques], redacts}
+        <<attr::size(prefix1)-binary, ":redact">> -> {[attr | attrs], uniques, [attr_name | redacts]}
+        _ -> {[attr | attrs], uniques, redacts}
+      end
```
josevalim, line 158: What do you think about a function that was recursive and did this:

```
defp split_flags(["unique" | rest], name, attrs, uniques, redacts),
  do: split_flags(rest, name, attrs, [name | uniques], redacts)

defp split_flags(["redact" | rest], name, attrs, uniques, redacts),
  do: split_flags(rest, name, attrs, uniques, [name | redacts])

defp split_flags(rest, _name, attrs, uniques, redacts),
  do: {Enum.join(Enum.reverse(rest), ":"), uniques, redacts}
```

and we would call it like this:

```
[attr_name | rest] = String.split(":")
attr_name = String.to_atom(attr_name)
split_flags(Enum.reverse(rest), attr_name, attrs, uniques, redacts)
```

?

josevalim (commented):
Thanks @TheFirstAvenger! I have dropped just two final comments.

One of them is because I think we are trying to do too much with the redact flag and I think it should only mean `redact: true` in the schema and nothing more. It definitely shouldn't be a password and I think we can continue emitting showing it on the generated forms/showpages/etc. The user can always remove it later (which is easier then adding, especially for new users). WDYT?

priv/templates/phx.gen.html/index.html.eex
TheFirstAvenger: That makes sense. Removing it manually after generation is easy, but manually adding it if we don't generate it is a pain. I will remove this change. I will do the same in the show template.

lib/mix/tasks/phx.gen.html.ex
josevalim: The thing is that you never want to persist a password field on the database (but rather its hash). And because this field is persisted on the database, I am afraid we can rather point people to the wrong direction by doing this.

josevalim: In my mind, the only way it would be ok to be a password would be if it is virtual and redacted but that feels a very corner case scenario that I am not sure if we should bother?

TheFirstAvenger:
@josevalim All set, I pulled everything back except the redact: true in the schema, and implemented the extraction as you suggested.

lib/mix/tasks/phx.gen.html.ex
chrismccord: Very good point about pointing folks in the wrong direction. I like the convenience of redact making a password field but José is right that it requires more than simply mapping a field back to the DB, so I think we need to drop the password approach as it's going to necessarily require more in user-land to be done properly

josevalim:

PR #4112, opened 2020-12-20
Update runtime configuratin guides

wojtekmach:


guides/deployment/releases.md
```diff
@@ -87,32 +87,11 @@ Release created at _build/prod/rel/my_app!
 
 And starting the release now should also successfully start the web server! Now you can get all of the files under the `_build/prod/rel/my_app` directory, package it, and run it in any production machine with the same OS and architecture as the one that assembled the release. For more details, check the [docs for `mix release`](https://hexdocs.pm/mix/Mix.Tasks.Release.html).
 
-But before we finish this guide, there are two features from releases most Phoenix applications will use, so let's talk about those.
-
-## Runtime configuration
```
wojtekmach, line 92: I decided to simply remove this section however also considered something like this:

```markdown
## Runtime configuration

If you're using Elixir 1.11+ and already have a `config/runtime.exs` file, you can skip this section as your runtime configuration is properly set up.
If you have an existing Phoenix app with `config/prod.secret.exs` or `config/releases.exs`, read on.

(...)
```

PR #4118, opened 2020-12-30
Adding @impl attributes to application modules

akoutmos:
Updates the application module installer templates  to include the impl attributes for Application behaviour callbacks.


chrismccord:
❤️❤️❤️🐥🔥

PR #4119, opened 2020-12-30
Change phx.new so that --no-html implies --no-webpack

wojtekmach:
I think when you don't have html files you won't use JS/CSS either so I think this would be a little bit more convenient, but I also see nothing wrong with explicitly typing the extra --no-webpack (though I almost never remember to do that! :))
﻿


chrismccord:
There should be several discussions around this if you search far enough back, but the tldr; is while your change is a very reasonable one, there is also the very common usecase of an SPA + API, where you indeed want the webpack toolchain so it's a toss up. Thanks!

PR #4121, opened 2020-12-31
Add flag to skip prompt when generator find an existing context

aaronrenner:
This adds new `--merge-with-existing-context` and `--no-merge-with-existing-context` options to the `phx.gen.context` family of generators. This allows us to have an integration test that merges`phx.gen.auth` generated code with an existing context (a prerequisite for me to clean up some duplicate fixture generation code). This test also found a formatting error where we were missing a blank newline between the existing context's functions/tests and `phx.gen.auth`'s functions/tests. I've also added unit tests and documentation for these new options. 

I was unable to add unit tests for `--no-merge-with-existing-context` because `System.halt/0` is called in the generator and kills the tests.


josevalim (approved):

chrismccord:
❤️❤️❤️🐥🔥

PR #4122, opened 2020-12-31
Fix code typo in Router documentation

szTheory:
This one tripped me up a bit coming from Rails. The path should be singular in Phoenix like it is in the other examples surrounding it, right? If I'm wrong sorry for the bad change, can just close the PR.


chrismccord:
❤️❤️❤️🐥🔥

PR #4129, opened 2021-01-06
Add moduledoc to not break when using credo

wesleimp:


chrismccord:
I'd prefer to leave this as is. Thanks!

PR #4130, opened 2021-01-07
CI: include `mix compile` in `test-setup`

mindreframer:
Problem: 
- when re-running `earthly` locally after changing a test Elixir file, we see a full recompilation of all the dependencies. 

Solution: 
- in the test-setup preparation step we can also compile all the packages with `MIX_ENV=test`. That way they get reused for all the dependent steps.

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -9,6 +9,7 @@ all-test:
 
 test:
     FROM +test-setup
+    RUN MIX_ENV=test mix deps.compile
     COPY --dir assets config installer lib integration_test priv test ./
     RUN mix test
 
@@ -102,4 +103,3 @@ test-setup:
    RUN mix local.rebar --force
    RUN mix local.hex --force
    RUN mix deps.get
-   RUN MIX_ENV=test mix compile
--- a/Earthfile
+++ b/Earthfile
@@ -104,3 +104,4 @@ test-setup:
    RUN mix local.rebar --force
    RUN mix local.hex --force
    RUN mix deps.get
+   RUN MIX_ENV=test mix compile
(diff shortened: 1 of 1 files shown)


aaronrenner:
Thanks, @mindreframer! Would you mind pulling the skip ci changes into a separate PR so we can work on both of these changes independently?

mindreframer:
@aaronrenner sure

josevalim (approved):

mindreframer:
@aaronrenner @josevalim skip-ci changes gone :) thanks for the approval!

Earthfile
```diff
@@ -104,3 +102,4 @@ test-setup:
    RUN mix local.rebar --force
    RUN mix local.hex --force
    RUN mix deps.get
+   RUN MIX_ENV=test mix compile
```
aaronrenner, line 105: Would it work to move this to the `:test` target and change it to `MIX_ENV=test mix deps.compile`? Since this is currently in the shared setup, it will also be run before the NPM tests which I don't believe need it.

mindreframer: Yep, seems like a good idea.

mindreframer: @aaronrenner would you mind taking another look?

aaronrenner: The update looks great. I just want to run this locally first. I'll try to get you feedback in the next evening or two.

mindreframer: @aaronrenner OK, thanks!

aaronrenner (approved):
Just tried it out locally and it significantly improves caching. Thanks @mindreframer! :fire:

mindreframer:
@aaronrenner great, thanks for merging!

PR #4138, opened 2021-01-09
Capitalize "Plug"

appdevnick:
When referencing the proper name of the middleware.

### DIFF
--- a/guides/controllers.md
+++ b/guides/controllers.md
@@ -64,7 +64,7 @@ While we can name our actions whatever we like, there are conventions for action
 
 Each of these actions takes two parameters, which will be provided by Phoenix behind the scenes.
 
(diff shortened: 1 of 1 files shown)


aaronrenner:
This makes sense to me, but it looks like we have a merge conflict. @appdevnick would you mind rebasing this off master?

appdevnick:
Apologies if I'm messing anything up. I'm new to Git so still figuring stuff out. I know rebasing is when you take your changes allow git to reapply them as if they are going on top of the master branch, but I've never done it before so please bear with me :)

It looks like what I did was not correct. Do I delete this PR and just resubmit on top of master, or is there a more idiomatic way of approaching it? I will try closing this issue and resubmitting a fresh one in the meantime! Sorry about that

PR #4144, opened 2021-01-10
Update word case

kianmeng:
@Gazler As discussed.


guides/request_lifecycle.md
```diff
@@ -238,7 +238,7 @@ And this is what the template should look like:
 </div>
 ```
 
-Our messenger appears as `@messenger`. We call "assigns" the values passed from the controller to views. It is a special bit of metaprogrammed syntax which stands in for `assigns.messenger`. The result is much nicer on the eyes and much easier to work with in a template.
+Our messenger appears as `@messenger`. We call "assigns" the values passed from the controller to views. It is a special bit of meta-programmed syntax which stands in for `assigns.messenger`. The result is much nicer on the eyes and much easier to work with in a template.
```
aaronrenner, line 241: Wikipedia shows metaprogramming as one word without the dash: https://en.wikipedia.org/wiki/Metaprogramming. Would you mind switching this one back?

aaronrenner (commented):
@kianmeng Thanks for this! Just one small request...

guides/request_lifecycle.md
kianmeng: > Wikipedia shows metaprogramming as one word without the dash: https://en.wikipedia.org/wiki/Metaprogramming. Would you mind switching this one back?

Done.

Gazler:
Thanks <3

PR #4147, opened 2021-01-12
heroku.md: Remove unnecessary "longpoll: false"

henrik:
Per https://hexdocs.pm/phoenix/Phoenix.Endpoint.html#socket/3, this is already the default, so perhaps not worth mentioning.

Unless I'm missing something and this setting is extra important in the context of Heroku – if so, perhaps the docs could say a bit more about that.


josevalim:

PR #4152, opened 2021-01-15
Implement put_new_flash/3

ream88:
When messing around with phx_gen_auth I needed to put a key/message into the flash but only if the key didn't not exist prior. What do you think of this change?


chrismccord:
This is probably best for a function in your own application. You can use `get_flash` to check for the prior existence of the key. Thanks!

ream88:
Thx for the quick feedback, it’s already my own function 😊

PR #4153, opened 2021-01-15
Update template app.scss for modal

cdock1029:
Vertical % for padding and margins is a percent of the width. When maximizing the browser to full screen especially with a large form, this causes modal to be pushed down the page.

"vh" units seems to be simplest fix here, as the modal is used full screen. If you wanted to generalize it to work within any container, you could do absolute and top/bottom percent (as that refers to the height).

https://css-tricks.com/oh-hey-padding-percentage-is-based-on-the-parent-elements-width/


Gazler:
Thanks <3

PR #4158, opened 2021-01-18
Fixing mix task name in Contexts.guide

srikanthkyatham:


guides/contexts.md
```diff
@@ -246,7 +246,7 @@ defmodule Hello.Accounts.User do
 end
 ```
 
-This is just what we saw before when we ran `mix phx.gen.schema`, except here we see a `@doc false` above our `changeset/2` function. This tells us that while this function is publicly callable, it's not part of the public context API. Callers that build changesets do so via the context API. For example, `Accounts.create_user/1` calls into our `User.changeset/2` to build the changeset from user input. Callers, such as our controller actions, do not access `User.changeset/2` directly. All interaction with our user changesets is done through the public `Accounts` context.
+This is just what we saw before when we ran `mix phx.gen.html`, except here we see a `@doc false` above our `changeset/2` function. This tells us that while this function is publicly callable, it's not part of the public context API. Callers that build changesets do so via the context API. For example, `Accounts.create_user/1` calls into our `User.changeset/2` to build the changeset from user input. Callers, such as our controller actions, do not access `User.changeset/2` directly. All interaction with our user changesets is done through the public `Accounts` context.
```
aaronrenner, line 249: This `mix phx.gen.schema` reference is correct assumes the user has completed the ecto guide. Maybe we should add a link to the ecto guide here?

```suggestion
This is just what we saw before when we ran `mix phx.gen.schema` (in [the Ecto guide](ecto.html)), except here we see a `@doc false` above our `changeset/2` function. This tells us that while this function is publicly callable, it's not part of the public context API. Callers that build changesets do so via the context API. For example, `Accounts.create_user/1` calls into our `User.changeset/2` to build the changeset from user input. Callers, such as our controller actions, do not access `User.changeset/2` directly. All interaction with our user changesets is done through the public `Accounts` context.
```

srikanthkyatham: I have changed the to mix phx.gen.html in the commit since it was the referenced earlier in the document. Perhaps you are correct there should be a link to ecto guide

josevalim:
@chrismccord is reworking those guides, so I would wait on this a bit for now. :) Thank you!

chrismccord:
This is now out of date with the context guide rewrite. Thank you! ❤️❤️❤️🐥🔥

PR #4159, opened 2021-01-18
Update `@unsent` Plug.Conn states to include `:set_chunked` and `:set_file`

mveytsman:
We use `@unsent` to raise an `AlreadySentError` if trying to change the view or the layout after a response is sent. This updates the module variable to include the other unsent states from https://github.com/elixir-plug/plug/blob/40efbe4680f98eb063ea1e17f079c4243b6593cf/lib/plug/conn.ex#L258

It *might* be nicer to expose `unsent?`/`sent?` as functions on Plug.Conn itself. What do you think?


josevalim:

PR #4161, opened 2021-01-19
Fix Documentation Links

thiagopromano:
The plug.md links to `localhost` are all broken on Hexdocs because it is including the trailing quote.

phoenixlinks


aaronrenner (approved):
Thanks @thiagopromano!

PR #4166, opened 2021-01-24
Add phx.gen.auth tests to main test suite

aaronrenner:
This adds for `phx.gen.auth` to the main test suite. I found that trying to debug using the integration tests is difficult because they are running in a different BEAM instance and tools like `IO.inspect/1` don't output anything to the console (because output is being captured).

## Things to note

1. In order to get these tests to work, I had to add an optional second parameter to `Mix.Tasks.Phx.Gen.Auth.run/1` that allows me to pass in the database adapter and skip the dependency check. With these two options set, I'm able to bypass the need to compile the app that's being tested and test this like the rest of the `phx.gen.*` tasks.
2. Since `phx.gen.auth` injects code into several files in an existing phoenix project, instead of trying to create those needed files in something like `in_tmp_live_project/1`, I found that I could just call `Mix.Tasks.Phx.New.run/1` and have the project generated for me automatically. This required me adding `phx_new` as a path-based, test dependency, but it doesn't seem to cause any issues and makes the test set up much simpler.
3. Now that I can use `phx.new` to generate projects, not every project needs to be named `:phoenix`. I modified `with_generator_env/3` to accept an optional app name.
4. This test currently takes about 30 seconds to run because of the size of some of the eex templates. My next task is to precompile these templates while still allowing templates to be overridden in the`priv` directory. I added a test in this PR that overrides a template in the priv directory and ensures the override works.
5. I added new tests for some of the user prompts and found some could use formatting improvements. I updated one of the prompts to make it look a little nicer.

Please let me know what you think. I hope these tests will make the developer experience working on `phx.gen.auth` much easier.

### DIFF
--- a/lib/mix/tasks/phx.gen.auth.ex
+++ b/lib/mix/tasks/phx.gen.auth.ex
@@ -85,7 +85,7 @@ defmodule Mix.Tasks.Phx.Gen.Auth do
              table: :string, merge_with_existing_context: :boolean]
 
   @doc false
-  def run(args) do
+  def run(args, test_opts \\ []) do
     if Mix.Project.umbrella?() do
       Mix.raise("mix phx.gen.auth can only be run inside an application directory")
     end
@@ -99,16 +99,21 @@ defmodule Mix.Tasks.Phx.Gen.Auth do
     {context, schema} = Gen.Context.build(context_args, __MODULE__)
     Gen.Context.prompt_for_code_injection(context)
 
-    # Needed so we can get the ecto adapter and ensure other
-    # libraries are loaded.
-    #
-    # As far as I can tell, everything after this must be tested with an
-    # integration test.
-    Mix.Task.run("compile")
+    if Keyword.get(test_opts, :validate_dependencies?, true) do
+      # Needed so we can get the ecto adapter and ensure other
+      # libraries are loaded.
+      Mix.Task.run("compile")
 
-    validate_required_dependencies!()
+      validate_required_dependencies!()
+    end
 
-    ecto_adapter = get_ecto_adapter!(schema)
+    ecto_adapter =
+      Keyword.get_lazy(
+        test_opts,
+        :ecto_adapter,
+        fn -> get_ecto_adapter!(schema) end
+      )
+
     migration = Migration.build(ecto_adapter)
 
     binding = [
@@ -311,7 +316,7 @@ defmodule Mix.Tasks.Phx.Gen.Auth do
       {:error, :unable_to_inject} ->
         Mix.shell().info("""
 
-        Add your #{inspect(mix_dependency)} dependency to #{file_path}:
+        Add your #{mix_dependency} dependency to #{file_path}:
 
             defp deps do
               [
--- a/installer/test/mix_helper.exs
+++ b/installer/test/mix_helper.exs
@@ -98,12 +98,27 @@ defmodule MixHelper do
     end
   end
 
-  def with_generator_env(new_env, fun) do
-    Application.put_env(:phoenix, :generators, new_env)
+  def modify_file(path, function) when is_binary(path) and is_function(function, 1) do
+    path
+    |> File.read!()
+    |> function.()
+    |> write_file!(path)
+  end
+
+  defp write_file!(content, path) do
+    File.write!(path, content)
+  end
+
+  def with_generator_env(app_name \\ :phoenix, new_env, fun) do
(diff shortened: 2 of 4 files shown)


josevalim:
> This adds for phx.gen.auth to the main test suite. I found that trying to debug using the integration tests is difficult because they are running in a different BEAM instance and tools like IO.inspect/1 don't output anything to the console (because output is being captured).

Btw, one option is to have an env var that makes you print the captured result as you capture them (or afterwards), but unit tests are also welcome.

lib/mix/tasks/phx.gen.auth.ex
```diff
@@ -99,16 +99,21 @@ defmodule Mix.Tasks.Phx.Gen.Auth do
     {context, schema} = Gen.Context.build(context_args, __MODULE__)
     Gen.Context.prompt_for_code_injection(context)
 
-    # Needed so we can get the ecto adapter and ensure other
-    # libraries are loaded.
-    #
-    # As far as I can tell, everything after this must be tested with an
-    # integration test.
-    Mix.Task.run("compile")
+    if Keyword.get(test_opts, :validate_dependencies?, true) do
```
josevalim, line 102: Maybe we can support a "--no-compile" regular flag?

josevalim: Actually, ignore me, we need test_opts anyway. :)

josevalim (approved):
This looks good to me! Just please double check if there is anything you can remove from the integration tests now that we have this, as we should definitely prefer unit test ones. :)

aaronrenner:
> Btw, one option is to have an env var that makes you print the captured result as you capture them (or afterwards), but unit tests are also welcome.

I thought about that as I was posting this PR. :sweat_smile: Seems like a useful option I should add in the future.

> Just please double check if there is anything you can remove from the integration tests now that we have this, as we should definitely prefer unit test ones. :)

That's a good idea and something I need to figure out the best way to do. I tried adding `Mix.Tasks.Format.run(~w(--check-formatted))` into one of the phx.gen.auth tests and it errored because I didn't have the dependencies installed.

```
     ** (Mix.Error) Unavailable dependency :ecto_sql given to :import_deps in the formatter configuration. The dependency cannot be found in the file system, please run "mix deps.get" and try again
```

I was able to install the dependencies with `Mix.Tasks.Deps.Get.run([])` and the formatter passed, but I realize since we're out of the integration test project, we don't currently have the ability to control the versions of the dependencies that are being installed. This may require some more thinking to figure out.

Also, I tried adding `Mix.Tasks.Compile.run(~w(--warnings-as-errors))` and got the following error
```
     ** (Mix.Error) Could not compile dependency :ranch, "/home/aaron/.mix/rebar3 bare compile --paths /home/aaron/projects/phoenix/tmp/ttmBOgMq4y/test works with apps generated with --live/my_app/_build/test/lib/*/ebin" command failed. You can recompile this dependency with "mix deps.compile ranch", update it with "mix deps.update ranch" or clean it with "mix deps.clean ranch"
```
I tried to see if I could run `Mix.Tasks.Local.Rebar.run(~w(--force))`, but that didn't fix it.

Another thing I'm weighing is the integration tests can be run concurrently, whereas the unit tests cannot (changing directories, environment config and maybe `Mix.shell(Mix.Shell.Process)`?).  I'm wondering if we figured out a way to move the formatting, compilation and test suite  checks into the unit tests, would it actually provide an overall speed improvement because we're not running them concurrently anymore? :thinking: 

Any suggestions on a better way to check formatting, code compilation and ensure the generated test suite passes would definitely be appreciated! :slightly_smiling_face:

josevalim:
I think those are all hard to check within unit tests because they assume a mix project, that’s why we started the integration suite. Only the formatting should be relatively easy to check, but then I would manually call Code.format_string and compare the results before and after. :)

aaronrenner:
> Only the formatting should be relatively easy to check, but then I would manually call Code.format_string and compare the results before and after.

I like that idea! I always have trouble with `mix format --check-formatted` because I wonder what the suggested changes are and can't see them without committing what I have to git, running `mix format` and then a `git diff`. It would be awesome to do some sort of diffing in a test helper like this. 👍 If I was going to do this, it seems like I'd need to extract some of the logic to load the `.formatter.exs` files from the generated phoenix app (so it doesn't try to add parentheses on the routes and migrations) and pass that configuration along into `Code.format_string`. I'll keep this improvement in mind too. 🙂

PR #4174, opened 2021-01-25
Update example to match with generated layout

chulkilee:
The custom error pages guide has HTML example for custom error page but it's outdated so different from the generated layout a lot.

This PR is to align the HTML example with the generated layout. Please note that it still does not use Routes helper (like the existing example)


Gazler:
Thanks <3

PR #4179, opened 2021-01-25
Fix bug in Plug guide code

eksperimental:
We will get an error when loading http://localhost:4000/,

```
 ArgumentError at GET /
assign @locale not available in eex template.
```

It is the PageController, not the HelloController where the plug needs to go.


Gazler:
Thanks <3

eksperimental:
> Thanks <3

no worries

PR #4185, opened 2021-01-27
Precompile generator templates

aaronrenner:
This PR precompiles all of the `phx.gen.*` templates making the generators much faster to run. For example, the `phx.gen.auth` test suite went from 30 seconds to 0.7 seconds. This also adds multiple template sources (`LocalApp`, `PhoenixCompiled` and `Phoenix`) so we can still override templates by placing files in `priv/templates`.

However there are a couple drawbacks with the current implementation
* Compilation time goes from 3 seconds to 13 seconds.
* Any time we change a template, `Mix.Phoenix.TemplateSources.PhoenixCompiled` must recompile all of its templates

I'm trying to figure out if we can improve performance by compiling these templates in parallel. I've compiling the eex with `Task.async_stream/1`, but that doesn't help. I've also tried partitioning the templates into multiple modules (using `:erlang.phash_2` to split them into 4 modules, and then calling `Module.create/3`) in hopes that we can take advantage of Elixir's parallel compiler, but `Module.create/3` seems to create the module sequentially. I'm curious if partitioning would still be good so if one template changed, we would only have to recompile that partition (maybe even so far as 1 module per template).

There's a ton of changes here, so it's broken into 2 commits. The first one switches all of the templates over to use assigns so it's compatible with precompiled templates. The second has the new TemplateSource idea including the precompilation logic.

I'm really curious what approaches I could take to make precompilation more efficient. And if it ends up not making sense, that's cool too. Thanks!

### DIFF
--- a/lib/mix/phoenix.ex
+++ b/lib/mix/phoenix.ex
@@ -11,12 +11,12 @@ defmodule Mix.Phoenix do
   def eval_from(apps, source_file_path, binding) do
     sources = Enum.map(apps, &to_app_source(&1, source_file_path))
 
-    content =
+    source =
       Enum.find_value(sources, fn source ->
-        File.exists?(source) && File.read!(source)
+        if File.exists?(source), do: source
       end) || raise "could not find #{source_file_path} in any of the sources"
 
-    EEx.eval_string(content, binding)
+    EEx.eval_file(source, assigns: binding)
   end
 
   @doc """
@@ -38,12 +38,12 @@ defmodule Mix.Phoenix do
 
       case format do
         :text -> Mix.Generator.create_file(target, File.read!(source))
-        :eex  -> Mix.Generator.create_file(target, EEx.eval_file(source, binding))
+        :eex  -> Mix.Generator.create_file(target, EEx.eval_file(source, assigns: binding))
         :new_eex ->
           if File.exists?(target) do
             :ok
           else
-            Mix.Generator.create_file(target, EEx.eval_file(source, binding))
+            Mix.Generator.create_file(target, EEx.eval_file(source, assigns: binding))
           end
       end
     end
--- a/lib/mix/phoenix.ex
+++ b/lib/mix/phoenix.ex
@@ -2,21 +2,16 @@ defmodule Mix.Phoenix do
   # Conveniences for Phoenix tasks.
   @moduledoc false
 
+  alias Mix.Phoenix.TemplateSources
+
   @doc """
   Evals EEx files from source dir.
 
   Files are evaluated against EEx according to
   the given binding.
   """
-  def eval_from(apps, source_file_path, binding) do
-    sources = Enum.map(apps, &to_app_source(&1, source_file_path))
-
-    source =
-      Enum.find_value(sources, fn source ->
-        if File.exists?(source), do: source
-      end) || raise "could not find #{source_file_path} in any of the sources"
-
-    EEx.eval_file(source, assigns: binding)
+  def eval_from(template_sources, source_file_path, binding) do
+    render_template!(template_sources, source_file_path, binding)
   end
 
   @doc """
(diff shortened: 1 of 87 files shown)


josevalim:
This is very interesting!

I think part of the reason why compilation is slower is because all templates are defined in a single module. Perhaps it would be best to introduce a macro that embeds chosen templates directly into the task module itself. For example, phx.gen.auth would embed its own templates and so on. Would you be willing to explore these change? In other words, when rendering a precompiled template, you would give the module it is supposed to exist at (which will generally be `__MODULE__`).

aaronrenner:
That's a good idea! I'll experiment with that approach and report back how it goes.

aaronrenner:
My latest commit added a macro like you suggested, while still using `Mix.Phoenix.TemplateSources.PhoenixCompiled` to as the common place to dispatch templates to the appropriate modules (could definitely be more efficient and I left some todos).

This helps compilation time a little, but the problem is phx.gen.auth has so many templates including a couple big ones that I still get a warning

```shell-session
$ mix compile
Compiling 70 files (.ex)
Compiling lib/mix/tasks/phx.gen.auth.ex (it's taking more than 10s)
Generated phoenix app
```

It is nice that if I'm working in anything other than phx.gen.auth, I don't get the big compilation hit. Ideally, though, it would only recompile an eex file if it's changed. 

A very hacky way would be to somehow write a file containing a module per eex file to a tmp folder and add that tmp folder to the project's elixirc_paths  (I was reading up on the parallel compiler and it doesn't seem like modules are compiled in parallel, only files). Then we're still left with what would we name these modules and how would we reference them in the code? I was looking at the leex and yecc compiler modules to see if there could be something similar for eex, but still wasn't sure how that would work.

I think I'm really getting off in the weeds here. :sweat_smile: Please let me know if you have any ideas.

aaronrenner:
On second thought, I could break the slow phx.gen.auth templates into their own modules and have the rest of them in a shared module. I'll try that and report back.

josevalim:
Yes, that would be my suggestion. Break the phx gen auth templates into two or three modules and that should be good enough. There is likely a good logical way to break them apart too.

but all of this makes me think that Erlang evaluator could be optimized and there are likely low hanging fruits in there.

aaronrenner:
@josevalim I just pushed a second commit that breaks the templates into separate modules. This brings phoenix compilation time down to 9 seconds. I think there's still some more that can be done to improve compilation speed because touching a template file causes 4 modules to be recompiled, but that could probably be addressed in a future PR.

```shell-session
$ touch priv/templates/phx.gen.auth/registration_controller.ex && mix compile --profile time 
Compiling 4 files (.ex)
[profile] lib/mix/tasks/phx.gen.embedded.ex compiled in 84ms
[profile] lib/mix/tasks/phx.gen.auth/injector.ex compiled in 100ms
[profile] lib/mix/tasks/phx.gen.schema.ex compiled in 159ms
[profile] lib/mix/tasks/phx.gen.auth/template_sources/primary.ex compiled in 2164ms
[profile] Finished compilation cycle of 4 modules in 2164ms
[profile] Finished group pass check of 18 modules in 40ms
```

Let me know what you think!

josevalim:
Hi @aaronrenner! This looks really good and I think we definitely reached the best we can do here. The compilation times are reasonable to me. However, my last remaining concern is that by having the templates embedded, we are increasing the size of the .beam files (pretty much doubling it) and this will translate to higher memory usage for templates we will never run in production. This is the only aspect I don't know yet how to work around.

Something that I also want to do is to profile the evaluation. Evaluation will always be slower but if it is orders of magnitude slower, it means that are probably some low hanging fruits in there. If I can make the phx.gen.auth tests 3 times faster, then I think it is good enough for us to not merge this.

I will keep you posted.

josevalim:
Hi @aaronrenner, good news, I could make the tests run ~5x faster with a relatively small change: https://github.com/erlang/otp/pull/3022

If this direction is approved by the Erlang/OTP team, then I propose to not merge this since the next Elixir version with Erlang/OTP 24 should already be considerably faster!

aaronrenner:
@josevalim Wow, that's a great improvement! Thanks for spending some time and looking into this!

> However, my last remaining concern is that by having the templates embedded, we are increasing the size of the .beam files (pretty much doubling it) and this will translate to higher memory usage for templates we will never run in production. This is the only aspect I don't know yet how to work around.

That's a really good point and I didn't even think of this issue. One way we could solve this is to pull the generators into a separate hex package (like phoenix_view) and only load them during development.

```elixir
  {:phoenix_generators, "~> 1.6", only: :dev}
```

I'd be happy to do this work, but I also understand if it over-complicates things. WDYT?

josevalim:
@aaronrenner the issue with the generators approach is that often we need to coordinate changes and that will complicate things. So I would hold on for now. My hope is that the eval situation will be addressed, if that doesn't work, then we can discuss how to tackle the last bit. :)

josevalim:
The OTP PR was accepted, so I think we can close this. I will make sure Elixir v1.12 uses Erlang's optimization. :)

aaronrenner:
That awesome! :tada: Thanks for figuring it out!

Closing this PR.

josevalim:
Thank you for finding out the root cause, now everyone will benefit from it! :D

PR #4187, opened 2021-01-28
Update Docker file sample

Shikanime:
Many new features have been added to Docker, including the integration of a caching feature that makes it much faster with dependency caching.

Furthermore, I believe it's best to also use multistage when building with multiple language, which can then also be versioned.


cw789:
That caching feature finnaly would make docker fast for CI.
Thanks, for making me aware of that.
That peace was the missing one within docker that have caused me to stop using docker.

josevalim:

PR #4191, opened 2021-01-31
Add modern entry point for npm package

yyx990803:
Many JavaScript bundlers nowadays support native ES modules syntax. In fact, ES modules syntax is preferred it is easier to statically analyze and allows tree-shaking to reduce bundle size.

Modern native ESM dev servers like [vite](https://vitejs.dev/) or [snowpack](https://www.snowpack.dev/) can also consume the native ESM version of the Phoenix client directly without having to convert it from cjs back to ESM again! However, this is only possible if the Phoenix package specifies the entry point to the file using modern syntax.

This PR adds:
- The `module` field: de-facto standard supported by almost all ESM-supporting bundlers.
- The `exports` field: as specified by [Node.js](https://nodejs.org/api/packages.html#packages_package_entry_points). Many modern bundlers are also converging on using this field to conditionally resolve package entry points.

### DIFF
--- a/package.json
+++ b/package.json
@@ -4,6 +4,11 @@
   "description": "The official JavaScript client for the Phoenix web framework.",
   "license": "MIT",
   "main": "./priv/static/phoenix.js",
+  "module": "./assets/js/phoenix.js",
+  "exports": {
+    "import": "./assets/js/phoenix.js",
+    "require": "./priv/static/phoenix.js"
+  },
   "repository": {
     "type": "git",
     "url": "git://github.com/phoenixframework/phoenix.git"


Gazler:
Thanks <3

ohmree:
This commit doesn't seem to be included in 1.5.9, is that on purpose?  
(Maybe for the same reason the sister-commit got reverted in LiveView?)

ATM Phoenix and LiveView are broken with Vite and while the error is too cryptic for me to understand I think there's a pretty good chance it has to do with this.

And even if this isn't the cause of my error (something about an undefined `e` in `phoenix.js`), with the proposed fixes I _think_ `phoenix.js` should be debuggable in the browser, allowing me to pinpoint the error.

PR #4192, opened 2021-02-01
Switch from Mix.Config to Config in docs

almirsarajcic:


chrismccord:
We will eventually go this route, but for backwards compatibility for older Elixir versions, we are waiting a bit. Thanks!

Issue #4194, opened 2021-02-02
log: false not respected


benwilson512:
### Environment

* Elixir version (elixir -v): 1.11.1
* Phoenix version (mix deps): 1.5.7

### Expected behavior

This route should not emit logs:
```
get("/health", HealthController, :ready, log: false)
```

### Actual behavior

[info] GET /health
[info] Sent 200 in 88µs

I think there is something missing with respect to how Phoenix router options and Plug.Telemetry interact. The code here https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/logger.ex#L186 seems to think that `options` will have the router options, but this is an event emitted by `Plug.Telemetry`, not the phoenix router. Plug Telemetry only emits options set on the plug itself: https://github.com/elixir-plug/plug/blob/master/lib/plug/telemetry.ex#L72. I don't see how the router options could end up on those events.

mcrumm:
> I think there is something missing with respect to how Phoenix router options and Plug.Telemetry interact.

@benwilson512 They do not interact. :) That is to say, `Phoenix.Logger` subscribes to `:telemetry` events executed from both `Plug.Telemetry` (assuming your event prefix is `[:phoenix, :endpoint]`) and from `Phoenix.Router`, but aside from that they are separate.

This is laid out in the [`Phoenix.Logger`](https://hexdocs.pm/phoenix/Phoenix.Logger.html#module-instrumentation) docs, though a little hard to find as it's embedded in the details of each event.

Code-wise, this means:

1. `Plug.Telemetry, event_prefix: [:phoenix, :endpoint], log: false` disables logging `[:phoenix, :endpoint, *]`
2. `get("/", Controller, :index, log: false)` disables logging `[:phoenix, :router_dispatch, *]`
3. To filter Endpoint logging by path, one needs to wrap `Plug.Telemetry` and conditionally set the log level per request.

Here's some example output - 

**Default output after `mix phx.new`:**

```sh
[info] GET /ready
[debug] Processing with YULogWeb.PageController.ready/2
  Parameters: %{}
  Pipelines: [:browser]
[info] Sent 200 in 779µs
```

**Router `log: :info`**
```ex
# lib/yulog_web/router.ex
get "/ready", PageController, :ready, log: :info
```

```sh
[info] GET /ready
[info] Processing with YULogWeb.PageController.ready/2
  Parameters: %{}
  Pipelines: [:browser]
[info] Sent 200 in 176µs
```

**Router `log: false`**

[...]

PR #4196, opened 2021-02-03
Dynamic Phoenix.Logger level

mcrumm:
Related to #4194, this change would let us remove a wrapper that needs to know too much about Plug.Telemetry internals and replace it with this:

```ex
# lib/my_app_web/endpoint.ex

plug Plug.Telemetry,
  event_prefix: [:phoenix, :endpoint],
  log: &__MODULE__.log_level/1

@quiet_paths ["alive", "health", "ready", "favicon.ico"]
def log_level(%{path_info: [head | _]}) when head in @quiet_paths, do: false
def log_level(_), do: :info
```

Incidentally, deploying the wrapper reduced our production log output by a non-insignificant amount (screenshot below) :)

image

Wdyt?

### DIFF
--- a/lib/phoenix/logger.ex
+++ b/lib/phoenix/logger.ex
@@ -13,7 +13,7 @@ defmodule Phoenix.Logger do
       * Metadata: `%{conn: Plug.Conn.t, options: Keyword.t}`
       * Options: `%{log: Logger.level | false}`
       * Disable logging: In your endpoint `plug Plug.Telemetry, ..., log: Logger.level | false`
-      * Configure logging per request: `plug Plug.Telemetry, ..., log: (Plug.Conn.t -> Logger.level | false)`
+      * Configure log level dynamically: `plug Plug.Telemetry, ..., log: {Mod, Fun, Args}`
 
     * `[:phoenix, :endpoint, :stop]` - dispatched by `Plug.Telemetry` in your
       endpoint whenever the response is sent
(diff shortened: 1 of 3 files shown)


josevalim:
Looks good. My main concern with this aproach though is that anonymous functions won't survive compilation unless they are in the format of `&Foo.bar/arity`. One alternative is either go with {Mod, fun} or {Mod, fun, args}. Thoughts?

benwilson512:
Agree with @josevalim, anonymous functions at compile time are a pain (see Absinthe). Phoenix already uses MFA tuples for things like check_origin, so I think an MFA makes sense here too.

benwilson512:
Looking at the proposal in more detail, is the `def log_level(_), do: nil` to indicate "leave the log level alone"? Should `nil` be `true` here?

mcrumm:
> go with {Mod, fun} or {Mod, fun, args}. Thoughts?

@josevalim :+1: Works for me!

> nil to indicate "leave the log level alone"?

@benwilson512 yeah, intent was "use the default", which is `:info` for endpoint and `:debug` for router_dispatch (though I mistakenly set the default to `:info` in the dynamic case in this PR - will fix).

josevalim:
I would rather explicitly declare the default and not necessarily support nil.

mcrumm:
@benwilson512 I thought about what you said and I made it `true` :)

@josevalim Support for `nil` has waned, ready for another look :):)

lib/phoenix/logger.ex
```diff
@@ -85,6 +88,30 @@ defmodule Phoenix.Logger do
...
+  In some cases you may wish to set the log level dynamically
+  on a per-request basis. To do so, set the `:log` option to
+  a tuple, `{Mod, Fun, Args}`. The `Plug.Conn.t()` for the
+  request will be prepended to the provided list of arguments.
+
+  When invoked, your function should return one of:
+
+    * a valid [`Logger.level()`](`t:Logger.level()/0`)
+    * `false` to disable logging
+    * `true` to use the event's default log level
```
josevalim, line 102: I would rather not support true for now and have users be explicit about it in their code. Thoughts?

mcrumm: My thoughts are it's late and I thought you meant _we_ explicitly declare the default. :)

mcrumm: Done :)

josevalim (approved):

PR #4197, opened 2021-02-03
Add inputs key to .formatter.exs

eksperimental:
Current version of Elixir requires either :inputs or :subdirectories fields

    $ mix format
    ** (Mix) Expected one or more files/patterns to be given to mix format or for a .formatter.exs to exist with an :inputs or :subdirectories key

The values in this commits are the default ones when a newly generated project is created by latest stable version of Elixir.

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -72,6 +72,7 @@ locals_without_parens = [
 ]
 
 [
+  inputs: ["{mix,.formatter}.exs", "{config,lib,test}/**/*.{ex,exs}"],
   locals_without_parens: locals_without_parens,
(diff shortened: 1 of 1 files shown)


josevalim:
We don't have it on purpoise because we have not formatted Phoenix yet. Thanks!

eksperimental:
So why is there a .formatter.exs file?

josevalim:
So others can do `import_deps: [:phoenix]` and get some functions without parens.

eksperimental:
> So others can do `import_deps: [:phoenix]` and get some functions without parens.

I see. Thank you.

PR #4202, opened 2021-02-06
Copy edit: Guide - Controllers

eksperimental:


guides/controllers.md
```diff
@@ -505,4 +513,16 @@ defmodule HelloWeb.MyController do
 end
 ```
 
-Whenever the `with` conditions do not match, `HelloWeb.MyFallbackController` will receive the original `conn` as well as the result of the action and respond appropriately.
+Whenever the `with` conditions do not match, `HelloWeb.MyFallbackController` will receive the original `conn` as well as the result of the action and respond accordingly.
+
+
+[welcome page]: http://localhost:4000/
```
Gazler, line 519: Not sure if anyone has any strong opinions on this, but in the rest of the guides I believe all of the markdown links are inline?

Other than that, this is good to merge.

eksperimental: It adds value because we are linking to functions that previously are just being surrounded by backticks, such in the case of `text/2`, and by just surrounding it in square brackets now it links to the docs. If I wanted to add the links inline, it would be too much I think.

eksperimental: ... on the `welcome page` case, it is a link that is used repeatedly, so I created a reference link just to simplify the code.

Gazler: Alright, I'm sold! Do you mind rebasing and then I can merge this in?

eksperimental: > Alright, I'm sold! Do you mind rebasing and then I can merge this in?

done!

Gazler:
Thanks <3

PR #4203, opened 2021-02-06
Standardize use of key-value pair(s)

eksperimental:
In the same vein as 12bd337cfdea7e5a7e38d35ae9c891857c6046d7


Gazler:
Thanks <3

PR #4207, opened 2021-02-13
Correct installation instructions & simplify release process

eksperimental:
I have noticed the instructions were wrong because in some files the Phoenix version was not updated,
so we can rely on `mix archive.install` to retrieve the latest version.


josevalim:

Issue #4156, opened 2021-01-17
Mix phx.gen.live generates invalid tests when a unique constraint is present


LostKobrakai:
### Environment

* Elixir version (elixir -v):
* Phoenix version (mix deps): 1.5.7
* NodeJS version (node -v):
* NPM version (npm -v):
* Operating system:

### Expected behavior

All tests pass.

### Actual behavior

```
1) test Index saves new registry (PourWeb.RegistryLiveTest)
     test/pour_web/live/registry_live_test.exs:32
     ** (RuntimeError) LiveView did not redirect
     code: |> follow_redirect(conn, Routes.registry_index_path(conn, :index))
     stacktrace:
       test/pour_web/live/registry_live_test.exs:48: (test)
```

Generated with `mix phx.gen.live Registries Registry registries name:string:unique type:string`. This happens because the setup and the test both generate a registry with `@create_attrs`, which then triggers the unique constraint.

josevalim:
@feliperenan do you want to pick this one up? :)

josevalim:
@aaronrenner beautiful!

a8t:
I'm also getting failing tests upon running the generator, due to conflicting HTML escaping. Is it somehow related? Otherwise I can make another issue.

<img width="717" alt="image" src="https://user-images.githubusercontent.com/7440689/115327446-f8503000-a15c-11eb-8a09-19cb039f28a4.png">

crova:
I see the same things as @a8t.

feliperenan:
This issue is solved on master by this PR: https://github.com/phoenixframework/phoenix/pull/3829. Maybe close this one @josevalim? 🤔

PR #4214, opened 2021-02-23
Generate fixture functions for unique attributes

aaronrenner:
This PR extends the fixture generator to create functions for generating unique attributes like is done with the email address in `mix phx.gen.auth`. I'm thinking of this as the first step in fixing #4156.

Below is the file generated by the integration test

```elixir
defmodule DefaultApp.BlogFixtures do
  @moduledoc """
  This module defines test helpers for creating
  entities via the `DefaultApp.Blog` context.
  """

  @doc """
  Generate a unique post body.
  """
  def unique_post_body, do: "some body" <> "#{System.unique_integer([:positive])}"

  @doc """
  Generate a unique post order.
  """
  def unique_post_order, do: System.unique_integer([:positive])

  @doc """
  Generate a unique post publish_at.
  """
  def unique_post_publish_at do
    DateTime.add(
      ~U[2021-02-22 04:29:00Z],
      System.unique_integer([:positive])
    )
  end

  @doc """
  Generate a unique post publish_at_usec.
  """
  def unique_post_publish_at_usec do
    DateTime.add(
      ~U[2021-02-22 04:29:00.000000Z],
      System.unique_integer([:positive])
    )
  end

  @doc """
  Generate a unique post publish_on.
  """
  def unique_post_publish_on do
    NaiveDateTime.add(
      ~N[2021-02-22 04:29:00],
      System.unique_integer([:positive])
    )
  end

  @doc """
  Generate a unique post publish_on_usec.
  """
  def unique_post_publish_on_usec do
    NaiveDateTime.add(
      ~N[2021-02-22 04:29:00.000000],
      System.unique_integer([:positive])
    )
  end

  @doc """
  Generate a unique post publish_time.
  """
  def unique_post_publish_time do
    Time.add(
      ~T[14:00:00],
      System.unique_integer([:positive])
    )
  end

  @doc """
  Generate a unique post publish_time_usec.
  """
  def unique_post_publish_time_usec do
    Time.add(
      ~T[14:00:00.000000],
      System.unique_integer([:positive])
    )
  end

  @doc """
  Generate a unique post score.
  """
  def unique_post_score, do: System.unique_integer([:positive]) * 1.0

  @doc """
  Generate a unique post status.
  """
  def unique_post_status, do: :draft

  @doc """
  Generate a unique post tags.
  """
  def unique_post_tags, do: []

  @doc """
  Generate a unique post title.
  """
  def unique_post_title, do: "some title" <> "#{System.unique_integer([:positive])}"

  @doc """
  Generate a post.
  """
  def post_fixture(attrs \\ %{}) do
    {:ok, post} =
      attrs
      |> Enum.into(%{
        author_name: "some author_name",
        body: unique_post_body(),
        order: unique_post_order(),
        publish_at: unique_post_publish_at(),
        publish_at_usec: unique_post_publish_at_usec(),
        publish_on: unique_post_publish_on(),
        publish_on_usec: unique_post_publish_on_usec(),
        publish_time: unique_post_publish_time(),
        publish_time_usec: unique_post_publish_time_usec(),
        score: unique_post_score(),
        status: unique_post_status(),
        tags: unique_post_tags(),
        title: unique_post_title()
      })
      |> DefaultApp.Blog.create_post()

    post
  end
end
```

If a special fixture function isn't defined for a specific type, the unique function will be generated but will just return the default value. These updates also follow existing pattern of not generating attributes for references associations.

## Other notes

This also adds a `Mix.Phoenix.prepend_newline/1` function, so we don't have to keep the leading blank line at the top of this template (which my editor wants to remove automatically). It's also used by the latest `phx.gen.auth` commit that I need to port over to phoenix.

### DIFF
--- a/lib/mix/tasks/phx.gen.context.ex
+++ b/lib/mix/tasks/phx.gen.context.ex
@@ -217,8 +217,9 @@ defmodule Mix.Tasks.Phx.Gen.Context do
       Mix.shell.info(
         """
 
-        Update the following fixture function(s) in #{context.test_fixtures_file}
-        with new implementations:
+        Some of the generated database columns are unique. Please provide
+        unique implementations for the following fixture function(s) in
+        #{context.test_fixtures_file}:
 
         #{
           fixture_functions_needing_implementations
--- a/lib/mix/tasks/phx.gen.context.ex
+++ b/lib/mix/tasks/phx.gen.context.ex
@@ -197,6 +197,7 @@ defmodule Mix.Tasks.Phx.Gen.Context do
 
     paths
     |> Mix.Phoenix.eval_from("priv/templates/phx.gen.context/fixtures.ex", binding)
+    |> Mix.Phoenix.prepend_newline()
     |> inject_eex_before_final_end(test_fixtures_file, binding)
   end
 
--- a/lib/mix/tasks/phx.gen.context.ex
+++ b/lib/mix/tasks/phx.gen.context.ex
@@ -199,6 +199,49 @@ defmodule Mix.Tasks.Phx.Gen.Context do
     |> Mix.Phoenix.eval_from("priv/templates/phx.gen.context/fixtures.ex", binding)
     |> Mix.Phoenix.prepend_newline()
     |> inject_eex_before_final_end(test_fixtures_file, binding)
+
+    maybe_print_unimplemented_fixture_functions(context)
+  end
+
+  defp maybe_print_unimplemented_fixture_functions(%Context{} = context) do
+    fixture_functions_needing_implementations =
+      Enum.flat_map(
+        context.schema.fixture_unique_functions,
+        fn
+          {_field, {_function_name, function_def, true}} -> [function_def]
+          {_field, {_function_name, _function_def, false}} -> []
+        end
+      )
+
+    if Enum.any?(fixture_functions_needing_implementations) do
+      Mix.shell.info(
+        """
+
+        Update the following fixture function(s) in #{context.test_fixtures_file}
+        with new implementations:
+
+        #{
+          fixture_functions_needing_implementations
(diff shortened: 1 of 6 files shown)


josevalim:
Thanks @aaronrenner, this looks great!

However I am thinking that stating that we support all data types is going to bite us with some unforeseen corner cases. Perhaps it is better to handle integer + string types and for the other cases, we emit a warning during generation and a code that raises? Something like this:

    defp unique_user_birthday() do
      raise "implement the logic to generate a unique user birthday"
    end

WDYT?

aaronrenner:
@josevalim I really like that idea. I'll work on an update!

lib/mix/tasks/phx.gen.context.ex
```diff
@@ -197,7 +197,51 @@ defmodule Mix.Tasks.Phx.Gen.Context do
...
+          {_field, {_function_name, _function_def, false}} -> []
+        end
+      )
+
+    if Enum.any?(fixture_functions_needing_implementations) do
+      Mix.shell.info(
+        """
+
+        Update the following fixture function(s) in #{context.test_fixtures_file}
+        with new implementations:
```
josevalim, line 221:
```suggestion
        Some of the generated database columns are unique. Please provide
        unique implementations for the following fixture function(s) in
        #{context.test_fixtures_file}:
```

josevalim (approved):

lib/mix/tasks/phx.gen.context.ex
aaronrenner: I like that much better! I just pushed an update. :slightly_smiling_face: 

Before I merge, I'm also going to spend a little time seeing if I can clean up the actual generator I wrote since I'm thinking it could be a little cleaner.

Issue #4216, opened 2021-02-24
[:router_dispatch, :stop] telemetry event does not fire for conns that are halted


fivetanley:
### Environment

* Phoenix version: <= 1.5.8 (this is also present on the current master branch)

### Expected behavior

The `[:phoenix, :router_dispatch, :stop]` telemetry event should fire for requests that get halted by a plug. Some common reasons you might want to halt a `Plug.Conn`. For instance: Validating request parameters (checking the presence of authorization tokens, that certain headers are present) or rate limiting. Halting is the only way to stop the request from going through the rest of the pipeline, as mentioned in the [Plug docs](https://hexdocs.pm/plug/Plug.Conn.html#send_resp/1):

> Note that this function does not halt the connection, so if subsequent plugs try to send another response, it will error out. Use `halt/1` after this function if you want to halt the plug pipeline.

This can technically be worked around by using some of the `[:plug]` telemetry events but then you miss out on some of the metadata from the phoenix router and it's also seems inconsistent and harder to document/remember over time that you need to listen for events in multiple places.

The fix for this looks pretty easy and I would be happy to open a pull request for it if the framework maintainers agree it is a good way to go

 It looks like we could add the telemetry event to the `%Plug.Conn{halted: true}` match in the case `statement` in `Phoenix.Router`.

The events for successful/non-halted-through-pipeline conns are set up here: 

https://github.com/phoenixframework/phoenix/blob/92e15520541795796855233e3f72e4422c2db9f3/lib/phoenix/router.ex#L355-L357

We could copy/refactor to a method and then call the method in the halted case here:

https://github.com/phoenixframework/phoenix/blob/92e15520541795796855233e3f72e4422c2db9f3/lib/phoenix/router.ex#L347-L349

### Actual behavior

The telemetry event `[:phoenix, :router_dispatch, :stop]` is never fired if the `Plug.Conn` is halted by a plug. This plug can be either in the router, the endpoint, or the controller.

This example code can be used to reproduce the lack of events for `halt`'d connections:

```elixir
defmodule MyApp.Router do
  use Phoenix.Router

  pipeline :api do 
    plug :check_bearer_token
  end

  scope "/api/v1", MyApp.Web do
    pipe_through [:api]
    get "/posts", PostsController, :index
  end

  def check_bearer_token(conn, _opts) do
    with "Bearer " <> token <- Plug.Conn.get_req_header("Authorization"),
          :ok <- MyApp.Token.validate(token) do
        conn
[...]

PR #4049, opened 2020-10-29
Consistently track telemetry measurement

davydog187:
Was reading the telemetry integration as an example for use in commanded, and noticed this inconsistency. Feel free to close if its too much of a "style-change", but I think it will improve readability

PR #4217, opened 2021-02-24
emit [:phoenix, :router_dispatch, :stop] on halted Plug.Conn

fivetanley:
Ensures the `[:phoenix, :router_dispatch, :stop]` event is emitted when either a plug in the router or a controller halts a conn using `Plug.Conn.halt/2`.

Fixes [#4216][1]

[1]: https://github.com/phoenixframework/phoenix/issues/4216

### DIFF
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -346,6 +346,9 @@ defmodule Phoenix.Router do
 
     case pipeline.(conn) do
       %Plug.Conn{halted: true} = halted_conn ->
+        measurements = %{duration: System.monotonic_time() - start}
+        metadata = %{metadata | conn: halted_conn}
+        :telemetry.execute([:phoenix, :router_dispatch, :stop], measurements, metadata)
         halted_conn
       %Plug.Conn{} = piped_conn ->
         try do
--- a/test/phoenix/router/routing_test.exs
+++ b/test/phoenix/router/routing_test.exs
@@ -23,6 +23,11 @@ defmodule Phoenix.Router.RoutingTest do
     def any(conn, _params), do: text(conn, "users any")
     def raise(_conn, _params), do: raise("boom")
     def exit(_conn, _params), do: exit(:boom)
+    def halt(conn, _params) do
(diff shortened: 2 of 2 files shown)


fivetanley:
Should I port this to a pull request for the `1.5` branch as well?

mcrumm:
@fivetanley this PR is fine -- we'll take care of the backport, thanks!

lib/phoenix/router.ex
```diff
@@ -346,32 +346,35 @@ defmodule Phoenix.Router do
 
     case pipeline.(conn) do
       %Plug.Conn{halted: true} = halted_conn ->
+        emit_router_dispatch_end(:stop, conn, start, metadata)
```
mcrumm, line 349: I would rather avoid the indirection and do the following:

```suggestion
        measurements = %{duration: System.monotonic_time() - start}
        :telemetry.execute([:phoenix, :router_dispatch, :stop], measurements, metadata)
```

Aside from the tests, this should be the only change we need :) Wdyt?

mcrumm (commented):
Hi @fivetanley - this is looking good! I think we can simplify a bit - I left one tiny comment. Thanks!

lib/phoenix/router.ex
fivetanley: Sounds good. My intention wasn't indirection, rather consistency.  :)  there was already a fix for making things consistent in [#4049](https://github.com/phoenixframework/phoenix/pull/4049). There may be a better way to help aid that goal, but since there aren't a lot of things to keep consistent, I agree it's probably better to leave that for a future refactoring if these measurements become more complex.

Updated!

mcrumm: Indirection was perhaps the wrong word. What I meant was, in this particular instance, especially for the purposes of review, less is more :) Thank you for the update!

mcrumm:
I'm pretty sure this is good to go, but I'd like to get a second opinion before we merge it :)


// @josevalim

josevalim:

PR #4220, opened 2021-02-25
Fix EchoSocket example code

maartenvanvliet:
The previous code would give an error `state` would not be available.


aaronrenner (approved):
Thanks @maartenvanvliet! :fire:

PR #4226, opened 2021-03-01
Fix grammar. Remove article "the" before controllers and views

eksperimental:
It replaces the use of "the `PageController`" with "the `PageControler` controller.
Same with views.

Previously we were using "the `PageController` module" to avoid this, but this is incorrect
since the module is `HelloWeb.PageController`.

### DIFF
--- a/guides/contexts.md
+++ b/guides/contexts.md
@@ -127,7 +127,7 @@ If we follow the "Back" link, we get a list of all users, which should contain t
 
 That little `mix phx.gen.html` command packed a surprising punch. We got a lot of functionality out-of-the-box for creating, updating, and deleting users. This is far from a full-featured app, but remember, generators are first and foremost learning tools and a starting point for you to begin building real features. Code generation can't solve all your problems, but it will teach you the ins and outs of Phoenix and nudge you towards the proper mind-set when designing your application.
 
-Let's first check out the `UserController` that was generated in `lib/hello_web/controllers/user_controller.ex`:
+Let's first check out `UserController` which was generated in `lib/hello_web/controllers/user_controller.ex`:
 
 
 ```elixir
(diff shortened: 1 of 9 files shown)


guides/controllers.md
```diff
@@ -22,7 +22,7 @@ end
 
 The first line below the module definition invokes the `__using__/1` macro of the `HelloWeb` module, which imports some useful modules.
 
-The `PageController` gives us the `index` action to display the Phoenix [welcome page] associated with the default route Phoenix defines in the router.
+The `PageController` controller gives us the `index` action to display the Phoenix [welcome page] associated with the default route Phoenix defines in the router.
```
eksperimental, line 25: Alternatively it can be expressed like this:
```suggestion
`PageController` gives us the `index` action to display the Phoenix [welcome page] associated with the default route Phoenix defines in the router.
```

eksperimental (commented):
Alternatively we can get rid of the article "the", and we can have instead of 
"The `ErrorView` view"
just
"`ErrorView`"

josevalim:
I don't think the `UserController` controller style reads well, so I would prefer to avoid it. If you need to quality if, module is fine. Thank you!

eksperimental:
Initially I used module, but there is not module named like that, it is always `HelloWeb.ErrorView`.
Then I would remove the use of the "The" article

josevalim:
Remove the "The" sounds good to me.

eksperimental:
> Remove the "The" sounds good to me.

@josevalim Please review again as many changes were done.

guides/contexts.md
```diff
@@ -127,7 +127,7 @@ If we follow the "Back" link, we get a list of all users, which should contain t
 
 That little `mix phx.gen.html` command packed a surprising punch. We got a lot of functionality out-of-the-box for creating, updating, and deleting users. This is far from a full-featured app, but remember, generators are first and foremost learning tools and a starting point for you to begin building real features. Code generation can't solve all your problems, but it will teach you the ins and outs of Phoenix and nudge you towards the proper mind-set when designing your application.
 
-Let's first check out the `UserController` that was generated in `lib/hello_web/controllers/user_controller.ex`:
+Let's first check out `UserController` which was generated in `lib/hello_web/controllers/user_controller.ex`:
```
josevalim, line 130:
```suggestion
Let's first check out the `UserController` that was generated in `lib/hello_web/controllers/user_controller.ex`:
```

josevalim: Reverting to avoid conflicts with the new context guides PR.

josevalim:

PR #4227, opened 2021-03-01
Add tmp folder to gitignore template

LostKobrakai:
https://github.com/elixir-lang/elixir/commit/e935cf078444ccc4291e02c6eb64563884ec4c9a#diff-9e3f7e25c9b25140c73745f6b8b0e7f5aa8c143a5289e33f08a95dcb9b65a6b4R296-R297

### DIFF
--- a/installer/templates/phx_single/gitignore
+++ b/installer/templates/phx_single/gitignore
@@ -32,3 +32,6 @@ npm-debug.log
 # we ignore priv/static. You may want to comment
 # this depending on your deployment strategy.
 /priv/static/
+
+# Temporary files for e.g. tests
+/tmp
--- a/installer/templates/phx_single/gitignore
+++ b/installer/templates/phx_single/gitignore
@@ -33,5 +33,5 @@ npm-debug.log
 # this depending on your deployment strategy.
 /priv/static/
 
-# Temporary files for e.g. tests
+# Temporary files, for example, from tests.
 /tmp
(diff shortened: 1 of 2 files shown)


installer/templates/phx_single/gitignore
```diff
@@ -32,3 +32,6 @@ npm-debug.log
 # we ignore priv/static. You may want to comment
 # this depending on your deployment strategy.
 /priv/static/
+
+# Temporary files for e.g. tests
+/tmp
```
eksperimental, line 37:
```suggestion
/tmp/
```

installer/templates/phx_single/gitignore
```diff
@@ -32,3 +32,6 @@ npm-debug.log
 # we ignore priv/static. You may want to comment
 # this depending on your deployment strategy.
 /priv/static/
+
+# Temporary files for e.g. tests
```
eksperimental, line 36: In Elixir core, we removed the use of Latin abbreviations to simplify the reading. I think in this case we used "for example"

LostKobrakai: This is a 1:1 copy of what is in elixir core though.

eksperimental: > This is a 1:1 copy of what is in elixir core though.

Then it should be fixed

eksperimental: I think "the agreement" was to do it with all Latin expressions, but we only implemented it for "etc."
The discussion is here: https://github.com/elixir-lang/elixir/pull/9559

LostKobrakai: Seems like it was indeed fixed in elixir core, so I copied the wording from there

mcrumm:
Thanks! <3

PR #4230, opened 2021-03-02
Fix code in How-to: Custom error pages

eksperimental:
The previous code will not show the HTML version,
but there is a caveat in using the name for the template suggested,
so it had to be updated.

### DIFF
--- a/guides/howto/custom_error_pages.md
+++ b/guides/howto/custom_error_pages.md
@@ -61,7 +61,7 @@ end
 
 But we can do even better.
 
-Phoenix generates an `ErrorView` for us, but it doesn't give us a `lib/hello_web/templates/error` directory. Let's create one now. Inside our new directory, let's add a template, `404.html.eex` and give it some markup – a mixture of our application layout and a new `div` with our message to the user.
+Phoenix generates an `ErrorView` view for us, but it doesn't give us a `lib/hello_web/templates/error/` directory. Let's create one now. Inside our new directory, let's add a template, `404_page.html.eex` and give it some markup – a mixture of our application layout and a new `<div>` with our message to the user.
 
 ```html
 <!DOCTYPE html>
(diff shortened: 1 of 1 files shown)


eksperimental:
Please let me know if there is a convention for naming these error files in this particular case, or if it can be other way so we can name the template `404.html.eex`.

josevalim:
> The previous code will not show the HTML version,

How come? I am looking at the previous code and it seems to be fine?

eksperimental:
> > The previous code will not show the HTML version,
> 
> How come? I am looking at the previous code and it seems to be fine?

it does not. And actually I looked for the same thing on github and all I found was people who renamed the template to make it work.

josevalim:
I cannot reproduce it. I created a new app, set debug_errors: false, created a file at templates/error/404.html.eex and I got this:

Screenshot 2021-03-07 at 21 14 40

eksperimental:
> I cannot reproduce it. I created a new app, set debug_errors: false, created a file at templates/error/404.html.eex and I got this:
> 
> Screenshot 2021-03-07 at 21 14 40

Do you mind sharing the code? So i can install locally and see what is going on
Which phoenix version are you using.

josevalim:
I am using master. I went to the installer directory and did `mix phx.new demo --dev` and then the steps above.

eksperimental:
> I cannot reproduce it. I created a new app, set debug_errors: false, created a file at templates/error/404.html.eex and I got this:
> 
> Screenshot 2021-03-07 at 21 14 40

Why are you getting this plain text, and not the Phoenix styled HTML page?

josevalim:
Because my template is only 404.html.eex. Once I add HTML to it, it works too.

eksperimental:
> Because my template is only the text "404.html.eex". Once I add HTML to it, it works too.

ok.
What is your render/2 function?

josevalim:
I didn't touch the generated error view at all. I just added the template.

eksperimental:
> I didn't touch the generated error view at all. I just added the template.

That's the error.
That the template will not work with a render function.
It is not clear, or at least that is what I understood, that the render/2 previously defined should be removed.

josevalim:
Oh fantastic, so let's document it has to be removed. WDYT?

eksperimental:
> Oh fantastic, so let's document it has to be removed. WDYT?

surely.

Another question is. Is bad practice to hardcode the title and the message in the template or that is done intentionally and considered good practice , or is it to avoid the circular errors that it describes bellow.

Because if it bad practice, also passing the title and the message in the template is most likely what user will need, so additionally I will include the changes in this PR.

josevalim:
There are multiple approaches. For example I could call a title/1 function in the template passing the assigns to generate the title for me. So I would keep this simple.

PR #4234, opened 2021-03-04
Update phx.gen.json test to use unique fixture functions

aaronrenner:
I'm working to updating the `mix phx.gen.json` generators to remove `@create_attrs` and `@update_attrs` so we can use the new unique functions introduced in this PR: https://github.com/phoenixframework/phoenix/pull/4214. This is also another step toward an approach to solve #4156.

This is my first attempt and would love some feedback before I update the generator code.

```diff
diff --git a/test/my_app_web/controllers/post_controller_test.exs b/test/my_app_web/controllers/post_controller_test.exs
index 87c55ec..c188ff0 100644
--- a/test/my_app_web/controllers/post_controller_test.exs
+++ b/test/my_app_web/controllers/post_controller_test.exs
@@ -5,16 +5,6 @@ defmodule MyAppWeb.PostControllerTest do
 
   alias MyApp.Blog.Post
 
-  @create_attrs %{
-    body: "some body",
-    status: :active,
-    title: "some title"
-  }
-  @update_attrs %{
-    body: "some updated body",
-    status: :draft,
-    title: "some updated title"
-  }
   @invalid_attrs %{body: nil, status: nil, title: nil}
 
   setup %{conn: conn} do
@@ -30,7 +20,15 @@ defmodule MyAppWeb.PostControllerTest do
 
   describe "create post" do
     test "renders post when data is valid", %{conn: conn} do
-      conn = post(conn, Routes.post_path(conn, :create), post: @create_attrs)
+      title = unique_post_title()
+
+      attrs = %{
+        body: "some body",
+        status: :active,
+        title: title
+      }
+
+      conn = post(conn, Routes.post_path(conn, :create), post: attrs)
       assert %{"id" => id} = json_response(conn, 201)["data"]
 
       conn = get(conn, Routes.post_path(conn, :show, id))
@@ -39,7 +37,7 @@ defmodule MyAppWeb.PostControllerTest do
                "id" => ^id,
                "body" => "some body",
                "status" => "active",
-               "title" => "some title"
+               "title" => ^title
              } = json_response(conn, 200)["data"]
     end
 
@@ -53,7 +51,15 @@ defmodule MyAppWeb.PostControllerTest do
     setup [:create_post]
 
     test "renders post when data is valid", %{conn: conn, post: %Post{id: id} = post} do
-      conn = put(conn, Routes.post_path(conn, :update, post), post: @update_attrs)
+      title = unique_post_title()
+
+      attrs = %{
+        body: "some updated body",
+        status: :draft,
+        title: title
+      }
+
+      conn = put(conn, Routes.post_path(conn, :update, post), post: attrs)
       assert %{"id" => ^id} = json_response(conn, 200)["data"]
 
       conn = get(conn, Routes.post_path(conn, :show, id))
@@ -62,7 +68,7 @@ defmodule MyAppWeb.PostControllerTest do
                "id" => ^id,
                "body" => "some updated body",
                "status" => "draft",
-               "title" => "some updated title"
+               "title" => ^title
              } = json_response(conn, 200)["data"]
     end
```

If there are no unique attributes, the `title = unique_post_title()` line (and trailing blank line would be gone, leaving the test to look like this:

```elixir
    test "renders post when data is valid", %{conn: conn} do
      attrs = %{
        body: "some body",
        status: :active,
        title: "some title"
      }

      conn = post(conn, Routes.post_path(conn, :create), post: attrs)
      assert %{"id" => id} = json_response(conn, 201)["data"]

      conn = get(conn, Routes.post_path(conn, :show, id))

      assert %{
               "id" => ^id,
               "body" => "some body",
               "status" => "active",
               "title" => "some title"
    end
```

Please let me know your thoughts!


josevalim:
Hi @aaronrenner! The diff looks good! :+1:

For completeness, there is another approach we could take here, which is to generate `valid_schema_attributes` functions into the fixtures module... but I am not sure if we want to go down this path. Thoughts?

aaronrenner:
@josevalim Thanks for the feedback!

I thought about doing the `valid_<schema>_attributes` approach and I guess the benefit we'd get is to be able to add additional fields after the generator has run without having to update the tests.

Here's what I'm thinking it would look like with this approach

```elixir
  describe "create post" do
    test "renders post when data is valid", %{conn: conn} do
      attrs =
        valid_post_attributes(
          body: "some body",
          status: :active,
          title: "some title"
        )

      conn = post(conn, Routes.post_path(conn, :create), post: attrs)
      assert %{"id" => id} = json_response(conn, 201)["data"]

      conn = get(conn, Routes.post_path(conn, :show, id))

      assert %{
               "id" => ^id,
               "body" => "some body",
               "status" => "active",
               "title" => "some title"
    end

    #...
  end

  describe "update post" do
    setup [:create_post]

    test "renders post when data is valid", %{conn: conn, post: %Post{id: id} = post} do
      title = unique_post_title()

      attrs =
        valid_post_attributes(
          body: "some updated body",
          status: :draft,
          title: title
        )

      conn = put(conn, Routes.post_path(conn, :update, post), post: attrs)
      assert %{"id" => ^id} = json_response(conn, 200)["data"]

      conn = get(conn, Routes.post_path(conn, :show, id))

      assert %{
               "id" => ^id,
               "body" => "some updated body",
               "status" => "draft",
               "title" => ^title
    end

    # ...
  end
```

I think it's important to list out the attributes that are being used to in both the create and update tests so the user knows why we're asserting the response should contain certain values. However, this makes those helpers do essentially nothing, then.

The `valid_<schema>_attributes` approach will allow developers to add additional fields without having to update the tests, but it seems like the purpose of these tests is to ensure all of the fields we want to be written to can be written to. My thought is the `valid_<schema>_attributes` approach makes sense in `mix phx.gen.auth` where we have tests that are focused on email/password combinations and don't care about the rest of the attributes it takes to create a user. However in the case of these CRUD actions, I'm still trying to come up with a scenario where having `valid_<schema>_attributes` would provide a big benefit to the developer.

Thoughts?

josevalim:
I agree with everything you said so let's go with baby steps. Let's fix this issue first and revaluate valid attributes whenever we want in the future. Thank you for the well thought out response! :heart:

chrismccord:
ping @aaronrenner where are we at on this effort?

chrismccord:
another ping for @aaronrenner :)

aaronrenner:
@chrismccord Sorry, this fell off my radar. I'll finish it up. :)

chrismccord:
@aaronrenner ping :)

aaronrenner:
> @aaronrenner ping :)

I've been super busy with the family, but I'll put this as next on my list.

PR #4235, opened 2021-03-04
Explain how to choose an application name in the Gigalixir guide

colinsmetz:
When following the Gigalixir deployment guide, I didn't know I could provide an application name because the guide does not mention it. It would have been fine if the name was editable, but it is not. I ended up losing "a lot" of time because I had to destroy my app and create a new one later.

I suggest this small update in the guide since I believe most people would prefer an easy-to-use name, rather than the long random names generated by default.


josevalim:

PR #4238, opened 2021-03-05
Fix singular form in context task message

eksperimental:


lib/phoenix/util.ex
```diff
@@ -0,0 +1,11 @@
+defmodule Phoenix.Util do
```
josevalim, line 1: We don't need a separate module for this, can you please define it direclty in phx.gen.context? Tks.

eksperimental: done

josevalim:

PR #4240, opened 2021-03-07
Helpful error message when accessing endpoint config at compile time

### LINKED ISSUE #2841: Using router helpers during compilation time throws an ets lookup error
### Environment

* Elixir version (elixir -v): 1.6.0 (compiled with OTP 20)
* Phoenix version (mix deps):  1.3.2
* Operating system: OS X

### Expected behavior
It's possible to pass a `path` as an argument to the plug

### Actual behavior
Using router helper during compilation throws an ETS lookup error

controller:

```
defmodule PathCacheWeb.PageController do
  use PathCacheWeb, :controller
  plug PathCacheWeb.Plug.DummyPlug, [redirect_path: page_path(PathCacheWeb.Endpoint, :index)]
```

backtrace: 
```
= Compilation error in file lib/path_cache_web/controllers/page_controller.ex ==
    (stdlib) :ets.lookup(PathCacheWeb.Endpoint, :__phoenix_path__)
    (phoenix) lib/phoenix/config.ex:45: Phoenix.Config.cache/3
    lib/phoenix/endpoint.ex:643: PathCacheWeb.Endpoint.path/1
    lib/path_cache_web/controllers/page_controller.ex:6: (module)
    (stdlib) erl_eval.erl:670: :erl_eval.do_apply/6
    (elixir) lib/kernel/parallel_compiler.ex:198: anonymous fn/4 in Kernel.ParallelCompiler.spawn_workers/6
```

I suspect that it’s because of the `:ets.lookup` tries to access non-existing table which is being created only in the runtime https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/config.ex#L102

Sample repository https://github.com/lessless/path_cache_bug

P.S.  simpler way to reproduce is just to add a module attribute

```
defmodule PathCacheWeb.PageController do
  use PathCacheWeb, :controller
  @lolo page_path(PathCacheWeb.Endpoint, :index)
```

TheFirstAvenger:
Fixes #2841

### DIFF
--- a/lib/phoenix/config.ex
+++ b/lib/phoenix/config.ex
@@ -46,7 +46,18 @@ defmodule Phoenix.Config do
   """
   @spec cache(module, term, (module -> {:cache | :nocache, term})) :: term
   def cache(module, key, fun) do
-    case :ets.lookup(module, key) do
+    try do
+      :ets.lookup(module, key)
+    rescue
+      e ->
+        case :ets.info(module) do
+          :undefined ->
+            raise "could not find ets table for endpoint #{inspect(module)}. Make sure your endpoint is started and note you cannot access endpoint functions at compile-time."
+
+          _ ->
+            reraise e, __STACKTRACE__
+        end
+    else
       [{^key, :cache, val}] ->
         val
 
--- a/test/phoenix/config_test.exs
+++ b/test/phoenix/config_test.exs
@@ -22,6 +22,12 @@ defmodule Phoenix.ConfigTest do
     assert :ets.lookup(meta.test, :custom)  == [custom: true]
   end
 
+  test "raises with warning about compile time when table not started" do
+    assert_raise RuntimeError,
+                 "could not find ets table for endpoint Fooz. Make sure your endpoint is started and note you cannot access endpoint functions at compile-time.",
+                 fn -> cache(Fooz, :foo, fn _ -> {:nocache, :bar} end) end
+  end
+
   test "can change configuration", meta do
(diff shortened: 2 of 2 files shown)


lib/phoenix/config.ex
```diff
@@ -46,18 +46,29 @@ defmodule Phoenix.Config do
...
+    rescue
+      e ->
+        case :ets.info(module) do
+          :undefined ->
+            raise "could not find ets table for endpoint #{inspect(module)}. Make sure your endpoint is started and note you cannot access endpoint functions at compile-time."
+
+          _ ->
+            reraise e, __STACKTRACE__
         end
     end
```
josevalim, line 73: We want to keep as little code as possible inside the try. WDYT about this:

```suggestion
    try do
      :ets.lookup(module, key)
    rescue
      e ->
        case :ets.info(module) do
          :undefined ->
            raise "could not find ets table for endpoint #{inspect(module)}. Make sure your endpoint is started and note you cannot access endpoint functions at compile-time."

          _ ->
            reraise e, __STACKTRACE__
        end
    else
      [{^key, :cache, val}] ->
        val

      [] ->
        case fun.(module) do
          {:cache, val} ->
            :ets.insert(module, {key, :cache, val})
            val

          {:nocache, val} ->
            val
        end
    end
```

PS: I have not tested it.

TheFirstAvenger:
@josevalim That worked. Also TIL try/else :).

josevalim:

PR #4241, opened 2021-03-07
Add compiler flag to nowarn on live_dashboard_path to generator

### LINKED ISSUE #4080: live_dashboard_path/2 is undefined or private warning in prod version
### Environment

### Expected behavior

The generated project shouldn't emit warnings.

### Actual behavior

When compiling in prod env, the following warning is emitted:

```
live_dashboard_path/2 is undefined or private
```

in app.html.eex

If I understand correctly this happens b/c in router a live_dashboard route is not defined in prod:

```elixir
# router.ex

if Mix.env() in [:dev, :test] do
  import Phoenix.LiveDashboard.Router

  scope "/" do
    pipe_through(:browser)
    live_dashboard("/dashboard", metrics: MySystemWeb.Telemetry)
  end
end
```

Which breaks the following code from the generated app.html.eex

```elixir
# app.html.eex

<%= if function_exported?(Routes, :live_dashboard_path, 2) do %>
  <li><%= link "LiveDashboard", to: Routes.live_dashboard_path(@conn, :home) %></li>
<% end %>
```

I'm not exactly sure what fix to recommend though. Checking at compile-time if the function is exported would be my first instinct, but I'm not sure if this would introduce unwanted compile-time deps. Another option is to copy-paste `Mix.env()` check, but that introduces a non-obvious coupling (if a dev wants to change the condition they have to do it in two different and mutually remote places, which is not obvious at all).

Either way I think this should be addressed, because the generator creates the code which might break on CI (that happened to us), and it's not clear what would be the proper way to fix this.

TheFirstAvenger:
Fixes #4080.

### DIFF
--- a/installer/templates/phx_web/views/layout_view.ex
+++ b/installer/templates/phx_web/views/layout_view.ex
@@ -1,4 +1,7 @@
 defmodule <%= @web_namespace %>.LayoutView do
   use <%= @web_namespace %>, :view
+
+  # Phoenix LiveDashboard is available only in development by default,
+  # so we instruct Elixir to not warn if the dashboard route is missing.
   @compile {:no_warn_undefined, {Routes, :live_dashboard_path, 2}}
 end
--- a/installer/templates/phx_web/views/layout_view.ex
+++ b/installer/templates/phx_web/views/layout_view.ex
@@ -1,3 +1,4 @@
 defmodule <%= @web_namespace %>.LayoutView do
   use <%= @web_namespace %>, :view
+  @compile {:no_warn_undefined, {Routes, :live_dashboard_path, 2}}
 end


josevalim:
This will add a compile-time dependency to the router. I will see if I can resolve this on Elixir side and, if so, merge it. Thank you!

installer/templates/phx_web/views/layout_view.ex
```diff
@@ -1,3 +1,4 @@
 defmodule <%= @web_namespace %>.LayoutView do
   use <%= @web_namespace %>, :view
+  @compile {:no_warn_undefined, {Routes, :live_dashboard_path, 2}}
```
josevalim, line 3:
```suggestion

  # Phoenix LiveDashboard is available only in development by default,
  # so we instruct Elixir to not warn if the dashboard route is missing.
  @compile {:no_warn_undefined, {Routes, :live_dashboard_path, 2}}
```

josevalim:
I have removed the compile-time dependency but I am still unsure. That's because users may remove the dashboard (or make it available in prod) and leave this annotation there forever. Do we care about this? Maybe we should move this to the home template instead of layout, which is more likely for users to change/remove, effectively getting rid of the warning?

In any case, if we want to keep things as is, this is definitely the best way to go.

josevalim (approved):

chrismccord:
❤️❤️❤️🐥🔥

PR #4243, opened 2021-03-08
Use so_reuseport to fix eaddrinuse test failures

aaronrenner:
This PR is an attempt to fix intermittent test `:eaddrinuse` test failures like these:

https://github.com/phoenixframework/phoenix/runs/2003562379#step:4:509

I first came across this approach with this PR to bypass: https://github.com/PSPDFKit-labs/bypass/pull/77. It appears that it takes a little bit of time for a socket to shut down after a connection is closed. When open up a connection to find an unused port, then close it so we can reopen it for the phoenix endpoint, if the OS hasn't finished closing down the socket we'll receive this error. By opening a port both times with `so_reuseport`, even if the OS isn't finished shutting down the original socket, we shouldn't receive an error.

Here are a few links I found useful about `so_reuseport`
* An in-depth stackoverflow answer about the difference between `SO_REUSEADDR` and `SO_REUSEPORT`: https://stackoverflow.com/a/14388707
* An example of how to configure `so_reuseport` in phoenix: https://stackoverflow.com/a/56952910

### DIFF
--- a/test/support/endpoint_helper.exs
+++ b/test/support/endpoint_helper.exs
@@ -29,7 +29,15 @@ defmodule Phoenix.Integration.EndpointHelper do
   Socket option to allow a port to be reused.
 
   When provided as a socket option, this allows for a port to be bound to by
-  multiple processes and prevents eaddrinuse errors.
+  multiple processes and prevents `:eaddrinuse` errors. This is useful in
+  Phoenix's integration tests because in order to get an unused port number from
+  the OS in `get_unused_port_numbers/1` we have to open and close a socket,
+  which temporarily makes the port number unavailable to other processes while
+  the OS is shutting it down (the shutdown process continues even after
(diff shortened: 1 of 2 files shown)


test/support/endpoint_helper.exs
```diff
@@ -24,4 +24,18 @@ defmodule Phoenix.Integration.EndpointHelper do
     :gen_tcp.close(socket)
     port_number
   end
+
+  @doc """
+  Socket option to allow a port to be reused.
+
+  When provided as a socket option, this allows for a port to be bound to by
+  multiple processes and prevents eaddrinuse errors.
```
josevalim, line 32: We should probably add a line to why we need this in Phoenix tests. :)

josevalim (approved):

aaronrenner:
Looks like this wasn't the fix for it. Tests are still failing: https://github.com/phoenixframework/phoenix/runs/2063131235#step:4:479

PR #4245, opened 2021-03-08
Fix code instructions in How-to: Custom Error Pages

eksperimental:
It tackles de problem described in https://github.com/phoenixframework/phoenix/pull/4230

Closes #4230

### DIFF
--- a/guides/howto/custom_error_pages.md
+++ b/guides/howto/custom_error_pages.md
@@ -96,7 +96,7 @@ Phoenix generates an `ErrorView` for us, but it doesn't give us a `lib/hello_web
 </html>
 ```
 
-Additionally, lets remove the `render/2` function we previously introduced in `lib/hello_web/views/error_view.ex`.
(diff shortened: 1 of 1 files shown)


guides/howto/custom_error_pages.md
```diff
@@ -96,6 +96,14 @@ Phoenix generates an `ErrorView` for us, but it doesn't give us a `lib/hello_web
 </html>
 ```
 
+Additionally, lets remove the `render/2` function we previously introduced in `lib/hello_web/views/error_view.ex`.
```
josevalim, line 99:
```suggestion
After you define the template disk, remember to remove the equivalent `render/2` clause for that template, as otherwise the function overrides the template. Let's do so for the 404.html clause we have previously introduced in `lib/hello_web/views/error_view.ex`:
```

eksperimental: template disk?
"template file" maybe?

josevalim:

eksperimental:
Awesome. It will save people a lot of time when going through the guides!

PR #4250, opened 2021-03-10
tolerate `mix fromat` typo

tank-bohr:
Most common typo. A small improvement can make a developer's life just a little better while saving a bit of time.


chrismccord:
This is great for your projects if it helps you out, but not something we should introduce for all projects :) Thanks!

PR #4251, opened 2021-03-10
Introduce Git revision and use it in docs and --version

eksperimental:
When generating docs in developments mode, the docs in ExDocs will link to a
branch that does not exist (1.6.0-dev) when trying to access the source code.

Additionally, to mimic `elixir --version`, now
`mix phx.new --version` while in development mode will include the revision number,
to aid in bug reporting.

    $ mix phx.new --version
    Phoenix v1.6.0-dev (d20fa47)


installer/lib/mix/tasks/phx.new.ex
```diff
@@ -112,7 +113,16 @@ defmodule Mix.Tasks.Phx.New do
 
   @impl true
   def run([version]) when version in ~w(-v --version) do
-    Mix.shell().info("Phoenix v#{@version}")
+    version_and_revision =
+      with revision when revision != "" <- @revision,
```
chrismccord, line 117: I don't believe we can yet support `with` based on our Elixir version support. Can you swap it?

eksperimental: > I don't believe we can yet support `with` based on our Elixir version support. Can you swap it?

`with` was introduced in v1.2.0
https://hexdocs.pm/elixir/1.2.0/Kernel.SpecialForms.html#with/1

PR #4253, opened 2021-03-10
Streamline mix.exs files

eksperimental:
When generating docs in developments mode, the docs in ExDocs will link to a
branch that does not exist (i.e. 1.6.0-dev) when trying to access the source code.

Additionally I tried to keep both mix.exs and installer/mix.exs structured as similar as possible.

Updates ExDoc dependency in main mix.exs.

Related PR: https://github.com/phoenixframework/phoenix/pull/4251


josevalim:
I am :-1:. It is adding complexity to the mix.exs file and one should expect that links in a dev version won't work because it hasn't been published. In any case, thanks for the PR!

eksperimental:
Can we please reopen it? I can remove the "link to master" part of the PR, but I think streamlining both mix.exs files is a good thing.

josevalim:
Sure, without the source_ref() stuff it looks good to me. We also want to bump to ExDoc v0.24. :)

eksperimental:
> Sure, without the source_ref() stuff it looks good to me. We also want to bump to ExDoc v0.24. :)

Done

josevalim:

PR #4254, opened 2021-03-10
Improve RELEASE instructions

eksperimental:
It mentions specifically that `mix test` needs to be run for Phoenix and the Phoenix installer.


Gazler:
Thanks <3

PR #4257, opened 2021-03-15
Update deployment guide

### LINKED ISSUE #4188: Deployment guide shouldn't use nobody user in the Dockerfile
The [deployment guide](https://github.com/phoenixframework/phoenix/blob/b973280f33f3403efbab14fa012b53596826d5a1/guides/deployment/releases.md) has a dockerfile example that suggests the use of the `nobody` user, but this user is not recommend to run any programs, it's reserved for use only with NFS, as per the  [Ubuntu wiki](https://wiki.ubuntu.com/nobody):

> User nobody on a Unix system is traditionally user id 65534. This user is used by NFS servers when they cannot trust the client-supplied uids and gids, or when the root-squash option is being used.
>
>See https://www.kernel.org/doc/ols/2006/ols2006v2-pages-59-72.pdf and https://www.oreilly.com/openbook/linag2/book/ch14.html for more information on user nobody.
>
>Some misguided programs or guides suggest that this user should be used for untrusted program execution or handling untrusted data. This is bad advice. Services should have their own, dedicated, user account. Even on sites where NFS is not being used, processes run as user nobody or files owned by user nobody may grant far more privileges than expected, especially if two services have been misconfigured in this fashion.
>
> **Do not use the user nobody for anything. It is for NFS. **

Using `nobody` as the user can even pose a security risk as explained in the above quote from the Ubuntu wiki. So instead a new and dedicated user should be created.

### Environment

N/A

### Expected behavior

```dockerfile
FROM hexpm/elixir:1.11.2-erlang-23.1.2-alpine-3.12.1 as build

# install build dependencies
RUN apk add --no-cache build-base npm git python3

# prepare build dir
WORKDIR /app

# install hex + rebar
RUN mix local.hex --force && \
    mix local.rebar --force

# set build ENV
ENV MIX_ENV=dev

# install mix dependencies
COPY mix.exs mix.lock ./
RUN mix deps.get --only $MIX_ENV
RUN mkdir config
# Dependencies sometimes use compile-time configuration. Copying
# these compile-time config files before we compile dependencies
# ensures that any relevant config changes will trigger the dependencies
# to be re-compiled.
COPY config/config.exs config/$MIX_ENV.exs config/
RUN mix deps.compile

# build assets
COPY assets/package.json assets/package-lock.json ./assets/
# install all npm dependencies from scratch
RUN npm --prefix ./assets ci --progress=false --no-audit --loglevel=error

COPY priv priv

# Note: if your project uses a tool like https://purgecss.com/,
# which customizes asset compilation based on what it finds in
# your Elixir templates, you will need to move the asset compilation step
# down so that `lib` is available.
COPY assets assets
# use webpack to compile npm dependencies - https://www.npmjs.com/package/webpack-deploy
RUN mix phx.digest

# compile and build the release
COPY lib lib
RUN mix compile
# changes to config/runtime.exs don't require recompiling the code
COPY config/runtime.exs config/
# uncomment COPY if rel/ exists
# COPY rel rel
RUN mix release

# Start a new build stage so that the final image will only contain
# the compiled release and other runtime necessities
FROM alpine:3.12.1 AS app
RUN apk add --no-cache openssl ncurses-libs

ENV USER="phoenix"
ENV HOME=/home/"${USER}"
ENV APP_DIR="${HOME}/app"

# Creates an unprivileged user to be used exclusively to run the Phoenix app
RUN \
  addgroup \
   -g 1000 \
   -S "${USER}" && \
  adduser \
   -s /bin/sh \
   -u 1000 \
   -G "${USER}" \
   -h "${HOME}" \
   -D "${USER}" && \

  su "${USER}" sh -c "mkdir ${APP_DIR}"

# Everything from this line onwards will run in the context of the unprivileged user.
USER "${USER}"

WORKDIR "${APP_DIR}"

COPY --from=build --chown="${USER}":"${USER}" /app/_build/prod/rel/my_app ./

ENTRYPOINT ["bin/my_app"]

# Usage:
#  * build: sudo docker build -t phoenix/my_app .
#  * shell: sudo docker run --rm -it --entrypoint "" -p 80:4000 -p 443:4040 phoenix/my_app sh
#  * run:   sudo docker run --rm -it -p 80:4000 -p 443:4040 --name my_app phoenix/my_app
#  * exec:  sudo docker exec -it my_app sh
#  * logs:  sudo docker logs --follow --tail 100 my_app
#
# Extract the production release to your host machine with:
#
# ```
# mkdir archive
# sudo docker run --rm -it --entrypoint "" -v "$PWD/archive:/home/phoenix/archive"  phoenix/my_app sh -c "tar zcf /home/phoenix/archive/app.tar.gz ."
# ls -al archive
# ````
CMD ["start"]
```

Build:

```
sudo docker build -t phoenix/my_app .
```

Check user:

```
$ sudo docker run --rm -it --entrypoint "" phoenix/my_app sh -c "id -u"                                                                     1 ↵
1000

$ sudo docker run --rm -it --entrypoint "" phoenix/my_app sh -c "id -un"
phoenix
```

Now the app will run with an unprivileged user that is not used anywhere else in the system as the `nobody` user his.

### Actual behavior

```
$ sudo docker run --rm -it --entrypoint "" phoenix/my_app sh -c "id -u" 
65534

$ sudo docker run --rm -it --entrypoint "" phoenix/my_app sh -c "id -un"
nobody
```

While `nobody` is still a unprivileged user, its a user that may be used by other parts of the system.

J0:
Fix #4188 

As mentioned in #4188, the deployment guide needs to be updated so that the user `nobody` is no longer used because the user `nobody` is reserved for NFS. The credit behind the discovery of this bug should go to @Exadra37 . I just decided to file a pull request given that the issue was created in January and a pull request was encouraged but there didn't seem to be a PR filed.

Will be leaving as draft for now as I have yet to test the updated guide(will do so in an hour after a meeting).

Joel


nickjj:
Here's a couple of informal suggestions:

1. You can use `Dockerfile` instead of `docker` to get syntax highlighting support in Markdown.

2. Might want to consider using the `elixir` user instead of `phoenix` because some official Docker images (such as the node one) will use the language run-time's name as a default user. This isn't a standard but it's something to think about.

2. Current commit has:

```Dockerfile
ENV USER="phoenix"
ENV HOME=/home/"${USER}"
ENV APP_DIR="${HOME}/app"
```

This will create 3 separate Docker layers but you can roll this up into 1 layer with:

```Dockerfile
ENV USER="phoenix" \
    HOME=/home/"${USER}" \
    APP_DIR="${HOME}/app"
```

4. You can move the `WORKDIR` up before the `RUN addgroup ...` instruction, and then consider removing the `APP_DIR` env variable and set the WORKDIR to `"/home/${USER}/app"`.

5. You have a hanging `&& \` at the end of the `RUN addgroup` instruction.

6. You have a hardcoded reference to `nobody` in `COPY --from=build --chown=nobody:nobody /app/_build/prod/rel/my_app ./`

7. You have a reference to the `phoenix` user in the usage docs which would need to change to `elixir` if you made the change from bullet number 2 in the end.

8. Might be worth also removing the `HOME` env variable all together because at the moment it feels like this is being used to dry things out in the Dockerfile. Since it's not a build `ARG` it's not something you can configure, so IMO it should be set straight to the `elixir` user (this is how other official Docker images do it btw such as postgres, redis, node, etc.).

9. Might be worth considering to install `curl` if you plan to use a Docker `HEALTHCHECK` which could be defined in a number of spots (`Dockerfile` or `docker-compose.yml`).

10. The `MIX_ENV` is very likely worth adding as a build `ARG` so you can customize it at build time to either build your image in dev or prod mode. Having it as an `ENV` var isn't enough to tweak it at build time. You can do that like this:

    ```Dockerfile
    ARG MIX_ENV="prod"
    ENV MIX_ENV="${MIX_ENV}"
    ```

    This way it defaults to prod but you can set a build arg with `MIX_ENV="dev"` in development. This also opens up some 
    interesting patterns like being able to only digest your static files when it's in prod mode. A fully working example of 
    this in the context of a Flask app can be found here: https://github.com/nickjj/docker-flask-example/blob/main/Dockerfile

    But the same exact concept would apply to Phoenix too. In the above repo's case the `docker-compose.yml` file sets the build args.

J0:
Hey @nickjj,

Thanks for taking the time to type that all out! It is really helpful.

I've read through the list of suggestions and will be taking most if not all of the suggestions.  

You seem to have extensive knowledge about Docker and I think that your expertise would be a great help in ensuring that the deployment guide is in the best state possible. Would it be okay if I circle back to you for a brief review once I'm done updating the PR?

Let me know.

nickjj:
Sure, just keep in mind I'm not a member of Phoenix so whatever Jose or Chris says goes but I'm happy to offer general feedback.

josevalim:
Hi @J0, thanks for the PR! and thanks for the review @nickjj!

Unfortunately this PR has many changes, some seem to be unrelated to the `nobody` issue. Can you please rebase and make sure the changes address only `nobody`? If you want to change other things, separate PRs are welcome!

J0:
Hey @nickjj,

I've made the changes that you've suggested and would like to request for a review. I've yet to insert Health checks as I'm not sure how to best go about doing so -- would love any suggestions.

Let me know what your thoughts are about the existing changes and if there are any other suggestions you have. Once again, thanks so much for taking the time to take a look at the PR!

Hey @josevalim,

Apologies about the unrelated changes and thanks for the catch. I've rebased and removed the stray changes introduced by my linter. Do let me know if there's anything that's still out of place that should be removed.

Thanks!

guides/deployment/releases.md
```diff
@@ -150,8 +150,11 @@ Elixir releases work well with container technologies, such as Docker. The idea
 
 Here is an example Docker file to run at the root of your application covering all of the steps above:
 
-```docker
-FROM hexpm/elixir:1.11.2-erlang-23.1.2-alpine-3.12.1 as deps
+```Dockerfile
+FROM hexpm/elixir:1.11.2-erlang-23.1.2-alpine-3.12.1 as build
+
+# install build dependencies
+RUN apk add --no-cache build-base npm git python3 curl
```
Exadra37, line 157: We can also use the latest update packages while building the release, just like I have recommend for the final release stage:

```
apk upgrade --no-cache
```

Exadra37 (changes requested):

guides/deployment/releases.md
```diff
@@ -222,20 +206,46 @@ RUN mix release
...
+RUN \
+  addgroup \
+   -g 1000 \
+   -S "${USER}" \
+  && adduser \
+   -s /bin/sh \
+   -u 1000 \
+   -G "${USER}" \
+   -h /home/elixir \
+   -D "${USER}" \
```
Exadra37, line 224: Please also use the variable `${USER}` here.

Exadra37 (changes requested):

guides/deployment/releases.md
```diff
@@ -222,20 +206,46 @@ RUN mix release
...
 
-RUN chown nobody:nobody /app
+WORKDIR "/home/${USER}/app"
+# Creates an unprivileged user to be used exclusively to run the Phoenix app
+RUN \
+  addgroup \
+   -g 1000 \
+   -S "${USER}" \
+  && adduser \
+   -s /bin/sh \
```
rktjmp, line 220: Should this be `-s /sbin/nologin`?

J0:
Hey @rkyjmp and @Exadra37  

Thanks for the pointers! I will be taking the suggestions you've provided and I just wanted to let you know that I'll be amending the PR shortly(hopefully by end of tomorrow).

achempion:
I think this approach complicates the image without any additional benefits.

The issue we are trying to resolve is two docker containers with the same user and access to the filesystem could read/modify each others files.

With the proposed approach we'll still have the same issue as user ids in containers build with such practice would be identical.

Basically, result the same but we now we would have more complex configuration in docs and the project still has to now about security caveats of the new approach.

About the NFS thing, when we attach volume to the docker container it could be NFS as well but anyway, it's really uncommon to have multiple services running in one docker container.

Exadra37:
> The issue we are trying to resolve is two docker containers with the same user and access to the filesystem could read/modify each others files.

Exadra37:
The issue we are trying to solve is about security. 

Please read the original issue:

https://github.com/phoenixframework/phoenix/issues/4188

achempion:
> Even on sites where NFS is not being used, processes run as user nobody or files owned by user nobody may grant far more privileges than expected, especially if two services have been misconfigured in this fashion.

@Exadra37 It does not state any security concerns though, only hints that could be issues if we have more than one service running from that user without describing those issues. I would appreciate if you point out to concrete security threats with this model.

We're preparing an image for docker container and wouldn't have many processes running and I'm trying to understand how 100 containers running in a cluster with user "elixir" would be more secure than 100 containers running with user "nobody".

guides/deployment/releases.md
achempion: why do we need `git` dependency here?

achempion: the same question for curl, you're not using it anywhere

josevalim: It is common to have `git` dependencies in your project mix.exs file. It is not needed in the final build though.

achempion:
I did a bit of research of `nobody` user and right now we only have links only to ubuntu wiki here even though we don't use ubuntu here.

If we take a look at the [Debian docs](https://wiki.debian.org/SystemGroups)

> Daemons that don't need to own any files can run as nobody.nogroup instead, and more complex or security conscious daemons run as dedicated users.

> nogroup (user: nobody): Daemons that need not own any files run as user nobody and group nogroup. Thus, no files on a system should be owned by this user or group.

guides/deployment/releases.md
achempion: Ok, we need `git` in case if we have dependencies in our `mix.exs` which we pull from git instead of hex.pm.

Just got used to include only what project really needs in docker image, I assume we're trying to provide more general image here with many things "just in case" instead of minimal base to grow for each individual project.

nickjj: @achempion `curl` was likely added based on a comment I wrote earlier in the thread:

> Might be worth considering to install curl if you plan to use a Docker HEALTHCHECK which could be defined in a number of spots (Dockerfile or docker-compose.yml).

It's really common to instruct Docker to curl a URL (`http://localhost:8000` or whatever) inside of your container to determine if your service is up or not.

achempion: @nickjj the final image does now have `curl` with current Dockerfile structure, it's included only on the first step and discard later because multi-stage builds do not preserve anything between stages

achempion: Anyway, even if a result image had curl installed, not sure how common it is, it seems that [kubernetes](https://kubernetes.io/docs/tasks/configure-pod-container/configure-liveness-readiness-startup-probes/) does not depend of curl presence inside the image.

nickjj: Yeah, if you planned to use Docker's health check feature you'd want it in the final image because Docker will run the health check command inside of the container and curl would need to be available there.

The health check ultimately becomes a command that Docker runs on a set interval (you can define the params of this) and then it'll determine if your app is healthy based on the exit code of the command you're running where 0 is healthy and 1 is unhealthy.

achempion: > Yeah, if you planned to use Docker's health check feature you'd want it in the final image because Docker will run the health check command inside of the container and curl would need to be available there.

Right, but the final image here still does not have curl installed, it's present only on the first stage of a build where we download/compile dependencies. It's unclear for me why do we need curl on the first stage.

nickjj: > It's unclear for me why do we need curl on the first stage.

Ah ok, that could go either way. Technically it's not needed, but it might be worth adding so you can test your health check in development where you would be using the build stage instead of a release stage.

But truthfully in practice I configure all of my health check commands with env variables and then use `/bin/true` in development to avoid Docker hitting my server every 30 seconds in development which can get a little noisy in the logs if you're trying to debug.

achempion: @nickjj Please, re-read my comments above, the final image built with current configuration **will not have curl installed**, i.e. **you could not use healthcheck feature** if you build your image with suggested dockerfile.

nickjj: Yeah we went over the final image question in a previous reply:

> Yeah, if you planned to use Docker's health check feature you'd want it in the final image because Docker will run the health check command inside of the container and curl would need to be available there.

You can do whatever you'd like to do in the end. Personally I install curl in both based on my recent replies here.

As for Kubernetes, you're right it doesn't need any special tools but if you plan to use Docker Compose or Swarm you do, unless you decide to include your own separate language specific script which does an HTTP request to a URL and then instruct your health check to call that instead of using curl.

rktjmp: I think if there is no HEATHCHECK set in the df, you shouldn't include provisions for it since who ever adds it will know what they want to use and may not use curl.

I would also personally drop git (and python3?, not sure what's depending on that, I haven't needed it in my df's), if someone is using specific git repo/branches in their mix file it becomes pretty clear why the build is failing and they're probably advanced enough to fix it. 

It is in the build stage though so really it comes down to how fastidious you are about having minimal stages.

nickjj: > I think if there is no HEATHCHECK set in the df

Yeah this is true, although you can use a health check without defining `HEALTHCHECK` in your `Dockerfile`. It's much more flexible to define it at runtime since it lets you change it without needing to rebuild your image, and certain container runtimes like Kubernetes ignore it.

Docker Compose / Swarm have support for it in `docker-compose.yml` which configures it at runtime or if you happen to be running `docker` directly it's a flag you can set.

Maybe it's worth calling that out in the docs (as documentation) since having a health check in the end is a really good idea no matter which container runtime strategy you use.

rktjmp: I don't think it's a bad idea to include a heathcheck in the dockerfile, just like, include it and curl _or_ include neither. I would even comment on why things are included:

```dockerfile
# install build dependencies
#  - git: for mix git deps
# ...

# install runtime dependencies
#   openssl for
#   ncur...
#   curl for health check 
# note: releases include a built in erlang/elixir don't need to install ...
```

Which does bloat the line count a bit but it helps make the dockerfile a bit less of a black box.

> Maybe it's worth calling that out in the docs (as documentation) since having a health check in the end is a really good idea no matter which container runtime strategy you use.

I think this is a good idea, maybe some links to docker best practices if a reliable, stable source for that exists.

josevalim:

PR #4258, opened 2021-03-15
Mention 'npm install --prefix assets' in more instructions

henrik:
It's a little easier than cd:ing into the dir, running it and then going back.

Some guides mention an alternative on Windows, so I did the same here. (I have no insight into whether this alternative is still needed.)

Also changing some '--prefix ./assets' to '--prefix assets' for consistency.

Did not change

> Phoenix developers typically only need to run `npm install` inside the assets directory.

in guides/directory_structure.md because it seemed like too much detail for that doc.

Did not change references in error output in lib/phoenix/endpoint/watcher.ex ('You may fix this by running "npm install" inside the "assets" directory.') because it seemed more of a one-off, and there a portable but less convenient command seems better.

This commit intentionally doesn't touch it, but I also noted that "mix setup" currently calls `npm install --prefix` and so possibly may have issues on Windows.

### DIFF
--- a/installer/templates/phx_single/README.md
+++ b/installer/templates/phx_single/README.md
@@ -4,7 +4,7 @@ To start your Phoenix server:
 
   * Install dependencies with `mix deps.get`<%= if @ecto do %>
   * Create and migrate your database with `mix ecto.setup`<% end %><%= if @webpack do %>
-  * Install Node.js dependencies with `npm install --prefix assets` (on Windows, you may need to do `cd assets && npm install && cd ..` instead)<% end %>
+  * Install Node.js dependencies with `npm install` inside the `assets` directory<% end %>
   * Start Phoenix endpoint with `mix phx.server` or inside IEx with `iex -S mix phx.server`
 
 Now you can visit [`localhost:4000`](http://localhost:4000) from your browser.
--- a/installer/templates/phx_single/README.md
+++ b/installer/templates/phx_single/README.md
@@ -4,7 +4,7 @@ To start your Phoenix server:
 
   * Install dependencies with `mix deps.get`<%= if @ecto do %>
   * Create and migrate your database with `mix ecto.setup`<% end %><%= if @webpack do %>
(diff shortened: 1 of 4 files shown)


installer/templates/phx_single/README.md
```diff
@@ -4,7 +4,7 @@ To start your Phoenix server:
 
   * Install dependencies with `mix deps.get`<%= if @ecto do %>
   * Create and migrate your database with `mix ecto.setup`<% end %><%= if @webpack do %>
-  * Install Node.js dependencies with `npm install` inside the `assets` directory<% end %>
+  * Install Node.js dependencies with `npm install --prefix assets` (on Windows, you may need to do `cd assets && npm install && cd ..` instead)<% end %>
```
josevalim, line 7:
```suggestion
  * Install Node.js dependencies with `npm install` inside the `assets` directory<% end %>
```

installer/templates/phx_umbrella/apps/app_name_web/README.md
```diff
@@ -4,7 +4,7 @@ To start your Phoenix server:
 
   * Install dependencies with `mix deps.get`<%= if @ecto do %>
   * Create and migrate your database with `mix ecto.setup`<% end %><%= if @webpack do %>
-  * Install Node.js dependencies with `npm install` inside the `assets` directory<% end %>
+  * Install Node.js dependencies with `npm install --prefix assets` (on Windows, you may need to do `cd assets && npm install && cd ..` instead)<% end %>
```
josevalim, line 7:
```suggestion
  * Install Node.js dependencies with `npm install` inside the `assets` directory<% end %>
```

josevalim:
Reverted the README because if we have to list both, then I would rather list one. :)

henrik:
Thank you, José! And alright, sure :) I didn't like the idea of someone reading the README and doing `cd assets && npm install && cd ..`all the time when `npm install --prefix assets` would do, but I see your point. Small diff in characters typed anyway.

PR #4260, opened 2021-03-16
Update directory_structure.md

joeapearson:
Hi Phoenix team, here are some updates to your documentation for your review.

Changes:

* Added some full-stops where needed for consistency.
* Added some notes about the `assets` directory

The goal I had in mind was to let devs know where they should start making changes to their front-end application (and not, for example, start making changes directly in the `priv/static` folder).

### DIFF
--- a/guides/directory_structure.md
+++ b/guides/directory_structure.md
@@ -20,19 +20,19 @@ When we use `mix phx.new` to generate a new Phoenix application, it builds a top
 
 We will go over those directories one by one:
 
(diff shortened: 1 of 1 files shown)


cw789:
What if Phoenix is used without Webpack (`--no-webpack`)?

joeapearson:
> What if Phoenix is used without Webpack (--no-webpack)?

I did think about that and you're right that's why I intentionally added the word "default".  It's easy to miss though... I just added some clarification.

josevalim:

PR #4261, opened 2021-03-16
use Snowpack to get JavaScript debugging in dev

ijt:
This change makes [Snowpack](https://www.snowpack.dev/) the asset builder while developing, so you can use an in-browser JavaScript debugger on non-minified code.

Webpack is still used for `npm run deploy` so you can have minified code in production.


chrismccord:
Please see the mailing list discussion for snowpack related decisions. Thanks!

PR #4268, opened 2021-03-20
Add `--database sqlite3` option to `phx.new`

kevinlang:
# Overview & Background
SQLite3 is an extraordinary powerful embedded database that is suitable for small sites. Since it is not a client/server database, it is much easier to setup and maintain. It is a great candidate for spinning up small Phoenix projects that do not need the full complexity of Postgres. It's primary disadvantage is that it cannot scale beyond multiple machines or nodes, but it is well suited to vertical scaling.

Until recently, there has not been any suitable [Ecto3 SQLite3 adapter](https://github.com/elixir-sqlite/ecto_sqlite3). I believe at some point in the past Phoenix used to support SQLite3 as an option, but removed it when it upgraded to Ecto3 as the existing drivers only supported Ecto2.

Now that there is a fully functional Ecto3 adapter, I thought it would be good to re-integrate SQLite3 into the Phoenix generators.

# Changes
Here I add a new `--database sqlite3` functionality to `phx.new`. 

I also updated the documentation where relevant, searching for any places where e.g., `mysql` or `myxql` was referenced, and adding relevant info about the `sqlite3` related functionality.

Because SQLite in typical usage does not require or need a username, password, or host, I had to fork some of the configuration logic for the generator to support the "embedded database" case where only a database path is needed. This also meant changing some code that assumed all adapters required a `DATABASE_URL`, for example, instead having the code generated refer to a `DATABASE_PATH` instead.

I've opted to have the database files stored at the project root by default. Another location I considered is `priv/` but per the [documentation](https://hexdocs.pm/phoenix/directory_structure.html), it looks like this is intended for more static/unchanging assets. In either case, I updated the `.gitignore` to exclude the db files that will be created in the root. 

One tricky part was umbrella projects, as running e.g., `mix test` from the application root (as opposed to the umbrella root) was creating database files _there_ at the app root, as the CWD would be there. I got around this by making the location of the database files be relative to the config directory in order to ensure they are always placed at the root of the umbrella project, as you can see in the changes. 

I've also opted to use the `.db` extension instead of say, a `.sqlite3` extension that e.g., Rails uses. However, I don't feel strongly either way - in SQLite3's view a file is a file :)


josevalim:
Those changes are absolutely great! Thank you for all of the amazing work on making sqlite3 firstclass for Elixir.

We just need to rebase the PR and I can merge it, thank you! :heart:

kevinlang:
@josevalim I rebased all the changes :+1:

josevalim:

matteing:
Bless, thank you for this. 

Question: will this allow starting a Phoenix app with SQLite as the default database for test and dev? It’s been a bit of a pain point as a fellow Django convert (that framework generates database files automatically on migration or test for local stuff, so it’s easy to get an app up and running quickly).

kevinlang:
By passing `--database sqlite3` when generating the project, it will be the default database for dev/test. But it will also be the database for prod as well (i.e., it is "all or nothing").

If you are asking about having some way of using SQLite3 for dev/test but another database (e.g., Postgres) for prod, it does not do that currently. I can see the appeal of using SQLite3 for dev/test and Postgres for prod, but I also think it is fraught with issues. Ecto is lower level than ActiveRecord (and likely whatever Django uses) and so I don't think it can serve as a great universal interface between the two, as the two adapters have different quirks (since the underlying databases have different quirks).

PR #4271, opened 2021-03-23
Add 'ecto.drop' to generated 'mix test' command

ericacheong:
In development process, schema are always updated and created. The current generated 'mix test' command did not drop repo and create new repo. If the repo has been updated and migration scripts have been updated, it would not run migration scripts again. There are many times that I would forget to run migration scripts again on test db. I propose to first drop the repo and create the new one, then apply all migrations every time when the tests are run.


josevalim:
Hi @ericacheong , thanks for the PR!

Unfortunately dropping the database before very test is going to add time to the test suite, specially on applications that have existed for a while and you have more than dozens of migrations.

Your test database should not get out of sync, unless you manually change previous migration, which in this case is best to call ecto.drop directly. For this reason, I will close this. Thanks!

PR #4275, opened 2021-03-24
Update Dockerfile instructions in releases.md

darraghenright:
Hi 👋 

The package install command `apk add python` no longer seems to work. The solution appears to be to explicitly specify `python2` or `python3` instead. There is some discussion of that here: https://stackoverflow.com/a/62652692/312962


josevalim:

PR #4276, opened 2021-03-25
Add SourceMaps to minified JS

maennchen:
As discussed in https://groups.google.com/g/phoenix-core/c/Dk2AbXVnzeg

When running `npm run build`, this will:

* Add a comment to `priv/static/phoenix.js` pointing to the sourcemap
* Add a new file `priv/static/phoenix.js.map`

I have verified that when using webpack in your application, it will incorporate this sourcemap and therefore create better error stacktraces.


gcauchon:
I'm not a Webpack expert, but should we also adapt the`webpack.config.js` template to include a sourcemap in the production bundle too?

https://github.com/phoenixframework/phoenix/blob/master/installer/templates/phx_assets/webpack.config.js#L27

```js
    devtool: devMode ? 'eval-cheap-module-source-map' : 'source-map',
```

maennchen:
@gcauchon I don't think so. The reason is that this is a decision that the user of the framework should make:

* Open Source Applications: Shoud probably have it
* Closed Source Applications: Most companies do not want to expose their unminified JS. If using Sentry for example a `hidden-source-map` can be used and uploaded instead.

rktjmp:
Same: https://github.com/phoenixframework/phoenix/pull/4265

maennchen:
@rktjmp I don't think so. The linked PR is for the phoenix generator. This PR only looks at the JS packaged with phoenix.

maennchen:
Rebased to current `master`

maennchen:
Rebased to current `master`

maennchen:
Amended compiled JS & Source Maps as discussed with @bcardarella

bcardarella:
I believe if you add to this block https://github.com/phoenixframework/phoenix/blob/dcc93986fdc13120c79b0ae49771a8692df4e533/installer/lib/phx_new/single.ex#L89-L97 to ensure the new sourcemap file is being copied in for `--no-webpack` projects this PR should be ready to merge.

chrismccord:
❤️❤️❤️🐥🔥

Issue #4280, opened 2021-03-31
no function clause matching in Keyword.has_key?/2


conradwt:
### Environment

* Elixir version (elixir -v):  Elixir 1.11.4 (compiled with Erlang/OTP 23)
* Phoenix version (mix deps): phoenix 1.5.8 (Hex package) (mix)
* Node.js version (node -v): v14.16.0
* NPM version (npm -v): 6.14.11
* Operating system: macOS 11.2.3

### Expected behavior

**When clicking**

movie/show.html.eex
```elixir
<%= link "Write Review", to: Routes.movie_review_path(@conn, :new, @movie) %>
```

**invokes**

review_controller.ex
```elixir
  def new(conn, %{"movie_id" => movie_id}) do
    movie = Catalogs.get_movie!(movie_id)

    changeset =
      movie
      |> Ecto.build_assoc(:reviews, %Review{})

    render(conn, "new.html", movie: movie, changeset: changeset)
  end
```

catalogs.ex
```elixir
 def get_movie!(id), do: Repo.get!(Movie, id)
```

**renders**

review/new.html.eex
```
<%= render "form.html", Map.put(assigns, :action, movie_review_path(@conn, :new, @movie)) %>
```

**invokes**

review/form.html.eex
```elixir
<%= form_for @changeset, movie_review_path(@conn, @action, @movie), fn f -> %> # blows up here

...

<% end %>
```

I'm expecting that the `review/new.html.eex` would properly render in the browser.

### Actual behavior

The actual result is as follows within the browser:

![Screen Shot 2021-03-30 at 10 55 56 PM](https://user-images.githubusercontent.com/20219/113097017-1accda00-91ab-11eb-8dae-7de4c785ef2f.png)

Finally, I have reviewed the limited documentation regarding `Nested Resources` but the docs didn't go into any detail about the actual view and template code.  I'm not sure if I'm doing something wrong here or there's a potential bug within Phoenix here.

eksperimental:
sorry, after the change you need to run `mix deps.compile phoenix`

conradwt:
@eksperimental There's only 87 lines within my `router.ex`.  However, the code that you're referring to here is code owned by the `Phoenix` project.

eksperimental:
> @eksperimental There's only 87 lines within my `router.ex`. However, the code that you're referring to here is code owned by the `Phoenix` project.

my bad.
`deps/lib/phoenix/rounter/helpers.ex`

[...]

PR #4282, opened 2021-04-01
Fix router helper

eksperimental:
Keyword.has_key?/2 is being passed a non atom key
as described in:
https://github.com/phoenixframework/phoenix/issues/4280

Fixes the issue by checking the type of key

/cc  @conradwt


josevalim:

PR #4284, opened 2021-04-08
add option `--prefix` to schema and migration generators

Apelsinka223:
I added a new option `--prefix` to the `mix phx.gen.schema` task and to the all related tasks: `phx.new`, `phx.gen.context`, `phx.gen.auth`. So it would be possible to generate files with pre-installed `prefix` option in the `table/2` function in migrations and `@schema_prefix` module attribute in schemas.


josevalim:

PR #4287, opened 2021-04-11
Fix presence example description

dmitrykleymenov:
Order of invokes was changed, but description wasn't. This synchronizes example with it's description


josevalim:

PR #4288, opened 2021-04-12
Change vsn query in assets url to be configurable

0urobor0s:
When generating `static_path` or `static_url` as well as when running `mix phx.digest`it would be nice to have the option of the generated urls to have the `vsn=d`query or not.
This PR adds the possibility and maintains the current behavior as the default, being opt-in.
The `vsn` option as added to the runtime configuration `:url` of `Phoenix.Endpoint` as it looked like an expected place for it.
This PR is also related to #1784

### DIFF
--- a/lib/mix/tasks/phx.digest.ex
+++ b/lib/mix/tasks/phx.digest.ex
@@ -33,19 +33,26 @@ defmodule Mix.Tasks.Phx.Digest do
     * app-eb0a5b9302e8d32828d8a73f137cc8f0.js.gz
     * cache_manifest.json
 
+  ## vsn
+
+  It is possible to digest the stylesheet asset references without the query
+  string "?vsn=d" with the option `--no-vsn`.
   """
+
+  @default_opts [vsn: true]
 
   @doc false
   def run(all_args) do
     Mix.Task.run "compile", all_args
-    {opts, args, _} = OptionParser.parse(all_args, switches: [output: :string], aliases: [o: :output])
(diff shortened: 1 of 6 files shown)


lib/phoenix/endpoint/supervisor.ex
```diff
@@ -389,10 +389,11 @@ defmodule Phoenix.Endpoint.Supervisor do
 
   defp warmup_static(endpoint, %{"latest" => latest, "digests" => digests}) do
     Phoenix.Config.put_new(endpoint, :cache_static_manifest_latest, latest)
+    with_vsn? = endpoint.config(:url)[:vsn]
```
josevalim, line 392: :vsn is not really a property of url. We would need to create a specific entry for it instead of putting it inside url. :)

josevalim:
Thanks for the PR @0urobor0s but I want to point out that vsn=d is a requirement if you are using Plug.Static, as it feels Plug.Static to set long expiry dates. You can only skip it if you are using something else to server assets. So given the potential footgun, I would prefer to not merge this unless there is a strong use case for this feature. Thank you!

lib/mix/tasks/phx.digest.ex
```diff
@@ -33,19 +33,27 @@ defmodule Mix.Tasks.Phx.Digest do
     * app-eb0a5b9302e8d32828d8a73f137cc8f0.js.gz
     * cache_manifest.json
 
+  ## vsn
+
+  It is possible to digest the stylesheet asset references without the query
+  string "?vsn=d" with the option `--no-vsn`.
+
```
josevalim, line 40:
```suggestion
```

josevalim:

PR #4289, opened 2021-04-12
Enable IPv6 for Ecto

mrkurt:
This is an intentionally small PR. I'm not sure how best to accomplish this, so I made a tiny guess.

It adds a default `socket_options` to allow IPv6 with Ecto. This is backwards compatible, it does not disable IPv4. It is necessary [so Phoenix can talk to databases over IPv6](https://elixirforum.com/t/use-fly-io-internal-dns-for-resolving-database-url/38889) by default.

Tests pass but I'm not sure this change would actually break an existing test.

### DIFF
--- a/installer/lib/phx_new/generator.ex
+++ b/installer/lib/phx_new/generator.ex
@@ -264,7 +264,8 @@ defmodule Phx.New.Generator do
     config :#{binding[:app_name]}, #{binding[:app_module]}.Repo,
       # ssl: true,
       url: database_url,
-      pool_size: String.to_integer(System.get_env("POOL_SIZE") || "10")
+      pool_size: String.to_integer(System.get_env("POOL_SIZE") || "10"),
+      socket_options: [:inet6]
     """)
   end
 
--- a/installer/templates/phx_ecto/repo.ex
+++ b/installer/templates/phx_ecto/repo.ex
@@ -1,6 +1,6 @@
 defmodule <%= @app_module %>.Repo do
   use Ecto.Repo,
     otp_app: :<%= @app_name %>,
-    adapter: <%= inspect @adapter_module %>
+    adapter: <%= inspect @adapter_module %>,
     socket_options: [:inet6]
 end
--- a/installer/templates/phx_ecto/repo.ex
+++ b/installer/templates/phx_ecto/repo.ex
@@ -2,4 +2,5 @@ defmodule <%= @app_module %>.Repo do
   use Ecto.Repo,
     otp_app: :<%= @app_name %>,
     adapter: <%= inspect @adapter_module %>
+    socket_options: [:inet6]
 end
--- a/installer/templates/phx_ecto/repo.ex
+++ b/installer/templates/phx_ecto/repo.ex
@@ -1,6 +1,5 @@
 defmodule <%= @app_module %>.Repo do
   use Ecto.Repo,
     otp_app: :<%= @app_name %>,
-    adapter: <%= inspect @adapter_module %>,
-    socket_options: [:inet6]
+    adapter: <%= inspect @adapter_module %>
 end


josevalim:
@mrkurt you most likely want to change this instead: https://github.com/phoenixframework/phoenix/blob/master/installer/lib/phx_new/generator.ex#L267

It is the same change, just in another place. :)

mrkurt:
Oh good call! Easy enough.

josevalim:

chunlea:
We should be careful with this. I know there are a lot of services that support IPv6 and I also love IPv6. But there still some other services that don't support IPv6, for example, DigitalOcean Managed Database. We changed this default behaviour which may break others.

https://github.com/elixir-ecto/ecto/issues/3593#issuecomment-806195037 And looks like `inet6` not working on all systems(According to https://erlang.org/doc/man/inet.html#setopts-2 at `{ipv6_v6only, Boolean}`). At least it not working on `alpine:3.13.3`, which I tested.

So I think we should add this change to Changelog and promote a message after creating a new Phoenix project to remind developers of this change.

josevalim:
The ipv6 option supports both ipv6 and ipv4, so we should be fine. This also only affects new apps, so old apps should not see any breaking behavior.

josevalim:
We reverted this for now because of #4389.

mrkurt:
Ahhh. That's too bad, what a "fun" bug.

josevalim:
We can bring it back, but we will have to do it in the adapter directly (Postgrex).

ruslandoga:
👋 

Sorry to "necropost" but I'm working on adding automatic IPv6 support to self-hosted Plausible installations (so that people don't need to provide `ECTO_IPV6` and co env vars) and the only problematic library seems to be Postgrex as others like Mint are doing a [manual inet fallback.](https://github.com/elixir-mint/mint/pull/270) In particular, I haven't been able to use inet6 in ipv4-only networks even with `ipv6_v6only` set to false. I tried websites from http://dual.tlund.se on a "dual stack" fly.io machine, and I've also tried them from ipv4-only Docker containers on AWS EC2. It always fails when trying to connect to an IPv4-only domain. Examples:

From fly.io machine:
``` elixir
iex> :inet.getifaddrs
{:ok,
 [
   {~c"lo",
    [
      flags: [:up, :loopback, :running],
      addr: {127, 0, 0, 1},
      netmask: {255, 0, 0, 0},
      addr: {0, 0, 0, 0, 0, 0, 0, 1},
      netmask: {65535, 65535, 65535, 65535, 65535, 65535, 65535, 65535},
      hwaddr: [0, 0, 0, 0, 0, 0]
    ]},
   {~c"dummy0", [flags: [:broadcast], hwaddr: [150, 232, 97, 230, 238, 118]]},
   {~c"eth0",
    [
      flags: [:up, :broadcast, :running, :multicast],
      addr: {172, 19, 136, 2},
      netmask: {255, 255, 255, 248},
      broadaddr: {172, 19, 136, 7},
      addr: {172, 19, 136, 3},
      netmask: {255, 255, 255, 248},
      broadaddr: {172, 19, 136, 7},
      addr: {9733, 19520, 344, 51602, 0, 62017, 27426, 1},
      netmask: {65535, 65535, 65535, 65535, 65535, 65535, 65535, 65534},
      addr: {64938, 0, 24663, 2683, 129, 62017, 27426, 2},
      netmask: {65535, 65535, 65535, 65535, 65535, 65535, 65535, 0},
      addr: {65152, 0, 0, 0, 56493, 55039, 65078, 12685},
      netmask: {65535, 65535, 65535, 65535, 0, 0, 0, 0},
      hwaddr: [222, 173, 214, 54, 49, 141]
    ]},
   {~c"teql0", [flags: []]}
 ]}
 
# try connecting to ipv4-only host from http://dual.tlund.se
iex> {:ok, socket} = :gen_tcp.connect(~c"ipv4.tlund.se", 80, [active: false])
iex> :inet.peername(socket)
{:ok, {{193, 15, 228, 195}, 80}}

# try inet6
iex> :gen_tcp.connect(~c"ipv4.tlund.se", 80, [:inet6, active: false])
{:error, :nxdomain}
 
 # ipv6_v6only doesn't help
iex> :gen_tcp.connect(~c"ipv4.tlund.se", 80, [:inet6, ipv6_v6only: false, active: false])
{:error, :nxdomain}

# inet doesn't work with ipv6-only host
iex> :gen_tcp.connect(~c"ipv6.tlund.se", 80, [active: false])
{:error, :nxdomain}

# inet6 works with ipv6-only host
iex> {:ok, socket} = :gen_tcp.connect(~c"ipv6.tlund.se", 80, [:inet6, active: false])
iex> :inet.peername(socket)
{:ok, {{10752, 2049, 15, 0, 0, 0, 0, 405}, 80}}
```

From IPv4-only AWS EC2 container:
```elixir
# no fallback to inet
iex> :gen_tcp.connect(~c"ipv6.tlund.se", 80, [:inet6, active: false])
{:error, :eaddrnotavail}
```

I know that it's possible to workaround this issue by using `:endpoints` option but I wonder if it'd OK for to try and open a PR to add a manual `inet6 -> inet` fallback like in Mint?

josevalim:
We should probably improve it. I would only do a bug report on Erlang/OTP first, and make sure that this indeed is not supposed to work. I think inet6 is meant to be both ipv6 and ipv4 after all.

Issue #1218, opened 2015-09-15
Running webpack executable watcher on Windows causes errors


johanalkstal:
I created a new Phoenix project without brunch, `mix phoenix.new yzerman --no-brunch` and added `webpack` through `npm`, then added it to the watchers list,

`watchers: [{ Path.expand("node_modules/webpack/bin/webpack.js"), ["--watch", "--colors", "--progress"] }]`

Also added a `webpack.config.js` and a simple `app.js` file to see that it works.

But I get this error,

```
D:\Projects\Faceoff\yzerman> mix phoenix.server
[info] Running Yzerman.Endpoint with Cowboy on http://localhost:4000
[info] Running Yzerman.Endpoint with Cowboy on http://localhost:4000
[error] Task #PID<0.248.0> started from Yzerman.Endpoint terminating
Function: &Phoenix.Endpoint.Watcher.watch/3
    Args: ["d:/Projects/Faceoff/yzerman", "d:/Projects/Faceoff/yzerman/node_modules/webpack/bin/webpack.js", ["--watch",
 "--colors", "--progress"]]
** (exit) an exception was raised:
    ** (ErlangError) erlang error: :eacces
        :erlang.open_port({:spawn_executable, 'd:/Projects/Faceoff/yzerman/node_modules/webpack/bin/webpack.js'}, [{:cd,
 "d:/Projects/Faceoff/yzerman"}, :stderr_to_stdout, :use_stdio, :exit_status, :binary, :hide, {:args, ["--watch", "--col
ors", "--progress"]}])
        (elixir) lib/system.ex:445: System.cmd/3
        (phoenix) lib/phoenix/endpoint/watcher.ex:13: Phoenix.Endpoint.Watcher.watch/3
        (elixir) lib/task/supervised.ex:74: Task.Supervised.do_apply/2
        (stdlib) proc_lib.erl:237: :proc_lib.init_p_do_apply/3
[error] Task #PID<0.250.0> started from Yzerman.Endpoint terminating
Function: &Phoenix.Endpoint.Watcher.watch/3
    Args: ["d:/Projects/Faceoff/yzerman", "d:/Projects/Faceoff/yzerman/node_modules/webpack/bin/webpack.js", ["--watch",
 "--colors", "--progress"]]
** (exit) an exception was raised:
    ** (ErlangError) erlang error: :eacces
[...]

PR #4291, opened 2021-04-12
interface npm scripts 

c4710n:
`scripts` section of `package.json` is a good place to provide commands which are used by Phoenix.

If we abstract the long commands into short npm scripts, we will:
+ have a unified interface to call node.js related commands.
+ avoid the long commands(such as `node node_modules/webpack/bin/webpack.js --mode development`) spreading in the code

And, there are two more advantage:
+ The developers who want to implement his own `assets` dir will have a better view on what they should implement - `just provides necessary NPM scripts, no need to touch the Phoenix code`.
+ When Phoenix Team decide to upgrade to Webpack 5, just update the npm scripts, there's nothing need to do with the Elixir code.

### DIFF
--- a/guides/introduction/up_and_running.md
+++ b/guides/introduction/up_and_running.md
@@ -38,7 +38,7 @@ When it's done, it will ask us if we want it to install our dependencies for us.
 Fetch and install dependencies? [Yn] Y
 * running mix deps.get
 * running mix deps.compile
-* running cd assets && npm install && node node_modules/webpack/bin/webpack.js --mode development
+* running cd assets && npm install && npm run dev
 
 We are almost there! The following steps are missing:
 
@@ -101,7 +101,7 @@ We are almost there! The following steps are missing:
 
     $ cd hello
     $ mix deps.get
-    $ cd assets && npm install && node node_modules/webpack/bin/webpack.js --mode development
+    $ cd assets && npm install && npm run dev
 
 Then configure your database in config/dev.exs and run:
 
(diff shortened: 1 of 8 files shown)


josevalim:
This would be our preferred approach but every time we made this change people reported issues, especially on OSes like Windows. So if someone can try this out on a Windows machine (under WSL, mingw, cygwin, etc), then we can merge it, otherwise it is probably best to leave it as is. :(

c4710n:
I will try this out on a Windows machine like you said.

c4710n:
## Running modified version `phx.new`

```
$ git clone -b adjust-assets https://github.com/c4710n/phoenix.git
$ cd phoenix/installer
$ mix deps.get
$ mix compile
$ mix phx.new /tmp/demo
```

## WSL

- Sub System: Ubuntu 20.04.2 LTS
- Elixir: 1.11.2 (Following the installation at [here](https://elixir-lang.org/install.html#gnulinux))
- Node.js: 10.19.0 (installed by `sudo apt-get install nodejs`)
- npm: 6.14.4 (installed by `sudo apt-get install npm`)

WSL works like a generic Linux box, no issue found.

## web installer

- Git: 2.3.1 (installed by [installer](https://git-scm.com/download/win) provided by https://git-scm.com)
- Elixir: 1.11.2 (installed by [installer](https://github.com/elixir-lang/elixir-windows-setup/releases/download/v2.1/elixir-websetup.exe) by Elixir team)
- Node.js: 14.16.1 (installed by [installer](https://nodejs.org/en/download/) provided by Node.js team)
- npm: 6.14.2 (included in above installer)

After the installation, `elixir` `mix` `npm` can be found in the `PATH` environment variable.

But, error occured, just like #1218.

> I think that's what you said above.

After serveral hours, I think I can't fix this problem.

## mingw / cygwin

Ignore them for now.

The above problem has not been solved, and there is no point in solving this problem. ;(

## Conclusion

Currently, there's no one perfect solution. I will give up this PR.

josevalim:
Thanks for checking, it is very appreciated! <3

PR #4297, opened 2021-04-19
Disable FLoC by default when creating a new project

Awlexus:
As the title says, this pull request tries to disable [FLoC](https://web.dev/floc/) on Google Chrome by default when creating a new project. I didn't add any tests yet, because I was ensure whether defining the function to disable this feature would fit into the `Phoenix.Endpoint` module and would appreciate further feedback. Thanks in advance!

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -905,4 +905,15 @@ defmodule Phoenix.Endpoint do
   def server?(otp_app, endpoint) when is_atom(otp_app) and is_atom(endpoint) do
     Phoenix.Endpoint.Supervisor.server?(otp_app, endpoint)
   end
+
+  @doc """
+  Disables FLoC on Google Chrome browsers.
+
+  This function adds the HTTP-header `Permission-Policy: interest-cohort=()` to the response,
+  which tells the browser to exclude the page from the the FLoC calculation.
+
+  For more information regarding FLoC, please visit https://web.dev/floc.
+  """
+  def disable_floc(conn, true), do: Plug.Conn.put_resp_header(conn, "permissions-policy", "interest-cohort=()")
+  def disable_floc(conn, false), do: conn
 end
--- a/installer/templates/phx_web/endpoint.ex
+++ b/installer/templates/phx_web/endpoint.ex
@@ -50,6 +50,6 @@ defmodule <%= @endpoint_module %> do
   plug Plug.MethodOverride
   plug Plug.Head
   plug Plug.Session, @session_options
-  plug :set_floc, false
+  plug :disable_floc, false
   plug <%= @web_namespace %>.Router
 end
--- a/installer/templates/phx_web/endpoint.ex
+++ b/installer/templates/phx_web/endpoint.ex
@@ -50,5 +50,6 @@ defmodule <%= @endpoint_module %> do
   plug Plug.MethodOverride
   plug Plug.Head
   plug Plug.Session, @session_options
+  plug :set_floc, false
   plug <%= @web_namespace %>.Router
 end


mcrumm:
Hi @Awlexus and thanks for the PR! I am 💯 on this, so much so that I began a similar effort last night (151fdd0d) :)

I stopped short of opening a PR because the more I read about [`Permissions-Policy`](https://github.com/w3c/webappsec-permissions-policy/blob/main/permissions-policy-explainer.md) the more it seemed like something that should be a Plug in its own right, something like `put_permissions_policy/2`:

```ex
    plug :put_permissions_policy, %{
      "geolocation" => ~w(self "https://foo.com"),
      "camera" => [],
      "fullscreen" => "*"
    }
```

Then we could default to `[{"interest-cohort", []}]` and build overrides similarly to `put_secure_browser_headers/2`.  Wdyt?

// @phoenixframework/phoenix-core

Awlexus:
Thanks @mcrumm, that's the feedback I was hoping for. 
> the more I read about Permissions-Policy the more it seemed like something that should be a Plug in its own right, something like put_permissions_policy/2

I honestly had only thought of disabling FLoC, but adding proper support for this header would be preferable and I like the approach you mentioned :heart: 

Glad too see that there are more people pushing for sane privacy defaults! Should I close this PR in favour your idea?

mcrumm:
To be clear, in retrospect I don't think my commit is the way to go, either. It was a "quick change" but I'd rather we were more explicit about the behaviour.

@Awlexus would you like to work on a PR for `Controller.put_permissions_policy/2`? 

Then we can add it to the router template in the installer just above :put_secure_browser_headers. In the future if someone doesn't want it they can just remove it from the pipeline, but adding it will be just as easy for existing installs :)

Awlexus:
> would you like to work on a PR for Controller.put_permissions_policy/2?

I'd gladly do so. I'd start a new PR and close this one. Thanks for the suggestions!

lib/phoenix/endpoint.ex
```diff
@@ -905,4 +905,15 @@ defmodule Phoenix.Endpoint do
   def server?(otp_app, endpoint) when is_atom(otp_app) and is_atom(endpoint) do
     Phoenix.Endpoint.Supervisor.server?(otp_app, endpoint)
   end
+
+  @doc """
+  Disables FLoC on Google Chrome browsers.
+
+  This function adds the HTTP-header `Permission-Policy: interest-cohort=()` to the response,
+  which tells the browser to exclude the page from the the FLoC calculation.
```
eksperimental, line 913:
```suggestion
  which tells the browser to exclude the page from the FLoC calculation.
```

lib/phoenix/endpoint.ex
```diff
@@ -905,4 +905,15 @@ defmodule Phoenix.Endpoint do
...
+
+  For more information regarding FLoC, please visit https://web.dev/floc.
```
eksperimental, line 915:
```suggestion
  For more information regarding FLoC, please visit <https://web.dev/floc>.
```

eksperimental (changes requested):

aselder:
While I like the privacy implication of this change, has the behavior of Chrome (and Chrome based browsers) been checked with this setting to make sure bad things don't happen. If behavior degrades, do we want Phoenix to be broken (but with better privacy) by default on those browsers? Also the documentation should have instructions for developers who might want to opt-in to FLoC.

mcrumm:
All great points @aselder, thank you for making them

@Awlexus I am going to close this in favor of the upcoming PR and we can continue the discussion there :)

josevalim:
Has the permission policy been accepted as part of a standard already? Or is it still in flux? I would avoid adding stuff to it to core until it is approved because we would have bitten by it in the past (such as the new cookie flags that changed right until becoming approved).

PR #4306, opened 2021-04-27
Fix multi-stage build args in dockerfile example

ntodd:
Docker ARGs are scoped per build stage (https://github.com/moby/moby/issues/37345#issuecomment-400245466). The `MIX_ENV` ARG is created in the `build` stage and is referenced in the `app` stage. This results in an incorrect `COPY` directory. The proposed change creates a global `MIX_ENV` ARG that is then usable in both stages.


mcrumm:
@ntodd Good catch, thanks!

PR #4309, opened 2021-04-30
Schedule heartbeat on reply to avoid intensive throttling

### LINKED ISSUE #4307: Chained timers causing websockets to be closed in hidden browser tabs
### Environment

* Elixir version (elixir -v): 1.11.2
* Node.js version (node -v): 14.15.5
* NPM version (npm -v): 7.10.0
* Operating system: macos 11.2.1

### Expected behavior

We have a phoenix server manages websocket connections to ember.js applications using https://github.com/mike-north/ember-phoenix. ember-phoenix wraps phoenix's js. 

* The user loads the ember app in Chrome browser, and it connects to N channels and topics on the phoenix server.
* The user opens another browser tab.
* Whenever the user returns to the browser tab with the application, the sockets are still connected, and have never been closed.

### Actual behavior

* The user loads the ember app in Chrome browser, and it connects to N channels and topics on the phoenix server.
* The user opens another browser tab.
* After several minutes (the range varies from 3-12m, but if you want to see the issue, wait for 12m) the user returns to the browser tab with the application, the sockets have been closed and reopened at least once.

This problem is mentioned with regard to Safari in this discussion post: https://elixirforum.com/t/pushing-notifications-to-your-browser-with-liveview/37553/2

However, as of version 88, Chrome now does ['heavy throttling' of chained JS timers](https://developer.chrome.com/blog/timer-throttling-in-chrome-88/). As far as I can tell, it's phoenix's JS itself that implements chained JS timers ([here perhaps](https://github.com/phoenixframework/phoenix/blob/master/assets/js/phoenix.js#L1041-L1044)), and is therefore the reason reason Chrome is choosing to kill the websockets. If that is the case, it would be ironic considering one of google's recommendations to get around changed timers is to use websockets.

Here is an example that – if left open for a long period – shows Chrome does not necessarily and always kill websockets on an interval: https://codepen.io/matt-west/pen/tHlBb

mcrumm:
Closes #4307 

According to a Chromium bug report, intensive throttling is not applied to timers that are scheduled from a "network response handler". This was in response to Chrome 88 breaking Microsoft SignalR in a similar way to what's been reported.

I refactored our heartbeat based on their reference:
https://bugs.chromium.org/p/chromium/issues/detail?id=1186569#c3

**EDIT:** I am adding the annotated workflow for this change to reduce any confusion about what is going on.

1. First, `resetHeartbeat()` is called from `onConnOpen`
2. `resetHeartbeat` schedules `sendHeartbeat()` to be invoked after the heartbeat interval (default 30s). Note: the returned timer is not tracked.
3. `sendHeartbeat` then...
    1. generates `pendingHeartbeatRef`
    2. pushes the heartbeat message to the channel
    3. schedules `heartbeatTimeout()` to be invoked after the heartbeat interval. Note: this timer is tracked by `heartbeatTimer`
4. Next in `onConnMessage`, if the message ref matches `pendingHeartbeatRef`, then...
    1. `pendingHeartbeatRef` is cleared
    2. `heartbeatTimer` is cancelled
    3. The next invocation of `sendHeartbeat()` is scheduled.
5. If it happens that `heartbeatTimeout()` is ever invoked, the connection will be closed and an timeout error logged.

tl;dr the initial `sendHeartbeat()` is always scheduled and invoked. Subsequent heartbeats are scheduled once the previous heartbeat response has been received.

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -1039,9 +1039,11 @@ export class Socket {
    */
 
   heartbeatTimeout(){
-    this.pendingHeartbeatRef = null
-    if (this.hasLogger()) this.log("transport", "heartbeat timeout. Attempting to re-establish connection")
-    this.abnormalClose("heartbeat timeout")
+    if(this.pendingHeartbeatRef){
+      this.pendingHeartbeatRef = null
+      if (this.hasLogger()) this.log("transport", "heartbeat timeout. Attempting to re-establish connection")
+      this.abnormalClose("heartbeat timeout")
+    }
   }
 
   resetHeartbeat(){ if(this.conn && this.conn.skipHeartbeat){ return }
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -1215,7 +1215,7 @@ export class Socket {
 
   abnormalClose(reason){
     this.closeWasClean = false
-    if(this.conn.readyState === SOCKET_STATES.open){ this.conn.close(WS_CLOSE_NORMAL, reason) }
+    if(this.isConnected()){ this.conn.close(WS_CLOSE_NORMAL, reason) }
   }
 
   flushSendBuffer(){
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -1041,7 +1041,7 @@ export class Socket {
   heartbeatTimeout(){
     if(this.pendingHeartbeatRef){
       this.pendingHeartbeatRef = null
-      if (this.hasLogger()) this.log("transport", "heartbeat timeout. Attempting to re-establish connection")
+      if(this.hasLogger()){ this.log("transport", "heartbeat timeout. Attempting to re-establish connection") }
       this.abnormalClose("heartbeat timeout")
     }
   }
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -1038,10 +1038,16 @@ export class Socket {
    * @private
    */
 
+  heartbeatTimeout(){
+    this.pendingHeartbeatRef = null
+    if (this.hasLogger()) this.log("transport", "heartbeat timeout. Attempting to re-establish connection")
+    this.abnormalClose("heartbeat timeout")
+  }
+
   resetHeartbeat(){ if(this.conn && this.conn.skipHeartbeat){ return }
     this.pendingHeartbeatRef = null
-    clearInterval(this.heartbeatTimer)
-    this.heartbeatTimer = setInterval(() => this.sendHeartbeat(), this.heartbeatIntervalMs)
+    clearTimeout(this.heartbeatTimer)
+    setTimeout(() => this.sendHeartbeat(), this.heartbeatIntervalMs)
   }
 
   teardown(callback, code, reason){
@@ -1090,7 +1096,7 @@ export class Socket {
   onConnClose(event){
     if (this.hasLogger()) this.log("transport", "close", event)
     this.triggerChanError()
-    clearInterval(this.heartbeatTimer)
+    clearTimeout(this.heartbeatTimer)
     if(!this.closeWasClean){
       this.reconnectTimer.scheduleTimeout()
     }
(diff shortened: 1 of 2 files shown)


jeregrine (approved):
Lgtm

jeregrine (approved):
I had to do something very similar when load testing to keep the socket alive with a custom phoenix.js adapter. Testing changes look good to me!

patrickberkeley:
@mcrumm Thanks for taking this on. I'm testing out your changes locally. 

Regarding the approach for fixing the issue: Would it be better to initiate the heartbeat on the server rather than on the client? A few reasons I think it makes sense: 

1. The server is a relatively contained/controlled environment and not susceptible to the whims of browser vendors.
2. One of the goals of using websockets is to eliminate polling from the UI. By initiating the heartbeat on a set interval in the UI, it is essentially polling over a socket.
3. A bunch of the code in phoenix.js could be removed. All it would need to do is reply to the server's heartbeat `ping`.

mcrumm:
Hi @patrickberkeley - please do let us know if this fixes your issue. I tested these changes manually on a project last night and didn't see any disconnects from my LiveView socket this morning (though I saw many from live_reload which was not using socket code with the fix).

As far as whether or not to initiate the ping from the server– the answer is above my pay grade, so I will leave that for @chrismccord :)  My hope is that regardless of whatever we may choose to do in the future, we can apply this patch to avoid a regression  in connectivity for background tabs in Chrome.

patrickberkeley:
@mcrumm after testing further, I see an error coming in periodically:

```
Uncaught TypeError: Cannot read property 'readyState' of null
```

Which points to this line: https://github.com/phoenixframework/phoenix/blob/edc730aa91b819eafc59230f88f3993616daf72e/assets/js/phoenix.js#L1218

mcrumm:
Thanks for the report @patrickberkeley - afaict the `conn` only gets null-ified in the teardown function, so somewhere teardown was invoked before the heartbeat timeout was triggered.

Can you provide any more context as to what happens before you see this error?

We could add a conditional to abnormalClose to check for `this.conn`, but before we do so I'd like to be sure that we are not masking a deeper problem.

patrickberkeley:
@mcrumm I spoke too soon. I had not correctly removed the old version of phoenix.js when I added the version you have in this pull request. Needless to say that caused cardiac arrhythmia 💓 

Sorry for the false report.

patrickberkeley:
@mcrumm I'm going back and forth here, but last night during our end-to-end test runs, we had 116 instances of the `Uncaught TypeError: Cannot read property 'readyState' of null`. These were against instances that included the changes in this PR (and did not also include the old phoenix.js code).

patrickberkeley:
Here is the stacktrace:

```log
phoenix.js:1501 Uncaught TypeError: Cannot read property 'readyState' of null
    at Socket.abnormalClose (phoenix.js:1501)
    at Socket.heartbeatTimeout (phoenix.js:1278)
    at phoenix.js:1495
abnormalClose @ phoenix.js:1501
heartbeatTimeout @ phoenix.js:1278
(anonymous) @ phoenix.js:1495
setTimeout (async)
sendHeartbeat @ phoenix.js:1495
(anonymous) @ phoenix.js:1289
setTimeout (async)
resetHeartbeat @ phoenix.js:1289
onConnOpen @ phoenix.js:1266
conn.onopen @ phoenix.js:1178
```

patrickberkeley:
@mcrumm I think the above error might be a timing issue where the connection is either not fully open or fully registered on the socket before the first heartbeat tick is called with the setTimeout. So the code executes all the way through to `abnormalClose` and blows up because there is no `this.conn` yet.

mcrumm:
Thanks @patrickberkeley! That was initial impresseion @jeregrine and I had as well.

assets/js/phoenix.js
```diff
@@ -1038,10 +1038,18 @@ export class Socket {
    * @private
    */
 
+  heartbeatTimeout(){
+    if(this.pendingHeartbeatRef){
+      this.pendingHeartbeatRef = null
+      if (this.hasLogger()) this.log("transport", "heartbeat timeout. Attempting to re-establish connection")
```
chrismccord, line 1044:
```suggestion
      if(this.hasLogger()){ this.log("transport", "heartbeat timeout. Attempting to re-establish connection") }
```

chrismccord:
❤️❤️❤️🐥🔥

patrickberkeley:
@chrismccord thanks for merging. Will you clarify why phoenix initiates the heartbeat on in the UI please? (more info: https://github.com/phoenixframework/phoenix/pull/4309#issuecomment-830213609)

chrismccord:
We want to avoid creating many many timers on the server

patrickberkeley:
Okay, I didn't realize creating timers on the server would be expensive

chrismccord:
More expensive than not doing them for every connection :)

patrickberkeley:
Understood ✅ Thanks for the explanation.

patrickberkeley:
@mcrumm would you mind publishing a new version to npm that includes the changes here please?

mcrumm:
It will be available in the next release, although we should probably make sure we backport 46dccae to the v1.5 branch too, thanks!

// @chrismccord

patrickberkeley:
✔️ okay thanks @mcrumm

patrickberkeley:
@mcrumm do you have any insight as to when the next release will be cut? We are hoping to remove our locally-patched code for this.

PR #4311, opened 2021-04-30
Update Phoenix Endpoint drainer configuration docs

axelson:
Was out of sync with Phoenix.Endpoint.Cowboy2Adapter
Also added a mention of "top-level" options to make it more clear that
the Phoenix.Endpoint.Cowboy2Adapter configuration lives at the same
level as all the other endpoint configuration and is nested under any
other key.

Also update the link to Plug.Cowboy.Drainer to be more specific/correct


axelson:
Note: the test failure is unrelated since this is just a doc PR.

mcrumm:
Thanks! <3

PR #4317, opened 2021-05-08
Update application filepath in Mix Tasks guide

allknowingfrog:
The current behavior of `mix phx.new hello` is to create an application at `lib/hello/application.ex`. The Mix Tasks guide referred to the application living in `lib/hello.ex` in a couple of places. Update these paths to reflect the current behavior and the current console output.

I'm very new to Phoenix and Elixir, so please let me know if this is just a misunderstanding on my part. I've been working my way through the guides and this one didn't quite match what I'm seeing.


chrismccord:
❤️❤️❤️🐥🔥

PR #4322, opened 2021-05-12
Implement Phoenix.Param for Float

dnsbty:
I tried to place a float in the query string of a URL using the router helpers like this:

```elixir
latitude = 32.149989
longitude = -110.835842
MyAppWeb.Router.Helpers.page_path(:search, %{latitude: latitude, longitude: longitude})
```

But I get the error

```
** (Protocol.UndefinedError) protocol Phoenix.Param not implemented for 32.149989 of type Float. This protocol is implemented for the following type(s): Any, Map, Integer, BitString, Atom
```

I've implemented this in my own application, but I felt like Float is a common enough data structure that it would make sense for me to add this change back upstream for everyone else too. I'm happy to make any changes necessary.

### DIFF
--- a/lib/phoenix/param.ex
+++ b/lib/phoenix/param.ex
@@ -58,6 +58,10 @@ defimpl Phoenix.Param, for: Integer do
   def to_param(int), do: Integer.to_string(int)
 end
 
+defimpl Phoenix.Param, for: Float do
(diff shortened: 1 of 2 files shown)


chrismccord:
❤️❤️❤️🐥🔥

Issue #4323, opened 2021-05-12
Trigger long poll channel messages in own event loop


chrismccord:
Simulates behavior of `websocket.onmessage` instead of firing accumulated messages in the same event loop

PR #4325, opened 2021-05-12
Fire each event in a separate task for the LongPoll transport

hansihe:
Should fix https://github.com/phoenixframework/phoenix/issues/4323.

I originally intended to create a promise and resolving it immediately in order to run in a microtask instead of a macrotask, but that causes the same ordering issue, just one more level down (promises fired from promises).

A slightly more optimal approach here would be to fire the first event within the current task, and any subsequent ones from a `setInterval`. This might be advantageous if the lower bound timeout of `setTimeout` in some modern browsers end up creating problems, but that might cause even more ordering issues.

Added a commend, but I can remove that if the source code is an inappropriate location.

### DIFF
--- a/assets/js/phoenix.js
+++ b/assets/js/phoenix.js
@@ -1302,7 +1302,29 @@ export class LongPoll {
 
       switch(status){
         case 200:
-          messages.forEach(msg => this.onmessage({data: msg}))
+          messages.forEach(msg => {
+            // Tasks are what things like event handlers, setTimeout callbacks,
+            // promise resolves and more are run within.
+            // In modern browsers, there are two different kinds of tasks,
+            // microtasks and macrotasks.
+            // Microtasks are mainly used for Promises, while macrotasks are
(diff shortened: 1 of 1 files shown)


hansihe:
Repro is located here: https://github.com/hansihe/phoenix_channel_microtask_repro/blob/main/assets/js/socket.js#L67-L83

With the WS transport, this produces `a b c d e`, while with the longpoll transport it produces a different order.

After this fix, both of them produce `a b c d e`.

mcrumm:
Thanks @hansihe, nice work!

Per our conversation yesterday, if my understanding of the issue described in #4309 holds, then these timeouts are not at risk of being throttled because they are scheduled within a network response handler, so the additional `setTimeout` calls should have zero impact in that regard.

hansihe:
Ahh, that makes sense, nice!

chrismccord:
❤️❤️❤️🐥🔥

PR #4326, opened 2021-05-12
Add *.eex suffix to *.ex template file names.

azizk:
When editing files that contain EEx template code you often have syntax highlighting that looks incorrect or broken.
I'd like to propose to keep it similar to `*.html.eex` templates and therefore rename the file endings to `*.ex.eex`.
So this PR will help editors (or git GUI tools) in choosing the correct syntax definition when showing EEx template files with Elixir code inside.

I didn't edit the places that refer to these files, so I'm sure I need to fix those as well. But first let's see what you think...


chrismccord:
Thank you but this change is very large and I would prefer to keep existing structure :)

azizk:
Yep, this one is pretty big, but I think it's still contained and manageable, isn't it? Do you think it's a lot of effort to make sure everything is 100% functional? I mean the tests are all green at the very least... Would it help to split it up by sub-folders and complete the whole thing in stages?

PR #4327, opened 2021-05-13
Fix dialyzer errors because of raise_with_help().

azizk:
:+1:


josevalim:
We don’t generate specs in any of our code, so we don’t have plans to change this. Thanks!

PR #4332, opened 2021-05-17
Draft of parametrization of the mix phx.server task

RobertDober:
I am well aware that this is not mergeable, it lacks testing, elegance, error handling, documentation, insight into how to do this best.

However I would be happy to _fix_ these points if the idea itself finds some acceptance.

Concerning YHS I need to add this code to almost all my `config/dev.exs` as I frequently run in dev mode, oldfashioned on a local network with many servers on some few machines.

So any guidance would be appreciated


josevalim:
Hi @RobertDober. We need to trade-off simplicity for new devs vs functionality, so I think this one is better handled directly in your apps. Thanks!

PR #4335, opened 2021-05-21
Use scripts defined in package.json for starting Webpack

sevenseacat:
I came across this while working with TailwindCSS and trying to get its JIT mode working, and figured it might be beneficial to contribute back.

There's already a `watch` script defined in package.json that starts webpack, so if we use it for the watcher then the command can be more easily tweaked (eg. to add `NODE_ENV=development` which Tailwind needs).

It works for npm (and yarn if using the comment version) - can we assume that users are using npm like this? I think maybe we can, because we use npm for assets if a user starts a new app and elects to install dependencies.

### DIFF
--- a/installer/templates/phx_single/config/dev.exs
+++ b/installer/templates/phx_single/config/dev.exs
@@ -14,11 +14,10 @@ config :<%= @app_name %>, <%= @endpoint_module %>,
   code_reloader: true,
   check_origin: false,
   watchers: <%= if @webpack do %>[
-    node: [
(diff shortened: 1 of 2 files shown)


cw789:
This would probably not work on Windows machines.
Was tried already some weeks ago: https://github.com/phoenixframework/phoenix/pull/4291

sevenseacat:
oh really? hmm

cw789:
Yes it is.

PR #4337, opened 2021-05-29
Asset pipeline rewrite

### LINKED ISSUE #4328: Improving the JS Asset build pipeline
I've already spoken with @chrismccord and got the :+1: on this. I'm opening this issue to track progress, indicate what I intend to do, how I intend to implement, and solicit feedback.

## What is being addressed

Right now the Webpack build pipeline pulls `phoenix.js`, `phoenix_html.js`, and `phoenix_live_view.js` from their respective Elixir package's `deps/<name>/priv/static` directories. See this screenshot from the compiled `app.js` file in a Phoenix 1.5.7 app:

Screen Shot 2021-05-15 at 8 49 12 AM

I've collapsed the code so the source paths are all visible. These files in `priv/static` are pre-compiled, minified, and obfuscated. Diagnosing app issues or just trying to gain a greater understanding of the JS side of things within an app is incredibly difficult. Adding developer ergonomics like sourcemaps don't make it down to the pre-compiled files. There have been changes to add sourcemaps to the build pipelines that produce these files but that is not the best solution. Instead the current Phoenix environment (dev, prod, test) should dictate how these dependencies are built rather than being pre-built.

In addition, `phoenix` and  `phoenix_live_view` are published to NPM. However they are included as `file` reference dependencies in `package.json`:

Screen Shot 2021-05-15 at 8 54 46 AM

This is fine as it eliminates the need to re-download from NPM. However, each package is set up with the `priv/static` files at as `main` in their `package.json`:

Screen Shot 2021-05-15 at 9 12 36 AM

So when `import { Socket } from "phoenix"` is called it is matching on the pre-compiled version only. And because the `file` reference in `package.json` just sets up a symlink:

Screen Shot 2021-05-15 at 8 56 29 AM

this is why the compiled `app.js` file references `..deps/<name>/priv/static/<file>.js`

There are two possible solutions:

1. Eliminate the pre-compiled JS files from the dependency packages. I know that part of the reasoning for them is so others that are not using Webpack can consume them. However, I would be interested to know if this is really a concern.
2. Change `main` to `assets/js` in each package

For the sake of causing the least amount of side-effects I'll move forward with the 2nd option for now. I'd like to get feedback on the reality of the 1st though.

bcardarella:
This is a comprehensive rewrite of the Phoenix asset pipeline. Closes #4328

LiveView has a similar pending PR: https://github.com/phoenixframework/phoenix_live_view/pull/1474

* Separated the single mono-js file into separate es6-style modules
* Updated all js dependencies
* Rewrote the webpack build pipeline for less complexity, modularity, and to take advantage of v5
* Implemented eslint to enforce @chrismccord's js style rules

This does include breaking changes for existing apps with legacy Webpack configurations. Upgrade instructions:

1. copy `package.json` from `installer/templates/phx_assets` to overwrite the existing `package.json` into your app's `assets/` directory.
2. `cd assets && rm -rf node_modules && rm package-lock.json`
3. `npm install`
4. copy `webpack.config.js` from `installer/templates/phx_assets` to overwrite the existing `webpack.config.js` in your `assets/` directory

**Note** if you have any customizations in your existing `webpack.config.js` you should port them to the new config. This PR includes an update from Webpack v4 to Webpack v5 so there may be incompatibilities with any plugins, loaders, etc... that you were using web Webpack v4

By default the asset pipeline will import the es6 modules for Phoenix (as well as Phoenix LiveView). This means it is no longer building from pre-compiled JS. The result is that in dev mode the Phoenix and LiveView JS is no longer minified and no longer uglified. Debugging and stepping through code at runtime is now much easier. In production the JavaScript is minified and uglified as well as the CSS.

Babel has been removed from the build pipeline for apps. This was done because the target minimum browser version for Phoenix is being bumped up. Targeting IE11 is no longer in scope and we no longe need to include the Babel polyfill anymore. If you still require Babel for your own app's needs you can easily add Babel to your `package.json` and add the correct rules/loaders to `webpack.config.js`.

Sourcemaps are now included out of the box. The previous sourcemap implementation favored compilation speed over how useful they were. Now sourcemaps are generated and can be accessible in the developer console:

Screen Shot 2021-05-29 at 2 01 11 PM

**TODO** for Sourcemaps: the naming of the asset files needs to be improved. They currently compile with their relative path names from the Phoenix and Phoenix LiveView directories (i.e. "js/phoenix/channel.js") This should be improved to just be "phoenix/channel.js"

ESlint was added with a set of rules that attempt to conform to @chrismccord's desired JS style. Currently the linter is not run with the test suite. You can run from the `assets/` directory: `npx eslint js` and you can add `--fix` if you trust the linter to auto-fix anything that isn't compliant. I do think that the linter should be part of the test suite to help prevent any future style fix requests on PRs

There should not be any changes necessary to consuming app's `app.js` to consume the new es6 modules. I made sure to export the necessary modules properly.

**TBD**

Anyone not using webpack please weigh in on this. The changes I've made were with webpack in mind. I removed quite a bit that you may be taking for granted. Let's discuss that. While adding complexity to this PR can be done it isn't desirable. Webpack is being seen as the anointed asset pipeline solution for Phoneix. If you are using something else I believe a npm module to auto-configure what you need is the best step forward (i.e. creating a `phoenix-rollup` package) rather than adding to what Phoenix generates. However, if there is something missing that is an absolute blocker necessary to build in your own environment let's discuss what that is and how best to support.

**TODO**

- [ ] CI if failing due to an attempt to copy the no longer existing babelrc file
- [x] LiveReload in dev mode attempts to find source maps... this likely isn't a problem for release as `priv/static/phoenix.js` should be a **production** build and not include any references to source maps.


tozz:
As someone who's interested in Phoenix and doing mostly frontend, a project like this shouldn't come with ESLint configuration. That is such an extremely team based thing and will only cause upgrade issues with diffs.

You could place the Phoenix files in their own directory outside the `js` path and alias them with `@phoenix` or similar for isolation.

bcardarella:
@tozz eslint isn't enforced for consuming apps

tozz:
@bcardarella Oh yeah, my bad, this is not the actual assets folder, it's just the naming of things that makes it confusing (things are named the same in source vs created project but are completely different) 😅

bcardarella:
@tozz fwiw I think having the `@phoenix` scoped package name would be great but I have no idea if that is something the core team wants to take on

bcardarella:
@chrismccord this test: https://github.com/phoenixframework/phoenix/blob/master/installer/test/phx_new_test.exs#L17-L22
is currently failing. Are the files in `installer/templates/phx_static/` intended to be duplicates of what is generated into `priv/static` of the phoenix repo? I'm wondering if the copying of `phoenix.js` is still necessary with this PR

assets/.babelrc
```diff
@@ -2,4 +2,4 @@
     "presets": [
         "@babel/preset-env"
     ]
-}
+}
```
AndrewDryga, line 5: Is it intentional to have no newlines at the end of all files?

bcardarella: It isn't. I bet my editor did that. I'll correct

bcardarella: This should be fixed now. I added an eslint rule to catch it for the js files in the future, I don't think it will catch on this config files though.

tomciopp:
@bcardarella Any thoughts on using [webpack-merge](https://github.com/survivejs/webpack-merge) for changing webpack configuration based on your current environment?

bcardarella:
@tomciopp if I am understanding the README correctly, all this plugin is doing is a left->right object merge. So this would be used to write your own custom webpack config in a different file then merge into the default? If this is the case I don't really see a need for this. Why not just change what is generated from the template?

bcardarella:
@tomciopp I see what you're saying but our goal here is to *reduce* complexity. For the majority of users they're unlikely to need to make changes to the webpack config. For those that want to they can add this plugin if they don't want to make many changes to the generated template. The asset pipeline got into its current state because things kept getting added and after a few conversations with @chrismccord to this point we're going to keep the default template as sparse and basic as possible. Nothing prevents people from adding whatever they'd like to customize away from the default.

tomciopp:
@bcardarella 

>  For the majority of users they're unlikely to need to make changes to the webpack config. 

I'm not so sure about the validity of this statement, based on personal experience I've found it to be a necessity for every project. YMMV. I understand your point of reducing complexity, but if you are right and the majority of users don't change the configuration, shouldn't we optimize for people who need to? This way we aren't leaving people in hell if they aren't power users of webpack and need to make changes. 

Perhaps a blog post would suffice for my opinionated guide for phoenix & webpack configuration for larger scale projects.

snewcomer (approved):
This is amazing!  Do we need the `babel.config.json` btw?

I also checked the asset size real quick.  Currently `phoenix.js` is around 25kb.  This PR it seems to have dropped to 20kb!

Also, what is the plan for versioning given the breaking changes?

installer/templates/phx_assets/webpack.config.js
```diff
@@ -1,19 +1,20 @@
...
-const MiniCssExtractPlugin = require('mini-css-extract-plugin');
-const TerserPlugin = require('terser-webpack-plugin');
-const OptimizeCSSAssetsPlugin = require('optimize-css-assets-webpack-plugin');
-const CopyWebpackPlugin = require('copy-webpack-plugin');
+const CssMinimizerPlugin = require('css-minimizer-webpack-plugin');
+const MiniCssExtractPlugin = require("mini-css-extract-plugin");
 
 module.exports = (env, options) => {
   const devMode = options.mode !== 'production';
+  console.log(`CURRENT MODE: ${devMode}`)
```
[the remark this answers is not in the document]
bcardarella: Good catch, will rm

package.json
```diff
@@ -3,11 +3,10 @@
...
   "description": "The official JavaScript client for the Phoenix web framework.",
   "license": "MIT",
-  "main": "./priv/static/phoenix.js",
-  "module": "./assets/js/phoenix.js",
+  "main": "priv/static/phoenix.js",
   "exports": {
-    "import": "./assets/js/phoenix.js",
-    "require": "./priv/static/phoenix.js"
+    ".": "./assets/js/phoenix/index.js",
+    "./": "./assets/js/phoenix/"
```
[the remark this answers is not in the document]
bcardarella: correct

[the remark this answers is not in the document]
bcardarella: if there is a pattern to do in 1 LOC I'm open to it

installer/templates/phx_assets/package.json
```diff
@@ -11,16 +11,13 @@
...
-    "babel-loader": "^8.0.0",
-    "copy-webpack-plugin": "^5.1.1",
-    "css-loader": "^3.4.2",
-    "hard-source-webpack-plugin": "^0.13.1",
-    "mini-css-extract-plugin": "^0.9.0",
-    "optimize-css-assets-webpack-plugin": "^5.0.1",
-    "terser-webpack-plugin": "^2.3.2",
-    "webpack": "4.41.5",
-    "webpack-cli": "^3.3.2"
+    "copy-webpack-plugin": "^9.0.0",
```
cw789, line 14: I do not see `copy-webpack-plugin` beeing used anymore within `webpack.config.js`.
Means it could be removed here.

But by removing `copy-webpack-plugin` I think in `installer/lib/phx_new/single.ex` copy path of `robots.txt`, `phoenix.png` and `favicon.ico` should be changed from `assets/static/` to `priv/static/`.

With this change `/priv/static` should get removed in the `.gitignore`.

installer/templates/phx_assets/package.json
```diff
@@ -11,16 +11,13 @@
...
+    "css-loader": "^5.2.6",
+    "css-minimizer-webpack-plugin": "^3.0.0",
+    "expose-loader": "3.0.0",
```
cw789, line 17: `expose-loader` not being used within `webpack.config.js`.

cw789 (changes requested):
Thanks for starting this topic. I took a quick look at the installer things, without testing or checking out this branch.

If any questions arises - please feel free.

installer/templates/phx_assets/package.json
bcardarella: moving them over to `priv/static` might make sense, but it depends upon what experience @chrismccord wants here. If all web assets should be in the `assets/` directory then copied over at build time I can bring back the copy rule.

installer/templates/phx_assets/package.json
bcardarella: I left it in because @chrismccord and I were going back and forth on if the `Phoenix` constant should be exposed. He fell on the side that it wasn't necessary but I wanted to give the community a chance to provide feedback if for some reason a need for that constant is being overlooked.

installer/templates/phx_assets/package.json
chrismccord: We need to support copying assets from `assets/static/` to `priv/static`, so we'lll need copy-webpack unless webpack can do something like this out of the box now. This allows folks to put all assets inside the assets directory, and build them into private/static, which is especially important for our digest task, which will digest the copied/built assets

cw789: Then it would be necessary to bring it back in.

I'm not aware of any simpler out of the box webpack solution.

One thing I once encountered but haven't evaluated in detail then,
`copy-webpack-plugin` caused to copy (or touch all copy files) if one of them canged. Means live reload reloads x times (or logged to reload).
But could be, that this was only true fot a certain version of the updated plugin.

bcardarella: @cw789 I'm working on putting it back right now, the API changed with the version bump so I just have to ensure that it's doing everything we want.

installer/templates/phx_assets/package.json
chrismccord: To be clear, the `Phoenix` constant needs to be on window when the built js is loaded, for folks not using an asset builder and for the live reload project. If we can ditch `expose-loader` and do this directly then great, but we need to make sure it is exposed from the bundle.

bcardarella: It could *technically* be done without `expose-loader` but I think it s going to be some crazy rule writing. I would vote to keep the dependency to do this.

bcardarella: To be clear, keeping it on the internal webpack config, not the one in the installer template

chrismccord: works for me

chrismccord:
❤️❤️❤️🐥🔥

gcauchon:
@bcardarella @chrismccord Is there a reason `assets/babel.config.json` was renamed (to follow the `.config.json` convention) and not deleted?

bcardarella:
@gcauchon babel is being used for the test suite for the js files

bcardarella:
FWIW if anyone want to PR to have the js test suite use Webpack to resolve the es6 modules so we can drop Babel there too I'm all for it. I just gave up fighting JS that day I think.

PR #4340, opened 2021-06-05
Update gitignore

andrecp:
Add gitignore to .DS_Store file


optikfluffel:
@andrecp I think that should be something that's in your [global gitignore file](https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/), since it's not really something project and/or Phoenix specific.

josevalim:
Closing per the above. :)

Issue #4343, opened 2021-06-07
Add mount_current_user on generated Auth module for LiveView auth


josevalim:
We already have fetch_current_user for plugs and we also need a mount_current_user to be used by the upcoming LiveView callbacks. The docs should also mention how to do the LV setup.

PR #4344, opened 2021-06-08
add: mount_current_user/2

mad42:
#4343


josevalim:
Thanks @mad42! The idea is to generate this inside the existing `<%= mod %>Auth` module. In any case, before we do this, we need to wait for LiveView to add the new hook APIs, otherwise this will likely be out of sync if what is added there. Thanks!

priv/templates/phx.gen.auth/auth_helpers.ex
```diff
@@ -0,0 +1,31 @@
...
+
+      # LiveView mount
+      def mount(_params, session, socket) do
+        socket = mount_current_<%= schema.singular %>(socket, session)
+        {:ok, socket}
+      end
+
+  """
+  def mount_current_<%= schema.singular %>(socket, %{"<%=schema.singular %>_token" => <%= schema.singular %>_token}) do
+    if connected?(socket) do
```
josevalim, line 19: Why the `connected?` branch? :)

priv/templates/phx.gen.auth/auth_helpers.ex
```diff
@@ -0,0 +1,31 @@
...
+        <%= inspect context.module %>.get_<%= schema.singular %>_by_session_token(<%= schema.singular %>_token)
+      end)
+    else
+      assign_new(socket, :current_<%= schema.singular %>, fn -> nil end)
+    end
+  end
+
+  def mount_current_<%= schema.singular %>(socket, _session) do
+    assign_new(socket, :current_<%= schema.singular %>, fn -> nil end)
+  end
```
josevalim, line 30: I would redirect if the current user is not available. In other words, I would have this be a strict helper by default, users can relax it. :)

priv/templates/phx.gen.auth/auth_helpers.ex
mad42: At first it was to prevent duplicating the call to get_user_by_session_token/1, but assign_new/3 can do the job alone :)

priv/templates/phx.gen.auth/auth_helpers.ex
mad42: I assumed the router already redirected using the :require_authenticated_user plug

mad42:
> Thanks @mad42! The idea is to generate this inside the existing `<%= mod %>Auth` module. In any case, before we do this, we need to wait for LiveView to add the new hook APIs, otherwise this will likely be out of sync if what is added there. Thanks!

My pleasure @josevalim :)
I've been following the wonderful work you guys did on the auth generator since the first draft/blog post and decided to share a function I use in every new app.

I've added that new file to keep LiveView files inside "live" folder, but since it will be one helper function, I see why you prefer to keep it in the `<%= mod %>Auth` module!

priv/templates/phx.gen.auth/auth_helpers.ex
josevalim: It is very important for LiveView to do their own authentication/authorization logic, especially in upcoming versions as we won’t be re-running the plug pipeline. :)

mad42:
In order to conditionally import `Phoenix.LiveView` into `<%= mod %>Auth` how can I know if LiveView is part of the actual app inside the generator ?

josevalim:
Oh, I see. This runs in the user project, so we can do a Code.endure_loaded? on Phoenix.LiveView before generating. Although the assign calls may conflict… should we import it only on the function that uses it? Or should we have a LiveAuth module?

mad42:
Code.ensure_loaded?/1 that what I was looking for 👍 

Importing only assign_new/3 in the `<%= mod %>Auth` may not rise conflict I guess, I think it can be good to have a separate modules for Auth and LiveAuth, and I also think it can be good to have all Auth helpers in the same place, your call ! 😄

josevalim:
What do you think about a UserLiveAuth module with two functions mount_current_user and mount_authenticate_user?

mad42:
I see the role of mount_current_user, but what will be the goal of mount_authenticate_user?

josevalim:
Sorry, mount_authenticated_user. The authenticated version would require it exists, opposite to the mount_current_user which could be nil… but maybe that’s not a good name.

mad42:
Ohh! something like maybe_mount_current_user and mount_current_user ?

josevalim:
Yeah but we don’t use the maybe terminology anywhere. In Plug we have two functions, one for getting the current user and another one for ensuring it is in assigns: fetch_current_user and ensure_authenticated_user. Maybe we should just use the same names, since it is a separate module? 🤔

mad42:
what about assign_current_user and ensure_assigned_user ?
using same name may be confusing even if it's a different module..

josevalim:
Right, if we want to avoid confusion, then we need to call it mount or something that only makes sense for LV, otherwise assign is not distinctive enough.

mad42:
Agreed! mount refers automatically to LiveView.
 
mount_current_user and ensure_mounted_current_user ? 😆

chrismccord:
This should take advantage of the new server hooks once they land https://github.com/phoenixframework/phoenix_live_view/pull/1511

priv/templates/phx.gen.auth/live_auth.ex
```diff
@@ -0,0 +1,53 @@
...
+      # LiveView mount
+      def mount(_params, session, socket) do
+        socket = ensure_mounted_current_<%= schema.singular %>(socket, session)
+        {:ok, socket}
+      end
+
+  """
+  def ensure_mounted_current_<%= schema.singular %>(socket, %{"<%=schema.singular %>_token" => <%= schema.singular %>_token}) do
+    assign_new(socket, :current_<%= schema.singular %>, fn ->
+      <%= inspect context.module %>.get_<%= schema.singular %>_by_session_token(<%= schema.singular %>_token)
```
henriquecf, line 46: I think that this is not ensuring a user, as this function may return nil. We would probably need to confirm after we get the result that it is not nil. Alternatively (and similar to what user_auth does) we could build this function on top of mount_current_user and only check if the current_user is assigned to the socket.

LostKobrakai: 👍 for the latter option.

mad42: fixed, now it halts and redirect to login route.

mad42:
`mix format --check-formatted` seems to fail,
I don't know why :man_shrugging:

priv/templates/phx.gen.auth/live_auth.ex
```diff
@@ -0,0 +1,61 @@
...
+      # In the LiveView file
+      defmodule DemoWeb.PageLive do
+        use Phoenix.LiveView
+
+        on_mount {<%= inspect context.web_module %>.<%= if schema.web_path, do: "#{Phoenix.Naming.humanize(schema.web_path)}.", else: "" %><%= Phoenix.Naming.humanize(schema.singular) %>LiveAuth, :mount_current_<%= schema.singular %>}
+      end
+
+  """
+  def mount_current_<%= schema.singular %>(socket, %{"<%=schema.singular %>_token" => <%= schema.singular %>_token}) do
+    socket = assign_new(socket, :current_<%= schema.singular %>, fn ->
```
connorlay, line 19: @mad42 I pulled your branch down locally and generated a new `user_live_auth.ex`, similar to the CI checks that are failing for this PR. Running mix format changes this line (and line 46) by moving the `assign_new/3` call onto a new line, like this:
```elixir
    socket =
      assign_new(socket, :current_user, fn ->
        DevApp.Accounts.get_user_by_session_token(user_token)
      end)
```
If you change this template to match that, I think the tests will start passing.

mad42: Thanks for pointing that 🙏

Gazler:
Closed by https://github.com/phoenixframework/phoenix/pull/4810

PR #4346, opened 2021-06-12
Require Elixir v1.12 for new apps

josevalim:


josevalim:
Hi @agbell, can you please give me a quick hand here? We need to do two changes:

1. The caching directory needs to be per Elixir version (if it isn't yet)

2. We should only run the installer tests on certain Elixir versions (in this case, v1.12) - how to split them and keep them efficiently?

Thanks!

agbell:
Hi @josevalim,
Yep, I'll take a look at it tomorrow.

agbell:
Check this out @josevalim https://github.com/phoenixframework/phoenix/pull/4348/files#diff-2705f6016a64189abd14041e2f3e8dbb829553295d9088e4eb42eccb5e61bf47R109

Earthfile
```diff
@@ -15,13 +15,16 @@ test:
     # Run unit tests
     RUN mix test
 
-    # Run installer tests
-    WORKDIR /src/installer
-    RUN mix test
+    IF $RUN_INSTALLER_TESTS == "1"
```
josevalim, line 18:
```suggestion
    IF [ $RUN_INSTALLER_TESTS = "1" ]
```

Earthfile
```diff
@@ -15,13 +15,16 @@ test:
     # Run unit tests
     RUN mix test
 
-    # Run installer tests
-    WORKDIR /src/installer
-    RUN mix test
+    IF [ $RUN_INSTALLER_TESTS = "1" ]
```
josevalim, line 18:
```suggestion
    IF [ "$RUN_INSTALLER_TESTS" = "1" ]
```

josevalim: @agbell please let me know if this is the way to go! Thank you so much! :heart:

Earthfile
```diff
@@ -15,13 +15,16 @@ test:
     # Run unit tests
     RUN mix test
 
-    # Run installer tests
-    WORKDIR /src/installer
-    RUN mix test
+    IF [ "$RUN_INSTALLER_TESTS" = "1" ]
```
josevalim, line 18: @agbell hrm, this is not working, but it should according to the docs. Is this perhaps a Earthfile version issue?

agbell: It must be, its a newer feature.  in the GitHub action, it has the download step, we could bump it to 
https://github.com/earthly/earthly/releases/tag/v0.5.16

josevalim: It works!!!

josevalim:

PR #4348, opened 2021-06-14
adding version check

agbell:
@josevalim I think this will do what you need

### DIFF
--- a/Earthfile
+++ b/Earthfile
@@ -15,9 +15,13 @@ test:
     # Run unit tests
     RUN mix test
 
+   IF echo $ELIXIR | version ">=1.12" 
     # Run installer tests
     WORKDIR /src/installer
     RUN mix test
+   ELSE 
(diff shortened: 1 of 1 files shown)


Earthfile
```diff
@@ -15,9 +15,13 @@ test:
     # Run unit tests
     RUN mix test
 
+   IF echo $ELIXIR | version ">=1.12"
```
josevalim, line 18: Thanks! I can also pass an env var called $RUN_INSTALLER_TESTS right?

josevalim:
Thanks! I ported this approach here: https://github.com/phoenixframework/phoenix/pull/4346/files#diff-2705f6016a64189abd14041e2f3e8dbb829553295d9088e4eb42eccb5e61bf47R18 :)

Earthfile
agbell: yep! The IFs should work like bash ifs

PR #4350, opened 2021-06-17
Add typespecs to Phoenix.Socket.Transport

sax:
I recently wrote a custom transport module to support WebSocket protocol extension, and initially thought that it was not possible to handle the extension via `Phoenix.Socket.Transport`.

After tracing through Phoenix sockets into `cowboy_websocket` and a lot of trial and error, I discovered all of the required behavior was actually possible, as Phoenix delegates most of the handling of these callbacks to cowboy. I thought that by documenting more of the valid return values as well as their consequences, I might save someone else the trouble of tracing the code paths.

Please let me know if there are any issues with these changes, or suggested modifications, and I would be happy to make changes.

### DIFF
--- a/lib/phoenix/socket/transport.ex
+++ b/lib/phoenix/socket/transport.ex
@@ -102,6 +102,37 @@ defmodule Phoenix.Socket.Transport do
 
   @type state :: term()
 
+  @type opcode :: :binary | :text
+
+  @typedoc """
+  Specifies a control frame, with opcode of `:ping` or `:pong`.
+
+  When replying from `c:handle_control/2`, a control frame must
+  be used.
+
+  Control frames may optonally be returned in replies from
+  `c:handle_in/2` or pushed to clients from `c:handle_info/2`.
+  This can be used to implement server-side keepalive, for instance.
+  """
+  @type control_frame ::
+          {:ping, message :: term}
+          | {:pong, message :: term}
+
+  @typedoc """
+  A tuple representing the frame type as well as its payload.
+
+  The frame type must be a valid `t:opcode/0` atom, which translates
+  to a specific 4-bit value defined by the WebSocket protocol.
+
+  When replying to a client or pushing a frame, using the tuple
+  `{:close, code, message}` will close the socket after pushing
+  the frame to the client.
+  """
+  @type frame ::
+          {opcode, message :: term}
+          | control_frame
+          | {:close, code :: integer, message :: binary}
+
   @doc """
   Returns a child specification for socket management.
 
@@ -175,9 +206,9 @@ defmodule Phoenix.Socket.Transport do
   `:binary` opcode and the message must be always iodata. Long polling only
   supports text opcode.
   """
-  @callback handle_in({message :: term, opts :: keyword}, state) ::
+  @callback handle_in({message :: term, [opcode: :binary | :text]}, state) ::
               {:ok, state}
-              | {:reply, :ok | :error, {opcode :: atom, message :: term}, state}
+              | {:reply, :ok | :error, frame, state}
               | {:stop, reason :: term, state}
 
   @doc """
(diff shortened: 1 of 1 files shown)


josevalim:
Hi @sax, thanks for the PR!

This is really well done but this is an extension of the contract we have today. Phoenix supports only text and binary opcodes and we don't support close or the control frames _by design_.

In other words, those changes are making our contract the same as the Cowboy one, but it is a subset of the Cowboy one on purpose, as we don't want to tie ourselves to the Cowboy implementation. Plus other transports may not even be able to handle those frames.

sax:
Thanks for the feedback, and for taking time to explain your reasoning.

FYI my need was to bridge the gap between Absinthe and a non-Phoenix GraphQL websocket protocol. I'm on a team building a customer-facing GraphQL API, and we don't believe we can can reasonably expect customers to implement the Phoenix channel API, if they're not building their products in Elixir.

I've worked on other products where a custom websocket extension (ie [RFC 7692](https://datatracker.ietf.org/doc/html/rfc7692.html)) would have resulted in a simpler solution than trying to force the clients into a Phoenix channel.

The three things that would make this easier/possible:
* Clarify that `Phoenix.Socket.Transport` is the thing to use when implementing a websocket in Phoenix. Blogs from Spring 2021 still tell the story that it's not possible in Phoenix—that one has to write the implementation in Cowboy and mess with the Phoenix dispatcher to write a non-Phoenix-Channel websocket implementation. The top section of the docs does say that `Phoenix.Socket` is just one implementation, but still implies that it's how to multiplex over channels.
* Support for server-side keepalive, ie initiating pings from the server.
* Close connections from the server with custom close code and message.

It would be nice if Phoenix supported these use cases in its contract, even if (maybe especially if?) not directly tied to the Cowboy implementation. If there was agreement on what that contract was, I would be happy to work on the code changes in Phoenix to support it—let me know and we can either continue the discussion here, or in another venue. I'll take silence as a decision that you don't want Phoenix to support this at all.

josevalim:
I mean, we could work towards supporting those, but then your Phoenix.Socket alternative is mostly a thin wrapper around Cowboy anyway, exactly because you said you won't be building on top of Phoenix Channels - for very valid reasons. So if Phoenix.Socket.Transport becomes a thin wrapper around Cowboy, why not just use Cowboy?

The downside is having to register a dispatcher, the upside is a much fine grained control over websockets, with no indirection for you to sort through. :)

josevalim:
Oh, one more thing: one reason why those changes could be justified is if you want your Channel-replacement to work across multiple transports, such as websocket, long-polling, and others. But your need for fine grained control probably means you want to be tied to a particular transport. Plus some of those likely won't work on some transports anyway.

sax:
I see. I was thinking this was not necessarily about Phoenix wrapping Cowboy, but in allowing `Phoenix.Socket.Transport` to support the full WebSocket protocol, which includes these use cases. Right now it's possible to take advantage of the fact that `Phoenix.Socket.Transport` in websocket mode _is_ a thin wrapper for Cowboy—it's just not declared in the Phoenix contract or documented.

If you were to close off the boundaries between Phoenix and Cowboy to only allow messages that _are_ specified in the contract, you would also be declaring that `Phoenix.Socket.Transport` only natively supports a subset of the WebSocket protocol. I can understand why you might want to do that, and that's one of my worries using this mechanism in production—it's possible that a future version of Phoenix might close off the ability to pass these messages through to Cowboy.

I think my homework is to read more about the different available transports, to see if what I'm asking for is supported cross-transport. If so, I can better argue for a change in the contract. If not, then I would suggest that the `Phoenix.Socket.Transport` documentation could clarify the fact that it only directly supports the subset of the protocols that work across all transports (and I'm happy to issue that as a pull request). If that's the case, then I should go the dispatcher route as you recommend.

Again, thanks for the conversation!

jasonkenney-drizly:
I can think of a valid use-case, if its what I think, I've been finding following the code a bit difficult.

I need to close/error a regularly functioning Phoenix websocket with channels that the client can discern means an expired token, or an incorrect token, as I want the socket to retry on normal disconnects, but not retry when its an authorization issue. Right now that does not seem possible.

PR #4352, opened 2021-06-23
Improve asset pipline docs

cw789:
This is a complementation of https://github.com/phoenixframework/phoenix/pull/4337.

Things done:
- Removed the Node.js version requirement from the installation guide
- Added instead the "engines" field with node >=14.x to the `package.json`  
  This should at least warn if an older Node.js version is used.
  But theoretically we still would support older versions.
  (webpack says it support 10.x - but you never really know with all the other dependencies)
- Changed the Node.js download page directly to the nodejs.org landing site (probably quicker for first time node users)


chrismccord:
❤️❤️❤️🐥🔥

PR #4355, opened 2021-06-30
Add mailer generator for phx.new mix task

philss:
This PR introduces a mailer generator for the `mix phx.new` command.

This generator sets up Swoosh for new applications. Both single and umbrella generators are capable of generating it.
The directory "emails" created inside web apps is a convention adopted by Swoosh.

Check https://hexdocs.pm/swoosh for more details.

PS: this feature was discussed before with José.


philss:
ping @josevalim

philss:
I'm working on the fix for the build and on changing some stuff for this PR. So please, don't review it yet.

josevalim:
@phills this looks ready to go to me! Quick question: should we wait for a version of Swoosh that does not require setting api_client to false or is it better to go ahead?

philss:
@josevalim I think it's OK to go ahead, because I couldn't focus on solving this yet. One question: should the other generator that we mention - the notifier one - go in another PR?

josevalim:
Let’s do another PR, yeah!

josevalim:

brightball:
I realize I'm late to the party on this, but I'd like to suggest some considerations for the future of this approach if we're standardizing email generation within the framework.

Email eventually becomes and application within the application as your project grows. It happens slowly as the application progresses and by the time you realize you need to untangle it, it's a project in and of itself.

Eventually you end up with a situation where email workers are polluting your general job queues. As the organization grows, non-developers take significantly more interest in the content and structure of email and those changes being pushed into the normal application pipeline, side tracking developers. You'll end up with a completely separate UI framework for emails like Zurb Foundation. You'll have dedicate queue and retry logic just for email. Eventually stats delivery and stats tracking will come in too, even when using a 3rd party service since usually your app will need to start factoring those stats into to certain functions (like user activity). Eventually, if you do have multiple services your emails will become a central service for all of them that just happens to exist inside whatever the first "core" app in your project was.

What I'm getting at with this is that if email is setup outside of the web app from the very beginning, it makes everything easier long term. I've talked about these challenges for a long time but Elixir is one language where it can be architecturally separated from the beginning without needing an entirely separate deployment or an external queuing system. 

Something along the lines of an email app within an umbrella could address this cleanly from the very beginning and then as  you grow, moving it to a fully standalone system would become trivial.

Just something to consider. 

I wrote[ a longer post about it for Codeship](https://www.cloudbees.com/blog/email-as-a-microservice) a few years ago that better explains my thoughts on the matter. I'm only commenting here because I think this is another area where Elixir can potentially make it easier to solve an issue experienced by many devs teams more effectively than any other language.

brightball:
ping @philss @josevalim for discussion sake since this has already been merged.

philss:
@brightball thanks for bringing those points! I think your concerns are valid but our intention is to start simple and build only the bootstrap for users to send emails. For most applications this is enough. 

The applications that are using _umbrella_ can use the "--context-app" flag when using the "phx.gen.notifier"  which allows the files to be generated in a specific app. This can achieve the isolation you mentioned.

For more complex cases I think it's up to the user to come with a better arrangement for its app, because it can vary a lot from case to case. I think the community can help with documentation - like your blog post, or guides - but the framework shouldn't delimit the lines for those scenarios IMHO.

brightball:
I can appreciate that @philss. I'll try out the `--context-app` flag when I get a chance and see if I can promote it's use with a blog post.

philss:
This reminds me that I didn't mention `--context-app` in the docs for that task. I will update in a upcoming PR :)

installer/test/phx_new_test.exs
```diff
@@ -191,6 +191,29 @@ defmodule Mix.Tasks.Phx.NewTest do
...
+      end
+
+      assert_file "phx_blog/lib/phx_blog/mailer.ex", fn file ->
+        assert file =~ "defmodule PhxBlog.Mailer do"
+        assert file =~ "use Swoosh.Mailer, otp_app: :phx_blog"
+      end
+
+      assert_file "phx_blog/config/config.exs", fn file ->
+        assert file =~ "config :swoosh"
+        assert file =~ "config :phx_blog, PhxBlogWeb.Mailer, adapter: Swoosh.Adapters.SMTP"
```
LostKobrakai, line 206: Is this an oversight? I would expect the namespace to be ˋPhxBlogˋ and not ˋ PhxBlogWebˋ given thats the module asserted for in line 200 of this test.

LostKobrakai: Seems like this has been fixed in a followup PR.

PR #4360, opened 2021-07-03
Fix default Swoosh adapter for mailer generator

philss:
This fix is needed because we don't want to include a new dependency
required for the "SMTP" adapter (gen_smtp). We choose to use the "Local"
adapter by default and mention Swoosh docs for more adapters.

### DIFF
--- a/installer/templates/phx_single/config/runtime.exs
+++ b/installer/templates/phx_single/config/runtime.exs
@@ -33,5 +33,27 @@ if config_env() == :prod do
   #     config :<%= @web_app_name %>, <%= @endpoint_module %>, server: true
   #
   # Then you can assemble a release by calling `mix release`.
-  # See `mix help release` for more information.
+  # See `mix help release` for more information.<%= if @mailer do %>
+
+  # ## Configuring the mailer
+  #
+  # In production you need to configure the mailer to use a different adapter.
+  # Also, you may need to configure the Swoosh API client of your choice if you
+  # are not using SMTP.
+  #
+  # Additionally you may need to add more dependencies required by your adapter.
+  # Here is an example of the configuration:
+  #
+  #     config :<%= @app_name %>, <%= @app_module %>.Mailer,
+  #       adapter: Swoosh.Adapters.Mailgun,
+  #       api_key: System.get_env("MAILGUN_API_KEY"),
+  #       domain: System.get_env("MAILGUN_DOMAIN")
+  #
+  # For this example you need include a HTTP client required by Swoosh API client.
+  # Swoosh supports Hackney and Finch, but you can write your own if you need.
+  # Finally configure Swoosh to use the given adapter:
+  #
+  #     config :swoosh, :api_client, Swoosh.ApiClient.Hackney
+  #
+  # See https://hexdocs.pm/swoosh/Swoosh.html#module-installation for details.<% end %>
 end
--- a/installer/templates/phx_single/config/runtime.exs
+++ b/installer/templates/phx_single/config/runtime.exs
@@ -39,10 +39,7 @@ if config_env() == :prod do
   #
   # In production you need to configure the mailer to use a different adapter.
   # Also, you may need to configure the Swoosh API client of your choice if you
-  # are not using SMTP.
-  #
-  # Additionally you may need to add more dependencies required by your adapter.
-  # Here is an example of the configuration:
+  # are not using SMTP. Here is an example of the configuration:
   #
   #     config :<%= @app_name %>, <%= @app_module %>.Mailer,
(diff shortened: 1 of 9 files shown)


philss:
ping @josevalim :)

josevalim:
We want to keep the configuration in dev because once they change the config.exs file, we want dev to stay the same.

Maybe we should even remove it from config.exs and move it to runtime.exs? This is because the config will likely have secrets, so runtime.exs is the best place for it. In this case, we have two options:

1. set local default in config.exs, nothing in dev, example comment in runtime.exs
2. nothing in config.exs, local in dev, local with comment in runtime.exs

also, given we are using local, I assume we no longer need to set api_client: false?

josevalim:
I believe I prefer the first option. The example in runtime.exs could be commented  out using an actual adapter and sample configuration.

philss:
> also, given we are using local, I assume we no longer need to set api_client: false?

unfortunately we can't now. This is a limitation from Swoosh and [changing the default would be a breaking change](https://github.com/swoosh/swoosh/issues/609). I couldn't think in an alternative for this :/ thoughts?

I added the commented snippet to `config/runtime.exs`. Sorry for the delay :grimacing: . Please check again.

installer/templates/phx_single/config/runtime.exs
```diff
@@ -33,5 +33,27 @@ if config_env() == :prod do
...
+  # Here is an example of the configuration:
+  #
+  #     config :<%= @app_name %>, <%= @app_module %>.Mailer,
+  #       adapter: Swoosh.Adapters.Mailgun,
+  #       api_key: System.get_env("MAILGUN_API_KEY"),
+  #       domain: System.get_env("MAILGUN_DOMAIN")
+  #
+  # For this example you need include a HTTP client required by Swoosh API client.
+  # Swoosh supports Hackney and Finch, but you can write your own if you need.
+  # Finally configure Swoosh to use the given adapter:
```
josevalim, line 54:
```suggestion
  # For this example you need include a HTTP client required by Swoosh API client.
  # Swoosh supports Hackney and Finch out of the box:
```

installer/templates/phx_umbrella/apps/app_name_web/config/runtime.exs
```diff
@@ -22,4 +22,26 @@ config :<%= @web_app_name %>, <%= @endpoint_module %>,
...
+# Here is an example of the configuration:
+#
+#     config :<%= @app_name %>, <%= @app_module %>.Mailer,
+#       adapter: Swoosh.Adapters.Mailgun,
+#       api_key: System.get_env("MAILGUN_API_KEY"),
+#       domain: System.get_env("MAILGUN_DOMAIN")
+#
+# For this example you need include a HTTP client required by Swoosh API client.
+# Swoosh supports Hackney and Finch, but you can write your own if you need.
+# Finally configure Swoosh to use the given adapter:
```
josevalim, line 43:
```suggestion
# For this example you need include a HTTP client required by Swoosh API client.
# Swoosh supports Hackney and Finch out of the box:
```

installer/templates/phx_umbrella/apps/app_name_web/config/runtime.exs
```diff
@@ -22,4 +22,26 @@ config :<%= @web_app_name %>, <%= @endpoint_module %>,
...
+# See `mix help release` for more information.<%= if @mailer do %>
+
+# ## Configuring the mailer
+#
+# In production you need to configure the mailer to use a different adapter.
+# Also, you may need to configure the Swoosh API client of your choice if you
+# are not using SMTP.
+#
+# Additionally you may need to add more dependencies required by your adapter.
+# Here is an example of the configuration:
```
josevalim, line 34:
```suggestion
# are not using SMTP. Here is an example of the configuration:
```

installer/templates/phx_single/config/runtime.exs
```diff
@@ -33,5 +33,27 @@ if config_env() == :prod do
...
+  # See `mix help release` for more information.<%= if @mailer do %>
+
+  # ## Configuring the mailer
+  #
+  # In production you need to configure the mailer to use a different adapter.
+  # Also, you may need to configure the Swoosh API client of your choice if you
+  # are not using SMTP.
+  #
+  # Additionally you may need to add more dependencies required by your adapter.
+  # Here is an example of the configuration:
```
josevalim, line 45:
```suggestion
  # are not using SMTP. Here is an example of the configuration:
```

josevalim:

princemaple:
I think I made swoosh too flexible that even adapter can be changed at runtime. So we have no way to know for sure from startup configuration whether api client is needed. 😂

PR #4361, opened 2021-07-05
Update request_lifecycle.md

kayuapi:
Update the wordings to make the concept clear.

### DIFF
--- a/guides/request_lifecycle.md
+++ b/guides/request_lifecycle.md
@@ -212,7 +212,7 @@ def show(conn, %{"messenger" => messenger}) do
 end
 ```
 
-Within the body of the `show` action, we also pass a third argument into the render function, a key-value pair where `:messenger` is the key, and the `messenger` variable is passed as the value.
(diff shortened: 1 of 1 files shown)


guides/request_lifecycle.md
```diff
@@ -212,7 +212,7 @@ def show(conn, %{"messenger" => messenger}) do
 end
 ```
 
-Within the body of the `show` action, we also pass a third argument into the render function, a key-value pair where `:messenger` is the key, and the `messenger` variable is passed as the value.
+Within the body of the `show` action, we also pass a third argument into the render function, a key-value pair where `messenger` is the key which is used in `show.html.eex` as `<%= @messenger %>`, and the `messenger` variable is passed as the value, taken from the second parameter (which is an Elixir map) of the `show` function.
```
anildigital, line 215: `: messenger` is the key. not `messenger`

kayuapi: What I meant is that when I wrote the following code:
```
def show(conn, %{"messenger" => messenger}) do
  render(conn, "show.html", messengerKey: messenger)
end
```
then the code in `show.html.eex` has to be changed to:
```
<div class="phx-hero">
  <h2>Hello World, from <%= @messengerKey %>!</h2>
</div>
```

kayuapi: Ok, I got it, my bad!

PR #4363, opened 2021-07-06
error in input tag usage

seasharps:


guides/realtime/channels.md
```diff
@@ -270,7 +270,7 @@ In `lib/hello_web/templates/page/index.html.eex`, we'll replace the existing cod
 
 ```html
 <div id="messages" role="log" aria-live="polite"></div>
-<input id="chat-input" type="text"></input>
+<input id="chat-input" type="text">
```
anildigital, line 273: It should be like this right?
```suggestion
<input id="chat-input" type="text"/>
```

josevalim:

PR #4366, opened 2021-07-10
Opinionated change regarding webpack

cw789:
- Update of mini-css-extract-plugin to v2.
- Reordered the webpack.config.js in a more "logical" order.?!
- Removed the entry point of `./vendor/**/*.js` but I'm unsure about it.
  Will phx.new created project have in any way a non empty "assets/vendor" directory?
  So I will probably also suggest to remove the creation of this directory.
  What are your thought about that?
- Again used 1.x version annotation where totally clear


josevalim:
Closing in favor of #4377, thanks!

PR #4368, opened 2021-07-13
Add mix phx.gen.notifier

philss:
This task generates email notifiers under a given context. It expects
that the app is configured with a "mailer" that is generated by `phx.new`.

PS: this feature was discussed before with José. It is related to https://github.com/phoenixframework/phoenix/pull/4355


philss:
ping @josevalim :)

lib/mix/tasks/phx.gen.notifier.ex
```diff
@@ -0,0 +1,176 @@
...
+    context: :boolean,
+    context_app: :string,
+    prefix: :string
+  ]
+
+  @default_opts [context: true]
+
+  alias Mix.Phoenix.Context
+
+  defmodule Notifier do
```
josevalim, line 30: We probably don't need this as a separate module? Private functions with a `notifier_` prefix should be enough. :)

lib/mix/tasks/phx.gen.notifier.ex
```diff
@@ -0,0 +1,176 @@
...
+
+        mix phx.gen.notifier Accounts User welcome reset_password
+
+    In this example the notifier will be called `UserNotifier` inside
+    the Accounts context. The functions `deliver_welcome/1` and
+    `reset_password/1` will be created inside this notifier.
+
+    The context serves as the API boundary for the given resource.
+    Multiple resources may belong to a context and a resource may be
+    split over distinct contexts (such as Accounts.User and Payments.User).
```
josevalim, line 150:
```suggestion
```

priv/templates/phx.gen.notifier/notifier.ex
```diff
@@ -0,0 +1,14 @@
+defmodule <%= inspect context.module %>.<%= inflections[:alias] %>Notifier do
+  import Swoosh.Email
+  alias <%= inspect context.base_module %>.Mailer<%= for message <- notifier_messages do %>
+
+  def deliver_<%= message %>(%{name: name, email: email}) do
+    new()
+    |> to({name, email})
+    |> from({"Phoenix Team", "team@phoenixframework.org"})
```
josevalim, line 8:
```suggestion
    |> from({"Phoenix Team", "team@example.com"})
```

lib/mix/phoenix/context.ex
```diff
@@ -46,6 +51,7 @@ defmodule Mix.Phoenix.Context do
       web_module: web_module(),
       basename: basename,
       file: file,
+      test_dir: test_dir,
```
josevalim, line 54: Do you need the test_dir because the test_file doesn't have the notifier suffix? What if you add the notifier suffix before you call `Context.new`?

philss: > Do you need the test_dir because the test_file doesn't have the notifier suffix? 

@josevalim the idea was to only get the context test dir, without the notifier. I changed to infer this from `test_file` now. WDYT? (I didn't get your suggestion though :thinking:)

lib/mix/tasks/phx.gen.notifier.ex
```diff
@@ -0,0 +1,167 @@
+defmodule Mix.Tasks.Phx.Gen.Notifier do
+  @shortdoc "Generates a notifier to send emails using the mailer"
+
+  @moduledoc """
+  Generates a notifier to send emails using the mailer
```
josevalim, line 5: This description is technically correct but it reads weird. It feels "notifier" is an empty word here and maybe we should go with `mix phx.gen.email` but I am not sure. The templates literally import `Swoosh.Email` though.

lib/mix/tasks/phx.gen.notifier.ex
```diff
@@ -0,0 +1,167 @@
...
+
+  @default_opts [context: true]
+
+  alias Mix.Phoenix.Context
+
+  @doc false
+  def run(args) do
+    if Mix.Project.umbrella?() do
+      Mix.raise(
+        "mix phx.gen.notifier must be invoked from within your *_web application root directory"
```
fuelen, line 34: Shouldn't email templates be separated from web?

josevalim: They are but the generators always run from web because thats where phoenix is defined.

lib/mix/tasks/phx.gen.notifier.ex
josevalim, line 5:
```suggestion
  @shortdoc "Generates a notifier that delivers emails by default"

  @moduledoc """
  Generates a notifier that delivers emails by default
```

jeregrine (approved):
This landed in a really good spot IMO. Thank you for doing the legwork on this.

josevalim:

PR #4369, opened 2021-07-14
Use Regex to extract token

benbarber:
The current token extraction method for the tests generated by `mix phx.gen.auth` fails if there is more than one occurrence of the token in the captured content. By using Regex to extract only the first occurrence of a token, the tests would still pass in the above case.

Using Regex seems a sensible default as the tests will now pass if the captured notification body is in either plain text or a HTML format requiring two occurrences of the token, like in a link for example. `<a href="[TOKEN]">[TOKEN]</a>`


LostKobrakai:
This could also be solved by using `[_, token, _] = String.split(captured.body, "[TOKEN]", parts: 3)` to limit the returned parts.

PR #4371, opened 2021-07-15
Doc:  Dockerfile intructions: copy lib folder before compiling assets.

sadjow:
I know that it is going to compile the assets again if we change the lib folder. But, when working with Tailwind(very popular) it needs to see the classes in the live view components before purging it on production.  I see it is a common problem with people trying to deploy tailwind with Live view to containers. https://pragmaticstudio.com/tutorials/adding-tailwind-css-to-phoenix

### DIFF
--- a/guides/deployment/releases.md
+++ b/guides/deployment/releases.md
@@ -192,12 +192,13 @@ COPY priv priv
 # your Elixir templates, you will need to move the asset compilation step
 # down so that `lib` is available.
 COPY assets assets
+COPY lib lib
 # use webpack to compile npm dependencies - https://www.npmjs.com/package/webpack-deploy
 RUN npm run --prefix ./assets deploy
 RUN mix phx.digest
 
 # compile and build the release
-COPY lib lib
+
 RUN mix compile
 # changes to config/runtime.exs don't require recompiling the code
 COPY config/runtime.exs config/
--- a/guides/deployment/releases.md
+++ b/guides/deployment/releases.md
@@ -198,7 +198,6 @@ RUN npm run --prefix ./assets deploy
 RUN mix phx.digest
 
 # compile and build the release
-
 RUN mix compile
 # changes to config/runtime.exs don't require recompiling the code
 COPY config/runtime.exs config/


sadjow:
Or perhaps, we could add instructions bellow.

guides/deployment/releases.md
```diff
@@ -192,12 +192,13 @@ COPY priv priv
...
 # down so that `lib` is available.
 COPY assets assets
+COPY lib lib
 # use webpack to compile npm dependencies - https://www.npmjs.com/package/webpack-deploy
 RUN npm run --prefix ./assets deploy
 RUN mix phx.digest
 
 # compile and build the release
-COPY lib lib
+
```
sadjow, line 201:
```suggestion
```

sadjow:
This is not a final solution. It is just some exposure of this need when working with Phoenix Live View + TailwindCSS.

guides/deployment/releases.md
```diff
@@ -192,12 +192,12 @@ COPY priv priv
 # your Elixir templates, you will need to move the asset compilation step
 # down so that `lib` is available.
 COPY assets assets
+COPY lib lib
```
vrcca, line 195: The instructions on line 190 already warns the dev when this change is needed. I wouldn't make it default, otherwise we would need to change the instructions to say the opposite. Wdyt?

sadjow: Sounds good. I did not see that comment and I lost some time trying to figure out why tailwind was not loaded. So the DX is not great. Perhaps, we could provide an alternative (ready to use) Dockerfile example. I think there are still people struggling about this. Like a section Dockerfile for PurgeCSS. I think this is good to provide some examples with some tools as Next.js does.

mcrumm:
As was pointed out, there are already instructions provided for folks who need to make this change when working with tools like Tailwind/PurgeCSS. The Dockerfile is provided purely for example purposes and it is understood that folks may need to make changes to fit their particular workflow. Since the comment is already there I think are good, thanks!

PR #4372, opened 2021-07-15
Switch to esbuild, package js as ESM

mveytsman:
Phoenix itself is only using webpack to compile javascript, we aren't shipping CSS or bundling any other assets. The built javascript is a module intended to be used in browser-side js code: it's preferable to package this as ESM. 

Replace webpack with esbuild for building phoenix.js and update our root package.json to expose phoenix.js as a ES module. I followed https://docs.skypack.dev/package-authors/package-checks#esm as a guideline for the package.json.

In my testing, existing client side code that is compiled with webpack and loads phoenix.js as `import {Socket} from "phoenix"` is not affected by this change. Tested by building the latest livebook off main.

This should be merged after https://github.com/phoenixframework/phoenix_live_reload/pull/112


mcrumm:
🔥

mveytsman:
We now build a IIFE version for being put in a script tag / CDNs and ESM for bundlers. I can confirm that livebook/npm7 work correctly with this branch.

priv/static/phoenix.esm.js.map
```diff
@@ -0,0 +1,7 @@
+{
+  "version": 3,
+  "sources": ["../../assets/js/phoenix/utils.js", "../../assets/js/phoenix/constants.js", "../../assets/js/phoenix/push.js", "../../assets/js/phoenix/timer.js", "../../assets/js/phoenix/channel.js", "../../assets/js/phoenix/ajax.js", "../../assets/js/phoenix/longpoll.js", "../../assets/js/phoenix/presence.js", "../../assets/js/phoenix/serializer.js", "../../assets/js/phoenix/socket.js"],
```
bcardarella, line 3: does esbuild provide a way to modify these paths? Chrome dev tools will parse these paths and group them according to the designated path hierarchy. So in dev tools you'll have to expand `..` then `..` then `assets` then `js` then `phoenix` to finally see the source file listings. If it is possible to re-write these during compilation time to just `phoenix/{filename}` this will be a better dev experience.

bcardarella:
I don't know if esbuild currently supports AMD output (issue from February: https://github.com/evanw/esbuild/issues/819) but if it does you may want to consider adding an AMD build as well. This way browsers that don't have complete or any module support  can still work. Browsers as recently as two years ago didn't have complete dynamic JS module import/export support: https://caniuse.com/es6-module-dynamic-import

mveytsman:
@bcardarella esbuild doesn't support amd unfortunately, but I included an iife build for CDNs. Do you think that solves the problem? Perhaps I should pick a the same target as we put in the generator...

mix.exs
```diff
@@ -82,7 +82,10 @@ defmodule Phoenix.MixProject do
 
       # Test dependencies
       {:phx_new, path: "./installer", only: :test},
-      {:websocket_client, git: "https://github.com/jeremyong/websocket_client.git", only: :test}
+      {:websocket_client, git: "https://github.com/jeremyong/websocket_client.git", only: :test},
+
+      # Dev dependencies
+      {:esbuild, "~> 0.1", path: "../esbuild/", only: :dev}
```
josevalim, line 88:
```suggestion
      {:esbuild, "~> 0.1", only: :dev}
```

joshchernoff:
I'd like to add that it would also be nice that we include source maps since enabling source maps on ESbuild will result in said error currently for LiveView. 

```
 > ../deps/phoenix_live_view/priv/static/phoenix_live_view.js:2:21: debug: Failed to read file "../deps/phoenix_live_view/priv/static/phoenix_live_view.js.map": open /Users/joshchernoff/Dev/elixir/elixir_group/deps/phoenix_live_view/priv/static/phoenix_live_view.js.map: no such file or directory
    2 │ //# sourceMappingURL=phoenix_live_view.js.map
      ╵                      ~~~~~~~~~~~~~~~~~~~~~~~~
 ```

**Comment by mveytsman** [comment]
@joshchernoff I thought we generate a sourcemap for the esm version which is what esbuild should be using -- is it not the case with the head of this branch?

**Comment by joshchernoff** [comment]
> @joshchernoff I thought we generate a sourcemap for the esm version which is what esbuild should be using -- is it not the case with the head of this branch?

Not sure what you mean but suing the ESM version. But my mix package is {:phoenix_live_view, "~> 0.15.1"},
and my app.js imports via `import { LiveSocket } from "phoenix_live_view"` and I'm building via 
```
watchers: [
    npx: [
      "esbuild",
      "./js/app.js",
      "--target=es2020",
      "--bundle",
      "--outdir=../priv/static/js",
      "--color=true",
      "--log-level=debug",
      "--sourcemap",
      "--watch",
      cd: Path.expand("../assets", __DIR__)
    ],
 ```
 
Using 
```
% ./assets/node_modules/.bin/esbuild --version
0.12.17
```

mveytsman:
https://github.com/phoenixframework/phoenix_live_view/pull/1533 should fix this.

chrismccord:
❤️❤️❤️🐥🔥

PR #4373, opened 2021-07-16
Add swoosh integration to phx.gen.auth generator

philss:
This commit adds the integration of the new mailer that uses Swoosh to
the existing `phx.gen.auth` generator. It also adds instructions for the
mailer installation if the app does not define the mailer module.


philss:
ping @josevalim :)

rickclare:
Perhaps the following is more of a discussion point than something that in the scope of this PR, but:

As the act of sending production emails (via Mailgun, Sendgrid, etc) can take significantly more time than regular Controller actions and Ecto operations, should any thought be given to using asynchronous email delivery as part of these new email generators? (e.g. as described in the Swoosh documentation https://github.com/swoosh/swoosh#async-emails).

In my Ruby-on-Rails days, we always found that sending email *synchronously* (eg during a user registration process) considerably degraded the user experience - HTTP requests that had an email-sending side effect were orders of magnitude slower, leading to user annoyance. On such projects we generally resorted to using solutions such as Redis-backed async queue (hopefully not necessary with Elixir).

I'm unsure of the best place location at which one might wrap mail delivery in something such as a `Task.async/1` call (eg whether at either Controller, Context or Mailer level?). Perhaps we can draw on advice from more seasoned Elixir engineers than myself on this?

josevalim:
If you are using Mailgun or Sendgrid, sending an email is relatively fast (< 100ms), as they don't block until the email is effectively sent. I would revisit this if using gen_smtp though. Plus, in Rails, you have to avoid blocking the request process as it may block other requests in the system, and that's not a concern here. So for now, my answer would be that they don't need to be made async. :)

rickclare:
That makes good sense (and is very reassuring). Thanks for the explanation Jose.

lib/mix/tasks/phx.gen.notifier.ex
```diff
@@ -49,12 +54,20 @@ defmodule Mix.Tasks.Phx.Gen.Notifier do
 
     prompt_for_conflicts(context)
 
-    copy_new_files(context, binding, paths)
+    if Keyword.get(test_opts, :compile_app?, true) do
+      # This is needed in order to detect if the mailer is defined in the project.
+      Mix.Task.run("compile")
+    end
```
josevalim, line 60: Instead of test_opts, do this:

```suggestion
    if "--no-compile" in args do
      Mix.Task.run("compile")
    end
```

lib/mix/tasks/phx.gen.notifier.ex
```diff
@@ -164,4 +177,44 @@ defmodule Mix.Tasks.Phx.Gen.Notifier do
...
+          end
+
+      It is also necessary to add "swoosh" as a dependency in your
+      "mix.exs" file:
+
+          def deps do
+            [{:swoosh, "~> 1.4"}]
+          end
+
+      An adapter needs to be set in your configuration:
```
josevalim, line 209:
```suggestion
      Finally, an adapter needs to be set in your configuration:
```

priv/templates/phx.gen.auth/notifier.ex
```diff
@@ -1,21 +1,29 @@
...
+      |> from({"MyApp", "contact@example.com"})
+      |> subject(subject)
+      |> text_body(body)
+      |> Mailer.deliver()
+
+    case result do
+      :ok -> {:ok, %{to: email, body: body}}
+      {:ok, _} -> {:ok, %{to: email, body: body}}
+      {:error, _} = error -> error
+    end
```
josevalim, line 19: Why so many different results? 🤔

josevalim: Btw, we should just return the whole email. We previously returned `%{to: email, body: body}` exactly so it matched the email fields. :)

lib/mix/tasks/phx.gen.notifier.ex
```diff
@@ -164,4 +177,44 @@ defmodule Mix.Tasks.Phx.Gen.Notifier do
...
+
+          use Mix.Config
```
wojtekmach, line 211: sorry if this was brought up already, can we replace Mix.Config with Elixir Config going forward? Or Phoenix itself (different than newly generated Phoenix app!) supports older version?

```suggestion
          import Config
```

josevalim: Good call, that's the default in new apps already.

ZombieHarvester:
@philss Why Swoosh and not Bamboo? I'm using Bamboo, I wonder why Swoosh is chosen, is it any better than Bamboo?

priv/templates/phx.gen.auth/notifier.ex
philss: Yeah, it was because the `Mailer.deliver/1` didn't have a clear API, so I was trying to match the standard result tuples. But I think it's fair to return just the result of the deliver :).

josevalim:
@ZombieHarvester they are pretty much equivalent from our point of view too. We had more experience with Swoosh, based on [Bytepack](https://github.com/dashbitco/bytepack_archive), so that's what we went with. :)

philss:
Sorry about the CI :disappointed: 
I always forget to run integration tests locally.

josevalim:
@philss don't worry about the integration tests :) I let the CI do the work most of the time for me :)

priv/templates/phx.gen.auth/notifier.ex
```diff
@@ -1,21 +1,26 @@
...
-    {:ok, %{to: to, body: body}}
+  # The result can be different based on the adapter,
+  # but will be `{:ok, term()} | {:error, term()}`.
+  defp deliver(email, subject, body) do
+    new()
+    |> to(email)
+    |> from({"MyApp", "contact@example.com"})
+    |> subject(subject)
+    |> text_body(body)
+    |> Mailer.deliver()
```
josevalim, line 16:
```suggestion
  defp deliver(recipient, subject, body) do
    email =
      new()
      |> to(recipient)
      |> from({"MyApp", "contact@example.com"})
      |> subject(subject)
      |> text_body(body)
    
    with {:ok, _metadata} <- Mailer.deliver(email) do
      {:ok, email}
    end
```

priv/templates/phx.gen.auth/context_fixtures_functions.ex
```diff
@@ -18,7 +18,7 @@
   end
 
   def extract_<%= schema.singular %>_token(fun) do
-    {:ok, captured} = fun.(&"[TOKEN]#{&1}[TOKEN]")
-    [_, token, _] = String.split(captured.body, "[TOKEN]")
+    {:ok, captured_email} = fun.(&"[TOKEN]#{&1}[TOKEN]")
+    [_, token, _] = String.split(captured_email.text_body, "[TOKEN]")
```
josevalim, line 22:
```suggestion
    [_, token | _] = String.split(captured_email.text_body, "[TOKEN]")
```

test/mix/tasks/phx.gen.auth_test.exs
```diff
@@ -89,7 +89,14 @@ defmodule Mix.Tasks.Phx.Gen.AuthTest do
       assert_file "lib/my_app/accounts.ex"
       assert_file "lib/my_app/accounts/user.ex"
       assert_file "lib/my_app/accounts/user_token.ex"
-      assert_file "lib/my_app/accounts/user_notifier.ex"
+      assert_file "lib/my_app/accounts/user_notifier.ex", fn file ->
+        assert file =~ "defmodule MyApp.Accounts.UserNotifier do"
+        assert file =~ "import Swoosh.Email"
+        assert file =~ "Mailer.deliver()"
```
josevalim, line 95:
```suggestion
        assert file =~ "Mailer.deliver(email)"
```

josevalim:

PR #4375, opened 2021-07-17
allow `node -r ...` in watchers

felix-starman:
This allows watchers that want to require a file inline, such as you might do if you are using `yarn`'s "Plug'n'Play" approach.

### DIFF
--- a/lib/phoenix/endpoint/watcher.ex
+++ b/lib/phoenix/endpoint/watcher.ex
@@ -59,6 +59,12 @@ defmodule Phoenix.Endpoint.Watcher do
 
   # We specially handle Node.js to make sure we
   # provide a good getting started experience.
+  # Although we assume folks know what they're doing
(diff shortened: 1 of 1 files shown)


felix-starman:
Alternatively, I guess, since people could be doing `--require` or other options, it might make more sense to just add some notes to either the docs or the error message regarding the use of the `NODE_OPTIONS` env. var.

mveytsman:
I think we should still validate the existence of node in this case.

Node's argument format is

```
node [options] [ script.js ] [arguments]
```
so there's definitely a bug here if we're just checking the first argument assuming its a script.

Im not sure we need to check the file exists at all - node will exit with a non-zero exit code, and we check for that above. 

I think we should cut the check for the file existing and add a better error message when the watcher process exits

josevalim:
We are removing this validation as we will be shipping with esbuild out of the box. Thanks!

PR #4381, opened 2021-07-19
Initial steps towards .heex

josevalim:
This just merges some of the html/live checks, but the bulk of the work (moving to .heex) and swapping of defaults still need to be done.


installer/templates/phx_assets/app.css
```diff
@@ -52,39 +86,4 @@
...
-}
-.alert:empty {
-  display: none;
-}
-.invalid-feedback {
-  color: #a94442;
-  display: block;
-  margin: -1rem 0 2rem;
-}
+}
```
josevalim, line 89:
```suggestion
}

```

Issue #4376, opened 2021-07-18
Add phx.gen.socket SocketModule


josevalim:
They have been removed from phx.new. It should generate the socket file and a .js file if assets exist. Channel and presence guides need to be updated.

josevalim:
At the end, this needs to be added to endpoint:

```
socket "/socket", <%= @web_namespace %>.UserSocket,
    websocket: true,
    longpoll: false
```

philss:
I can take this one :)

chrismccord:
We may also consider adding this functionality to `phx.gen.channel` which already exists and it can create the `UserSocket` (or flag for socket module name) if it doesn't exist w/ instructions for the endpoint? This approach gets bonus points too because we could add the first channel definition to the created socket module matching the channel module. Thoughts?

philss:
@chrismccord good point! I added a suggestion when the default `UserSocket` file does not exist. I liked the idea to create the socket file matching the channel name, but for each channel we would create a new socket file, right? I can change the task if you think this is preferable. Please let me know! :)

josevalim:
Closing this in favor of PR!

PR #4383, opened 2021-07-21
Add phx.gen.socket to generate a socket file

philss:
The socket generation was removed from the installer, so this task is needed
when the user wants to work with _websockets_ outside LiveView.

This also changes the `mix phx.gen.channel` task to include instructions regarding
this new task when the default socket - `user_socket.ex` does not exist under `/channels`.

Closes https://github.com/phoenixframework/phoenix/issues/4376.

### DIFF
--- a/lib/mix/tasks/phx.gen.channel.ex
+++ b/lib/mix/tasks/phx.gen.channel.ex
@@ -42,12 +42,31 @@ defmodule Mix.Tasks.Phx.Gen.Channel do
       {:eex, "channel_test.exs", Path.join(test_prefix, "channels/#{binding[:path]}_channel_test.exs")},
     ]
 
-    Mix.shell().info """
+    user_socket_path = Mix.Phoenix.web_path(context_app, "channels/user_socket.ex")
 
-    Add the channel to your `#{Mix.Phoenix.web_path(context_app, "channels/user_socket.ex")}` handler, for example:
+    instructions =
+      if File.exists?(user_socket_path) do
+        """
 
-        channel "#{binding[:singular]}:lobby", #{binding[:module]}Channel
-    """
+        Add the channel to your `#{user_socket_path}` handler, for example:
+
+            channel "#{binding[:singular]}:lobby", #{binding[:module]}Channel
+        """
+      else
+        """
+
+        The default socket handler - `#{binding[:web_module]}.UserSocket` - was not found
+        in its default location. To create it, please run the mix task:
+
+            mix phx.gen.socket User
+
+        Then add the channel to the newly created file, at `#{user_socket_path}`:
+
+            channel "#{binding[:singular]}:lobby", #{binding[:module]}Channel
+        """
+      end
+
+    Mix.shell().info(instructions)
   end
 
   @spec raise_with_help() :: no_return()
--- a/lib/mix/tasks/phx.gen.channel.ex
+++ b/lib/mix/tasks/phx.gen.channel.ex
@@ -22,12 +22,16 @@ defmodule Mix.Tasks.Phx.Gen.Channel do
 
   """
   use Mix.Task
+  alias Mix.Tasks.Phx.Gen
 
   @doc false
   def run(args) do
     if Mix.Project.umbrella?() do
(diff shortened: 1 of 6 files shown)


philss:
ping @josevalim :)

lib/mix/tasks/phx.gen.channel.ex
```diff
@@ -42,12 +42,31 @@ defmodule Mix.Tasks.Phx.Gen.Channel do
...
+        """
 
-        channel "#{binding[:singular]}:lobby", #{binding[:module]}Channel
-    """
+        Add the channel to your `#{user_socket_path}` handler, for example:
+
+            channel "#{binding[:singular]}:lobby", #{binding[:module]}Channel
+        """
+      else
+        """
```
josevalim, line 56: Let's ask the user if he wants us to create it for him. If he doesn't, then we print this error message. In fact, there is one cool thing, if we create the socket for the user, we can automatically add the channel line for it. :)

lib/mix/tasks/phx.gen.socket.ex
```diff
@@ -0,0 +1,94 @@
...
+  For a regular application:
+
+    * a socket in `lib/my_app_web/channels`
+    * a socket test in `test/my_app_web/channels`
+
+  For an umbrella application:
+
+    * a socket in `apps/my_app_web/lib/app_name_web/channels`
+    * a socket test in `apps/my_app_web/test/my_app_web/channels`
+
```
josevalim, line 22:
```suggestion
  For a regular application:

    * a client in `assets/js`
    * a socket in `lib/my_app_web/channels`
    * a socket test in `test/my_app_web/channels`

  For an umbrella application:

    * a client in `apps/my_app_web/assets/js`
    * a socket in `apps/my_app_web/lib/app_name_web/channels`
    * a socket test in `apps/my_app_web/test/my_app_web/channels`

```

priv/templates/phx.gen.socket/socket.js
```diff
@@ -0,0 +1,66 @@
...
+// you uncomment its entry in "assets/js/app.js".
+
+// To use Phoenix channels, the first step is to import Socket,
+// and connect at the socket path in "<%= web_prefix %>/endpoint.ex".
+//
+// Pass the token on params as below. Or remove it
+// from the params if you are not using authentication.
+import {Socket} from "phoenix"
+
+let socket = new Socket("/socket", {params: {token: window.userToken}})
```
josevalim, line 11:
```suggestion
// Bring in Phoenix channels client library:
import {Socket} from "phoenix"

// And connect to the path in "lib/web/endpoint.ex". We pass the
// token for authentication. Read below how it should be used.
let socket = new Socket("/socket", {params: {token: window.userToken}})
```

josevalim:
@philss this looks great! I have added two minor comments and one comment about automatically invoking the socket generator on channel.

lib/mix/tasks/phx.gen.socket.ex
philss: I changed in the commit faaf240e189e9b2622640460d1f067da597ca256

lib/mix/tasks/phx.gen.channel.ex
philss: Good point! I added the feature to ask the user. Please check again :)

lib/mix/tasks/phx.gen.socket.ex
```diff
@@ -0,0 +1,126 @@
...
+  @doc false
+  def run(args) do
+    if Mix.Project.umbrella?() do
+      Mix.raise(
+        "mix phx.gen.socket must be invoked from within your *_web application root directory"
+      )
+    end
+
+    [socket_name, pre_existing_channel] = validate_args!(args)
```
josevalim, line 36: Can we pass the channel under an option instead? Something like `--from-channel Room`. I want to avoid passing the channel as accident... unless we want to make it the channel argument a public API.

josevalim:

lib/mix/tasks/phx.gen.socket.ex
josevalim: I changed this locally!

PR #4388, opened 2021-07-26
Implement LiveAuth module for phx.gen.auth generator

henriquecf:
This builds upon what was discussed on #4344 

- Create LiveAuth module with mount_current_[name] and require_mounted_[name] functions.
- Documented the use cases for these function using the on_mount callback
- Create LiveAuthTest module to test the two functions
- Added tests to confirm file creations on many scenarios

I was thinking about injecting the LiveAuth module in the main Web module to be imported in :live_view modules. As I saw that AuthModule did not do it, I was unsure if this was desired.

Also, I believe the tests for require_mounted_[name] function can be improved, but not sure how:
- Did a direct check on socket.redirected to test redirection. Not sure if this is a good way to do it
- Confirming the flash by getting it in socket.assigns.flash. Also not sure this is a good way to do it

Please let me know if there is anything else I can do here.

### DIFF
--- a/priv/templates/phx.gen.auth/live_auth_test.exs
+++ b/priv/templates/phx.gen.auth/live_auth_test.exs
@@ -1,0 +1,47 @@
+defmodule <%= inspect live_auth_module %>Test do
+  use <%= inspect context.web_module %>.ChannelCase
+
+  alias <%= inspect context.module %>
+  alias <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web_namespace, schema.alias) %>LiveAuth
+  alias <%= inspect context.web_module %>.Router.Helpers, as: Routes
+  import <%= inspect context.module %>Fixtures
+
+  setup do
+    socket = %Phoenix.LiveView.Socket{
+      assigns: %{flash: %{}},
+      endpoint: <%= inspect context.web_module %>.Endpoint
+    }
+
+    %{<%= schema.singular %>: <%= schema.singular %>_fixture(), socket: socket}
+  end
+
+  describe "mount_current_<%= schema.singular %>/2" do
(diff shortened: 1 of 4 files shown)


priv/templates/phx.gen.auth/live_auth_test.exs
```diff
@@ -0,0 +1,47 @@
...
+  alias <%= inspect context.module %>
+  alias <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web_namespace, schema.alias) %>LiveAuth
+  alias <%= inspect context.web_module %>.Router.Helpers, as: Routes
+  import <%= inspect context.module %>Fixtures
+
+  setup do
+    socket = %Phoenix.LiveView.Socket{
+      assigns: %{flash: %{}},
+      endpoint: <%= inspect context.web_module %>.Endpoint
+    }
```
josevalim, line 13: We should not build this by hand, we probably need an API for this. /cc @chrismccord

josevalim (commented):
Thank you for the PR, I have added some comments.

priv/templates/phx.gen.auth/live_auth_test.exs
henriquecf: Thanks José. That's the main point really, but I was not able to find helper methods to build the socket, nor assert the flash or the redirection (the current ones require a view).

The option here would be to ask the use for a path that requires authentication, but that way we would have a failing test out of the box. My preference was to let it as a unit test, but we can make changes also.

henriquecf:
Was already implemented in #4810

henriquecf:
Was already implemented in https://github.com/phoenixframework/phoenix/pull/4810

PR #4392, opened 2021-07-29
Use latest esbuild, closes #4379

### LINKED ISSUE #4379: Address zombie processes on esbuild
A PR has already been sent. If it is not merged before release, we should change `esbuild` to provide a script and wrap the `esbuild` call on said script if the `--watch` flag is given.

josevalim:


josevalim:

PR #4394, opened 2021-08-01
Expand the plug only in plug_init_mode :runtime

wojtekmach:
Otherwise, in the :compile mode, the router is not recompiled if a *module* plug changes.

cc @josevalim

### DIFF
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -676,7 +676,12 @@ defmodule Phoenix.Router do
   See `pipeline/2` for more information.
   """
   defmacro plug(plug, opts \\ []) do
-    plug = Macro.expand(plug, %{__CALLER__ | function: {:init, 1}})
+    plug =
+      if Phoenix.plug_init_mode() == :runtime do
+        Macro.expand(plug, %{__CALLER__ | function: {:init, 1}})
+      else
+        plug
+      end
 
     quote do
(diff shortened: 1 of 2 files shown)


lib/phoenix/router.ex
```diff
@@ -676,7 +676,12 @@ defmodule Phoenix.Router do
   See `pipeline/2` for more information.
   """
   defmacro plug(plug, opts \\ []) do
-    plug = Macro.expand(plug, %{__CALLER__ | function: {:init, 1}})
+    plug =
+      if Phoenix.plug_init_mode() == :runtime do
```
wojtekmach, line 680: we have a couple other calls to this function https://github.com/phoenixframework/phoenix/blob/2e663e28934a81635e82c7e494fcb15d4e86e76a/lib/phoenix/router.ex#L534, https://github.com/phoenixframework/phoenix/blob/2e663e28934a81635e82c7e494fcb15d4e86e76a/lib/phoenix/router.ex#L647, the ets lookup is of course plenty fast but I wonder if conceptually it would make sense to set it just once in a module attribute and just refer to it.

wojtekmach: actually scratch that, it might be too restrictive if we want to test things and in some weird meta programming situations :)

josevalim:

PR #4400, opened 2021-08-05
Add --all flag to phx.digest.clean

mcrumm:
This PR allows `mix phx.digest` to be fully reversed by the `phx.digest.clean` task:

```sh
mix phx.digest.clean --all
```

When run with the `--all` flag, the following files are cleaned:

* All digested versions listed in the `cache_manifest.json` file (including latest).
* All compressed originals (e.g `robots.txt.gz`).
* The `cache_manifest.json` file itself.

### DIFF
--- a/lib/phoenix/digester.ex
+++ b/lib/phoenix/digester.ex
@@ -94,6 +94,10 @@ defmodule Phoenix.Digester do
     File.write!(Path.join(output_path, "cache_manifest.json"), manifest_content)
   end
 
+  defp remove_manifest(output_path) do
+    File.rm!(Path.join(output_path, "cache_manifest.json"))
+  end
+
   defp generate_digests(files) do
     Map.new(
       files,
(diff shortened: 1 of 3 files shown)


lib/phoenix/digester.ex
```diff
@@ -294,6 +298,36 @@ defmodule Phoenix.Digester do
...
+        for {_, versions} <- grouped_digests,
+            file <- Enum.map(versions, fn {path, _attrs} -> path end),
+            do: file
+
+      remove_files(files, path)
+      remove_compressed_files(logical_paths, path)
+      remove_manifest(path)
+      :ok
+    else
+      {:error, :invalid_path}
```
josevalim, line 327: I think we should just return `:ok`. If there is no manifest, it is because nothing was generated, which means `--all` was removed?

lib/phoenix/digester.ex
```diff
@@ -294,6 +298,36 @@ defmodule Phoenix.Digester do
...
+
+  ## Arguments
+
+    * `path` - The path where the compiled/compressed files are saved
+
+  """
+  @spec clean_all(String.t()) :: :ok | {:error, :invalid_path}
+  def clean_all(path) do
+    if File.exists?(path) do
+      %{"digests" => digests} = load_manifest(path)
```
josevalim, line 312: I was wondering if implementation wise we could simply call `clean()` with age of 0 and keep of 0? And the additional `--compressed` flag? But if this is simpler, then it is good for me too!

lib/phoenix/digester.ex
mcrumm: Ah, good point! :+1:

mcrumm: Okay, so I made `clean_all/1` a no-op, but I did so by checking map_size(digests) > 0 because we always get a map back from load_manifest.  Wdyt? If it's too noisy I can re-organize.

mcrumm: Ah, or we could just `rm` and not check for existence :) 🙈

josevalim (approved):

Issue #4358, opened 2021-07-02
Getting started guide raises an error: npm install, node-gyp python2 couldn't be found


adrien-mg:
### Environment

* Elixir version (elixir -v):

```
Erlang/OTP 24 [erts-12.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

Elixir 1.12.1 (compiled with Erlang/OTP 24)
```

* Phoenix version (mix deps):

```
* phoenix 1.5.9 (Hex package) (mix)
  locked at 1.5.9 (phoenix) 7e4bce20
* phoenix_ecto 4.3.0 (Hex package) (mix)
  locked at 4.3.0 (phoenix_ecto) 0ac49192
* phoenix_html 2.14.3 (Hex package) (mix)
  locked at 2.14.3 (phoenix_html) efd697a7
* phoenix_live_dashboard 0.4.0 (Hex package) (mix)
  locked at 0.4.0 (phoenix_live_dashboard) 8d52149e
* phoenix_live_reload 1.3.2 (Hex package) (mix)
  locked at 1.3.2 (phoenix_live_reload) 6d6f1d35
* phoenix_live_view 0.15.7 (Hex package) (mix)
  locked at 0.15.7 (phoenix_live_view) 3a756cf6
* phoenix_pubsub 2.0.0 (Hex package) (mix)
  locked at 2.0.0 (phoenix_pubsub) c52d948c
```

* Node.js version (node -v):

```
v16.4.1
```

* NPM version (npm -v):

```
7.17.0
```

* Operating system: 

Arch Linux, uname -a: 

```
Linux shash-mind 5.12.14-arch1-1 #1 SMP PREEMPT Thu, 01 Jul 2021 07:26:06 +0000 x86_64 GNU/Linux
```

### Expected behavior

The getting started guide doesn't fail in early steps.

### Actual behavior

Just started following the up and running guide on the phoenix documentation for version 1.5.9 at: https://hexdocs.pm/phoenix/up_and_running.html#content. 

Running the first step is fine:

```
mix phx.new antenna
```

... which logs, including instructions:

```
* creating antenna/config/config.exs
* creating antenna/config/dev.exs
* creating antenna/config/prod.exs
* creating antenna/config/prod.secret.exs
* creating antenna/config/test.exs
* creating antenna/lib/antenna/application.ex
* creating antenna/lib/antenna.ex
[...]

PR #4404, opened 2021-08-07
Remove node.js sass dependency from generated projects

udanieli:
This is related to #4358 and https://github.com/phoenixframework/phoenix/issues/4359#issuecomment-873135861. For a newcomer, being stopped at the first page of the manual by a node.js dependency nightmare is **really** frustrating.

### DIFF
--- a/installer/templates/phx_assets/package.json
+++ b/installer/templates/phx_assets/package.json
@@ -18,8 +18,6 @@
     "babel-loader": "^8.0.0",
     "copy-webpack-plugin": "^5.1.1",
     "css-loader": "^3.4.2",
-    "sass-loader": "^8.0.2",
-    "node-sass": "^4.13.1",
     "hard-source-webpack-plugin": "^0.13.1",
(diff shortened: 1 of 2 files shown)


Gazler:
Controversial branch name. :laughing: 

1.6 will be dropping soon which does not include sass.

udanieli:
> Controversial branch name. laughing

Well...how would you have named it?

> 
> 1.6 will be dropping soon which does not include sass.

Yes, this was already clear [in the last comment of #4359](https://github.com/phoenixframework/phoenix/issues/4359#issuecomment-873228353). But until that _soon_ some user will swear like a sailor at the first page of the manual :-(

chrismccord:
`useless` => `use less` :) 

We are working to get 1.6 out, so hang tight and smooth sailing soon!

udanieli:
> `useless` => `use less` :)

OMG

> We are working to get 1.6 out, so hang tight and smooth sailing soon!

Wait wait...can I say that too? 1.6 will be released soon! :-D

PR #4406, opened 2021-08-07
Remove duplicate background attribute in .phx-modal class

wingyplus:


a8t:
https://caniuse.com/css3-colors probably safe to remove

wingyplus:
I update the description of why I send a patch to remove it.

Issue #4347, opened 2021-06-12
Improving semantic html of default templates


vnglst:
Hi everyone!

I just started learning Phoenix using the `Programming Phoenix` book and I love it so far!

I'm a frontend developer by day and I noticed something in the html of the standard templates that could be improved. It correctly uses semantic html elements like `main` and `nav`, but it also specifies the implicit aria roles of `main` and `nav`. 

I don't think that is necessary anymore (according to the [spec](https://html.spec.whatwg.org/multipage/grouping-content.html#the-main-element) the `main` element has an implicit aria role and you should not set any role). Explicitly setting the role used to be a workaround for older browsers (released before 2013).

Do you also see this as an improvement and is this something I can help fix? I'd love to help!

Basically I would suggest making the following change for `main` and `nav`:

```diff
-    <main role="main" class="container">
+   <main class="container">
      <p class="alert alert-info" role="alert"><%%= get_flash(@conn, :info) %></p>
      <p class="alert alert-danger" role="alert"><%%= get_flash(@conn, :error) %></p>
      <%%= @inner_content %>
    </main>
```

josevalim:
Yes, please, a PR is welcome!

vnglst:
That’s great news! I’ll open a PR as soon as I get back from my holidays.

PR #4407, opened 2021-08-07
Remove implicit aria roles for navigation and main elements

vnglst:
Closes issue #4347

# Background

In the html templates the aria roles of `main` and `nav` are explicitly set. According to the spec both `main` and `nav` elements have an implicit aria role and you should not set any role. Explicitly setting the role used to be a workaround for older browsers (released before 2013, most notably IE11).

In the PR I basically make the following change for main and nav elements:

```diff
-   <main role="main" class="container">
+   <main class="container">
      <p class="alert alert-info" role="alert"><%%= get_flash(@conn, :info) %></p>
      <p class="alert alert-danger" role="alert"><%%= get_flash(@conn, :error) %></p>
      <%%= @inner_content %>
    </main>
```

Further reading on explicitly setting role="main" can be found in the MDN article: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/main#browser_compatibility

### DIFF
--- a/guides/howto/custom_error_pages.md
+++ b/guides/howto/custom_error_pages.md
@@ -77,7 +77,7 @@ Phoenix generates an `ErrorView` for us, but it doesn't give us a `lib/hello_web
   <body>
     <header>
       <section class="container">
-        <nav role="navigation">
+        <nav>
           <ul>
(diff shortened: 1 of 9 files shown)


mcrumm (approved):

josevalim:

PR #4411, opened 2021-08-13
Add phx_routes_router config for umbrella applications

TheFirstAvenger:
For those of us running umbrella applications, `mix phx.routes` requires manually entering the router every time. This small change recognizes a `:phoenix` config entry of `:phx_routes_router` so `mix phx.routes` works without requiring it to be specified each time. Manual passing of a router to the command still overrides this config.


chrismccord:
Perfect fit for a mix task alias, so no feature/code required on the phoenix side :)

TheFirstAvenger:
Thanks @chrismccord. That's a really simple solution... which I didn't think of at all 🤣 . Added #4417 to document this for others who similarly didn't think of it.

PR #4412, opened 2021-08-13
Include Phoenix as an external to the esbuild args

clone1018:
When creating a new project with the current `master` I was having an issue with the `--live` flag working as expected. The initial install / build would complete, and then on the first run of `mix phx.server` a dependency error would occur inside esbuild. 

```
 > js/app.js:26:21: error: Could not resolve "phoenix" (mark it as external to exclude it from the bundle)
    26 │ import {Socket} from "phoenix"
       ╵                      ~~~~~~~~~
```

My assumption here is that this is caused because Phoenix is a library, and is not natively included in the phoenix install. Adding this command line argument immediately fixed it.

If my assumption is incorrect, then ignore this PR and we can move it to a bug report instead :D!

### DIFF
--- a/installer/templates/phx_single/config/config.exs
+++ b/installer/templates/phx_single/config/config.exs
@@ -24,7 +24,7 @@ config :<%= @app_name %>, <%= @endpoint_module %>,
 config :esbuild,
   version: "0.12.18",
   default: [
-    args: ~w(js/app.js --bundle --target=es2016 --outdir=../priv/static/assets),
+    args: ~w(js/app.js --bundle --target=es2016 --outdir=../priv/static/assets --external:phoenix),
     cd: Path.expand("../assets", __DIR__),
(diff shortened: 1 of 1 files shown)


clone1018:
On retrospect, I'm not sure why `phoenix_live_view` does not trigger this as well. Maybe it's an issue when following the instructions here: https://github.com/phoenixframework/phoenix/blob/master/README.md#generating-a-phoenix-project-from-unreleased-versions

clone1018:
I just confirmed this only happens when creating an app using the installer. 

Does not work out of the box:
```
$ cd phoenix/installer
$ mix phx.new dev_app --dev --live
```

Does work out of the box:
```
$ cd installer
$ MIX_ENV=prod mix do archive.build, archive.install
$ mix phx.new dev_app --live
```

josevalim:
The proper fix here is to convert the import to a relative path that points to the Phoenix file in the parent directory. Only if —dev is given. Can you please update the PR?

PR #4417, opened 2021-08-16
Add documentation suggesting alias for umbrella phx.routes

TheFirstAvenger:
Added documentation suggesting an alias for those of us who use umbrella apps and didn't think to automate the Router specification. Result of #4411.


chrismccord:
❤️❤️❤️🐥🔥

PR #4419, opened 2021-08-21
Two step confirmation

josevalim:
Instead of confirming when the link in the email is clicked,
we explicitly require the user to press a button, that is to
avoid crawlers from automatically visiting emails and confirming
accounts.


lib/mix/tasks/phx.gen.auth.ex
```diff
@@ -226,6 +226,7 @@ defmodule Mix.Tasks.Phx.Gen.Auth do
       {:eex, "auth_test.exs", Path.join([web_test_prefix, "controllers", web_path, "#{schema.singular}_auth_test.exs"])},
```
josevalim, line 226: Users backporting this change to their apps can ignore the changes to this file.

test/mix/tasks/phx.gen.auth_test.exs
```diff
@@ -167,7 +167,8 @@ defmodule Mix.Tasks.Phx.Gen.AuthTest do
             delete "/users/log_out", UserSessionController, :delete
```
josevalim, line 167: Users backporting this change to their apps can ignore the changes to this file.

PR #4426, opened 2021-08-23
Removes unused module attribute @phoenix_log_level

grzuy:
Hi!

Apparently became unused after the deprecation of the instrument API: 19b0b01e2fc751901e06fcd74db1fc5b39672d26

---

Was looking at options to disable/change log level per route or controller and reading these lines of code made me think `use Phoenix.Controller, log: false` was still a valid option, when apparently no longer is. So hopefully cleaning this up avoids any future confusion for anyone :-)
Unless this is in fact used and I am missing something.

Thanks!


josevalim:

PR #4440, opened 2021-08-27
Update docker build in Deploying with Releases documentation

wingyplus:
In docker build instructuion should create homedir depends on $USER not
strict with `/home/elixir`.


josevalim:

PR #4444, opened 2021-08-30
log_out redirects to the / route

### LINKED ISSUE #4441: logout redirects to root of domain instead of root of app
Elixir 1.12.0 , phoenix 1.6.0-rc.0 , Ubuntu 20.04.1

When using phx.gen.auth and setting a url path in a config, such as

   `url: [host: "example.com", path: "/a/b", port: 80],`

successfully logging in, and then clicking logout, users get redirected to `example.com/` instead of `example.com/a/b/`

starkeepers:
on log_out, redirect to the / route (which includes a possible prefix path) rather than to the root of the domain. Closes #4441 . Thanks to @theMarinac for the fix.

### DIFF
--- a/priv/templates/phx.gen.auth/auth.ex
+++ b/priv/templates/phx.gen.auth/auth.ex
@@ -81,7 +81,7 @@ defmodule <%= inspect auth_module %> do
     conn
     |> renew_session()
     |> delete_resp_cookie(@remember_me_cookie)
-    |> redirect(to: "/")
+    |> redirect(to: Router.page_path(conn, :index))
   end
 
   @doc """


josevalim:
This is tricky because there is no guarnatee a page_path exists. I prefer to leave it hardcoded to / and people can change it accordingly. :) It is expected that you will customize the code after all!

josevalim:
Thanks!

starkeepers:
> This is tricky because there is no guarnatee a page_path exists. I prefer to leave it hardcoded to / and people can change it accordingly. :) It is expected that you will customize the code after all!

That was also my first reaction (see the issue), but there *is* actually a guarantee that page_path exists after phx.gen.auth has run, so the proposed fix is actually less surprising to a new user than hardcoding "/". I know I was thrown off when all of phoenix used my prefix-path correctly except the logout link. But of course, if you prefer to keep "/", no worries.

josevalim:
> but there is actually a guarantee that page_path exists after phx.gen.auth has run

I may be missing something, but how is this guaranteed? Anyone can remove the PageController and its routes from the router.

starkeepers:
Again, I actually removed that route from my app myself, and had the same answer as you did when I first saw this solution. But the fact remains (as far as I understand): when one runs a vanilla phx.gen.auth, one gets a PageController and its routes, and since this code would be part of what one gets with phx.gen.auth, one can count on them. (And more importantly, it improves the generators wrt the "principle of least surprise".) That said, I am new to this code and you're not, so I don't mean to push anything.

josevalim:
The PageController is part of a new app generator, not of the auth generator, which means someone may have removed it between commands. You can see all auth routes here: https://github.com/phoenixframework/phoenix/blob/master/priv/templates/phx.gen.auth/routes.ex

starkeepers:
Makes total sense, thanks. Supposing there were some sensible way to pull out the path value out of a config of the type:

```
config :app, AppWeb.Endpoint,
  url: [host: "example.com", path: "/a/b", port: 80],
  ...
```
perhaps based on what phoenix already does in its *_path helpers, would you be open to some future PR that uses this in the redirection after logout? That would be independent of any route, and I think more in line with what the rest of phoenix does. Of course, this approach runs the risk that the user has not set any route handling the root of their app, but that seems like such an edge case, that users setting up their app in that way can be expected to also know how to modify the logout redirect. The 99% of other users would have their redirect handled automatically without having to touch anything, even if their app is not at the root of their domain. No worries if that's also not a good option -- just probing whether there is a point to think further about this in case some time frees up.

PR #4445, opened 2021-08-30
Fix length error messages for password

HusseinMorsy:
because the password validation has changed to `count: bytes` we get the following error messages:

```
left:  "should be at least 12 character(s)"
right: ["should be at least 12 byte(s)"]
...
left:  "should be at most 72 character(s)"
right: ["should be at most 72 byte(s)"]
```


josevalim:
Thank you but the current tests are also valid. The bytes one is an additional validation and only if you are using bcrypt.

HusseinMorsy:
Thank you for the clarification.

PR #4446, opened 2021-08-30
Only validate bcrypt bytes if changeset is valid

josevalim:


josevalim:

PR #4447, opened 2021-08-30
Fix dockerfile

LostKobrakai:
This is the error I got on fly.io
```
Error error building: error rendering build status stream: When using COPY with more than one source file, the destination must be a directory and end with a /
```


mcrumm:
Thanks! 💚

PR #4450, opened 2021-08-30
skip CSRF check in WS session load if cookie is secure

azhi:
A follow-up to [this elixir forum thread](https://elixirforum.com/t/revisiting-accessing-session-cookie-in-phoenix-socket-connect-for-samesite-cookies/41601) to start a discussion.

In this implementation, cookie is considered to be 'secure' if it was set using `sameSite`
= `Lax` or `Strict`, `httpOnly` = `true` and `secure` = `true`.

We have no way to say for sure whether cookie was set with these opts - we are using data from `{:session, session_opts}` passed to `connect_info` option for socket transport.

Since phoenix encourages users to pass an exact copy of the arguments given to Plug.Session as session_opts, we can rely on this to get cookie options.

Note that this is not final implementation - need to change docs for Endpoint's `socket`, add entry to changelog, etc.

I will finish this if you are OK with general idea.

### DIFF
--- a/lib/phoenix/socket/transport.ex
+++ b/lib/phoenix/socket/transport.ex
@@ -270,10 +270,10 @@ defmodule Phoenix.Socket.Transport do
     key = Keyword.fetch!(session_config, :key)
     store = Plug.Session.Store.get(Keyword.fetch!(session_config, :store))
     init = store.init(Keyword.drop(session_config, [:store, :key]))
-    {key, store, init}
+    {key, store, init, session_config}
   end
 
-  defp init_session({_, _, _} = mfa)  do
+  defp init_session({_, _, _} = mfa) do
     {:mfa, mfa}
   end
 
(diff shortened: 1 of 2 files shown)


chrismccord:
Do you have a usecase which this helps support? On its own even if same site protects us against CSWSH (I need to do more esearch to verify this), I won't be convinced to avoid the csrf code path unless a usecase warrants it. Thanks!

azhi:
I'm working on a phoenix project that exposes JSON API (with a WS) for a separate frontend app, both hosted on same hostname.

As of now we are implementing our own JWT-token authorization for WS.
We would want to simplify things for frontend - it can use the same cookie it is using for HTTP API authentication.
For now the only option to do this is passing CSRF token in a separate HTTP API action, which defeats the purpose of it being a CSRF check anyway.

josevalim:
> We have no way to say for sure whether cookie was set with these opts

This is the issue with this approach. For example, imagine someone is building a Django application and they start using Phoenix for realtime and they port the Django session storage to Elixir. They can even set all of those flags, but there is no guarantee the Elixir backend set the cookies.

Or maybe someone can do a cookie injection in the browser, and now you are sending a cookie to the server that is actually not safe.

The only way we can accept WS cookies is if we verify them to be valid using the current mechanism. Any other mechanism is bound to bite someone at some point. If you don't want to use the session, then you can pass your own token in the headers or parameters. :)

PR #4452, opened 2021-08-31
Sync Diff function in Elixir just like JS

dewetblomerus:
I felt the need for this with a LiveView process that needed to keep in sync with Presence.

I was told there is a built-in function for this, after some searching, I discovered that the built-in function only exists in JS.


lib/phoenix/presence.ex
```diff
@@ -137,9 +137,9 @@ defmodule Phoenix.Presence do
 
   """
 
-  @type presences :: %{String.t => %{metas: [map()]}}
-  @type presence :: %{key: String.t, meta: map()}
-  @type topic :: String.t
+  @type presences :: %{String.t() => %{metas: [map()]}}
```
dewetblomerus, line 140: All the changes until the end of the file was `mix format`, I'm assuming we want to run that everywhere but I put it in a separate commit in case we don't.

josevalim:
I haven’t reviewed it yet (sorry) but I want to add a reminder that due to races we should discard joins of entries that we already know and discard the leaves of entries we don’t know. Both are possible given the list operation will always race updates.

chrismccord:
This would be a great addition and something I've recently been thinking about, so perfect timing :)

Have you thought about what the interface would look like to allow the user to hook into joins and leaves, similar to the JavaScript API? For example, for LiveView you likely want some kind of reduce operation that allows you to handle each join and leave of the sync_diff operation and accumulate specific socket state or `send_update` to a child component. I need to think a bit more about what an API could look like and play with some pseudo code, but will need to wait until next week. I'd love to hear your ideas since you're in this space! Thanks!

lib/phoenix/presence.ex
```diff
@@ -421,4 +430,64 @@ defmodule Phoenix.Presence do
...
+    Enum.reduce(leaves, state, fn leave, new_state ->
+      delete_meta_or_presence(new_state, leave)
+    end)
+    |> Map.merge(joins, fn _k, state_presence, joined_presence ->
+      %{metas: add_or_replace_metas(state_presence.metas, joined_presence.metas)}
+    end)
+  end
+
+  defp delete_meta_or_presence(state, {key, %{metas: left_metas}}) do
+    state_metas = Map.get(state, key, %{}) |> Map.get(:metas, [])
```
dewetblomerus, line 465: @josevalim you mentioned _"discard the leaves of entries we don’t know"_

That happens in this function.

This will be an empty list`[]` if we receive such a leave.

So we will perform two operations on it
```elixir
# 1. Try to remove it from an empty list
[] -- [%{meta: :we_dont_know}]

# 2. Try to delete it from the state Map.
Map.delete(state, :key_we_dont_know)
```

Should I do some kind of early return (probably another function with a pattern match) if we do not know the meta to make this more performant? Or is removing from an empty list and deleting non-existent keys from a Map already very performant?

lib/phoenix/presence.ex
```diff
@@ -421,4 +430,64 @@ defmodule Phoenix.Presence do
...
+
+    case remaining_metas do
+      [] -> Map.delete(state, key)
+      _ -> Map.put(state, key, %{metas: remaining_metas})
+    end
+  end
+
+  defp add_or_replace_metas(state_metas, joined_metas) do
+    Enum.filter(state_metas, fn state_meta ->
+      keep_existing_meta?(state_meta, joined_metas)
```
dewetblomerus, line 476: @josevalim you mentioned _"discard joins of entries that we already know"_

This happens inside of a `Map.merge/3`'s 3rd argument function. So it might feel like it is not very performant but this code only runs when there is a conflict, and the lists of metas should only be as big as how many devices a single user signed in from.

Before discarding anything, we need to be sure that it is the exact same entry that we already know.

It felt easier to discard our existing entries that match the incoming entry`phx_ref` at the same time as discarding entries that might not be identical but need to be replaced because they match the `phx_ref_prev` and then `++ joined_metas`.

dewetblomerus:
> This would be a great addition and something I've recently been thinking about, so perfect timing :)
> 
> Have you thought about what the interface would look like to allow the user to hook into joins and leaves, similar to the JavaScript API? For example, for LiveView you likely want some kind of reduce operation that allows you to handle each join and leave of the sync_diff operation and accumulate specific socket state or `send_update` to a child component. I need to think a bit more about what an API could look like and play with some pseudo code, but will need to wait until next week. I'd love to hear your ideas since you're in this space! Thanks!

Thanks for your encouragement 😄 
I have not thought any further than getting the whole list with `Presence.list` during `mount` [which I do here](https://gitlab.com/dewetblomerus/quickaverage/-/blob/main/lib/quick_average_web/live/average_live.ex#L19) and then keeping it updated with `Presence.sync_diff` every time the LiveView receives a `"presence_diff"` message [which I do here](https://gitlab.com/dewetblomerus/quickaverage/-/blob/main/lib/quick_average_web/live/average_live.ex#L113)

I don't feel or understand the need for any other API but would like to learn about it if someone does.

Also, I've never used the Javascript API, if it does something other than just patch the presence list, I'm willing to go take a look and see if I can implement that as well.

chrismccord:
We now have an elixir presence client in master, which uses a `handle_metas` callback. Thanks for the initial work on this!

Issue #4378, opened 2021-07-18
Convert phx.new templates to .heex and merge --live into --html


josevalim:

silathdiir:
Hi @josevalim and @chrismccord , I just wonder if I could try to contribute to this wonderful project by starting from the second task of this issue `Migrate all templates to .heex`. I couldn't confirm if you have started to work on this. Thanks.

PR #4454, opened 2021-08-31
Update CHANGELOG to cover phx.new changes

samgaw:
Include changes to `mix phx.new` generator, highlighting that `--live` is now the default and the home page has been reverted to a traditional controller  [#4378]


mcrumm:
Thanks! 💚

PR #4455, opened 2021-09-01
Update heroku.md to remove broken hook command

JonRowe:
The `esbuild` step as proposed in the Heroku guide is broken and produces the error:

```
The input path "&&" does not exist
```

As it seems to not be interpreting the string as a shell command. This in turn causes the boot error:

```
[error] Could not find static manifest at "/app/_build/prod/rel/<app>/lib/<app>-0.1.0/priv/static/cache_manifest.json". Run "mix phx.digest" after building your static files or remove the configuration from "config/prod.exs".
```

Simply dropping the cleanup step from build "solves" the problem, but maybe the cleanup needs to happen inside the generated mix task?

### DIFF
--- a/guides/deployment/heroku.md
+++ b/guides/deployment/heroku.md
@@ -105,7 +105,7 @@ erlang_version=24.0.3
 
 # Invoke assets.deploy defined in your mix.exs to deploy assets with esbuild
 # Note we nuke the esbuild executable from the image
-hook_post_compile="mix assets.deploy"
+hook_post_compile="eval mix assets.deploy && rm -f _build/esbuild"
 ```
 
 Finally, let's tell the build pack how to start our webserver. Create a file named `Procfile` at the root of your project:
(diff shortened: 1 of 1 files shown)


josevalim:
Hrm, I have deployed exactly with this config and it worked. I wonder if I was doing something special? Has anyone else ran into this?

josevalim:
Thanks for sharing btw!

JonRowe:
This was a fresh project with a single buildpack: `https://github.com/HashNuke/heroku-buildpack-elixir`

guides/deployment/heroku.md
```diff
@@ -105,7 +105,7 @@ erlang_version=24.0.3
 
 # Invoke assets.deploy defined in your mix.exs to deploy assets with esbuild
 # Note we nuke the esbuild executable from the image
-hook_post_compile="mix assets.deploy && rm -f _build/esbuild"
+hook_post_compile="mix assets.deploy"
```
josevalim, line 108:
```suggestion
hook_post_compile="eval mix assets.deploy && rm -f _build/esbuild"
```

josevalim:
Oh, I get what is missing, sorry about that.

josevalim:

JonRowe:
Fantastic thanks!

PR #4457, opened 2021-09-02
Improve esbuild plugins guide

karolsluszniak:
- recognize `--watch` and `--deploy` mode via `process.argv`
  (called via `mix cmd ...` so env vars are out)

- minify assets in deploy mode
  (just like default phx_new app does)

- enable inline sourcemaps in watch mode
  (just like default phx_new app does)

- just use `logLevel: 'info'`
  (equivalent to how CLI usage in default phx_new app behaves)

- add placeholder for loaders aside to the one for plugins
  (since it's a very common need with npm based fonts etc)

- update rest of the guide accordingly
  (list all script modes, watcher hook, assets.deploy alias)

### DIFF
--- a/guides/asset_management.md
+++ b/guides/asset_management.md
@@ -65,23 +65,44 @@ Next, add a custom Javascript build script. We'll call the example `assets/build
 ```js
 const esbuild = require('esbuild')
 
-const bundle = true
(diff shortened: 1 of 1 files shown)


josevalim:

PR #4458, opened 2021-09-03
Add --no-runtime-config option to esbuild

aswinmohanme:
Running `assets.deploy` in a CI/CD pipeline fails because `esbuild` reads the `config/runtime.exs` file which raises because the CI environment doesn't have access to the environment variables. The flag `--no-runtime-config` was added in https://github.com/phoenixframework/esbuild/commit/9779f2bc445dc2ae8413e0345bd35050f6cb9872 which would mitigate this issue.


josevalim:
Thanks for the PR!

I think as a default it is better to read the runtime config and if you need to read env vars on all environments, you can add this flag.

Although I am wondering if we should actually not load runtime config by default. Especially with Elixir v1.13 that won’t recompile the whole app when config changes. 🤔

josevalim:
esbuild installer 0.3.0 is out and it no longer loads the runtime config by default. Thank you!

aswinmohanme:
Updated to 0.3 and it works as expected. Thanks for making the change and making all this possible.

Issue #3906, opened 2020-06-18
Compilation error in file installer/lib/phx_new/ecto.ex


eksperimental:
Running `mix test` generates a compilation error in one of the tests

### Environment

* Erlang version: Erlang/OTP 23 [erts-11.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [hipe]
* Elixir version (elixir -v): Elixir 1.10.3 (compiled with Erlang/OTP 23)
* Phoenix version (mix deps): master
```
mix deps:
* ranch 1.7.1 (Hex package) (rebar3)
  locked at 1.7.1 (ranch) 451d8527
  ok
* telemetry 0.4.1 (Hex package) (rebar3)
  locked at 0.4.1 (telemetry) 4738382e
  ok
* jason 1.0.0 (Hex package) (mix)
  locked at 1.0.0 (jason) b96c400e
  ok
* phoenix_pubsub 2.0.0 (Hex package) (mix)
  locked at 2.0.0 (phoenix_pubsub) c52d948c
  ok
* cowlib 2.8.0 (Hex package) (rebar3)
  locked at 2.8.0 (cowlib) 79f954a7
  ok
* cowboy 2.7.0 (Hex package) (rebar3)
  locked at 2.7.0 (cowboy) 04fd8c6a
  ok
* mime 1.3.1 (Hex package) (mix)
  locked at 1.3.1 (mime) 6cbe761d
  ok
* plug_crypto 1.1.2 (Hex package) (mix)
  locked at 1.1.2 (plug_crypto) 6b8b608f
  ok
* plug 1.10.3 (Hex package) (mix)
  locked at 1.10.3 (plug) 01f9037a
  ok
* plug_cowboy 2.3.0 (Hex package) (mix)
  locked at 2.3.0 (plug_cowboy) bc595a18
  ok
* phoenix_html 2.14.2 (Hex package) (mix)
  locked at 2.14.2 (phoenix_html) 58061c8d
  ok
```
* NodeJS version (node -v): 6.14.5
* NPM version (npm -v): v14.4.0
* Operating system: Arch Linux 64 bits

### Expected behavior
Not to print any errors

### Actual behavior
While all the tests pass, there is a compilation error that gets printed out.

```
...
[...]

PR #4462, opened 2021-09-04
Remove archive phx_new during compilation

MaSven:
This was an issue raised in #3906 and closed in 496627f. But the test file that fixed this was removed. This commit also fixes the compilation error for all environments. Jose mentioned it in the issue that the line should be moved to somewhere else. 

I don't know if this is the right place for the line.


MaSven:
My mistake this was already resolved in b097e816e10bba969d0a3e167caf21631d71ecb8

PR #4468, opened 2021-09-07
No secret key base fallback

### LINKED ISSUE #4465: `secret_key_base`, `signing_salt` in version control considered harmful
Howdy Phoenix!

I was in the middle of writing a tutorial about implementing OAuth and session persistence in a Phoenix application, when I noticed something a little ... silly, I think is the right word for it.

I won't beat around the bush, as this issue ticket's title has already gotten down to brass tacks: `config/config.exs` and `/lib/app_web/endpoint.ex` are _terrible_ places to make the default homes for `AppWeb.Endpoint`'s `:secret_key_base` and `:signing_salt`! There's all kinds of opportunity for these keys to unwittingly enter the codebase, and having it be in there as early as the very first `git commit`. I didn't even realize I had committed `secret_key_base` to my own repo until I'd already made 30 commits on `main`!

I think the simplest way to go about fixing this issue is to migrate the secret values into `/config/dev.secret.exs`. Here's how I've implemented it in my own project:

```elixir
# /config/config.exs
import Config

env = config_env()

...

# Import environment specific config. This must remain at the bottom
# of this file so it overrides the configuration defined above.
import_config "#{env}.exs"

with secret_config <- "#{env}.secret.exs",
  true <- File.exists?("#{__DIR__}/#{secret_config}"),
  do: import_config(secret_config)
```

Then, in `.gitignore`:

```ignore
...
# anything with `.secret` as an extension
*.secret.*
*.secret
...
```

The ignore rule used to be `/config/*.secret.exs`, but then I thought, why limit the scope of an extension I anticipate _only_ using for files I want to keep out of version control?

---

I don't know how important this ranks on the Phoenix team's priorities, but it just feels like an easy thing to correct that'll have reasonable payoff in the short and long-term. Not saying that any other framework is the thing Elixir should strive to become, but it's a strong enough security concern that Rails generates a `/config/master.key` inside of new projects, and immediately adds it to the project's `.gitignore`. It seems reasonable to have similar defaults within new Phoenix apps. Plus, since Elixir releases only contain compiled .beam files, `{env}.secret.exs` config files come with a much greater degree of security, as their source code is fundamentally never needed for an app after the build phase.

Anyway, that's all I got this time. Phoenix 1.6.0 is pretty spiffy so far! I'd already split out Sass's behavior from Webpack in 1.5, and am so happy to finally have an uncomplicated asset pipeline with it and esbuild.

josevalim:
Closes #4465.


chrismccord:
❤️❤️❤️🐥🔥

PR #4470, opened 2021-09-08
Update `Phoenix.Socket` docs to clarify defaults

joshprice:
`Phoenix.Endpoint.socket/3` sets `websocket` on by default and `longpoll` is off by default. Previous phrasing implied that both were on by default.

https://hexdocs.pm/phoenix/Phoenix.Endpoint.html#socket/3

`:longpoll` - controls the longpoll configuration. Defaults to false

https://github.com/phoenixframework/phoenix/blob/v1.5.12/lib/phoenix/endpoint.ex#L649-L650


josevalim:

PR #4471, opened 2021-09-10
Set argon2 as default hashing algorithm for mix.gen.auth

goncalotomas:
Argon2 is now recommended as the go-to algorithm by OWASP, being classified as modern and secure hashing algorithm. The [OWASP Password Storage page](https://cheatsheetseries.owasp.org/cheatsheets/Password_Storage_Cheat_Sheet.html) states that the first choice should be Argon2, listing bcrypt as a second option when Argon2 is not available.

This Pull Request sets Argon2 as a default algorithm used in `mix.gen.auth` and updates the test suite accordingly.

I decided not to update the default option for Windows as I am unaware of the stability of the Argon2 library on that Operating System.

I tried to follow the contribution guidelines but seeing as this is my first PR to Phoenix I will be happy to change anything that is not as expected.

### DIFF
--- a/lib/mix/tasks/phx.gen.auth.ex
+++ b/lib/mix/tasks/phx.gen.auth.ex
@@ -14,7 +14,7 @@ defmodule Mix.Tasks.Phx.Gen.Auth do
 
   ## Password hashing
 
-  The password hashing mechanism defaults to `bcrypt` for
+  The password hashing mechanism defaults to `argon2` for
   Unix systems and `pbkdf2` for Windows systems. Both
   systems use the [Comeonin interface](https://hexdocs.pm/comeonin/).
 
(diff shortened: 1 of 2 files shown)


josevalim:

chrismccord:
We may need to revert this as some users are experiencing OOM failures on smaller instances. It seems argon2 is enough to OOM 256MB instances easily, so I think unfortunately it's a poor default when bcrypt remains a great option.

ref: https://elixirforum.com/t/staging-environment-how-to-debug-out-of-memory-errors-in-production-on-fly-io/42763

I will touch base with the team tomorrow and cut a new release if everything checks out.

josevalim:
I have reverted it on master but added notes that users should consider using argon2. :+1:

PR #4482, opened 2021-09-12
Mention fly's remote builder on deployment guide

mauroveron:
I ran into some issues when trying to deploy phoenix 1.6-rc0 to fly from a Mac M1 (Apple Silicon)

The problem seems to be with building the docker image locally. On Apple Silicon computers, Docker uses qemu as an emulation layer to build Linux platform images which seems to cause issues.

I solved the problem by using fly's remote builder (just using `fly deploy --remote-only`) and thought it would be good to update the docs in case someone else runs into the same issue.

### DIFF
--- a/guides/deployment/fly.md
+++ b/guides/deployment/fly.md
@@ -230,6 +230,19 @@ Our project is now ready to be deployed to Fly.io.
 $ fly deploy
 ```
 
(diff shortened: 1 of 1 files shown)


roughike:
+1 for this.

I have an M1, and I almost gave up with the deployment process as it's really flaky and fails most times. `--remote-only` fixes it.

chrismccord:
❤️❤️❤️🐥🔥

PR #4483, opened 2021-09-14
Fix LongPoll readyState check

mcrumm:
The `Ajax` class is never instantiated, so `this.states` would fail with the following error:
Screen Shot 2021-09-14 at 10 26 26 AM

To fix, I simply moved the ready states into a constant and check from there :)


jeregrine (approved):

chrismccord:
❤️❤️❤️🐥🔥

PR #4491, opened 2021-09-19
doc: Add dependency documentation for Swoosh

dkarter:
Considering that the name of the extension is not very indicative of its purpose and even if you do know what it is you may be wondering why it's included by default.

I had to do some light Github spelunking, but figured I'd save some time for the next person.

The list of dependencies generally seems to be out of date, this PR chips away at it :)

### DIFF
--- a/lib/phoenix.ex
+++ b/lib/phoenix.ex
@@ -12,6 +12,9 @@ defmodule Phoenix do
 
     * [ExUnit](https://hexdocs.pm/ex_unit) - Elixir's built-in test framework
 
+    * [Gettext](https://hexdocs.pm/gettext) - Internationalization and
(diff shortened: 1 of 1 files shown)


lib/phoenix.ex
```diff
@@ -35,12 +38,11 @@ defmodule Phoenix do
     * [Plug](https://hexdocs.pm/plug) - a specification and conveniences
       for composable modules in between web applications
 
+    * [Swoosh](https://hexdocs.pm/swoosh) - email delivery for `phx.gen.auth`
```
josevalim, line 41:
```suggestion
    * [Swoosh](https://hexdocs.pm/swoosh) - a library for composing,
    delivering and testing emails, also used by `mix phx.gen.auth`
```

josevalim:

PR #4492, opened 2021-09-21
Create phx.gen.ex

BarbaricCorgi:
This task describe in detail what every generator do. specially the differences between the resource related tasks

### DIFF
--- a/lib/mix/tasks/phx.gen.ex
+++ b/lib/mix/tasks/phx.gen.ex
@@ -1,0 +1,50 @@
+defmodule Mix.Tasks.Phx.Gen do
+  use Mix.Task
+
+  @shortdoc "Explains in detail what every phx.gen task do"
+
+  @moduledoc """
+    ## Explanation of the different types of phoenix generators
+
+    If you're creating an entity, you can decide whether you need to validate it, store it in db, have some helpers in the form of an API boundary (repository helper methods), have an HTML crud, expose a REST JSON API, have a LiveView index/show. depending in which one you need, you may select:
+    - `phx.gen.schema` : Generates a db backed entity (includes migration, schema)
+    - `phx.gen.embedded` : Generates an schema just for validation (includes schema, no migration)
+    - `phx.gen.context` : Generates a context a.k.a. API boundary (includes migration, schema and context )
+    - `phx.gen.html` : Generates a HTML CRUD (includes migration, schema, context, controller, view and html CRUD templates)
+    - `phx.gen.json` : Generates a REST JSON API (includes migration, schema, context, controller and view)
+    - `phx.gen.live` : Generates a LiveView resource (includes migration, schema, live view index/show, live component form/modal and view)
+
+    Other generators:
+    - `phx.gen.auth` : Generates authentication logic for a resource.
+    - `phx.gen.presence` : Generates a Presence tracker
+    - `phx.gen.channel` : Generates a Phoenix channel and its corresponding tests
+    - `phx.gen.notifier` : Generates a notifier that delivers emails by default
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/phx.gen.ex
```diff
@@ -0,0 +1,50 @@
...
+    phx.gen.auth : Generates authentication logic for a resource.
+    phx.gen.presence : Generates a Presence tracker
+    phx.gen.channel : Generates a Phoenix channel and its corresponding tests
+    phx.gen.notifier : Generates a notifier that delivers emails by default
+    phx.gen.socket : Generates a Phoenix socket handler
+
+    Static generators:
+    phx.gen.cert : Generates a self-signed certificate for HTTPS testing
+    phx.gen.secret : Generates a random key 64 characters long
+    "
```
josevalim, line 48:
```suggestion
    Mix.Task.run("help", ["--search", "phx.gen."])
```

josevalim: The above should achieve a similar result, without requiring us to remember to keep the task up to date. WDYT?

lib/mix/tasks/phx.gen.ex
```diff
@@ -0,0 +1,50 @@
+defmodule Mix.Tasks.Phx.Gen do
+  use Mix.Task
+
+  @shortdoc "Explains in detail what every phx.gen task do"
```
josevalim, line 4:
```suggestion
  @shortdoc "Lists all available Phoenix generators"
```

lib/mix/tasks/phx.gen.ex
```diff
@@ -0,0 +1,50 @@
...
+    Other generators:
+    - `phx.gen.auth` : Generates authentication logic for a resource.
+    - `phx.gen.presence` : Generates a Presence tracker
+    - `phx.gen.channel` : Generates a Phoenix channel and its corresponding tests
+    - `phx.gen.notifier` : Generates a notifier that delivers emails by default
+    - `phx.gen.socket` : Generates a Phoenix socket handler
+
+    Static generators:
+    - `phx.gen.cert` : Generates a self-signed certificate for HTTPS testing
+    - `phx.gen.secret` : Generates a random key 64 characters long
```
josevalim, line 26:
```suggestion
  Lists all available Phoenix generators.
```

lib/mix/tasks/phx.gen.ex
BarbaricCorgi: yes and we will keep the tasks updated. great choice

lib/mix/tasks/phx.gen.ex
BarbaricCorgi: I use hexdocs a lot and considering the mix tasks help lists all the commands alphabetically, I had to read every generator one by one to realize how many of them do the same as the previous one plus a little bit more (which is awesome)

so i though, i should improve the documentation (in fact, now i think i should use a markdown table showing the relationship between the CRUD related tasks) and when I went for it I realized I needed to modify Phoenix tasks to achieve it. and considering this is a cross-cutting bit of information, I ended up suggesting this tasks.

In that sense, your idea of automating the listing of tasks is great, I would only add a table describing the CRUD related tasks by order of complexity. will commit them in a minute.

BarbaricCorgi (commented):
@josevalim please, may you check this change? do you think documentation may improve this way

this is my first contribution to the project, so your guidance is very enlightening for me

josevalim:
@BarbaricCorgi that's a great idea, i just pruned the table a little bit :)

josevalim:

PR #4501, opened 2021-09-25
guides: update 'request life-cycle' page

milmazz:
This PR updates the `pipeline` structure in `HelloWeb.Router` based on the `mix phx.new hello` output using v1.6.0. 

This PR also removes the (ab)use of the _simply_ word on the same page. I would suggest avoiding saying that things "are easy" or that they're not challenging, everyone learns at its own pace, and that's fine. See https://justsimply.dev/ for more examples.


cw789 (approved):

Gazler:
Thanks <3

PR #4504, opened 2021-09-25
Update releases.md

arbaaz:
Fixes
(File.Error) could not read file "/app/config/prod.secret.exs"

image


cw789:
I think `config/prod.secret.exs` isn't anymore a default of gernerated Phoenix projects starting with v1.6.0 (in preference of `config/runtime.exs` - doesn't mean you should put secrets directly in there). So it's not listed in the latest docs.

PR #4510, opened 2021-09-28
guides: update 'Controllers' guide

milmazz:
This PR includes the following changes:

* Replace `put_layout/2` with `put_root_layout/2` in the _Assigning
  Layouts_ section, so, that way the sample works as expected.
* Update the code examples to match Phoenix v1.6.0
* Reduce the verbosity in some links, instead we use the _autolink_
  feature from ExDoc (See: https://github.com/elixir-lang/ex_doc/#auto-linking)
* Sort the link references at the end of the document


josevalim:

Issue #4506, opened 2021-09-25
latest pheonix package can not be used in node.js


wangshijun:
### Environment

* Elixir version (elixir -v): Elixir 1.9.4 (compiled with Erlang/OTP 22)
* Phoenix version (mix deps): Not Applicable
* Operating system: MacOS

### Expected behavior

https://www.npmjs.com/package/phoenix can work in node.js on 1.5.x, it should work in node.js for 1.6 release.

### Actual behavior

but when 1.6.0 is out, the node.js runtime throws an error:

```
(node:2076) UnhandledPromiseRejectionWarning: Error [ERR_PACKAGE_PATH_NOT_EXPORTED]: No "exports" main defined in /usr/local/lib/node_modules/@abtnode/cli/node_modules/phoenix/package.json
```

jamesvl:
I believe I'm having a similar issue when updating to Phoenix 1.6. An existing project that is still using brunch (it started back in the early Phoenix days and I haven't had a need to change to webpack) can no longer compile the Phoenix JS asset, when previously it did for 1.5.x.

### Environment
```
elixir          1.12.3-otp-24
erlang          24.1
nodejs          16.9.1
rebar           3.16.1
phoenix         1.6.0
```

npm package versions for Phoenix are

* phoenix_html@3.0.4 -> ./../deps/phoenix_html
* phoenix_live_view@0.16.4 -> ./../deps/phoenix_live_view
* phoenix@1.6.0 -> ./../deps/phoenix

The error message plus JS stack trace is

```
$LOGGY_STACKS=1 npm run deploy

> deploy
> brunch build --production

08:00:07 - error: Processing of js/app.js failed. Error: Could not load module 'phoenix' from '/ex/assets/js'. Possible solution: run `npm install`.
    at friendlyRequireError (/ex/assets/node_modules/deppack/lib/resolve.js:69:9)
    at /ex/assets/node_modules/deppack/lib/resolve.js:90:17
    at /ex/assets/node_modules/browser-resolve/index.js:265:24
    at /ex/assets/node_modules/browser-resolve/node_modules/resolve/lib/async.js:46:14
    at process (/ex/assets/node_modules/browser-resolve/node_modules/resolve/lib/async.js:173:43)
    at ondir (/ex/assets/node_modules/browser-resolve/node_modules/resolve/lib/async.js:188:17)
    at load (/ex/assets/node_modules/browser-resolve/node_modules/resolve/lib/async.js:69:43)
[...]

PR #4512, opened 2021-09-28
Build CommonJS module along with ESM and IIFE

hubertlepicki:
I honestly don't know if this is the best approach but I am quite stuck with trying to upgrade to Phoenix 1.6 on a fairly large project that uses "Jest" for testing of their JavaScript code (elixirforum thread here: https://elixirforum.com/t/phoenix-1-6-0-npm-package-cannot-be-imported-in-jest/42670/4)

I tried various hacks and third party tools suggested on Jest thread but I wasted a whole day and fixing one thing breaks another, typical Node.js developer story. So, I decided to build CJS package of phoenix.js that Jest can understand like it used to up to 1.5.9.

Now I suspect there was a good reason for not including CJS file, but maybe it's just an omission - for the later case I provide this PR below. Note: in this project "phoenix" is the only package that doesn't provide CJS fallback, many other packages do ship with CJS and EMS and I tried to follow the pattern here.

esbuild can build CJS modules as well as ESM, and for the people using
build tools / or testing tools (such as Jest) that do not yet support
ESM modules, it is useful to have fallback when ESM is not availalbe.

This PR includes command to build CJS module along with currently
supported types, and adds entry "main" in package.json to point towards
that file.

This has been tested with Jest, which currently has very poor support
for ESM modules and allows for including "phoenix" in tests and also
stubbing it if needed.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -20,6 +20,7 @@ if Mix.env() == :dev do
   config :esbuild,
     version: "0.12.18",
     module: esbuild.(~w(--format=esm --sourcemap --outfile=../priv/static/phoenix.esm.js)),
+    main: esbuild.(~w(--format=cjs --sourcemap --outfile=../priv/static/phoenix.cjs.js)),
     cdn:
       esbuild.(
         ~w(--target=es2016 --format=iife --global-name=Phoenix --outfile=../priv/static/phoenix.js)
--- a/mix.exs
+++ b/mix.exs
@@ -198,7 +198,7 @@ defmodule Phoenix.MixProject do
   defp aliases do
     [
       docs: ["docs", &generate_js_docs/1],
-      "assets.build": ["esbuild module", "esbuild cdn", "esbuild cdn_min"],
+      "assets.build": ["esbuild module", "esbuild cdn", "esbuild cdn_min", "esbuild main"],
       "assets.watch": "esbuild module --watch"
     ]
   end
--- a/package.json
+++ b/package.json
@@ -4,6 +4,7 @@
   "description": "The official JavaScript client for the Phoenix web framework.",
   "license": "MIT",
   "module": "./priv/static/phoenix.esm.js",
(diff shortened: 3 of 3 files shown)


mveytsman:
This looks good by me, although the two compiled files in priv/static should be taken out of this PR and built as part of the release cycle.

hubertlepicki:
@mveytsman just removed those from the PR.

hubertlepicki:
So @mveytsman I've been thinking of it and I'm not sure if this should be built as part of release cycle or earlier - when this PR is being merged. If we did the later, then the CJS file could be used intermediately by folks pointing to GH repo master branch, and that'd be some extra testing + we have copies of the other versions of phoenix.js already commited into master as well, so it'd be consistent. Either way it's fine by me.

mcrumm:
@mveytsman does this fix #4506?

@hubertlepicki The assets will be built when the PR is merged. Max just meant that, as a security measure, the build needs to be done by someone on the Phoenix team :)

hubertlepicki:
Glancing at it, it may be that these are 3 separate but related issues, and
yes, Node.js does not natively support ESM paclages (yet there is
experimental flag to turn it on). I think at least second person's issue
should be reaolved as I was getting the same error as @jamesvl but with
"jest" not "brunch". Webpack handles those natively, and we already
upgraded to webpack while ago so it wasn't the issue.

hubertlepicki:
@mcrumm I wonder if I (or you if you want to do it yourself) should prepare a branch to run by those folks and see if their issues are resolved when pointing to that branch? This is, if this PR is not going to be merged to master shortly, otherwise they can just point to master to test.

mveytsman:
@hubertlepicki yeah I agree - the file doesn't only need to be built on releases. I more wanted to emphasize that we should keep changes to built artifacts out of PRs like this and as part of an action taken by a team member.

@mcrumm I have not tested but I believe that should fix that!

mcrumm:
Thanks! 💚💚💚

PR #4513, opened 2021-09-28
guides: Update documented server console log

faheempatel:
This PR includes the following changes:
- Update the documented server log to remove mention of Webpack and replace it with one which is more accurate


mcrumm:
Thanks! 💚

PR #4528, opened 2021-10-03
Fix doc semantic error

qrede:
Fix doc semantic error


josevalim:
Hi @qrede! Apparently both versions are correct, so we should probably keep this as is? https://english.stackexchange.com/questions/25742/is-it-supposed-to-be-a-html-or-an-html

Thank you!

PR #4529, opened 2021-10-04
small fixup in the channel guide

ConaGo:
Hi, i noticed the html as shown in the guide would not compile. So I changed the docu to use the {}-notation and also changed the path from js/app.js to assets/app.js


josevalim:

PR #4531, opened 2021-10-04
Fix Dockerfile in guides/deployment/releases.md

onixus74:
The current `Dockerfile` example seems to be not working out of the box. 

This is due to the `COPY` command which copies `my_app/` release directory instead of its content, and when referenced in the `ENTRYPOINT`, it was expecting `bin/` folder.

To fix this I added `/*` to reference the directory content like in this PR

OR

We could reference the directory in `ENTRYPOINT`
```docker
ENTRYPOINT ["my_app/bin/my_app"]
```

But imho, I think the PR solution is cleaner

### DIFF
--- a/guides/deployment/releases.md
+++ b/guides/deployment/releases.md
@@ -217,7 +217,7 @@ RUN \
 # Everything from this line onwards will run in the context of the unprivileged user.
 USER "${USER}"
 
-COPY --from=build --chown="${USER}":"${USER}" /app/_build/"${MIX_ENV}"/rel/my_app ./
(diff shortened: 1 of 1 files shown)


josevalim:

c4710n:
In the [doc about COPY](https://docs.docker.com/engine/reference/builder/#copy), it mentions:

> The directory itself is not copied, just its contents.

I don't know how you produce this issue, but previous  version of `Dockerfile`  should work as expected.

@onixus74

josevalim:
I tried locally without this change and it indeed worked. So reverting. Thanks @c4710n!

PR #4542, opened 2021-10-10
Clarify language in phx.gen.auth

t0mhaines:
One word change to clarify that the generator doesn't work with projects generated with **either** `--no-ecto` **or** `--no-html`.


josevalim:

PR #4550, opened 2021-10-14
Update contexts.md & mix_tasks.md

angrycandy:
* mainly punctuation & grammar fixes
* introduce task args with `:` instead of `,`
* standardize task args description to match `mix phx.gen.auth`


Gazler:
Thanks <3

PR #4556, opened 2021-10-21
Update channels.md

angrycandy:
Correct instruction to match v1.6.2 `assets/js/app.js`:
```javascript
// If you want to use Phoenix channels, run `mix help phx.gen.channel`
// to get started and then uncomment the line below.
// import "./user_socket.js"

// You can include dependencies in two ways.
```

### DIFF
--- a/guides/real_time/channels.md
+++ b/guides/real_time/channels.md
@@ -267,11 +267,11 @@ channel.join()
 export default socket
 ```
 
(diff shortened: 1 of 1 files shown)


aaronrenner (approved):

aaronrenner:
Great find! Thank you! :heart:

PR #4559, opened 2021-10-22
Accidental use of match for checking against empty map

hauleth:
Ref elixir-lang/elixir#11331


aaronrenner (approved):

aaronrenner:
Thanks for catching this! :sunglasses:

PR #4576, opened 2021-11-02
Remove already mentioned comment about runtime.exs

eksperimental:
Its is already mentioned some lines before within the same section of the guide.


josevalim:

PR #4584, opened 2021-11-09
Tweak Channel Generator Output

optikfluffel:
I added some parentheses, so the generated code passes `mix format --check-formatted`.


josevalim:

PR #4589, opened 2021-11-10
Fix link to phoenix javascript files

axelson:
phoenix.js is no longer in a single js file, it is now a directory of files. Alternatively we could link to the index.js as the entry point but I think linking to the entire directory is easier to navigate from.


josevalim:

Issue #4597, opened 2021-11-13
Safari downloads the page instead of rendering using   send_resp(conn, 201, "")


khanakia:
### Environment

* Elixir version (elixir -v): Elixir 1.12.3 (compiled with Erlang/OTP 24)
* Phoenix version (mix deps):  phoenix 1.6.2
* Operating system: OSX Big Sur

### Expected behavior
I am learning the phoenix and based on the tutorial here (https://hexdocs.pm/phoenix/controllers.html#sending-responses-directly) it should show the blank page according to the docs  
`Reloading http://localhost:4000 should show us a completely blank page. The network tab of our browser's developer tools should show a response status of "201" (Created).`

### Actual behavior
Safari keeps downloading the page instead of showing the empty page.
Chrome is working fine.

https://user-images.githubusercontent.com/2174170/141608910-a802adef-312d-4a92-a17a-8b98385374bf.mov

petermm:
Hi Aman,

Safari downloads it as there is no content_type and it doesn't know what to do with the response.. if you add ` |> put_resp_content_type("text/plain")` to as in the next step, safari will know it's text and render it..

btw: the issue tracker is usually for "actionable" items - you can check out elixirforum.com and post questions there..

all the best🙌

petermm:
realized the doc was wrong, and it wasn't a question.. so made PR correcting docs.

PR #4598, opened 2021-11-13
update controllers.md (safari no content type)

petermm:
The docs are not precise, when using safari - as it will not render without a content_type..

More than open for better wording.

fixes https://github.com/phoenixframework/phoenix/issues/4597

### DIFF
--- a/guides/controllers.md
+++ b/guides/controllers.md
@@ -264,7 +264,7 @@ end
 
 Reloading [http://localhost:4000](http://localhost:4000) should show us a completely blank page. The network tab of our browser's developer tools should show a response status of "201" (Created). Some browsers (Safari) will download the response, as the content type is not set.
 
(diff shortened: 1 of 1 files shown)


guides/controllers.md
```diff
@@ -262,9 +262,10 @@ def index(conn, _params) do
 end
 ```
 
-Reloading [http://localhost:4000](http://localhost:4000) should show us a completely blank page. The network tab of our browser's developer tools should show a response status of "201" (Created).
+Reloading [http://localhost:4000](http://localhost:4000) should show us a completely blank page. The network tab of our browser's developer tools should show a response status of "201" (Created). Some browsers (Safari) will download the response, as the content type is not set.
+
+So we would like to be specific about the content type, and we can use [`put_resp_content_type/2`] in conjunction with [`send_resp/3`].
```
josevalim, line 267:
```suggestion
To be specific about the content type, we can use [`put_resp_content_type/2`] in conjunction with [`send_resp/3`].
```

josevalim:

PR #4601, opened 2021-11-17
bugfix: allow creating enum with a single value

milmazz:
While I don't see why someone would want to do this, the DB
engines (e.g., PostgreSQL, MySQL) allow setting an enum type with a single value and don't throw
errors.

Fixes: #4594


chrismccord:
❤️❤️❤️🐥🔥

PR #4609, opened 2021-11-23
Add phx.gen.release task for release/docker based deployments

chrismccord:
I'm not super firm on env names, so would love feedback. The migrate and server shims are necessary for deployments where it's difficult for infrastructure to know what the OTP app name is. If we are happy with this direction, we can add bat scripts for windows to match migrate/server.


chrismccord:
The `bin/server` shim also aids folks who forget to set `server: true` in their config. This has been very frequent stumbling block over the years, so by generating runtime config that looks for the server env, we can also smooth over this particular stumbling block.

priv/templates/phx.gen.docker/Dockerfile.eex
```diff
@@ -0,0 +1,93 @@
...
+# your Elixir templates, you will need to move the asset compilation
+# step down so that `lib` is available.
+COPY assets assets
+
+# For Phoenix 1.6 and later, compile assets using esbuild
+RUN mix assets.deploy
+
+# For Phoenix versions earlier than 1.6, compile assets npm
+# RUN cd assets && yarn install && yarn run webpack --mode production
+# RUN mix phx.digest
```
mcrumm, line 58: I think we can drop the old asset commands for clarity, but maybe you had a specific reason for keeping them around?

```suggestion
# compile assets
RUN mix assets.deploy
```

chrismccord: no, great call. Copy pasta :)

installer/lib/phx_new/generator.ex
```diff
@@ -349,10 +349,12 @@ defmodule Phx.New.Generator do
           environment variable DATABASE_URL is missing.
           For example: ecto://USER:PASS@HOST/DATABASE
           \"""
+
+      ipv6? = System.get_env("IPV6") == "true"
```
Gazler, line 353: Is `IPV6` the correct environment variable here? Since this is specific to the db socket, wouldn't something like `DATABASE_IPV6` be more appropriate?

lib/mix/tasks/phx.gen.docker.ex
```diff
@@ -0,0 +1,99 @@
...
+    ])
+    File.chmod!("rel/overlays/bin/migrate", 0o700)
+    File.chmod!("rel/overlays/bin/server", 0o700)
+
+    Mix.shell().info("""
+
+    Your application is ready to be deployed in a release!
+
+        # To start your system
+        _build/dev/rel/live_beats/bin/live_beats start
```
Gazler, line 74: `s/live_beats/#{app}`

lib/mix/tasks/phx.gen.docker.ex
```diff
@@ -0,0 +1,99 @@
...
+      {:eex, "env.sh.eex", "rel/env.sh.eex"},
+      {:eex, "remote.vm.args.eex", "rel/remote.vm.args.eex"},
+      {:eex, "vm.args.eex", "rel/vm.args.eex"},
+      {:eex, "migrate.sh.eex", "rel/overlays/bin/migrate"},
+      {:eex, "server.sh.eex", "rel/overlays/bin/server"}
+    ])
+    File.chmod!("rel/overlays/bin/migrate", 0o700)
+    File.chmod!("rel/overlays/bin/server", 0o700)
+
+    Mix.shell().info("""
```
Gazler, line 69: Shouldn't this shell info explain something about building for Docker vs release information from mix?

lib/mix/tasks/phx.gen.docker.ex
```diff
@@ -0,0 +1,99 @@
...
+    ]
+
+    Mix.Phoenix.copy_from(paths(), "priv/templates/phx.gen.docker", binding, [
+      {:eex, "Dockerfile.eex", "Dockerfile"},
+      {:eex, "dockerignore.eex", ".dockerignore"},
+      {:eex, "release.ex", Mix.Phoenix.context_lib_path(app, "release.ex")}
+    ])
+
+    Mix.Phoenix.copy_from(paths(), "priv/templates/phx.gen.docker/rel", binding, [
+      {:eex, "env.bat.eex", "rel/env.bat.eex"},
```
Gazler, line 59: Is there a way we could leverage these from the existing `mix release.init` task to keep any changes in sync with the Elixir version?

priv/templates/phx.gen.docker/Dockerfile.eex
```diff
@@ -0,0 +1,89 @@
...
+#
+#
+# This file is based on these images:
+#
+#   - https://hub.docker.com/r/hexpm/elixir/tags - for the build image
+#   - https://hub.docker.com/_/debian?tab=tags&page=1&name=bullseye-20210902-slim - for the release image
+#   - https://pkgs.org/ - resource for finding needed packages
+#   - Ex: hexpm/elixir:1.12.3-erlang-24.1.4-debian-bullseye-20210902-slim
+#
+ARG BUILDER_IMAGE="hexpm/elixir:1.12.3-erlang-24.1.4-debian-bullseye-20210902-slim"
```
Gazler, line 15: Should we leverage:

`Application.spec(:elixir, :vsn)` and `:erlang.system_info(:version)` (not sure how to get the minor version here)

So that the build image matches the elixir version in use?

This might have a downside as we can't ensure the image exists vs using a hard-coded string.

Gazler (commented):
Good idea, covers most of the common cases. Big +1 for the `server: true` change.

priv/templates/phx.gen.docker/Dockerfile.eex
```diff
@@ -0,0 +1,89 @@
...
+COPY config/config.exs config/${MIX_ENV}.exs config/
+RUN mix deps.compile
+
+COPY priv priv
+
+# note: if your project uses a tool like https://purgecss.com/,
+# which customizes asset compilation based on what it finds in
+# your Elixir templates, you will need to move the asset compilation
+# step down so that `lib` is available.
+COPY assets assets
```
ZombieHarvester, line 51: Can asset compilation always be after `lib`? Won't it cover both cases with and without purgecss? 

I don't know if there are any downsides of permanently moving it a step down, but this could make the generated dockerfile clearer.

mrkurt: The downside is slower builds when you change something in lib. This would trigger and asset recompilation.

There's a way to do both with a multistage Dockerfile though, it's a good idea.

installer/templates/phx_single/config/runtime.exs
```diff
@@ -28,6 +32,8 @@ if config_env() == :prod do
       ip: {0, 0, 0, 0, 0, 0, 0, 0},
       port: String.to_integer(System.get_env("PORT") || "4000")
     ],
+    server: server?,
+    check_origin: ["//#{host}"],
```
josevalim, line 36: Isn't this the default?

lib/mix/tasks/phx.gen.docker.ex
josevalim: I don't see why we need to generate them. Could we tell them to use `mix release.init` if necessary?

lib/mix/tasks/phx.gen.docker.ex
```diff
@@ -0,0 +1,99 @@
...
+      {:eex, "vm.args.eex", "rel/vm.args.eex"},
+      {:eex, "migrate.sh.eex", "rel/overlays/bin/migrate"},
+      {:eex, "server.sh.eex", "rel/overlays/bin/server"}
+    ])
+    File.chmod!("rel/overlays/bin/migrate", 0o700)
+    File.chmod!("rel/overlays/bin/server", 0o700)
+
+    Mix.shell().info("""
+
+    Your application is ready to be deployed in a release!
```
josevalim, line 71: We need to tell them to run the release command first.

priv/templates/phx.gen.docker/Dockerfile.eex
josevalim: Good call. We can do `-erlang-24.1-`, which perhaps is the safest choice.

lib/mix/tasks/phx.gen.docker.ex
josevalim: And this information is also shown by the release command.

lib/mix/tasks/phx.gen.docker.ex
josevalim: Alternatively, we call this `mix phx.gen.release` and support a `--docker` flag?

installer/templates/phx_single/config/runtime.exs
```diff
@@ -19,7 +19,11 @@ if config_env() == :prod do
       You can generate one by calling: mix phx.gen.secret
       """
 
+  host = System.get_env("URL_HOST") || "example.com"
+  server? = System.get_env("SERVER") == "true"
```
josevalim, line 23: IMO:

```suggestion
  server? = System.get_env("PHX_SERVER") == "true"
```

and `PHX_HOST`.

installer/templates/phx_single/config/runtime.exs
```diff
@@ -19,7 +19,11 @@ if config_env() == :prod do
...
+
   config :<%= @app_name %>, <%= @endpoint_module %>,
+    url: [host: host, port: 80],
```
josevalim, line 26: Not sure if I am happy with the hardcoded port here. I am wondering if instead we could support `url` as a string which we will parse. So we could do: `url: System.get_env("PHX_URL")`, and we set `PHX_URL=example.com:80`. The code would parse the string and discard all nil fields?

lib/mix/tasks/phx.gen.docker.ex
```diff
@@ -0,0 +1,99 @@
...
+
+  See the `Mix.Release` docs for more details.
+  """
+
+  use Mix.Task
+
+  @doc false
+  def run(_args) do
+    if Mix.Project.umbrella?() do
+      Mix.raise("mix phx.gen.docker is not supported in umbrella applications")
```
josevalim, line 42: Maybe mention they should generate it within their web app instead?

chrismccord:
Updates:

- The task has been renamed to `phx.gen.release` with an optional `--docker` flag for the Dockerfile.
- Env vars included in `config/runtime.exs` for new apps are: `ECTO_IPV6`, `PHX_HOST`, and `PHX_SERVER`
- Warnings are printed after the task is run if above env vars are not found in the file with instructions
- Default `:url` port for production is set to 443, which is something we should have defaulted to a long time ago
- The `mix release.init` files are no longer generated as they are not necessary and a user can opt-in to them in the future by running `mix release.init`

installer/lib/phx_new/generator.ex
```diff
@@ -349,12 +349,14 @@ defmodule Phx.New.Generator do
           environment variable DATABASE_URL is missing.
           For example: ecto://USER:PASS@HOST/DATABASE
           \"""
+
+      ipv6? = System.get_env("ECTO_IPV6") == "true"
```
josevalim, line 353: Usually we don't check for true specifically, because it is common to set it to 1 or similar? We could simply set if the variable is set, but that will also enable it if it is empty... which might be ok? DATABASE_URL would be used even for empty. So maybe something like this:

```suggestion
      maybe_ipv6 = if System.get_env("ECTO_IPV6"), do: [:inet6], else: []
```

And below you do `socket_options: maybe_ipv6`?

installer/templates/phx_single/config/runtime.exs
```diff
@@ -19,15 +19,21 @@ if config_env() == :prod do
       You can generate one by calling: mix phx.gen.secret
       """
 
+  server? = System.get_env("PHX_SERVER") == "true"
```
josevalim, line 22: Same as above?

```suggestion
  server? = System.get_env("PHX_SERVER") != nil
```

lib/mix/tasks/phx.gen.release.ex
```diff
@@ -0,0 +1,141 @@
...
+  defp post_install_instructions(path, matching, msg) do
+    case File.read(path) do
+      {:ok, content} ->
+        unless content =~ matching, do: Mix.shell().info(msg)
+
+      {:error, _} ->
+        Mix.shell().info(msg)
+    end
+  end
+end
```
josevalim, line 141: This file is gorgeous.

priv/templates/phx.gen.release/Dockerfile.eex
```diff
@@ -0,0 +1,89 @@
...
+#
+#
+# This file is based on these images:
+#
+#   - https://hub.docker.com/r/hexpm/elixir/tags - for the build image
+#   - https://hub.docker.com/_/debian?tab=tags&page=1&name=bullseye-20210902-slim - for the release image
+#   - https://pkgs.org/ - resource for finding needed packages
+#   - Ex: hexpm/elixir:1.12.3-erlang-24.1.4-debian-bullseye-20210902-slim
+#
+ARG BUILDER_IMAGE="hexpm/elixir:1.12.3-erlang-24.1.4-debian-bullseye-20210902-slim"
```
josevalim, line 15: We need to get the Elixir and OTP version from the user system.

We should also add a note to the RELEASE file to bump the debian version here on new release (we should also equally add a note to bump the esbuild version).

lib/mix/tasks/phx.gen.release.ex
```diff
@@ -0,0 +1,141 @@
...
+    app = Mix.Phoenix.otp_app()
+    app_namespace = Mix.Phoenix.base()
+
+    binding = [
+      app_namespace: app_namespace,
+      otp_app: app
+    ]
+
+    Mix.Phoenix.copy_from(paths(), "priv/templates/phx.gen.release", binding, [
+      {:eex, "rel/migrate.sh.eex", "rel/overlays/bin/migrate"},
```
josevalim, line 50: Should we check if the user has Ecto? Maybe the app was installed with --no-ecto? 🤔

priv/templates/phx.gen.release/rel/migrate.sh.eex
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+cd -P -- "$(dirname -- "$0")"
+./<%= otp_app %> eval <%= app_namespace %>.Release.migrate
```
josevalim, line 3: I am not a bash especialist, but should we use `exec` here to avoid the additional process?

```suggestion
exec ./<%= otp_app %> eval <%= app_namespace %>.Release.migrate
```

priv/templates/phx.gen.release/rel/server.sh.eex
```diff
@@ -0,0 +1,3 @@
+#!/bin/sh
+cd -P -- "$(dirname -- "$0")"
+PHX_SERVER=true ./<%= otp_app %> start
```
josevalim, line 3: Similar here:

```suggestion
exec PHX_SERVER=true ./<%= otp_app %> start
```

I don't know if `exec` should come before or after the env var.

lib/mix/tasks/phx.gen.release.ex
chrismccord: yes, great catch!

lib/mix/tasks/phx.gen.release.ex
```diff
@@ -0,0 +1,169 @@
...
+    major = to_string(:erlang.system_info(:otp_release))
+    path = Path.join([:code.root_dir(), "releases", major, "OTP_VERSION"])
+
+    with {:ok, content} <- File.read(path),
+         {:ok, %Version{} = vsn} <- Version.parse(String.trim_trailing(content)) do
+      "#{vsn.major}.#{vsn.minor}.#{vsn.patch}"
+    else
+      _ ->
+        IO.warn("unable to read OTP minor version at #{path}. Falling back to #{major}.0.0")
+        "#{major}.0.0"
```
josevalim, line 166:
```suggestion
        IO.warn("unable to read OTP minor version at #{path}. Falling back to #{major}.0")
        "#{major}.0"
```

lib/mix/tasks/phx.gen.release.ex
```diff
@@ -0,0 +1,169 @@
...
+    end
+  end
+
+  def otp_vsn do
+    major = to_string(:erlang.system_info(:otp_release))
+    path = Path.join([:code.root_dir(), "releases", major, "OTP_VERSION"])
+
+    with {:ok, content} <- File.read(path),
+         {:ok, %Version{} = vsn} <- Version.parse(String.trim_trailing(content)) do
+      "#{vsn.major}.#{vsn.minor}.#{vsn.patch}"
```
josevalim, line 162: OTP versioning is... tricky. There is no 24.1.0, it is just 24.1. But i guess you could do this:

```suggestion
    with {:ok, content} <- File.read(path) do
      String.trim(content)
```

lib/mix/tasks/phx.gen.release.ex
```diff
@@ -0,0 +1,169 @@
...
+    Mix.shell().info("""
+
+    Your application is ready to be deployed in a release!
+
+        # To start your system
+        _build/dev/rel/#{app}/bin/#{app} start
+
+        # To start your system with the Phoenix server running
+        _build/dev/rel/#{app}/bin/server
+        #{ecto? && ecto_instructions(app)}
```
josevalim, line 82:
```suggestion
    #{ecto? && ecto_instructions(app)}
```

lib/mix/tasks/phx.gen.release.ex
```diff
@@ -0,0 +1,169 @@
...
+        _build/dev/rel/#{app}/bin/#{app}
+    """)
+
+    ecto? &&
+      post_install_instructions("config/runtime.exs", ~r/ECTO_IPV6/, """
+      [warn] Conditional IPV6 support missing from runtime configuration.
+
+      Add the following to your config/runtime.exs:
+
+          ipv6? = System.get_env("ECTO_IPV6") == "true"
```
josevalim, line 102: You need to update this.

lib/mix/tasks/phx.gen.release.ex
```diff
@@ -0,0 +1,169 @@
...
+            ...,
+            socket_options: if(ipv6?, do: [:inet6], else: [])
+      """)
+
+    post_install_instructions("config/runtime.exs", ~r/PHX_SERVER/, """
+    [warn] Conditional server startup is missing from runtime configuration.
+
+    Add the following to your config/runtime.exs:
+
+        server? = System.get_env("PHX_SERVER") == "true"
```
josevalim, line 114:
```suggestion
        server? = System.get_env("PHX_SERVER") != nil
```

PR #4615, opened 2021-11-27
ecto guide: update iex example output and add missing concept in Data persistence section

netProphET:
These changes bring the Ecto Guide's iex examples in line with the current Phoenix version.


josevalim:

PR #4619, opened 2021-12-04
Make live_modal id dynamic

mustela:
If you have more than one calls to the `live_modal` in a view, it will fail with `...elixir phoenix found duplicate ID :modal for component,...` that's because the `id` inside the function is hardcoded. 

This PR will introduce a chance to use the `id` from the declaration. Falling back to `:modal` in case it is not set.


chrismccord:
We have reworked the modals to use new function components, so this is no longer up to date. Thanks!

PR #4621, opened 2021-12-07
Update Copyright date

tsrandrei:


cw789:
Copyright names the year of first publication. Not the current year.
So it's correct as is.

chrismccord:
Thanks!

Issue #4618, opened 2021-12-03
Wrong Channel is joined due to the `channel` order


andersonmcook:
### Environment

* Elixir version (elixir -v): Elixir 1.12.3 (compiled with Erlang/OTP 24)
* Phoenix version (mix deps): 1.6.2
* Operating system: MacOS

### Expected behavior

For multiple channels the channel with a more specific topic should match first, or the order should not be reversed when creating the channels in https://github.com/phoenixframework/phoenix/blob/v1.6.2/lib/phoenix/socket.ex#L352

```elixir
  channel "a:b*", ABStar
  channel "a:*", AStar
```

For "a:b1" the `ABStar` channel is joined.

### Actual behavior

For "a:b1" the `AStar` channel is joined.

I've created an example app with a test
https://github.com/andersonmcook/channel_order

https://github.com/andersonmcook/channel_order/blob/main/lib/channel_order_web/channels/user_socket.ex
https://github.com/andersonmcook/channel_order/blob/main/test/channel_order_web/channels/user_socket_test.exs

josevalim:
Can you please provide a pull request with tests? :) Let me know if you need assistance. :)

andersonmcook:
@josevalim I can definitely provide a pull request with tests.  

My initial thought was just to report this as a fix reversing the order could cause a regression in applications that use Phoenix.  I know for our app I added tests very similar to the example app to check if the order ever changed.

We might have to order by specificity.

josevalim:
I think we should reverse it and list it as a bug fixes. It will only be available on v1.7 though.

josevalim:
Closing in favor of PR, thank you! :heart:

PR #4622, opened 2021-12-07
Reverse order of definition of channel handler functions

andersonmcook:
References #4618


josevalim (approved):
@chrismccord, before you merge this, you should start a v1.6 branch. This should be v1.7+ only just in case this accidentally breaks stuff for someone.

josevalim:

PR #4625, opened 2021-12-08
Update the Phoenix.Token.verify/4 docs to mention the default value for max_age

sudochop:
The default should be called out so it doesn't lead to confusion or bugs when working with this function. Verbiage taken directly from https://github.com/elixir-plug/plug_crypto/blob/v1.2.2/lib/plug/crypto.ex#L257


josevalim:

PR #4626, opened 2021-12-08
Place hostname before database in config

eksperimental:
This follows the orders used in DATABASE_URL, which is
1. username
2. password
3. host
4. database


chrismccord:
❤️❤️❤️🐥🔥

PR #4628, opened 2021-12-11
Leverage MIX_ENV on the runner image

leandrocp:
The builder image reuses `MIX_ENV` to get deps and copy configs, but the runner image assumes the app was built on the "prod" env, which may not be case if one changes the `MIX_ENV` env value on the build image. This PR just make it a bit easier to spot that the env should be same on both build and runner images.


josevalim (approved):

chrismccord:
❤️❤️❤️🐥🔥

Issue #4629, opened 2021-12-12
Phoenix 1.6.4: Esbuild watcher bug when running in dev container on M1


restlessronin:
### Environment

* Elixir version (elixir -v): 1.13.0
* Phoenix version (mix deps): 1.6.4
* OTP: 24.1.7
* Operating system: MacOS Big Sur 11.6.1
* Hardware M1 chip
* Base Docker Image: `hexpm/elixir:1.13.0-erlang-24.1.7-ubuntu-focal-20210325`

### Steps to Reproduce

1. `mix phx.new . --app my_app --no-ecto --no-mailer`
2. `mix phx.server`
3.  browse to http://localhost:4000

### Expected behavior

See the Phoenix Welcome Screen

### Actual behavior

Works as expected when running natively on MacOS

In container there is no page received.

### Workaround

Comment out the esbuild watcher in container
```
  watchers: [
    # Start the esbuild watcher by calling Esbuild.install_and_run(:default, args)
#    esbuild: {Esbuild, :install_and_run, [:default, ~w(--sourcemap=inline --watch)]}
  ]
```
The page is then received as normal

### Possible cause

This may be related to the fact that `esbuild` uses `esbuild-darwin-arm64` natively and `esbuild-linux-arm64` in container.

```
mix assets.deploy
```

works correctly on native, but segfaults in container

chrismccord:
I know folks are still having various issues with M1 and docker, so I don't believe this is anything for us to handle on our side. Thanks!

restlessronin:
Fair Enough. If I have some time, I'll look into `esbuild` itself.

PR #4630, opened 2021-12-14
Fixes m1 container segfault (Issue #4629)

restlessronin:
The `esbuild` installer is by default using version `0.14.0`. However `phx.new` is generating an old version `0.13.5` for the :dev config. When there is a mismatch, esbuild downloads the version that is in the config.

Seemingly esbuild v 0.13.5 segfaults in container, but 0.14.0 does not. Changing the version number in the config fixed my problem.

I made the changes to the `phx.new` installer templates, as well as the `config.ex` in the tree, and am submitting them as this PR.

Thanks for the great work.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -18,7 +18,7 @@ if Mix.env() == :dev do
   end
 
   config :esbuild,
-    version: "0.13.5",
+    version: "0.14.0",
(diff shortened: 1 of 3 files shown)


restlessronin:
🙏🏾🙏🏾❤️

PR #4632, opened 2021-12-15
Support opt-in host origin checking

chrismccord:


lib/phoenix/socket/transport.ex
```diff
@@ -588,13 +591,15 @@ defmodule Phoenix.Socket.Transport do
     end
   end
 
-  defp origin_allowed?({module, function, arguments}, uri, _endpoint),
+  defp origin_allowed?({module, function, arguments}, uri, _endpoint, _conn),
     do: apply(module, function, [uri | arguments])
-  defp origin_allowed?(_check_origin, %{host: nil}, _endpoint),
+  defp origin_allowed?(:host, uri, _endpoint, %Plug.Conn{} = conn),
+    do: uri.host == conn.host
```
josevalim, line 597: IIRC you also need to check port and scheme.

Issue #4401, opened 2021-08-06
Telemetry 1.0 warnings


sneako:
Since upgrading to Telemetry 1.0, I am now seeing the following warnings emitted:
```
15:57:30.098 [info]  Function passed as a handler with ID {Phoenix.Logger, [:phoenix, :channel_handled_in]} is local function.
This mean that it is either anonymous function or capture of function without module specified. That may cause performance penalty when calling such handler. For more details see note in `telemetry:attach/4` documentation.

https://hexdocs.pm/telemetry/telemetry.html#attach-4

15:57:30.116 [info]  Function passed as a handler with ID {Phoenix.Logger, [:phoenix, :channel_joined]} is local function.
This mean that it is either anonymous function or capture of function without module specified. That may cause performance penalty when calling such handler. For more details see note in `telemetry:attach/4` documentation.

https://hexdocs.pm/telemetry/telemetry.html#attach-4

15:57:30.116 [info]  Function passed as a handler with ID {Phoenix.Logger, [:phoenix, :endpoint, :start]} is local function.
This mean that it is either anonymous function or capture of function without module specified. That may cause performance penalty when calling such handler. For more details see note in `telemetry:attach/4` documentation.

https://hexdocs.pm/telemetry/telemetry.html#attach-4

15:57:30.116 [info]  Function passed as a handler with ID {Phoenix.Logger, [:phoenix, :endpoint, :stop]} is local function.
This mean that it is either anonymous function or capture of function without module specified. That may cause performance penalty when calling such handler. For more details see note in `telemetry:attach/4` documentation.

https://hexdocs.pm/telemetry/telemetry.html#attach-4

15:57:30.116 [info]  Function passed as a handler with ID {Phoenix.Logger, [:phoenix, :error_rendered]} is local function.
This mean that it is either anonymous function or capture of function without module specified. That may cause performance penalty when calling such handler. For more details see note in `telemetry:attach/4` documentation.

https://hexdocs.pm/telemetry/telemetry.html#attach-4

15:57:30.116 [info]  Function passed as a handler with ID {Phoenix.Logger, [:phoenix, :router_dispatch, :start]} is local function.
This mean that it is either anonymous function or capture of function without module specified. That may cause performance penalty when calling such handler. For more details see note in `telemetry:attach/4` documentation.

https://hexdocs.pm/telemetry/telemetry.html#attach-4

[...]

PR #4634, opened 2021-12-16
Simplify code

marcandre:


lib/phoenix/logger.ex
```diff
@@ -123,13 +123,13 @@ defmodule Phoenix.Logger do
...
   def install do
     handlers = %{
-      [:phoenix, :endpoint, :start] => &__MODULE__.phoenix_endpoint_start/4,
-      [:phoenix, :endpoint, :stop] => &__MODULE__.phoenix_endpoint_stop/4,
-      [:phoenix, :router_dispatch, :start] => &__MODULE__.phoenix_router_dispatch_start/4,
-      [:phoenix, :error_rendered] => &__MODULE__.phoenix_error_rendered/4,
-      [:phoenix, :socket_connected] => &__MODULE__.phoenix_socket_connected/4,
-      [:phoenix, :channel_joined] => &__MODULE__.phoenix_channel_joined/4,
-      [:phoenix, :channel_handled_in] => &__MODULE__.phoenix_channel_handled_in/4
+      [:phoenix, :endpoint, :start] => &phoenix_endpoint_start/4,
```
thiamsantos, line 126: @marcandre I think it is the intention to be this way in order to avoid telemetry warning. Ref #4401

marcandre:
Thanks @thiamsantos, I should have done a git blame 😅

Very interesting, it would not have crossed my mind that these could mean different things. If I understand, one difference is that if the module is reloaded with a different version of the function, then without the explicit module, the original function that was passed at creation time will be called, while with the explicit module, the up-to-date function will be called ([ref](https://www.erlang.org/doc/man/erlang.html#fun_info-1)). It's quite unclear to me why that would have an impact on performance (and the doc [does not indicate](https://www.erlang.org/doc/efficiency_guide/functions.html#function-calls) anything to that regard), so I'm very curious to understand more.

thiamsantos:
@marcandre I was also curious to understand why we would have any performance impact with either approach. I could found this PR https://github.com/beam-telemetry/telemetry/pull/43 with a benchmark demonstrating some difference in performance. 

Quoting from there:

> It turned out, that when anonymous functions are copied into the process heap from ETS, a global reference counter is incremented, which we suspect is the cause of contention.

I don't fully understand the internals, but this seems to be something specific of how telemetry store the references of the handlers.

PR #4638, opened 2021-12-18
Let watchers execute batch files on Windows

### LINKED ISSUE #4617: It's hard to write cross-platform watchers
### Environment

* Elixir version (elixir -v): Elixir 1.12.1 (compiled with Erlang/OTP 22)
* Phoenix version (mix deps): 1.6.2
* Operating system: Windows

### Situation

Most of our team is on Linux and Mac. A colleague had adopted Phoenix's `watchers` feature, and added the following:

```ex
  watchers: [
    npm: ["run", "assets:watch:js", cd: Path.expand("..", __DIR__)],
    npm: ["run", "assets:watch:css", cd: Path.expand("..", __DIR__)]
  ]
```

### Expected behavior

I expected phoenix to start normally on my Windows computer

### Actual behavior

I got spawn errors and the watchers did not run.

### Underlying problem

I don't grok unix all too deeply so pardon me if I get details wrong. I think Phoenix spawns the watchers directly as OS child processes. On unixes, I think anything that is `+x` can be spawned that way, and if it's a script (and not a binary executable), then the OS reads the hashbang to figure out which shell to use.

On Windows, however, you cannot do this as easily: the OS only lets you spawn an actual binary executable. I noticed that the Phoenix watcher documentation example works around this limitation by launching `node`, which is a binary everywhere, as opposed to some .cmd or .bat file such as `npm.cmd` and `iex.bat`. But that's not very developer friendly (and I couldn't figure how to launch `npm` using `node yadayada` in a cross-platform way, although I'm sure there's a way).

To my understanding, Windows treats batch files as regular text files and not like executables at all. They are associated on the shell level to open in `cmd.exe` by default - exactly like how you open a .txt file in Notepad when double-clicking it.

I was able to solve the problem on my machine by passing `npm.cmd` to `cmd.exe` explicitly, eg:

```ex
  watchers: [
    cmd: ["/c", "npm", "run", "assets:watch:js", cd: Path.expand("..", __DIR__)],
    cmd: ["/c", "npm", "run", "assets:watch:css", cd: Path.expand("..", __DIR__)]
  ]
```

But now it's Windows-only, so I hacked a little local helper function to make it cross-platform:

```ex
npm = fn args ->
  case :os.type() do
    {:win32, _} -> {:cmd, ["/c", "npm"] ++ args}
    _ -> {:npm, args}
  end
end

# ... 

  watchers: [
    npm.(["run", "assets:watch:js", cd: Path.expand("..", __DIR__)]),
    npm.(["run", "assets:watch:css", cd: Path.expand("..", __DIR__)])
  ]
```

This solved it for us, but I still thought it's worth reporting. 

### Possible solutions

I looked around the npm sources a bit, and it looks like they always launch their `npm run` commands through an OS shell (bash on unix, cmd.exe on windows): https://github.com/npm/run-script#usage. Maybe it's worth doing the same for Phoenix? 

That said, doing it like that also imports the questionable NPM design that encourages people to `&&` entire scripts together inside package.json, carefully frankensteined such that a single command is legal bash _and_ legal cmd syntax. I myself have become quite adept at doing that, but it's the kind of skill that nobody should've had to need if our tools were designed better :-) 

An alternative would probably be something like "are we on windows? and does the result of `:os.find_executable('npm')` end with `.cmd` or `.bat`? then launch it with cmd.exe and not directly". I'd be happy to attempt a PR that does this.

eteeselink:
Turns out your ping helped Jose! As discussed in #4617, here's an attempt to let Phoenix watchers run batch files more easily.

I took care to not blanket pass any command to `cmd /c` on Windows, because that's a step in the direction of the same kind of hell that writing package.json scripts is, with people trying to write scripts that are both valid sh and valid cmd or something like that.

This should keep the old behavior in tact *unless*:
1. the user is on Windows
2. the executable specified is found in the system PATH (plus `opts[:cd]`) and ends with `.cmd` or `.bat`

I like to think that that can't cause breaking changes since any watcher that matches point 2 above would've failed previously (`System.cmd`'s `find_executable` call would find it, but subsequently fail to launch it) (come to think of it, that's a bit of a "bad error message" kind of bug in `System.cmd` isn't it?)

Note that this is my first proper Elixir ecosystem contribution and I'm not a very seasoned Elixir dev by any means. Any code commentary (or straights edits) will be warmly appreciated.

### DIFF
--- a/lib/phoenix/endpoint/watcher.ex
+++ b/lib/phoenix/endpoint/watcher.ex
@@ -32,6 +32,8 @@ defmodule Phoenix.Endpoint.Watcher do
     {args, opts} = Enum.split_while(args, &is_binary(&1))
     opts = Keyword.merge([into: IO.stream(:stdio, :line), stderr_to_stdout: true], opts)
 
+    {cmd, args} = maybe_prefix_shell(:os.type(), cmd, args, opts)
+
     try do
       System.cmd(cmd, args, opts)
     catch
@@ -57,4 +59,39 @@ defmodule Phoenix.Endpoint.Watcher do
   end
 
   defp cd(opts), do: opts[:cd] || File.cwd!()
+
+  defp maybe_prefix_shell({:win32, _}, cmd, args, opts) do
+    # This lets us call batch files on Windows, which aren't proper executables and can't be simply
+    # passed to `System.cmd`.
+    case find_executable(cmd, opts[:cd]) do
+      nil ->
+        {cmd, args}
+
+      path ->
+        if String.ends_with?(path, [".cmd", ".bat"]) do
+          {"cmd", ["/c", Path.basename(path)] ++ args}
+        else
+          {cmd, args}
+        end
+    end
+  end
+
+  defp maybe_prefix_shell(_, cmd, args, _opts) do
+    {cmd, args}
+  end
+
+  defp find_executable(cmd, cd) do
+    path =
+      if cd do
+        search_path = System.get_env("PATH") <> ";" <> cd
+        :os.find_executable(String.to_charlist(cmd), String.to_charlist(search_path))
+      else
+        :os.find_executable(String.to_charlist(cmd))
+      end
+
+    case path do
+      false -> nil
+      other -> List.to_string(other)
+    end
+  end
 end
--- a/test/phoenix/endpoint/echo.bat
+++ b/test/phoenix/endpoint/echo.bat
@@ -1,0 +1,5 @@
+:: batch file used in Phoenix watcher test. serves 2 purposes:
+:: 1. polyfills the watcher test's use of `echo`, which isn't 
+::    an executable on windows but an internal command
+:: 2. verifies that the watcher code can launch batch files on windows.
+@echo %*
--- a/test/phoenix/endpoint/watcher_test.exs
+++ b/test/phoenix/endpoint/watcher_test.exs
@@ -4,12 +4,20 @@ defmodule Phoenix.Endpoint.WatcherTest do
   alias Phoenix.Endpoint.Watcher
   import ExUnit.CaptureIO
 
-  test "starts watching and writes to stdio with args" do
+  test "starts watching and writes to stdio with args (batch file when on windows)" do
     assert capture_io(fn ->
-      {:ok, pid} = Watcher.start_link({"echo", ["hello", cd: File.cwd!()]})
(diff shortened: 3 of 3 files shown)


lib/phoenix/endpoint/watcher.ex
```diff
@@ -57,4 +59,39 @@ defmodule Phoenix.Endpoint.Watcher do
...
+        end
+    end
+  end
+
+  defp maybe_prefix_shell(_, cmd, args, _opts) do
+    {cmd, args}
+  end
+
+  defp find_executable(cmd, cd) do
+    path =
```
josevalim, line 84: `cmd` is either an absolute path or something in the OS path. So i think you need this:

```
if Path.type(cmd) == :relative do
  System.find_executable(cmd)
else
  cmd
end
```

No need to mangle `cd`. :)

lib/phoenix/endpoint/watcher.ex
```diff
@@ -57,4 +59,39 @@ defmodule Phoenix.Endpoint.Watcher do
...
+
+  defp maybe_prefix_shell({:win32, _}, cmd, args, opts) do
+    # This lets us call batch files on Windows, which aren't proper executables and can't be simply
+    # passed to `System.cmd`.
+    case find_executable(cmd, opts[:cd]) do
+      nil ->
+        {cmd, args}
+
+      path ->
+        if String.ends_with?(path, [".cmd", ".bat"]) do
```
josevalim, line 71:
```suggestion
        if Path.extname(path) in [".cmd", ".bat"] do
```

?

lib/phoenix/endpoint/watcher.ex
```diff
@@ -57,4 +59,39 @@ defmodule Phoenix.Endpoint.Watcher do
...
+          {"cmd", ["/c", Path.basename(path)] ++ args}
```
josevalim, line 72: Are you sure we want the basename here? And not the full path to the script? 🤔

eteeselink:
Hey, sorry about taking so long. I've had some pandemic related family setbacks. I still very much intend to address your comments and get this thing mergeable.

josevalim:
@eteeselink sorry to hear and no worries! Thanks for the updates! :heart:

chrismccord:
ping @eteeselink . Are you in a place to move this forward? No worries if this needs to wait. Thanks!

chrismccord:
@eteeselink if you are interested in moving this across the finish line, please let us know! Closing in the meantime. Thanks!

PR #4646, opened 2021-12-22
docs: add missing error for `Phoenix.Token.verify/4`

c4710n:
There are three types of errors, this PR adds the missing one. ;)

### DIFF
--- a/lib/phoenix/token.ex
+++ b/lib/phoenix/token.ex
@@ -163,7 +163,7 @@ defmodule Phoenix.Token do
   returned a tuple of type `{:ok, user_id}`. The server can now proceed with
   the request.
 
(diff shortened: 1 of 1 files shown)


lib/phoenix/token.ex
```diff
@@ -163,7 +163,7 @@ defmodule Phoenix.Token do
   returned a tuple of type `{:ok, user_id}`. The server can now proceed with
   the request.
 
-  However, if the client had sent an expired or otherwise invalid token
+  However, if the client had sent an expired or invalid token or otherwise `nil`,
```
josevalim, line 166:
```suggestion
  However, if the client had sent an expired token, an invalid token, or `nil`,
```

josevalim:

PR #4649, opened 2021-12-26
Fix broken `check_origin: :conn`

c4710n:
In the internal implementation, the related function is at [here](https://github.com/phoenixframework/phoenix/blob/8d2b33ac9691bd624ede602088d213f89600d233/lib/phoenix/socket/transport.ex#L596):

```elixir
defp origin_allowed?(:conn, uri, _endpoint, %Plug.Conn{} = conn)
```

`uri` is a struct which is generated by `URI.parse/1`, the type of `scheme`
field is a string.

`conn` is a `%Plug.Conn{}`, the type of `scheme` field is an atom.

Because of that, this check will always failed in real case.

* * *

This commit fixes above issue, and fix related tests.

### DIFF
--- a/lib/phoenix/socket/transport.ex
+++ b/lib/phoenix/socket/transport.ex
@@ -593,8 +593,13 @@ defmodule Phoenix.Socket.Transport do
 
   defp origin_allowed?({module, function, arguments}, uri, _endpoint, _conn),
     do: apply(module, function, [uri | arguments])
-  defp origin_allowed?(:conn, uri, _endpoint, %Plug.Conn{} = conn),
-    do: uri.host == conn.host and uri.scheme == conn.scheme and uri.port == conn.port
+
+  defp origin_allowed?(:conn, uri, _endpoint, %Plug.Conn{} = conn) do
+    uri.host == conn.host and
+      uri.scheme == Atom.to_string(conn.scheme) and
+      uri.port == conn.port
+  end
+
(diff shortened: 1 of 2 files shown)


lib/phoenix/socket/transport.ex
```diff
@@ -593,8 +593,13 @@ defmodule Phoenix.Socket.Transport do
 
   defp origin_allowed?({module, function, arguments}, uri, _endpoint, _conn),
     do: apply(module, function, [uri | arguments])
-  defp origin_allowed?(:conn, uri, _endpoint, %Plug.Conn{} = conn),
-    do: uri.host == conn.host and uri.scheme == conn.scheme and uri.port == conn.port
+
+  defp origin_allowed?(:conn, uri, _endpoint, %Plug.Conn{} = conn) do
+    uri.host == conn.host and
+      String.to_atom(uri.scheme) == conn.scheme and
```
josevalim, line 599: `String.to_atom` is dangerous We can either do `String.to_existing_atom` or convert `conn.scheme` to a string. Thoughts?

c4710n: I prefer the later one - converting `conn.scheme` to a string by `Atom.to_string/1`.

Is that fine?

josevalim: Yes, please!

c4710n: Done!

chrismccord:
❤️❤️❤️🐥🔥

PR #4650, opened 2021-12-27
Fix docker stop timeout with phx.gen.release

AstolfoKawaii:
Use _exec form_ instead of _shell form_ [https://docs.docker.com/engine/reference/builder/](https://docs.docker.com/engine/reference/builder/) because the _shell form_ does not forward signals (used by `docker stop`).


chrismccord:
❤️❤️❤️🐥🔥

PR #4653, opened 2022-01-08
Context Guide: Add "Category" alias in catalog.ex

hamvocke:
I was following the [In-context relationships guide](https://hexdocs.pm/phoenix/contexts.html#in-context-relationships) and noticed a problem when testing things out.

Creating a new product or editing an existing one would run into an `UndefinedFunctionError` in my `catalog.ex`:

```
function Category.__schema__/1 is undefined (module Category is not available)
```

The error appears here:

```elixir
def list_categories_by_id(category_ids) do
  Repo.all(from c in Category, where: c.id in ^category_ids)
end
```

Adding an `alias Hello.Catalog.Category` in my `catalog.ex` context solves the problem.

I didn't find the `alias` call being mentioned in the Context guide. This PR adds that `alias` call to the documentation, hoping that it'll help improve the guide a small bit.

### DIFF
--- a/guides/contexts.md
+++ b/guides/contexts.md
@@ -449,6 +449,8 @@ With our schema associations set up, we can implement the selection of categorie
 
 
 ```diff
++ alias Hello.Catalog.Category
+
 - def get_product!(id), do: Repo.get!(Product, id)
 + def get_product!(id) do
 +   Product |> Repo.get(id) |> Repo.preload(:categories)


josevalim:
Thanks for the PR! 

There is an alias on line 431: https://github.com/phoenixframework/phoenix/blob/1ca3c242f58d8bb8210c64954e738e04ded729ac/guides/contexts.md?plain=1#L431

Is that not the alias you need? :)

hamvocke:
It sure is :) Sorry, I missed that.

Maybe it was unfortunate that `diff` code blocks don't use syntax highlighting on Hexdocs.

GitHub's Markdown code preview highlights additions in green, removals in red. Having the same on Hexdocs might have made spotting my mistake easier:

| GitHub | Hexdocs |
| - | - |
|diffHighlights | diffNoHighlight |

Anyways, that's only tangentially related. Thanks for the pointer and sorry again for my silly mistake.

josevalim:
In case you want to help solve this problem, syntax highlighting is done with `makeup`. So you could implement `makeup_diff` project that handles diffs. The list of existing projects is here: https://github.com/elixir-makeup/

PR #4659, opened 2022-01-14
Add upgrade instructions to phx.gen.live task

### LINKED ISSUE #4642: Warn on phx.gen.live if LiveView.JS is missing?
The latest generator relies on `Phoenix.LiveView.JS` which requires LiveView v0.17+. Without the module, running `mix phx.server` after running the generator will emit a compilation error:

```sh
== Compilation error in file lib/fiddle_web/live/live_helpers.ex ==
** (CompileError) lib/fiddle_web/live/live_helpers.ex:55: Phoenix.LiveView.JS.__struct__/1 is undefined, cannot expand struct Phoenix.LiveView.JS. Make sure the struct name is correct. If the struct name exists and is correct but it still cannot be found, you likely have cyclic module usage in your code
    lib/fiddle_web/live/live_helpers.ex:55: (module)
```

Should we check whether or not `LiveView.JS` is loaded and print a warning for the user to upgrade? They will need to update deps for LiveView and LiveDashboard:

```elixir
defp deps do
 [
    {:phoenix_live_view, "~> 0.17"},
    {:phoenix_live_dashboard", "~> 0.6"}
    # deps...
  ]
end
```

mcrumm:
Closes #4642

### DIFF
--- a/lib/mix/tasks/phx.gen.live.ex
+++ b/lib/mix/tasks/phx.gen.live.ex
@@ -101,6 +101,23 @@ defmodule Mix.Tasks.Phx.Gen.Live do
       Mix.raise "mix phx.gen.live must be invoked from within your *_web application root directory"
     end
 
+    unless Code.ensure_loaded?(Phoenix.LiveView.JS) do
+      Mix.raise """
+      mix phx.gen.live requires :phoenix_live_view v0.17+.
+
+      Update your LiveView (and, if present, LiveDashboard) deps in mix.exs:
+
+          def deps do
+            [
+              {:phoenix_live_view, "~> 0.17.0"},
+              {:phoenix_live_dashboard, "~> 0.6.0"}
+              ...
+            ]
+          end
+
+      """
+    end
+
     {context, schema} = Gen.Context.build(args)
     Gen.Context.prompt_for_code_injection(context)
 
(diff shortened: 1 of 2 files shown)


lib/mix/tasks/phx.gen.live.ex
```diff
@@ -219,6 +219,31 @@ defmodule Mix.Tasks.Phx.Gen.Live do
...
+      Mix.shell().info """
+
+      Note the generated code requires Phoenix.LiveView v0.17+.
+
+      Before proceeding, update your :phoenix_live_view and,
+      if present, :phoenix_live_dashboard deps in mix.exs:
+
+          def deps do
+            [
+              {:phoenix_live_view, "~> 0.17.0"},
```
josevalim, line 236: This will get outdated. :( Should we say "we require Phoenix.LiveView v0.17+ and Phoenix.LiveDashboard v0.6+" and leave at that? We could also specify said requirements but maybe too broad?

mcrumm: Oops I forgot to leave a note saying to reword as you see fit :)

> This will get outdated

I was concerned about that, too :(

lib/mix/tasks/phx.gen.live.ex
```diff
@@ -219,6 +219,31 @@ defmodule Mix.Tasks.Phx.Gen.Live do
...
+              {:phoenix_live_dashboard, "~> 0.6"},
+              ...
+            ]
+          end
```
josevalim, line 240:
```suggestion
      You must update :phoenix_live_view to v0.17 or later and
      :phoenix_live_dashboard to v0.6 or later to use the features
      in this generator.
```

mcrumm: done!

josevalim (approved):

PR #4660, opened 2022-01-15
Normalise phx.gen.auth function declaration names

0urobor0s:
Use `schema.singular` on the context function declaration names as most of then already use it.
This is useful when the `phx.gen.auth` is used multiple times, for example:
```
mix phx.gen.auth Accounts User users --web auth
mix phx.gen.auth Accounts Admin admins --web admin
```

In test file, I'm not sure what is the best description for the `inspect/2` function.


josevalim:

PR #4665, opened 2022-01-21
Prefix unused variable with an underscore

quentin-bettoum:
Prefix unused "cart" variable with an underscore in contexts guide.


mcrumm:
Thanks!

PR #4667, opened 2022-01-23
Add flash function component to mix phx.new generator.

paulo-silva:
The idea is to have a function component to `flash` messages, so when no messages are present, we avoid printing an empty `<p>` tag.

```heex
<!-- input -->
<.flash type="info" message="User created." />

<!-- output -->
<p class="alert alert-info" role="alert">User created.</p>
```


priv/templates/phx.gen.live/live_helpers.ex
```diff
@@ -4,6 +4,26 @@ defmodule <%= inspect context.web_module %>.LiveHelpers do
...
+  If message is empty, the element is not rendered.
+
+  ## Examples
+
+      <.flash type="info" message="User created" />
+  """
+  def flash(assigns) do
+    ~H"""
+    <%= if is_binary(@message) and byte_size(String.trim(@message)) > 0 do %>
+      <p class={"alert alert-#{@type}"} role="alert"><%= @message %></p>
```
feliperenan, line 22:
```suggestion
      <p class={"alert alert-#{@type}"} role="alert">
        <%= @message %>
      </p>
```

That's how it would be formatted by the formatter :)

paulo-silva: yeah, good point 😄

priv/templates/phx.gen.live/live_helpers.ex
```diff
@@ -4,6 +4,28 @@ defmodule <%= inspect context.web_module %>.LiveHelpers do
...
+  inner HTML.
+  If message is empty, the element is not rendered.
+
+  ## Examples
+
+      <.flash type="info" message="User created" />
+  """
+  def flash(assigns) do
+    ~H"""
+    <%= if is_binary(@message) and byte_size(String.trim(@message)) > 0 do %>
```
josevalim, line 21: You can move the if outside of HEEx. I also don't think you need to trim the message, as you shouldn't have empty strings as messages. :)

priv/templates/phx.gen.live/live_helpers.ex
```diff
@@ -4,6 +4,28 @@ defmodule <%= inspect context.web_module %>.LiveHelpers do
...
+      <p class={"alert alert-#{@type}"} role="alert">
+        <%= @message %>
+      </p>
+    <% end %>
+    """
```
josevalim, line 26:
```suggestion
    if assigns.message do
      ~H"""
      <p class={"alert alert-#{@type}"} role="alert">
       <%= @message %>
      </p>
       """
    end
```

chrismccord:
Thanks for this! We need to also handle the `live.html.heex` layout. I am thinking something that accepts the entire flash would be better, which is what I do in my own applications, for example:

# live.html.heex
```elixir
    <.flash flash={@flash} kind={:info}/>
    <.flash flash={@flash} kind={:error}/>
```

# app.html.heex
```elixir
    <.flash flash={get_flash(@conn)} kind={:info}/>
    <.flash flash={get_flash(@conn)} kind={:error}/>
```

Thoughts?

paulo-silva:
> Thanks for this! We need to also handle the `live.html.heex` layout. I am thinking something that accepts the entire flash would be better, which is what I do in my own applications, for example:
> # live.html.heex
> 
> ```elixir
>     <.flash flash={@flash} kind={:info}/>
>     <.flash flash={@flash} kind={:error}/>
> ```
> 
> # app.html.heex
> 
> ```elixir
>     <.flash flash={get_flash(@conn)} kind={:info}/>
>     <.flash flash={get_flash(@conn)} kind={:error}/>
> ```
> 
> Thoughts?

oh, it looks great to me! we don't need to care about asking for the specific flash key while using the flash function component, we can do it inside of the function! :)

Thanks for the idea @chrismccord 🙇, I'll implement that

chrismccord:
Also we don't currently generate the `LiveHelpers` module until the first time you run `mix phx.gen.live`, so we either need to put this function elsewhere, or generate the live_helpers file on phx.new

paulo-silva:
> Also we don't currently generate the `LiveHelpers` module until the first time you run `mix phx.gen.live`, so we either need to put this function elsewhere, or generate the live_helpers file on phx.new

oh, I see, we could have a file `helpers.ex` on root of `app_web`, but I'm not sure if only this function would deserve to create a new file, do you think it would be ok to generate live_helpers in `mix phx.new`?

chrismccord:
Let's use this moment to consolidate things. We can generate a single `MyAppWeb.Helpers` module in phx.new and it will replace `ErrorHelpers` and `LiveHelpers`. Out of the box the flash, modal, and error_tag helpers will be there. And while we're here, we can also replace `<%= error_tag ...` with `<.error` :)

Sorry I know this ballooned. Are you okay to make those changes?

paulo-silva:
> 

Definitely! yeah, not a problem at all :) - it's my first contribution to phoenix, so I'm more than glad to help 😄

I'll work in the proposed changes, thanks for the clarification 🙇

priv/templates/phx.gen.live/live_helpers.ex
```diff
@@ -4,6 +4,29 @@ defmodule <%= inspect context.web_module %>.LiveHelpers do
...
+  inner HTML, if any exists.
+
+  ## Examples
+
+      <.flash type="info" message="User created" />
+  """
+  def flash(assigns) do
+    message = Map.get(assigns.flash, assigns.kind)
+
+    if is_binary(message) do
```
v0idpwn, line 21:
```suggestion
    if message = Map.get(assigns.flash, assigns.kind) do
```
Currently, phoenix supports anything that implements `Phoenix.HTML.Safe` as flash messages. While I agree that displaying `nil` is not the best idea, the function component should probably support iolists, live components or anything else.

chrismccord: if the key is there, we should use it, so we can ignore the type check entirely

paulo-silva:
@chrismccord I'm unsure about my changes 😅 - would you mind taking a look and checking if I'm on the right track? thanks in advance 🙇

installer/templates/phx_web/views/helpers.ex
```diff
@@ -1,10 +1,79 @@
-defmodule <%= @web_namespace %>.ErrorHelpers do
+defmodule <%= @web_namespace %>.Helpers do
   @moduledoc """
   Conveniences for translating and building error messages.
   """<%= if @html do %>
 
   use Phoenix.HTML
 
+  import Phoenix.LiveView
+  import Phoenix.LiveView.Helpers
```
mcrumm, line 9: You can simplify this a bit with `use Phoenix.Component`:

```suggestion
  use Phoenix.Component
```

installer/templates/phx_web/views/helpers.ex
```diff
@@ -1,10 +1,79 @@
...
+
+  @doc """
+  Renders a live component inside a modal.
+
+  The rendered modal receives a `:return_to` option to properly update
+  the URL when the modal is closed.
+
+
+  """
+  def modal(assigns) do
```
feliperenan, line 42: We also have `live_helpers`: https://github.com/phoenixframework/phoenix/blob/master/priv/templates/phx.gen.live/live_helpers.ex#L1

I guess we should either move these functions to there or remove this that file if we want to keep it here 🤔

priv/templates/phx.gen.live/live_helpers.ex
```diff
@@ -1,60 +0,0 @@
...
-      <.modal return_to={Routes.<%= schema.singular %>_index_path(@socket, :index)}>
-        <.live_component
-          module={<%= inspect context.web_module %>.<%= inspect Module.concat(schema.web_namespace, schema.alias) %>Live.FormComponent}
-          id={@<%= schema.singular %>.id || :new}
-          title={@page_title}
-          action={@live_action}
-          return_to={Routes.<%= schema.singular %>_index_path(@socket, :index)}
-          <%= schema.singular %>: @<%= schema.singular %>
-        />
-      </.modal>
```
aptinio, line 24: This example got left out.

chrismccord:
This exists in `core_components.ex` on master. Thanks!

PR #4670, opened 2022-01-26
Switch to .mjs files for ES modules

mveytsman:
This fixes #4662

It looks like Node doesn't treat .js files as being ES modules unless we set the filename to `.mjs` or set `"type": "module"` in the package.json (see https://nodejs.org/api/packages.html#determining-module-system)

I read through this [Hybrid npm packages (ESM and CommonJS)](https://2ality.com/2019/10/hybrid-npm-packages.html )and spot checked a few package.json files from js frameworks that target the browser and support both esm and cjs

- [Vue](https://github.com/vuejs/core/blob/main/packages/vue/package.json) - no type in package.json, uses .mjs files
- [Svelte](https://github.com/sveltejs/svelte/blob/master/package.json) - no type in package.json, uses .mjs files
- [Alpine](https://github.com/alpinejs/alpine/blob/main/packages/alpinejs/package.json)  - no type in package.json, uses .esm.js but I had the same issues as #4662 getting it to work inside a svelte project.

Based on the above I think this change solves the problem in the most backwards-compatible way.

PR #4671, opened 2022-01-26
Get rid of top-level `this` in constants.js

mveytsman:
This is a quick drive-by I'm making while I was building the javascript in #4670 

Building our javascript gives the following warning:

```
 > js/phoenix/constants.js:3:49: warning: Top-level "this" will be replaced with undefined since this file is an ECMAScript module
     3 │ export const global = globalSelf || phxWindow || this
```

Based on [globalThis - JavaScript | MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/globalThis#search_for_the_global_across_environments) I this the intention here is to use `global` if `window` and `self` both fall through.


chrismccord:
❤️❤️❤️🐥🔥

PR #4675, opened 2022-01-31
Improve app.css formatting

oskarkook:
Looks nicer this way, I think :)


mcrumm:
Thanks, it is certainly more consistent this way :)

PR #4676, opened 2022-02-01
Set `live_socket_id` when restoring session from cookie

hlindset:
When restoring user sessions from the "remember me"-cookie, the `live_socket_id` value is not set, so the `log_out_*`-function does not disconnect sockets for users that are on a "restored" session.

This PR sets a `live_socket_id` on the session when creating a new session from the "remember me"-cookie and adds an assert in the test suite for this behaviour.

### DIFF
--- a/priv/templates/phx.gen.auth/auth_test.exs
+++ b/priv/templates/phx.gen.auth/auth_test.exs
@@ -102,6 +102,7 @@ defmodule <%= inspect auth_module %>Test do
         |> <%= inspect schema.alias %>Auth.fetch_current_<%= schema.singular %>([])
 
       assert get_session(conn, :<%= schema.singular %>_token) == <%= schema.singular %>_token
+      assert get_session(conn, :live_socket_id) == "<%= schema.plural %>_sessions:#{Base.url_encode64(<%= schema.singular %>_token)}"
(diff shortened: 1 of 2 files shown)


priv/templates/phx.gen.auth/auth_test.exs
```diff
@@ -102,6 +102,7 @@ defmodule <%= inspect auth_module %>Test do
         |> <%= inspect schema.alias %>Auth.fetch_current_<%= schema.singular %>([])
 
       assert get_session(conn, :<%= schema.singular %>_token) == <%= schema.singular %>_token
+      assert get_session(conn, :live_socket_id) == "<%= schema.plural %>_sessions:#{Base.url_encode64(<%= schema.singular %>_token)}"
       assert conn.assigns.current_<%= schema.singular %>.id == <%= schema.singular %>.id
```
josevalim, line 106:
```suggestion
      assert conn.assigns.current_<%= schema.singular %>.id == <%= schema.singular %>.id
      assert get_session(conn, :<%= schema.singular %>_token) == <%= schema.singular %>_token

      assert get_session(conn, :live_socket_id) ==
               "<%= schema.plural %>_sessions:#{Base.url_encode64(<%= schema.singular %>_token)}"
```

josevalim:

PR #4677, opened 2022-02-02
Update security practices

josevalim:
* Include same_site: "lax". Albeit the default,
  it is not the default in Safari

* x-xss-protection has been removed in most browsers
  in favor of Content Security Policy

* cross-origin-window-policy was not actualy implemented
  in browsers and SameSite handles more cases

### DIFF
--- a/installer/templates/phx_web/endpoint.ex
+++ b/installer/templates/phx_web/endpoint.ex
@@ -7,7 +7,8 @@ defmodule <%= @endpoint_module %> do
   @session_options [
     store: :cookie,
     key: "_<%= @web_app_name %>_key",
-    signing_salt: "<%= @signing_salt %>"
+    signing_salt: "<%= @signing_salt %>",
+    same_site: "Lax"
   ]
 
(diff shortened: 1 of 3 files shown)


josevalim:
@GriffinMB, can you please double check if those look good for you?

GriffinMB (approved):
Would probably be worth including a test to verify SameSite is included by default as expected. Otherwise looks good to me!

josevalim:
Thanks! The same_site was added to generated new app code, so we can only test the generated code itself and I am not sure that will add much. :)

josevalim:

AviKav:
> * cross-origin-window-policy was not actualy implemented
>   in browsers and SameSite handles more cases

As far as I can tell, `SameSite` doesn't cover the same semantics/features as `Cross-Origin-Opener-Policy` (What `Cross-Origin-Window-Policy` was ultimately named)
https://chromestatus.com/feature/5432089535053824

I also only just learned that same-site != same-origin

PR #4679, opened 2022-02-03
Add ecto.seed alias to generator

nelsonmestevao:
I create an `ecto.seed` alias in every app that I use, I think it is very useful to make it available in the aliases because a lot of the time we want to add more seeding data without necessary running any migration.

It is also much easier and fast to type.

### DIFF
--- a/installer/templates/phx_ecto/seeds.exs
+++ b/installer/templates/phx_ecto/seeds.exs
@@ -1,6 +1,6 @@
 # Script for populating the database. You can run it as:
 #
-#     mix run priv/repo/seeds.exs
+#     mix ecto.seed
 #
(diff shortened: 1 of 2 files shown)


wojtekmach:
I think this is a good addition but only if your seed.exs can be run multiple times. In my that wasn't a case more often than not. For example, in seeds I'd create unique records like the admin user and stuff like that. Now, I'd sometimes have seed.exs which would first truncate all the relevant tables so that I could easily seed multiple times but I feel like more often I'd just go ahead and run (a bit slower) `$ mix ecto.reset` instead.

chrismccord:
I actually like that we are able to show an example of `mix run`, and you don't need to invoke this too often, so the current is preferable. Thanks!

PR #4682, opened 2022-02-05
Use optimized multicolumn index as an example

sato11:
Hi. First of all thanks for preparing a substantial guide. It's helping me a lot to get introduced to phoenix.

In this PR I'd like to propose that we use the optimized schema design in guides. The focus is on multicolumn index.

The current version promotes to create one index for each foreign key plus one composite unique index, which to me seems too much. Since PostgreSQL (and MySQL as well) can use leftmost prefix of multicolumn index to plan query, two indexes are just enough, and the third single-column index becomes only the source of overhead on write. This page in PostgreSQL Tutorial can be helpful for that behavior:  https://www.postgresqltutorial.com/postgresql-indexes/postgresql-multicolumn-indexes/

That's one thing and the other is about the column order of the composite index. The current version defines `[:product_id, :category_id]` but I think this should be reversed.

Because naturally the amount of categories compared to that of products is expected to be permanently smaller, we can say our `category_id` has lower cardinality than `product_id`. It's generally nice to order columns based on the order of cardinality so the data is more efficiently arranged in the index. That's why I have dropped the index on `category_id` and not the other way round.

So, that's my intensions. To be honest I was not sure if we should be aware of these things at this point in this guide, but then also I as a reader feel the style of this guide is trying to be inclusive about web development's better practices, and eventually I come to think it does only good that we take care a little about schema design at this point. After all, one day someone might refer to this guide to design their own real-world app, and design indexes based on this, then I would be more glad if this guide could present them a careful option.

Also I notice that there is a variation of this issue in the following section, where we create `cart_items`. If this looks good I'd like to move on, but before that I'd like to ask for review.


chrismccord:
❤️❤️❤️🐥🔥

PR #4685, opened 2022-02-10
Prioritize user `:http` configuration for ranch

sax:
This puts an application's `:http` configuration for their Endpoint
first in the list to be sent into the ranch supervisor. When running on
a newer commit of ranch, this allows an application to specific
`:inet_backend` for the TCP sockets spawned by ranch.

Without this change, the application configuration is always _after_
port and IP. Since `:inet_backend` is only valid when specified as the
first argument to `:gen_tcp.listen/2`, this results in a `:badarg`
error.

### DIFF
--- a/lib/phoenix/endpoint/cowboy2_adapter.ex
+++ b/lib/phoenix/endpoint/cowboy2_adapter.ex
@@ -61,6 +61,7 @@ defmodule Phoenix.Endpoint.Cowboy2Adapter do
           raise "aborting due to nil port"
         end
 
+        # Ranch options are read from the top, so we keep the user opts first.
         opts = :proplists.delete(:port, opts) ++ [port: port_to_integer(port), otp_app: otp_app]
         child_spec(scheme, endpoint, opts)
       end
--- a/lib/phoenix/endpoint/cowboy2_adapter.ex
+++ b/lib/phoenix/endpoint/cowboy2_adapter.ex
@@ -61,7 +61,7 @@ defmodule Phoenix.Endpoint.Cowboy2Adapter do
           raise "aborting due to nil port"
         end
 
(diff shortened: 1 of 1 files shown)


sax:
This is pretty specific, so I could see if you did not want this back in the main branch. My use case is fairly clear, though: as a part of a load test I'm currently doing, I would like to switch `gen_tcp` sockets to use the new `socket` implementation. That only works if `inet_backend` is the first option passed to `:gen_tcp.listen/2`.

This pull request makes this possible, through the undocumented interaction between Phoenix and Ranch. 😅

I was considering doing more work to make this a real feature that could be specified in endpoint config, but that runs into a couple of issues:
* This only works with a version of Ranch committed 5 months ago.
* Phoenix pins a decently older Ranch.
* The `inet_backend` configuration of sockets in OTP is temporary, and presumably will be ignored in a soon-to-be version of OTP.

Let me know what you think! Also, no worries if this is closed without merging. It is a change to allow me to exploit a loophole.

lib/phoenix/endpoint/cowboy2_adapter.ex
```diff
@@ -61,7 +61,7 @@ defmodule Phoenix.Endpoint.Cowboy2Adapter do
           raise "aborting due to nil port"
         end
 
-        opts = [port: port_to_integer(port), otp_app: otp_app] ++ :proplists.delete(:port, opts)
+        opts = :proplists.delete(:port, opts) ++ [port: port_to_integer(port), otp_app: otp_app]
```
josevalim, line 64:
```suggestion
        # Ranch options are read from the top, so we keep the user opts first.
        opts = :proplists.delete(:port, opts) ++ [port: port_to_integer(port), otp_app: otp_app]
```

josevalim:

sax:
Yay! Thank you!

Issue #4669, opened 2022-01-26
Do not generate asset handling in docker file if no asset directory exists


chrismccord:
This accommodates users who run with `--no-assets`

PR #4688, opened 2022-02-15
Don't generate asset handling in Docker when directory doesn't exist

timothyvanderaerden:
Checks if the `assets` directory is present and adds asset handling to the Dockerfile if needed.
Not sure if this is the desired direction.

Ref: #4669


chrismccord:
❤️❤️❤️🐥🔥

PR #4692, opened 2022-02-17
Config url scheme to "https" to match port 443

entertainyou:
This fixes the following unexpected log upon start
```
Access EuronextWeb.Endpoint at http://example.com:443
```

### DIFF
--- a/installer/templates/phx_single/config/runtime.exs
+++ b/installer/templates/phx_single/config/runtime.exs
@@ -29,7 +29,7 @@ if config_env() == :prod do
   port = String.to_integer(System.get_env("PORT") || "4000")
 
   config :<%= @app_name %>, <%= @endpoint_module %>,
-    url: [host: host, port: 443],
+    url: [host: host, port: 443, scheme: "https"],
     http: [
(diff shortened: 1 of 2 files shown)


mcrumm:
Thanks!

chrismccord:
I thought we inferred the scheme from the port? Can we double check this wasn't an issue on the log output side? For example, the `_url` route helpers should generate https without this addition, no?

mcrumm:
I will double-check!

mcrumm:
> ...the _url route helpers should generate https without this addition, no?

Unfortunately no :(


We derive `scheme` and `port` based on the `:http` and `:https` keys on the Endpoint config:

https://github.com/phoenixframework/phoenix/blob/8833ecc9b36b6f13aeb4a5809ca937169104f28d/lib/phoenix/endpoint/supervisor.ex#L284-L300


The log output _just_ prints `endpoint.url()`:

https://github.com/phoenixframework/phoenix/blob/8833ecc9b36b6f13aeb4a5809ca937169104f28d/lib/phoenix/endpoint/supervisor.ex#L438

Issue #4637, opened 2021-12-18
Change "use Phoenix.HTML" to "import Phoenix.HTML" and "import Phoenix.HTML.Form"


josevalim:
The goal is to improve the clarity over what is imported. Don’t send a pull request for this yet as it requires https://github.com/phoenixframework/phoenix_html/issues/372 to be completed and many helpers to be rewritten.

fmterrorf:
Hello José :wave: I've created a PR above to address this issue

PR #4701, opened 2022-02-24
Change "use Phoenix.HTML" to "import Phoenix.HTML" and "import Phoenix.HTML.Form"

fmterrorf:
Subject says it all. Only merge when https://github.com/phoenixframework/phoenix_html/issues/372 is completed

Closes https://github.com/phoenixframework/phoenix/issues/4637


josevalim:
Hi @fmterrorf! Thanks for the PR but unfortunately Many changes will be necessary to make this work, as now many HTML helpers used by the generators won’t be available. I will leave a note in the original issue mentioning that we need to wait, unfortunately.

PR #4702, opened 2022-02-25
Update "phx.gen.live"-related docs

dannyh79:
It's all document changes. :sweat_smile: 

## Changed
- Convert live_helpers.ex template modal/1 doc's EEX Syntax to HEEX
- Update phx.gen.live module doc to prune removed "ModalComponent"

### DIFF
--- a/priv/templates/phx.gen.live/live_helpers.ex
+++ b/priv/templates/phx.gen.live/live_helpers.ex
@@ -18,8 +18,8 @@ defmodule <%= inspect context.web_module %>.LiveHelpers do
           id={@<%= schema.singular %>.id || :new}
           title={@page_title}
           action={@live_action}
+          <%= schema.singular %>={@<%= schema.singular %>}
           return_to={Routes.<%= schema.singular %>_index_path(@socket, :index)}
-          <%= schema.singular %>: @<%= schema.singular %>
         />
       </.modal>
   """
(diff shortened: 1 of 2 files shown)


priv/templates/phx.gen.live/live_helpers.ex
```diff
@@ -19,7 +19,7 @@ defmodule <%= inspect context.web_module %>.LiveHelpers do
           title={@page_title}
           action={@live_action}
           return_to={Routes.<%= schema.singular %>_index_path(@socket, :index)}
-          <%= schema.singular %>: @<%= schema.singular %>
+          <%= schema.singular %>={@<%= schema.singular %>}
```
c4710n, line 22: I am trying to change this in another PR, too.

The generated code is something like:

```
    <.live_component
      module={AdminWeb.VersionLive.FormComponent}
      id={@version.id}
      title={@page_title}
      action={@live_action}
      version={@version} <- schema.singular
      return_to={Routes.version_show_path(@socket, :show, @version)}
    />
```

As you can see, `schema.singular` comes before `return_to`. I think it would be better to adjust the order. Then, I can close my own PR. ;)

dannyh79: > I think it would be better to adjust the order

Not sure if I follow it. Would you mind providing the reasoning behind the ordering change in #4704?

c4710n: [The template for generating the final code](https://github.com/phoenixframework/phoenix/blob/bc6ece3a632552f2114441d4ff306f7d044ff0fe/priv/templates/phx.gen.live/show.html.heex#L10) is using this order.

dannyh79: Got it. Sure it maybe could use some adjustments. :+1:

josevalim:

PR #4703, opened 2022-02-25
fix incorrect html

tilleryd:
This PR fixes outdated HTML in the Request Lifecycle Guide. Without this change, a user will see an empty section on the hello page which makes the guide and/or framework seem broken.


josevalim:

PR #4707, opened 2022-02-28
Handle presence metas

bemesa21:


lib/phoenix/presence.ex
```diff
@@ -343,29 +343,203 @@ defmodule Phoenix.Presence do
...
+            presence_diff
+          )
         end
       end)
 
       {:ok, state}
     end
   end
 
+  defmodule Client do
```
chrismccord, line 386: Rather than an extra process here, we can put everything that this genserver is doing inside of the `Phoenix.Presence` module. For example, `init` here will move up inside `Phoenix.Presence.init`, and then we can handle the state inside the same process without extra copying

lib/phoenix/presence.ex
```diff
@@ -343,29 +343,203 @@ defmodule Phoenix.Presence do
...
         for {topic, {joins, leaves}} <- diff do
-          Phoenix.Channel.Server.local_broadcast(pubsub_server, topic, "presence_diff", %{
+          presence_diff = %{
             joins: module.fetch(topic, Phoenix.Presence.group(joins)),
             leaves: module.fetch(topic, Phoenix.Presence.group(leaves))
-          })
+          }
+
+          if presence_client do
+            Phoenix.Presence.Client.handle_diff(topic, presence_diff)
```
chrismccord, line 370: we'll handle this inline in the same module here (can call a private function)

lib/phoenix/presence.ex
```diff
@@ -343,29 +343,203 @@ defmodule Phoenix.Presence do
...
       {:ok, state}
     end
   end
 
+  defmodule Client do
+    use GenServer
+    require Logger
+
+    @callback init(state :: term) :: {:ok, new_state :: term}
+    @callback handle_join(topic :: String.t(), key :: String.t(), meta :: [map()], state :: term) ::
```
chrismccord, line 391: as discussed, let's consolidate `handle_join` and `handle_leave` into `handle_metas`, which can look something like this in user land:

```elixir
      def handle_metas(topic, %{joins: _, leaves: _} = diff, presences, state) do
        for {user_id, meta} <- diff.joins do
          presence = Map.fetch!(presences, user_id)
          local_broadcast(topic, {__MODULE__, %{user_joined: presence}})
        end

        for leave <- diff.leaves do
          local_broadcast(topic, {__MODULE__, %{user_left: presence}})
        end
        {:ok, state}
      end
```

lib/phoenix/presence.ex
```diff
@@ -343,26 +354,182 @@ defmodule Phoenix.Presence do
...
+      Phoenix.Tracker.start_link(
+        __MODULE__,
+        {module, task_supervisor, pubsub_server},
+        opts
+      )
     end
 
-    def init(state) do
-      {:ok, state}
+    def init({module, task_supervisor, pubsub_server}) do
```
chrismccord, line 364: Since we have quite a bit of implementation here now, we should delegate back out of the injected callbacks (`init`, `handle_diff`) and all of the private functions into the `Phoenix.Presence` module. That way we aren't injecting all this code into the users module. For example, init could look like: 

`def init(state), do: Phoenix.Presence.__init__(state)` and so on.

In general in a `__using__` block, we should inject as little as code as possible. Also, I was guilty of violating this principle in the current/original `handle_diff` :)

Make sense?

chrismccord:
This was merged in [87f1405](https://github.com/phoenixframework/phoenix/commit/87f14057b6da601ad3a47f7c4ecb905ead227b28)

❤️❤️❤️🐥🔥

PR #4709, opened 2022-03-02
Deprecate {:system, env_var} tuples for configuring endpoint urls

aaronrenner:
I was reading through the endpoint supervisor code and saw the following TODO so I thought I'd try tackling it.
```
# TODO: Deprecate {:system, env_var} once we require Elixir v1.9+
```
This adds the following log message when an endpoint's `:url` or `:static_url` function has a deprecated `{:system, env_var}` tuple.

```
00:14:41.051 [warn] :url configuration containing values of {:system, env_var} tuples for Phoenix.Endpoint.EndpointTest.SystemTupleEndpoint is deprecated.

Configuration:

    config :phoenix, Phoenix.Endpoint.EndpointTest.SystemTupleEndpoint,
      [host: {:system, "ENDPOINT_TEST_HOST"}, port: {:system, "ENDPOINT_TEST_PORT"}]

Either move these config values into config/runtime.exs:

    config :phoenix, Phoenix.Endpoint.EndpointTest.SystemTupleEndpoint,
      host: System.get_env("ENDPOINT_TEST_HOST"),
      port: System.get_env("ENDPOINT_TEST_PORT")

Or read them during initialization inside of the Phoenix.Endpoint.EndpointTest.SystemTupleEndpoint.init/2 callback.

    def init(:supervisor, config) do
      config
      |> put_in([:url, :host], System.get_env("ENDPOINT_TEST_HOST"))
      |> put_in([:url, :port], System.get_env("ENDPOINT_TEST_PORT"))

      {:ok, config}
    end
```

Please let me know what you think!

### DIFF
--- a/lib/phoenix/endpoint/supervisor.ex
+++ b/lib/phoenix/endpoint/supervisor.ex
@@ -367,37 +367,26 @@ defmodule Phoenix.Endpoint.Supervisor do
     deprecated_configs = Enum.filter(conf[key] || [], &match?({_, {:system, _}}, &1))
 
     if Enum.any?(deprecated_configs) do
-      runtime_exs_config_lines =
-        for {key, {:system, env_var}} <- deprecated_configs do
-          ~s|#{key}: System.get_env("#{env_var}")|
-        end
-
-      init_suggestion_lines =
-        for {setting, {:system, env_var}} <- deprecated_configs do
-          ~s/|> put_in(#{inspect([key, setting])}, System.get_env("#{env_var}"))/
-        end
+      deprecated_config_lines = for {k, v} <- deprecated_configs, do: "#{k}: #{inspect(v)}"
+      runtime_exs_config_lines = for {key, {:system, env_var}} <- deprecated_configs, do: ~s|#{key}: System.get_env("#{env_var}")|
 
       Logger.warn """
-      #{inspect(key)} configuration containing values of {:system, env_var} tuples for #{inspect(mod)} is deprecated.
+      #{inspect(key)} configuration containing {:system, env_var} tuples for #{inspect(mod)} is deprecated.
 
-      Configuration:
-
-          config #{inspect(otp_app)}, #{inspect(mod)},
-            #{inspect(deprecated_configs)}
-
-      Either move these config values into config/runtime.exs:
+      Configuration with deprecated values:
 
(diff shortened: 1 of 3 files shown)


lib/phoenix/endpoint/supervisor.ex
```diff
@@ -350,15 +354,54 @@ defmodule Phoenix.Endpoint.Supervisor do
...
+
+          config #{inspect(otp_app)}, #{inspect(mod)},
+            #{inspect(deprecated_configs)}
+
+      Either move these config values into config/runtime.exs:
+
+          config #{inspect(otp_app)}, #{inspect(mod)},
+            #{runtime_exs_config_lines |> Enum.join(",\r\n      ")}
+
+      Or read them during initialization inside of the #{inspect(mod)}.init/2 callback.
```
josevalim, line 393: I would not necessarily mention this approach. I would go a 100% in on config/runtime.exs. WDYT?

aaronrenner: I was curious if I should include both approaches, so I appreciate your feedback. I just pushed an updated commit.

aaronrenner:
I just reworked the log message because I realized some of the code samples were wrong. Here's the updated log message:

```
08:16:54.563 [warn] :url configuration containing {:system, env_var} tuples for Phoenix.Endpoint.EndpointTest.SystemTupleEndpoint is deprecated.

Configuration with deprecated values:

    config :phoenix, Phoenix.Endpoint.EndpointTest.SystemTupleEndpoint,
      url: [
        host: {:system, "ENDPOINT_TEST_HOST"},
        port: {:system, "ENDPOINT_TEST_PORT"}
      ]

Move this configuration into config/runtime.exs and replace the {:system, env_var} tuples
with System.get_env/1 function calls:

    config :phoenix, Phoenix.Endpoint.EndpointTest.SystemTupleEndpoint,
      url: [
        host: System.get_env("ENDPOINT_TEST_HOST"),
        port: System.get_env("ENDPOINT_TEST_PORT")
      ]
```

josevalim (approved):

josevalim:
No need to backport imo.

mtrudel:
There are also an instance of this at https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/endpoint/cowboy2_adapter.ex#L124. I'll work up a quick PR for that one too.

PR #4711, opened 2022-03-04
[phx.gen.release] Skip generating migration files when ecto_sql is not installed

aaronrenner:
When running `mix phx.gen.release` in an app that has `ecto` installed but not `ecto_sql`, it still generate the migration-related files, causing compilation warnings and files that need to be deleted. This PR does the following:

* Instead of looking to ensure the `Ecto` module is loaded, it looks through `Application.loaded_applications` for `:ecto_sql`.
* Adds integration tests for `phx.gen.release` to test various combinations of `ecto` and `ecto_sql` being installed.
  * In order to properly test this, we needed to remove the unused deps that are copied in from the integration test project. This is done automatically, but I also added a `skip_clean_unused_deps: true` option to `generate_phoenix_app/4` in the cases where we want to delay this automatic cleanup.
* Adds the `--no-ecto` flag to skip generating the migration files even if `ecto_sql` is installed.
* Updated the documentation for the new options.

This PR is best reviewed commit by commit. The first commit updates the integration tests to clean up unused deps when generating a sample project (but has an option to disable this commit if deps are needed later like in phx.gen.auth). The second commit is where the issue is actually fixed in the migrator and the new `--no-ecto` option is added.

### DIFF
--- a/lib/mix/tasks/phx.gen.release.ex
+++ b/lib/mix/tasks/phx.gen.release.ex
@@ -21,7 +21,6 @@ defmodule Mix.Tasks.Phx.Gen.Release do
   To skip generating the migration-related files, use the `--no-ecto` flag. To
   force these migration-related files to be generated, the use `--ecto` flag.
 
-
   ## Docker
 
   When the `--docker` flag is passed, the following docker files are generated:
@@ -65,7 +64,7 @@ defmodule Mix.Tasks.Phx.Gen.Release do
       {:eex, "rel/server.bat.eex", "rel/overlays/bin/server.bat"}
     ])
 
-    if opts.ecto? do
+    if opts.ecto do
       Mix.Phoenix.copy_from(paths(), "priv/templates/phx.gen.release", binding, [
         {:eex, "rel/migrate.sh.eex", "rel/overlays/bin/migrate"},
         {:eex, "rel/migrate.bat.eex", "rel/overlays/bin/migrate.bat"},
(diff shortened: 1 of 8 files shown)


josevalim:
To be honest, I think the complexity we add to the tests is not worth what we are testing: a boolean check. So I would fix the issue and skip the integration tests. :)

aaronrenner:
> To be honest, I think the complexity we add to the tests is not worth what we are testing: a boolean check. So I would fix the issue and skip the integration tests. :)

@josevalim Thanks for the feedback and that makes sense. I removed the integration tests in 64a6567.

lib/mix/tasks/phx.gen.release.ex
```diff
@@ -18,6 +18,12 @@ defmodule Mix.Tasks.Phx.Gen.Release do
   Note, the `rel/overlays` directory is copied into the release build by default when
   running `mix release`.
 
+  To skip generating the migration-related files, use the `--no-ecto` flag. To
+  force these migration-related files to be generated, the use `--ecto` flag.
+
+
```
josevalim, line 24:
```suggestion
```

lib/mix/tasks/phx.gen.release.ex
```diff
@@ -143,6 +148,18 @@ defmodule Mix.Tasks.Phx.Gen.Release do
...
+  defp parse_args(args) do
+    args
+    |> OptionParser.parse!(strict: [ecto: :boolean, docker: :boolean])
+    |> elem(0)
+    |> Keyword.put_new_lazy(:ecto, &ecto_sql_installed?/0)
+    |> Keyword.put_new(:docker, false)
+    |> Map.new(fn
+      {:ecto, val} -> {:ecto?, val}
+      {:docker, val} -> {:docker?, val}
+    end)
```
josevalim, line 160:
```suggestion
    |> Map.new()
```

josevalim (approved):
Two small comments and ship it :)

lib/mix/tasks/phx.gen.release.ex
josevalim: Of course this requires changing the rest of the code but given that we don't rewrite options like this anywhere else, I would skip this. :)

josevalim (approved):

Issue #4683, opened 2022-02-07
protocol String.Chars not implemented for enum in mix phx.gen.context


jb3:
### Environment

* Elixir version (elixir -v): 1.13.0
* Phoenix version (mix deps): 1.6.6
* Operating system: Arch Linux, kernel 5.16.4

### Expected behavior

Similar to `mix phx.gen.schema`, `mix phx.gen.context` should be able to process the command:

```
$ mix phx.gen.context Accounts User users email:text:unique password:text:redact status:enum:verified:unverified:disable
```

### Actual behavior

```elixir
$ mix phx.gen.context Accounts User users email:text:unique password:text:redact status:enum:verified:unverified:disabled
* creating lib/portal/accounts/user.ex
* creating priv/repo/migrations/20220207191418_create_users.exs
* creating lib/portal/accounts.ex
** (Protocol.UndefinedError) protocol String.Chars not implemented for {:enum, :"verified:unverified:disabled"} of type Tuple
    (elixir 1.13.0) lib/string/chars.ex:3: String.Chars.impl_for!/1
    (elixir 1.13.0) lib/string/chars.ex:22: String.Chars.to_string/1
    (stdlib 3.17) erl_eval.erl:685: :erl_eval.do_apply/6
    (stdlib 3.17) erl_eval.erl:446: :erl_eval.expr/5
    (stdlib 3.17) erl_eval.erl:123: :erl_eval.exprs/5
    (stdlib 3.17) erl_eval.erl:270: :erl_eval.expr/5
    (stdlib 3.17) erl_eval.erl:446: :erl_eval.expr/5
    (stdlib 3.17) erl_eval.erl:123: :erl_eval.exprs/5
```

The `mix phx.gen.schema` of this command works absolutely fine.

chrismccord:
this necessarily involves some arbitrary HTML input decisions for representing and handling the enum, so we don't intend to accept enum types in the HTML generators. Thanks!

jb3:
Thanks for the response! Surely `mix phx.gen.schema` shouldn't be doing any HTML generation? It's just there to generate the context module for use however you want. `mix phx.gen.html` would be doing HTML generation, in which I can understand not supporting enums.

chrismccord:
bah, I read it as the html generator :)

milmazz:
@jb3 I think this is already fixed, I just tried in a new app created with `mix phx.new` and this is working:

```console
[...]

PR #4713, opened 2022-03-05
chore(gen.context): add unit test including enums

milmazz:
As commented [here](https://github.com/phoenixframework/phoenix/issues/4683#issuecomment-1059713876). The issue #4683 should be already fixed, but I didn't see any unit test to generate a context that includes enums like:

```
mix phx.gen.context Accounts User users email:text:unique password:text:redact status:enum:verified:unverified:disable
```

So, the idea behind this unit test is to avoid regressions in the future.


josevalim:

PR #4722, opened 2022-03-15
Add support for custom debuggers

tmbb:
This PR adds a new option to the `use MyApp.Endpoint` macro which allows you to choose a custom debugger instead of the default `Plug.Debugger` debugger. For example, this is me running a custom debugger with a different logo and syntax highlighting:

Homepage; let's click the naughty link:

image

Oh, no! You've found a bug. Look at how the Elixir code is highlighted:

image

But we can go deeper; note that HEEX is also highlighted:

image

And finally, it supports Erlang too:

image

I also tried to add support for replaying requests and automatically setting breakpoints, but it turns out it didn't work that well (in the naïve way I've implemented it). It is the kind of thing you can implement with custom debuggers.

Note: the custom debugger logo is a ghost that eats bugs (of course!) and it's a modification of the default Phoenix logo. I think it's different enough for it not to be considered copyright infringement, but if you think it is, I can replace it by something else.


aaronrenner:
@tmbb Thanks for the PR! It looks like there are quite a few formatting changes mixed in with your commit. Would you mind removing those formatting changes so the changes you are proposing are easier to review?

tmbb:
> Would you mind removing those formatting changes so the changes you are proposing are easier to review?

Ok, I'll have to turn off the format-on-save somehow, but I can revert the formatting changes

tmbb:
Done. Some tests are failing, but they succeed on my machine (and in my experience they're usually pretty flaky unfortunately)

lib/phoenix/endpoint.ex
```diff
@@ -62,6 +62,35 @@ defmodule Phoenix.Endpoint do
       YourAppWeb.Endpoint.config(:port)
       YourAppWeb.Endpoint.config(:some_config, :default_value)
 
+  ### Module-level configuration
```
josevalim, line 65: I don't think we should introduce a new type of configuration. Instead, we could use `debug_errors` to also accept an atom (or a tuple), besides booleans, and act accordingly. WDYT?

josevalim (changes requested):

lib/phoenix/endpoint.ex
tmbb: > I don't think we should introduce a new type of configuration.

Whenever possible, I prefer "module-level" configuration to "application" configuration (when it makes sense for the configuration to be done at compile-time, of course). But that's just a preference. But I guess I could have it either way.

> Instead, we could use debug_errors to also accept an atom (or a tuple), besides booleans, and act accordingly

I don't think we should reuse the `debug_errors` configuration option for this.  IMO, choosing whether to show the debugger when there is an error and which debugger to show are two different concerns. Allowing for the possibility that `debug_errors` can be either a boolean or a tuple also complicates the code with the current architecture. I've actually tried to implement this feature by making `debug_errors` a tuple, and I got into a pretty heavy mess between runtime and compile-time values which I couldn't solve at all (but that may be just my inexperience with compile-time elixir programming).

josevalim: Agreed that we don’t need to reuse the configuration but we should not introduce a new type of configuration just for this, so let’s move that part to the app. :)

Issue #4721, opened 2022-03-14
`mix phx.routes` for live_view pages are all shown as dispatched by `Phoenix.LiveView.Plug`


RudolfMan:
### Environment

* Elixir 1.13.3
* Phoenix 1.6.4 (also on "master"):
* Operating system: macOS Big Sur 11.6.2

For router defined as:
```elixir
  scope "/", AppWeb do
    pipe_through :browser

    scope "/blog" do
      live "/", PostLive.Index, :index
      live "/:id", PostLive.Show, :show
    end
  end
```
### Expected behavior
When running `mix phx.routes` for LiveView routes instead of `Phoenix.LiveView.Plug` it prints the live_view module name that the route is dispatched to.

```
$ mix phx.routes
    post_index_path  GET  /blog                    AppWeb.PostLive.Index :index
     post_show_path  GET  /blog/:id                AppWeb.PostLive.Show :show
live_dashboard_path  GET  /dashboard               Phoenix.LiveDashboard.PageLive :home
live_dashboard_path  GET  /dashboard/:page         Phoenix.LiveDashboard.PageLive :page
live_dashboard_path  GET  /dashboard/:node/:page   Phoenix.LiveDashboard.PageLive :page
```

### Actual behavior
For all "live" routes it prints `Phoenix.LiveView.Plug`

```
$ mix phx.routes
    post_index_path  GET  /blog                     Phoenix.LiveView.Plug :index
     post_show_path  GET  /blog/:id                 Phoenix.LiveView.Plug :show
live_dashboard_path  GET  /dashboard                Phoenix.LiveView.Plug :home
live_dashboard_path  GET  /dashboard/:page          Phoenix.LiveView.Plug :page
live_dashboard_path  GET  /dashboard/:node/:page    Phoenix.LiveView.Plug :page
```

I've put up together a codechange that allows passing a new option `module` to `Route` and can be set with "route_options". And when the route is built without that option we could fallback to `plug`. 
 - Code change for phoenix: https://github.com/phoenixframework/phoenix/commit/eeb0dd1f5983a8e12a092c654c114ca91fb44d45
 - And corresponding tiny change for phoenix_live_view: https://github.com/phoenixframework/phoenix_live_view/commit/9edafe5933ae8cc1454e7f2f41eb47111dd8ffc9

[...]

Issue #4686, opened 2022-02-10
Support log_module in router metadata


chrismccord:

mcrumm:
Closed by #4723 🎉

PR #4723, opened 2022-03-15
Support log_module in router metadata

RudolfMan:
Initially was working on https://github.com/phoenixframework/phoenix/issues/4721, but turns out there is already an issue related https://github.com/phoenixframework/phoenix/issues/4686


lib/phoenix/router/scope.ex
```diff
@@ -40,6 +40,7 @@ defmodule Phoenix.Router.Scope do
       opts
       |> Keyword.get(:metadata, %{})
       |> Map.put(:log, Keyword.get(opts, :log, top.log))
+      |> Map.put_new(:log_module, plug)
```
josevalim, line 43: I don't think we should put this by default, so we generate less metadata. Then in the formatter we do this:

    metadata[:log_module] || plug

WDYT?

josevalim:
Thank you for the PR! I dropped a comment. Note we also need to change here: https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/logger.ex#L282

lib/phoenix/router/scope.ex
RudolfMan: I like the idea! 
will make the change

josevalim:
Looks great! Also feel welcome to send a PR to LiveView that makes use of this new option. :)

josevalim:

PR #4731, opened 2022-03-22
Hide maybe_print_mailer_installation_instructions/1 from docs

aaronrenner:
This removes the `maybe_print_mailer_installation_instructions/1` function from the `mix phx.gen.notifier` docs. It seems like this is an internal phoenix function and not an API contract we want to maintain, I'm thinking it should probably be marked with `@doc false`.

If this is intentionally in the API docs so other libraries can depend on it, we can close this PR.

### DIFF
--- a/lib/mix/tasks/phx.gen.notifier.ex
+++ b/lib/mix/tasks/phx.gen.notifier.ex
@@ -177,10 +177,6 @@ defmodule Mix.Tasks.Phx.Gen.Notifier do
     |> Mix.Phoenix.prompt_for_conflicts()
   end
 
-  # Print mailer instructions if mailer is not defined.
-  #
-  # This is useful for applications that were created without the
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/phx.gen.notifier.ex
```diff
@@ -177,12 +177,11 @@ defmodule Mix.Tasks.Phx.Gen.Notifier do
...
-  @doc """
-  Print mailer instructions if mailer is not defined.
-
-  This is useful for applications that were created without the
-  mailer.
-  """
+  # Print mailer instructions if mailer is not defined.
+  #
+  # This is useful for applications that were created without the
+  # mailer.
```
josevalim, line 183:
```suggestion
```

josevalim (approved):

PR #4732, opened 2022-03-22
Use app name as from name for phx.gen.auth notifier

aaronrenner:
After running `mix phx.gen.auth` and registering a user, I was surprised to see the confirmation email come from the literal "MyApp" instead of the actual name of my app. This updates the generators to set from name to the actual name of the app.

## Question
I noticed `mix phx.gen.notifier` uses the from name "Phoenix Team" and email "team@example.com". 

https://github.com/phoenixframework/phoenix/blob/90a6494ffc028f2d5aa84a8a3d8f5a022fbd97d3/priv/templates/phx.gen.notifier/notifier.ex#L8

Do we want to update the default from name and email in both of these generators to match?

### DIFF
--- a/priv/templates/phx.gen.auth/notifier.ex
+++ b/priv/templates/phx.gen.auth/notifier.ex
@@ -8,7 +8,7 @@ defmodule <%= inspect context.module %>.<%= inspect schema.alias %>Notifier do
     email =
       new()
       |> to(recipient)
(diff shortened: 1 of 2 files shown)


nelsonmestevao (approved):

josevalim (approved):
Pls backport to v1.6 too!

PR #4733, opened 2022-03-23
[phx.gen.auth] Improve references to current module SchemaToken module

aaronrenner:
Currently `mix phx.gen.auth` creates a token module where its references to the current module look like this

```elixir
defmodule MyApp.Accounts.UserToken do
 # ...
    {token, %MyApp.Accounts.UserToken{token: token, context: "session", user_id: user.id}}
  #...
end
```
This seems pretty verbose, especially with a longer application name.

## My current approach

This PR switches the references to the current module from using the fully qualified name to use `__MODULE__` instead. This keeps these lines shorter, makes renaming this module easier, and reduces the formatting errors in the generated code if the application name is too long.

```elixir
defmodule MyApp.Accounts.UserToken do
 # ...
    {token, %__MODULE__{token: token, context: "session", user_id: user.id}}
  #...
end
```

## Another approach

I didn't find any other generated code that built structs for the current module, so I'm not sure what the best practices are. Another approach could be:

```elixir
defmodule MyApp.Accounts.UserToken do
  alias MyApp.Accounts.UserToken
 # ...
    {token, %UserToken{token: token, context: "session", user_id: user.id}}
  #...
end
```
This approach does repeat the current module's name when building the struct, but then sometimes I wonder "is this another module named UserToken from elsewhere, or is it the module I'm currently in?" I historically haven't taken this approach because  `__MODULE__` approach doesn't leave room for confusion and also removes the possibility of naming collisions.

## Questions

Which pattern is preferred for building a struct within the module that defines it?

### DIFF
--- a/priv/templates/phx.gen.auth/schema_token.ex
+++ b/priv/templates/phx.gen.auth/schema_token.ex
@@ -1,6 +1,7 @@
 defmodule <%= inspect schema.module %>Token do
   use Ecto.Schema
   import Ecto.Query
+  alias <%= inspect schema.module %>Token
 
   @hash_algorithm :sha256
   @rand_size 32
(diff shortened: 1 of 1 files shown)


josevalim:
I prefer "Another approach" because that's how we would refer to it if we moved the code to another module but any way is fine by me.

josevalim (approved):

PR #4735, opened 2022-03-25
Add .git to dockerignore

aaronrenner:
This PR adds `/.git/` to `.dockerignore` when running `mix phx.gen.release --docker`. The `.git` folder is not used in the `Dockerfile` we generate, so it seems unnecessary to have it included in the build context. Also, when running on a larger repo, the `.git` folder can become pretty big and it would be nice not to have to send it to the docker daemon.

This also removes `.dockerignore` from the list of excluded files because it's a small, single file and doesn't seem like it gives us much benefit to explicitly exclude it from the build context.

### DIFF
--- a/priv/templates/phx.gen.release/dockerignore.eex
+++ b/priv/templates/phx.gen.release/dockerignore.eex
@@ -14,10 +14,8 @@
 # More information on using .dockerignore is available here:
 # https://docs.docker.com/engine/reference/builder/#dockerignore-file
 
-.dockerignore
-# There are valid reasons to keep .git, namely so that you can get the
-# current commit hash
-#/.git/
+# Git repository
+/.git/
 
 # Common development/test artifacts
 /cover/


almirsarajcic (approved):
Seems like a reasonable default. 👍
We can always comment the line for the specific projects.

chrismccord:
I'll loop back to this one soon. I have some ideas for the future that may benefit from referencing the git hash, so including it by default now would be better, but TBD on whether it is necessary. Thanks!

Gazler:
You should be able to achieve this by only including `.git/HEAD` and `.git/refs`.

```
$ git show HEAD --pretty=full | head -n 1
commit d0b8727759e1e0e7aa3d41707d12376e373d5ecc

$ cat .git/HEAD
ref: refs/heads/master

$ cat .git/HEAD | awk '{print ".git/"$2}' | xargs cat
d0b8727759e1e0e7aa3d41707d12376e373d5ecc
```

PR #4736, opened 2022-03-26
Removes build_params/1 method since it is unused

ghousemohamed:
While going through the router code to understand its internals, I noticed that `build_params/1`
method is being unused. I thought there might be something I must be missing, so I removed
it and ran the test suite. But that did not seem to break anything. In case this method is needed for a
particular reason, I'd be happy to extend the test cases to account for this.

Thank you 😄


josevalim:

PR #4747, opened 2022-04-07
Use build arguments for docker base image versions

dvic:
This PR refactors the Dockerfile to use build arguments for the base images (with defaults). This allows for easier customisation in CI. For example in Ci we do something like:

```yaml
      - name: Docker build
        uses: docker/build-push-action@v2
        with:
          build-args: |
            ELIXIR_VERSION=${{ steps.asdf-versions.outputs.elixir }}
            ERLANG_VERSION=${{ steps.asdf-versions.outputs.erlang }}
```


chrismccord:
❤️❤️❤️🐥🔥

PR #4748, opened 2022-04-08
Add --info option to mix phx.routes

fchabouis:
Hi!

I wrote some code allowing a user to locate in its code base the controller function called, given an URL. Following @josevalim [suggestion](https://twitter.com/josevalim/status/1509408277460836354) to open a PR, here is a first step.

For the moment, I modified the `phx.routes.ex` file, allowing to pass a new option `--info` to the task. For example: 

```
mix phx.routes --info http://localhost:4000/

Module: Elixir.RouteInfoTestWeb.PageController
Function: index

/home/francis/projects/phoenix/installer/route_info_test/lib/route_info_test_web/controllers/page_controller.ex:4
``` 

Before going further (documentation, tests, edge cases,...) I would appreciate a first feedback about this proposition to know if it seems interesting to the Phoenix team and if the code (location, method used to find the code location) looks valid to you.

To locate the code, I use:
- `Phoenix.Router.route_info` to get the module and function called
- `module_info()` to get the file where the module is defined
- `Code.fetch_docs` to get the function definition line number in the file

Thank you


chrismccord:
This is looking great! For non plug routes, ie LiveViews, Phoenix supports a `:log_module` in the router metadata. I have added a commit here to see how you can support it:
https://github.com/phoenixframework/phoenix/commit/8c3c2c80dff25cd2cbb995db0340d3a4be22e6eb

To get the proper routed function for controllers and LiveViews, we're going to need to add a `:log_function` to the router metadata as well. You should be able to mirror everything we do + docs for the `:log_module`, then we'll be good to go. Btw I tested this with a project and the jump to line by clicking the output is just perfect. Great idea to show lines!

fchabouis:
Thanks for the feedback! I'll have a look at the `:log_function` as you suggest.

fchabouis:
Here is an update with the suggested changes @chrismccord.
See also https://github.com/phoenixframework/phoenix_live_view/pull/1970

I'm not very familiar with live_view. I tried in a new phoenix project to run 
```
mix phx.routes --info /dashboard
```
And got:
```
Module: Phoenix.LiveDashboard.PageLive
Function: :home
/home/francis/projects/phoenix/installer/route_info_test/deps/phoenix_live_dashboard/lib/phoenix/live_dashboard/page_live.ex
```
I understand, the information I passed in `:log_function` is in fact an "action", so there is no function called `:home` in the module. Is the function called always `:mount`?

Should we log the action name, but link in the module to the `:mount` function definition then?

josevalim:
I think action is better and we call it “log_action”. :)

fchabouis:
Hi @chrismccord, please let me know if there is anything else to add. Thanks!

chrismccord:
❤️❤️❤️🐥🔥

PR #4750, opened 2022-04-08
Allowing the use of nested resources with name: nil

dmitryv:
This PR allows for the resources _name_ option to be nil. I currently need to add a nested route with a custom ID name and don't see any other option of doing it.

Example:
```
    resources "/products", ProductController, param: "uuid" do
      resources "/comments", Api.CommentController
    end
```

This will generate routes like `products/:product_uuid/comments` and I need to have `:uuid` instead of `:product_uuid`.

So I've tried this:
```
    resources "/products", ProductController, param: "uuid", name: nil do
      resources "/comments", Api.CommentController
    end
```

and got the following warning

`warning: the underscored variable "_uuid" is used after being set. A leading underscore indicates that the value of the variable should be ignored. If this is intended please rename the variable to remove the underscore
  test/phoenix/router/resources_test.exs: Phoenix.Router.ResourcesTest.Router.Helpers.comment_path/3`

plus the route itself turned to be slightly wrong `products/:_uuid/comments` and also the name of the path didn't have the product word in it. This PR solves all the 3 problems.

### DIFF
--- a/lib/phoenix/router/resource.ex
+++ b/lib/phoenix/router/resource.ex
@@ -32,7 +32,7 @@ defmodule Phoenix.Router.Resource do
     alias   = Keyword.get(options, :alias)
     param   = Keyword.get(options, :param, @default_param_key)
     name    = Keyword.get(options, :name, Phoenix.Naming.resource_name(controller, "Controller"))
-    as      = Keyword.get(options, :as, name)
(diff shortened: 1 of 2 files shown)


chrismccord:
The `:name` also controls the named helper, so `nil` doesn't really  make sense in that capacity. Your best bet is to use the `:product_uuid` param or  build the routes yourself. Thanks!

PR #4783, opened 2022-04-14
Add execute permission to server in a container

N-0x90:
Generate the image from Azure leave the /app/bin/server without execute flag.
Its needed to place this explicit to make it work correctly.

### DIFF
--- a/priv/templates/phx.gen.release/Dockerfile.eex
+++ b/priv/templates/phx.gen.release/Dockerfile.eex
@@ -37,7 +37,7 @@ ENV MIX_ENV="prod"
 
 # install mix dependencies
 COPY mix.exs mix.lock ./
-RUN mix deps.get --only $MIX_ENV
+RUN mix deps.get --only ${MIX_ENV}
 RUN mkdir config
 
 # copy compile-time config files before we compile dependencies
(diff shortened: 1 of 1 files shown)


chrismccord:
It's not clear why a build on azure would result in the server not being executable. `chmod +x` is what we would want if this is required, but it would be nice to know more about why this isn't working out of the box. Thanks!

N-0x90:
I don't make a full in deep analyses about this, but this is what I see...
Making a total new project and prepare the deploy and pipeline... then deploy it over web app and get

message: OCI runtime create failed: container_linux.go:380: starting container process caused: exec: \"/app/bin/server\": permission denied: unknown"

Its a strange because if i make the same build from my pc all works flawless, but when the pipeline execute the copied files inside the image only have 

-rwxr-xr-x 1 nobody root 5308 Apr 14 06:04 project
-rwxr-xr-x 1 nobody root 6461 Apr 14 06:04 project.bat
-rw-r--r-- 1 nobody root   83 Apr 14 06:04 migrate
-rw-r--r-- 1 nobody root   49 Apr 14 06:04 migrate.bat
**-rw-r--r-- 1 nobody root   77 Apr 14 06:04 server**
-rw-r--r-- 1 nobody root   46 Apr 14 06:04 server.bat

then if add +x explicit all works flawless again...

dvic:
And have you checked the file itself on your computer? Does it have the execute flag?

N-0x90:
on my pc "windows" when i see the generated image it have 755 permisions...

chrismccord:
Closing this until we have more insight on if this is really required for azure. Thanks!

PR #4791, opened 2022-04-15
Store Phoenix version and Elixir requirement in text file.

eksperimental:
Simplify release process by storing Phoenix version and Elixir requirements in files

This PRs simplifies the release process by storing as single source of truth for:
- The Phoenix and Phoenix installer version in `VERSION`
- The Elixir requirement for Phoenix in `ELIXIR_REQUIREMENT`
- The Elixir requirement for Phoenix installer in `installer/ELIXIR_REQUIREMENT`

It removes the need to update the templates in the Phoenix installer.
Only .json and .md are required to be updated manually.


mix.exs
```diff
@@ -8,13 +8,15 @@ defmodule Phoenix.MixProject do
     end
   end
 
-  @version "1.6.4"
+  @version File.read!("VERSION")
```
cw789, line 11: Doesn't this also require to use `@external_resource`

eksperimental: it's an .exs file. It gets evaluated every time it is run. But I also wondered myself if there is a corner case where it should be required. Thank you for asking this.

cw789: You're right it's an `.exs`. But I think something then isn't working as expected (at least with Elixir v1.13.4).
If it means `mix test` should also always reevaluate `mix.exs` then it's definitely not working for me.

I think probably some special behavior for `mix.exs` coming with Elixir v1.13 has introduces something that this does not work as expexted. Probably related to https://github.com/elixir-lang/elixir/issues/11704.

eksperimental: It does not work with other elixir versions down to 1.10, either OTP 23 or 24,
Actually adding `@external_resource` does not solve the issue.

eksperimental: Apparently this feature has never worked in Elixir
with external_resource: https://github.com/eksperimental/debug_me/runs/6055378901?check_suite_focus=true
without it: https://github.com/eksperimental/debug_me/actions/runs/2180695163

here's the script that checks for it.
https://github.com/eksperimental/debug_me/blob/db6e04feeab1f29c4dbddd5a100c40661c7e93fe/check.sh

eksperimental: @cw789 The Mix tasks were storing the version number at compile time and that module didn't have a dependency on the VERSION file, so they file didn't get updated. Thanks @wojtekmach  for finding out the root cause.
Could you test it, I have, and it works as expected. Thank you.

cw789: Perfect. I've learned some new insights. Thank you.

chrismccord:
Thanks, but I think we're well covered by the current approach.

eksperimental:
> Thanks, but I think we're well covered by the current approach.

It is not.
There was a mismatching in the versions (that's actually why I came up with the PR).
If you look at these lines you will see what I mean:
Here the intaller says it is version 1.6.4 https://github.com/phoenixframework/phoenix/pull/4791/files#diff-268de1872950ff2850ee46314ffcf08e1e4c28b48a230000a3415fd423059a85L9

When the version tat the time of the PR was 1.6.8-dev https://github.com/phoenixframework/phoenix/pull/4791/files#diff-7b60b8e351cbb80c47459ffe2c79f1a26404871f49294780fe47ad0e58c09350

PR #4793, opened 2022-04-16
Add HTMLFormatter in formatter.exs templates

leandrocp:
That's an attempt to close #4749 

The config is added only if `@html` and Elixir >= 1.13.4 to meet formatter requirements.

PR #4794, opened 2022-04-16
CI - Bump Elixir to latest v1.13.4

leandrocp:
Just realized that CI won't test https://github.com/phoenixframework/phoenix/pull/4793 completely so It might be worth updating max Elixir version already.


leandrocp:
Closing it for now to debug locally

PR #4796, opened 2022-04-18
Fix flaky phx.gen.release ecto_sql check

alisinabh:
Sometimes in a project where everything is already compiled, The `mix phx.gen.release` task cannot correctly detect if ecto_sql is already installed or not. This is because the ecto_sql is in fact not loaded yet.

To resolve this we can either try to load ecto_sql using `Application.load(:ecto_sql)` before fetching the loaded apps or simply
use the `Mix.Project.deps_apps` to see if `ecto_sql` is there.

This PR currently implemented the latter.


chrismccord:
❤️❤️❤️🐥🔥

Issue #4451, opened 2021-08-30
Use Application.compile_env on all compile time config


josevalim:
Phoenix v1.7 can depend on Elixir v1.10 which means we can use Application.compile_env/3. Therefore, for all compile time configuration Phoenix uses on the endpoint, Phoenix should explicitly mark it as such. For example, for `debug_errors`, we should do:

    debug_errors? = Application.compile_env(otp_app, [__MODULE__, :debug_errors], false)

The benefit of doing this is that, if the user now configures debug errors at runtime.exs instead of config/config.exs, we will let them know the value changed at runtime while it is used at compile time.

Over the years, we removed most of the compile time configuration in Phoenix, but the ones that have stayed should be handled as above.

escobera:
Hey @josevalim 

Seems like these are the compile time configs for the Phoenix Endpoint

https://github.com/phoenixframework/phoenix/blob/510061e24bdd1b53a5ca1ca82cfbd27a3863eda1/lib/phoenix/endpoint/supervisor.ex#L201-L204

Those are the ones that are intended to be marked as `compile_env`?

josevalim:
Yes!

josevalim:
Closing in favor of PR!

PR #4801, opened 2022-04-20
Use `Application.compile_env` for compile-time config values on the endpoint

escobera:
This is an attempt at #4451 

Any feedback is welcome!


escobera:
Unfortunately I don't think this is the right solution, maybe I haven't grokked how `compile_env` works on this scenario. 

I had some expectations that were proven false:

1 - Changing one of the values at `runtime.exs` would trigger an error. Not the case.
2 - Calling `Application.put_env/2` inside the generated `DevAppWeb.Endpoint ` would prevent the application from booting up.  Didn't happen either.

I'm a little lost right now, do you have any hints how to go about doing this @josevalim ?

josevalim:
1. This should happen but you need to call compile_env inside the generated code (not inside Phoenix).
2. This is not verified because put_env is a regular runtime call we can't check

escobera:
Got it! Thanks for the clarification. I'll move the compile-time checks to the generated `Endpoint` and submit another version of this.

escobera:
Ok, now it seem to be working. But I think I hit some kind of race condition on the app initialization process.

Some times it just errors out without showing the complete error message like so:
```
$ _build/dev/rel/dev_app/bin/dev_app start
{"init terminating in do_boot",{<<"aborting boot">>,[{'Elixir.Config.Provider',boot,2,[]}]}}
init terminating in do_boot ({,[{Elixir.Config.Provider,boot,2,[]}]})

Crash dump is being written to: erl_crash.dump...done
```

Most of the times it works as it should:
```
$ _build/dev/rel/dev_app/bin/dev_app start
ERROR! the application :dev_app has a different value set for path [:debug_errors] inside key DevAppWeb.Endpoint during runtime compared to compile time. Since this application environment entry was marked as compile time, this difference can lead to different behaviour than expected:

  * Compile time value was set to: true
  * Runtime value was set to: "test"

To fix this error, you might:

  * Make the runtime value match the compile time one

  * Recompile your project. If the misconfigured application is a dependency, you may need to run "mix deps.compile dev_app --force"

  * Alternatively, you can disable this check. If you are using releases, you can set :validate_compile_env to false in your release configuration. If you are using Mix to start your system, you can pass the --no-validate-compile-env flag


{"init terminating in do_boot",{<<"aborting boot">>,[{'Elixir.Config.Provider',boot,2,[]}]}}
init terminating in do_boot ({,[{Elixir.Config.Provider,boot,2,[]}]})

Crash dump is being written to: erl_crash.dump...done
```

Steps to reproduce:

1 - Clone this branch
2 - Generate a new phoenix app 
```
$ cd phoenix/installer
$ mix phx.new dev_app --dev
```
3 - Add `config :dev_app, DevAppWeb.Endpoint, debug_errors: "test"` to `config/runtime.exs`
4 - `mix release`
5 - run `_build/dev/rel/dev_app/bin/dev_app start` a bunch of times.

josevalim:
Hi @escobera! This is the way to go but you want to read those values inside the `quote do` sections in `lib/phoenix/endpoint.ex`. That will have the same effect of running inside the user endpoint.

You will also need to bump Phoenix minimum Elixir require to v1.10 first.

escobera:
Sorry about the confusion, I moved the `compile_env` calls back into `lib/phoenix/endpoint.ex`  and it is working as intended. I also bumped the required Elixir version to 1.10.

About it not showing the output properly I was able to reproduce the problem without phoenix, so I'll open an issue at the Elixir repo.

Thanks for the guidance @josevalim , if you need me to change anything just let me know.

lib/phoenix/endpoint.ex
```diff
@@ -398,8 +398,16 @@ defmodule Phoenix.Endpoint do
...
+
       var!(config) = Phoenix.Endpoint.Supervisor.config(@otp_app, __MODULE__)
-      var!(code_reloading?) = var!(config)[:code_reloader]
+
+      # Compile-time configuration checkg
+      #
+      # This ensures that if a compile-time configuration is overwritten at runtime the application won't boot.
+      var!(code_reloading?) = Application.compile_env(@otp_app, [__MODULE__, :code_reloader], var!(config)[:code_reloader])
+      var!(debug_errors?) =  Application.compile_env(@otp_app, [__MODULE__, :debug_errors], var!(config)[:debug_errors])
+      var!(render_errors) =  Application.compile_env(@otp_app, [__MODULE__, :render_errors], var!(config)[:render_errors])
```
josevalim, line 409: I was wrong. This one should be force_ssl, and not render_errors. :)

josevalim, line 409:
```suggestion
      var!(code_reloading?) = Application.compile_env(@otp_app, [__MODULE__, :code_reloader])
      var!(debug_errors?) =  Application.compile_env(@otp_app, [__MODULE__, :debug_errors])
      var!(force_ssl) =  Application.compile_env(@otp_app, [__MODULE__, :force_ssl])
```

josevalim:
I have dropped a comment but it seems we may need to require Elixir v1.13.0.

lib/phoenix/endpoint.ex
escobera: I was going to suggest the removal of the checks done at the Endpoint supervisor but I see you already did that! =)

lib/phoenix/endpoint.ex
```diff
@@ -398,11 +398,19 @@ defmodule Phoenix.Endpoint do
...
       @otp_app unquote(opts)[:otp_app] || raise "endpoint expects :otp_app to be given"
+
       var!(config) = Phoenix.Endpoint.Supervisor.config(@otp_app, __MODULE__)
-      var!(code_reloading?) = var!(config)[:code_reloader]
+
+      # Compile-time configuration checkg
+      #
+      # This ensures that if a compile-time configuration is overwritten at runtime the application won't boot.
+      var!(code_reloading?) = Application.compile_env(@otp_app, [__MODULE__, :code_reloader], var!(config)[:code_reloader])
+      var!(debug_errors?) =  Application.compile_env(@otp_app, [__MODULE__, :debug_errors], var!(config)[:debug_errors])
```
josevalim, line 408: I think we may not need defaults.

```suggestion
      var!(code_reloading?) = Application.compile_env(@otp_app, [__MODULE__, :code_reloader])
      var!(debug_errors?) =  Application.compile_env(@otp_app, [__MODULE__, :debug_errors])
```

josevalim:
Unfortunately we cannot merge this yet because Elixir v1.11 still required the app to be a compile-time atom. This has been fixed on v1.13, so we will need to wait for that release.

lib/phoenix/config.ex
```diff
@@ -84,7 +84,6 @@ defmodule Phoenix.Config do
 
   @doc """
   Reads the configuration for module from the given OTP app.
```
josevalim, line 86:
```suggestion
  Reads the configuration for module from the given OTP app.

```

josevalim (approved):

escobera:
@josevalim I had to reintroduce `Phoenix.Endpoint.Supervisor.config/2` not to break tests.

SteffenDE:
> @josevalim I had to reintroduce `Phoenix.Endpoint.Supervisor.config/2` to not break tests.

My question is: if we only need it for a test, why not remove the test instead?

escobera:
> > @josevalim I had to reintroduce `Phoenix.Endpoint.Supervisor.config/2` to not break tests.
> 
> My question is: if we only need it for a test, why not remove the test instead?

Makes sense. I've removed it again and the test.

josevalim (approved):

PR #4802, opened 2022-04-20
docs(contexts): clearer diff in order_controller.ex

TommasoAmici:
When working through the guide, it wasn't clear we're changing Orders.create_order to Orders.complete_order in this step

### DIFF
--- a/guides/contexts.md
+++ b/guides/contexts.md
@@ -1204,7 +1204,8 @@ To complete an order, our cart page can issue a POST to the `OrderController.cre
 
 ```elixir
   def create(conn, _) do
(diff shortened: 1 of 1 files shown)


Gazler:
Thanks for the PR.

I believe the description following the code snippet describes this change:

> We rewrote the `create` action to call an as-yet-implemented `Orders.complete_order/1` function. The code that phoenix generated had a generic `Orders.create_order/1` call. Our code is technically "creating" an order, but it's important to step back and consider the naming of your interfaces.

Having diffs could be useful in the guides, but we don't have any at the moment, so it would be inconsistent.

PR #4803, opened 2022-04-21
Add typespecs to phoenix token functions

sorentwo:
Each `Phoenix.Token` function takes multiple arguments with ambiguous types. Furthermore, the functions delegate to `Plug.Crypto` which also lacks typespecs. This adds types and specs to make working with `Token` clearer, along with a minor adjustment to option documentation to keep the order consistent between functions.

### DIFF
--- a/lib/phoenix/token.ex
+++ b/lib/phoenix/token.ex
@@ -116,7 +116,7 @@ defmodule Phoenix.Token do
     * `:signed_at` - set the timestamp of the token in seconds.
       Defaults to `System.system_time(:second)`
     * `:max_age` - the default maximum age of the token. Defaults to
-      86400 seconds (1 day) and it may be overridden on `Plug.Crypto.verify/4`.
+      86400 seconds (1 day) and it may be overridden on `verify/4`.
 
   """
   @spec sign(context, binary, term, [shared_opt | signed_at_opt | max_age_opt]) :: binary
--- a/lib/phoenix/token.ex
+++ b/lib/phoenix/token.ex
@@ -115,11 +115,9 @@ defmodule Phoenix.Token do
       when generating the encryption and signing keys. Defaults to `:sha256`
     * `:signed_at` - set the timestamp of the token in seconds.
       Defaults to `System.system_time(:second)`
-    * `:max_age` - the default maximum age of the token. Defaults to
-      86400 seconds (1 day) and it may be overridden on `verify/4`.
 
   """
-  @spec sign(context, binary, term, [shared_opt | signed_at_opt | max_age_opt]) :: binary
+  @spec sign(context, binary, term, [shared_opt | signed_at_opt]) :: binary
   def sign(context, salt, data, opts \\ []) when is_binary(salt) do
     context
     |> get_key_base()
--- a/lib/phoenix/token.ex
+++ b/lib/phoenix/token.ex
@@ -92,8 +92,18 @@ defmodule Phoenix.Token do
 
   require Logger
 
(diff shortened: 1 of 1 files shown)


lib/phoenix/token.ex
```diff
@@ -92,6 +92,16 @@ defmodule Phoenix.Token do
...
+  @type shared_opt ::
+          {:key_iterations, pos_integer}
+          | {:key_length, pos_integer}
+          | {:key_digest, :sha256 | :sha384 | :sha512}
+
+  @type max_age_opt :: {:max_age, pos_integer}
+  @type signed_at_opt :: {:signed_at, pos_integer}
+
   @doc """
   Encodes  and signs data into a token you can send to clients.
```
leandrocp, line 106:
```suggestion
  Encodes and signs data into a token you can send to clients.
```

leandrocp (changes requested):
👀

sorentwo:
@leandrocp Anything else needed here?

lib/phoenix/token.ex
```diff
@@ -109,6 +119,7 @@ defmodule Phoenix.Token do
       86400 seconds (1 day) and it may be overridden on verify/4.
```
leandrocp, line 119:
```suggestion
      86400 seconds (1 day) and it may be overridden on `Plug.Crypto.verify/4`.
```

That fn belongs to another lib/context so IMO it's worth adding a proper link as it's done on other parts of this doc.

leandrocp (commented):
Just my 2 cents

leandrocp:
> @leandrocp Anything else needed here?

LGTM but it's up to the core team to review it 🙂

sorentwo:
Can any maintainers take a look? Our devs keep bumping into a lack of specs while referencing the token docs.

lib/phoenix/token.ex
```diff
@@ -106,9 +116,10 @@ defmodule Phoenix.Token do
     * `:signed_at` - set the timestamp of the token in seconds.
       Defaults to `System.system_time(:second)`
     * `:max_age` - the default maximum age of the token. Defaults to
-      86400 seconds (1 day) and it may be overridden on verify/4.
+      86400 seconds (1 day) and it may be overridden on `Plug.Crypto.verify/4`.
```
josevalim, line 119:
```suggestion
      86400 seconds (1 day) and it may be overridden on `verify/4`.
```

josevalim:

nshafer:
I just upgraded a project to Phoenix 1.7, and now I have dialyzer complaining that my call to Phoenix.Token.sign/4 breaks the contract. My code: 

```elixir
Phoenix.Token.sign(MyAppWeb, "test", true, max_age: max_age)
```

Results in:

```
The function call will not succeed.

Phoenix.Token.sign(
  _context :: any(),
  <<116,101,115,116>>,
  true,
  [{:max_age, _}, ...]
)

breaks the contract
(context(), binary(), term(), [shared_opt() | signed_at_opt()]) :: binary()
```

The code works fine, because it just calls Plug.Crypto.sign/4 which allows `max_age` in the opts, and indeed is useful functionality since you can use it to override the default 86400 that's stored in the token itself. I noticed that `max_age` was also removed from the docs for this and the encrypt/4 function.

I don't mind submitting a PR to fix the type spec and restore `max_age` to the docs. However, before I do that, I figured there was probably a good reason they were removed?

josevalim:
You should prefer to validate the max age on verification / decryption.

SteffenDE:
Related to https://github.com/phoenixframework/phoenix/pull/5444

PR #4805, opened 2022-04-25
phx.gen.auth: please credo

inoas:
`mix credo` returns some warnings for me on freshly baked code:

```
The alias `MyApp.Accounts.UserToken` is not alphabetically ordered among its group.
```

Should I also fix `modules should have a @moduledoc tag.`? for:

```
APPWeb.USERAuth
APP.ACCOUNTS.USERToken
APP.ACCOUNTS.USER
APP.ACCOUNTS.USERNotifier
```

(variables in uppercase)


leandrocp:
Most of the generated templates have a `@moduledoc` so it seems reasonable to keep the pattern of adding either `@moduledoc false` or some doc to explain about the generated module when that's the case.

chrismccord:
Alphabetic ordering is not a rule I would follow. You are of course free to follow whatever practices you feel is best after you run the generator. Thanks!

PR #4808, opened 2022-04-27
Implement Bandit HTTP server adapter

mtrudel:
Provides support for using Phoenix with the Plug-native HTTP server [Bandit](https://github.com/mtrudel/bandit). This adapter currently only supports HTTP(S) connections; WebSocket support is a work-in-progress which will be significantly facilitated by the work herein. 

Additionally, test coverage is unchanged here (though tests are, of course, green). I was unable to find any substantive tests which vary according to the underlying web server; happy to take guidance on this.

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -235,6 +235,37 @@ defmodule Phoenix.Endpoint do
       Defaults to `[]`, which will start a drainer process for each configured endpoint,
       but can be disabled by setting it to `false`.
 
+  Phoenix also provides the `Phoenix.Endpoint.BanditAdapter` adapter which works with the
+  [Bandit](https://hexdocs.pm/bandit) web server. It can be configured via the following top-level
+  options:
+
+    * `:http`: the configuration for the HTTP server. Accepts the following options:
+      * `port`: The port to run on. Defaults to 4000
+      * `ip`: The address to bind to. Can be specified as `{127, 0, 0, 1}`, or using `{:local,
+        path}` to bind to a Unix domain socket. Defaults to {127, 0, 0, 1}.
+      * `transport_options`: Any valid value from `ThousandIsland.Transports.TCP`
+    
+      Defaults to `false`, which will cause Bandit to not start an HTTP server.
+
+    * `:https`: the configuration for the HTTPS server. Accepts the following options:
+      * `port`: The port to run on. Defaults to 4040
+      * `ip`: The address to bind to. Can be specified as `{127, 0, 0, 1}`, or using `{:local,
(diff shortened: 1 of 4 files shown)


mtrudel:
Thanks for the quick response! I'll implement this provisional version of the Bandit adapter outside of Phoenix, as you suggest. Just so I'm better able to plan future WebSocket work in Bandit and beyond, I have a couple of finer points I'd greatly appreciate your perspective on.

Doing this work outside of Phoenix puts Bandit's WebSocket development in a bit of a chicken and egg scenario. A large part of my motivation for writing Bandit in the first place was to use it as a proving ground for developing a WebSocket version of Plug (that is, a server API abstraction of WebSockets in the same way Plug is an abstraction of HTTP). Any such API would of course have to consider Phoenix as the prototypical server, and as such would likely hew very closely to the existing semantics of Phoenix's internal socket mount / interface API. 

However, this is somewhat exacerbated by the fact that the overall structure of Phoenix's WebSocket support (although well isolated in `cowboy_handler`) is quite tightly coupled with Cowboy / Plug.Cowboy's take on the matter, and is more 'inside out' than 'outside-in' as a proper generic API would likely look like. Without having a clearer idea of what approach is likely to be favourably received by the Phoenix project, it's hard to know which approach to take to make a lot of progress here.


Perhaps I'm missing a perspective here that unlocks the complexity of this; if you have any insight it would be *greatly* appreciated. If there's a better method for me to drive this discussion & plan more of this up from with the Phoenix team, I'm happy to drive that. 

Lastly, assuming all of this comes to pass & Phoenix exposes / uses a public API for socket mounting, is it your expectation that the equivalent cowboy shim code continue to exist within Phoenix? Setting aside the choice of Cowboy as Phoenix's out-of-the-box default server, it would seem to follow from your statement that "WebSocket support is the only custom code that would require a Phoenix shim to exist" that the current Cowboy implementation within Phoenix would make more sense in an external library. I ask not just out of naked self-interest in having Bandit be seen as equal in stature to Cowboy, but also to understand what you envision as an ideal separation of concerns.

Thanks again!

chrismccord:
Yes, ideally we'd have a behavior/contract that would operate at the plug levels, and plug_cowboy and bandit would use that interface for websockets. Then Phoenix could ditch our cowboy specific code and our socket stuff would simply use the plug interfaces. I don't have specific API ideas at the moment, but where I would start is reviewing the cowboy_websocket behaviour and deciding if you like that direction. From there, I'd write my own bandit websocket behavior per my liking, and then see what it would take for a shared interface sitting on top. Assuming a shared interface is achievable, that's where the plug work would begin. Hope that helps!

PR #4809, opened 2022-04-28
Fix reference to UserSocket module in setup block

patreeceeo:
If I'm not mistaken, this code won't work otherwise.

### DIFF
--- a/guides/testing/testing_channels.md
+++ b/guides/testing/testing_channels.md
@@ -16,10 +16,24 @@ As we progress through this guide, it would help to have a concrete example we c
 $ mix phx.gen.channel Room
 * creating lib/hello_web/channels/room_channel.ex
 * creating test/hello_web/channels/room_channel_test.exs
+* creating test/support/channel_case.ex
 
-Add the channel to your `lib/hello_web/channels/user_socket.ex` handler, for example:
+The default socket handler - HelloWeb.UserSocket - was not found.
 
-    channel "room:lobby", HelloWeb.RoomChannel
+Do you want to create it? [Yn]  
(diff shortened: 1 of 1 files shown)


leandrocp:
Hi @patreeceeo your fix is correct and you may want to update the rest of that document too as it has changed a bit.

1. The output of `mix phx.gen.channel Room` now is:
```
mix phx.gen.channel Room
* creating lib/hello_web/channels/room_channel.ex
* creating test/hello_web/channels/room_channel_test.exs
* creating test/support/channel_case.ex

The default socket handler - HelloWeb.UserSocket - was not found.

Do you want to create it? [Yn] 
* creating lib/hello_web/channels/user_socket.ex
* creating assets/js/user_socket.js

Add the socket handler to your `lib/hello_web/endpoint.ex`, for example:

    socket "/socket", HelloWeb.UserSocket,
      websocket: true,
      longpoll: false

For the front-end integration, you need to import the `user_socket.js`
in your `assets/js/app.js` file:

    import "./user_socket.js"
```

2. The content of `HelloWeb.ChannelCase` now is:
```elixir
defmodule HelloWeb.ChannelCase do
  use ExUnit.CaseTemplate

  using do
    quote do
      # Import conveniences for testing with channels
      import Phoenix.ChannelTest
      import HelloWeb.ChannelCase

      # The default endpoint for testing
      @endpoint HelloWeb.Endpoint
    end
  end

  setup tags do
    Hello.DataCase.setup_sandbox(tags)
    :ok
  end
end
```

Please double check but I believe it's worth reviewing that doc, thanks!

PR #4810, opened 2022-04-28
Add mount_current_user on generated Auth module

bemesa21:
Added `mount_current_user/2` and `on_mount(:current_user, _params, session, socket)` callback to mount the current user in the socket assigns.

Assuming a `User` schema is defined, it generates the following code:

```elixir
  def on_mount(:mount_current_user, _params, session, socket) do
    {:cont, mount_current_user(session, socket)}
  end

  def on_mount(:ensure_authenticated, _params, session, socket) do
    socket = mount_current_user(session, socket)
    case  socket.assigns.current_user do
      nil ->
        {:halt, LiveView.redirect(socket, to: Routes.user_session_path(socket, :new))}

      _ ->
        {:cont, socket}
    end
  end

  defp mount_current_user(session, socket) do
    case session do
      %{"user_token" => user_token} ->
        LiveView.assign_new(socket, :current_user, fn -> Accounts.get_user_by_session_token(user_token) end)

      %{} ->
        LiveView.assign_new(socket, :current_user, fn -> nil end)
    end
  end
```


priv/templates/phx.gen.auth/auth.ex
```diff
@@ -107,6 +108,24 @@ defmodule <%= inspect auth_module %> do
     end
   end
 
+  @doc """
+  Authenticates and mounts in the socket assigns the
+  <%= schema.singular %> by looking into the session.
+  """
+  def mount_current_<%= schema.singular %>(session, socket) do
```
chrismccord, line 115: we can make this private. Let's also add an `on_mount(:ensure_authenticated, ...` clause which enforces authentication. It can call the shared private `mount_current_user` function

priv/templates/phx.gen.auth/auth.ex
```diff
@@ -107,6 +108,24 @@ defmodule <%= inspect auth_module %> do
...
+    case session do
+      %{"<%= schema.singular %>_token" => <%= schema.singular %>_token} ->
+        LiveView.assign_new(socket, :current_<%= schema.singular %>, fn -> <%= inspect context.alias %>.get_<%= schema.singular %>_by_session_token(<%= schema.singular %>_token) end)
+
+      %{} ->
+        LiveView.assign(socket, :current_<%= schema.singular %>, nil)
+    end
+  end
+
+  def on_mount(:current_<%= schema.singular %>, _params, session, socket) do
```
chrismccord, line 125: Add docs here and for the 2nd `on_mount` clause with examples on usage

priv/templates/phx.gen.auth/auth_test.exs
```diff
@@ -116,6 +116,52 @@ defmodule <%= inspect auth_module %>Test do
...
+
+  describe "on_mount: ensure_authenticated" do
+    test "authenticates current_<%= schema.singular %> based on a valid <%= schema.singular %>_token ", %{conn: conn, <%= schema.singular %>: <%= schema.singular %>} do
+      <%= schema.singular %>_token = <%= inspect context.alias %>.generate_<%= schema.singular %>_session_token(<%= schema.singular %>)
+      session = conn |> put_session(:<%= schema.singular %>_token, <%= schema.singular %>_token) |> get_session()
+      {:cont, updated_socket} = <%= inspect schema.alias %>Auth.on_mount(:ensure_authenticated, %{}, session, %Phoenix.LiveView.Socket{})
+      assert updated_socket.assigns.current_<%= schema.singular %>.id == <%= schema.singular %>.id
+    end
+
+    test "redirects to login page if there isn't a valid <%= schema.singular %>_token ", %{conn: conn} do
```
bemesa21, line 149: This test is generated:

```elixir
 test "redirects to login page if there isn't a valid user_token ", %{conn: conn} do
      user_token = "invalid_token"
      session = conn |> put_session(:user_token, user_token) |> get_session()
      socket = %Phoenix.LiveView.Socket{endpoint: TestAuthWeb.Endpoint}
      {:halt, updated_socket} = UserAuth.on_mount(:ensure_authenticated, %{}, session, socket)
      assert updated_socket.assigns.current_user == nil
    end
```
Is it a correct way to setup the socket?

chvanikoff:
Hey, thanks for adding this by default!
Quick question/suggestion: shouldn't this functionality be separated from the auth module in the /controller space?
All the Auth module functions are expecting `conn` now and it feels ok to keep it in the /controller. With the added functionality, it feels a bit off.
What would you think about moving the new functions into a separate liveview-related module in the /live space, or to move the whole `auth.ex` file out of /controller?

bemesa21:
> Hey, thanks for adding this by default! Quick question/suggestion: shouldn't this functionality be separated from the auth module in the /controller space? All the Auth module functions are expecting `conn` now and it feels ok to keep it in the /controller. With the added functionality, it feels a bit off. What would you think about moving the new functions into a separate liveview-related module in the /live space, or to move the whole `auth.ex` file out of /controller?

Hey! thanks for the suggestion!

I agree, they look a bit off. I think those new functions should be in the /live space, I'll create new module for them. 👌

priv/templates/phx.gen.auth/auth_test.exs
chrismccord: perfect!

bemesa21:
I left all the auth functions inside `auth.ex`, the auth file will be created inside the app's web directory instead of the controllers directory.

chrismccord:
❤️❤️❤️🐥🔥

PR #4818, opened 2022-05-06
Add phx-no-format to templates/layout/*.heex

karangejo:
I added the new heex formatter to my project and then ran ```mix format``` and then when I ran my project the alert boxes for the liveview flash messages where now displayed all the time but with no text this is because the formatter adds some whitespace and the css for the flash alerts has this:
```css
.alert:empty {
    display: none;
}
```
but the added whitespace from the formatter makes it so that it is no longer empty and so is always displayed. So I added phx-no-format to the alert tags so they won't get formatted incorrectly according to @leandrocp  as on this thread: https://github.com/phoenixframework/phoenix_live_view/issues/2001


feliperenan:
Hi @karangejo thanks for you PR.

But this issue will be solved whenthis PR: https://github.com/phoenixframework/phoenix/pull/4667 gets merged 🙌🏻

karangejo:
Ok cool! No problem!

PR #4819, opened 2022-05-07
Add # back for .dockerignore comment

wkirschbaum:
Perhaps I am mistaken, but it seems like the `$ cat ...` line is not intended.


Gazler:
Was definitely intended to be a comment. Thanks.

PR #4830, opened 2022-05-19
Fix JSON response @spec

tomekowal:
Hi!

I have an API that returns lists of things in the API without any wrapping. After parsing becomes a list:
`[object1, object2, object]`
and it breaks dialyzer because `json_response` specifies that parsed response might be only a map.

I've searched the project for lines that begin with `@spec`, have `json` in the name and `map` in the return value and it looks like it is the only place where `phoenix` assumes the shape of json.

### DIFF
--- a/lib/phoenix/test/conn_test.ex
+++ b/lib/phoenix/test/conn_test.ex
@@ -400,6 +400,8 @@ defmodule Phoenix.ConnTest do
     body
   end
 
+  @type json_value :: integer | float | boolean | nil
+  @type json_response :: map | list | json_value
   @doc """
   Asserts the given status code, that we have a json response and
   returns the decoded JSON response if one was set or sent.
@@ -410,7 +412,7 @@ defmodule Phoenix.ConnTest do
       assert "can't be blank" in body["errors"]
 
   """
-  @spec json_response(Conn.t, status :: integer | atom) :: map | list | no_return
+  @spec json_response(Conn.t, status :: integer | atom) :: json_response | no_return
   def json_response(conn, status) do
     body = response(conn, status)
     _    = response_content_type(conn, :json)
--- a/lib/phoenix/test/conn_test.ex
+++ b/lib/phoenix/test/conn_test.ex
@@ -410,7 +410,7 @@ defmodule Phoenix.ConnTest do
       assert "can't be blank" in body["errors"]
 
   """
-  @spec json_response(Conn.t, status :: integer | atom) :: map | no_return
+  @spec json_response(Conn.t, status :: integer | atom) :: map | list | no_return
   def json_response(conn, status) do
     body = response(conn, status)
     _    = response_content_type(conn, :json)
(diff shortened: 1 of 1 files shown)


lib/phoenix/test/conn_test.ex
```diff
@@ -410,7 +410,7 @@ defmodule Phoenix.ConnTest do
       assert "can't be blank" in body["errors"]
 
   """
-  @spec json_response(Conn.t, status :: integer | atom) :: map | no_return
+  @spec json_response(Conn.t, status :: integer | atom) :: map | list | no_return
```
fuelen, line 413: to be fair, other types can be there as well, like string, nil, integer, float, boolean

tomekowal: You are totally right. So I went through JSON spec https://www.json.org/json-en.html and thankfully apart from arrays (lists) and objects (maps) there are only there are only a couple of primitives:
```
string - String.t
number - integer | float
true | false - boolean
null - nil
```
That should be exhaustive.

I was wandering about stealing it from Jason, but actually, Jason uses `term`...
Maybe we shouldn't be so granular and also return `term`?

fuelen: `Poison` uses more precise definition:

https://github.com/devinus/poison/blob/de07407fd35293d136ef3c7b2a140eaa29ff5617/lib/poison/parser.ex#L69-L75

Let's wait what maintainers say.
But I'd use this definition:
```
@type json :: boolean() | nil | float() | integer() | String.t() | [json()] | %{String.t() => json()}
```

tomekowal: I've thought about it some more and I've changed it to `term`.
For e.g. `Jason` allows decoding floats either to `native` or to `Decimal` so that type would not be covered by the above definition.
We would need to add every decoding possibility from every library to cover all edge cases and make dialyzer happy.
I think it is better to go back to a more generic `term`.

Thanks, @fuelen for the fruitful discussion!

fuelen: Converting to `Decimal` is possible only if `decode/2` is used, which is not the case

chrismccord:
❤️❤️❤️🐥🔥

lib/phoenix/test/conn_test.ex
```diff
@@ -410,7 +410,7 @@ defmodule Phoenix.ConnTest do
       assert "can't be blank" in body["errors"]
 
   """
-  @spec json_response(Conn.t, status :: integer | atom) :: map | no_return
+  @spec json_response(Conn.t, status :: integer | atom) :: term | no_return
```
wojtekmach, line 413: while at it, consider dropping ` | no_return`. :) `no_return` is for functions that _never_ return: they always raise/throw/exit, stop the VM, loop infinitely, etc. If we were to add ` | no_return` to functions that may raise, we'd have to do that for majority of them as most would on invalid args, etc.

PR #4832, opened 2022-05-20
Deprecate port_to_integer method in Cowboy2Adapter

J0:
Might be confused but seems like we currently require elixir 1.9 as per the [mix.exs](https://github.com/phoenixframework/phoenix/blob/master/mix.exs) and should be able to deprecate the method


chrismccord:
We still support apps with the older config, so this stays for now. Thanks!

PR #4834, opened 2022-05-21
Move swoosh api_client config out of config.exs to ENV.exs

Gazler:
Since the `:swoosh, :api_client` is set to false in `config.exs`, this
requires setting both the adapter and api_client in either `prod.exs` or
`runtime.exs` to ensure mails are sent in a production environment.
However, if this is not configured, the following error is raised:

```
** (UndefinedFunctionError) function false.post/4 is undefined
(module false is not available)
```

It is not entirely clear from the error how to track the value back to
this configuration setting.

The `:api_client` is not set in `dev.exs` and `test.exs` separately, as
these are the environments where this value is less likely to be
configured. Now in production, it will default to the swoosh default
(currently hackney). If the user does not have hackney installed in
mix.exs then the error message is more friendly, prompting them to add
`{:hackney, "~>..."}` to their `mix.exs` file.

### DIFF
--- a/installer/templates/phx_single/config/config.exs
+++ b/installer/templates/phx_single/config/config.exs
@@ -26,10 +26,7 @@ config :<%= @app_name %>, <%= @endpoint_module %>,
 #
 # For production it's recommended to configure a different adapter
 # at the `config/runtime.exs`.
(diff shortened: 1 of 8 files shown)


chrismccord:
❤️❤️❤️🐥🔥

PR #4850, opened 2022-06-03
Deprecate {:system, env_var} configuration

MichalDolata:
As https://github.com/phoenixframework/phoenix/commit/f0049f009237af56693efc0b52c1b46431f45568 bumped up required Elixir version to 1.9+, we can deprecate `{:system, env_var}` tuple in configuration


josevalim:
The deprecation you have removed is only on master, we have not shipped it yet, so we should wait probably 2 or 3 years to give people plenty of time to catch-up.

PR #4851, opened 2022-06-05
Fix minor typo in Context guide

wceolin:
There's a minor typo in the Context guide where "to handle" is in the past (to handled).


Gazler:
Thanks <3

PR #4866, opened 2022-06-13
Reword longpoll options

kianmeng:
This clear the confusion of setting the value to `10_000ms` where we
should be using `10_000`.


josevalim:

kianmeng:

PR #4867, opened 2022-06-19
updates phx.gen.auth docs and mix task help info

Morzaram:
Hey there, 

This is my first time doing this thing and I saw some information was missing in regards to doing multiple accounts within phx.gen.auth. I believe this is the best way to do this hence my update of the documentation. However, I'm not the most advanced writer, nor developer. So if you have any issues then that's okay by me. I do think that some of the mix tasks should cover this though since contexts and scaffolders are powerful but not too much is there when it comes to scaffolding into folders as I cover in the paragraphs I've written. 

Thank you and I hope you find this useful!

### DIFF
--- a/guides/authentication/mix_phx_gen_auth.md
+++ b/guides/authentication/mix_phx_gen_auth.md
@@ -43,18 +43,18 @@ $ mix phx.server
 If you have a User and Admin setup, ie: where an app needs moderation and such then it's suggested that you use the commands like so:
 
 ```console
-mix phx.gen.auth Accounts.Admin Admin admins
-mix phx.gen.auth Accounts.User User users
+mix phx.gen.auth Accounts.Admins Admin admins
+mix phx.gen.auth Accounts.Users User users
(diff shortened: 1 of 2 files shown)


guides/authentication/mix_phx_gen_auth.md
```diff
@@ -4,6 +4,8 @@ The `mix phx.gen.auth` command generates a flexible, pre-built authentication sy
 
 ## Getting started
 
+As a recommendation, commit all of your work before running this command because this modifies/creates 39 files that if you want to undo, will be quite a time consumer if you have other files uncommited.
```
josevalim, line 7:
```suggestion
Considering committing your work before running this command as it generates many files.
```

josevalim:
Thank you @Morzaram, i pushed some related notes. I think we should not necessarily recommend separate resources for admin/users (sometimes authorization is better than authentication), so I just added a quick mention to it.

PR #4868, opened 2022-06-19
Add missing gettext error messages to pot file

### LINKED ISSUE #4849: Some translation messages are missing in the generated `priv/gettext/errors.pot`
### Environment

* Phoenix version (mix deps):1.6.9

### Expected behavior

All error translations in the generated po-files correspond with a message in the template pot-files. Running `mix gettext.extract --merge` does not introduce changes to the po-files, in a newly generated phoenix project.

### Actual behavior

Not all translations in `priv/gettext/en/LC_MESSAGES/errors.po` appear in the template `priv/gettext/errors.pot`. These are missing in `errors.pot`:

- `should be %{count} byte(s)`
- `should be at least %{count} byte(s)`
- `should be at most %{count} byte(s)`

Running `mix gettext.extract --merge` results in these messages being removed from the po-file, which in fact is not required, as they are referenced in some validations.

Also, the sorting is a bit different in the generated po and pot-files, which is a bit confusing because a change is created when first using `mix gettext.extract --merge` on a clean project.

linusdm:
Not all translations in errors.po appear in the template errors.pot.
These are added in errors.pot:

- should be %{count} byte(s)
- should be at least %{count} byte(s)
- should be at most %{count} byte(s)

This avoids these messages being removed when running
`mix gettext.extract --merge` on a clean project.

The order is also consistent, again to avoid changes
when running gettext on a clean project.

Closes #4849


chrismccord:
❤️❤️❤️🐥🔥

PR #4882, opened 2022-07-10
Add `--repo` option to schema generator

d3mcfadden:
I added a `--repo` option to the schema generator so apps that have multiple repos can have the ability to select what repo the schema/migration should use.

I noticed `ecto.gen.migration` supports this flag and assumed it would work in phoenix as well. It did not, so here it is :-)

This works on the schema generator and other "parent" generators that call it (eg: `phx.gen.context`)

### NOTE to reviewer

Ideally, I wanted to support a custom `priv` config that may be set on the repo to determine the `migration_path`, however I noticed that will require me ensuring the repo module is compiled first. I wanted to get some advice on that before doing anything.

### DIFF
--- a/lib/mix/tasks/phx.gen.schema.ex
+++ b/lib/mix/tasks/phx.gen.schema.ex
@@ -86,6 +86,14 @@ defmodule Mix.Tasks.Phx.Gen.Schema do
   Generated migration can use `binary_id` for schema's primary key
   and its references with option `--binary-id`.
 
+  ## repo
+
+  By default, the migration will be generated for the primary repo. If your app
+  has multiple repos you can choose the repo the migration is generated for with
+  the `--repo` option.
+
+      $ mix phx.gen.schema Blog.Post posts --repo MyApp.SecondaryRepo
+
   ## Default options
 
   This generator uses default options provided in the `:generators`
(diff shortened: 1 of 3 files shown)


josevalim:
I would rather support setting a custom migration directory and that's it. Changing the directory itself should be done manually IMO because you can end-up in scenarios where a context has multiple repos aliases `as: Repo` and that can be confusing and trying to address such corner use case through generators are not going to scale IMO. :)

d3mcfadden:
@josevalim thanks for the feedback! That makes sense to me also. I'm working on an app that has 2 repos and manually have to move stuff around after using the generators. It is not _that_ big of a deal, but I figured it may be useful to us and others.

I can update this to support `--migration-path` instead if you think that change would make it in. Otherwise, we can just close this out. 

Thanks!

josevalim:
The migration path bit is welcome if you think it is helpful. Your call!

josevalim:
Closing this./ If you want to send a PR for the custom migration path, it would be welcome!

MikaAK:
I realised this was around after i created #4905 which I needed for a generator project I'm working on, I've implemented both though if you want me to strip one of the options lemme know

PR #4888, opened 2022-07-14
Use Telemetry as first child to not miss init telemetry events

robsonpeixoto:
The `Ecto.Repo` trigger some `:init` events that can be lost if the Telemetry start after the Repo.


josevalim:

PR #4890, opened 2022-07-18
docs: fix diff on ecto contexts walk through

pedromss:
Following the guide, eventually I got to this point:

```
$ mix phx.server
Compiling 3 files (.ex)
warning: Category.__schema__/1 is undefined (module Category is not available or is yet to be defined)
  lib/hello/catalog.ex:114: Hello.Catalog.list_categories_by_id/1
```

The flow of creating a product with categories then also fails.

The alias fixed it.


josevalim:

PR #4895, opened 2022-07-22
Add verified routes

chrismccord:
Still needs to support interpolated query string form, as well as moduledocs

### DIFF
--- a/lib/phoenix/verified_routes.ex
+++ b/lib/phoenix/verified_routes.ex
@@ -527,8 +527,8 @@ defmodule Phoenix.VerifiedRoutes do
   end
 
   # we reached the dynamic query string, return to call with rewritten query
-  defp verify_segment(["?" | rest], route, acc) do
-    {Enum.reverse(acc), verify_query(rest, route, ["?"])}
+  defp verify_segment(["?" <> _ = static_query_segment | rest], route, acc) do
+    {Enum.reverse(acc), verify_query(rest, route, [static_query_segment])}
   end
 
   defp verify_segment([segment | _], route, _acc) when is_binary(segment) do
--- a/lib/phoenix/verified_routes.ex
+++ b/lib/phoenix/verified_routes.ex
@@ -542,20 +542,15 @@ defmodule Phoenix.VerifiedRoutes do
            | rest
          ],
          route,
-         acc
-       ) do
-    unless is_binary(Enum.at(acc, 0)) do
-      raise ArgumentError,
-            "a dynamic ~p interpolation must be followed by a static segment, got: #{Macro.to_string(route)}"
-    end
-
+         [prev | _] = acc
+       ) when is_binary(prev) do
     rewrite = {:"::", m1, [{{:., m2, [__MODULE__, :__encode_segment__]}, m3, [dynamic]}, bin]}
     verify_segment(rest, route, [rewrite | acc])
   end
 
-  defp verify_segment([_other | _], route, _acc) do
+  defp verify_segment([_ | _], route, _acc) do
     raise ArgumentError,
-          "verified routes require a compile-time string, got: #{Macro.to_string(route)}"
+          "a dynamic ~p interpolation must follow a static segment, got: #{Macro.to_string(route)}"
   end
 
   # we've reached the end of the path without finding query, return to caller
--- a/lib/phoenix/verified_routes.ex
+++ b/lib/phoenix/verified_routes.ex
@@ -584,6 +584,17 @@ defmodule Phoenix.VerifiedRoutes do
     {type, test_path, path_ast, static_ast}
   end
 
+  defp attr!(env, :endpoint) do
+    Module.get_attribute(env.module, :endpoint) ||
+      raise """
+      expected @endpoint to be set. For dynamic endpoint resolution, use path/2 instead.
+
+      for example:
+
+          path(conn_or_socket, ~p"/my-path")
+      """
+  end
+
   defp attr!(env, name) do
     Module.get_attribute(env.module, name) || raise "expected @#{name} module attribute to be set"
   end
--- a/lib/phoenix/verified_routes.ex
+++ b/lib/phoenix/verified_routes.ex
@@ -12,8 +12,7 @@ defmodule Phoenix.VerifiedRoutes do
     endpoint: AppWeb.Endpoint,
     statics: ~(images)
   """
-  defstruct endpoint_ctx: nil,
-            router: nil,
+  defstruct router: nil,
             route: nil,
             inspected_route: nil,
             stacktrace: nil,
@@ -59,7 +58,7 @@ defmodule Phoenix.VerifiedRoutes do
 
   @doc false
   def __verify__(routes) when is_list(routes) do
-    Enum.each(routes, fn {_kind, %__MODULE__{} = route} ->
+    Enum.each(routes, fn %__MODULE__{} = route ->
       unless match_route?(route.router, route.test_path) do
         IO.warn(
           "no route path for #{inspect(route.router)} matches #{route.inspected_route}",
@@ -103,29 +102,35 @@ defmodule Phoenix.VerifiedRoutes do
     |> inject_path(__CALLER__)
   end
 
-  defp inject_path({%__MODULE__{} = route, type, _route_ast, path_ast, static_ast}, env) do
+  defp inject_path(
+         {%__MODULE__{} = route, type, _endpoint_ctx, _route_ast, path_ast, static_ast},
+         env
+       ) do
     case type do
       :static ->
         static_ast
 
       other when other in [:match, :error] ->
-        Module.put_attribute(env.module, :phoenix_verified_routes, {:path, route})
+        Module.put_attribute(env.module, :phoenix_verified_routes, route)
         path_ast
     end
   end
 
-  defp inject_url({%__MODULE__{} = route, type, route_ast, path_ast, _static_ast}, env) do
+  defp inject_url(
+         {%__MODULE__{} = route, type, endpoint_ctx, route_ast, path_ast, _static_ast},
+         env
+       ) do
     case type do
       :static ->
         quote do
-          unquote(__MODULE__).static_url(unquote_splicing([route.endpoint_ctx, route_ast]))
+          unquote(__MODULE__).static_url(unquote_splicing([endpoint_ctx, route_ast]))
         end
 
       other when other in [:match, :error] ->
-        Module.put_attribute(env.module, :phoenix_verified_routes, {:url, route})
+        Module.put_attribute(env.module, :phoenix_verified_routes, route)
 
         quote do
-          unquote(__MODULE__).unverified_url(unquote_splicing([route.endpoint_ctx, path_ast]))
+          unquote(__MODULE__).unverified_url(unquote_splicing([endpoint_ctx, path_ast]))
         end
     end
   end
@@ -507,14 +512,13 @@ defmodule Phoenix.VerifiedRoutes do
       rewrite_path(route_ast, endpoint_ctx, router, statics)
 
     route = %__MODULE__{
-      endpoint_ctx: endpoint_ctx,
       router: router,
       stacktrace: Macro.Env.stacktrace(env),
       inspected_route: Macro.to_string(og_ast),
       test_path: test_path
     }
 
-    {route, type, route_ast, path_ast, static_ast}
+    {route, type, endpoint_ctx, route_ast, path_ast, static_ast}
   end
 
   defp rewrite_path(route, endpoint, router, statics) do
--- a/lib/phoenix/verified_routes.ex
+++ b/lib/phoenix/verified_routes.ex
@@ -571,7 +571,7 @@ defmodule Phoenix.VerifiedRoutes do
   defp verify_query(["&" <> _ = param | rest], route, acc) do
     unless String.contains?(param, "=") do
       raise ArgumentError,
-            "expected query string param key to end with =, got: #{inspect(param)}"
+            "expected query string param key to end with = or declare a static key value pair, got: #{inspect(param)}"
     end
 
     verify_query(rest, route, [param | acc])
--- a/lib/phoenix/verified_routes.ex
+++ b/lib/phoenix/verified_routes.ex
@@ -12,17 +12,12 @@ defmodule Phoenix.VerifiedRoutes do
     endpoint: AppWeb.Endpoint,
     statics: ~(images)
   """
-  defstruct endpoint: nil,
+  defstruct endpoint_ctx: nil,
             router: nil,
-            statics: nil,
             route: nil,
             inspected_route: nil,
             stacktrace: nil,
-            test_path: nil,
-            route_ast: nil,
-            path_ast: nil,
-            static_ast: nil,
-            type: nil
+            test_path: nil
 
   defmacro __using__(opts) do
     opts =
@@ -33,7 +28,11 @@ defmodule Phoenix.VerifiedRoutes do
       end
 
     quote do
-      Module.register_attribute(__MODULE__, :phoenix_verified_routes, accumulate: true, persist: false)
+      Module.register_attribute(__MODULE__, :phoenix_verified_routes,
+        accumulate: true,
+        persist: false
+      )
+
       opts = unquote(opts)
 
       @before_compile unquote(__MODULE__)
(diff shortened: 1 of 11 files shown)


lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,607 @@
...
+
+      opts = unquote(opts)
+
+      @before_compile unquote(__MODULE__)
+      @endpoint Keyword.get(opts, :endpoint)
+      @router Keyword.get(opts, :router)
+      @statics Keyword.get(opts, :statics, [])
+      import unquote(__MODULE__)
+    end
+  end
```
josevalim, line 36:
```suggestion
    quote do
      unquote(__MODULE__).__using__(__MODULE__, unquote(opts))
      import unquote(__MODULE__)
    end
  end
  
  def __using__(mod, opts) do
    Module.register_attribute(mod, :phoenix_verified_routes, accumulate: true)
    Module.put_attribute(mod, :before_compile, __MODULE__)
    Module.put_attribute(mod, :router, Keyword.fetch!(opts, :router))
    Module.put_attribute(mod, :endpoint, Keyword.get(opts, :endpoint))
    Module.put_attribute(mod, :phoenix_verified_statics, Keyword.get(opts, :statics)
  end
```

Note I did two extra changes:

1. Made the :router required
2. Renamed :statics to :phoenix_verified_statics

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,607 @@
...
+  @moduledoc ~S"""
+  TODO
+    - [ ] ~p"/posts?page=#{page}"
+
+  use Phoenix.VerifiedRoutes,
+    router: AppWeb.Router,
+    endpoint: AppWeb.Endpoint,
+    statics: ~(images)
+  """
+  defstruct router: nil,
```
josevalim, line 11:
```suggestion
  @doc false
  defstruct router: nil,
```

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,607 @@
...
+      redirect(to: path(conn, ~p"/users/top"))
+
+      redirect(to: path(conn, ~p"/users/#{@user}"))
+
+      ~H"""
+      <.link to={path(@uri, "/users?#{[page: @page]}")}>profile</.link>
+      """
+  '''
+  defmacro path(
+             endpoint,
```
josevalim, line 166: I think this should also be: `conn_or_socket_or_endpoint_or_uri`. And the examples use `path/2`, not `path/3`.

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,607 @@
...
+  ## Examples
+
+      use Phoenix.VerifiedRoutes, endpoint: MyAppWeb.Endpoint, router: MyAppWeb.Router
+
+      redirect(to: path(conn, ~p"/users/top"))
+
+      redirect(to: path(conn, ~p"/users/#{@user}"))
+
+      ~H"""
+      <.link to={path(@uri, "/users?#{[page: @page]}")}>profile</.link>
```
josevalim, line 232: wrong example?

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,607 @@
...
+      script_name = router.__forward__(forward_router)
+      match_route?(forward_router, split_path -- script_name)
+    else
+      true
+    end
+  end
+
+  defp build_route(route_ast, og_ast, env, endpoint_ctx, router) do
+    statics = Module.get_attribute(env.module, :statics, [])
+    {endpoint_ctx, router, statics} = Macro.expand({endpoint_ctx, router, statics}, env)
```
josevalim, line 528: This won't work because you are building a weird tuple that looks like an AST. Plus `Macro.expand` is not recursive. I also think that you already manually expand router and endpoint when they are given as arguments.

```suggestion
```

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,607 @@
...
+
+    {type, test_path, path_ast, static_ast} =
+      rewrite_path(route_ast, endpoint_ctx, router, statics)
+
+    route = %__MODULE__{
+      router: router,
+      stacktrace: Macro.Env.stacktrace(env),
+      inspected_route: Macro.to_string(og_ast),
+      test_path: test_path
+    }
```
josevalim, line 538: Nice!

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,607 @@
...
+    static_ast =
+      quote do
+        unquote(__MODULE__).static_path(unquote_splicing([endpoint, rewrite_route]))
+      end
+
+    {type, test_path, path_ast, static_ast}
+  end
+
+  defp attr!(env, name) do
+    Module.get_attribute(env.module, name) || raise "expected @#{name} module attribute to be set"
```
josevalim, line 578: I think we need a better error message here.

1. Router should always be given
2. The issue is that endpoint is optional but, if the endpoint is missing, we should tell people to use `path(conn_or_socket, ...)`. So a good error message should be tailored to endpoint and say that they should either pass it on use or use `path/2`

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,607 @@
...
+    router = Macro.expand(router, __CALLER__)
+
+    route
+    |> build_route(og_ast, __CALLER__, endpoint, router)
+    |> inject_path(__CALLER__)
+  end
+
+  defmacro path(endpoint, router, {:sigil_p, _, [str, extra]} = og_ast) when is_binary(str) do
+    validate_sigil_p!(extra)
+    endpoint = Macro.expand(endpoint, __CALLER__)
```
josevalim, line 181: I don't think you want to expand the endpoint... we always treat it as AST. But you want to expand the router and make sure it expands to an atom, otherwise you need to ask them to use unverified_path.

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,607 @@
...
+      """
+  '''
+  defmacro path(
+             endpoint,
+             router,
+             {:sigil_p, _, [{:<<>>, _meta, _segments} = route, extra]} = og_ast
+           ) do
+    validate_sigil_p!(extra)
+    endpoint = Macro.expand(endpoint, __CALLER__)
+    router = Macro.expand(router, __CALLER__)
```
josevalim, line 172: I don't think you want to expand the endpoint... we always treat it as AST. But you want to expand the router and make sure it expands to an atom, otherwise you need to ask them to use unverified_path.

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,607 @@
...
+
+    route
+    |> build_route(og_ast, __CALLER__, endpoint, router)
+    |> inject_path(__CALLER__)
+  end
+
+  defmacro path(endpoint, router, {:sigil_p, _, [str, extra]} = og_ast) when is_binary(str) do
```
josevalim, line 179: I don't think this branch will ever be invoked. `sigil_p` is always called with `:<<>>` AST exactly to make it easier to parse:

```
iex(1)> quote do: ~p"foo"
{:sigil_p, [delimiter: "\""], [{:<<>>, [], ["foo"]}, []]}
```

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,607 @@
...
+           ) do
+    validate_sigil_p!(extra)
+    router = attr!(__CALLER__, :router)
+
+    route
+    |> build_route(og_ast, __CALLER__, conn_or_socket_or_endpoint_or_uri, router)
+    |> inject_path(__CALLER__)
+  end
+
+  defmacro path(conn_or_socket_or_endpoint_or_uri, {:sigil_p, _, [route, extra]} = og_ast)
```
josevalim, line 203: You can kill this branch too.

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,607 @@
...
+  defmacro url(conn_or_socket_or_endpoint_or_uri, {:sigil_p, _, [route, _]} = og_ast)
+           when is_binary(route) do
+    router = attr!(__CALLER__, :router)
+
+    route
+    |> build_route(og_ast, __CALLER__, conn_or_socket_or_endpoint_or_uri, router)
+    |> inject_url(__CALLER__)
+  end
+
+  defmacro url(_conn_or_socket_or_endpoint_or_uri, other), do: raise_invalid_route(other)
```
josevalim, line 275: Are we not having `url/3`? I don't remember.

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,719 @@
...
+  ## Examples
+
+      import Phoenix.VerifiedRoutes
+
+      redirect(to: path(conn, MyAppWeb.Router, ~p"/users/top"))
+
+      redirect(to: path(conn, MyAppWeb.Router, ~p"/users/#{@user}"))
+
+      ~H"""
+      <.link to={path(@uri, MyAppWeb.Router, "/users?#{[page: @page]}")}>profile</.link>
```
josevalim, line 229:
```suggestion
      <.link to={path(@uri, MyAppWeb.Router, "/users?page=#{@page}")}>profile</.link>
```

There may be other uses. :)

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,719 @@
...
+      test_path: test_path
+    }
+
+    {route, type, endpoint_ctx, route_ast, path_ast, static_ast}
+  end
+
+  defp rewrite_path(route, endpoint, router, statics) do
+    {rewrite_route, test_path} =
+      case route do
+        route when is_binary(route) ->
```
josevalim, line 648: I don't think this branch will ever match.

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,720 @@
...
+    router = attr!(__CALLER__, :router)
+
+    route
+    |> build_route(og_ast, __CALLER__, endpoint, router)
+    |> inject_url(__CALLER__)
+  end
+
+  defmacro url(other), do: raise_invalid_route(other)
+
+  defmacro url(
```
josevalim, line 313: Missing docs.

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,720 @@
...
+    router = attr!(__CALLER__, :router)
+
+    route
+    |> build_route(og_ast, __CALLER__, conn_or_socket_or_endpoint_or_uri, router)
+    |> inject_url(__CALLER__)
+  end
+
+  defmacro url(_conn_or_socket_or_endpoint_or_uri, other), do: raise_invalid_route(other)
+
+  defmacro url(
```
josevalim, line 326: Missing docs.

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,757 @@
...
+    router =
+      case Macro.expand(router, env) do
+        mod when is_atom(mod) ->
+          mod
+
+        other ->
+          raise ArgumentError, """
+          expected router to be to module, got: #{inspect(other)}
+
+          If you want to generate a compile-time router, use unverified_path/2 instead.
```
josevalim, line 672:
```suggestion
          If your router is not defined at compile-time, use unverified_path/3 instead.
```

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,757 @@
...
+          """
+      end
+
+    statics =
+      case Macro.expand(statics, env) do
+        list when is_list(list) -> list
+        other -> raise ArgumentError, "expected statics to be a list, got: #{inspect(other)}"
```
josevalim, line 679: Statics should already be expanded at this point, pls move the validation to `__using__/2`.

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,757 @@
...
+    unless is_binary(Enum.at(acc, 0)) do
+      raise ArgumentError,
+            "a dynamic ~p interpolation must be followed by a static segment, got: #{Macro.to_string(route)}"
+    end
+
+    rewrite = {:"::", m1, [{{:., m2, [__MODULE__, :__encode_segment__]}, m3, [dynamic]}, bin]}
+    verify_segment(rest, route, [rewrite | acc])
+  end
+
+  defp verify_segment([_other | _], route, _acc) do
```
josevalim, line 549: I don't think this clause can ever be triggered.

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,757 @@
...
+    raise ArgumentError,
+          "verified routes require a compile-time string, got: #{Macro.to_string(route)}"
+  end
```
josevalim, line 552: I don't think this last clause is ever used. In any case, we can move things to a guard and reuse it:

```suggestion
  defp verify_segment(
         [
           {:"::", m1, [{{:., m2, [Kernel, :to_string]}, m3, [dynamic]}, {:binary, _, _} = bin]}
           | rest
         ],
         route,
         [prev | _] = acc
       ) when is_binary(prev) do
    rewrite = {:"::", m1, [{{:., m2, [__MODULE__, :__encode_segment__]}, m3, [dynamic]}, bin]}
    verify_segment(rest, route, [rewrite | acc])
  end

  defp verify_segment([_ | _], route, _acc) do
    raise ArgumentError,
          "a dynamic ~p interpolation must follow a static segment, got: #{Macro.to_string(route)}"
  end
```

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,757 @@
...
+
+  # we reached the static query string, return to caller
+  defp verify_segment(["?" <> _ = query], _route, acc) do
+    {Enum.reverse(acc), [query]}
+  end
+
+  # we reached the dynamic query string, return to call with rewritten query
+  defp verify_segment(["?" | rest], route, acc) do
+    {Enum.reverse(acc), verify_query(rest, route, ["?"])}
+  end
```
josevalim, line 525: I think this case is broken:

    /posts/#{post.id}?foo=bar&#{key}=#{value}&baz=bat

If so, this might be a fix:

```suggestion
  # we reached the dynamic query string, return to call with rewritten query
  defp verify_segment(["?" <> _ = query | rest], route, acc) do
    {Enum.reverse(acc), verify_query(rest, route, [query])}
  end
```

josevalim (approved):
I dropped last batch of comments, ship it!!!

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,757 @@
...
+
+  Unmatched routes will issue compiler warnings:
+
+      warning: no route path for AppWeb.Router matches "/postz/#{post}"
+        lib/app_web/controllers/post_controller.ex:100: AppWeb.PostController.show/2
+
+  Additionally, interpolated ~p values are encoded via the `Phoenix.Param` protocol.
+  For example, a `%Post{}` struct in your application may derive the `Phoenix.Param`
+  protocol to generate slug-based paths rather than ID based ones. This allows you to
+  use `~p"/posts/#{post}"` rather than `~p"/posts/#{post.id}"` throughout your
```
Malian, line 27: To be coherent with the first part of the sentence, isn't `post.slug` instead of `post.id`?
```suggestion
  use `~p"/posts/#{post}"` rather than `~p"/posts/#{post.slug}"` throughout your
```

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,757 @@
...
+      params = %{page: 1, direction: "asc"}
+      ~p"/posts?#{params}"
+
+  Like path segments, query strings params are proper URL encoded and may be interpolated
+  directly into the ~p string.
+
+  ## Options
+
+  To verify routes in your application modules, such as controller, templates, and views,
+  `use Phoenx.VerifiedRoutes`, which supports the following options:
```
Malian, line 46:
```suggestion
  `use Phoenix.VerifiedRoutes`, which supports the following options:
```

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,757 @@
...
+  While this one will not allow the compiler to see the full path:
+
+      type = "posts"
+      ~p"/media/#{type}/#{post}"
+
+  In such cases, it's better to write a function such as `media_path/1` which branches
+  on different `~p`'s to handle each type.
+
+  Like any other compilation warning, the Elixir compiler will warn any time the file
+  that a ~p resides in changes, or if the router is changed. To view prevoiusly issued
```
Malian, line 96:
```suggestion
  that a ~p resides in changes, or if the router is changed. To view previously issued
```

lib/phoenix/verified_routes.ex
```diff
@@ -0,0 +1,757 @@
+defmodule Phoenix.VerifiedRoutes do
+  @moduledoc ~S"""
+  Provies route generation with compile-time verificiation.
```
optikfluffel, line 3:
```suggestion
  Provides route generation with compile-time verificiation.
```

lenileiro: *verification

Fudoshiki:
with `get "*segments", PageController, :index` all routes always verified

josevalim:
Yes. We want to add an annotation to say certain routes should not count towards verification but I believe that’s not yet implemented. :)

chrismccord:
Pass `warn_on_verify: false` to that route definition

Fudoshiki:
> Pass `warn_on_verify: false` to that route definition

I can't do that for `live` routes

chrismccord:
Fixed on LV master

Issue #4893, opened 2022-07-20
:cache_static_manifest configuration is added in prod.exs with --no-assets phx.new option


maximemenager:
### Environment

* Elixir version (elixir -v):
  ```
  Erlang/OTP 25 [erts-13.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit:ns]
  
  Elixir 1.13.4 (compiled with Erlang/OTP 25)
  ```

* Phoenix version (mix deps):
  ```
  * phoenix 1.7.0-dev (../..) (mix)
    ok
  ```
* Operating system:
  ```
  Linux gabichon 5.4.0-121-generic #137-Ubuntu SMP Wed Jun 15 13:33:07 UTC 2022 x86_64 x86_64 x86_64 GNU/Linux
  ```
### Steps to reproduce

* Create a new phoenix app
```
mix phx.new foobar --no-ecto --no-assets --no-html --no-gettext --no-dashboard --no-live --no-mailer
```
* Launch the server with MIX_ENV set to "prod"

```
$ cd foobar && mix deps.get && SECRET_KEY_BASE="foobar" MIX_ENV=prod mix phx.server
```

### Expected behavior

No error message

### Actual behavior

```
➜  foobar git:(master) ✗ SECRET_KEY_BASE="foobar" MIX_ENV=prod mix phx.server                                                      
13:13:44.052 [info] Running FoobarWeb.Endpoint with cowboy 2.9.0 at :::4000 (http)
13:13:44.057 [error] Could not find static manifest at "/tmp/phoenix/installer/foobar/_build/prod/lib/foobar/priv/static/cache_manifest.json". Run "mix phx.digest" after building your static files or remove the "cache_static_manifest" configuration from your config files.
13:13:44.058 [info] Access FoobarWeb.Endpoint at https://example.com
```

I'm not sure if this is an expected behaviour, but I would have expected [these lines](https://github.com/phoenixframework/phoenix/blob/master/installer/templates/phx_single/config/prod.exs#L7-L12=) to be omitted from the template when the `--no-assets` option is given for phx.new

What do you think ?

chrismccord:
Can you send a PR? Thanks!

maximemenager:
> Can you send a PR? Thanks!

Done here: https://github.com/phoenixframework/phoenix/pull/4902

PR #4902, opened 2022-07-28
fix: skip cache_static_manifest config option with --no-assets flag

maximemenager:
This PR skips the `cache_static_manifest` option which was added to the `config/prod.exs` even if the `--no-assets` flag was given during `mix phx.new`

Fixes https://github.com/phoenixframework/phoenix/issues/4893


chrismccord:
❤️❤️❤️🐥🔥

PR #4872, opened 2022-06-23
Support --live to phx.gen.auth

bemesa21:
Added --live option to `phx.gen.auth` command.
With this new option, in addition to generating an authentication system with `Phoenix.View`s, you can also generate an authentication system using `LiveViews`.

To generate an authentication system, the following options can be used:
- `--live` to generate a system with `LiveView`
- `--no-live` to generate a system with `Phoenix.View`

Additionally, if neither of these options is given, a prompt will be displayed so that the user can choose between these two options.

TODO
  - [x] Add interactive Y/n prompt for LiveView generation on `mix phx.gen.auth` if `--live` and `--no-live` is not present

PR #4903, opened 2022-07-28
Remove hardcoded context ref in phx.gen.auth

samgaw:
Quick fix for `mount_current_user/2` which was calling an `Accounts` context regardless of what the generator created.


samgaw:
Fixed in #4872

PR #4905, opened 2022-07-29
Allow --repo option & --migration-dir for phx.gen.schema 

MikaAK:
This allows us to provide a `--repo` flag to `mix phx.gen.schema`

It seems support is already built in for this with `Mix.Phoenix.Schema` so I simply added it to the `opts` in the format it seems is expected and made it utilise that `repo` for the migration path as well

Through options i've also added a `--migration_dir` which is passed in and used instead of the generated repo path

If this is wanted lemme know and I'll add tests to get it merged

### DIFF
--- a/lib/mix/tasks/phx.gen.schema.ex
+++ b/lib/mix/tasks/phx.gen.schema.ex
@@ -89,14 +89,14 @@ defmodule Mix.Tasks.Phx.Gen.Schema do
   ## repo
 
   Generated migration can use `repo` to set the migration repository
-  folder with option `--repo`.
+  folder with option `--repo`:
 
       $ mix phx.gen.schema Blog.Post posts --repo MyApp.Repo.Auth
 
   ## migration_dir
 
-  Generated migrations can be added to a specific `migration-dir` which sets the
-  migration folder path
+  Generated migrations can be added to a specific `--migration-dir` which sets
+  the migration folder path:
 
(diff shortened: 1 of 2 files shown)


lib/mix/tasks/phx.gen.schema.ex
```diff
@@ -86,6 +86,21 @@ defmodule Mix.Tasks.Phx.Gen.Schema do
...
+
+  Generated migration can use `repo` to set the migration repository
+  folder with option `--repo`.
+
+      $ mix phx.gen.schema Blog.Post posts --repo MyApp.Repo.Auth
+
+  ## migration_dir
+
+  Generated migrations can be added to a specific `migration-dir` which sets the
+  migration folder path
```
josevalim, line 99:
```suggestion
  Generated migrations can be added to a specific `--migration-dir` which sets the
  migration folder path:
```

lib/mix/tasks/phx.gen.schema.ex
```diff
@@ -86,6 +86,21 @@ defmodule Mix.Tasks.Phx.Gen.Schema do
   Generated migration can use `binary_id` for schema's primary key
   and its references with option `--binary-id`.
 
+  ## repo
+
+  Generated migration can use `repo` to set the migration repository
+  folder with option `--repo`.
```
josevalim, line 92:
```suggestion
  folder with option `--repo`:
```

PR #4907, opened 2022-08-02
Fix a bug that remove the Telemetry when ecto is disabled

robsonpeixoto:
The bug was injected by this PR https://github.com/phoenixframework/phoenix/pull/4888


mcrumm:
Thanks!

PR #4912, opened 2022-08-11
Support multiple scope hosts. Closes #4744

### LINKED ISSUE #4744: Support `host: :*` and list of hosts on router scope
I need to look into any implications of this, but I'd like to support router scopes where you can match on the present of any host, as well as multiple hosts. For example, using LiveView's `live` routes today where you would like to route to different LiveViews based on the root host being accessed `example.com` vs `username.example.com` is not possible without branching to different routers in the endpoint. Supporting something like this would solve the issue:

```elixir
# match www, no subdomain, and localhost
scope "/", MyAppWeb, host: ["www.", "example.com", "localhost"] do
  live_session :default do
    live "/", LandingLive, :new
  end
end

# matched logged in subdomain user homepages
scope "/", MyAppWeb, host: :* do
  live_session :authenticated, on_mount: [{UserAuth, :ensure_authenticated}] do
    live "/", HomeLive, :new
  end
end


```

chrismccord:


lib/phoenix/router/route.ex
```diff
@@ -48,7 +48,7 @@ defmodule Phoenix.Router.Route do
   """
   @spec build(non_neg_integer, :match | :forward, atom, String.t, String.t | nil, atom, atom, atom | nil, list(atom), map, map, map, boolean, boolean) :: t
   def build(line, kind, verb, path, host, plug, plug_opts, helper, pipe_through, private, assigns, metadata, trailing_slash?, warn_on_verify?)
-      when is_atom(verb) and (is_binary(host) or is_nil(host)) and
+      when is_atom(verb) and (is_binary(host) or is_nil(host) or is_list(host)) and
```
josevalim, line 51: The host is only created in `Scope.push`. So my suggestion is to normalize the host to a list as soon as possible in there, by calling `List.wrap`. Everywhere else, the host is a list. Then you can rename the `host` field to `hosts` both there and here.

PR #4914, opened 2022-08-12
Put the overview guides link top of moduledoc

DianaOlympos:
I had people today that were discovering Phoenix and just thought that this was the main hub as it was linked as "Docs" from the phoenixframework.org website. I think that for a newcomer that end up there, having the guide link at the top probably would make more sense.


josevalim:

PR #4917, opened 2022-08-18
Remove repeated alias

adamwiggall:
When a project is generated, `alias Phoenix.LiveView.JS` is duplicated in the `view_helpers/0` function in `installer/templates/phx_single/lib/app_name_web.ex`

Both aliases are created inside a conditional that check `@html`.

This PR removes the second conditional / alias.


chrismccord:
I have this fixed on another branch. Thank you!

PR #4924, opened 2022-08-23
Always reconnect after connection closes

laurglia:
Phoenix creates a new WebSocket connection if the current one closes for
some reason. However, for a new connection to be created, it is required
that the following conditions hold:
- The connection did not exit cleanly. This is needed to not make the
socket reconnect after it is manually disconnected with
`Socket#disconnect`.
- The WebSocket exit code was not normal (something other than 1000).

In some cases, the connection can close with the 1000 code and still
require reconnecting. For example, when turning the internet off for
around 1 minute, and turning it back on, then the connection sometimes
closes with code 1000.

You can test that behavior on my test site: http://34.219.19.64:5000/

Here is also a video of the issue reproduced by me: https://www.youtube.com/watch?v=j76SvA3kglQ

This commit removes the requirement for the condition that the
connection has to close with code 1000 to make reconnections more
reliable. I found that the condition was first added in November 2021
with no explanation on why it was needed in the first place:
https://github.com/phoenixframework/phoenix/commit/470337db9162284b892202ce8f6018a153aa83fc

If there is a good reason why this condition is set in place, do let me
know, but otherwise I would go ahead and remove it.

The new behavior can be tested on my second test site: http://34.219.19.64:5001/

Vide of the new behavior: https://www.youtube.com/watch?v=q06kPuTln9Q


laurglia:
I may have found the reason why this behavior was needed: https://github.com/phoenixframework/phoenix/pull/2832

So basically when the server disconnects the user with `Endpoint.broadcast(socket.assigns.id, "disconnect", %{})`, then they should not reconnect.

I checked that this commit would break that behavior and start reconnecting even then. I think maybe we could use a custom close code on the server side to indicate that the connection was closed by the Phoenix server and that client should not reconnect.

jeregrine:
Hey I think I might have handled this in a slightly different way here in #5024 while trying to solve a different bug.

majeed-platogo:
We actually would like to have exactly this in our product

PR #4928, opened 2022-08-30
Disable busy waiting for docker deployments

AstolfoKawaii:
Containerized applications are designed to run in shared environments, so I suggest disabling busy waiting, not to interfere with other applications and to allow kernel to properly estimate the load (for autoscaling/...)


AstolfoKawaii:
It seems like CI is broken in master 61a4371f26d4d1d128cac3e6deeb846a7c413a95

josevalim:
I don't think we should do this by default as this may reduce response time. You may be in a virtual environment, and still have an allocated quota which busy wait would certainly be below. :) Thanks!

PR #4929, opened 2022-08-30
Return type of c:Endpoint.Cowboy2Handler.websocket_init/1

satom99:
[The documentation for `:cowboy_websocket`](https://ninenines.eu/docs/en/cowboy/2.4/manual/cowboy_websocket/) states that the `websocket_init/1` callback can return values specified in the `CallResult` type.

However, the current implementation of `Endpoint.Cowboy2Handler` **forces** the return from the handler/transport `Socket.Transport.init/1` callback to be `{:ok, state}` and cannot be other returns — like for example the simple `{:push, {:close, 4000, "Custom error"}, state}`.

Other callbacks like `websocket_handle/2` share the same `CallResult` result type, and use the private function `handle_reply/2` to translate the replies returned from the handler/transport. This PR changes the implementation of `websocket_init/2` to use that same private function and allow the aforementioned results to be returned.

---

With this change, we also want to modify the return type of `c:Socket.Transport.init/1` to account for the change in `websocket_init/1`. I have done so, and also modified the types of the rest of callbacks.

Though this last change sort of affects longpoll as well... because websocket and longpoll do not seem to share the same result types for the callbacks. **(?)**

---

**Background:** I need this because I am writing a custom websocket server and need to close the socket with custom codes in case of error on `init/1`. I.e. per example do `{:push, {:close, 4000, "Unknown user"}, state}`.

### DIFF
--- a/lib/phoenix/endpoint/cowboy2_handler.ex
+++ b/lib/phoenix/endpoint/cowboy2_handler.ex
@@ -137,8 +137,7 @@ defmodule Phoenix.Endpoint.Cowboy2Handler do
   ## Websocket callbacks
 
   def websocket_init([handler | state]) do
-    {:ok, state} = handler.init(state)
-    {:ok, [handler | state]}
+    handle_reply(handler, handler.init(state))
   end
 
   def websocket_handle({opcode, payload}, [handler | state]) when opcode in [:text, :binary] do
--- a/lib/phoenix/socket/transport.ex
+++ b/lib/phoenix/socket/transport.ex
@@ -158,7 +158,10 @@ defmodule Phoenix.Socket.Transport do
   This must be executed from the process that will effectively
   operate the socket.
   """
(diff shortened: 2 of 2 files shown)


mtrudel:
See the work in https://github.com/phoenixframework/phoenix/pull/5030

josevalim:
We keep a subset on purpose as it needs to work on an transports. Thanks! :)

mtrudel:
Technically, your desired return type will work with websocket transports on 1.7+ (the underlying [`WebSock`](https://github.com/phoenixframework/websock/blob/main/lib/websock.ex#L74) contract explicitly supports it), but as José says you'd be violating the overlying `Phoenix.Socket.Transport` behaviour by doing so. 

If you *know* you don't need any other transports (or want to special-case your code) you can do what you want today, you're just 'on your own' if you choose to do so.

PR #4935, opened 2022-09-04
Fix syntax highlight type

sharils:


sharils:
Hi @chrismccord , I just found that some of the diff are diff while some are elixir. IMO, they should be consistent, but which one?

chrismccord:
I think the elixir highlight is what we want in this case. Thank you!

PR #4927, opened 2022-08-30
Update debain to bullseye-20220801-slim

AstolfoKawaii:
bullseye-20210902-slim is outdated I suggest using latest for hexpm/elixir

PR #4937, opened 2022-09-05
Update Dockerfile template

brainlid:
Use newer Debian base image.

Newer Elixir + Erlang releases do not exist on the older 2021 base. At least not as created by the hexpm builder.

Building a new Phoenix app with Elixir 1.14.0 and Erlang 25.0.4 fails to build the docker image because Hexpm has not defined an image for the new Elixir+Erlang versions on the older Bullseye 2021 base.


brainlid:
Similar to PR #4927

chrismccord:
❤️❤️❤️🐥🔥

PR #4938, opened 2022-09-09
Permit customising HTML error page

frerich:
This PR introduces support for a new endpoint configuration setting, `:debug_banner_hooks`.

It can be used to specify functions (as MFA tuples) to be invoked when rendering the debug banner. The functions can contribute additional markup to permit customising the error page.

The motivation behind this is the desire to build the package https://github.com/frerich/phoenix_live_debug_console which (by taking advantage of https://github.com/frerich/underthehood) can be included in Phoenix projects to augment the error view with an interactive IEx terminal (based on LiveView), right in the error page.

I previously brought this idea up in https://elixirforum.com/t/embedding-interactive-iex-shell-in-debugging-page-shown-for-development-builds/49554 and the (little) feedback I got encouraged me to give this a shot. 😊 

I'd ❤️  to hear what you think!

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -215,6 +215,25 @@ defmodule Phoenix.Endpoint do
 
     * `:log_access_url` - log the access url once the server boots
 
+    * `:debug_banner_hooks` - a list of functions provided as MFA (`{module,
+      function, args}`) tuples. These functions are invoked in case
+      `:debug_errors` is set to `true` and an error occurs, in which case they may
+      contribute additional markup to be included in the error page. Each function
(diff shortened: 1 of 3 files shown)


chrismccord:
Thank you for this! We are planning a our own development toolbar/enhanced error page so it's probably premature to move forward with this at the moment. Once we have new features underway we can loop back to see if new APIs accommodate your needs. Thanks!

frerich:
> We are planning a our own development toolbar/enhanced error page

Nice! ❤️

I wasn't aware of work being done on a new and improved development toolbar/error page. That sounds very interesting -- is there a way to watch or maybe even join that effort? Maybe there's even a public Git repository to peek at?

PR #4820, opened 2022-05-07
Fix and format all templates files

leandrocp:
- Format template files, ref https://github.com/phoenixframework/phoenix/pull/4793#issuecomment-1119687158
- Fix what's currently broken to make CI pass
- ~Auto-format .ex and .exs on demand~ that would require more changes on tests, I'll work on #4939 later

Depends on:
- https://github.com/phoenixframework/phoenix/pull/4977
- https://github.com/phoenixframework/phoenix/pull/4976
- https://github.com/phoenixframework/phoenix/pull/4981
- https://github.com/phoenixframework/phoenix/issues/4998

PR #4939, opened 2022-09-09
Auto format generated templates

leandrocp:
While working on https://github.com/phoenixframework/phoenix/pull/4820 I've realized that manually formatting templates with a default config won't work well because the user's project may have a different config for line length, generating a file that won't be formatted in that project context, forcing the user to run `mix format` to fix it anyways. Besides that a project may have custom plugins that is expected to modify those files.

So why not format on demand? The test shows how it could work. WDYT?

### DIFF
--- a/installer/lib/phx_new/generator.ex
+++ b/installer/lib/phx_new/generator.ex
@@ -81,7 +81,9 @@ defmodule Phx.New.Generator do
 
         :eex ->
           contents = mod.render(name, source, project.binding)
-          create_file(target, contents)
+          file = create_file(target, contents)
+          :ok = Mix.Tasks.Format.run([target])
+          file
       end
     end
   end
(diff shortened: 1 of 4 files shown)


josevalim:
Actually, there are complications:

1. On phx.new, we don't have the HEEx formatter, so we can't format them. I still think the above would be helpful for Elixir. So I suggest checking the file extension for `.ex` and `.exs` and apply automatic formatting then.

2. For phx.gen.*, if there are plugins and the plugins are not available, then we will call compile, but the odds are low because we need phoenix compiled to invoke the generators. So it should be ok with formatting any `.ex`, `.exs`, `.heex` file.

josevalim:
Actually, there is no guarantee the user will have a formatter for .heex files, so we can truly only format .ex and .exs for now in both cases. It is a great help, but the rest has to be manually. :)

chrismccord:
Closing this for now as it has gone stale. We may pick this back up in the future. Thanks!

PR #4950, opened 2022-09-16
Add @flash assign

chrismccord:


lib/phoenix/test/conn_test.ex
```diff
@@ -274,13 +274,17 @@ defmodule Phoenix.ConnTest do
   Gets the whole flash storage.
   """
   @spec get_flash(Conn.t) :: map
-  defdelegate get_flash(conn), to: Phoenix.Controller
+  @deprecated "get_flash/1 is deprecated. Use the @flash assign provided by the :fetch_flash plug"
```
josevalim, line 277:
```suggestion
  @deprecated "get_flash/1 is deprecated. Use conn.assigns.flash instead"
```

lib/phoenix/controller.ex
```diff
@@ -1411,6 +1412,7 @@ defmodule Phoenix.Controller do
       "Welcome Back!"
 
   """
+  @deprecated "get_flash/2 is deprecated. Use Phoenix.Flash.get/2 instead"
```
josevalim, line 1415:
```suggestion
  @deprecated "get_flash/2 is deprecated. Use Phoenix.Flash.get(@flash, key) instead"
```

lib/phoenix/flash.ex
```diff
@@ -0,0 +1,33 @@
...
+  """
+  def get(%Plug.Conn{}, key) when is_atom(key) or is_binary(key) do
+    raise ArgumentError, """
+    expected a map of flash data, but got a %Plug.Conn{}
+
+    Use the @flash assign set by the :fetch_flash plug instead:
+
+        <%= Phoenix.Flash.get(@flash, :#{key}) %>
+    """
+  end
```
josevalim, line 22:
```suggestion
  def get(%mod{}, key) when is_atom(key) or is_binary(key) do
    raise ArgumentError, """
    expected a map of flash data, but got a %#{inspect(mod)}{}

    Use the @flash assign set by the :fetch_flash plug instead:

        <%= Phoenix.Flash.get(@flash, :#{key}) %>
    """
  end
```

lib/phoenix/flash.ex
```diff
@@ -0,0 +1,33 @@
...
+
+  def get(%{} = flash, key) when is_atom(key) or is_binary(key) do
+    Map.get(flash, to_string(key))
+  end
+
+  def get(other, key) do
+    raise ArgumentError, """
+    expected a flash map with atom or string key, but got: #{inspect(other)} with key: #{inspect(key)}\
+    """
+  end
```
josevalim, line 32:
```suggestion
```

I would let FunctionClauseError do the job. :)

josevalim (approved):

PR #4894, opened 2022-07-20
Update to latest gettext and remove all compilers

josevalim:

PR #4952, opened 2022-09-17
Added the --no-gettext arg to the mix phx.new command. This is to get…

ECAllen:
… rid of the annoying gettext warnings when running mix phx.server.

Below is the warning I am trying to avoid. 

warning: the :gettext compiler is no longer required in your mix.exs.

Please find the following line in your mix.exs and remove the :gettext entry:

    compilers: [..., :gettext, ...] ++ Mix.compilers(),

  (gettext 0.20.0) lib/mix/tasks/compile.gettext.ex:5: Mix.Tasks.Compile.Gettext.run/1
  (mix 1.12.2) lib/mix/task.ex:394: anonymous fn/3 in Mix.Task.run_task/3
  (mix 1.12.2) lib/mix/tasks/compile.all.ex:92: Mix.Tasks.Compile.All.run_compiler/2
  (mix 1.12.2) lib/mix/tasks/compile.all.ex:72: Mix.Tasks.Compile.All.compile/4
  (mix 1.12.2) lib/mix/tasks/compile.all.ex:59: Mix.Tasks.Compile.All.with_logger_app/2
  (mix 1.12.2) lib/mix/tasks/compile.all.ex:36: Mix.Tasks.Compile.All.run/1

### DIFF
--- a/guides/introduction/up_and_running.md
+++ b/guides/introduction/up_and_running.md
@@ -7,7 +7,7 @@ Before we begin, please take a minute to read the [Installation Guide](installat
 We can run `mix phx.new` from any directory in order to bootstrap our Phoenix application. Phoenix will accept either an absolute or relative path for the directory of our new project. Assuming that the name of our application is `hello`, let's run the following command:
 
 ```console
-$ mix phx.new hello
+$ mix phx.new hello --no-gettext
 ```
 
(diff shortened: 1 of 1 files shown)


mcrumm:
Hi @ECAllen, thanks for the PR. Unfortunately this change would recommend users skip installing gettext, rather than us fixing the `mix phx.new` task so that it generates an app that does not add the unnecessary compiler. :)

The generators were updated in the following PR:

- #4894 

...so I think we are good, thanks!

PR #4962, opened 2022-09-21
Fix: Don't alias Phoenix.LiveView.JS twice

### LINKED ISSUE #4961: Duplicate alias Phoenix.LiveView.JS
https://github.com/phoenixframework/phoenix/blob/fbc6f54190ed0b3c90fb45ed1b2d980623599868/installer/templates/phx_single/lib/app_name_web.ex#L115

I was looking at the updated generators and looks like this is aliased twice

luhagel:
Removes the duplicated alias and adds some basic indentation

Fixes #4961


chrismccord:
Fixed on master. Thank you!

PR #4967, opened 2022-09-21
Explicitly callout password input type

jsmestad:
@josevalim easy one for doc updates.


chrismccord:
Fixed on master with `:include` thanks!

PR #4970, opened 2022-09-21
Emit telemetry span events for channel join and handle_in

arkgil:
This PR adds telemetry span events for `Phoenix.Channel` `join/3` and `handle_in/3` callbacks. The new events help with instrumenting channels with distributed traces (e.g. via OpenTelemetry).

The following new events were added:

- `[:phoenix, :channel, :join, :start]`
- `[:phoenix, :channel, :join, :stop]`
- `[:phoenix, :channel, :join, :exception]`
- `[:phoenix, :channel, :handle_in, :start]`
- `[:phoenix, :channel, :handle_in, :stop]`
- `[:phoenix, :channel, :handle_in, :exception]`

I haven't added documentation for the events yet, as I want to understand if this is something that the Phoenix team would accept as a feature, and if yes, what metadata to include the events.

### DIFF
--- a/lib/phoenix/channel/server.ex
+++ b/lib/phoenix/channel/server.ex
@@ -297,12 +297,37 @@ defmodule Phoenix.Channel.Server do
     }
 
     start = System.monotonic_time()
-    {reply, state} = channel_join(channel, topic, auth_payload, socket)
-    duration = System.monotonic_time() - start
-    metadata = %{params: auth_payload, socket: socket, result: elem(reply, 0)}
-    :telemetry.execute([:phoenix, :channel_joined], %{duration: duration}, metadata)
-    GenServer.reply(from, reply)
-    state
+    measurements = %{system_time: System.system_time()}
+    metadata = %{topic: topic, params: auth_payload, socket: socket}
+    :telemetry.execute([:phoenix, :channel, :join, :start], measurements, metadata)
+
+    try do
+      {{result, reply}, state} = channel_join(channel, topic, auth_payload, socket)
+
+      measurements = %{duration: System.monotonic_time() - start}
+
+      :telemetry.execute(
+        [:phoenix, :channel, :join, :stop],
+        measurements,
+        Map.merge(metadata, %{result: result, reply: reply, socket: get_socket(state)})
+      )
+
+      :telemetry.execute([:phoenix, :channel_joined], measurements, %{
+        params: auth_payload,
+        socket: socket,
+        result: result
+      })
+
+      GenServer.reply(from, {result, reply})
+      state
+    catch
+      kind, reason ->
+        measurements = %{duration: System.monotonic_time() - start}
+        stacktrace = __STACKTRACE__
+        metadata = Map.merge(metadata, %{kind: kind, reason: reason, stacktrace: stacktrace})
+        :telemetry.execute([:phoenix, :channel, :join, :exception], measurements, metadata)
+        :erlang.raise(kind, reason, stacktrace)
+    end
   end
 
   def handle_info(%Message{topic: topic, event: "phx_leave", ref: ref}, %{topic: topic} = socket) do
--- a/lib/phoenix/channel/server.ex
+++ b/lib/phoenix/channel/server.ex
@@ -298,7 +298,7 @@ defmodule Phoenix.Channel.Server do
 
     start = System.monotonic_time()
     measurements = %{system_time: System.system_time()}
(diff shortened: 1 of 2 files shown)


lib/phoenix/channel/server.ex
```diff
@@ -314,11 +339,33 @@ defmodule Phoenix.Channel.Server do
...
+
+      :telemetry.execute(
+        [:phoenix, :channel, :handle_in, :stop],
+        measurements,
+        Map.merge(metadata, %{socket: get_socket(state), result: result})
+      )
+
+      :telemetry.execute([:phoenix, :channel_handled_in], measurements, metadata)
+
+      state
```
arkgil, line 360: This isn't state of the GenServer, but the return value of handle_info when joining the channel is also called "state". I'm happy to change it if someone has a better idea.

lib/phoenix/channel/server.ex
```diff
@@ -314,11 +339,33 @@ defmodule Phoenix.Channel.Server do
...
+    try do
+      result = socket.channel.handle_in(event, payload, put_in(socket.ref, ref))
+      state = handle_in(result)
+
+      measurements = %{duration: System.monotonic_time() - start}
+
+      :telemetry.execute(
+        [:phoenix, :channel, :handle_in, :stop],
+        measurements,
+        Map.merge(metadata, %{socket: get_socket(state), result: result})
```
arkgil, line 355: I extract the socket from the return value of handle_in so that the stop event carries the updated socket (e.g. in case new assigns were added).

arkgil, line 355: Having the full return value of `handle_in/3` in the metadata is very helpful. I'm not sure about naming though, because the `[.., :join, :stop]` event also has the `:result` key, but it's either `:ok` or `:error` (to be consistent with the existing `:channel_joined` event).

josevalim:
Thank you!

Our position so far is that we don't instrument GenServer by default, so we shouldn't do it here, but I would like to hear opinions. In particular, I am concerned about this and LiveView now emitting duplicate events (since LiveView does trace some of those). So I would say go with LiveView traces for the regular case and, for the channels one, it is on the developer. :)

arkgil:
Ah, I forgot about LiveView. Could we emit LiveView's events based on these events? I.e. install a handler for the channel events that triggers the LiveView events.

From my perspective, having these events emitted by Phoenix makes instrumenting channels much more convenient and robust (e.g.. we can have Phoenix validate the return values of the callbacks and emit either a `stop` or `exception`). Implementing this in the app's codebase requirse a wrapper behaviour around `Phoenix.Channel`, introducing a new concept, which makes the codebase slightly harder to understand for newcomers.

Of course I understand if this is not something you want to maintain. Thanks for the feedback!

josevalim:
> Ah, I forgot about LiveView. Could we emit LiveView's events based on these events? I.e. install a handler for the channel events that triggers the LiveView events.

Those events are too low-level for LiveView and they will end-up emitting LiveView's private state, which is one additional concern for this PR: if we emit the socket, we will expose LiveView internals. If we don't emit the socket, then the events are not very useful. :(

So we would need, at a minimum, to opt-in/opt-out somehow.

arkgil:
Is emitting the internal LiveView state a huge concern? The regular Phoenix socket also has a bunch of data no one should fiddle with, but it's very easy to access and modify.

---

Regarding opt-in/out, we could make this configurable on a per-channel basis, similar to how we can enable/disable events for specific routes:

```elixir
defmodule MyWeb.Channel do
  use Phoenix.Channel, telemetry_span_events: true
end
```

The option would be `false` by default, to not introduce new events in existing apps. Depending on how useful you think this is, the option could be explicitly set to `true` in newly generated Phoenix projects (I don't have a strong opinion here).

arkgil:
@josevalim what do you think about ☝️ ?

josevalim:
> Is emitting the internal LiveView state a huge concern?

Yes, it is. The Phoenix.Socket is structured as a struct with "private" fields. You can access it but those fields are hidden. Phoenix.LiveView internal state can be regular maps and tuples, and forcing it to be a special struct to hide information that is being accidentally exposed by Phoenix.Channel feels backwards.

About the opt-in/opt-out API, I leave that decision to Phoenix maintainers. :)

josevalim:
Closing this for now, this discussion surfaced again and my concerns outweight the potential benefits. :) IF you really want to instrument all events, it is easy to do with  `def handle_event(...) do :telemetry.span(fn -> do_handler_event(...) end) end`.  Thank you!

PR #4972, opened 2022-09-23
Fix markdown issues

kianmeng:
Found via `markdownlint --ignore installer --ignore integration_test --disable MD004 MD007 MD013 M  D014 MD024 MD026 MD028 -- **/*.md`

See a detailed description of the rules is available at https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md

Not sure about the `instaler` folder and its usage, hence I ignored it from checking.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -26,7 +26,7 @@ making it easier for developers to [contribute to Phoenix](#pull-requests).
 ## Bug reports
 
 A bug is either a _demonstrable problem_ that is caused by the code in the repository,
-or indicate missing, unclear, or misleading documentation. Good bug reports are extremely 
+or indicate missing, unclear, or misleading documentation. Good bug reports are extremely
 helpful - thank you!
 
 Guidelines for bug reports:
@@ -134,8 +134,10 @@ in order to craft an excellent pull request:
    ```bash
    # Clone your fork of the repo into the current directory
    git clone https://github.com/<your-username>/phoenix
+
    # Navigate to the newly cloned directory
    cd phoenix
+
    # Assign the original repo to a remote called "upstream"
    git remote add upstream https://github.com/phoenixframework/phoenix
    ```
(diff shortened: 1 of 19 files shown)


josevalim:
Hi @kianmeng!

ExDoc recognizes `shell` but not `console`, and it uses the `$` to make the prompt recognizable but not selectable. So I would keep using `bash` instead of `console` and keep using the prompts. :)

kianmeng:
> ExDoc recognizes `shell` but not `console`, and it uses the `$` to make the prompt recognizable but not selectable. So I would keep using `bash` instead of `console` and keep using the prompts. :)

ExDoc have supported the `console` name in shell lexer since v0.25.1? See
https://github.com/elixir-lang/ex_doc/commit/8b754f45160e65d38b4b792bded1c4893b3a892a

The reason for removing `$` from some code indent block without output, for example, the `README.md`, is when viewing these Markdown files from a browser in GitHub, the `$` is always selectable when we click the code indent block for copying. Either we've to manually remove the `$` after copying, or drag and select the command without `$`.

And to follow the [contribution guideline](https://github.com/phoenixframework/phoenix/blob/master/CONTRIBUTING.md#guides):

> We use the "console" code fence for iex and shell commands.

Also, see https://github.com/markdownlint/markdownlint/blob/master/docs/RULES.md#md014---dollar-signs-used-before-commands-without-showing-output

josevalim:
I will update the guides. :) I would like to keep `$`, because I think it is important context. :)

josevalim:
And thanks for letting me know that `console` works, it is ok to migrate to it then. :)

kianmeng:
@josevalim Done. I maintain the `$` prompt in those documents in `guides` folder, but remove it in `README.md`, so it matches with the instruction in `CONTRIBUTING.md`.

josevalim:

kianmeng:

PR #4973, opened 2022-09-23
Add support for Sock API

mtrudel:
This PR adds support for the [Sock](https://github.com/mtrudel/sock) API as a generic abstraction over WebSockets. This allows web server implementations to fully support Phoenix without it having to provide facilities for each server implementation (as is currently the case with `Phoenix.Endpoint.Cowboy2{Adpater,Handler}`). By way of example, given the content of this PR [Bandit](https://github.com/mtrudel/bandit) is now able to host the entirety of Phoenix with no Bandit-specific code in this project. Providing support for Sock in Phoenix is also the first step towards removing Cowboy-specific handler code in Phoenix, rough plans for which are outlined  at https://github.com/mtrudel/sock/issues/1.

~Sock conformance is implemented in the `Phoenix.Endpoint.Sock` module, which is `use`d in application endpoints in a similar manner to how `Plug.Builder` is. The Sock implementation itself is quite self-contained, however the socket path building functions within Endpoint are required when building out matching functions at compile time, and so have been factored out to their own module.~ This is no longer the case; see below.


lib/phoenix/endpoint/socket_path_helpers.ex
```diff
@@ -0,0 +1,67 @@
...
+    end_path_fragment = Keyword.fetch!(config, :path)
+
+    {vars, path} =
+      String.split(path <> "/" <> end_path_fragment, "/", trim: true)
+      |> Enum.join("/")
+      |> Plug.Router.Utils.build_path_match()
+
+    conn_ast =
+      if vars == [] do
+        quote context: module do
```
mtrudel, line 49: `module` here is the eventual context in which this AST will be applied. I don't believe that the usual `Macro.var(:conn, nil)` approach will work here, since the final point-of-use for this intends for `conn` to bind against a variable defined in the containing macro itself (see https://github.com/phoenixframework/phoenix/pull/4973/files#diff-9f7279f1d916a972e876499435496ca3c1ea0aa688cc7228ce1e3ba4043f5994R165). I'd love to be proven wrong here, but 'functions used to build quotes which are later included in other macros across module boundaries' is rather at the outer limits of my meta programming universe :). 

A similar case exists on line 59, just below.

test/phoenix/integration/websocket_socket_test.exs
```diff
@@ -11,15 +11,6 @@ defmodule Phoenix.Integration.WebSocketTest do
   @port 5907
```
mtrudel, line 11: Changes to this file are a large diff, but amount to nesting existing tests inside a comprehension over bandit and cowboy servers (in order to test both the existing `Cowboy2Handler` call path, and the new `Sock` based one.

test/phoenix/integration/websocket_channels_test.exs
```diff
@@ -11,15 +11,6 @@ defmodule Phoenix.Integration.WebSocketChannelsTest do
   @moduletag :capture_log
```
mtrudel, line 11: Changes to this file are a large diff, but amount to nesting existing tests inside a comprehension over bandit and cowboy servers (in order to test both the existing `Cowboy2Handler` call path, and the new `Sock` based one.

test/phoenix/integration/websocket_channels_test.exs
```diff
@@ -482,50 +597,72 @@ defmodule Phoenix.Integration.WebSocketChannelsTest do
         WebsocketClient.close(sock)
 
         assert_receive {:DOWN, _, :process, ^channel, shutdown}
-                       when shutdown in [:shutdown, {:shutdown, :closed}]
+                       when shutdown in [
+                              :shutdown,
+                              {:shutdown, :closed},
+                              {:shutdown, :local_closed}
```
mtrudel, line 603: This is a net-new line, since Sock returns a different tuple

test/phoenix/integration/websocket_channels_test.exs
```diff
@@ -539,58 +676,80 @@ defmodule Phoenix.Integration.WebSocketChannelsTest do
...
         assert_receive {:DOWN, _, :process, ^sock, :normal}
         assert_receive {:DOWN, _, :process, ^chan1, shutdown}
-        #shutdown for cowboy, {:shutdown, :closed} for cowboy 2
-        assert shutdown in [:shutdown, {:shutdown, :closed}]
+
+        # shutdown for cowboy, {:shutdown, :closed} for cowboy 2, {:shutdown, :peer_closed} for bandit
+        assert shutdown in [:shutdown, {:shutdown, :closed}, {:shutdown, :peer_closed}]
         assert_receive {:DOWN, _, :process, ^chan2, shutdown}
-        assert shutdown in [:shutdown, {:shutdown, :closed}]
+        assert shutdown in [:shutdown, {:shutdown, :closed}, {:shutdown, :peer_closed}]
```
mtrudel, line 683: This is a net-new line, since Sock returns a different tuple

test/phoenix/integration/websocket_channels_test.exs
mtrudel, line 11: Changes to this file are a large diff, but largely amounts to nesting existing tests inside a comprehension over bandit and cowboy servers (in order to test both the existing Cowboy2Handler call path, and the new Sock based one.

There are two exceptions to this, outlined below.

test/phoenix/integration/websocket_channels_test.exs
```diff
@@ -482,7 +510,7 @@ defmodule Phoenix.Integration.WebSocketChannelsTest do
         WebsocketClient.close(sock)
 
         assert_receive {:DOWN, _, :process, ^channel, shutdown}
-                       when shutdown in [:shutdown, {:shutdown, :closed}]
+                       when shutdown in [:shutdown, {:shutdown, :closed}, {:shutdown, :local_closed}]
```
mtrudel, line 513: This is a net-new line, since Sock returns a different tuple

test/phoenix/integration/websocket_channels_test.exs
```diff
@@ -540,9 +568,9 @@ defmodule Phoenix.Integration.WebSocketChannelsTest do
         assert_receive {:DOWN, _, :process, ^sock, :normal}
         assert_receive {:DOWN, _, :process, ^chan1, shutdown}
         #shutdown for cowboy, {:shutdown, :closed} for cowboy 2
-        assert shutdown in [:shutdown, {:shutdown, :closed}]
+        assert shutdown in [:shutdown, {:shutdown, :closed}, {:shutdown, :peer_closed}]
         assert_receive {:DOWN, _, :process, ^chan2, shutdown}
-        assert shutdown in [:shutdown, {:shutdown, :closed}]
+        assert shutdown in [:shutdown, {:shutdown, :closed}, {:shutdown, :peer_closed}]
```
mtrudel, line 573: These are net-new, since Sock returns a different tuple

test/phoenix/integration/websocket_socket_test.exs
mtrudel, line 11: Changes to this file are a large diff, but amount to nesting existing tests inside a comprehension over bandit and cowboy servers (in order to test both the existing Cowboy2Handler call path, and the new Sock based one, with two changes outlined below. 

No changes to tests other than indenting / wrapping in a comprehension

lib/phoenix/endpoint/socket_path_helpers.ex
mtrudel, line 49: `module` here is the eventual context in which this AST will be applied. I don't believe that the usual `Macro.var(:conn, nil)` approach will work here, since the final point-of-use for this intends for `conn` to bind against a variable defined in the containing macro itself (see [lib/phoenix/endpoint/sock.ex:165](https://github.com/phoenixframework/phoenix/pull/4973/files#diff-9f7279f1d916a972e876499435496ca3c1ea0aa688cc7228ce1e3ba4043f5994R165)). I'd love to be proven wrong here, but 'functions used to build quotes which are later included in other macros across module boundaries' is rather at the outer limits of my meta programming universe :). 

A similar case exists on line 59, just below.

josevalim:
Hi @mtrudel!

Thank you for the PR! :heart:

I believe this is a direction we should explore but I would like to see the Sock module be a concretion instead of an abstraction. I.e. I would like to define a single `Phoenix.Endpoint.Sock` and be able to use it across multiple endpoints, similar to how the Cowboy handler is the same across all endpoints/web apps.

If you go down this path, I would actually think integrating it would be easier and require fewer changes to Phoenix. Perhaps it could even be used via a `phoenix_sock` library.

I am also worried about the name (sorry!). We already have:

* transports
* sockets
* channels

And I think introducing something called `Sock` will add to the confusion. Especially if it is WebSockets only. But don't rename it for now, this is just food for thought. We can even consider calling it PlugWS and make it official part of Plug depending on how it goes?

mtrudel:
In terms of Sock being a concretion, do you mean something similar to [this](https://github.com/mtrudel/phoenix/blob/274c0ad391c7415ee1374e4112fc234164c964b3/lib/phoenix/endpoint/sock.ex)? I'd originally implemented Sock support this way earlier in the week, but pivoted to the compile-time macro style on this PR as it would make it easier to refactor the innards of `Endpoint` once `(Plug.)Cowboy` is made to support Sock [1]. If you feel it's a better approach to keep it structurally separate I'm happy to resurrect the structure represented above into this PR, I'm just trying to be mindful of where future refactors are likely to lead. The test changes in the above PR would stand, I think, in order to ensure that `Phoenix.Endpoint.Sock` is covered.

I don't however think that it would make sense to do this work in a separate library, as then the ultimate external API of `Phoenix.Endpoint` becomes the `__handlers__` function.

In terms of naming, the existing presence of `Socket` in Phoenix is exactly why I didn't name the API `Socket`, choosing instead the more diminutive and distinct `Sock` specifically to avoid confusion. It's more of a proper name than than a common one.

[1] The Sock API presumes that the implementing server is responsible for determining if the client request is a WebSocket request or not *before* making the call to Sock (that is, the server has run [RFC6455§4.2.1](https://datatracker.ietf.org/doc/html/rfc6455#section-4.2.1) before a call to `c:Sock.negotiate/2` is made, and no call to `Plug` is ever undertaken). This is in contrast with how Cowboy does things, where the handler implementation is expected to be the one to determine whether or not it's a WebSocket connection, and to return a `{:websocket, ...}` tuple if so. As such, if/once we can do this work upstream of Phoenix (presumably in the `Plug.Cowboy` project), all request will come to Phoenix already broken out into HTTP requests or WebSocket requests, via Plug and Sock respectively. At this point, the need for `__handlers__` can disappear, replaced with compile-time generation of the relevant `do_sock_negotiate` functions. In short, the end goal here is to have Phoenix's `Sock` implementation look an awful lot like its (well, `Plug.Builder`'s) `Plug` implementation.

josevalim:
> In terms of Sock being a concretion, do you mean something similar to [this](https://github.com/mtrudel/phoenix/blob/274c0ad391c7415ee1374e4112fc234164c964b3/lib/phoenix/endpoint/sock.ex)?

Yes! I would rather meta-program a tiny bit (handlers) rather than the whole module. Can we move to this direction? I think it will give me a better sense of the moving parts.

mtrudel:
No problem - it'll be done in a few hours (I mostly need to work the upstream changes in Sock and Bandit). The funny thing is that I actually waffled on submitting this implementation thinking it wasn't clever enough ('need more ~cowbell~ meta-programming')

mtrudel:
Updated to use a concrete `Phoenix.Endpoint.Sock` that looks up WebSockets via `Phoenix.Endpoint.__handlers__` at runtime. The test updates stand, as they provide coverage of this new Sock endpoint.

josevalim:
Thank you @mtrudel!

This approach is definitely better but unfortunately it confirms my suspicions that we are adding a new layer. Today, Phoenix works like this:

     CoAP
              \
    WS Cowboy -- Transport -- Socket -- Channel
              /
    WS Bandit

But with Sock:

    WS Cowboy \         
               - Sock \
    WS Bandit /        - Transport -- Socket -- Channel
                 CoAP /  


So from an abstraction point of view, we are introducing a new concept, while we already have a concept (called transports) that works both with WS and other protocols such as CoAP.

Therefore, if we were to move in this direction, I would rather make the transport-socket communication the abstraction. The downside is that it is a little bit less WebSocket specific (so it may be less useful to Bandit?) but I think that's what would make sense for Phoenix.

Thank you!

mtrudel:
Thanks for the thoughtful notes José.

This discussion has quite helpfully drawn a line for me between the ‘routing/negotiation part’ and the ‘socket part’ of the websocket lifecycle and how Sock can better accommodate them. I’d originally thought that something like `Plug.Router` would be the ultimate realization of routing within Sock, but I’ve come to realize that the routing aspects of a websocket request stand quite apart from the post-connection aspects, and that splitting them up into discrete APIs will result in a cleaner implementation.

To this end, a modest proposal:

I will split apart the `Sock` API into a `Sock.Negotiation` behaviour and a `Sock.Socket` behaviour (better names TBD). Generally speaking, the `c:Sock.negotiate/2` callback will end up in the negotiation API, and everything else will end up in the socket API. This provides one API that can answer ‘which socket implementation should I use for a given connection’, and another that can handle the lifecycle of an actual connection.

This allows Phoenix (and other Sock implementors) to think of the two aspects separately, as they really do represent different abstractions (one is an endpoint concern, the other is a socket concern). I’m not sure your proposal would suitably account for the ‘routing/negotiation’ requirements of a WebSocket adapter otherwise.

On the negotiation side, I propose to implement `Sock.Negotiation` behaviour largely as-is within the `Phoenix.Endpoint.Sock` module. Ultimately, this should be done at compile-time within the Endpoint as a simple call to `Phoenix.Transports.WebSocket.connect/4`, but I would propose that this work is better done if / after `Cowboy2Handler` goes away in favour of a pure Plug/Sock public API.

This callback would return an `{module, state}` pair to Sock, where `module` is that defined in the relevant `socket` directive within Phoenix.Endpoint. Sock will then run a socket lifecycle directly against the `Sock.Socket` implementation in `module`.

This approach would allow for an eventual collapse of most of the routing concerns to be a compile-time concern isolated within Endpoint, more or less replacing `__handler__` with `c:Sock.negotiate/2` calls (not as part of this PR, just looking to the future). This was the tack that I’d originally taken in this PR, and I’m happy to revisit it (again) if you think it’s more reasonable given the reduced scope of this new negotiation behaviour. 

On the socket side, I would model `Sock.Socket` on the subset of `Phoenix.Transport` that makes sense. Candidly, I have some reservations with this approach (primarily, I disagree with modelling a transport-specific API such as Sock on a general-purpose API such as Phoenix.Socket.Transport that is (by design) a lowest-common denominator compromise between WebSockets & other transports). However I’m happy to disagree and commit if it means getting this over the line. 

WDYT?

mtrudel:
> Yes, I agree with this concern. I don't think our abstraction is generally useful for Bandit but it is generally useful for Phoenix, which is part of the problem. It may be that a bandit_phoenix is the best option, but I will be glad to abstract small chunks away, especially when it comes to negotiation.

To my mind, it's not so much a matter of 'useful for Bandit' as it is 'useful for a generalized WebSocket abstraction' [1].
`Phoenix.Socket` is similar to a WebSocket insofar as it represents an abstract notion of 'an end to end pipe that can send and receive bytes', but the devil is - as always - in the details. Things which are squarely WebSocket concerns (connection close codes are a good example) should absolutely be exposed in a WebSocket API such as Sock, and should also equally absolutely *not* be exposed to `Phoenix.Socket`. The interfacing of the two concerns is, of course, the job of a transport module (`Phoenix.Transports.WebSocket` in this case).

Put another way, I think it should be the role of `Phoenix.Transports.WebSocket` to mediate between a WebSocket connection and a `Phoenix.Socket` instance (much as it's the role of `Phoenix.Transports.LongPoll` et al to mediate between a Plug client and a `Phoenix.Socket` instance). 

The difference from my most recent proposal is modest - I think it ought to be `Phoenix.Transports.WebSocket` which implements `Sock.WebSocket`, and which thinly wraps calls to `Phoenix.Socket` instances. This allows the `Sock.WebSocket` API to stay fully featured, and also (IMO anyway) actually makes Phoenix's code cleaner (the first place people are going to go looking to read about WebSocket support in Phoenix is likely going to be `Phoenix.Transports.WebSocket`, and seeing a nice concrete implementation there would be a great jumping off point for people).

WDYT?

[1] The (well, 'my') eventual goal here is to have Sock be *the* interface that Phoenix exposes for WebSocket support Specifically, I'm aiming to eventually push up a PR to remove `Cowboy2{Adapter,Handler}` from Phoenix in favour of Plug.Cowboy taking over that responsibility (Or perhaps even folding Sock into something like PlugWS, as you'd suggested above. Either way it's a conversation for a different PR).

mtrudel:
I've put up a rough WiP at https://github.com/phoenixframework/phoenix/compare/master...mtrudel:phoenix:split_sock_api to demonstrate the intent I'm going for viz a viz:

* Having `Phoenix.Endpoint.Sock` implement `Sock.Connector` (with the intent of having this become a compile-time concern within Endpoint in the near future, pending getting Sock awareness landed in Plug.Cowboy).
 
*Having `Phoenix.Transports.WebSocket` implement `Sock.WebSocket` (functioning as a adapter between `Sock.WebSocket`'s WebSocket-focused API & `Phoenix.Socket.Transport`'s more generic API)

I think that consolidating the WebSocket implementation that had previously been spread between `Cowboy2Handler` and `Phoenix.Transports.WebSocket` into a single module under the Transport umbrella makes the WebSocket concerns much more centralized and easy to reason about (to your earlier point, this doesn't add another abstraction so much as it fleshes out & centralizes WebSocket code within the Transport abstraction).

I'm less in love with the `Sock.Connector` routing abstraction here, but I think that once it has a chance to get compile-time evaluated in Endpoint it'll make a LOT more sense. Hopefully the existence of the `Phoenix.Endpoint.Sock` module is a short term thing on the road to that.

Note that this branch is *very* rough and none of the accompanying work has been done in Sock (or Bandit, for that matter). This is just to illustrate the approach to avoid wasted work.

josevalim:
> [1] The (well, 'my') eventual goal here is to have Sock be the interface that Phoenix exposes for WebSocket support Specifically,

Right, to be clear, this is a non-goal for us. Because we already have an API in Phoenix that works for WebSocket and other transports. So introducing an additional contract for us is really a diminishing return. Looking at the proposed PR, it still has to implement a new set of callbacks and, even if we keep the number of implementations inside Phoenix the same (replacing the Cowboy2Handler by the Sock one), we are increasing the number of abstractions. I understand and agree with the benefits (encapsulating everything in transport) but I don't think (at the moment) the pros offset the cons.

To me the big question is if we can make the contract become this: https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/socket/transport.ex#L105. If Sock exposes a superset of this contract (we can augment the current one in Phoenix to bring pending features that Sock need, we can even deprecate some contracts), then it is perfect because we will keep the exact same number of abstractions.

So my question is: how would you change the existing Phoenix.Socket.Transport so it is actually usable by Sock? And what would you change in Sock to make this easier? Do you think we can pursue this direction?

josevalim:
And before I forget, thank you for the replies and interesting discussion!

mtrudel:
> Right, to be clear, this is a non-goal for us.

For clarity, do you mean 'we won't entertain declaring Sock conformance under any circumstances' or 'we won't entertain declaring Sock conformance if it means introducing another abstraction beyond Transport'?  The answer to this really helps clarify if the best path forward here is to either:

1. Continue the work here with the goal of having Phoenix explicitly declare Sock conformance directly, or
2. Tighten up / formalize some of Phoenix's existing public APIs and have a `phoenix_sock` project that declares Sock conformance based on top of these public APIs

FWIW, my preference (and TBH I think the easiest outcome for all of us to maintain / understand) is squarely the first option.

As previously, I think it's useful to consider the endpoint and transport concerns separately. 

## Endpoint concerns

Currently, Phoenix doesn't provide any public way to expose which transport handler (if any) should be used to handle a given upgrade request. The `__handler__` function is more or less this, but it would need some work (and a rename) to really be fit for public use. It's my goal with the `Sock.Connector` behaviour to define a clear public API for this purpose, and to have `Phoenix.Endpoint.Sock` (and ideally eventually `Phoenix.Endpoint` at compile-time) implement this behaviour so that servers can turn a client's HTTP request path to a `Sock.WebSocket` instance that can handle the connection itself.

## Transport concerns

To your point, Phoenix already exposes a (nearly) fit for purpose public API in the form of `Phoenix.Socket.Transport`. For the parts of the WebSocket lifecycle that it covers I would be perfectly OK using it as-is. Enumerating the places where it falls short (it's a small list):

* Improved support for timeout / connection error conditions. If `c:Phoenix.Socket.Transport.terminate/2` defined some extra atoms beyond `:closed`, (say `':timeout`, `:remote_error`, `:shutdown` as a first-pass) I think that would be sufficient. I would be fine foregoing explicit connection close codes (that's admittedly a bit of a completionist niche anyway).

* I was really hoping to have Sock provide access to a live Socket as a passed-in parameter that could be used to send data directly on (rather than having to return back a payload as part of a `:push` response value), but having to implement support for that on top of Cowboy's websocket support was going to be a battle. I'm fine to let that go (note that, in my most recent workup above, this is already done).

* `c:Phoenix.Socket.Transport.child_spec/1` is unrelated to Sock's needs (TBH, I'd always wondered why it's part of `Phoenix.Socket.Transport` and not `Phoenix.Socket`). I mention this primarily because your statement above that "If Sock exposes a superset of this contract" isn't quite correct; Sock and Transport's APIs would look like a classic two-circle Venn diagram.

* There's also one concern that I'm not sure should be handled as an endpoint or a transport concern, and that's where and how to capture the work done in `Phoenix.Transports.WebSocket.connect/4`. On the one hand, I see this as an endpoint concern that should be called as part of `Sock.Connector`'s responsibilities. On the other hand, it's ultimately the caller of `Phoenix.Socket.connect/1`. My only real requirement here (if we go with option 1 above) is to have this captured *somewhere* in a Sock callback.

Other than that last point, I think all of the other issues are quite surmountable. To my mind, it comes down to how you'd answer the question I posed at the top of this reply. At the end of the day I'd like to see Bandit (and Cowboy!) be able to treat Phoenix as it would any other application, via a clear and public Plug/Sock API (and ideally within Phoenix itself). I'm happy to hammer Sock into whatever shape makes sense to make that happen.

mtrudel:
> It is option 2: "we won't entertain declaring Sock conformance if it means introducing another abstraction beyond Transport".

🥳 

> FWIW, we should avoid moving things to compile time, except the minimum necessary (such as the `__handler__` function).

(Assuming it's not intended to serve as a public API), the only use for `__handler__` is to construct the data used to dispatch WebSocket requests (which should eventually only be done via the `Sock.Connector` implementation) & long poll requests (which (I'm pretty sure?) could be inlined as `plug` macro calls. Anyway, a discussion for another day :smile:
 
> Another option is to introduce `handle_close`, which is specific about connection closes, and that by default returns `{:stop, :closed, state}`. WDYT?

I like this! (also maybe `handle_timeout`?) Sock would need to translate the reason from the connection close code into an atom, otherwise we'd be leaking RFC6455 concerns into the `Phoenix.Socket.Transport` API. That's easy to do though.

> Definitely agreed this would be a better API. But maybe we can support the current one and, once support is added to Cowboy, we figure out a way to return from anywhere? (it would need a process dictionary though).

Agreed that this is fine to defer.

> > c:Phoenix.Socket.Transport.child_spec/1 is unrelated to Sock's needs (TBH, I'd always wondered why it's part of Phoenix.Socket.Transport and not Phoenix.Socket)
> 
> Agreed. We would need to remove this one.

:+1:

> Given Sock is meant to be transport agnostic, then I think this is squarely outside of the Sock callbacks. The idea is to have different transports that end-up using the same Sock. So we would need a contract that is us specifically telling the webserver, early on, if something is meant to be websockets or regular request.

I assume you mean 'Given Phoenix.Socket is meant to be transport agnostic', as Sock is *very* much focused on WebSockets. Regardless, the more I think about it, the more I think it's a connector level concern and not a socket concern (that is, the set of Sock.WebSocket callbacks starts with `init/1`, not `connect/1`).

> We could also handle this by adding `Plug.Conn.upgrade/2` to Plug. This way the whole negotiation is a regular Plug request/response.

Now *this* is interesting.

Bandit's upgrade flow is (unsurpringly) based entirely on Plug.Conn calls (https://github.com/mtrudel/bandit/blob/main/lib/bandit/websocket/handshake.ex) and in fact uses a callback response to make the switch from HTTP to RFC6455 (https://github.com/mtrudel/bandit/blob/main/lib/bandit/http1/handler.ex#L24). It would be quite easy to move the upgrade flow to be Plug-powered. Maybe something like this:

* Define a `Plug.WebSocketUpgrade` plug gets run that checks the connection to see if the conn represents an upgrade request
  * This plug is expected to match a single WebSocket path, and takes arguments specifying the socket that handles connections to it (along with options).
  * If it is indeed a valid upgrade, it calls`Plug.Conn.upgrade/2` to signal that HTTP processing should halt and the adapter should do an upgrade to WebSocket, passing control to the indicated socket

From Phoenix.Endpoint's perspective, it would look like:

* `socket` macros translate directly into at most two `Plug.Router` calls (one for long polling (already a Plug), and one to `Plug.WebSocketUpgrade` with the socket details passed as arguments)
* WebSocket negotiation becomes an entirely Plug-mediated concern. `__handler__` goes away entirely, as does `Sock.Connector`. Plug.Router does it all
* All the param extraction stuff (https://github.com/phoenixframework/phoenix/blob/master/lib/phoenix/endpoint.ex#L668-L722) folds into the implementation in Plug.Router

In addition, it simplifies the Sock API to only have to consider the socket concerns (ie: Sock.Connector would go away entirely)

It would also make the eventual changes to Plug.Cowboy to add Sock support much simpler, as it sits well with how Cowboy expects upgrades to be indicated.

The socket side of things would be largely unchanged from what we've just discussed.

Exciting!

There's obviously a lot of loose ends & things unaddressed above, but I think the idea has a LOT of potential and I'm happy to spike out some work to investigate. WDYT?

josevalim:
The only part that is not clear to me is what are the arguments we pass to `Plug.Conn.upgrade/2`? Do we leave it per adapter, as to not force `Sock`/`Transport` contract?

josevalim:
For Cowboy, the args needs to be `{arg, opts}`. Should we also hardcode what is the type of upgrade? So something like:

    Plug.Conn.upgrade(conn, :websockets, {arg, opts})

And what happens from there is Cowboy specific. Then Sock can provide its own upgrade that wraps the adapter and make it agnostic to Bandit/Cowboy?

    Sock.upgrade(conn, :websockets, arg, opts)

josevalim:
> based entirely on Plug.Conn calls (https://github.com/mtrudel/bandit/blob/main/lib/bandit/websocket/handshake.ex) 

Nice! Btw, you want to use `String.downcase(..., :ascii)` almost everywhere as it is more performant and most likely the only use case for you?

mtrudel:
I wonder if the answer here is to bring Sock into Plug? (`Plug.WebSocket`, `Plug.PlugWS`?) The spec would be:

```
@spec upgrade(Plug.Conn.t(), module(), opts()) :: Plug.Conn.t()
```

mtrudel:
> Nice! Btw, you want to use `String.downcase(..., :ascii)` almost everywhere as it is more performant and most likely the only use case for you?

Good to know! Thanks! This is a hot path now (on every request), but if we end up going this upgrade route it would only get called by the upgrade plug, which would only get called on HTTP requests to the web socket endpoint

mtrudel:
It's a good call that we need to be able to pass opts back to the server (`timeout, compress`, etc). I'll make sure that that gets handled.

josevalim:
> Do you mean that having Plug own a Plug.WebSocket definition is too restrictive

Yes. I think imposing a particular contract for sockets. For example, the issue you mentioned that Cowboy cannot do out-of-band sending of text/binary. It is fine for Phoenix, fine for Sock, but not for Plug. :)

---

So here is my suggestion:

1. Let's add `Plug.Conn.adapter_upgrade/3`
2. Let's then update `Plug.Cowboy` to accept `Plug.Conn.adapter_upgrade(conn, :websockets, {handler, arg, opts})` (does this second argument make sense to you?)
3. Let's then refactor Phoenix to use it

This should solve half of the problem. Then we touch point again to discuss about Sock. WDYT?

mtrudel:
Great plan overall! I think there may be a slight shortcoming in the API in step 2:

> Let's then update Plug.Cowboy to accept Plug.Conn.adapter_upgrade(conn, :websockets, {handler, arg, opts}) (does this second argument make sense to you?)

The second argument is conveying two things to the adapter:

First it's saying 'you should upgrade to the specified protocol' (to be concrete, in the context of `Cowboy2Handler` this would amount to what we use as the first element of the return tuple to Cowboy)

Second, it's implicitly saying 'you should communicate with `handler` via the specified API' (to be concrete, in the context of `Cowboy2Handler` this would amount to what functions it calls on handler in the body of `websocket_` callbacks from Cowboy). This is a concern that needs to be communicated back to the adapter (and it's obviously something that the adapter needs to support). 

I think we'd need to either have an API like `Plug.Conn.adapter_upgrade(conn, :websockets, Sock.WebSocket, {handler, arg, opts})`, or else `Plug.Conn.adapter_upgrade(conn, Sock.WebSocket, {handler, arg, opts})` (I *think* I have a preference for the second option). Basically, we need to tell the adapter 'upgrade to this protocol, and communicate with `handler` using this API'. 

Of course, an adapter that doesn't support a given upgrade type / API is free to return an error tuple.

josevalim:
The idea is that the third argument is defined by the adapter. It can mean anything. You have to consult the adapter documentation.

Sock can then provide a high-level upgrade per adapter later.

mtrudel:
I see your intention. I'm not entirely convinced that it won't be a bit fragile, but let's wait until we're there and see how it shakes out. I'll be happy to be proven wrong.

I'll get to work on the work as discussed above. I'll be moving a bit slower as I'm back to my real job (company hack week last week, which is why I got so much done :). Expect a WiP in the next several weeks.

In the meantime I'll close off this PR. Thanks once again for a most fruitful discussion.

josevalim:
Perfect. Please ping me in the new PR!

PR #4976, opened 2022-09-24
Fix: Verified route in mix phx.gen.auth confirmation_edit.html.heex

icrayix:
For reference
https://github.com/phoenixframework/phoenix/blob/64ab5740a4c69337590a3c859ed52ac0744741a4/priv/templates/phx.gen.auth/routes.ex#L52-L53


chrismccord:
❤️❤️❤️🐥🔥

PR #4977, opened 2022-09-25
Update select tag generator with new components syntax

shamanime:
This fixes two issues when generating HTML for a schema with a select tag.

The first is the slot within a `for` comprehension.

The other is the missing attribute `value` for the `:option` slot:
```
warning: undefined attribute "value" in slot "option" for component PhxBlogWeb.Components.input/1
  lib/phx_blog_web/templates/post/edit.html.heex:17: (file)
```

### DIFF
--- a/lib/mix/tasks/phx.gen.html.ex
+++ b/lib/mix/tasks/phx.gen.html.ex
@@ -195,9 +195,9 @@ defmodule Mix.Tasks.Phx.Gen.Html do
         ~s"""
         <.input field={{f, #{inspect(key)}}} type="select" label="#{label(key)}">
           <:option>Choose a value</:option>
-          <%= for value <- Ecto.Enum.values(#{inspect(schema.module)}, #{inspect(key)}) do %>
-            <:option value={value}><%= value %></:option>
-          <% end %>
+          <:option :for={value <- Ecto.Enum.values(#{inspect(schema.module)}, #{inspect(key)})} value={value}>
+            <%= value %>
+          </:option>
         </.input>
         """
       {key, _}  ->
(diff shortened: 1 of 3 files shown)


leandrocp:
For reference, that previous code would fail at phx.gen tests like: https://github.com/phoenixframework/phoenix/blob/64ab5740a4c69337590a3c859ed52ac0744741a4/integration_test/test/code_generation/app_with_defaults_test.exs#L30

I'll get https://github.com/phoenixframework/phoenix/pull/4820 ready for review soon so we can test all templates.

lib/mix/tasks/phx.gen.html.ex
```diff
@@ -195,9 +195,9 @@ defmodule Mix.Tasks.Phx.Gen.Html do
         ~s"""
         <.input field={{f, #{inspect(key)}}} type="select" label="#{label(key)}">
           <:option>Choose a value</:option>
-          <%= for value <- Ecto.Enum.values(#{inspect(schema.module)}, #{inspect(key)}) do %>
```
chrismccord, line 198: We want the for comprehension here to build a template with an individual `<:option>...</:option>` per value. Can you make the change?

shamanime: @chrismccord could you elaborate? There's a `:for` on the new code and we get the expected individual option per value.

chrismccord: ah sorry, I thought this was the .exs template, where we want invidiual `<:option>`'s, but I see the sis the injected body of the enum

chrismccord:
❤️❤️❤️🐥🔥

PR #4981, opened 2022-09-26
Use schema for atom names in auth templates.

jbosse:
I ran the generator with `mix phx.gen.auth Accounts Player players` and my tests were failing because the templates are hard coded to `:user`.

```
test register player renders errors for duplicated email (TankTurnTacticsWeb.PlayerRegistrationLiveTest)
     test/tank_turn_tactics_web/live/player_registration_live_test.exs:58
     ** (ArgumentError) could not find non-disabled input, select or textarea with name "player[email]" within:
```

The old form was using schema so it seems like the `:user` was an oversight in https://github.com/phoenixframework/phoenix/pull/4955/files#diff-2e85c1d43c43cc817baa18f588d767a32cda37aa39425e44aa6e488237d147e1

I though about adding a test for a mix call with different names but I figure if user still works then the code is correct. I smoke tested it in my app and it worked.

### DIFF
--- a/priv/templates/phx.gen.auth/confirmation_live_test.exs
+++ b/priv/templates/phx.gen.auth/confirmation_live_test.exs
@@ -33,7 +33,7 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
 
       assert {:ok, conn} = result
       assert Phoenix.Flash.get(conn.assigns.flash, :info) =~
-        "#{<%= inspect schema.alias %>} confirmed successfully"
+        "<%= inspect schema.alias %> confirmed successfully"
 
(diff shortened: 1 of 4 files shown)


priv/templates/phx.gen.auth/confirmation_live_test.exs
```diff
@@ -33,7 +33,7 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
 
       assert {:ok, conn} = result
       assert Phoenix.Flash.get(conn.assigns.flash, :info) =~
-        "User confirmed successfully"
+        "<%= inspect schema.alias %> confirmed successfully"
```
jbosse, line 36: It feels like this should match what the developer is naming the "user" as the UI probably matches their domain.

chrismccord:
❤️❤️❤️🐥🔥

PR #4987, opened 2022-09-27
Add more input types

x4lldux:
Allowed types for `input` element were missing most of supported types which even generated warnings in templates generated with `phx.gen.html`.


chrismccord:
❤️❤️❤️🐥🔥

PR #4988, opened 2022-09-27
Fix bug in sigil_p query encoding

mveytsman:
Before this PR `~p"/foo?bar=#{struct}"` would give us `"/foo?bar==123"` because structs would be interpreted as maps of param values.


chrismccord:
Great catch! ❤️❤️❤️🐥🔥

PR #4990, opened 2022-09-28
Fetch docker release image from API

chrismccord:


priv/templates/phx.gen.release/Dockerfile.eex
```diff
@@ -8,13 +8,13 @@
 # This file is based on these images:
 #
 #   - https://hub.docker.com/r/hexpm/elixir/tags - for the build image
-#   - https://hub.docker.com/_/debian?tab=tags&page=1&name=bullseye-20220801-slim - for the release image
+#   - https://hub.docker.com/_/debian?tab=tags&page=1&name=<%= debian %>-<%= debian_vsn %>-slim - for the release image
 #   - https://pkgs.org/ - resource for finding needed packages
-#   - Ex: hexpm/elixir:<%= elixir_vsn %>-erlang-<%= otp_vsn %>-debian-bullseye-20220801-slim
+#   - Ex: hexpm/elixir:<%= elixir_vsn %>-erlang-<%= otp_vsn %>-debian-<%= debian %>-<%= debian_vsn %>-slim!
```
chrismccord, line 13:
```suggestion
#   - Ex: hexpm/elixir:<%= elixir_vsn %>-erlang-<%= otp_vsn %>-debian-<%= debian %>-<%= debian_vsn %>-slim
```

mix.exs
```diff
@@ -93,7 +95,8 @@ defmodule Phoenix.MixProject do
       {:mint_web_socket, "~> 1.0.0", only: :test},
 
       # Dev dependencies
-      {:esbuild, "~> 0.5", only: :dev}
+      {:esbuild, "~> 0.5", only: :dev},
+      {:castore, ">= 0.0.0", only: [:dev, :test]}
```
josevalim, line 99: We will either need it as a dependency OR add it to the generated mix.exs files. I will add it as a dep. We can drop it on OTP 25+ as Erlang has its own certs now.

chrismccord:
❤️❤️❤️🐥🔥

PR #4991, opened 2022-09-29
Add "for" to input global includes

clarkware:
The `phx.gen.auth` generator currently generates a "for" attribute for the `input` component:

```
<.input
        field={{f, :current_password}}
        name="current_password"
        type="password"
        label="Confirm new password"
        for="current_password_for_password"
        id="current_password_for_password"
        value={@current_password}
        required
      />
```

Adding `for` to the list of global includes silences the warning about the "for" attribute being undefined.


Eiji7 (approved):

chrismccord:
`for` isn't a valid input attribute so the fix was to remove it :) Thanks for taking a look at this!

clarkware:
Ah, of course! 😀

PR #4992, opened 2022-09-30
Add test for static paths

mveytsman:
This is a just an extra passing test - I thought I found a bug here but I was confused -- having an explicit test here might be useful for the future.


chrismccord:
❤️❤️❤️🐥🔥

PR #4993, opened 2022-09-30
Add get without trailing bang to phx.gen.context

nikitalocalhost:
Add get_$SCHEMA/1 to mix task phx.gen.context (there was only get_$SCHEMA!/1)

Signed-off-by: Nikita Localhost <nikita.localhost@gmail.com>


chrismccord:
You are free to implement this in your own app as needed, and it's absolutely fine if you want a non raising version. Thanks!

PR #4999, opened 2022-10-03
mix phx.digest to return non zero error code on failure

youroff:
Currently `mix phx.digest` exits with 0 on error which makes it harder to detect that something went wrong in CI/CD. This PR adds exit with 1 in such case.


chrismccord:
❤️❤️❤️🐥🔥

PR #5000, opened 2022-10-03
Fix select input losing the selected value when validating other fields

shamanime:
This addresses 2 issues:

1) The missing selected attribute would reset the HTML after validating because of phx-update.
2) The missing `value=""` for the initial option would send the value as "Choose a value" triggering "is invalid" error.


lib/mix/tasks/phx.gen.html.ex
```diff
@@ -171,7 +171,7 @@ defmodule Mix.Tasks.Phx.Gen.Html do
       {key, :decimal} ->
         ~s(<.input field={{f, #{inspect(key)}}} type="number" label="#{label(key)}" step="any" />)
       {key, :boolean} ->
-        ~s(<.input field={{f, #{inspect(key)}}} type="checkbox" label="#{label(key)}" />)
+        ~s|<.input field={{f, #{inspect(key)}}} type="checkbox" label="#{label(key)}" checked={Ecto.Changeset.get_field(@changeset, #{inspect(key)})} />|
```
leandrocp, line 174:
```suggestion
        ~s"""
        <.input
          field={{f, #{inspect(key)}}}
          type="checkbox"
          label="#{label(key)}"
          checked={Ecto.Changeset.get_field(@changeset, #{inspect(key)})}
        />
        """
```

priv/templates/phx.gen.live/components.ex
```diff
@@ -270,11 +271,19 @@ defmodule <%= @web_namespace %>.Components do
   def input(%{type: "checkbox"} = assigns) do
     ~H"""
     <label phx-feedback-for={@name} class="flex items-center gap-4 text-sm leading-6 text-zinc-600">
+      <input
+        type="hidden"
+        id={@id || @name}
+        name={@name}
+        value="false"
+      />
```
leandrocp, line 279:
```suggestion
      <input type="hidden" id={@id || @name} name={@name} value="false" />
```

leandrocp (approved):
Tested locally and it looks good 👍🏻 
Just a few suggestions to format the tags as expected by mix format.

installer/templates/phx_web/components.ex
```diff
@@ -270,11 +271,14 @@ defmodule <%= @web_namespace %>.Components do
...
     ~H"""
     <label phx-feedback-for={@name} class="flex items-center gap-4 text-sm leading-6 text-zinc-600">
+      <input type="hidden" id={@id || @name} name={@name} value="false" />
       <input
         type="checkbox"
         id={@id || @name}
         name={@name}
+        value="true"
         class="rounded border-zinc-300 text-zinc-900 focus:ring-zinc-900"
+        checked={@rest[:checked]}
```
elliottneilclark, line 281: Since we know the possible values for a check box input why not directly use the value for checked attribute?

```
checked={@value || @value =="true"}
```

elliottneilclark (commented):
I tested the select solution and it's great. I had one question about checkbox.

installer/templates/phx_web/components.ex
shamanime: This would indeed work for a boolean field but sometimes custom logic is required to mark the value as checked, using the global @rest lets the developer do it.

I'm upgrading our prod app to LV 0.18 and picking some new stuff from Phoenix as well and select/checkboxes/radios are some of the most annoying fields to work with in some scenarios. So far this is what worked for us.

elliottneilclark: Then I would suggest that it's worth the complexity to have the default be computed. So something like:

```elixir
  defp input_checked(%{checked: checked}, _value) when not is_nil(checked), do: checked
  defp input_checked(_rest, value) when is_boolean(value), do: value
  defp input_checked(_rest, value), do: to_string(value) == "true"

  defp option_selected(%{selected: selected}, _value), do: selected
  defp option_selected(%{value: option_value}, value), do: to_string(option_value) == value


  def input(%{type: "checkbox"} = assigns) do
    ~H"""
      <input type="hidden" id={@id || @name} name={@name} value="false" />
      <input
        type="checkbox"
        id={@id || @name}
        name={@name}
        class="checkbox checkbox-secondary"
        value="true"
        checked={input_checked(@rest, @value)}
      />
      <%= @label %>
    """
  end

  def input(%{type: "select"} = assigns) do
    ~H"""
      <.label for={@id}><%= @label %></.label>
      <select id={@id} name={@name} autocomplete={@name} {@rest}>
        <option
          :for={opt <- @option}
          {assigns_to_attributes(opt, [:selected])}
          selected={option_selected(opt, @value)}
        >
          <%= render_slot(opt) %>
        </option>
      </select>
    """
  end
```

That lets most component users get the correct defaults computed, while people that need to set a value are able to override pretty easily.

shamanime: This looks promising, I may be able to test this snipped with our app this week.

shamanime: Nice! I was able to use this and add a couple extra stuff while testing in our app. I'll update the PR early next week.

priv/templates/phx.gen.live/components.ex
```diff
@@ -270,11 +271,14 @@ defmodule <%= @web_namespace %>.Components do
   def input(%{type: "checkbox"} = assigns) do
     ~H"""
     <label phx-feedback-for={@name} class="flex items-center gap-4 text-sm leading-6 text-zinc-600">
+      <input type="hidden" id={@id || @name} name={@name} value="false" />
```
monooso, line 274: This will result in a "duplicate ID" error in the browser. Assuming we even need an ID on the hidden field, maybe do something like this?

```elixir
<input type="hidden" id="#{@id || @name}_hidden" name={@name} value="false" />
```

shamanime: Good one. I think it doesn't even need the id. I'll test tomorrow when I push more changes.

PR #5005, opened 2022-10-05
Fix method in edit settings form

ayrat555:


chrismccord:
I don't believe this is necessary any more. Thanks!

PR #5006, opened 2022-10-05
Remove checks for Elixir v1.13

leandrocp:
- ~Leverage a common fn to resolve min version~
- Update installer templates
- Update docs

Ref https://github.com/phoenixframework/phoenix/pull/4820#issuecomment-1267375088

/cc @josevalim


installer/mix.exs
```diff
@@ -11,13 +11,15 @@ defmodule Phx.New.MixProject do
...
-  #   1. the mix.exs generated by the installer
-  #   2. guides/introduction/installation.md
-  #   3. test/test_helper.exs at the root
-  #   4. installer/lib/phx.new.ex
+  #   1. guides/introduction/installation.md
+  #   2. guides/deployment/releases.md
   #
   @elixir_requirement "~> 1.14"
 
+  def elixir_requirement do
```
leandrocp, line 19: I can revert this change if you prefer to keep static values on templates.

lib/phoenix/verified_routes.ex
```diff
@@ -139,7 +139,7 @@ defmodule Phoenix.VerifiedRoutes do
     Module.put_attribute(mod, :phoenix_verified_statics, statics)
   end
 
-  @after_verify_supported Version.match?(System.version(), ">= 1.14.0-dev")
```
leandrocp, line 142: This line was added before the v1.14 release so I believe it's about time to update it.

installer/lib/mix/tasks/phx.new.ex
```diff
@@ -337,8 +337,10 @@ defmodule Mix.Tasks.Phx.New do
   end
 
   defp elixir_version_check! do
-    unless Version.match?(System.version(), "~> 1.14") do
-      Mix.raise "Phoenix v#{@version} requires at least Elixir v1.14.\n " <>
+    [_, min_version] = Phx.New.MixProject.elixir_requirement() |> String.split()
+
+    unless Version.match?(System.version(), Phx.New.MixProject.elixir_requirement()) do
```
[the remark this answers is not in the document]
leandrocp: It doesn't work :)
Reverted.

PR #5008, opened 2022-10-06
FEAT: Allow filtering of socket messages

Zarathustra2:
Hey there,

This is a draft for adding support to filtering messages based on user/socket params as briefly described here: https://groups.google.com/g/phoenix-core/c/kBAQai5y0jo

The motivation is simply that intercept is not suitable as a filter approach as then the user will encode the same payload n times. 

Filtering can be very useful, when you are sending lots of payloads to clients and each client has or can have filters which determine whether a payload should go out to the user or not.

Let me know what you think about it!

Best regards,
Dario


Zarathustra2:
ups sorry @chrismccord , then I will wait until someone replies on the mailing thread that I started, mb.

Issue #4997, opened 2022-10-03
`nil` query params showing in URLs generated by route helpers


elvismelkic:
### Environment

* Elixir version (elixir -v): 1.14.0 (compiled with Erlang/OTP 25)
* Phoenix version (mix deps): 1.6.12
* Operating system: Linux Ubuntu

### Expected behavior

I have this snippet in `router.ex`:
```ex
scope "/orgs/:organisation_id", BackendWeb do
  pipe_through([:browser, :require_authenticated_user])
   
  ...
  live("/holidays/pending", HolidayLive.Pending, :index)
  ...
end
```

When I call
```ex
Routes.holiday_pending_path(socket, :index, org_id,
  page_number: page_number,
  year: nil
)
```
the returned path with query params should look like `/orgs/{org_id}/holidays/pending?page_number=1`.

### Actual behavior

The returned path with query params looks like `/orgs/{org_id}/holidays/pending?page_number=1&year=`.

What's strange is that when I pass an empty list instead of `nil` as `year` param, so
```ex
Routes.holiday_pending_path(socket, :index, org_id,
  page_number: page_number,
  year: []
)
```
the returned path with query params doesn't contain `year` query param, so it looks like  
`/orgs/{org_id}/holidays/pending?page_number=1`.

I believe that passing `nil` query param to route helpers shouldn't add those params to created URL, or at least the behavior for passing `nil` and `[]` should be consistent.

(This is my first time opening an issue, I hope I explained everything well. If not, please let me know, so that I can improve it. :slightly_smiling_face: )

geeksilva97:
I can take a look.

geeksilva97:
Still investigating it. Found out that it's related to the `lib/phoenix/param.ex` file. Today I'm gonna do some experiments on that.

josevalim:
I would say this is intentional. A list of strings `["foo", "bar", "baz"]` is rendered as:

    ?list[]=foo&list[]=bar&list[]=baz

So it makes sense that nothing is rendered for empty list.

`nil` means there is a key but not value. You can remove nil yourself if desired.

PR #5009, opened 2022-10-07
Handle query mounting for empty lists

geeksilva97:
This PR fixes issue #4997 by making the empty lists have the same behavior as `nil`.


josevalim:
Thanks for the PR but I believe the behaviour is correct.

PR #5013, opened 2022-10-09
Update components.ex

LostKobrakai:
Rounded corners should be set relative to the row size not relative to what is in certain cells.

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -447,23 +447,25 @@ defmodule <%= @web_namespace %>.CoreComponents do
           <tr
             :for={row <- @rows}
             id={"#{@id}-#{Phoenix.Param.to_param(row)}"}
-            class="group hover:bg-zinc-50"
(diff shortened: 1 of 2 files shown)


chrismccord:
@LostKobrakai I have to revert this as it breaks safari and chrome row styling. Safari and chrome both lose the rounded bg padding/corners, and chrome makes the first row taller than the rest. Firefox looks perfect. If you have time, let me know if you can revisit, but I'll be reverting for now. Thanks!

LostKobrakai:
Changing one broken version to the other 🫣, but I see you. Cross browser table styling is 🤬. At least for safari the issue seems to be that it doesn't like `position: relative` on `tr`, which might mean knowing which `td` is the last one per row, which for the component depends on if there are actions and how many. I'll need to investige that a bit more.

jdmarshall:
This rounded hover effect is making it very difficult to do other styling of tables.

Why does this feature exist? What is it accomplishing except making tables look very odd when hovering? What is the goal?

PR #5020, opened 2022-10-13
Set missing `lib_web_name` in `phx.new.web`

moogle19:
When generating a new `web_app` via `mix phx.gen.web` the `lib_web_name` was missing, resulting in invalid paths e.g. for the `root` in `use Phoenix.View`


chrismccord:
There are some major changes to this file that will be merged soon, so let's double back after the merge. Thanks!

Issue #5011, opened 2022-10-07
Duplicate flash messages in root/app templates


shamanime:
### Environment

* Elixir version (elixir -v): 1.14
* Phoenix version (mix deps): from master
* Operating system: Linux

### Expected behavior

No duplicate flash messages in root and app templates.

### Actual behavior

The browser is complaining about duplicate IDs "#disconnected" on the page and I found out this issue.

The commit https://github.com/phoenixframework/phoenix/pull/4955/files#diff-2076b17871608fced1ef34220b23da5c55a927e475ddd04a9d74cb87ba119a1f added `plug :put_layout, false when action in [:index]` to PageController so we can get a nice landing page. This led to the duplicate flash messages being inserted on https://github.com/phoenixframework/phoenix/pull/4820/files#diff-e34c19dd158377da0ec2b93b7217bd54bcc8cc2e99d17cb8264a1abd882bb600 to get the PR working.

This is a leftover that needs to be fixed.

I'll inspect the issue and open a PR.

mindreframer:
I hit the same error 3 days ago.

Eiji7:
That's probably because both `root.html.heex` and `app.html.heex` renders `flash` massages.

jeregrine:
Oh man I searched the PRs but I didn't go through issues. Sorry!

PR #5022, opened 2022-10-14
Remove extra flash

jeregrine:
The app and the root both have the same flash component. 

https://github.com/phoenixframework/phoenix/blob/master/installer/templates/phx_web/components/layouts/root.html.heex#L15-L27 

And

https://github.com/phoenixframework/phoenix/blob/master/installer/templates/phx_web/components/layouts/app.html.heex#L40-L52

I am removing from the root.

### DIFF
--- a/installer/templates/phx_web/components/layouts/root.html.heex
+++ b/installer/templates/phx_web/components/layouts/root.html.heex
@@ -12,19 +12,6 @@
     </script>
   </head>
   <body class="bg-white antialiased">
-    <.flash kind={:info} title="Success!" flash={@flash} />
-    <.flash kind={:error} title="Error!" flash={@flash} />
-    <.flash
-      id="disconnected"
-      kind={:error}
-      title="We can't find the internet"
-      close={false}
-      autoshow={false}
-      phx-disconnected={show("#disconnected")}
-      phx-connected={hide("#disconnected")}
-    >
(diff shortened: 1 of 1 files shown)


chrismccord:
❤️❤️❤️🐥🔥

shamanime:
I noticed this and opened https://github.com/phoenixframework/phoenix/issues/5011.

The only issue is that the phx.gen.auth generators redirects to "/" with a flash message when it fails and since it doesn't use the app layout now it won´t displays the flash.

Thoughts @chrismccord?

Perhaps the flash could be moved to the template instead for this case?

leandrocp:
For eg this flash won't be displayed:

https://github.com/phoenixframework/phoenix/blob/16c3e9e818be9122ea6b666f6beecf11c4548fd1/priv/templates/phx.gen.auth/reset_password_controller_test.exs#L118-L119

Due to https://github.com/phoenixframework/phoenix/blob/16c3e9e818be9122ea6b666f6beecf11c4548fd1/installer/templates/phx_web/controllers/page_controller.ex#L4

chrismccord:
We can put flash in the `home.html.heex` template or leave it out. The landing page in the app is considered "throw away" as you necessarily replace it in your own application. The controller does `plug :put_layout, false when action in [:home]`

PR #5023, opened 2022-10-14
Milligram css is not required in the 1.7 version

cvkmohan:
the milligram css file is not needed anymore. updated the templates to exclude the `phoenix.css`


josevalim:
I _think_ we can clean up the app.css too. :)

chrismccord:
❤️❤️❤️🐥🔥

PR #5024, opened 2022-10-14
Handle websocket close events differently

jeregrine:
Right now today when we "disconnect" a websocket we simply use stop reply and let cowboy handle the rest, where as cowboy has a [reply shutdown](https://ninenines.eu/docs/en/cowboy/2.6/guide/ws_handlers/#_closing_the_connection) that allows you to send the status code and a message. 

I chose 3000 as it is currently registered with the [iana as "unauthorized"](https://www.iana.org/assignments/websocket/websocket.xml#close-code-number) the entire 3000-3999 status code space is left open for libraries and frameworks to register. We could claim 3001 for "kicked" or something for example.

This solves #4924 as we can simply check for error code 3000 in the ws onConnClose case.

The second issue is that when a user navigates away from the page SOME browsers will initate a websocket close event with status 1001 (going away) and isTrusted set to true. This was causing the connection status flash in liveview to flicker if you navigated away using a standard a tag href. 

Work left to do:
- [x] Test in Firefox
- [x] Test in chrome 
- [x] Test in Safari
- [x] Test in mobile Safari
- [ ] Test in various android browsers?


chrismccord:
@jeregrine we can revisit this when you have some time now that the bandit adapter work has settled. Note that the flash issue was solved in LV so we don't necessarily need that part of the fix here.

mtrudel:
I've got some additive changes to websock that I'm hoping to find time for in the next few weeks (see https://github.com/phoenixframework/websock_adapter/issues/2) which would help somewhat with this (at the very least, it should be coordinated)

mtrudel:
I just pushed some new proposed functionality up to address https://github.com/phoenixframework/websock_adapter/issues/2. May be of interest here as well.

jeregrine:
@chrismccord I agree we don't need most of this code ATM but I do think we can use close codes to handle the "user is kicked" scenario more cleanly.

PR #5026, opened 2022-10-15
Replace "setup tags" with "setup context"

eksperimental:
As this is the current terminology used in Elixir. Tags is reserved for the `@tag` attribute.


josevalim:
I think in those examples we mostly use it for the tags, so I would keep the naming for now. :)

Issue #5021, opened 2022-10-14
Phoenix.ConnTest.redirected_params/1 raises an error when the HTTP response code is not a 302 (like a 303)


Tricote:
### Environment

* Elixir version (elixir -v): Elixir 1.13.4 (compiled with Erlang/OTP 24)
* Phoenix version (mix deps): 1.6.11
* Operating system: Ubuntu 22.04.1 LTS 

### Context

I'm using [Turbo](https://turbo.hotwired.dev/) in my Phoenix application, and it requires to redirect with the HTTP status code [303 See Other](https://developer.mozilla.org/fr/docs/Web/HTTP/Status/303) in order to follow the redirection.

cf: https://turbo.hotwired.dev/handbook/drive#redirecting-after-a-form-submission

> After a stateful request from a form submission, Turbo Drive expects the server to return an [HTTP 303 redirect response](https://en.wikipedia.org/wiki/HTTP_303), which it will then follow and use to navigate and update the page without reloading.

So in my controller I need to override the default `302` status using `put_status(conn, 303)` like this:

```elixir
conn
|> put_flash(:info, gettext("Your request was proceed"))
|> put_status(303)
|> redirect(to: Routes.object_path(conn, :show, object))
```

### Expected behavior

In my controller tests, I would like to use the `Phoenix.ConnTest.redirected_params/1` function to retrieve the path params. If I had a  `302` HTTP status, this would work for example: 

```elixir
assert %{id: id} = redirected_params(conn)
# so that I can do after: assert redirected_to(conn, 303) == Routes.object_path(conn, :show, id) 
```

And I expected that it would work whatever the HTTP redirect code: 301, 302, 303, 307, 308,...

### Actual behavior

But currently the `Phoenix.ConnTest.redirected_params/1` function raises because the status is not `302`: 

```
** (RuntimeError) expected redirection with status 302, got: 303
```

Actually it's `Phoenix.ConnTest.redirected_to/2` that raises because the function relies on `Phoenix.ConnTest.redirected_to/2` to get the `Location` of the redirection, and it expects by default a `302`, cf: https://github.com/phoenixframework/phoenix/blob/v1.6.14/lib/phoenix/test/conn_test.ex#L585

### Possible improvement ?

* Add an optional `status` param to `Phoenix.ConnTest.redirected_params/1` to pass the status code to `Phoenix.ConnTest.redirected_to/2` ?
[...]

PR #5027, opened 2022-10-16
Add status code to ConnTest redirected_params (#5021)

linomassarani:
Add status code to `Phoenix.ConnTest.redirected_params/2` so that can also be used when status code needs to be different from `302`. Started on (#5021).
#hacktoberfest


josevalim:

Issue #5003, opened 2022-10-03
Allow Phoenix to provide generic WebSocket support


mtrudel:
This issue serves as an omnibus overview of the work being undertaken to improve Phoenix's ability to be hosted by web servers other than Cowboy, and as a follow-on benefit, simplifying the WebSocket implementation in Phoenix. The goals of this workup are to:

* Allow Phoenix to provide WebSocket functionality via a Plug-mediated upgrade flow, using Plug's new `Plug.Conn.upgrade_adapter/3`
* Expose sockets via the `Sock` API
* Simplify the implementation of `Phoenix.Endpoint` by leveraging `Plug.Router`, now that everything is done via Plug
* Remove the necessity for `Phoenix.Endpoint.Cowboy2Handler`, now that everything is done via Plug & Sock

The work to enable this spans a number of PRs across the Plug, Plug.Cowboy & Bandit repos:

* elixir-plug/plug#1119 defines`Plug.Conn.upgrade_adapter/3`
* elixir-plug/plug_cowboy#88 adds support for WebSocket upgrades to Plug.Cowboy
* mtrudel/bandit#38 adds support for WebSocket upgrades to Bandit
* mtrudel/sock#2 refactors the `Sock` API to be drop-in compatible with `Phoenix.Socket.Transport`
* #5030 updates Phoenix to use all of the above

All of the above work is complete, testing green, and functional (with the exception of commits to update Mix dependencies as the various PRs above are merged). The primary concern now is staging PR reviews in the correct order. To that end, a proposal:

## Review & Merge Proposal

1. Plug folks to review & merge changes in Plug PR above and push a point release of Plug once merged
2. @mtrudel to update Plug.Cowboy & Bandit PRs to depend on this new released version
3. Plug.Cowboy folks to review & merge Plug.Cowboy PR above
3. ~@moogle19 to review Bandit PR above~ done
4. @mtrudel to update Phoenix PR to depend on this newly released Plug.Cowboy version
5. Phoenix folks to review Phoenix PR

mtrudel:
Update: this is done 'on the bench' and working exactly as expected. I'll need a week or so of evenings to get PRs cleaned up & ready for proper review, but it's looking really really good.

mtrudel:
This work is now done! There are a number of followup tasks which are small enough to be handled via standalone PRs:

* Landing WebSocket upgrade/handling support via Sock
* Moving `socket` declarations into Router
[...]

PR #5030, opened 2022-10-16
Implement WebSocket negotiation support based on Plug.Conn.upgrade_adapter/3

mtrudel:
The work here is the first step towards implementation of the Phoenix work outlined in #5003. It implements:

* Plug-mediated support for both WebSocket and long poll socket endpoints. WebSocket requests are upgraded via `Plug.Conn.upgrade_adapter/3`
* Removal of the HTTP aspects of `Phoenix.Endpoint.Cowboy2Handler` as they are no longer required

Note that this PR is not entirely complete yet; it will need to have its dependencies on Plug & Plug.Cowboy updated to releases which incorporate elixir-plug/plug#1119 and elixir-plug/plug_cowboy#88 respectively. CI obviously won't pass until this is done.


lib/phoenix/endpoint.ex
```diff
@@ -456,7 +456,8 @@ defmodule Phoenix.Endpoint do
 
   defp plug() do
     quote location: :keep do
-      use Plug.Builder, init_mode: Phoenix.plug_init_mode()
+      @before_compile {Phoenix.Endpoint, :__add_fallback_route__}
```
mtrudel, line 459: I'm not nuts about how I'm (ab)using Plug.Router throughout Endpoint. The main problem is effectively adding a fallback route that is both performant and doesn't use any private parts of Plug.Router's implementation. 

I'm thinking that a `continue_on_no_match` option to Plug.Router may be a good option, which would add a no-op `do_match` clause at the top of the `Plug.Router.__before_compile` macro. This would enable the general purpose use of Plug.Router matches in a non-comprehensive manner, allowing Plug.Router's plugs to be more effectively interleaved with other Plugs. Regardless, this isn't a blocking concern for the other PRs in https://github.com/phoenixframework/phoenix/issues/5003

lib/phoenix/endpoint.ex
```diff
@@ -475,6 +476,12 @@ defmodule Phoenix.Endpoint do
           ]
       end
 
+      plug :match
+      plug :dispatch_if_matched
+
+      def dispatch_if_matched(%Plug.Conn{} = conn, opts),
+        do: if(Map.has_key?(conn.private, :plug_route), do: dispatch(conn, opts), else: conn)
```
mtrudel, line 483: ibid.

lib/phoenix/endpoint.ex
```diff
@@ -657,68 +654,14 @@ defmodule Phoenix.Endpoint do
...
-                not match?("_" <> _, param),
-                do: {param, Macro.var(var, nil)}
-
-          quote do
-            params = %{unquote_splicing(params)}
-            %Plug.Conn{conn | path_params: params, params: params}
-          end
-        end
-
-    {conn_ast, path}
```
mtrudel, line 721: Now that we're 100% Plug, this is replaced with use of Plug.Router's `match` clauses directly in the `socket` DSL API below.

lib/phoenix/endpoint/cowboy2_handler.ex
```diff
@@ -1,183 +0,0 @@
-defmodule Phoenix.Endpoint.Cowboy2Handler do
```
mtrudel, line 1: The majority of this functionality has moved to Plug.Cowboy in https://github.com/elixir-plug/plug_cowboy/pull/88

The reload catch & retry is still pending before this PR is ready for proper review - it can easily land in Endpoint, I think.

Again, not a blocking concern for other work on https://github.com/phoenixframework/phoenix/issues/5003

lib/phoenix/socket.ex
```diff
@@ -190,6 +190,30 @@ defmodule Phoenix.Socket do
...
+  supervision tree associated to the socket.
+
+  It receives the socket options from the endpoint,
+  for example:
+
+      socket "/my_app", MyApp.Socket, shutdown: 5000
+
+  means `child_spec([shutdown: 5000])` will be invoked.
+  """
+  @callback child_spec(keyword) :: :supervisor.child_spec()
```
mtrudel, line 215: This is moved unchanged from `Phoenix.Socket.Transport`

lib/phoenix/socket/transport.ex
```diff
@@ -13,6 +13,11 @@ defmodule Phoenix.Socket.Transport do
   many conveniences that invokes the underlying socket API to make
   it easier to build custom transports.
 
+  Of note, this behaviour is a superset of the `Sock` behaviour, and
+  as such all implementations of `Phoenix.Socket.Transport` are also
+  implementations of `Sock`. It is not possible to note this in code
+  without raising warnings about conflicting implementations.
+
```
mtrudel, line 20: I'd love to know if there's a way to properly declare Sock conformance here

lib/phoenix/transports/websocket.ex
```diff
@@ -1,5 +1,9 @@
 defmodule Phoenix.Transports.WebSocket do
   @moduledoc false
+  @behaviour Plug
```
mtrudel, line 3: This is now a Plug that gets routed to from the websocket routes within `socket` declarations in Endpoint. It uses the `Plug.Conn.upgrade_adapter/3` function to signal upgrades back to the HTTP adapter.

lib/phoenix/endpoint.ex
```diff
@@ -909,13 +852,38 @@ defmodule Phoenix.Endpoint do
...
   """
   defmacro socket(path, module, opts \\ []) do
-    module = Macro.expand(module, %{__CALLER__ | function: {:__handler__, 2}})
+    module = Macro.expand(module, __CALLER__)
+    websocket = Keyword.get(opts, :websocket, true)
+    longpoll = Keyword.get(opts, :longpoll, false)
 
     quote do
+      unquote(socket_route(path, module, websocket, Phoenix.Transports.WebSocket))
+      unquote(socket_route(path, module, longpoll, Phoenix.Transports.LongPoll))
```
mtrudel, line 861: WebSocket and long polling are now exposed directly as plugs within Endpoint, matched via `Plug.Router.match/5` that we return directly from this macro

lib/phoenix/endpoint.ex
```diff
@@ -657,68 +654,14 @@ defmodule Phoenix.Endpoint do
 
       @doc false
       def __sockets__, do: unquote(Macro.escape(sockets))
-
-      @doc false
-      def __handler__(%{path_info: path} = conn, opts), do: do_handler(path, conn, opts)
-      unquote(dispatches)
-      defp do_handler(_path, conn, opts), do: {:plug, conn, __MODULE__, opts}
```
mtrudel, line 664: We don't need to worry about rolling up all the socket endpoints as special handlers any longer; they're now dispatched as Plugs directly within Endpoint's Plug pipeline (which is now based on Plug.Router and not Plug.Builder)

mtrudel:
Regarding the Code Reload retry that was previously in Cowboy2Handler, I've managed to reproduce it locally (it's a race, and a hard one to get): 

```
[error] GenServer #PID<0.910.0> terminating
** (UndefinedFunctionError) function BanditSocketsWeb.Endpoint.call/2 is undefined (module BanditSocketsWeb.Endpoint is not available)
    (bandit_sockets 0.1.0) BanditSocketsWeb.Endpoint.call(%Plug.Conn{adapter: {Bandit.HTTP1.Adapter, :...}, assigns: %{}, body_params: %Plug.Conn.Unfetched{aspect: :body_params}, cookies: %Plug.Conn.Unfetched{aspect: :cookies}, halted: false, host: "localhost", method: "GET", owner: #PID<0.910.0>, params: %Plug.Conn.Unfetched{aspect: :params}, path_info: [], path_params: %{}, port: 4000, private: %{}, query_params: %Plug.Conn.Unfetched{aspect: :query_params}, query_string: "", remote_ip: {127, 0, 0, 1}, req_cookies: %Plug.Conn.Unfetched{aspect: :cookies}, req_headers: [{"connection", "keep-alive"}, {"accept-encoding", "gzip, deflate"}, {"accept-language", "en-CA,en-US;q=0.9,en;q=0.8"}, {"user-agent", "Mozilla/5.0 (Macintosh; Intel Mac OS X 10_15_7) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/15.4 Safari/605.1.15"}, {"cookie", "_bandit_sockets_key=SFMyNTY.g3QAAAABbQAAAAtfY3NyZl90b2tlbm0AAAAYYXQ0bnQzNG5SMktsWE1jaTU1NEVEcXo0.64HYCShvl4RSXnT9VPKAJebkptTsxdx6nSfZggFt3vg; lsid=XcBoGMLmuvKhhtMZ"}, {"upgrade-insecure-requests", "1"}, {"accept", "text/html,application/xhtml+xml,application/xml;q=0.9,*/*;q=0.8"}, {"host", "localhost:4000"}], request_path: "/", resp_body: nil, resp_cookies: %{}, resp_headers: [{"cache-control", "max-age=0, private, must-revalidate"}], scheme: :http, script_name: [], secret_key_base: nil, state: :unset, status: nil}, [])
    (bandit 0.5.6) lib/bandit/http1/handler.ex:56: Bandit.HTTP1.Handler.call_plug/2
    (bandit 0.5.6) lib/bandit/http1/handler.ex:15: Bandit.HTTP1.Handler.handle_data/3
    (bandit 0.5.6) lib/bandit/delegating_handler.ex:18: Bandit.DelegatingHandler.handle_data/3
    (bandit 0.5.6) lib/thousand_island/handler.ex:326: Bandit.DelegatingHandler.handle_continue/2
    (stdlib 4.1) gen_server.erl:1123: :gen_server.try_dispatch/4
    (stdlib 4.1) gen_server.erl:865: :gen_server.loop/7
    (stdlib 4.1) proc_lib.erl:240: :proc_lib.init_p_do_apply/3
Last message: {:continue, :handle_connection}
State: {%ThousandIsland.Socket{socket: #Port<0.23>, transport_module: ThousandIsland.Transports.TCP, connection_id: "AA725AD2C3D5", acceptor_id: "A7646B3A724C", read_timeout: 15000}, %{handler_module: Bandit.InitialHandler, plug: {BanditSocketsWeb.Endpoint, []}}}
```

as expected, the top frame in the stack is now the call to the endpoint's `call/2` function (since it's the same underlying issue of the endpoint itself not being defined). The thing to note here is that we can't fix this by moving the retry into Endpoint, since the error is that Endpoint itself isn't defined. Wherever this code is, it'll need to be in a module that never gets dynamically recompiled.

We can accomplish this by defining a 'ReloadPlug' in Phoenix that wraps the same retry logic as was in Cowboy2Handler, and have *that* be the main plug of the app as passed to Plug.Cowboy / Bandit. We'll likely want to make this conditional on whether or not code reloading is in place, since otherwise it's one extra plug in the stack that needn't be there.

lib/phoenix/endpoint/attempt_code_reload_plug.ex
```diff
@@ -0,0 +1,36 @@
...
+
+  def init({endpoint, opts}), do: {endpoint, endpoint.init(opts)}
+
+  def call(conn, {endpoint, opts}), do: do_call(conn, endpoint, opts, true)
+
+  defp do_call(conn, endpoint, opts, retry?) do
+    try do
+      endpoint.call(conn, opts)
+    rescue
+      exception in [UndefinedFunctionError] ->
```
mtrudel, line 24: Note we're rescuing against an `UndefinedFunctionError` rather than catching a throw as previously. Unsure where that implementation came from, but this captures the same set of errors.

mtrudel:
@josevalim as requested, this work is done and ready for review! See https://github.com/phoenixframework/phoenix/issues/5003 for an overview & merge proposal.

josevalim:
Thanks, I got the PRs. I am swamped with other work but I will review once I have time!

lib/phoenix/endpoint/attempt_code_reload_plug.ex
```diff
@@ -0,0 +1,36 @@
...
+  @moduledoc ~S"""
+  Wraps an Endpoint, attempting to sync with Phoenix's code reloader if 
+  an exception is raising which indicates that we may be in the middle of a reload.
+
+  We detect this by looking at the raised exception and seeing if it indicates
+  that the endpoint is not defined. This indicates that the code reloader may be 
+  mid way through a compile, and that we should attempt to retry the request
+  after the compile has completed. This is also why this must be implemented in
+  a separate module (one that is not recompiled in a typical code reload cycle),
+  since otherwise it may be the case that the endpoint itself is not defined.
```
josevalim, line 11: I wonder if we need this now?

Before we needed it because the handler would call the endpoint separately. But in this approach the handler no longer calls the endpoint, so it won't run into races.

In fact, the race we can run into is that we won't find the user handler (i.e. the UserSocket module), so we would need to wrap that instead.

lib/phoenix/endpoint/attempt_code_reload_plug.ex
```diff
@@ -0,0 +1,36 @@
+defmodule Phoenix.Endpoint.AttemptCodeReloadPlug do
```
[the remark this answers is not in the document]
josevalim: I wonder if we need this now?

Before we needed it because the handler would call the endpoint separately. But in this approach the handler no longer calls the endpoint, so it won't run into races.

In fact, the race we can run into is that we won't find the user handler (i.e. the UserSocket module), so we would need to wrap that instead.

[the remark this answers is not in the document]
josevalim: I guess we can try this:

1. Remove the protection from the old code, use your snippet above to reproduce the failure. Can you reproduce it?

2. If you can reproduce it, remove this protection and try again, can you reproduce it?

3. And then, try the snippet above changing `user_socket` and see if it happens in the new code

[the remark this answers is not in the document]
mtrudel: We do. I can actually reproduce in both Cowboy and Bandit without this. It's the same issue, though it now occurs at the boundary b/w the underlying server and Phoenix (ie: Cowboy is looking to call `App.Endpoint.call/2`, but `App.Endpoint` doesn't exist since it's being recompiled). 

The intent of this change is to ensure that the module the web server is calling never goes away due to recompilation. `Cowboy2Handler` played that role previously.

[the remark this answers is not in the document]
mtrudel: There's a similar race existing in the boundary between the underlying server and the socket implementation within Phoenix, but those aren't likely to get recompiled very often so it's less of a concern in practice.

josevalim:
Thank you @mtrudel!

I think this looks great but I would break this PR in two. The first PR is to use `upgrade_adapter` without using `Sock`, simply by calling the `handler` directly which will be our existing Cowboy implementation - without touching the bulk of the handler code (i.e. we will remove the call to `endpoint.__handler__` but leave the rest pretty much as is).

This in itself will already allow you to plug bandit in, in a transparent way. Then we can have another PR to discuss and move to Sock altogether. :)

lib/phoenix/endpoint.ex
```diff
@@ -657,68 +654,14 @@ defmodule Phoenix.Endpoint do
...
-            params = %{unquote_splicing(params)}
-            %Plug.Conn{conn | path_params: params, params: params}
-          end
-        end
-
-    {conn_ast, path}
+  @doc false
+  defmacro __add_fallback_route__(_) do
+    quote do
+      defp do_match(conn, _, _, _), do: conn
```
josevalim, line 663: This is private API and relying on implementation details of Plug.Router, we most likely shouldn't do this.

lib/phoenix/endpoint.ex
```diff
@@ -909,13 +852,38 @@ defmodule Phoenix.Endpoint do
 
   """
   defmacro socket(path, module, opts \\ []) do
-    module = Macro.expand(module, %{__CALLER__ | function: {:__handler__, 2}})
+    module = Macro.expand(module, __CALLER__)
```
josevalim, line 855: We need to keep the %{function : ...} bit to avoid compile-time dependencies. If we need to pick another name, it can be `call/2`, as it is now a regular Plug.

lib/phoenix/endpoint.ex
mtrudel: Yep. I was expecting this to raise some red flags. I float a proposal to deal with this above. See https://github.com/phoenixframework/phoenix/pull/5030#discussion_r996490316

lib/phoenix/endpoint.ex
mtrudel: This is past the limit of my macro brain :) You're saying basically
```suggestion
    module = Macro.expand(module, %{__CALLER__ | function: {:call, 2}})
```

?

lib/phoenix/endpoint.ex
josevalim: I am almost sure we are using a private API here too.

Perhaps it is best to ditch `Plug.Router` and build a mini router, which we already have in `__handler__`, except we will call `:socket_dispatch` here. The difference is that `socket_dispatch`  invokes the underlying plugs instead of returning metadata. 

Then all we need to do is to add a `plug :socket_dispatch` here.

mtrudel:
> Thank you @mtrudel!
> 
> I think this looks great but I would break this PR in two. The first PR is to use `upgrade_adapter` without using `Sock`, simply by calling the `handler` directly which will be our existing Cowboy implementation - without touching the bulk of the handler code (i.e. we will remove the call to `endpoint.__handler__` but leave the rest pretty much as is).

Very good!

lib/phoenix/transports/websocket.ex
```diff
@@ -13,36 +17,52 @@ defmodule Phoenix.Transports.WebSocket do
...
-  def connect(%{method: "GET"} = conn, endpoint, handler, opts) do
+  def init(opts), do: opts
+
+  def call(%{method: "GET"} = conn, {endpoint, handler, opts}) do
     conn
-    |> Plug.Conn.fetch_query_params()
+    |> fetch_query_params()
     |> Transport.code_reload(endpoint, opts)
     |> Transport.transport_log(opts[:transport_log])
     |> Transport.force_ssl(handler, endpoint, opts)
```
josevalim, line 27: I think we no longer need this because the endpoint will always call `force_ssl` before we go down this path?

mtrudel: It depends on where the `socket` clause ends up within the Endpoint, but yes you're correct.

lib/phoenix/endpoint.ex
josevalim: I dropped a comment here about how to implement it: https://github.com/phoenixframework/phoenix/pull/5030#discussion_r1008718222

TL;DR - convert `__handler__` to a dispatch function and plug it in. :)

lib/phoenix/endpoint.ex
josevalim: Yes! Or `:socket_dispatch` instead of `:call` if we go down that route. The name doesn't matter.

lib/phoenix/endpoint.ex
mtrudel: You're correct that this is also part of Plug.Router's private API, though it's part of the same problem that we have with the private `defp do_match` called out below.

The main reason I wanted to appeal to Plug.Router here is to get the benefit of its excellent `/path/:id` parameter parsing, which we'd need to replicate. 

I would think that adding in a `use Plug.Router, continue_on_no_match: true` style option would solve this quite handily, and also make Plug.Router useful within plug pipelines in a general case (this is something I've come up against before as well, in other 'pure Plug' projects).

mtrudel: Is there a reason that you're opposed to adding that to Plug.Router? To be clear, I'm happy to do the work for that (it's already done other than tests locally).

lib/phoenix/transports/websocket.ex
josevalim: In fact, I realize now we can mostly get rid of this module if we move the socket upgrades to happen much later in the Plug stack, which is now possible. But that's a separate conversation...

For now, let's focus on getting `upgrade_adapter` without using Sock, so we can ship this in v1.7 and unblock a good part of Bandit integration, and then we can explore how to evolve all of this later!

lib/phoenix/endpoint.ex
josevalim: 1. We already have the match logic implemented today
2. We won't use private APIs
3. We won't rock the boat too much which means...
4. We will get this merged sooner

:)

josevalim: In fact, I am even thinking we may even move the whole `socket` stuff to the router, so not rocking the boat too much sounds even better if the goal is to deprecate this eventually.

lib/phoenix/endpoint/attempt_code_reload_plug.ex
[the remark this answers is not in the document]
josevalim: Perfect!

lib/phoenix/endpoint/attempt_code_reload_plug.ex
```diff
@@ -0,0 +1,36 @@
...
+  @moduledoc ~S"""
```
josevalim, line 2: Please keep this as `@moduledoc false`, as it should be private, but keep those notes as code comments!

lib/phoenix/endpoint/attempt_code_reload_plug.ex
josevalim, line 1:
```suggestion
defmodule Phoenix.Endpoint.SyncCodeReloadPlug do
```

lib/phoenix/endpoint/attempt_code_reload_plug.ex
josevalim: Ah, you would need to use this from Bandit? Which then means it has to be public?

mtrudel: Yup!

lib/phoenix/endpoint/attempt_code_reload_plug.ex
mtrudel: Yup!

lib/phoenix/transports/websocket.ex
mtrudel: Sounds good! There are a couple of things above that I'm not sure I agree with just yet, but I think they'll be clearer if we can get all the other PRs in this stack done and dusted so we can focus on just this one more narrowly!

mtrudel:
Branch updated with a provisional version of the two-PR split @josevalim requested previously. It's functional, though some gaps remain:

* Endpoint is still using private `Plug.Router` details
* I'm chasing down a new intermittent race condition on websocket setup

However I wanted to push this up to demonstrate the overall structure. Specifically, note that:

* `Phoenix.Endpoint.Cowboy2Handler` is still around, but the scope is now narrowed to just providing `:cowboy_websoket` conformance. This will disappear into Sock in a subsequent PR
* The Cowboy-specific setup code in `Phoenix.Transports.WebSocket` makes this version still Cowboy-specific (again, this will disappear into Sock in a subsequent PR)

Is this what you'd envisioned, @josevalim ?

lib/phoenix/transports/websocket.ex
```diff
@@ -13,36 +17,67 @@ defmodule Phoenix.Transports.WebSocket do
...
 
         case handler.connect(config) do
-          {:ok, state} -> {:ok, conn, state}
-          :error -> {:error, Plug.Conn.send_resp(conn, 403, "")}
+          {:ok, state} ->
+            conn
+            |> upgrade_adapter(
+              :websocket,
+              {Phoenix.Endpoint.Cowboy2Handler, {handler, process_flags, state}, cowboy_opts}
+            )
```
mtrudel, line 67: This will all disappear into Sock in short order, though for now this keeps Phoenix as Cowboy only

lib/phoenix/transports/websocket.ex
```diff
@@ -13,36 +17,69 @@ defmodule Phoenix.Transports.WebSocket do
...
+          opts
+          |> Keyword.take([:fullsweep_after])
+          |> Map.new()
 
         case handler.connect(config) do
-          {:ok, state} -> {:ok, conn, state}
-          :error -> {:error, Plug.Conn.send_resp(conn, 403, "")}
+          {:ok, state} ->
+            # Cures a Cowboy race condition where it doesn't see our declared websocket_init/1
+            _ = Code.ensure_loaded?(Phoenix.Endpoint.Cowboy2Handler)
```
mtrudel, line 64: Interesting!

lib/phoenix/endpoint.ex
mtrudel: Done!

lib/phoenix/endpoint.ex
mtrudel: Done!

lib/phoenix/endpoint/cowboy2_adapter.ex
```diff
@@ -80,6 +80,9 @@ defmodule Phoenix.Endpoint.Cowboy2Adapter do
       Application.ensure_all_started(:ssl)
     end
 
+    # Cures a Cowboy race condition where it doesn't see our declared websocket_init/1
+    _ = Code.ensure_loaded?(Phoenix.Endpoint.Cowboy2Handler)
```
mtrudel, line 84: This was reliably buggy on the server's first WebSocket upgrade without this - switching the handler within Cowboy seems to cause a race with the new handler not being available right away, and so the new handler's (optional!) websocket_init/1 was not being called.

Since this isn't a module that never gets reloaded we're OK having this just load once at startup.

mtrudel:
mix deps updated for plug 1.14 and plug_cowboy 2.6.0. Should be good to go!

lib/phoenix/endpoint/cowboy2_adapter.ex
josevalim: I can see why this happens: we should fix it in Plug.Cowboy. We should call `handler.module_info(:module)` before passing it to the websocket upgrade. :)

josevalim: I have republished v2.6.0. We should be able to remove this line.

mtrudel:
Integration test failure looks to be unrelated

josevalim:
This PR looks great to me. I only need to figure out a way to make the websocket bits inside the WebSocket transport Pluggable. I will tidy it up and I should merge it tomorrow!

lib/phoenix/endpoint/cowboy2_adapter.ex
mtrudel: Confirmed that this fixes the issue. Branch updated

mtrudel: Just so I understand for next time, does that change fix it due to the side effect of ensuring the module is in fact loaded before calling `upgrade/5`?

mtrudel:
If you want to take the option, I've got a good deal of that stuff cleaned up in my 'part 2' PR (still pending). I'm expecting to have it out in the next day or so.

josevalim:
Thank you @mtrudel, this is huge! :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

josevalim:
Hi @mtrudel! I have pushed 08c6e61b9b382fbc4143c75458ae77b7b764d255, which should allow you to inject a upgrade so you can inject `sock` for now. I don't expect this to be available in the long term, it is temporary for experimentation, so make sure that you depend on Phoenix ~> 1.7.x.

I think we want to move the socket connections all the way down to the router. This way we get logging, session handling, and more for free. The same way we just got SSL handling for free. This also means LiveDashboard no longer has to ask users to plug stuff in the endpoint. The same for Oban and pretty much every LiveView lib out there. This will be a huge improvement thanks to your work!

Since there is some API work necessary to make this happen, that's better in the hands of @chrismccord! So I would expect this to happen only after v1.7 is out. :)

mtrudel:
AMAZING!

I see two streams of followup work coming from this:

First, I'll get to work on implementing generic upgrade / handler support via Sock as we've talked about above. I suspect that should be up for review within a week or so; I'll be dropping a few points for early discussion down below to make sure we're aligned on this. Do you see any reason that this couldn't make it into 1.7 assuming I get the work done soon?

Second, in terms of moving socket down into router, that's going to simplify things a ton, both within Phoenix and for users as you say. Given that this will change the user facing socket definition API, I agree that this should be tackled post-1.7.

josevalim:
>  Do you see any reason that this couldn't make it into 1.7 assuming I get the work done soon?

I can give feedback once the PR is up. I am worried about the second item affecting the API from the first item but maybe that won't be the case depending on how isolated it is.

mtrudel:
I don't think there's going to be any real overlap between the two items, but I agree, let's wait and see how the PR shakes out.

My main uncertainty as this point is whether the Sock API should subsume the 'Phoenix.Socket.Transport.connect/1` function, and call it as part of the upgrade process (that is, should Sock manage the 'take a conn and some config and turn it into initial socket state' flow or not)? 

The way it is now in `master`, the work to build initial socket state is spread between `Phoenix.Socket.connect/1`, `Phoenix.Socket.Transport.connect_info/3`, and `Phoenix.Transports.WebSocket.call/2`. Instead, if the responsibility for that moved into `Phoenix.Socket.Transport.connect` (whose signature/arity would need to change), it would allow all that disparate setup code to be consolidated. Essentially, the `Phoenix.Socket.Transport.connect` contract would be broadened to include evaluating the conn for any required values (I'm 90% sure this is common between WebSockets and long poll implementations).

This could always be accommodated later on by adding an optional step of this nature to the Sock lifecycle & moving some stuff in `WebSocket.call/2` around, so I don't think it's a blocking concern to getting Sock support in here. 

Any strong feelings one way or the other?

mtrudel:
Also, thanks again for being so amazing to work with. You really do make it look easy!

josevalim:
> My main uncertainty as this point is whether the Sock API should subsume the 'Phoenix.Socket.Transport.connect/1` function, and call it as part of the upgrade process (that is, should Sock manage the 'take a conn and some config and turn it into initial socket state' flow or not)?

Right now, I am thinking that the Sock contract would exactly be the `phoenix_websocket_upgrade` contract that we have in master right now.

I also really think we should have "Web" in the name of the project if it is going to be WebSocket specific. :) WebSocket, WebSock, WS are all fine by me. :)

mtrudel:
> Right now, I am thinking that the Sock contract would exactly be the `phoenix_websocket_upgrade` contract that we have in master right now.

Agree that that's a fine place to start. My concern is really more of 'future optimization' than anything fundamentally structural.

> I also really think we should have "Web" in the name of the project if it is going to be WebSocket specific. :) WebSocket, WebSock, WS are all fine by me. :)

Instead of 'Sock', you mean? The good names all mostly already taken in Hex, sadly. 

Though `WebSock` isn't! And it's a shoutout to a name I haven't heard for a long, long time! https://en.wikipedia.org/wiki/Trumpet_Winsock

mtrudel:
Getting this working on Bandit was a [breeze](https://github.com/mtrudel/bandit/pull/38/commits/652836d044fd2b2b912c52d0ae7d324294e47564)!

It's official: Bandit 0.5.7 (just released) will FULLY support Phoenix as of 1.7 (even if Sock support doesn't land)! Fantastic!

Thanks again for all your hard work on this @josevalim!

PR #4986, opened 2022-09-26
Support new controller options

chrismccord:
Ready for early review, but not ready for merge. Start of #4975

PR #5034, opened 2022-10-18
Replace view with html for umbrella generators

Gazler:
This seems to have been overlooked when replacing Phoenix.View in https://github.com/phoenixframework/phoenix/pull/4986


josevalim (approved):

chrismccord:
Fixed on master. Thanks!

PR #5035, opened 2022-10-18
Skip non-phoenix projects when running mix phx.digest

Gazler:
In an umbrella application, there can be non-phoenix applications that are not intended to have static assets. Previously these would warn that the path does not exist.

Commit f49acd74f47b7abcc9d011f9ec3156295ce3feb1 changed the exit code of a failed digest to 1. This means that an umbrella application that had any non-phoenix applications would exit. This could be a problem in a CI/CD pipeline.

Any non-phoenix applications are now skipped when running mix phx.digest


elliottneilclark:
This looks like it would totally fix my issues. However I also think the test failures are related.

PR #5042, opened 2022-10-19
Improve table component example

mveytsman:
The id attr is required, so we should include it in the example.


chrismccord:
❤️❤️❤️🐥🔥

PR #5045, opened 2022-10-20
add a hint to direct mix release users away from prod.exs

trashhalo:
There are so many examples online of calling System.get_env to pull production variables. Even 10 lines down in this file there is an example. I think we would do our new users, like me, a solid if you redirect them to runtime.exs

Don't feel strongly on wording, feel free to edit.


josevalim:
Thank you @trashhalo! I moved SSL to the runtime.exs and add a small note!

PR #5047, opened 2022-10-20
Fix redirected_params helper to forward status code

### LINKED ISSUE #5021: Phoenix.ConnTest.redirected_params/1 raises an error when the HTTP response code is not a 302 (like a 303)

Isaius:
This PR addresses and Closes #5021 highlighted issues.

### What was done
Now the helper function `redirected_params` to forward the status code received from the connection when calling the `redirected_to` helper which defaults to `302`, resulting in the error when a different value was received.

#hacktoberfest

### DIFF
--- a/lib/phoenix/test/conn_test.ex
+++ b/lib/phoenix/test/conn_test.ex
@@ -586,7 +586,7 @@ defmodule Phoenix.ConnTest do
   @spec redirected_params(Conn.t) :: map
   def redirected_params(%Plug.Conn{} = conn) do
     router = Phoenix.Controller.router_module(conn)
-    %URI{path: path, host: host} = conn |> redirected_to() |> URI.parse()
+    %URI{path: path, host: host} = conn |> redirected_to(conn.status) |> URI.parse()
 
     case Phoenix.Router.route_info(router, "GET", path, host || conn.host) do
       :error ->
--- a/test/phoenix/test/conn_test.exs
+++ b/test/phoenix/test/conn_test.exs
@@ -460,6 +460,18 @@ defmodule Phoenix.Test.ConnTest do
         |> redirected_params()
       end
     end
+    
+    test "with custom status code" do
+      Enum.each(300..308, fn status ->
+        conn =
+          build_conn(:get, "/")
+          |> RedirRouter.call(RedirRouter.init([]))
+          |> put_resp_header("location", "/posts/123")
+          |> send_resp(status, "foo")
+
+        assert redirected_params(conn) == %{id: "123"}
+      end)
+    end
   end
 
   describe "path_params/1" do


linomassarani:
I think this one is better than mine #5027!
👏🏻👏🏻👏🏻👏🏻👏🏻

josevalim:
Closing in favor of #5027.

josevalim:
I prefer #5027 because it makes explicit the status code. :)

PR #5048, opened 2022-10-21
Fix missing end when use --no-mailer alone

feng19:


josevalim:
Thank you @feng19! I wanted to unify those blocks for a while, so I pushed a commit that does so!

PR #5050, opened 2022-10-22
Fix authentication forms (LiveView variant)

Cervajz:
The `mix phx.gen.auth` generates broken forms in the LiveView variant. On a fresh project after generating the authentication code:

```shell
mix test

Compiling 13 files (.ex)
warning: undefined function input_value/2 (expected WhWeb.UserResetPasswordLive to define such a function or for it to be imported, but none are available)
  lib/wh_web/live/user_reset_password_live.ex:30

warning: undefined function input_value/2 (expected WhWeb.UserSettingsLive to define such a function or for it to be imported, but none are available)
  lib/wh_web/live/user_settings_live.ex:62

warning: undefined function input_value/2 (expected WhWeb.UserSettingsLive to define such a function or for it to be imported, but none are available)
  lib/wh_web/live/user_settings_live.ex:55
```

I believe the `value={input_value(f, :password)}` is not needed anymore as it is part of [core_components.ex's input function (component)](https://github.com/phoenixframework/phoenix/blob/master/priv/templates/phx.gen.live/core_components.ex#L266) ?


Cervajz:
Rebased against current master

chrismccord:
❤️❤️❤️🐥🔥

PR #5056, opened 2022-10-27
Restore web namespace name when generating core components

shamanime:
The commit https://github.com/phoenixframework/phoenix/commit/9f0b9eca871c9816cd33090177f77b82d152afe6 replaced the web namespace with a fixed name so the generated core component module would have the wrong namespace defined.


chrismccord:
❤️❤️❤️🐥🔥

PR #5057, opened 2022-10-27
Fix integration_test and format a few more templates

leandrocp:
This won't fix all issues as templates are WIP but it does allow integration_test to run again and does fix a few formatting issues.


chrismccord:
❤️❤️❤️🐥🔥

PR #5063, opened 2022-11-02
Websock support

mtrudel:
This is part two of #5030, adding in support for (newly renamed) WebSock facilitated upgrades. This makes Phoenix entirely agnostic to the underlying server, so long as it provides Plug & WebSock support.

Note that WebSock 0.4.0 (which this depends on) doesn't have full test coverage yet (though it works fine & shouldn't change as a result); I wanted to get this up for review in a timely manner.


lib/phoenix/socket/transport.ex
```diff
@@ -13,6 +13,11 @@ defmodule Phoenix.Socket.Transport do
   many conveniences that invokes the underlying socket API to make
   it easier to build custom transports.
 
+  Of note, this behaviour is a superset of the `WebSock` behaviour, and
+  as such all implementations of `Phoenix.Socket.Transport` are also
+  implementations of `WebSock`. It is not possible to note this in code
+  without raising warnings about conflicting implementations.
```
josevalim, line 19: Ah... I see now. Phoenix.Socket.Transport in Phoenix is generic. WebSock is an implementation of Phoenix.Socket.Transport for WebSockets + a generic abstraction for adapters. So now that I think about it, I don't think we should list `WebSock` here.

josevalim: I.e. we should just remove this whole paragraph and instead say that WebSock implements Phoenix.Socket.Transport.

josevalim:
Yeah, this is as good as it gets. :) I will talk to @chrismccord about his feelings on this and the ownership. Would you be ok with moving WebSock to the Phoenix org? You'd still remain as admin but we would be around to help/backup.

lib/phoenix/socket/transport.ex
mtrudel: > I.e. we should just remove this whole paragraph and instead say that WebSock implements Phoenix.Socket.Transport.

Do you mean that we should say that here, just in a differently worded comment, or that `WebSock` should actually `@impl` Phoenix.Socket.Transport over in the WebSock project?

mtrudel:
I have no problem with that! I've actually been thinking about the longer-term plans for the entire Thousand Island / Bandit / WebSock stack and how there should be an `elixir-networking` org to host all of that (a bit of an overlap with `elixir-plug`; maybe it would be a better home)?

In any case, having been bit firsthand by 'the bus effect' of maintainers just disappearing I'm more than happy to have stuff live in an org.

lib/phoenix/socket/transport.ex
josevalim: Right, the thing is that Phoenix.Socket.Transport is meant to be transport agnostic, so it doesn't make sense to say it implements WebSock.

At the same time, we don't want WebSock to depend on Phoenix, because people can use it without Phoenix.

So I think we should just informally says in WebSocket that it implements Phoenix.Socket.Transport and that's it. WDYT?

josevalim: This is also the reason why I think it should exist in the Phoenix org (I was originally considering it should be Plug but I think it is more aligned to Phoenix).

mtrudel: Happy to note it at both ends (in fact, I'm pretty sure I already do note the similarity in the WebSock docs, at least as a source of inspiration). My main worry (speaking from experience as I pieced together an understanding of all of Phoenix's internals) is that without some sort of note here, it's not at all obvious how a fundamental part of Phoenix's WebSocket implementation works. Maybe the note could instead live in Phoenix.Transport.WebSocket, as a code comment near the `WebSock.upgrade/4` call?

mtrudel: I actually think that's better, TBH. As I've mentioned before, someone trying to understand how WebSockets work in Phoenix is going to check out Phoenix.Transports.WebSocket first (just given the name), and so as many breadcrumbs as possible should be left there

mtrudel: As an aside, the ability for behaviours to compose other behaviours is something I've wished for on multiple occasions (a concrete example is ThousandIsland.Handler wanting to be a superset of GenServer). Has this been proposed before / are there plans for it / is it something you'd entertain in the language?

josevalim: > Maybe the note could instead live in Phoenix.Transport.WebSocket, as a code comment near the WebSock.upgrade/4 call?

Agreed!

> Has this been proposed before / are there plans for it / is it something you'd entertain a PR for in the language?

Nothing we can do. Behaviours are somewhat tied to Dialyzer and Dialyzer will complain anyway.

mtrudel:
Docs updated, merge conflicts resolved. 

Let's wait to hear @chrismccord 's take on the matter, but I'm happy to move the WebSock project to any org you'd like

mtrudel:
I'm running into an issue with Hex dependencies now that WebSock depends on Cowboy. I'd originally conceived of WebSock as being strictly behaviour, but now that it's an actual abstraction on _top_ of Cowboy & Bandit there's an inverted dependency that makes it so Bandit ends up having a transitive dependency on Cowboy (not great). 

To fix this I need Bandit to be able to depend on *just* the WebSock behaviour and not pull in any of the adapter dependencies. Similar to the split between plug and plug_cowboy, I think I'm going to end up cutting a `websock_adapter` package to contain the actual adapter (this is what Phoenix will end up depending on).

Let's hold off on merging this until I get that lined up. Should be later today (EDT).

mtrudel:
No - the whole intent of WebSock was to *be* that general WebSocket API (and in my mind, it still mostly is). If you look at WebSock's 'adapter' to Bandit, it just wires Bandit directly into the handler (see https://github.com/mtrudel/websock/blob/main/lib/websock.ex#L167-L168).

The discussion we had previously that ended up reshaping the WebSock API in the mold of Transport was more about the tradeoffs of seeing the API from the perspective of a server implementor (what it was originally) or the perspective of a framework implementor (what it is now). I still think it stands as a generalized WebSocket API.

Before I get too far into this, do you have a better idea?

mtrudel:
That's more or less the split that I have in mind, just that the base behaviour doesn't try to be generic (it's more specific than a generalized socket interface, since it knows about frame types). 

The split is actually turning out to be pretty straightforward, and enables everything I was hoping to in terms of declaring conformance and testability. Should be up shortly.

josevalim:
Hi @mtrudel! One question: is there a particular reason why you preferred to not go with `sock` for spec and `websock` for the implementations? Is it because the current contract has websock specific functionality?

mtrudel:
Yep - the `sock` -> `websock` rename was, to my mind, a literal rename. Nothing changed semantically, it was just a matter of renaming things. I never intended it to mean anything more.

I always conceived the WebSock (née Sock) API to be a WebSocket abstraction and nothing more (specifically, I never thought of it as a generalization over other transports such as long polling). This is in contrast to Transport, which intends to be generic over all transports. The WebSock spec is pretty specific to WebSockets (at least insofar as it knows about frame types and the overall lifecycle).

The tl;dr here is that I think this difference of opinion (if I can call it that) comes from the perspective of a server-first or framework-first perspective. From my perspective, I set out for WebSock to be the solution to 'how can I generalize different servers' WebSocket implementation?' whereas from a framework perspective may be 'how can I generalize 'sockets' across different transports?'. WebSock is not and never has been trying to be the latter of those two [1]. It's mostly a happy accident that WebSock and Transport have managed to line up so well.
 
The split between the WebSock behaviour and the WebSockAdapter adapter lib reflected here is purely a matter of necessity, to allow libs both 'above' and 'below' the adapter layer to be able to interact with the spec. 

[1] Having a library that generalized over multiple transports could be an interesting thing to explore in the future, despite path dependencies having brought us to where we are

mtrudel:
Updated to WebSock 0.4.3 which provides support for return values to specify multiple frames to send:

```
{:push, [{:text, "TEXT"}, {:binary, "BINARY"}], state}
```

Gazler:
Thanks for all the hard work @mtrudel I just tried out this PR on master and can confirm that cowboy and bandit are both working with a fresh liveview project, it was seamless!

I've reviewed this PR as well as `websock` and `websock_adapter` and it all looks good to me.

josevalim:
@mtrudel, can you please add me as admin to both projects so I transfer them to the Phoenix org and merge this?

mtrudel:
I'll have to add you as an admin to the hex projects too. Will do that later this morning!

mtrudel:
@josevalim I added you to the hex packages. In terms of the repos, it [looks like](https://docs.github.com/en/repositories/creating-and-managing-repositories/transferring-a-repository#transferring-a-repository-owned-by-your-personal-account) I need to be the one to do the transfer, and I can only do so to another personal account or to an org where I have permission to create repos. Probably easiest if I just transfer it to `josevalim` and you can bounce it over to the org from there. Will do so now.

mtrudel:
@josevalim done. You should now have two invitations to transfer the repos to your personal account, from where you can immediately transfer them to phoenixframework. Note that you'll probably need to add me to in the org, as I don't think my permissions will transfer all the way through.

lib/phoenix/socket/transport.ex
```diff
@@ -94,30 +94,6 @@ defmodule Phoenix.Socket.Transport do
...
-  supervision tree associated to the socket.
-
-  It receives the socket options from the endpoint,
-  for example:
-
-      socket "/my_app", MyApp.Socket, shutdown: 5000
-
-  means `child_spec([shutdown: 5000])` will be invoked.
-  """
-  @callback child_spec(keyword) :: :supervisor.child_spec()
```
josevalim, line 119: Why was this moved to Phoenix.Socket? I don't think it is actually part of Phoenix.Socket, because the user does not need to define a `child_spec` for `Phoenix.Socket`. :)

josevalim (commented):
One comment and I think we can ship this.

lib/phoenix/socket/transport.ex
mtrudel: (Note that we've [talked about this before](https://github.com/phoenixframework/phoenix/pull/4973#issuecomment-1261041615), but I suppose never came to a conclusion on it)

The actual concrete implementation of this callback hasn't changed - it's still `use`d into the Socket at https://github.com/mtrudel/phoenix/blob/af00565585bdbc5644f9ed0a1077d52d87ac3f0f/lib/phoenix/socket.ex#L309-L311. The only thing that changed here is which behaviour this is part of. The scope of this callback based on its docs makes this seem like a better home for it. Not strictly blocking, though.

josevalim: It should be part of `Phoenix.Socket.Transport` because someone can implement `Phoenix.Socket.Transport` so they bypass the whole channel machinery and I believe that, without this callback, it won't work.

mtrudel: Updated!

josevalim:

mtrudel:
A M A Z I N G 

Now that this is finalized I've published the last of the Bandit 0.5 series (`0.5.10`), which wraps up WebSocket and Phoenix support! Being able to say 'Bandit works with Phoenix 1.7, no strings attached' is *huge* for adoption. Thank you all so much for helping to make it happen.

I've got a number of follow-on PRs relating to completing the transition to Cowboy being 'one among many' possible servers for Phoenix, but they can & should wait until after 1.7 is out and Bandit's integration has had some time to bake in

PR #5065, opened 2022-11-02
Bugfix: Render inner block instead of @message

andreaseriksson:
This seems like a bug and was broken when I ran the `mix phx.gen.live ..` and then `mix test`


chrismccord:
❤️❤️❤️🐥🔥

PR #5068, opened 2022-11-03
Guides: request lifecycle and custom error pages

rktjmp:


guides/request_lifecycle.md
```diff
@@ -273,7 +283,9 @@ And this is what the template should look like:
 </section>
 ```
 
-Our messenger appears as `@messenger`. We call "assigns" the values passed from the controller to views. It is a special bit of metaprogrammed syntax which stands in for `assigns.messenger`. The result is much nicer on the eyes and much easier to work with in a template.
+Our messenger appears as `@messenger`.
+
+The values we passed to the view from the controller are collectively called our "assigns". We could access our messenger value via `assigns.messenger` but through some metaprogramming, Phoenix gives us the much cleaner `@` syntax for use in templates.
```
rktjmp, line 288: "We call "assigns" the values passed from the controller to views" read a bit backwards.

guides/request_lifecycle.md
```diff
@@ -265,6 +273,8 @@ For the last piece of this puzzle, we'll need a new template. Since it is for th
 
 To do that, we'll use the special EEx tags for executing Elixir expressions: `<%=  %>`. Notice that the initial tag has an equals sign like this: `<%=` . That means that any Elixir code that goes between those tags will be executed, and the resulting value will replace the tag in the HTML output. If the equals sign were missing, the code would still be executed, but the value would not appear on the page.
 
+Remember our templates are written in HEEx (HTML+EEx). HEEx is a superset of EEx which is why it shares the `<%= %>` syntax.
+
```
rktjmp, line 277: New users *might* get tripped up by "EEx tags" vs "aren't we writing HEEx?" so this tries to head that off at the pass.

guides/request_lifecycle.md
```diff
@@ -143,7 +147,11 @@ defmodule HelloWeb.HelloHTML do
 end
 ```
 
- Here we are telling `Phoenix.Component` to embed all `.heex` templates found in the sibling `hello` directory into our module as function definitions. Next, we need to add files to the `lib/hello_web/controllers/hello_html` directory. Note the controller name (`HelloController`), the view name (`HelloHTML`), and the template directory (`hello`) all follow the same naming convention and are named after each other. They are also collocated together in the directory tree:
+ Here we are telling `Phoenix.Component` to embed all `.heex` templates found in the sibling `hello` directory into our module as function definitions.
+
+ Next, we need to add files to the `lib/hello_web/controllers/hello_html` directory.
+
+ Note the controller name (`HelloController`), the view name (`HelloHTML`), and the template directory (`hello`) all follow the same naming convention and are named after each other. They are also collocated together in the directory tree:
```
rktjmp, line 154: Just whitespace

guides/request_lifecycle.md
```diff
@@ -117,7 +117,11 @@ defmodule HelloWeb.HelloHTML do
 end
 ```
 
-Now in order to add templates to this view, we can either define them as function components directly in the module:
```
rktjmp, line 120: This "either ..." was missing its friend, so i *colocated* both parts, functions and files then expand on them.

guides/request_lifecycle.md
```diff
@@ -117,7 +117,11 @@ defmodule HelloWeb.HelloHTML do
...
+To add templates to this view, we can define them as function components in the module or in separate files. Function components are great for smaller templates and separate files are a good choice when you have a lot of markup or your functions start to feel unmanageable.
+
+Both functional components and template files are compiled to an internal format, there is no runtime or performance difference between them.
```
rktjmp, line 122: Pretty sure this is true and should ease some new users questions ("are files slower? does it read each time?")

guides/howto/custom_error_pages.md
```diff
@@ -27,7 +27,7 @@ defmodule HelloWeb.ErrorHTML do
   end
 end
 
-defmodule HelloWeb.ErrorJSOn do
+defmodule HelloWeb.ErrorJSON do
```
rktjmp, line 30: whoops

guides/howto/custom_error_pages.md
```diff
@@ -167,9 +167,9 @@ However, implementing the `Plug.Exception` protocol by hand can be convenient in
 
 ## Actionable errors
 
-Exception actions are functions that can be triggered by the error page, and they're basically a list of maps defining a `label` and a `handler` to be executed.
+Exception actions are functions that can be triggered from the error page, and they're basically a list of maps defining a `label` and a `handler` to be executed. As an example, Phoenix will display an error if you have pending migrations and will provide a button on the error page to perform the pending migrations.
 
-They are rendered in the error page as a collection of buttons and follow the format of:
+When `debug_errors` is `true`, they are rendered in the error page as a collection of buttons and follow the format of:
```
rktjmp, line 172: The user may have turned this off previously in the guide, so important to note.

Eiji7 (approved):

guides/request_lifecycle.md
Gazler: I think instead of this line we could say:

> Template files are compiled into the module, so there is no runtime or performance difference between using a separate file or the `~H` sigil.

Gazler (commented):
Good changes. I've made one small suggestion that clarifies slightly what happens with template files.

guides/request_lifecycle.md
rktjmp: I think using the term "~H sigil" might be premature as we've not discussed `~H` or sigils yet (it is in the next post-code para though). The reader might go "what is a function component? wait what is ~H?" vs reusing the same terms we just outlaid.

How do you feel about this?

> To add templates to this view, we can define them as function components in the module or in separate files. Function components are great for smaller templates and separate files are a good choice when you have a lot of markup or your functions start to feel unmanageable.
>
> Template files are compiled into the module as function components themselves, there is no runtime or performance difference between the two styles.

The only real key point I think we need to make is that the reader shouldn't stress about picking between the two!

Gazler: Perfect!

rktjmp:
Fixed merge conflict & applied feedback

Gazler:
Thanks <3

PR #5069, opened 2022-11-03
Fix select multiple name

LostKobrakai:


installer/templates/phx_web/components/core_components.ex
```diff
@@ -258,14 +258,18 @@ defmodule <%= @web_namespace %>.CoreComponents do
...
+                                   pattern placeholder readonly required size step)
   slot :inner_block
 
   def input(%{field: {f, field}} = assigns) do
     assigns
     |> assign(field: nil)
-    |> assign_new(:name, fn -> Phoenix.HTML.Form.input_name(f, field) end)
+    |> assign_new(:name, fn ->
+      name = Phoenix.HTML.Form.input_name(f, field)
+      if assigns.multiple, do: name <> "[]", else: name
```
chrismccord, line 271: let's move this into the input select clause since it only applies there

chrismccord:
❤️❤️❤️🐥🔥

PR #5070, opened 2022-11-04
Remove reference to Milligram and replace with Tailwind

hwatkins:
Remove reference to Milligram and replace with Tailwind


Gazler:
Good catch, thanks!

PR #5071, opened 2022-11-04
Add `--adapter` flag

moogle19:
Adds a `--adapter` flag to `mix phx.new` to choose the web adapter for a new application (allows cowboy or bandit, defaults to cowboy).

I am not quite happy with having the versions in the `generator.ex`, but not sure if using `>= 0.0.0` (like for the ecto_adapters) will lead to problems in the future.


Gazler:
I'm not sure if we're ready to merge this yet, but it's worth updating the following files too:

https://github.com/phoenixframework/phoenix/blob/master/installer/templates/phx_umbrella/apps/app_name_web/config/config.exs
https://github.com/phoenixframework/phoenix/blob/master/installer/templates/phx_umbrella/apps/app_name_web/mix.exs

mtrudel:
I was planning on submitting a PR for this after 1.7 was out, on the reasoning that we should probably let bandit support bake in for a bit before having Phoenix but any sort of 'official' stamp on it. This feels like a 1.7.x type thing, lest Bandit support get a bad name during the inevitable teething process. 

The PR looks great though!

installer/templates/phx_single/mix.exs
```diff
@@ -54,7 +54,7 @@ defmodule <%= @app_module %>.MixProject do
       {:telemetry_poller, "~> 1.0"},<%= if @gettext do %>
       {:gettext, "~> 0.20"},<% end %>
       {:jason, "~> 1.2"},
-      {:plug_cowboy, "~> 2.5"}
+      {<%= inspect @web_adapter_app %>, <%= inspect @web_adapter_version %>}
```
mtrudel, line 57: I wonder (as in I'm legitimately not sure) if this would be better expressed via a fixed `">= 0.0.0"` here, and letting phoenix itself declare specific version dependencies via an `:optional` dependency on bandit (it already has one on plug_cowboy). This would keep the onus on Phoenix to manage server dep versions, as we [just did](https://github.com/phoenixframework/phoenix/pull/5030/files#diff-dfa6f4ed74c90e5d4fda283d547d366586e690387289bcfd473e3fa5f9ace308R83) for WebSock support.

mtrudel:
There's also test coverage to be added here - feel free to take at will from the (mostly done, but not heavily tested) copy of the work I've done on this!  https://github.com/mtrudel/phoenix/tree/bandit_generator

mtrudel:
Looks great!

josevalim (approved):

mtrudel:
This continues to look good to me; I'll leave it at the discretion of @josevalim et al about timing, but from my perspective this is ready to go anytime.

chrismccord:
❤️❤️❤️🐥🔥

mtrudel:
Woooo!

PR #5072, opened 2022-11-04
Fix input component when multiple flag isn't given

doughsay:
When `multiple` isn't given, inputs fail to render because `name` uses `assigns.multiple` when generating the name.


LostKobrakai:
This has been merged here: https://github.com/phoenixframework/phoenix/pull/5073

PR #5075, opened 2022-11-05
Remove unnecessary \ from phx.gen.auth info

zolrath:
While swapping from
`- Using Phoenix.View` to `- Using Phoenix.Controller only` an erroneous `\` was added.


josevalim:
It is not erronous. That avoids an additional new line from being added.

PR #5076, opened 2022-11-06
Properly generate swoosh configuration for finch adapter

LostKobrakai:
The generators should generate `MyApp.Finch`, not `my_app.Finch`


josevalim:

aus70:
Just a note: with Phoenix 1.7.0-rc.0, Swoosh 1.8.2, in `application.ex`, I had to use `{Finch, name: Swoosh.Finch}` instead of the generated `{Finch, name: MyApp.Finch}` and  in the config, `config :swoosh, :api_client, Swoosh.ApiClient.Finch`  instead of `config :swoosh, :api_client, MyApp.Finch`, otherwise I would get a `(ArgumentError) unknown registry: Swoosh.Finch`.

PR #5082, opened 2022-11-08
Fix broken live_reload in v1.7.0-rc.0

c4710n:
Following directories change in 1.7:
+ `views/` and `templates/` are replaced by `controllers/`.
+ `components/` is added.

This PR makes live_reload work with new file structure.

### DIFF
--- a/installer/templates/phx_single/config/dev.exs
+++ b/installer/templates/phx_single/config/dev.exs
@@ -48,8 +48,7 @@ config :<%= @app_name %>, <%= @endpoint_module %>,
     patterns: [
       ~r"priv/static/.*(js|css|png|jpeg|jpg|gif|svg)$",<%= if @gettext do %>
       ~r"priv/gettext/.*(po)$",<% end %>
-      ~r"lib/<%= @lib_web_name %>/(live|views)/.*(ex)$",
-      ~r"lib/<%= @lib_web_name %>/templates/.*(eex)$"
+      ~r"lib/<%= @lib_web_name %>/(controllers|live|components)/.*(ex|heex)$"
     ]
   ]<% end %>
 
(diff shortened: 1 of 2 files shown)


Eiji7 (approved):

Eiji7:
This PR closes #5085

dfalling:
In my project I added `layouts` as well to hot load the app and root files:
```
~r"lib/ido_web/(controllers|live|components|layouts)/.*(ex|heex)$"
```

installer/templates/phx_single/config/dev.exs
```diff
@@ -48,8 +48,7 @@ config :<%= @app_name %>, <%= @endpoint_module %>,
     patterns: [
       ~r"priv/static/.*(js|css|png|jpeg|jpg|gif|svg)$",<%= if @gettext do %>
       ~r"priv/gettext/.*(po)$",<% end %>
-      ~r"lib/<%= @lib_web_name %>/(live|views)/.*(ex)$",
-      ~r"lib/<%= @lib_web_name %>/templates/.*(eex)$"
+      ~r"lib/<%= @lib_web_name %>/(controllers|live|components)/.*(ex|heex)$"
```
mcrumm, line 51:
```suggestion
      ~r"lib/<%= @lib_web_name %>/(controllers|live|components)/.*(ex)$"
```

c4710n: I got your point - `.*(ex)` will match `.ex` file and `.heex` file. 

But I prefer to use `.*(ex|heex)`, which **tells the user explicitly that we are matching `.ex` file and `.heex` file.**

If you think `.*(ex|heex)` is making the regexp redundant. Maybe we can use a explicit one, like `.*\.(ex|heex)`. Then the final suggestion is:
```elixir
~r"lib/<%= @lib_web_name %>/(controllers|live|components)/.*\.(ex|heex)$"
```
What do you think?

c4710n: For code consistency, it is also possible to adjust other two lines:
```elixir
~r"priv/static/.*\.(js|css|png|jpeg|jpg|gif|svg)$",<%= if @gettext do %>
~r"priv/gettext/.*\.(po)$",<% end %>
```

mcrumm: I prefer the simpler regex, and if defining both `ex` and `heex` seems better for clarity then I would vote to leave it as-is, meaning ignore my suggestions :)

chrismccord:
❤️❤️❤️🐥🔥

PR #5086, opened 2022-11-08
Fix "Deploy your application" link

zimt28:
The current link redirects to https://fly.io/docs/elixir/getting-started/iex-into-running-app/.


Gazler:
Thanks <3

PR #5090, opened 2022-11-08
Unify `gen.auth` and `phx.new` menu styles

vanderhoop:
Hi all,
Pulled the 1.7rc, and noticed that the `gen.auth` templates aren't stylistically aligned with the awesome work that the tailwind devs applied to the rest of the `phx.new` templates. There are a handful of additional updates I would make, but I wanted to keep the PRs small, so this one just updates the auth-related link styles so that they're identical to the link styles from `phx.new`.

## Before
before_styling_updates

## After

after_styling_updates

## Possible Additional Update

I _think_ it would make sense to replace the `@phoenixframework`, `GitHub` and `Getting Started` links if a user has already taken the step to generate authentication, as that implies that the developer already has a level of fluency in the elixir/phoenix ecosystem. Then these registration links would sit in their place, inline with the phoenix logo, but that change feels like a bigger discussion and might be non-trivial given the existing structure of the generators and templates.

### DIFF
--- a/lib/mix/tasks/phx.gen.auth/injector.ex
+++ b/lib/mix/tasks/phx.gen.auth/injector.ex
@@ -149,6 +149,7 @@ defmodule Mix.Tasks.Phx.Gen.Auth.Injector do
   """
   def app_layout_menu_help_text(file_path, %Schema{} = schema) do
     {_dup_check, code} = app_layout_menu_code_to_inject(schema)
+
     """
     Add the following #{schema.singular} menu items to your #{Path.relative_to_cwd(file_path)} layout file:
 
@@ -172,17 +173,38 @@ defmodule Mix.Tasks.Phx.Gen.Auth.Injector do
           <%= @current_#{schema.singular}.email %>
         </li>
         <li>
-          <.link href={~p"#{schema.route_prefix}/settings"} class="#{link_tailwind_classes}">Settings</.link>
+          <.link
+            href={~p"#{schema.route_prefix}/settings"}
+            class="#{link_tailwind_classes}"
+          >
+            Settings
+          </.link>
         </li>
         <li>
-          <.link href={~p"#{schema.route_prefix}/log_out"} method="delete" class="#{link_tailwind_classes}">Log out</.link>
+          <.link
+            href={~p"#{schema.route_prefix}/log_out"}
+            method="delete"
+            class="#{link_tailwind_classes}"
+          >
+            Log out
+          </.link>
         </li>
       <% else %>
         <li>
-          <.link href={~p"#{schema.route_prefix}/register"} class="#{link_tailwind_classes}">Register</.link>
+          <.link
+            href={~p"#{schema.route_prefix}/register"}
+            class="#{link_tailwind_classes}"
+          >
+            Register
+          </.link>
         </li>
         <li>
-          <.link href={~p"#{schema.route_prefix}/log_in"} class="#{link_tailwind_classes}">Log in</.link>
+          <.link
+            href={~p"#{schema.route_prefix}/log_in"}
+            class="#{link_tailwind_classes}"
+          >
+            Log in
+          </.link>
         </li>
       <% end %>
     </ul>\
--- a/lib/mix/tasks/phx.gen.auth/injector.ex
+++ b/lib/mix/tasks/phx.gen.auth/injector.ex
@@ -166,7 +166,7 @@ defmodule Mix.Tasks.Phx.Gen.Auth.Injector do
     link_tailwind_classes = "#{base_tailwind_classes} font-semibold hover:text-zinc-700"
 
     template = """
-    <ul class="flex items-center gap-4 px-4 sm:px-6 lg:px-8">
+    <ul class="flex items-center gap-4 px-4 sm:px-6 lg:px-8 justify-end">
       <%= if @current_#{schema.singular} do %>
         <li class="#{base_tailwind_classes}">
           <%= @current_#{schema.singular}.email %>
--- a/lib/mix/tasks/phx.gen.auth/injector.ex
+++ b/lib/mix/tasks/phx.gen.auth/injector.ex
@@ -163,7 +163,7 @@ defmodule Mix.Tasks.Phx.Gen.Auth.Injector do
   def app_layout_menu_code_to_inject(%Schema{} = schema, padding \\ 4, newline \\ "\n") do
     already_injected_str = "#{schema.route_prefix}/log_in"
 
-    base_tailwind_classes = "text-[0.8125rem] leading-6 text-zinc-900 hover:text-zinc-700"
+    base_tailwind_classes = "text-[0.8125rem] leading-6 text-zinc-900"
     link_tailwind_classes = "#{base_tailwind_classes} font-semibold hover:text-zinc-700"
 
     template = """
--- a/lib/mix/tasks/phx.gen.auth/injector.ex
+++ b/lib/mix/tasks/phx.gen.auth/injector.ex
@@ -175,7 +175,7 @@ defmodule Mix.Tasks.Phx.Gen.Auth.Injector do
           <.link href={~p"#{schema.route_prefix}/settings"} class="#{link_tailwind_classes}">Settings</.link>
         </li>
         <li>
-          <.link href={~p"#{schema.route_prefix}/log_out"} class="#{link_tailwind_classes}" method="delete">Log out</.link>
+          <.link href={~p"#{schema.route_prefix}/log_out"} method="delete" class="#{link_tailwind_classes}">Log out</.link>
         </li>
       <% else %>
         <li>
--- a/lib/mix/tasks/phx.gen.auth/injector.ex
+++ b/lib/mix/tasks/phx.gen.auth/injector.ex
@@ -162,24 +162,27 @@ defmodule Mix.Tasks.Phx.Gen.Auth.Injector do
   def app_layout_menu_code_to_inject(%Schema{} = schema, padding \\ 4, newline \\ "\n") do
     already_injected_str = "#{schema.route_prefix}/log_in"
 
+    base_tailwind_classes = "text-[0.8125rem] leading-6 text-zinc-900 hover:text-zinc-700"
+    link_tailwind_classes = "#{base_tailwind_classes} font-semibold hover:text-zinc-700"
+
     template = """
     <ul>
       <%= if @current_#{schema.singular} do %>
-        <li>
+        <li class="#{base_tailwind_classes}">
           <%= @current_#{schema.singular}.email %>
         </li>
         <li>
-          <.link href={~p"#{schema.route_prefix}/settings"}>Settings</.link>
+          <.link href={~p"#{schema.route_prefix}/settings"} class="#{link_tailwind_classes}">Settings</.link>
         </li>
         <li>
-          <.link href={~p"#{schema.route_prefix}/log_out"} method="delete">Log out</.link>
+          <.link href={~p"#{schema.route_prefix}/log_out"} class="#{link_tailwind_classes}" method="delete">Log out</.link>
         </li>
       <% else %>
         <li>
-          <.link href={~p"#{schema.route_prefix}/register"}>Register</.link>
+          <.link href={~p"#{schema.route_prefix}/register"} class="#{link_tailwind_classes}">Register</.link>
         </li>
         <li>
-          <.link href={~p"#{schema.route_prefix}/log_in"}>Log in</.link>
+          <.link href={~p"#{schema.route_prefix}/log_in"} class="#{link_tailwind_classes}">Log in</.link>
         </li>
       <% end %>
     </ul>\
--- a/lib/mix/tasks/phx.gen.auth/injector.ex
+++ b/lib/mix/tasks/phx.gen.auth/injector.ex
@@ -166,7 +166,7 @@ defmodule Mix.Tasks.Phx.Gen.Auth.Injector do
     link_tailwind_classes = "#{base_tailwind_classes} font-semibold hover:text-zinc-700"
 
     template = """
-    <ul>
+    <ul class="flex items-center gap-4 px-4 sm:px-6 lg:px-8">
       <%= if @current_#{schema.singular} do %>
         <li class="#{base_tailwind_classes}">
           <%= @current_#{schema.singular}.email %>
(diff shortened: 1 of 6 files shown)


lib/mix/tasks/phx.gen.auth/injector.ex
```diff
@@ -162,24 +162,27 @@ defmodule Mix.Tasks.Phx.Gen.Auth.Injector do
   def app_layout_menu_code_to_inject(%Schema{} = schema, padding \\ 4, newline \\ "\n") do
     already_injected_str = "#{schema.route_prefix}/log_in"
 
+    base_tailwind_classes = "text-[0.8125rem] leading-6 text-zinc-900 hover:text-zinc-700"
```
vanderhoop, line 165: I'm new to tailwind, so I'm not sure what the best practice is for reusing classes, as these are the classes applied to the links in the header from the phx.new generator. It might make sense to house these classes in a common place, just not sure what they convention would be within a phoenix app.

lib/mix/tasks/phx.gen.auth/injector.ex
```diff
@@ -162,24 +162,27 @@ defmodule Mix.Tasks.Phx.Gen.Auth.Injector do
...
+    link_tailwind_classes = "#{base_tailwind_classes} font-semibold hover:text-zinc-700"
+
```
vanderhoop, line 167: __Note:__ these updates cause the `test/mix/tasks/phx.gen.auth/injector_test.exs` tests to fail. The tests appear to be snapshot tests. Does the team we want to keep them that way? It feels bizarre to me anytime I have to copy the output from a new update and paste-update a test assertion, especially given the noise that all of these classes will end up adding to the snapshot.

My suggestion would be to update those tests so that they assert particular behaviors, which I'm happy to do, but if the team wants to keep the snapshot approach, happy to update those, too.

derpycoder:
Here’s my take on a Unfied Auth look:

[](https://elixirforum.com/t/phoenix-1-7-0-rc-0-is-out/51625/15?u=derpycoder) 

6D736FEE-5975-4FD3-8F11-BC9FC8D83D23
4FE852EB-F605-49F7-8329-17B80F0E7D8A
A069A580-D0E8-4B76-822E-9F9858ACB300

derpycoder:
3108B327-A1A8-4E8E-A000-626B05E41BC3

Eiji7:
Any update for this? @derpycoder proposal looks great and I'm looking forward to test `phx.new` with this update.

derpycoder:
I am a beginner, I just made some styling changes in the front end and showed them here. I did not contribute to the project, sorry about that.

I just thought I can influence @vanderhoop to make a slight modification, as he took the initiative to make this look awesome.

I don't know if the design I proposed, will work with the phoenix auth generator or not, so I left it to experienced developers.

lib/mix/tasks/phx.gen.auth/injector.ex
vanderhoop: Based on the comment [here](https://github.com/phoenixframework/phoenix/issues/5176#issuecomment-1387733898), sounds like this can be resolved by using the components in core_components. Gonna play around and verify.

vanderhoop:
hey @derpycoder + @Eiji7! I've been punting on this, as I was waiting word on how the maintainers wanted the tests updated (see [comment](https://github.com/phoenixframework/phoenix/pull/5090#discussion_r1016927569) above). That said, I'll pull the latest RC and see if I can implement some of the styles proposed without much of a lift, as this would close issue #5176. Updating the test assertions in the snapshot sense would be good enough, as that's the status quo, and assertion updates could follow at any time.

lib/mix/tasks/phx.gen.auth/injector.ex
vanderhoop: Okay, so played around with that, and there are no base `link` components defined in core components. The `.link` component comes from Phoenix.LiveView.Helpers, and I'm hesitant to change global link styles across the entire phoenix stack for this change.

test/mix/tasks/phx.gen.auth_test.exs
```diff
@@ -192,16 +192,16 @@ defmodule Mix.Tasks.Phx.Gen.AuthTest do
 
       assert_file("lib/my_app_web/components/layouts/root.html.heex", fn file ->
         assert file =~
-                 ~S|<.link href={~p"/users/settings"}>Settings</.link>|
+                 ~r|<.link href={~p"/users/settings"}.*>Settings</.link>|
```
vanderhoop, line 195: Rather than duplicate the tailwind classes in all of these assertions, I shifted the assertions to regexes to be more permissive to tailwind style changes down the road.

The longer, multi-line snapshot tests (below) have been updated with explicit matches.

test/mix/tasks/phx.gen.auth_test.exs
```diff
@@ -192,16 +192,16 @@ defmodule Mix.Tasks.Phx.Gen.AuthTest do
 
       assert_file("lib/my_app_web/components/layouts/root.html.heex", fn file ->
         assert file =~
-                 ~S|<.link href={~p"/users/settings"}>Settings</.link>|
```
vanderhoop, line 195: Note that I've killed off the inner text in these assertions due to mess it'd have to make of the regexp matchers due to the additional whitespace added by the formatter. I'm comfortable with the removal of the text and closing-tag from these assertions due to their being covered in the snapshot tests in this file and the injector tests, both of which due 1:1 string equality checks on the entirety of the output.

priv/templates/phx.gen.auth/forgot_password_live_test.exs
```diff
@@ -12,8 +12,8 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
       {:ok, _lv, html} = live(conn, ~p"<%= schema.route_prefix %>/reset_password")
 
       assert html =~ "Forgot your password?"
-      assert html =~ "Register</a>"
-      assert html =~ "Log in</a>"
```
vanderhoop, line 16: @aaronrenner + @josevalim Note that I've killed off the inner text in these assertions due to mess it'd have to make of the regexp matchers due to the additional whitespace added by the formatter. I'm comfortable with the removal of the text and closing-tag from these assertions due to their being covered in the snapshot tests in the injector tests, which do 1:1 string equality checks on the entirety of the output.

test/mix/tasks/phx.gen.auth_test.exs
```diff
@@ -192,16 +192,16 @@ defmodule Mix.Tasks.Phx.Gen.AuthTest do
...
+                 ~r|<.link.*href={~p"/users/settings"}.*>|s
```
vanderhoop, line 195: Updated these assertions to regexes to accommodate the multi-lining needed to satisfy the formatter. Explicit matches are still covered in the injector test.

vanderhoop:
hey @derpycoder + @Eiji7! I'm going to keep this PR scoped to just updating the link styles, and avoid changing their placement in the layouts for now, as the injector's logic for placement is non-trivial and getting the various CI test suites to pass took more time than I expected 😅.

If and when _this_ PR gets merged, I think I can open a follow-up getting these links get added to the _existing_ set of links in the header. It won't _quite_ meet the design you proposed up above for the header, but I think it would be an overall aesthetic improvement.

lib/mix/tasks/phx.gen.auth/injector.ex
codebycaleb: @vanderhoop I was just taking a look at the tests and noticed that `hover:text-zinc-700` is defined in `base_tailwind_classes` and right here, so it ends up being duplicated in the `link_tailwind_classes` output:

```
text-[0.8125rem] leading-6 text-zinc-900 hover:text-zinc-700 font-semibold hover:text-zinc-700
                                         ^^^^^^^^^^^^^^^^^^^               ^^^^^^^^^^^^^^^^^^^
```

vanderhoop: 🤦 . Great catch, @codebycaleb. I'll update the code and snapshots this weekend.

vanderhoop: @codebycaleb I've fixed the redundancy in https://github.com/vanderhoop/phoenix/commit/952421d8fc62d6ed8cf5dd060d3b2608029de4af. @jeregrine I think this PR is now ready to go

priv/templates/phx.gen.auth/forgot_password_live_test.exs
```diff
@@ -12,8 +12,8 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
...
+      assert html =~ "/users/register"
```
chrismccord, line 15: has_element? should work on these and give us the best of both worlds
```suggestion
      assert has_element?(lv, ~s|a[href="#{~p"/users/register"}"]|, "Register")
```

chrismccord:
Once we replace the html checks with `has_element?`, this will be ready to go. 

> I think it would make sense to replace the @phoenixframework, GitHub and Getting Started links if a user has already taken the step to generate authentication, as that implies that the developer already has a level of fluency in the elixir/phoenix ecosystem. Then these registration links would sit in their place, inline with the phoenix logo, but that change feels like a bigger discussion and might be non-trivial given the existing structure of the generators and templates.

Yes, the design was made exactly for the login/signed in nav to go where those links are, but the issue is `phx.gen.auth` must work for all apps – out-of-the-box apps, tailwind apps that are completely custom, apps that use other css altogether, etc.

So we can't easily inject a matching nav into the design. The safest and most portable option to to inject it just after <body> like we do now, and let the user customize. Your approach of doing that + light tailwind styling strikes a good balance. Thanks for being patient on this one!

priv/templates/phx.gen.auth/forgot_password_live_test.exs
vanderhoop: resolved in https://github.com/phoenixframework/phoenix/pull/5090/commits/e86b6b372e285f6488a5cdd3f31aa1c88c233cc5

Eiji7:
@chrismccord When reading your reply something come to my mind … Can we add `--live` option to `phx.new` task as well? Here are the main points:
1. `phx.gen.auth` stays as you said generic for all projects - if we add `--live` option then there is no need to call `phx.gen.auth` right after `phx.new` which means that the `target` is reduced only to custom apps and `phx.new` based apps that are already modified
2. `phx.new` with `--live` would give a simple way using as one task generator for new apps with `auth` support
3. We can unify styles only for new apps (i.e. in `phx.new`) which is reasonable as of 1st point

What do you think about it?

chrismccord:
No there is still little guarantee that the auth generator menu will match what the user has in their app and we shouldn’t run auth by default. It’s just a starting point so looking decent is a solid stating point.

priv/templates/phx.gen.auth/registration_controller_test.exs
```diff
@@ -8,8 +8,8 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
       conn = get(conn, ~p"<%= schema.route_prefix %>/register")
       response = html_response(conn, 200)
       assert response =~ "Register"
-      assert response =~ "Log in</a>"
-      assert response =~ "Register</a>"
+      assert response =~ "/users/log_in"
```
chrismccord, line 11:
```suggestion
      assert response =~ ~p"/users/log_in"
```

priv/templates/phx.gen.auth/registration_controller_test.exs
```diff
@@ -8,8 +8,8 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
       conn = get(conn, ~p"<%= schema.route_prefix %>/register")
       response = html_response(conn, 200)
       assert response =~ "Register"
-      assert response =~ "Log in</a>"
-      assert response =~ "Register</a>"
+      assert response =~ ~p"/users/log_in"
+      assert response =~ "/users/register"
```
chrismccord, line 12:
```suggestion
      assert response =~ ~p"/users/register"
```

priv/templates/phx.gen.auth/registration_controller_test.exs
```diff
@@ -36,8 +36,8 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
       conn = get(conn, ~p"/")
       response = html_response(conn, 200)
       assert response =~ email
-      assert response =~ "Settings</a>"
-      assert response =~ "Log out</a>"
+      assert response =~ "/users/settings"
```
chrismccord, line 39:
```suggestion
      assert response =~ ~p"/users/settings"
```

priv/templates/phx.gen.auth/registration_controller_test.exs
```diff
@@ -36,8 +36,8 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
       conn = get(conn, ~p"/")
       response = html_response(conn, 200)
       assert response =~ email
-      assert response =~ "Settings</a>"
-      assert response =~ "Log out</a>"
+      assert response =~ ~p"/users/settings"
+      assert response =~ "/users/log_out"
```
chrismccord, line 40:
```suggestion
      assert response =~ ~p"/users/log_out"
```

priv/templates/phx.gen.auth/session_controller_test.exs
```diff
@@ -36,8 +36,8 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
       conn = get(conn, ~p"/")
       response = html_response(conn, 200)
       assert response =~ <%= schema.singular %>.email
-      assert response =~ "Settings</a>"
-      assert response =~ "Log out</a>"
+      assert response =~ "/users/settings"
```
chrismccord, line 39:
```suggestion
      assert response =~ ~p"/users/settings"
```

priv/templates/phx.gen.auth/session_controller_test.exs
```diff
@@ -36,8 +36,8 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
...
+      assert response =~ "/users/log_out"
```
chrismccord, line 40:
```suggestion
      assert response =~ ~p"/users/log_out"
```

chrismccord:
❤️❤️❤️🐥🔥

PR #5091, opened 2022-11-08
Fix extra whitespace in textarea in core_components

salva-ruiz:
The _textarea_ input type in `core_components.ex` adds a new line before the field value which also causes that a blank to be saved when submitting the form.

This error occurs on Chrome and Safari.

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -329,9 +329,7 @@ defmodule <%= @web_namespace %>.CoreComponents do
           "phx-no-feedback:border-zinc-300 phx-no-feedback:focus:border-zinc-400 phx-no-feedback:focus:ring-zinc-800/5"
         ]}
         {@rest}
-      >
-
-    <%%= @value %></textarea>
+      ><%%= @value %></textarea>
       <.error :for={msg <- @errors}><%%= msg %></.error>
     </div>
     """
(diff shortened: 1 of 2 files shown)


salva-ruiz:
@LostKobrakai I think no because _textarea_ is an inline element and according to the formatter code it doesn't force line breaks of inline elements: https://github.com/phoenixframework/phoenix_live_view/blob/703170844436ce7a1327b3a111a3967673b684c8/lib/phoenix_live_view/html_formatter.ex#L190

chrismccord:
This is intentional because browsers add a newline before the textarea always, which means if we made the proposed change a user would see their newlines disappear after submitting a form if they prefixed their inputs with a new line. The formatter respects whitespace in a textarea. Thanks!

theenglishway:
Hi all !

I hope it's not bad form to ask a question here even though the issue is closed.
I stumbled upon this issue myself after wondering why I got extra `"\n"` in the inputs of the `textarea` component from `core_components.ex` of the 1.7-rc0.

I myself encountered two problems with that component that I solved with the exact same fix as @salva-ruiz : 
* it produces a non-empty value (`"\n"`) if I submit it without inputting any text
* selecting it to input a value brings me in the middle of the textarea (instead of the top left), which I thought was a CSS issue at first
* the produced value always starts with a `"\n"` (expect of course if I delete it manually)

I actually didn't even realize the problem at first while playing around :eyes: 

Being still kinda newbie in the webdev world, is this behaviour really expected with a textarea, and if so, what's the advised way to get a textarea that outputs a `""`, since the solution provided in that PR does not seem to be the right one ?

(I run Mozilla Firefox 105.0.1 if that's of any use)

mxgrn:
Can't wrap my head around how this can be intentional. It's not the browser which adds an empty line, it's the way the HTML is being generated. Every time I edit a record via a form with a `textarea` it keeps adding a blank line at the top. It's a bug.

jadengis:
I have been using this in practice and it is definitely a bug. If you populate a textarea with content from a changeset (say updating data), this behavior will insert an extra newline at the top of the content, mutating the data in the form field. This means submitting the form without the user changing the content in the textarea will actually change the data in the database. Furthermore, every time you submit again will add an additional newline. This is not correct.

PR #5094, opened 2022-11-10
Typo Fix: Missing curly bracket added.

reza-naq:
~p"/users/17?#{[admin: true]" -> ~p"/users/17?#{[admin: true]}"


Gazler:
Thanks <3

PR #5095, opened 2022-11-10
Fix layouts module in component guide

kevPo:
First time contributor here. 😅  

Working through the guide and ran into errors as it was written. I believe I've updated the guide properly but happy to make changes if not.


josevalim:

PR #5096, opened 2022-11-10
Add a call to `humanize` for input form generation for consistency

zolrath:
The template generation for tables in `phx.gen.html/index.html.heex` uses a call to `Phoenix.Naming.humanize` to generate its labels while the generator for input forms only calls `to_string`.

This modifies the `label` function to call humanize as well for consistency.

## Before:
Listing Page
generated form

## After:
image


zolrath:
Just as a sanity check, Phoenix 1.6's `label` function humanized the labels as well.
image

josevalim (approved):

chrismccord:
❤️❤️❤️🐥🔥

PR #5097, opened 2022-11-10
Clarify usage of push/2 vs replies

nathanl:


lib/phoenix/channel.ex
```diff
@@ -46,43 +46,44 @@ defmodule Phoenix.Channel do
...
         {:noreply, socket}
       end
 
   ## Replies
 
-  In addition to pushing messages out when you receive a `handle_in` event,
-  you can also reply directly to a client event for request/response style
-  messaging. This is useful when a client must know the result of an operation
-  or to simply ack messages.
+  Replies are useful for acking a client's message or responding with the
```
rktjmp, line 82: `acking` may be a difficult term for non-english readers, perhaps we should just use `acknowledge` directly or `respond`?

Could this also use mention of `{:noreply, socket}`? 

> If you do not wish to send a response back to the client, you can return `{:noreply, socket}`

or similar?

nathanl: Changed @rktjmp

lib/phoenix/channel.ex
```diff
@@ -586,19 +633,24 @@ defmodule Phoenix.Channel do
...
-  callbacks. `reply/2` will be used in the rare cases you need to perform work in
-  another process and reply when finished by generating a reference to the push
-  with `socket_ref/1`.
-
-  *Note*: In such cases, a `socket_ref` should be generated and
-  passed to the external process, so the `socket` itself is not leaked outside
-  the channel. The `socket` holds information such as assigns and transport
-  configuration, so it's important to not copy this information outside of the
-  channel that owns it.
+  Replies to a socket push. This is useful for asynchronous replies.
```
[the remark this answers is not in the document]
nathanl: True. Updated.

Gazler (approved):

chrismccord:
❤️❤️❤️🐥🔥

PR #5100, opened 2022-11-11
Remove blank favicon.ico

cw789:
We all agree, favicon is something app specific. But...

The blank favicon.ico was added to avoid Browsers requesting non existing favicon.ico file. Mainly to avoid error logs.
But the blank favicon.ico is still downloaded (without any benefit). And it makes the favicon space in Browser tabs blank.
The blank space doesn't look good and even during dev time it makes the tab harder to distinguish within the browser tab bar.

This variant doesn't download any favicon and uses a Browser default.

Before:
grafik
grafik

After:
grafik
grafik

### DIFF
--- a/installer/templates/phx_single/lib/app_name_web.ex
+++ b/installer/templates/phx_single/lib/app_name_web.ex
@@ -17,7 +17,7 @@ defmodule <%= @web_namespace %> do
   and import those modules here.
   """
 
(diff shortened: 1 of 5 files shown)


chrismccord:
It is also there to let folks know how to set it/customize it. Thanks!

PR #5101, opened 2022-11-12
Disable ability to scroll background while in modal

zolrath:
## Problem
Currently while the modal is open a user can still scroll the primary page in the background. 
This is usually unexpected and also leads to two scrollbars being rendered by Chrome.
image

## Solution
In order to prevent the user from scrolling the background, one common practice is to set the `body` element to `overflow: hidden` while the modal is open.  

One issue with this method is that it causes the sidebar to disappear, triggering a layout shift of the pages content.

Per [CanIUse](https://caniuse.com/mdn-css_properties_scrollbar-gutter), all browsers (aside from Safari which uses floating scrollbars which do not take up space) have implemented support for `scrollbar-gutter` so we can use `scrollbar-gutter: stable` to prevent the shift.

## Issues
`scrollbar-gutter` works perfectly well in all browsers tested, aside from a [known issue in Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=1318404#c4) which prevents the scrollbar for the modal from rendering. The temporary hidden gutter is drawn above any `fixed` position element, causing the modals scrollbar to be hidden.

The `scrollbar-gutter` portion could be removed, keeping the layout shift from the sidebar disappearing but resolving the background scrolling/double scrollbar issue.

## Alternatives
The HTML5 [Dialog Element](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/dialog) allows modals to be shown with a `showModal` command. Per [CanIUse](https://caniuse.com/mdn-api_htmldialogelement_showmodal) this has support in all browsers aside from IE11. It performs proper focus trapping, backdrop colors, prevents scrolling of backgrounds etc.

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -568,6 +568,7 @@ defmodule <%= @web_namespace %>.CoreComponents do
       transition: {"transition-all transform ease-out duration-300", "opacity-0", "opacity-100"}
     )
     |> show("##{id}-container")
+    |> JS.add_class("overflow-hidden", to: "body")
     |> JS.focus_first(to: "##{id}-content")
   end
 
@@ -579,6 +580,7 @@ defmodule <%= @web_namespace %>.CoreComponents do
     )
     |> hide("##{id}-container")
     |> JS.hide(to: "##{id}", transition: {"block", "block", "hidden"})
+    |> JS.remove_class("overflow-hidden", to: "body")
     |> JS.pop_focus()
   end
 
--- a/installer/templates/phx_web/components/layouts/root.html.heex
+++ b/installer/templates/phx_web/components/layouts/root.html.heex
@@ -1,5 +1,5 @@
 <!DOCTYPE html>
-<html lang="en">
+<html lang="en" class="[scrollbar-gutter:stable]">
   <head>
     <meta charset="utf-8" />
     <meta name="viewport" content="width=device-width, initial-scale=1" />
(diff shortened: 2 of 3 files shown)


rktjmp:
Just asking, is there a reason to not toggle the scrollbar-gutter in the JS event vs always in the template? Probably kind of race-y with the layout change?

zolrath:
In my testing setting scrollbar-gutter it in the JS event causes a flash of multiple scrollbars when closing the modal.

Having `scrollbar-gutter: stable` set also prevents things like opening/closing accordions or dynamically added content that causes the scrollbar to appear/disappear from creating a layout shift as well, so it's quite useful even outside of the context of the modal.

shamanime:
I found this issue in our production app and implemented this exact same code to add and remove the overflow to the body, the issue is there's some circumstances where the modal may close without firing the close event (a redirect or patch that removes it from the HTML for instance) and the body stays locked. 

To prevent this I created a hook that ensures the overflow is removed on the destroyed() event.

zolrath:
Hmm, I suppose we have two directions we could go with this then.
1. We could add more logic for ensuring the body `overflow` classes are added/removed at the appropriate times in other circumstances.
2. Go more minimal with this and add `overscroll-none` to the `role="dialog"` div. This keeps the second scrollbar, but the browser won't automatically scroll the background page when the user hits the bottom of the modal.

zolrath:
I love the `has` option conceptually as it's not something we need to toggle, though it becomes a convention the user needs to understand since it happens without any obvious cause.

I know that it's [available everywhere but Firefox (and IE11)](https://caniuse.com/?search=css-has) but I've been waiting for that selector for years. Really happy to see it's made so much progress recently!

Even in situations where the selector isn't supported, it would just mean the current behavior of allowing scrolling returns.

chrismccord:
The modal component should have had `phi-remove`, so that takes care of the overflow failing to be reset on things like navigation. I went ahead and fixed that and applied @zolrath's changes in my own commit, (but added you as a co-author :) .Thanks!

❤️❤️❤️🐥🔥

PR #5105, opened 2022-11-13
Fix live_reload config tests

aaronrenner:
This fixes the following test failures we've been seeing on master

```
   1) test new with umbrella and defaults (Mix.Tasks.Phx.New.UmbrellaTest)
Error:      test/phx_new_umbrella_test.exs:32
     Assertion with =~ failed
     code:  assert file =~ "lib/#{@app}_web/(live|views)/.*(ex)"

  2) test new with defaults (Mix.Tasks.Phx.NewTest)
Error:      test/phx_new_test.exs:34
     Assertion with =~ failed
     code:  assert file =~ "lib/phx_blog_web/(live|views)/.*(ex)"
```

It appears this was changed in #5082 and we forgot to update the tests.

### DIFF
--- a/installer/test/phx_new_test.exs
+++ b/installer/test/phx_new_test.exs
@@ -149,8 +149,7 @@ defmodule Mix.Tasks.Phx.NewTest do
 
       assert_file("phx_blog/config/dev.exs", fn file ->
         assert file =~ "esbuild: {Esbuild,"
-        assert file =~ "lib/phx_blog_web/(live|views)/.*(ex)"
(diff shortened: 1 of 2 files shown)


aaronrenner:
The test failure seems to be coming from an unrelated websocket channel test that has been failing for several builds

```
  1) test with 2.0.0 serializer Phoenix.Socket.V2.JSONSerializer does not log user socket connect when disabled (Phoenix.Integration.WebSocketChannelsTest)
Error:      test/phoenix/integration/websocket_channels_test.exs:424
     Assertion with == failed
     code:  assert log == ""
```

Issue #5058, opened 2022-10-28
mix phx.gen.live with array field generates invalid tests


aidalgol:
### Environment

* Elixir version (elixir -v): 1.13.4
* Phoenix version (mix deps): 1.6.15
* Phoenix LiveView version (mix deps): 0.17.12
* PostgreSQL version: 14.5
* Operating system: NixOS 22.05
* Browsers you attempted to reproduce this bug on (the more the merrier): N/A
* Does the problem persist after removing "assets/node_modules" and trying again? Yes/no: N/A

### Actual behavior
From a fresh Phoenix LiveView project, adding a resource with an array field, using the `phx.gen.live` task, generates invalid tests.

After generating a new project and running `mix setup`,
```console
$ mix phx.gen.live Animals Pets pets name:string routines:array:string
```
Add live routes to the project's router.
```console
$ mix ecto.migrate
$ mix test
[…]
  1) test Index saves new pets (JunkWeb.PetsLiveTest)
     test/junk_web/live/pets_live_test.exs:26
     ** (RuntimeError) LiveView did not redirect
     code: |> follow_redirect(conn, Routes.pets_index_path(conn, :index))
     stacktrace:
       test/junk_web/live/pets_live_test.exs:42: (test)
[…]
```

### Expected behavior
All tests to pass.

leejarvis:
Looks like either the schema generator should not add array fields to `validate_required`, or [`type_to_default`](https://github.com/phoenixframework/phoenix/blob/e520e888ee8e67582322b02c7bd230cd954a87b6/lib/mix/phoenix/schema.ex#L280-L321) should do something like `{:array, _} -> ["some #{key}"]`

paridin:
I'm trying to solve this issue, but I need direction to solve `:array:integer` case.

The `:array:string` is covered by doing some like

`def live_form_value([]), do: ["option1"]`

Because [phx.gen.html:214](https://github.com/phoenixframework/phoenix/blob/master/lib/mix/tasks/phx.gen.html.ex#L214) injects that value, but [phx.gen.html:204](https://github.com/phoenixframework/phoenix/blob/master/lib/mix/tasks/phx.gen.html.ex#L204-L205) covers the `integer` part which I figure it's also an issue since it's handled as `datetime-local` type.

PR #5108, opened 2022-11-14
Solves array issues for phx generator. #5058

paridin:
Adds values for an array of strings and integers, which solve issue #5058


chrismccord:
❤️❤️❤️🐥🔥

PR #5113, opened 2022-11-16
Remove space in test name, referencing `current_user` instead of `current_ user`

linusdm:
Reference `current_user` instead of `current_ user` in test name.

Just a small typo I noticed.


Gazler:
Thanks <3

PR #5114, opened 2022-11-17
Fix a typo in the phx.gen.release docs

rspeicher:
Is this the least significant change ever committed to this wonderful project? It might just be! 😆


Gazler:
Still counts! Thanks <3

PR #5115, opened 2022-11-17
Remove leftover backticks in testing guide

optikfluffel:
The "Introduction to Testing" guide is kind of [broken in `1.7.0-rc.0`](https://hexdocs.pm/phoenix/1.7.0-rc.0/testing.html#view-tests) because of those extra backticks for which I couldn't find a matching pair.


Gazler:
Good spot. Thanks <3

PR #5116, opened 2022-11-18
Add dark mode for page#home.

wintermeyer:
I am pretty sure that @adamwathan will not be pleased by this attempt. But it is better than not having a dark mode at all. Maybe somebody with more knowledge and talent can fix this.

Bildschirm­foto 2022-11-18 um 08 02 25


chrismccord:
Thanks for this! The bigger work would be making all the components in `core_components.ex` support dark mode, which are the ones used throughout the UI/generators. I'm not sure if we'll ship the extra classes or not, but at the very least the community could have a drop in copy-paste replacements for `core_components.ex`. If someone would like to pick this up, please let us know!

PR #5117, opened 2022-11-18
Add empty value to select input prompt option

shamanime:
Without this the input sends the prompt text which is not desirable.
The current behaviour won't trigger an empty validation.


josevalim:

PR #5118, opened 2022-11-19
Update contexts.md

GuiHeurich:
`get_product!` function should use `get!` instead of `get` otherwise `assert_raise Ecto.NoResultsError, fn -> Catalog.get_product!(product.id) end` test will fail.


josevalim:

PR #5119, opened 2022-11-20
Fix inconsistency in controllers guide

randycoulman:
The `Overriding rendering formats` section of the controllers guide talks about accepting `json`, but the example code was using `text.

This updates the example code to also use `json`. It also wraps the output with backticks to make it more clear that it's a literal response.

### DIFF
--- a/guides/controllers.md
+++ b/guides/controllers.md
@@ -263,7 +263,7 @@ defmodule HelloWeb.Router do
   use HelloWeb, :router
 
   pipeline :browser do
-    plug :accepts, ["html", "text"]
+    plug :accepts, ["html", "json"]
     plug :fetch_session
     plug :fetch_live_flash
     plug :put_root_layout, {HelloWeb.LayoutView, :root}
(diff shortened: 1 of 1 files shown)


guides/controllers.md
```diff
@@ -273,7 +273,7 @@ defmodule HelloWeb.Router do
 ...
 ```
 
-If we go to [`http://localhost:4000/?_format=text`](http://localhost:4000/?_format=text), we will see %{"message": "this is some JSON"}.
+If we go to [`http://localhost:4000/?_format=json`](http://localhost:4000/?_format=json), we will see `%{"message": "this is some JSON"}`.
```
josevalim, line 276: I think it makes more sense to use text because json would typically require a separate pipeline. Maybe we use text and change what is rendered? 🤔

randycoulman: That argument makes a lot of sense to me, but I'm not sure how best to rewrite the entire section to match that. Everything in the section is about rendering JSON, including defining a `PageJSON` module, etc, so I was just trying to make the last part of the example match the rest.

Does `put_view/2` work with a `text:` format? If so, is it sufficient to essentially replace `json` with `text` in the whole section and have it all be correct?

josevalim: Ah, good point. @chrismccord, your call. :)

chrismccord:
❤️❤️❤️🐥🔥

PR #5120, opened 2022-11-20
Add reference to Components guide

randycoulman:
In the `A new view` section of the `Request life-cycle`, the reader was being taken directly to the `Phoenix.Component` API documentation.

However, there is now a separate guide for components and HEEx templates (which then links to the API docs).

This links the reader to the components guide instead of the API documentation.


josevalim:

PR #5121, opened 2022-11-20
Fix dead reference to old Views guide

randycoulman:
The Controllers guide was still referencing the Views guide, which no longer exists.

It appears to have been renamed to `components.md` (and significantly updated) in cd049a3325313280530315af9014fcc359d940f4, so this updates the reference accordingly.


josevalim:

PR #5125, opened 2022-11-22
If a schema has no properties then the form functions fail to match

krainboltgreene:
Found while using the generated form components (in 1.7) on a schemas that just references other schemas. Willing to change as needed or close if the team feels like this isn't the right change to make.


chrismccord:
It's not clear under which circumstances you'd have a form without form inputs? Can you elaborate on where this change would be necessary? Thanks!

krainboltgreene:
Sure so I'm building an app to track my cluster headaches. A cluster headache itself has no properties, but it has entries, medicines, etc.

PR #5128, opened 2022-11-24
Controller docs. Add json to accepts instead of text

Ivor:
Noticed that the guide mentions adding JSON but the example adds TEXT.


zimt28:
This is already being discussed here: #5119

Ivor:
Closing because issue yas already been raised.

PR #5133, opened 2022-11-27
docs: typo: replace `text` by `json`

kingshalaby1:
the docs explains adding `json` format and uses `PageJSON` module, but the code example adds the format `text` 
for `text` format to work, we need to add it in `HelloWeb` too
```
def controller do
    quote do
      use Phoenix.Controller,
        namespace: EmrWeb,
        formats: [:html, :json, :text],
        layouts: [html: HelloWeb.Layouts]
   # ....
  end
end
```
I've tried it and it works, but this part is not mentioned in the docs, so I kept with `json`

### DIFF
--- a/guides/controllers.md
+++ b/guides/controllers.md
@@ -263,7 +263,7 @@ defmodule HelloWeb.Router do
   use HelloWeb, :router
 
   pipeline :browser do
-    plug :accepts, ["html", "text"]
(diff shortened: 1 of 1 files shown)


kingshalaby1:
just found a similar PR https://github.com/phoenixframework/phoenix/pull/5119
closing this as being a duplicate

PR #5138, opened 2022-11-29
Add entries to changelog for v1.6.14 v1.6.15

eldano:
These entries appear only in the specific tags but not in master


chrismccord:
❤️❤️❤️🐥🔥

PR #5147, opened 2022-12-03
Extract ErrorHelpers from CoreComponent

### LINKED ISSUE #5136: mix phx.gen.json fails on --no-html app
It assumes CoreComponents is available but it may not be. Maybe we want to check if it is available and, if not, define our own version inline?

Alternatively just always define the version inline.

50kudos:
Fix #5136 based on assumption that it breaks like the following screenshot:

core_components

So that `mix phx.gen.json` work with `mix phx.new --no-html` app because ErrorHelpers is always generated.

Note: The screenshot is `mix test` ran on [`mix phx.gen.json` on top of `mix phx.new --no-html` app] code.

---
Let me know if it breaks anything else or any convention I should follow. Thanks!


chrismccord:
Fixed on master. Thank you!

PR #5151, opened 2022-12-03
Fix slack link in homepage template

benjreinhart:
Looks like the heroku links stopped working. Happy to change this to another link if available or close the PR if the link comes back online.


chrismccord:
The link is still valid. Thanks!

PR #5052, opened 2022-10-23
Add Swoosh.Finch

christopherlai:
Adds the usage of the `Swoosh.Finch` adapter for newly generated projects. Related to #4876

PR #5166, opened 2022-12-14
Correct the swoosh api client configuration

woodward:
The current swoosh API client config code doesn't work on fly.io, and this fixes it.  After some experimentation, I finally figured out that this works.  Note that the name in for Finch in application.ex is <%= @app_module %>.Finch, and so the finch_name on the adapter needs to be set to this same value in prod.exs.

Another solution is to just call the Finch service "Swoosh.Finch" in application.ex, and then you do not have to give a name for the api client in prod.exs, but the changes I'm suggesting seem more in line with the current code. 

Thanks!

-- Greg


Gazler (approved):
Looks like this was an oversight between the reviews in the initial PR.

https://github.com/phoenixframework/phoenix/pull/5052

chrismccord:
❤️❤️❤️🐥🔥

PR #5170, opened 2022-12-19
fix: Extra newline in textarea component causes new textareas to be non-empty.

jadengis:
The `textarea` input template in `core_components.ex` has an additional newline character as a child of the `<textarea>` block. This causes the textarea to always be rendered with this additional character, even when a value is supplied (like with a changeset). Removing this line fixes the issue for new generated projects.


benjreinhart:
Hey @jadengis, heads up I think there is a [previous PR](https://github.com/phoenixframework/phoenix/pull/5091) with some discussion around it on this topic.

PR #5171, opened 2022-12-20
Improve docs on custom errors from `Socket.connect/3`

nathanl:


lib/phoenix/endpoint.ex
```diff
@@ -866,7 +866,10 @@ defmodule Phoenix.Endpoint do
       the error handler must be a MFA tuple that receives a `Plug.Conn`, the
       error reason, and returns a `Plug.Conn` with a response. For example:
 
-          error_handler: {MySocket, :handle_error, []}
+          socket "/socket", MySocket,
+              websocket: [
+                error_handler: {MySocket, :handle_error, []}
```
nathanl, line 871: Note: this does introduce a compile time dependency that is flagged by `mix xref graph --label compile-connected --fail-above 0`. Using another module, like `error_handler: {SocketErrorHandler, :handle_error, []}`, will fix it. Not sure if this example should be changed accordingly or not.

chrismccord:
❤️❤️❤️🐥🔥

PR #5172, opened 2022-12-20
Fixes flaky test case - test/mix/tasks/phx.gen.json_test.exs:166

grzuy:
Hi!

Test cases in `phx.gen.json_test.exs` are isolated given gen tasks are run in different `tmp` subdirectories, however it seems this particular recently-added test case having a run time compilation of a `PhoenixWeb.CoreComponents` module is leaking to other test case given the module is not purged/deleted.

In particular it seems to affect the test case titled `"with json --web namespace generates namespaced web modules and directories"` in the same test file.

So the intermittence in the failure depends on the random order of the test case run.

Failure can be reproduced consistently with some seeds. E.g.

```
$ mix test test/mix/tasks/phx.gen.json_test.exs --seed 911231
```

Also, this is the failure that is making `master` last commit and a few pull requests fail: https://github.com/phoenixframework/phoenix/actions/runs/3696703936/jobs/6260760004.

### DIFF
--- a/test/mix/tasks/phx.gen.json_test.exs
+++ b/test/mix/tasks/phx.gen.json_test.exs
@@ -340,7 +340,7 @@ defmodule Mix.Tasks.Phx.Gen.JsonTest do
       end
       """)
 
-      Code.compile_file("lib/phoenix_web/components/core_components.ex")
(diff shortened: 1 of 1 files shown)


josevalim:

PR #5175, opened 2022-12-21
Proposal: alias Phoenix.HTML in CoreComponets to avoid Credo warnings.

leandrocp:
That would improve the experience as Credo is widespread used :)


josevalim:
I don't personally think the code is better with this change, perhaps it is not a warning that should be enabled by default. :)

PR #5178, opened 2022-12-23
feat: Add method offAll for Socket to remove all callback registrations

rnike:
Add a method for socket to remove all the callback registrations

usage
```js
socket.offAll();
```

### DIFF
--- a/assets/js/phoenix/socket.js
+++ b/assets/js/phoenix/socket.js
@@ -480,6 +480,15 @@ export default class Socket {
   }
 
   /**
+   * Removes all callback registrations.
+   */
+  offAll(){
+    for(let key in this.stateChangeCallbacks){
+      this.stateChangeCallbacks[key] = []
+    }
+  }
+
+  /**
    * Initiates a new channel for the given topic
    *
    * @param {string} topic
--- a/assets/test/socket_test.js
+++ b/assets/test/socket_test.js
@@ -392,6 +392,36 @@ describe("with transports", function(){
     })
   })
 
+  describe("offAll", function(){
+    it("removes all the callback registrations from the socket", function(){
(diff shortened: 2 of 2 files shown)


benjreinhart:
Curious, what is your use case for removing all callbacks?

If this functionality is accepted, wondering if it's nicer to use the existing `off` method with no arguments?

```js
// remove individual callback
socket.off(ref)

// remove everything
socket.off()
```

rnike:
@benjreinhart To match our application's lifecycle, we are opening and closing the socket based on the user's interaction, when closing it we clean up the socket instance.

we are currently doing like this

```js
let socket = null;
let channel = null;
let refs = null;

function disconnect() {
  if (channel) {
    channel.disconnect();
    channel = null;
  }

  if (socket) {
    refs && socket.off(refs);
    socket.disconnect();
    socket = null;
    refs = null;
  }
}

function connect(channelName, token) {
  disconnect();

  socket = new Socket('url', {
    params: {
      token,
    },
  });

  refs = [
    socket.onMessage(() => {
      // ...
    }),
    socket.onOpen(() => {
      // ...
    }),
    socket.onError(() => {
      // ...
    }),
    socket.onClose(() => {
      // ...
    }),
  ];

  socket.connect();
  channel = socket.channel(channelName);
  channel.join();
}
```
I'm wondering if it is necessary to clean up before `socket.disconnect()`, is there any document about it?

chrismccord:
Would it be better in our case then to instantiate a new socket instead of trying to clear out the old one?

chrismccord:
I'm not convinced yet this is needed because you can rebuild the socket state as needed when you want to teardown and reup the connection. You can also write a simple wrapper that collects the refs yourself and calls off. Thanks!

PR #5180, opened 2022-12-25
doc: renaming :index to :home in routing.md

kingshalaby1:
altering routing docs, as the generated PageController has its only action changed to :home


josevalim:

PR #5183, opened 2022-12-28
Configurable csrf token session key

cjbottaro:
Using a custom `:session_key` with the `:protect_from_forgery` plug causes the session to not be loaded by a Phoenix.Socket.

This PR allows you to specify the session key in your Endpoint.

```elixir
# Router
plug :protect_from_forgery, session_key: "_my_csrf_token"

# Endpoint
socket "/live", Phoenix.LiveView.Socket, websocket: [
  connect_info: [
    session: Keyword.merge(@session_options,
      csrf_token_key:  "_my_csrf_token"
    )
  ]
]
```

I added docs and a test.


cjbottaro:
Shoot, I think I branched off of v1.6.15. Trying to fix that...

lib/phoenix/endpoint.ex
```diff
@@ -814,6 +814,10 @@ defmodule Phoenix.Endpoint do
           config in runtime `{MyAppWeb.Auth, :get_session_config, []}`. Otherwise
           the session will be `nil`.
 
+        * `{:csrf_token_session_key, csrf_token_session_key}` - Name of the session
```
josevalim, line 817: I would move this inside the `session_config`, as this is session specific. :) And we can call it `csrf_token_key`.

test/phoenix/socket/transport_test.exs
```diff
@@ -288,5 +288,70 @@ defmodule Phoenix.Socket.TransportTest do
...
+    end
+
+    Application.put_env :phoenix, __MODULE__.EndpointWithCustomCSRFToken,
+      force_ssl: [],
+      url: [host: "host.com"],
+      check_origin: ["//endpoint.com"],
+      secret_key_base: @secret_key_base
+
+    test "loads the session with custom :csrf_token_session_key" do
+      {:ok, pid} = EndpointWithCustomCSRFToken.start_link()
```
josevalim, line 323: Instead of defining a new endpoint, you can try to emulate routes in the previous endpoint. Something like, set this particular session key based on `conn.path_info`.

lib/phoenix/endpoint.ex
cjbottaro: I'm having some trouble with this. The original `session_config` is lost because `Transport.load_config` calls `init_session/1` which in turn calls `Plug.Session.Store.init/1` which is not guaranteed to pass through any options it doesn't care about.

I thought about putting `:csrf_token_key` back in, but `Plug.Session.Store.init/1` doesn't even have to return a keyword list or map.

I ended up wrapping the return of `init_session/1`. I think it's kind of hacky though, what do you think?

test/phoenix/socket/transport_test.exs
cjbottaro: I can't figure out how to do this. The CSRF session token seems to be generated by a `before_send` callback defined on the `conn`. So I couldn't figure out how to make my own plug in the existing Endpoint that removes `"_csrf_token"` and adds my own based on path info.

josevalim: Hi @cjbottaro! I don't want to leave you hanging but I don't have answers for your questions right now and I would have to play with the code, but lacking time to do so at the moment, sorry.

cjbottaro: No worries at all. I'm just using my own branch for now, so no rush... 👍

Schultzer: What about something like this:

```elixir
  defmodule Endpoint do
    use Phoenix.Endpoint, otp_app: :phoenix
    @session_config [
      store: :cookie,
      key: "_hello_key",
      signing_salt: "change_me"
    ]
    def session_config, do: @session_config
    plug Plug.Session, @session_config
    plug :fetch_session
    plug :put_csrf
    plug :put_session

    defp put_csrf(conn, _) do
      Plug.CSRFProtection.call(conn, session_key: Enum.join(conn.path_info, "_"))
    end

    defp put_session(conn, _) do
      conn
      |> put_session(:from_session, "123")
      |> send_resp(200, Plug.CSRFProtection.get_csrf_token())
    end
  end

```

cjbottaro: @Schultzer Cool, thanks, I got that working with a little fiddling. 👍

Do you have any opinions on the implementation of this feature? It works, but I feel it is hacky. The issue is that Phoenix doesn't store the _original_ session store options anywhere. It only stores the options after they have passed through `Plug.Session.Store.init/1`.

```elixir
orig_options = [csrf_token_key: "custom_key", ...]
init_options = MySessionStore.init(orig_options)
# ^ is what Phoenix stores, which isn't required to have the orig_options at all.
```

My hack does something like this...
```elixir
orig_options = [csrf_token_key: "custom_key", ...]
init_options = MySessionStore.init(orig_options)
{orig_options, init_options}
# ^ Phoenix stores a tuple of orig_options and init_options,
# that way we have access to csrf_token_key downstream.

cc @josevalim

Comment (cjbottaro) [reply test/phoenix/socket/transport_test.exs]: Actually, now that I articulated that though and coded it... it seems like a good solution, and not as hacky as I did it before. Returning both session_config and the results of init is more flexible in the long run, and it's conceptually simple.

**Comment by cjbottaro** [comment]
I think this PR is ready to go. The tests are much better thanks to @Schultzer 's help. And I think the implementation is simple and easy to understand at this point.

File: lib/phoenix/endpoint.ex [hunk 3]
```diff
@@ -835,6 +839,8 @@ defmodule Phoenix.Endpoint do
             ]
       ```
 
+
+
```
josevalim, line 843:
```suggestion
```

lib/phoenix/socket/transport.ex
```diff
@@ -258,11 +258,13 @@ defmodule Phoenix.Socket.Transport do
...
   end
 
+  # The original session_config is returned in addition to init value so we can
+  # access special config like :csrf_token_key downstream.
   defp init_session(session_config) when is_list(session_config) do
     key = Keyword.fetch!(session_config, :key)
     store = Plug.Session.Store.get(Keyword.fetch!(session_config, :store))
     init = store.init(Keyword.drop(session_config, [:store, :key]))
-    {key, store, init}
+    {key, store, {session_config, init}}
```
josevalim, line 267: We only need the `csrf_token_key`, so what if we only pass it here?

```suggestion
    csrf_token_key = Keyword.get(session_config, :csrf_token_key, "_csrf_token")
    {key, store, {csrf_token_key, init}}
```

josevalim:
Awesome job @Schultzer  and @cjbottaro ! This looks great, only some minor nits.

josevalim:

PR #5186, opened 2023-01-02
Fixes missing :references generation to phx.gen.embedded

Morzaram:
As mentioned in the [docs](https://hexdocs.pm/phoenix/1.7.0-rc.0/Mix.Tasks.Phx.Gen.Embedded.html#module-attributes) `:references` should be generated when calling it within
 `mix phx.gen.embedded Blog.Comment  author_id:references:author` 
but it only creates an empty schema. I've added the population of `:references` as well as a new test. 

I'm not sure if there needs to be anything else since it's an embedded schema and not an actual schema which is why I left out the primary_key and foreign_key assignments. 

Hope this does it!


chrismccord:
❤️❤️❤️🐥🔥

PR #5188, opened 2023-01-03
bugfix: flash component's phx-click hide should go to @id, not #flash

benswift:
`#flash` is the default `id`, so this will work in that case, but it seems as though the hide message should actually go to `@id`?


chrismccord:
❤️❤️❤️🐥🔥

PR #5190, opened 2023-01-04
touchup styles, add navigation links

bemesa21:
- Added some CSS classes that were missing in some places. 
- The navigation links were not unified between the --live option and the --no--live option, so I added them to both.


chrismccord:
❤️❤️❤️🐥🔥

PR #5191, opened 2023-01-05
Fix sentence

simonelnahas:
The order is not different.

### DIFF
--- a/guides/routing.md
+++ b/guides/routing.md
@@ -98,7 +98,7 @@ DELETE  /users/:id       HelloWeb.UserController :delete
 ...
 ```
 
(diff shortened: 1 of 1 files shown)


guides/routing.md
```diff
@@ -98,7 +98,7 @@ DELETE  /users/:id       HelloWeb.UserController :delete
 ...
 ```
 
-This is the standard matrix of HTTP verbs, paths, and controller actions. For a while, this was known as RESTful routes, but most consider this a misnomer nowadays. Let's look at them individually, in a slightly different order.
+This is the standard matrix of HTTP verbs, paths, and controller actions. For a while, this was known as RESTful routes, but most consider this a misnomer nowadays. Let's look at them individually in the same order.
```
benjreinhart, line 101: Personally, I would just omit anything about "order" here since it doesn't add much and is brittle anytime docs are updated.

> Let's look at them individually:

simonelnahas: Agreed.

PR #5203, opened 2023-01-08
Improved function component discussion.

CharlesIrvineKC:
Improved function component discussion. Removed reference to out-of-place reference to title/2 function component. Since I'm new to Phoenix, someone more knowledgable should look this over and perhaps elaborate a bit.


guides/components.md
```diff
@@ -52,7 +52,21 @@ defmodule HelloWeb.HelloHTML do
 end
 ```
 
-We declared the attributes we accept via `attr` provided by `Phoenix.Component`, then we defined our `title/1` function which returns the HEEx template. When we reload our home page, we should see our new title. Since templates are compiled inside the view, we can invoke the view function simply as `<.greet messenger="..." />`, but we can also type `<HelloWeb.HelloHTML.greet messenger="..." />` if the component was defined elsewhere.
+We declared the attributes we accept via `attr` provided by `Phoenix.Component`, then we defined our `greet/1` function which returns the HEEx template. 
+
+Next we need to recode `show.html.heex` to look like:
```
chrismccord, line 57:
```suggestion
Next we need to update `show.html.heex`:
```

guides/components.md
```diff
@@ -52,7 +52,21 @@ defmodule HelloWeb.HelloHTML do
...
 ```
 
-We declared the attributes we accept via `attr` provided by `Phoenix.Component`, then we defined our `title/1` function which returns the HEEx template. When we reload our home page, we should see our new title. Since templates are compiled inside the view, we can invoke the view function simply as `<.greet messenger="..." />`, but we can also type `<HelloWeb.HelloHTML.greet messenger="..." />` if the component was defined elsewhere.
+We declared the attributes we accept via `attr` provided by `Phoenix.Component`, then we defined our `greet/1` function which returns the HEEx template. 
+
+Next we need to update `show.html.heex`:
+
+```elixir
+<section>
+    <.greet messenger={@messenger} />
```
chrismccord, line 61:
```suggestion
  <.greet messenger={@messenger} />
```

chrismccord:
❤️❤️❤️🐥🔥

PR #5205, opened 2023-01-10
Update injector.ex

joshchernoff:
Make else block render.


joshchernoff:
I'm not sure why this worked to get it to render but for what ever reason it looked like I am requited to use `<%= else` I'm gonna try one more time to confirm but I'm was not aware that `<%= else` was required

joshchernoff:
Ok that was a fluke on me end my bad, nothing to see here :facepalm:

PR #5216, opened 2023-01-14
fix: formatting for core_components generated file

joshchernoff:
This should address the integration failing tests as a result of formatting.


joshchernoff:
looks like a noop, closing as not needed.

PR #5221, opened 2023-01-16
Fix unused variable on topbar callback

drselump14:
The typescript warns of the unused variable `info` on these two lines. It's better to add an underscore to silent this warning.


josevalim:

PR #5223, opened 2023-01-18
Use the new input component in the Context guide

wceolin:
The instructions in the [Contexts guide](https://hexdocs.pm/phoenix/1.7.0-rc.2/contexts.html) aren't using the new `input` component. This PR updates the guide to use it.

It also looks like we're not generating the `form.html.heex` file anymore. So, for now, I've also updated the instructions to mention we need to update both `edit.html.heex` and `new.html.heex`.

However, I think it may be a good idea to bring back the `form` file to avoid duplicating those inputs.

Feel free to make any changes you see fit to this PR.

### DIFF
--- a/guides/contexts.md
+++ b/guides/contexts.md
@@ -493,30 +493,27 @@ Next, let's expose our new feature to the web by adding the category input to ou
 defmodule HelloWeb.ProductHTML do
   use HelloWeb, :html
 
-  def category_select(f, changeset) do
+  def category_select(changeset) do
     existing_ids =
       changeset
       |> Ecto.Changeset.get_change(:categories, [])
       |> Enum.map(& &1.data.id)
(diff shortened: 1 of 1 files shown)


guides/contexts.md
```diff
@@ -493,30 +493,27 @@ Next, let's expose our new feature to the web by adding the category input to ou
 defmodule HelloWeb.ProductHTML do
   use HelloWeb, :html
 
-  def category_select(f, changeset) do
+  def category_select(changeset) do
```
leandrocp, line 496: Just nitpicking but it could be renamed to `category_options` with those changes:

```suggestion
  def category_options(changeset) do
```

wceolin: Makes sense to me.

PR #5225, opened 2023-01-20
Change `assets.setup` to run esbuild & tailwind

wojtekmach:
There are two reasons for this change:

1. This way when we launch the Web app for the first time the assets
   would have been built and so we avoid the "flash" from the page
   without assets to the one with.

2. I think if there are any assets build errors, they should be a little
   bit easier to debug in a terminal window that runs `$ mix setup` than
   in the terminal window that runs `$ mix phx.server`, the latter
   containing much more log output and such.

### DIFF
--- a/installer/templates/phx_single/mix.exs
+++ b/installer/templates/phx_single/mix.exs
@@ -71,10 +71,10 @@ defmodule <%= @app_module %>.MixProject do
       "ecto.reset": ["ecto.drop", "ecto.setup"],
       test: ["ecto.create --quiet", "ecto.migrate --quiet", "test"]<% end %><%= if @assets do %>,
       "assets.setup": [
-        "tailwind.install --if-missing",
         "esbuild.install --if-missing",
-        "tailwind default",
-        "esbuild default"
+        "tailwind.install --if-missing",
+        "esbuild default",
+        "tailwind default"
       ],
       "assets.deploy": ["tailwind default --minify", "esbuild default --minify", "phx.digest"]<% end %>
     ]
--- a/installer/templates/phx_single/mix.exs
+++ b/installer/templates/phx_single/mix.exs
@@ -70,7 +70,12 @@ defmodule <%= @app_module %>.MixProject do
       "ecto.setup": ["ecto.create", "ecto.migrate", "run priv/repo/seeds.exs"],
       "ecto.reset": ["ecto.drop", "ecto.setup"],
       test: ["ecto.create --quiet", "ecto.migrate --quiet", "test"]<% end %><%= if @assets do %>,
-      "assets.setup": ["tailwind.install --if-missing", "esbuild.install --if-missing"],
+      "assets.setup": [
+        "tailwind.install --if-missing",
+        "esbuild.install --if-missing",
+        "tailwind default",
+        "esbuild default"
+      ],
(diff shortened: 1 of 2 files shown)


installer/templates/phx_single/mix.exs
```diff
@@ -70,7 +70,12 @@ defmodule <%= @app_module %>.MixProject do
       "ecto.setup": ["ecto.create", "ecto.migrate", "run priv/repo/seeds.exs"],
       "ecto.reset": ["ecto.drop", "ecto.setup"],
       test: ["ecto.create --quiet", "ecto.migrate --quiet", "test"]<% end %><%= if @assets do %>,
-      "assets.setup": ["tailwind.install --if-missing", "esbuild.install --if-missing"],
+      "assets.setup": [
+        "tailwind.install --if-missing",
+        "esbuild.install --if-missing",
+        "tailwind default",
+        "esbuild default"
```
wojtekmach, line 77: Does the order matter? If it doesn't my gut tells me esbuild should run first as it seems lower level. lmk!

```suggestion
        "esbuild.install --if-missing",
        "tailwind.install --if-missing",
        "esbuild default",
        "tailwind default"
```

installer/templates/phx_umbrella/apps/app_name_web/mix.exs
```diff
@@ -62,7 +62,12 @@ defmodule <%= @web_namespace %>.MixProject do
     [
       setup: ["deps.get"<%= if @assets do %>, "assets.setup"<% end %>]<%= if @ecto do %>,
       test: ["ecto.create --quiet", "ecto.migrate --quiet", "test"]<% end %><%= if @assets do %>,
-      "assets.setup": ["tailwind.install --if-missing", "esbuild.install --if-missing"],
+      "assets.setup": [
+        "tailwind.install --if-missing",
+        "esbuild.install --if-missing",
+        "tailwind default",
+        "esbuild default"
```
wojtekmach, line 69:
```suggestion
        "esbuild.install --if-missing",
        "tailwind.install --if-missing",
        "esbuild default",
        "tailwind default"
```

Gazler (approved):

installer/templates/phx_umbrella/apps/app_name_web/mix.exs
```diff
@@ -60,9 +60,10 @@ defmodule <%= @web_namespace %>.MixProject do
   # See the documentation for `Mix` for more info on aliases.
   defp aliases do
     [
-      setup: ["deps.get"<%= if @assets do %>, "assets.setup"<% end %>]<%= if @ecto do %>,
+      setup: ["deps.get"<%= if @assets do %>, "assets.setup", "assets.build"<% end %>]<%= if @ecto do %>,
       test: ["ecto.create --quiet", "ecto.migrate --quiet", "test"]<% end %><%= if @assets do %>,
-      "assets.setup": ["tailwind.install --if-missing", "esbuild.install --if-missing"],
+      "assets.setup": ["esbuild.install --if-missing", "tailwind.install --if-missing"],
+      "assets.build": ["esbuild default", "tailwind default"],
```
josevalim, line 66: Sorry, but assets.deploy lists tailwind first. Should we make things consistent? :P

josevalim: Actually, i think there was a reason to have tailwind first, but i can't recall it right now.

josevalim:

Issue #4831, opened 2022-05-20
Channel `handle_in/3` to emit more standardized telemetry events


RudolfMan:
Hi there! I was working on the instrumentation of a project, particularly on the incoming events via channels. And apparently, learned that `handle_in` only emits `[:phoenix, :channel_handeled_in]` telemetry event after success.

I'd like to propose standardized set of `:telemetry` events for channel `handle_in/3` callback, similar to those emitted by `Phoenix.Router`:
 - `[:phoenix, :channel_handele_in, :start]`
 - `[:phoenix, :channel_handele_in, :stop]`
 - `[:phoenix, :channel_handele_in, :exception]`
 
As if they were emitted by `:telemetry.span/3`. And, probably, "soft-deprecate" the existing `[:phoenix, :channel_handeled_in]` in favor of new events.

That would provide with better info for observability.

I put together a code change, in case it's helpful: https://github.com/phoenixframework/phoenix/compare/master...RudolfMan:handle_in_telemetry_events

aselder:
I like it!!!

jeregrine:
I think this makes sense, open up a PR and we can get more feedback on it!

RudolfMan:
Closing in favor of https://github.com/phoenixframework/phoenix/pull/5229

PR #5229, opened 2023-01-23
Channel `handle_in/3` to emit more standardized telemetry events

RudolfMan:
_Opening a PR for https://github.com/phoenixframework/phoenix/issues/4831 as suggested by @jeregrine_

Currently `handle_in` only emits `[:phoenix, :channel_handeled_in]` telemetry event after success.

I'd like to propose standardized set of `:telemetry` events for channel `handle_in/3` callback, similar to those emitted by `Phoenix.Router`:
 - `[:phoenix, :channel_handele_in, :start]`
 - `[:phoenix, :channel_handele_in, :stop]`
 - `[:phoenix, :channel_handele_in, :exception]`
 
As if they were emitted by `:telemetry.span/3`. And, probably, "soft-deprecate" the existing `[:phoenix, :channel_handeled_in]` in favor of new events.

That would provide with better info for observability.

### DIFF
--- a/lib/phoenix/channel/server.ex
+++ b/lib/phoenix/channel/server.ex
@@ -313,12 +313,26 @@ defmodule Phoenix.Channel.Server do
         %Message{topic: topic, event: event, payload: payload, ref: ref},
         %{topic: topic} = socket
       ) do
-    start = System.monotonic_time()
-    result = socket.channel.handle_in(event, payload, put_in(socket.ref, ref))
-    duration = System.monotonic_time() - start
     metadata = %{ref: ref, event: event, params: payload, socket: socket}
-    :telemetry.execute([:phoenix, :channel_handled_in], %{duration: duration}, metadata)
-    handle_in(result)
+    start = System.monotonic_time()
+    measurements = %{system_time: System.system_time()}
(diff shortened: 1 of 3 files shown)


chrismccord:
I think we had conversations that folks doing more complex things will necessarily be measuring themselves? But it has been awhile. For example, highly latent operations would be sent to another process, so it's not clear if we need two events here. Do you remember anything in particular @josevalim ?

josevalim:
Correct. The events were added as simple hooks to aid logging only.

RudolfMan:
> highly latent operations would be sent to another process

Just in case, it might be not the best practice, but in the project I worked at that time, we were often using channels in "request/response style messaging" as shown in the example in docs https://hexdocs.pm/phoenix/Phoenix.Channel.html#module-replies

Surely, we wrapped the internals of `handle_in` to `Opentelemetry.Tracer.with_span` and I was wondering if I could attach telemetry handlers similarly as how [`opentelemetry_phoenix`](https://hexdocs.pm/opentelemetry_phoenix/OpentelemetryPhoenix.html) does with `[:phoenix, :endpoint]` events.

PR #5231, opened 2023-01-24
fix pluralization in live generator delete template

rbritom:
There is a small error in the template of the Liveview generator.

```elixir
test "deletes post in listing", %{conn: conn, post: post} do
  {:ok, index_live, _html} = live(conn, ~p"/posts")

  assert has_element?(index_live, "#post-#{post.id}")
  assert index_live |> element("#posts-#{post.id} a", "Delete") |> render_click()
  refute has_element?(index_live, "#post-#{post.id}")
end
```
this fails because "#post-#{post.id}" was never there in the first place, yet in `master` template this is passing. Making the selectors match fixes the false positive.


josevalim:

PR #5232, opened 2023-01-24
Fix CHANGELOG.md paragraph about route helpers

BartOtten:
Updated the changelog to reflect the new default of not generating helpers.

https://github.com/phoenixframework/phoenix/blob/master/installer/templates/phx_single/lib/app_name_web.ex

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -32,8 +32,8 @@ will work properly and warn on new compilations, but changes to the router file
 will not issue new warnings.*
 
 This feature replaces the `Helpers` module generated in your Phoenix router, but helpers
-will continue to work and be generated. You can disable router helpers by passing the
-`helpers: false` option to `use Phoenix.Router`.
+will continue to work when enabled. You can enable router helpers by passing the
+`helpers: true` option to `use Phoenix.Router`.
 
 ## 1.7.0-rc.2 (2023-01-13)
 


chrismccord:
Helpers are generated by default, for backwards compatibility. You have to explicitly opt out with `helpers: false`. Leaving of ff the option, such as all previously generated apps < 1.7 will default to true. Thanks!

BartOtten:
Hi @chrismccord, maybe I am ignorant but at this moment helpers are _not_ generated when creating a new app with RC.2. The flag is set to false by default so 'true' would be opt-in. Although this PR does not cover the problem as the current paragraph seems to match your explanation, the template does not.

```
mix phx.new --version
Phoenix installer v1.7.0-rc.2

mix phx.new foo

grep -R "helpers:" foo
foo/lib/foo_web.ex:      use Phoenix.Router, helpers: false
```

```
Generated foo app
  GET  /                                      FooWeb.PageController :home
  GET  /dev/dashboard                         Phoenix.LiveDashboard.PageLive :home
  GET  /dev/dashboard/:page                   Phoenix.LiveDashboard.PageLive :page
  GET  /dev/dashboard/:node/:page             Phoenix.LiveDashboard.PageLive :page
  *    /dev/mailbox                           Plug.Swoosh.MailboxPreview []
  WS   /live/websocket                        Phoenix.LiveView.Socket
  GET  /live/longpoll                         Phoenix.LiveView.Socket
  POST  /live/longpoll                         Phoenix.LiveView.Socket
  ```

chrismccord:
Existing projects upgrading to 1.7 will necessarily have no `:helpers` option because it didn't exist, and therefore `use Phoenix.Router` will default to `helpers: true`, so existing projects are all good with their use of helpers on 1.7 :)

PR #5234, opened 2023-01-24
Add guard to skip map attributes when generating heex templates

codeanpeace:
Skip over map data type attributes from jsonb columns that currently cause the mix tasks `phx.gen.live` and `phx.gen.html` to generate broken heex templates. Here's an [old stackoverflow question](https://stackoverflow.com/questions/35549618/mix-phoenix-gen-html-with-map-field-type) on this.

### DIFF
--- a/priv/templates/phx.gen.html/index.html.heex
+++ b/priv/templates/phx.gen.html/index.html.heex
@@ -7,7 +7,7 @@
   </:actions>
 </.header>
 
-<.table id="<%= schema.plural %>" rows={@<%= schema.collection %>} row_click={&JS.navigate(~p"<%= schema.route_prefix %>/#{&1}")}><%= for {k, _} <- schema.attrs do %>
+<.table id="<%= schema.plural %>" rows={@<%= schema.collection %>} row_click={&JS.navigate(~p"<%= schema.route_prefix %>/#{&1}")}><%= for {k, v} when not is_map(v) <- schema.attrs do %>
   <:col :let={<%= schema.singular %>} label="<%= Phoenix.Naming.humanize(Atom.to_string(k)) %>"><%%= <%= schema.singular %>.<%= k %> %></:col><% end %>
   <:action :let={<%= schema.singular %>}>
     <div class="sr-only">
--- a/priv/templates/phx.gen.html/show.html.heex
+++ b/priv/templates/phx.gen.html/show.html.heex
@@ -8,7 +8,7 @@
   </:actions>
 </.header>
 
-<.list><%= for {k, _} <- schema.attrs do %>
+<.list><%= for {k, v} when not is_map(v) <- schema.attrs do %>
   <:item title="<%= Phoenix.Naming.humanize(Atom.to_string(k)) %>"><%%= @<%= schema.singular %>.<%= k %> %></:item><% end %>
 </.list>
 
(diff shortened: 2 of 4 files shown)


priv/templates/phx.gen.html/index.html.heex
```diff
@@ -7,7 +7,7 @@
   </:actions>
 </.header>
 
-<.table id="<%= schema.plural %>" rows={@<%= schema.collection %>} row_click={&JS.navigate(~p"<%= schema.route_prefix %>/#{&1}")}><%= for {k, _} <- schema.attrs do %>
+<.table id="<%= schema.plural %>" rows={@<%= schema.collection %>} row_click={&JS.navigate(~p"<%= schema.route_prefix %>/#{&1}")}><%= for {k, v} when not is_map(v) <- schema.attrs do %>
```
josevalim, line 10: What about something like this:

```
if Phoenix.HTML.Safe.impl_for(v) do
  <%= v %>
else
  <%= inspect v %>
end
```

This way we keep showing the value but we properly escape it before?

josevalim: Also, do forms work with maps? I am thinking they wouldn't work either, right?

codeanpeace: Hello there!!

Oh neat, so because `Phoenix.HTML.Safe.impl_for(some_map_value)` would return `nil`, it'd instead `inspect some_map_value` which would render a string eg `"%{\"foo\" => \"bar\"}"`. That would fix the error in the generated template. Off top of my head, some tradeoffs would be readibility as large and/or deeply nested maps would affect the table layout while perhaps not being very useful to display.

That's a good point, fixing the rendering error in forms while still displaying the map fields as strings may just push the error into the form submission step. AFAIK, there's no checking for and converting of strings into maps upon form submission, so it'd probably raise an error on `cast` because the map attribute would still be a string.

codeanpeace: Also on that note, I just took a look at the form component templates and realized that it'd also require some additional changes. Currently, the form inputs are created by calling `<%= Mix.Tasks.Phx.Gen.Html.indent_inputs(inputs, 8) %>` in [form_component.ex#L23](https://github.com/phoenixframework/phoenix/blob/master/priv/templates/phx.gen.live/form_component.ex#L23) with  `inputs` set in the mix task via `inputs: Gen.Html.inputs(schema)` in [phx.gen.live.ex#L109](https://github.com/phoenixframework/phoenix/blob/master/lib/mix/tasks/phx.gen.live.ex#L109).

For map attributes defined in the schema, `Gen.Html.inputs` currently generates a text input since it does not explicitly pattern match on `:map` types so it matches the catchall pattern `{key, _} -> ~s(<.input field={{f, #{inspect(key)}}} type="text" label="#{label(key)}" />)` in
[phx.gen.html.ex#L223-L224](https://github.com/phoenixframework/phoenix/blob/master/lib/mix/tasks/phx.gen.html.ex#L223-L224). This also results in the generated form component rendering the `(Protocol.UndefinedError) protocol Phoenix.HTML.Safe not implemented for %{...}`

Potential approaches:
	– pattern match on `{key, :map} -> ` and inspect to display map as string, but not as form input/field
	– pattern match on `{key, :map} -> ` and skip map type attributes when generating html inputs

PR #5241, opened 2023-01-28
Do not render th for actions if actions are empty

hzeus:
When the actions slot is not used, the table header renders an empty `th`, but the table body rows do not render a `td` element.

I added the same condition as used within the table body to fix this.

I also split the lines, as the formatter also did this for me in a project (with gettext enabled).


hzeus:
I just noticed that my first commit made the tests fail and also adjusted the corresponding test. 

I hope it's okay that I amended the change to the previous commit

chrismccord:
❤️❤️❤️🐥🔥

PR #5243, opened 2023-01-30
Remove code from deprecated `mix compile.phoenix`

wojtekmach:
The installer requires Elixir `~> 1.14` and Phoenix itself `~> 1.11`.


josevalim:
The previous code checked `__phoenix_recompile__?` instead of `__mix_recompile__?`, so I would still leave it around for a couple more versions. Thanks :)

wojtekmach:
Oops. I can submit another patch that will warn if phoenix.compile finds a `__phoenix_recompile__?`. I'm happy to let it go too. :)

josevalim:
The compiler itself is deprecated, which means people should not be using it in the `mix.exs` anyway and this way we will catch any missing usage of `__phoenix_recompile__?`, so I think we are fine. :)

PR #5250, opened 2023-02-03
Cross browser rounded hover effect

LostKobrakai:
Second stab at #5013

This one seems to work for me in ff/safari/chrome on mac: https://kobrakai.de/scratchpad/table

### DIFF
--- a/priv/templates/phx.gen.live/core_components.ex
+++ b/priv/templates/phx.gen.live/core_components.ex
@@ -452,8 +452,12 @@ defmodule <%= @web_namespace %>.CoreComponents do
       <table class="mt-11 w-[40rem] sm:w-full">
         <thead class="text-left text-[0.8125rem] leading-6 text-zinc-500">
           <tr>
-            <th :for={col <- @col} class="p-0 pb-4 pr-6 font-normal"><%%= col[:label] %></th>
-            <th class="relative p-0 pb-4"><span class="sr-only"><%= if @gettext do %><%%= gettext("Actions") %><% else %>Actions<% end %></span></th>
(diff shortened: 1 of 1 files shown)


chrismccord:
The current impl works across chrome, firefox, safari just fine for me, so closing this unless you have examples where the existing setup fails. Thanks!

LostKobrakai:
I've just update the example with the latest component code from main. The action column head is now correctly skipped if not necessary, but all the issues around the missing or incorrectly sized rounded borders on the right side of rows are still present.

https://kobrakai.de/scratchpad/table still shows those. Feel free to poke at the source at https://github.com/LostKobrakai/kobrakai_elixir/blob/e945ca27b7e842e9842ef16418a3a66d3cf155af/lib/kobrakai_web/components/core_components.ex#L541 / https://github.com/LostKobrakai/kobrakai_elixir/blob/e945ca27b7e842e9842ef16418a3a66d3cf155af/lib/kobrakai_web/live/scratchpad/table.ex

jdmarshall:
@chrismccord I don't think I agree that it works as is. If you set a dark background, the final column is one pixel too short vertically and the rounded end is completely missing.

I feel like this whole feature is coming at a great expense to the user being able to do their own styling and layout for the tables.  

I'm trying to make modifications to the table template and definition in order to provide attributes to the columns, such as class, but to get any CSS involving layout to work I'm having to gut this feature to get anything to work.

If you're giving people a table to work with, every cell shouldn't come with three dom elements inside it that they didn't ask for, just do do a weird bit of chrome, that I don't even understand the purpose of.

PR #5264, opened 2023-02-10
Update testing_channels.md

scottwoodall:
The docs mention a call to setup the database, but that call is not in the code sample.


josevalim:

PR #5265, opened 2023-02-10
Set default Endpoint check_origin value to :conn

TheFirstAvenger:
Per [this conversation](https://community.fly.io/t/phoenix-liveview-constantly-refreshes-with-custom-domain/3384/5) the `check_origin: :conn` option introduced in Phoenix 1.6.6 is superior to `check_origin: true` and comes with no drawbacks. The post mentions the desire for this to be the default value going forward. This PR sets that default in both of the templates and in the endpoint supervisor defaults.

### DIFF
--- a/installer/templates/phx_single/config/config.exs
+++ b/installer/templates/phx_single/config/config.exs
@@ -14,6 +14,7 @@ config :<%= @app_name %><%= if @namespaced? do %>,
 
 # Configures the endpoint
 config :<%= @app_name %>, <%= @endpoint_module %>,
+  check_origin: :conn,
   url: [host: "localhost"],
   render_errors: [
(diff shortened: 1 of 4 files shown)


chrismccord:
This is a great default for most folks, but we haven't shipped it with new apps because it only works securely if you have a forward proxy in front of your app. Most apps do these days, but we can't guarantee that's how folks ship their apps to prod, so it would be an insecure default in that regard. thanks!

TheFirstAvenger:
@chrismccord Would it be good to put a writeup of this potential security issue in the documentation for that flag? That way someone doesn't implement the flag in an insecure way based on the documentation without knowing the risks and required actions.

Issue #5247, opened 2023-02-03
`phx.gen.auth` Confirm Email Flash Message Not Rendering 


theshapguy:
### Environment

* Elixir version (elixir -v): Elixir 1.14.2
* Phoenix version (mix deps): 1.7.0-rc.2
* Operating system: macOS 13.1

### Expected behavior

In regards to `user_confirmation_controller.ex` generated via `phx.gen.auth`. Issue exists in both LiveView and Controller only generator.

- Flash message should be rendered after clicking on the "Confirm my Account" button. However page is being redirected to the root `/`

```elixir
  def create(conn, %{"user" => %{"email" => email}}) do
    if user = Accounts.get_user_by_email(email) do
      Accounts.deliver_user_confirmation_instructions(
        user,
        &url(~p"/users/confirm/#{&1}")
      )
    end

    conn
    |> put_flash(
      :info,
      "If your email is in our system and it has not been confirmed yet, " <>
        "you will receive an email with instructions shortly."
    )
    |> redirect(to: ~p"/")
  end
```

Just one instance is shown above however, `user_confirmation_controller.ex ` has three instances of it being redirected to the root page `/`.

### Actual behavior

Flash Message is not rendering. No Error Message or Success Message is shown due to the page being redirect to root `/`

https://user-images.githubusercontent.com/1960718/216501430-867dd97b-0bfd-449f-8fae-01c782c7982e.mov

### Possible Solution

https://user-images.githubusercontent.com/1960718/216501972-2e67b065-2641-4ad9-a47f-f59930ac4e01.mov

1. Redirect it to the same page so that the flash message is rendered. I used this to solve the issue.  

```elixir
|> redirect(to: ~p"/users/confirm")

&

|> redirect(to: ~p"/users/confirm/#{token}")
```
2. Implement.`<.flash/>` on the root page.

theshapguy:
I'd be happy to fix this issue once comments are made as this might be related to user authentication didn't want to just submit a pull without discussion. Is there any issue on being redirect to self (if token invalid) or login page (after token confirmation)?

josevalim:
I think we should show flash messages on root! PR is welcome!

theshapguy:
On it!

theshapguy:
I'm working on the PR now

theshapguy:
[...]

PR #5270, opened 2023-02-12
Changed <.flash_group> location to root.html.heex

theshapguy:
Pull Request for #5247
_I made an error while merging in the earlier pull request #5256 so creating a new one as the master changed._

As master has changed to `<.flash_group>`; it is updated accordingly. Since flash_group has also been added to `home.html.heex` since we last discussed, I have added it `<.flash_group>` to `root.html.heex` which supersedes all layouts and remove it from everywhere else. 

I hope this is the correct way to go about it.


chrismccord:
Having the flash group in the app layout and the home page (which does `layout: false`) will support the flash messages showing for all of our generators, so I don't see the need for this change. Thanks!

PR #5271, opened 2023-02-13
Detailing archive uninstall and deps fetching for new contributors

krainboltgreene:
I recently ran into the archive issue detailed in #3906 and realized both of these instructions were missing for new contributors.


josevalim:
@krainboltgreene we have commands at the top of the mix.exs that removes any archive from the path. Could this code not be working for you for some reason? If so, can you please amend that code snippet instead?

krainboltgreene:
@josevalim I think I know what happened, on my fork my `master` branch was behind.

PR #5272, opened 2023-02-13
Show the git diff of any bad formatting to make debugging significantly easier

krainboltgreene:
When tests are failing due to the generators creating files that don't pass the formatter check it's incredibly hard to find read the bad output . This is a caveman approach, using git to show the difference.


krainboltgreene:
Ugh, now realizing that this context isn't a git repository. There's two paths here:

  1. Move the diff into `mix format` itself
  2. Create and commit to a repository to see the diff

josevalim:
This is already in Elixir main ;)

krainboltgreene:
Oh nice.

Issue #5278, opened 2023-02-16
phoenix 1.7-rc.3 dialyzer error in put_new_layout


rubas:
I'm seeing typespec errors in our project since RC3 (locked at 1.7.0-rc.3 (phoenix) ae651c30).  RC2 (1.7.0-rc.2 (phoenix) 71abde2f) does not have this issue.

Looks like the same issue as [#5031.](https://github.com/phoenixframework/phoenix/issues/5031#issue-1412712123)

```elixir
lib/marko_web/controllers/home_controller.ex:1:call
The function call will not succeed.

Phoenix.Controller.put_new_layout(
  _ :: %Plug.Conn{
    :adapter => {atom(), _},
    :assigns => %{atom() => _},
    :body_params => %Plug.Conn.Unfetched{:aspect => atom(), binary() => _},
    :cookies => %Plug.Conn.Unfetched{:aspect => atom(), binary() => _},
    :halted => boolean(),
    :host => binary(),
    :method => binary(),
    :owner => pid(),
    :params => %Plug.Conn.Unfetched{:aspect => atom(), binary() => _},
    :path_info => [binary()],
    :path_params => %{binary() => binary() | [any()] | map()},
    :port => char(),
    :private => %{atom() => _},
    :query_params => %Plug.Conn.Unfetched{
      :aspect => atom(),
      binary() => binary() | [any()] | map()
    },
    :query_string => binary(),
    :remote_ip =>
      {byte(), byte(), byte(), byte()}
      | {char(), char(), char(), char(), char(), char(), char(), char()},
    :req_cookies => %Plug.Conn.Unfetched{:aspect => atom(), binary() => binary()},
    :req_headers => [{_, _}],
    :request_path => binary(),
    :resp_body =>
      nil
      | binary()
      | maybe_improper_list(
          binary() | maybe_improper_list(any(), binary() | []) | byte(),
          binary() | []
        ),
    :resp_cookies => %{binary() => map()},
    :resp_headers => [{_, _}],
    :scheme => :http | :https,
    :script_name => [binary()],
    :secret_key_base => nil | binary(),
    :state =>
      :chunked | :file | :sent | :set | :set_chunked | :set_file | :unset | :upgraded,
    :status => nil | non_neg_integer()
  },
  {MarkoWeb.LayoutView, :app}
)

breaks the contract
(Plug.Conn.t(), [{format :: atom(), layout()}]) :: Plug.Conn.t()

________________________________________________________________________________
[...]

PR #5280, opened 2023-02-16
Fix typespec for Phoenix.put_new_layout

rubas:
Relates to: https://github.com/phoenixframework/phoenix/issues/5278#issuecomment-1432896755

```elixir
@spec put_new_layout(Plug.Conn.t(), [{format :: atom, layout}] | layout) :: Plug.Conn.t()
  def put_new_layout(%Plug.Conn{state: state} = conn, layout)
      when (is_tuple(layout) and tuple_size(layout) == 2) or is_list(layout) or layout == false do
```

The typespec is now in sync with the guards and allows to pass a `layout()` without a `format`.

# Reason
The problem starts when [default layouts](https://github.com/phoenixframework/phoenix/blob/v1.7.0-rc.3/lib/phoenix/controller.ex#L230) are set in the Phoenix.Controller and the error case of [__view__](https://github.com/phoenixframework/phoenix/blob/v1.7.0-rc.3/lib/phoenix/controller.ex#L1823)
and [__layout__ ](https://github.com/phoenixframework/phoenix/blob/v1.7.0-rc.3/lib/phoenix/controller.ex#L1860), don't return a `format`. This makes the return incompatible with new layout definition.

I guess all those function will be removed very soon.

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -690,7 +690,7 @@ defmodule Phoenix.Controller do
 
   Raises `Plug.Conn.AlreadySentError` if `conn` is already sent.
   """
-  @spec put_new_layout(Plug.Conn.t(), [{format :: atom, layout}]) :: Plug.Conn.t()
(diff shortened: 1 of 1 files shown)


josevalim:

PR #5282, opened 2023-02-16
reference from form.html.heex to new.html.heex

vectorfrog:
There isn't a product_html/form.html.heex file, so I've just updated the reference to product_html/new.html.heex.


josevalim:
We are adding form back, thanks!

PR #5283, opened 2023-02-17
Batch longpoll messages to resolve race conditions

chrismccord:


lib/phoenix/transports/long_poll.ex
```diff
@@ -72,7 +72,22 @@ defmodule Phoenix.Transports.LongPoll do
...
     case read_body(conn, []) do
       {:ok, body, conn} ->
-        status = transport_dispatch(endpoint, server_ref, body, opts)
+        # we need to match on both v1 and v2 protocol, as well as wrap for backwards compat
+        batch =
+          if String.starts_with?(body, ~w(["[)) do
+            Phoenix.json_library().decode!(body)
+          else
+            [body]
+          end
```
josevalim, line 81: Sorry, I am a bit confused. Why don't we need to `decode!` on the else branch?

josevalim: Ah, I think I got it. We pass the body to the socket.handle_in, which then calls deserialize.

So unfortunately there is something wrong with this pull request in that we are now calling `socket.handle_in({text, op: :text}, ...)` but `text` is already an encoded document. Which means we need to find another way to encode the json document. Maybe we could use [NDJSON](http://ndjson.org/)? And then here it is just a matter of splitting on newlines. We need to check if JSON.stringify guarantees to encode newlines as `\n`.

test/phoenix/integration/long_poll_channels_test.exs
```diff
@@ -159,7 +159,13 @@ defmodule Phoenix.Integration.LongPollChannelsTest do
   """
   def poll(method, path, vsn, params, json \\ nil, headers \\ %{}) do
     {serializer, json} = serializer(vsn, json)
-    headers = Map.merge(%{"content-type" => "application/json"}, headers)
+    headers =
+      if is_list(json) do
+        Map.merge(%{"content-type" => "application/ndjson"}, headers)
```
mcrumm, line 164: According to the spec this should be `application/x-ndjson`
https://github.com/ndjson/ndjson-spec#33-mediatype-and-file-extensions

assets/js/phoenix/longpoll.js
```diff
@@ -123,7 +123,7 @@ export default class LongPoll {
 
   batchSend(messages){
     this.awaitingBatchAck = true
-    this.ajax("POST", JSON.stringify(messages), () => this.onerror("timeout"), resp => {
+    this.ajax("POST", "application/ndjson", messages.join("\n"), () => this.onerror("timeout"), resp => {
```
mcrumm, line 126: According to the spec this should be [`application/x-ndjson`](https://github.com/ndjson/ndjson-spec#33-mediatype-and-file-extensions)

mcrumm: Ahhh, you already fixed it 🙈

PR #5285, opened 2023-02-17
Use distributed messaging for longpoll if available

chrismccord:


lib/phoenix/transports/long_poll.ex
```diff
@@ -182,8 +184,14 @@ defmodule Phoenix.Transports.LongPoll do
 
   ## Helpers
 
-  defp server_ref(endpoint_id, id, pid, topic) do
-    if endpoint_id == id and Process.alive?(pid), do: pid, else: topic
+  def distributed_node?, do: node() != :"nonode@nohost"
+
+  defp server_ref(endpoint_id, id, pid, topic) when is_pid(pid) do
+    if distributed_node?() do
```
josevalim, line 190:
```suggestion
    if distributed_node?() and node(pid) in nodes() do
```

lib/phoenix/transports/long_poll.ex
```diff
@@ -182,8 +184,14 @@ defmodule Phoenix.Transports.LongPoll do
...
+      pid
+    else
+      if endpoint_id == id and Process.alive?(pid), do: pid, else: topic
+    end
```
josevalim, line 194:
```suggestion
    cond do
      node(pid) in Node.list() -> pid
      endpoint_id == id and Process.alive?(pid) -> pid
      true -> topic
    end
```

josevalim (approved):

PR #5291, opened 2023-02-18
Fix incorrect example of disabling root layout in the Components guide

a3kov:
After completing _Request life-cycle_ guide the reader doesn't have `index` _action_ inside `PageController`. He has `home` action there, and `index` action in another controller. The easiest fix is to change the action name (instead of rewriting the whole thing for another controller)


a3kov:
Same issue with `put_layout` example

josevalim:

PR #5297, opened 2023-02-21
Fix saving product categories example in the Contexts guide

a3kov:
I think this part of the guide is outdated, it refers to a state where there was a single form template


josevalim:
We are adding form back, thanks!

PR #5304, opened 2023-02-25
Fix typo and traces of old options from docs

mathieuprog:


mathieuprog:
Actually I'm not so certain anymore that these options do not exist anymore

josevalim:
Both options exist. They are handled in the Mix.Phoenix.Schema module. :)

PR #5308, opened 2023-02-26
Pin esbuild version to match the build.js script

conradfr:
esbuild v0.17 deprecated the watch option: https://github.com/evanw/esbuild/releases/tag/v0.17.0

As such the build.js in the documentation does not work with it.

(next step would be to have an updated version of it)


conradfr:
I guess that would be fixed by #5399

leandrocp:
Fixed by https://github.com/phoenixframework/phoenix/pull/5404
This issue can be closed :)

conradfr:
Fixed.

PR #5310, opened 2023-02-27
Fix warning for string keys on form for phx.gen.auth on login_live.ex

gabrielmancini:
warning: a map with atom keys was given to a form. Maps are always considered parameters and therefore must have string keys, got: %{email: nil}
  (phoenix_html 3.3.1) lib/phoenix_html/form_data.ex:83: Phoenix.HTML.FormData.Map.name_params_and_opts/2
  (phoenix_html 3.3.1) lib/phoenix_html/form_data.ex:53: Phoenix.HTML.FormData.Map.to_form/2


josevalim:

PR #5311, opened 2023-02-27
Clarify allowed arguments for pipe_through & make router auth imports explicit

nathanl:
I was confused about why `:require_authenticated_user` works, and after investigating, wanted to document it.

Also, update the generated code so that the functions the router `import`s from the auth module are explicitly listed, so that people are not encouraged to `import GiantPileOfFunctions` and make their router pipelines harder to understand.


nathanl:
I'm getting test failures like `test phx.gen.auth + argon2 has no compilation or formatter warnings (--no-live) (Phoenix.Integration.CodeGeneration.AppWithMySqlAdapterTest)`. Any guidance on how to run the generator myself from this repo and examine the output?

chrismccord:
❤️❤️❤️🐥🔥

PR #5318, opened 2023-03-01
fix `phx.digest.clean` only cleaning `.gz` compressed versions

breunigs:


lib/mix/tasks/phx.digest.clean.ex
```diff
@@ -44,6 +44,7 @@ defmodule Mix.Tasks.Phx.Digest.Clean do
     keep = opts[:keep] || @default_keep
     all? = opts[:all] || false
 
+    Mix.Task.run("compile", args)
```
josevalim, line 47: Why this change?

josevalim, line 47:
```suggestion
```

josevalim:

lib/mix/tasks/phx.digest.clean.ex
breunigs: @josevalim honest question, does this actually work for you without it?

Setting up a new phoenix 1.7 installation and manually getting in the patch:
```
mix phx.new
cd hello
curl -s https://raw.githubusercontent.com/breunigs/phoenix/f60c13bfc29a0d50492ecbdebd6f61469620106c/lib/phoenix/digester.ex > ./deps/phoenix/lib/phoenix/digester.ex
mix deps.clean phoenix --build
```

Using this test digester:
```elixir
defmodule DummyDigest do
  def compress_file(_file_path, _content), do: :error

  def file_extensions do
    IO.puts("reading exts!")
    [".foo"]
  end
end

# and config.exs:
config :phoenix, static_compressors: [DummyDigest]
```

results in this behaviour for me:

```
$ mix phx.digest
reading exts!
reading exts!
reading exts!
reading exts!
reading exts!
Check your digested files at "priv/static"
```

this looks good. However:

```
$ mix phx.digest.clean --all
** (UndefinedFunctionError) function DummyDigest.file_extensions/0 is undefined (module DummyDigest is not available)
    DummyDigest.file_extensions()
    (elixir 1.14.2) lib/enum.ex:4249: Enum.flat_map_list/2
    (phoenix 1.7.0) lib/phoenix/digester.ex:386: Phoenix.Digester.remove_compressed_file/2
    (phoenix 1.7.0) lib/phoenix/digester.ex:377: anonymous fn/3 in Phoenix.Digester.remove_files/2
    (elixir 1.14.2) lib/enum.ex:2468: Enum."-reduce/3-lists^foldl/2-0-"/3
    (phoenix 1.7.0) lib/phoenix/digester.ex:372: Phoenix.Digester.remove_files/2
    (phoenix 1.7.0) lib/phoenix/digester.ex:341: Phoenix.Digester.clean_all/1
    (phoenix 1.7.0) lib/mix/tasks/phx.digest.clean.ex:51: Mix.Tasks.Phx.Digest.Clean.run/1
    (mix 1.14.2) lib/mix/task.ex:421: anonymous fn/3 in Mix.Task.run_task/4
    (mix 1.14.2) lib/mix/cli.ex:84: Mix.CLI.run_task/2
    (elixir 1.14.2) lib/code.ex:1260: Code.require_file/2
```

Adding this line back:
```
$ code ./deps/phoenix/lib/mix/tasks/phx.digest.clean.ex
(line added)

$ mix deps.clean phoenix --build                       
* Cleaning phoenix

$ mix phx.digest.clean --all                           
==> phoenix
Compiling 71 files (.ex)
Generated phoenix app
==> hello
Generated hello app
reading exts!
reading exts!
reading exts!
reading exts!
Clean complete for "priv/static"
```

I assume this should also fail the same fashion with vanilla `phx.digest.clean` assuming one sets up the digests to be outdated enough and such. Hence the `IO.puts` to easily see if this code path was actually run.

Looking at `phx.digest`, it runs these:
```
    Mix.Task.run "compile", all_args
    # …
    Mix.Task.run "deps.loadpaths", all_args
```

which is where I cargo culted it from.

breunigs: I accidentally answered in change suggestion: https://github.com/phoenixframework/phoenix/pull/5318#discussion_r1123429862

breunigs: Funnily enough, you had the same question back when the `phx.digest` part was introduced, and answered yourself:
https://github.com/phoenixframework/phoenix/pull/3924/files#r450940231

Maybe we should add a comment on top why this is necessary, in both places, as it's not obvious why this line is needed. Adding an integration test could also work; glancing at the tooling already available for it, it doesn't seem _that_ much effort. Then again, running `phx.digest` probably will download `esbuild`, not sure if that is desired during CI.

josevalim: ah, fantastic, i pushed a fix to add the compile back.

josevalim: honestly, i think adding the code with a comment does its job. the line is relatively simple, the hard part is just knowing why it is there!

breunigs: Ehm, you broke the build :)

https://github.com/phoenixframework/phoenix/commit/c76ad1fa6c915fa584d516588677e1839b501b3f#diff-3b8c1801d19226d48473d25172c9c4c2bf869fc3f070fa6a9167d0cbc39ed63eR48

it's just `args` for the clean variant.

breunigs: no multi task, only multi heart 😉

PR #5322, opened 2023-03-03
Fix typo on template

drselump14:
It looks like there is a typo in this template. The compilation failed in my local without this fix


josevalim:
The code looks correct to me. It will generate `<%= Application.spec(:phoenix, :vsn) %>`, which is what we want. What error are you seeing? How is compilation failing?

drselump14:
Oh, I'm sorry. Maybe it's because I just copy-pasted it. 😓 
The original code was supposed to be generated by the generator, wasn't it?
Let me close this PR, sorry for bothering you

josevalim:
No problem!

PR #5323, opened 2023-03-03
Prevent duplicate DOM ids.

jbosse:
The template for user settings renders two DOM nodes with the same `id`. One is the visible email input on the change email form and the other is a hidden input on the change password form.

This PR gives the hidden email input a specific id of `hidden_user_email` to ensure valid HTML and prevent the error "Multiple IDs detected: user_email. Ensure unique element ids." from being displayed in the developer console.

### DIFF
--- a/priv/templates/phx.gen.auth/settings_live.ex
+++ b/priv/templates/phx.gen.auth/settings_live.ex
@@ -39,7 +39,7 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
       phx-submit="update_password"
       phx-trigger-action={@trigger_submit}
     >
-      <.input field={@password_form[:email]} type="hidden" value={@current_email} />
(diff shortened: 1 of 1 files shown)


josevalim:
Thanks for the PR! The integration tests are failing because the generated file will no longer be formatted. Can you please run the formatter on the generated template  on your machine  and then apply the changes to the template as well?

jbosse:
Absolutely. I wasn’t sure if they were red from my change.

-Jimmy

jbosse:
@chrismccord I tried simply removing the hidden input altogether because I don't understand why is is needed for a password change when you have the user from the socket. That broke a bunch of tests and I assume it is needed somewhere down in the changes validation but I wasn't ready to go on a deep dive today.

jbosse:
Closing until I can get it green.

jbosse:
**Replaced by https://github.com/phoenixframework/phoenix/pull/5330**

PR #5324, opened 2023-03-03
update heroicons upgrade instructions

oohnoitz:
Due to some differences between BSD and GNU `tar`, it may be better to use the `tar.gz` artifact so that the instructions remain consistent across different environments when updating heroicons. This should address the error below when following the instructions with GNU tar installed.

```
tar: This does not look like a tar archive
tar: Skipping to next header
tar: heroicons-2.0.16/optimized: Not found in archive
tar: Exiting with failure status due to previous errors
```

Should the generator be updated to use `optimized.tar.gz` and `:erl_tar.extract/2`?


josevalim:

josevalim:
For the generator it does not matter. :)

PR #5330, opened 2023-03-06
Rename id to ensure all ids are unique.

jbosse:
**Replaces https://github.com/phoenixframework/phoenix/pull/5323**

The template for user settings renders two DOM nodes with the same id. One is the visible email input on the change email form and the other is a hidden input on the change password form.

This PR gives the hidden email input a specific id of hidden_user_email to ensure valid HTML and prevent the error "Multiple IDs detected: user_email. Ensure unique element ids." from being displayed in the developer console.

_I am not sure how I messed up the first branch but this is now a clean branch that should be green. On the bright side my fork now has the actions running so I should avoid red PRs in the future_.


josevalim:

PR #5333, opened 2023-03-07
includes check of socket, longpoll verbs when calculating column widths for routes

nhessler:
If you run `mix phx.routes` when all your non-socket routes have Verb/Method lengths less than `Post` you will get an issue with the formatting of socket longpoll routes.

Screenshot 2023-03-07 at 3 07 42 PM

This is because when calculating column widths, verb length of socket routes is not taken into account. This PR fixes that.

Screenshot 2023-03-07 at 5 05 29 PM

**Note:** *this is my first PR and I'm just getting back into Elixir after a long hiatus. any feedback of course is welcome*

### DIFF
--- a/lib/phoenix/router/console_formatter.ex
+++ b/lib/phoenix/router/console_formatter.ex
@@ -4,6 +4,11 @@ defmodule Phoenix.Router.ConsoleFormatter do
   @doc """
   Format the routes for printing.
   """
+
+  @socket_verb "WS"
+
+  @longpoll_verbs ["GET", "POST"]
+
   def format(router, endpoint \\ nil) do
     routes = Phoenix.Router.routes(router)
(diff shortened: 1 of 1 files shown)


lib/phoenix/router/console_formatter.ex
```diff
@@ -106,4 +112,13 @@ defmodule Phoenix.Router.ConsoleFormatter do
...
   defp verb_name(verb), do: verb |> to_string() |> String.upcase()
+
+  defp socket_verbs(socket_opts) do
+    if socket_opts[:longpoll] != false do
+      [@socket_verb | @longpoll_verbs]
+    else
+      [@socket_verb]
+    end
+  end
+
```
josevalim, line 123:
```suggestion
```

josevalim:

PR #5334, opened 2023-03-08
update to defmodule Mix.Tasks.Phx.Gen.Html documentation

BradS2S:
I added to the Mix.Tasks.Phx.Gen.Html documentation. I also have some minor cosmetic changes on here suggested by credo.


lib/mix/tasks/phx.gen.html.ex
```diff
@@ -2,54 +2,66 @@ defmodule Mix.Tasks.Phx.Gen.Html do
...
+  changing the value of the `:context_app` configuration option.
+  
+  Phoenix generators will generate web files directly in your lib and test folders since 
+  the application is assumed to be isolated to web specific functionality.
+  
+  Example configuration in `my_app_umbrella/config/config.exs`:
+
+    config :my_app_web,
+      ecto_repos: [Stuff.Repo],
+      generators: [context_app: :my_app]
```
josevalim, line 51:
```suggestion
  By default, Phoenix injects both web and domain specific functionality into the same
  application. When using umbrella applications, those concerns are typically broken
  into separate apps. You can teach Phoenix to place your context and migration files
  into a different application via the `:context_app` configuration option.

  Example configuration in `my_app_umbrella/config/config.exs`:

    config :my_app_web,
      ecto_repos: [Stuff.Repo],
      generators: [context_app: :my_app]
```

lib/mix/tasks/phx.gen.html.ex
```diff
@@ -2,54 +2,66 @@ defmodule Mix.Tasks.Phx.Gen.Html do
...
 
   Alternatively, the `--context-app` option may be supplied to the generator:
 
-      mix phx.gen.html Sales User users --context-app warehouse
+    mix phx.gen.html Sales User users --context-app my_app
```
josevalim, line 55:
```suggestion
      mix phx.gen.html Sales User users --context-app my_app
```

lib/mix/tasks/phx.gen.html.ex
```diff
@@ -2,54 +2,66 @@ defmodule Mix.Tasks.Phx.Gen.Html do
...
+  
+  If you delete the `:context_app` configuration option, Phoenix will automatically put generated web files in
+  `my_app_umbrella/apps/my_app_web_web`.
```
josevalim, line 58: Is the `_web_web` a typo or what happens? If so, it is a bug.

lib/mix/tasks/phx.gen.html.ex
josevalim: I found the paragraphs above to be wrong. umbrella apps generated by phoenix does not assume it all goes into the same directory. the above is an attempt to simplify it but the previous text had more details, so we probably want to keep it.

josevalim:
Hi @BradS2S! Thank you for the PR.

However, it has too many different changes. I think we can break it in three:

1. Code style change
2. Cody editing to mix phx.gen.html
3. Changes to the umbrella apps (which may either be a bug in Phoenix or something that needs further clarification)

lib/mix/tasks/phx.gen.html.ex
BradS2S: That is what happens right now. It appends a _web to whatever is in the congfig file

BradS2S:
> Hi @BradS2S! Thank you for the PR.
> 
> However, it has too many different changes. I think we can break it in three:
> 
> 1. Code style change
> 2. Cody editing to mix phx.gen.html
> 3. Changes to the umbrella apps (which may either be a bug in Phoenix or something that needs further clarification)

Sounds good. I revise and resubmit.

lib/mix/tasks/phx.gen.html.ex
josevalim: I see. It seems what is happening here is the following: umbrellas assume you will have two applications, `my_app` and a `my_app_web`, where `my_app` is the context app. So perhaps we should document this assumption explicitly instead?

lib/mix/tasks/phx.gen.html.ex
```diff
@@ -2,54 +2,64 @@ defmodule Mix.Tasks.Phx.Gen.Html do
...
+  By default, Phoenix injects both web and domain specific functionality into the same
+  application. When using umbrella applications, those concerns are typically broken
+  into separate apps. You can teach Phoenix to place your context and migration files
+  into a different application via the `:context_app` configuration option.
+
+  Example configuration in `my_app_umbrella/config/config.exs`:
+
+    config :my_app_web,
+      ecto_repos: [Stuff.Repo],
+      generators: [context_app: :my_app]
```
josevalim, line 49:
```suggestion
  By default, Phoenix injects both web and domain specific functionality into the same
  application. When using umbrella applications, those concerns are typically broken
  into two separate apps, your context application - let's call it `my_app` - and its web
  layer, which Phoenix assumes to be `my_app_web`.

  You can teach Phoenix to use this style via the `:context_app` configuration option
  in your `my_app_umbrella/config/config.exs`:

      config :my_app_web,
        ecto_repos: [Stuff.Repo],
        generators: [context_app: :my_app]
```

lib/mix/tasks/phx.gen.html.ex
josevalim: In any case, besides documenting the assumptions, it seems the my_app_web_web has to be addressed. Please file a bug report. For simplicity, we can actually assume the web app to be whatever app the generator is running on.

chrismccord:
This has gone stale. If you'd like to pick things up based on the latest feedback, please let us know. Thanks!

PR #5337, opened 2023-03-09
Fetch and extract heroicons in one step

c4710n:
In UNIX like system, it's a common practice to alias `rm` to less dangerous commands, like:
+ `alias rm="trash"`
+ `alias rm="echo 'please use trash'"`
+ ...

If the developer indeed did so, the commands in `UPGRADE.md` may fail.

A good solution is to not use the "rm" command.

---

This PR is trying to fetching and extracting heroicons in one step, without using `rm`.


josevalim:
Merged manually, thank you!

PR #5338, opened 2023-03-09
update PageController index action to home in Controllers Guide

nhessler:
not sure how it was setup before Phoenix 1.7, but the default action for "/" is now `PageController.home` and not `PageController.index`. I've gone through and updated the controller guide docs changing appropriate references to the `home` action and adding `layout: false` in the code snippets as well. 

In the flash section I removed the `info` flash message as it cannot be seen along with the `error` flash message and might cause confusion and updated the reference to the `flash` code in the app template to point at the `flash_group` component added to the top of the `home.html.heex` file.

resolves issue #5365

### DIFF
--- a/guides/controllers.md
+++ b/guides/controllers.md
@@ -14,15 +14,15 @@ A newly generated Phoenix app will have a single controller named `PageControlle
 defmodule HelloWeb.PageController do
   use HelloWeb, :controller
 
-  def index(conn, _params) do
-    render(conn, :index)
+  def home(conn, _params) do
+    render(conn, :home, layout: false)
   end
 end
 ```
 
 The first line below the module definition invokes the `__using__/1` macro of the `HelloWeb` module, which imports some useful modules.
 
-`PageController` gives us the `index` action to display the Phoenix [welcome page] associated with the default route Phoenix defines in the router.
+`PageController` gives us the `home` action to display the Phoenix [welcome page] associated with the default route Phoenix defines in the router.
 
 ## Actions
 
@@ -372,7 +372,7 @@ For our convenience, a `flash_group` component is already available and added to
 <.flash_group flash={@flash} />
 ```
 
-When we reload the [welcome page], our messages should appear in the top right corner of the page.
+When we reload the [welcome page], our message should appear in the top right corner of the page.
 
 The flash functionality is handy when mixed with redirects. Perhaps you want to redirect to a page with some extra information. If we reuse the redirect action from the previous section, we can do:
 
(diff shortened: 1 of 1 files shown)


guides/controllers.md
```diff
@@ -351,48 +351,34 @@ Sometimes we need to communicate with users during the course of an action. Mayb
...
     conn
-    |> put_flash(:info, "Welcome to Phoenix, from flash info!")
     |> put_flash(:error, "Let's pretend we have an error.")
-    |> render(:index)
+    |> render(:home, layout: false)
   end
 end
 ```
 
 In order to see our flash messages, we need to be able to retrieve them and display them in a template layout. We can do that using [`Phoenix.Flash.get/2`] which takes the flash data and the key we care about. It then returns the value for that key.
```
phoebe100, line 367: As long as multiple flash messages are not shown, I like the idea to put only one flash message in the example.
But then, we should also say "our flash message" instead of "our flash messages"

guides/controllers.md
```diff
@@ -351,48 +351,34 @@ Sometimes we need to communicate with users during the course of an action. Mayb
...
-  phx-disconnected={show("#disconnected")}
-  phx-connected={hide("#disconnected")}
->
-  Attempting to reconnect <.icon name="hero-arrow-path" class="ml-1 w-3 h-3 inline animate-spin" />
-</.flash>
+<.flash_group flash={@flash} />
 ```
 
-When we reload the [welcome page], our messages should appear just above "Welcome to Phoenix!"
+When we reload the [welcome page], our messages should appear in the top right corner of the page.
```
phoebe100, line 375: Since you adjusted the example so that only one flash message is put, it should be "our message" instead of "our messages" here

guides/controllers.md
phoebe100, line 1: Will close #5365

guides/controllers.md
```diff
@@ -351,48 +351,34 @@ Sometimes we need to communicate with users during the course of an action. Mayb
...
 The flash functionality is handy when mixed with redirects. Perhaps you want to redirect to a page with some extra information. If we reuse the redirect action from the previous section, we can do:
 
 ```elixir
-  def index(conn, _params) do
+  def home(conn, _params) do
     conn
-    |> put_flash(:info, "Welcome to Phoenix, from flash info!")
     |> put_flash(:error, "Let's pretend we have an error.")
     |> redirect(to: ~p"/redirect_test"))
   end
```
phoebe100, line 384: Since you adjusted the example so that only one flash message is put, the following sentence should be adjusted as well:

"you will be redirected and the flash messages will be shown once more"
should become
"you will be redirected and the flash message will be shown once more"

guides/controllers.md
nhessler: fixed, thank you!

guides/controllers.md
nhessler: thanks! it should be good to go now.

guides/controllers.md
nhessler: thanks for the heads up on these. issues corrected. if I resolve this and it unlinks the connection to the issue. I'll re-link it in the description.

chrismccord:
❤️❤️❤️🐥🔥

PR #5340, opened 2023-03-10
Fix typo: peeking, not peaking

seth:
Minor spelling fix. I think you want "peek" here, as in to look and not "peak" as in summit.


josevalim:

PR #5341, opened 2023-03-10
[heroicons] Fix inconsistent names

c4710n:
`heroicons` is a more appropriate name, because:
+ the official packages is using it, such as `@heroicons/react` and `@heroicons/vue`.
+ the `UPGRADE.md` is using it.
+ ...

Maybe, it is better to make them consistent.

This PR is trying to eliminate this inconsistency, and fix some bad comments.


chrismccord:
❤️❤️❤️🐥🔥

PR #5349, opened 2023-03-11
Fix hard coded paths in tests generated by `phx.gen.auth`

sebroeder:
In some tests generated by `phx.gen.auth` the paths are hard coded to `~p"/users/..."`, so the tests fail when a different schema module or web namespace is used.

To fix this problem, interpolate `<%= schema.route_prefix %>` into `~p` instead of hard coding `/users`.
 
I tested this PR manually with two examples from the official documentation:

```
$ mix phx.gen.auth Accounts User users --web Warehouse
$ mix phx.gen.auth Backoffice Admin admins
```

Please let me know if you would like additional assertions in `test/mix/tasks/phx.gen.auth_test.exs` for this PR. By reading through the existing tests I came to the conclusion that this bug is too specific to warrant its own regression test. Maybe adding a test for a schema module other than `User` might be a good idea, though?


josevalim:

PR #5351, opened 2023-03-12
Adjust text size of login and register links.

robindrost:
Both the login and register links on the bottom of the forgot / reset password screens is larger then any of the other text. I have changed them to the same size as the other text on the page.


josevalim:

PR #5354, opened 2023-03-13
Add as comments the phx.new arguments in mix.exs

bougueil:
This PR allows the main mix.exs file to be documented with the `phx.new` arguments.
Think it might be useful.

The output generated by phx.new is similar to this:
```
 # Generated with:
 #   phx.new my_app --no-ecto
  def project do
    [
      app: :my_app,
      ...
    ]
```

### DIFF
--- a/installer/lib/mix/tasks/phx.new.ex
+++ b/installer/lib/mix/tasks/phx.new.ex
@@ -149,7 +149,7 @@ defmodule Mix.Tasks.Phx.New do
 
       {opts, [base_path | _]} ->
         generator = if opts[:umbrella], do: Umbrella, else: Single
-        generate(base_path, generator, :project_path, opts)
+        generate(base_path, generator, :project_path, opts, argv)
     end
   end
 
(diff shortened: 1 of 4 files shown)


Gazler:
Hi, thanks for the PR.

I think that whilst this may initially be useful when generating a new project, it is the sort of thing that could fall out of sync with the project as the codebase or dependencies change.

Personally, I'd put the command used to generate the project in the commit message for the initial commit.

bougueil:
Hi Gazler,
yes, documenting how a file / project was generated is common practice.
I would be glad to adapt this to a more suitable form as you suggest.
Could you elaborate a little about this unitial commit as for instance, newcomers may use phx.new with different options and observe their differences without committing ...

Gazler:
Hi,

So for me, when creating a new project, if I wanted to document how it was generated, I'd do something like:

```
mix phx.new hello --no-ecto
cd hello
git init
git commit -am "initial commit generated with `mix phx.new hello --no-ecto`
```

rmoorman:
Maybe the [readme file ](https://github.com/phoenixframework/phoenix/blob/main/installer/templates/phx_single/README.md) would be a more suitable place to document this, as it also contains information that is very much prone to need amendment as the project evolves and is usually one of the places to document things that do not relate directly to a source code file itself?

Issue #5356, opened 2023-03-14
Core component <.flash_group> causes scroll to top when showing flash


Hannes-Steffenhagen-bd:
### Environment

* Elixir version (elixir -v): Elixir 1.14.0 (compiled with Erlang/OTP 25)
* Phoenix version (mix deps):
  + phoenix (Hex package) (mix)
     locked at 1.7.1 (phoenix) ea9d4a85
* Operating system: MacOS

### Expected behavior

After a `put_flash`, the flash message should appear and our scroll position should stay the same

### Actual behavior

The message appears but causes a scroll to top.

### Reason/potential fix

The way the `flash_group` is written it just inserts the flashes directly into the DOM without a wrapper. Since the `flash` elements inside the group aren't rendered at all if the respective flash isn't set, this adds a new sibling element and causes everything else at the same level in the DOM to be rerendered, which in turn causes a loss of the current scroll position.

The simplest fix I can think of would be to simply wrap a wrapper `div` around the `<.flash>` elements in `<.flash_group>`:

```diff
  def flash_group(assigns) do
    ~H"""
+    <div>
      <.flash kind={:info} title="Success!" flash={@flash} />
      <.flash kind={:error} title="Error!" flash={@flash} />
      <.flash
        id="disconnected"
        kind={:error}
        title="We can't find the internet"
        phx-disconnected={show("#disconnected")}
        phx-connected={hide("#disconnected")}
        hidden
      >
        Attempting to reconnect <.icon name="hero-arrow-path" class="ml-1 h-3 w-3 animate-spin" />
      </.flash>
+    </div>
    """
  end
```

Alternatively the default project could also put that `<div>` around `<.flash_group>` inside the generated `app.html.heex`, but I can't think of a reason right now why you wouldn't just put it into the component.

<details>
  <summary>Link to source</summary>

https://github.com/phoenixframework/phoenix/blob/8ab705603ac695779bf40668b0c63a46ebcc19e5/installer/templates/phx_web/components/core_components.ex#L144-L159

</details>

PR #5357, opened 2023-03-14
Fix weird scrolling behaviour with flash_group

Hannes-Steffenhagen-bd:
If a sibling element appears or disappears other elements without an id will be rerendered, which can cause weird issues, including strange scrolling behaviour. Adding the extra <div> makes it so that a flash appearing/disappearing only affects other flashes in the group.

Fixes https://github.com/phoenixframework/phoenix/issues/5356


chrismccord:
❤️❤️❤️🐥🔥

PR #5358, opened 2023-03-14
Update core input component to support list attribute

bkilshaw:
Adds the "list" attribute to the allowed attributes you can pass into the input component. The list attribute is used in conjunction with the `datalist` element to generate auto-complete suggestions on text inputs.

```html
<input name="fruit"  list="fruit-suggestions" />

<datalist id="fruit-suggestions">
  <option value="Apple" />
  <option value="Orange" />
  <option value="Banana" />
</datalist>
```


josevalim:

PR #5362, opened 2023-03-15
Adding info log if they have http/https set but not server: true set

jeregrine:
When a user misconfigures their app to not set `server: true` on the endpoint with releases the app just hangs and has no reason to log anything by default. 

See https://github.com/dwyl/phoenix-liveview-chat-example/pull/111#discussion_r1137410990 this should warn them if they have HTTP/HTTPS set but not server: true.

### DIFF
--- a/lib/phoenix/endpoint/supervisor.ex
+++ b/lib/phoenix/endpoint/supervisor.ex
@@ -123,16 +123,17 @@ defmodule Phoenix.Endpoint.Supervisor do
   end
 
   defp server_children(mod, config, server?) do
-    if server? do
-      adapter = config[:adapter] || Phoenix.Endpoint.Cowboy2Adapter
-      adapter.child_specs(mod, config)
-    else
-      if config[:http] || config[:https] do
+    cond do
+      server? ->
+        adapter = config[:adapter] || Phoenix.Endpoint.Cowboy2Adapter
+        adapter.child_specs(mod, config)
+
+      config[:http] || config[:https] ->
         Logger.info(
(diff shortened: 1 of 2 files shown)


lib/phoenix/endpoint/supervisor.ex
```diff
@@ -127,6 +127,11 @@ defmodule Phoenix.Endpoint.Supervisor do
       adapter = config[:adapter] || Phoenix.Endpoint.Cowboy2Adapter
       adapter.child_specs(mod, config)
     else
+      if config[:http] || config[:https] do
```
josevalim, line 130:
```suggestion
      unless config[:http] || config[:https] do
```

josevalim: Although I would probably use a cond.

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -127,6 +127,11 @@ defmodule Phoenix.Endpoint.Supervisor do
...
+        Logger.info(
+          "Configuration :server is false or not set for #{inspect(mod)}, will not be starting a webserver"
```
josevalim, line 132:
```suggestion
          "Configuration :server was not enabled for #{inspect(mod)}, http/https services won't start"
```

jeregrine:
@josevalim <3 thanks

josevalim (approved):

chrismccord:
❤️❤️❤️🐥🔥

PR #5368, opened 2023-03-19
flash_group: Use Gettext

wceolin:
The `flash_group` core component isn't using `Gettext`. Note: I'm getting a compile error on line `156`: `** (CompileError) templates/phx_web/components/core_components.ex:156: undefined function gettext/1 (expected Phx.New.Web to define such a function or for it to be imported, but none are available)`

Not sure why since calling `gettext` seems to be working for the other calls, just this one is broken. I'd appreciate if someone could take a look.


chrismccord:
Gettext is there are a starting point and we include some examples but we don't localize all templates out of the box. Thanks!

PR #5373, opened 2023-03-22
updated phx.gen.html copy

BradS2S:
Added more details about how the generator works.

### DIFF
--- a/lib/mix/tasks/phx.gen.html.ex
+++ b/lib/mix/tasks/phx.gen.html.ex
@@ -2,54 +2,62 @@ defmodule Mix.Tasks.Phx.Gen.Html do
   @shortdoc "Generates context and controller for an HTML resource"
 
   @moduledoc """
-  Generates controller, HTML views, and context for an HTML resource.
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/phx.gen.html.ex
```diff
@@ -2,54 +2,62 @@ defmodule Mix.Tasks.Phx.Gen.Html do
...
+  the context, context test, and context test helper modules.
+
+  ## Umbrella app configuration
+
+  By default, Phoenix injects both web and domain specific functionality into the same
+  application. When using umbrella applications, those concerns are typically broken
+  into two separate apps, your context application - let's call it `my_app` - and its web
+  layer, which Phoenix assumes to be `my_app_web`.
+
+  You can teach Phoenix to use this style via the `:context_app` configuration option
```
Gazler, line 45: Small thing here, "teach" implies that Phoenix is capable of learning. I'd change it to something like:

`Phoenix can be configured to use this style via the...`

Gazler (changes requested):

PR #5374, opened 2023-03-22
Change Hero Icons to Heroicons

zimt28:
Use the spelling that is used by the Tailwind/Heroicons team: https://tailwindcss.com/docs/stroke-width#setting-the-stroke-width


chrismccord:
❤️❤️❤️🐥🔥

PR #5377, opened 2023-03-22
Notes about payload serialization

nathanl:
I was confused by some code which set a struct as a payload until I recalled that payloads get serialized this way and saw that the struct implemented the required protocol. I think this makes the expectation clearer.

### DIFF
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -347,7 +347,8 @@ defmodule Phoenix.Channel do
   alias Phoenix.Socket
   alias Phoenix.Channel.Server
 
-  @type payload :: map | {:binary, binary}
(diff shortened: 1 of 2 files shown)


lib/phoenix/channel.ex
```diff
@@ -347,7 +347,8 @@ defmodule Phoenix.Channel do
   alias Phoenix.Socket
   alias Phoenix.Channel.Server
 
-  @type payload :: map | {:binary, binary}
+  @typedoc "Payload can be any term which the configured serializer can encode."
```
chrismccord, line 350:
```suggestion
```

chrismccord:
❤️❤️❤️🐥🔥

PR #5393, opened 2023-04-02
updated folder structures to match 1.7

nhessler:
updated folder structure of `priv/static` and `assets/` as part of `mix.digest` guide to match output from phoenix 1.7


phoebe100 (approved):
looks good - thank you

PR #5399, opened 2023-04-04
Update asset_management.md for esbuild v0.17

leandrocp:
`watch` opt was removed on v0.17 https://github.com/evanw/esbuild/blob/main/CHANGELOG.md#0170

This change is to make use of the new `context` method.

### DIFF
--- a/guides/asset_management.md
+++ b/guides/asset_management.md
@@ -103,42 +103,40 @@ const plugins = [
   // Add and configure plugins here
 ]
 
-let opts = {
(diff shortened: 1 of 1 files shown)


josevalim:
Sorry, I merged another PR for this. Please double check if main is good or if you would like to see other changes.

leandrocp:
No worries, it looks good. The only nitpick is that usually docs don't use `;` so there's no need to add it, but again a small nitpick.

PR #5402, opened 2023-04-05
Update app_name_web.ex

RaymondLoranger:
Allows a single line instead of multiple lines.
Example:
  `use HelloWeb, [:live_view, :imports, :aliases]`
  instead of
  `use HelloWeb, :live_view`
  `use HelloWeb, :imports`
  `use HelloWeb, :aliases`


josevalim:
We don't plan to support this by default in new apps but you are welcome to incorporate the changes to your own. :)

PR #5407, opened 2023-04-06
Improve comment of :watchers option

c4710n:
`esbuild` and `tailwind` can be disabled in `1.7.2`.

Therefore, the comment **"For example, we use it with esbuild to bundle .js and .css sources."**  is not always accurate.

This PR adjusts the comment to make it reasonable in all cases.


josevalim:

PR #5412, opened 2023-04-09
Fix Heroicons upgrade instructions

aptinio:


josevalim:
I don't think we need the -z option for extracting:

>     -z, --gunzip, --gzip
>             (c mode only) Compress the resulting archive with gzip(1).  In extract or
>             list modes, this option is ignored.  Note that this tar implementation
>             recognizes gzip compression automatically when reading archives.

PR #5420, opened 2023-04-16
Allow to return close code and detail from Phoenix.Socket.Transport callbacks

mickel8:
I am not sure if those changes make any sense but [cowboy_websocket](https://ninenines.eu/docs/en/cowboy/2.6/manual/cowboy_websocket/), which is implemented in [cowboy_adapter](https://github.com/phoenixframework/websock_adapter/blob/main/lib/websock_adapter/cowboy_adapter.ex) allows for returning specific close code and additional info when closing a WebSocket, and it would be nice to allow for returning those values from modules implementing `Phoenix.Socket.Transport` too.

The problem is that those values can't be returned when longpolling is used.

On the other hand, the similar discrepancy is with the message which is marked as term but websocket transport can accept only binary. 

We would benefit from those changes in our media server called [Jellyfish](https://github.com/jellyfish-dev/jellyfish/blob/main/lib/jellyfish_web/peer_socket.ex#L68) where we rely on plain WebSockets as we would like to return more specific close details, at least for debugging purposes.

### DIFF
--- a/lib/phoenix/socket/transport.ex
+++ b/lib/phoenix/socket/transport.ex
@@ -136,9 +136,9 @@ defmodule Phoenix.Socket.Transport do
   Connects to the socket.
 
   The transport passes a map of metadata and the socket
-  returns `{:ok, state}`. `{:error, reason}` or `:error`. 
-  The state must be stored by the transport and returned 
-  in all future operations. `{:error, reason}` can only 
+  returns `{:ok, state}`. `{:error, reason}` or `:error`.
+  The state must be stored by the transport and returned
+  in all future operations. `{:error, reason}` can only
   be used with websockets.
 
   This function is used for authorization purposes and it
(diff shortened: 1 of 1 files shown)


lib/phoenix/socket/transport.ex
```diff
@@ -176,16 +176,24 @@ defmodule Phoenix.Socket.Transport do
...
+
+  When stopping the built-in websocket transport, specific `close_code` and
+  additional `detail` can be returned.
   """
   @callback handle_in({message :: term, opts :: keyword}, state) ::
               {:ok, state}
               | {:reply, :ok | :error, {opcode :: atom, message :: term}, state}
               | {:stop, reason :: term, state}
+              | {:stop, reason :: term, close_code :: integer(), state}
+              | {:stop, reason :: term, {close_code :: integer(), detail :: binary()}, state}
```
mickel8, line 196: Not sure if `detail` should be `binary` `iodata` or `term` :thinking:

PR #5430, opened 2023-04-20
Reintroduce the File Uploads guide

mcrumm:
If you search for `elixir phoenix upload` your first result will be a page that is about six years old

Screenshot 2023-04-20 at 4 35 47 PM

If you are very unlucky, you might find a link on the forums to a guide on `phoenixframework.org` and it will 404

This PR reintroduces the File Uploads guide, updated to work with the current Contexts guide.

PR #5433, opened 2023-04-21
Add file upload attributes to generated CoreComponents

mcrumm:
Adds the `multipart` attribute and file-specific attributes to the form and input components respectively.

* Related to #5430


mcrumm:
The input attrs are not strictly speaking required– happy to drop them if preferred :)

PR #5438, opened 2023-04-25
Improve user prompt for gen auth

louiscb:
When running the gen auth mix command, e.g `mix phx.gen.auth Accounts User users` the user prompt can be somewhat confusing. In particular the use of the term twice "authentication system" in both the option description and the question posed to the user.

Current prompt:

```
An authentication system can be created in two different ways:
- Using Phoenix.LiveView (default)
- Using Phoenix.Controller only
Do you want to create a LiveView based authentication system? [Yn] n
```

This pr proposes:

```
An authentication system can be created in two different ways:
  1) Using Phoenix.LiveView (default)
  2) Using Phoenix.Controller only
Which option (1,2) do you want to use? (Leave empty for default): 
```

### DIFF
--- a/guides/authentication/mix_phx_gen_auth.md
+++ b/guides/authentication/mix_phx_gen_auth.md
@@ -12,13 +12,13 @@ Let's start by running the following command from the root of our app (or `apps/
 $ mix phx.gen.auth Accounts User users
 
 An authentication system can be created in two different ways:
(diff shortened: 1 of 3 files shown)


chrismccord:
Thanks! I like this change, but my only hesitation is piping with `yes` won't work so I'm not sure if it's worth the change.

PR #5441, opened 2023-04-27
Fix put_root_layout typespec

### LINKED ISSUE #5437: Wrong spec for `put_root_layout/2` method
### Environment

* Elixir version (elixir -v): Elixir 1.14.0 (compiled with Erlang/OTP 25)
* Phoenix version (mix deps): phoenix 1.7.2 (Hex package) (mix) locked at 1.7.2 (phoenix) 1ebca94b
* Operating system: macOS  Ventura version 13.3.1 (22E261)

### Expected behavior

Should be able to use the `put_root_layout/2` method with the value `false` to replace the root layout as specified in the documentation, without encountering any Dialyzer warnings.

Example:
```elixir
def print(conn, _params) do
    conn
    |> put_root_layout(false)
    |> put_layout("empty.html")
    |> render("print.html")
end
```

### Actual behavior

A Dialyzer warning is observed since the spec does not allow the value `false`.

#### Current spec
```
@spec put_root_layout(Plug.Conn.t(), [{format :: atom(), layout()}]) :: Plug.Conn.t()
```

#### Warning
```
The call 'Elixir.Phoenix.Controller':put_root_layout
         (_conn@1 :: any(),
          'false') breaks the contract ('Elixir.Plug.Conn':t(),[{format::atom(),layout()}]) -> 'Elixir.Plug.Conn':t()ElixirLS Dialyzer
```

Possibly it should have the same spec as the `put_layout/2` method:
```
@spec put_layout(Plug.Conn.t(), [{format :: atom(), layout()}] | false) :: Plug.Conn.t()
```

axelson:
Fixes #5437

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -755,7 +755,7 @@ defmodule Phoenix.Controller do
 
   Raises `Plug.Conn.AlreadySentError` if `conn` is already sent.
   """
-  @spec put_root_layout(Plug.Conn.t(), [{format :: atom, layout}]) ::
+  @spec put_root_layout(Plug.Conn.t(), [{format :: atom, layout}] | false) ::
           Plug.Conn.t()
   def put_root_layout(%Plug.Conn{state: state} = conn, layout) do
     if state in @unsent do


josevalim:

PR #5444, opened 2023-04-29
Add default max_age to Phoenix.Token sign and encrypt typespec

SteffenDE:
https://github.com/phoenixframework/phoenix/commit/0587b9d0dc03267b6acf6d96ad14fa0095247407 added typespecs to Phoenix.Token, but did not include the `max_age` option for `Phoenix.Token.sign/4` and `Phoenix.Token.encrypt/4`. It also removed the documentation for `max_age` from these functions - I don't know why.
https://github.com/elixir-plug/plug_crypto/blob/5884b466766676f7db7f79d62cadead47f161d36/lib/plug/crypto.ex#L171 includes max_age for sign and encrypt.

This PR re-adds the documentation and also adds the max_age option to the typespec.

### DIFF
--- a/lib/phoenix/token.ex
+++ b/lib/phoenix/token.ex
@@ -115,9 +115,11 @@ defmodule Phoenix.Token do
       when generating the encryption and signing keys. Defaults to `:sha256`
     * `:signed_at` - set the timestamp of the token in seconds.
       Defaults to `System.system_time(:second)`
+    * `:max_age` - the default maximum age of the token. Defaults to
+      86400 seconds (1 day) and it may be overridden on verify/4.
 
   """
-  @spec sign(context, binary, term, [shared_opt | signed_at_opt]) :: binary
+  @spec sign(context, binary, term, [shared_opt | max_age_opt | signed_at_opt]) :: binary
   def sign(context, salt, data, opts \\ []) when is_binary(salt) do
     context
     |> get_key_base()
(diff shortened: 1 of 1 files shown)


SteffenDE:
Sorry to bother again, but I think this should be discussed a little bit more.
I'm fine with encouraging giving the max-age on verification/decryption, but this issue seems more like a mistake in #4803 than a deliberate choice, as I find no discussion/reasoning about removing support for this.

It also breaks existing workflows, for example I have a small project that lets the user specify the validity of the token using max_age when signing, so a hardcoded value on decrypt does not work and I'd have to store this somewhere else (which is of course possible, but I currently don't see a reason for this).

As long as Plug supports this, I think Phoenix should do it as well. If there is a strong reason not to support it any more, it should be clearly stated so. Currently this seems like a breaking change not documented in any changelog.

Of course that's only my opinion though :)

josevalim:
You have a convincing opinion :) :green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart:

nshafer:
Thank you @SteffenDE  and @josevalim, I appreciate both of your efforts to resolve this. I agree that in typical cases the max_age should be specified in `verify/4` which overrides any max_age in the token itself. And this could lead to some major footguns if improperly used, but I thought I'd just throw in my opinion as well to hopefully keep this functionality and not deprecate it in Phoenix.Token or Plug.Crypto.

I also have a facility that lets the user create a new Phoenix.Token with a specified max_age (constrained to reason). These tokens are then distributed by my user to dozens, hundreds, possibly thousands of clients to authorize a specific action (along with a second factor) via whatever mechanism: GPO, shared drive, thumb drive, floppy disk, carrier pigeon, whatever. I want the user to specify the max_age on creation to suit their use case for that specific token. Then on the server when it gets a token it can trust that the Plug.Crypto and Phoenix.Token code has been well vetted to verify that token by its max age. Without that, I would need to either ignore the warning, use Plug.Crypto directly, or duplicate the max_age (that's already in the token anyways) in the token payload that may not be as well implemented. (I always prefer to trust library code for security related functions as much as possible over my own.)

So thanks again, you're both legends. :heart:

PR #5445, opened 2023-04-29
Do not re-raise on NoRouteError

mtrudel:
Based on the commentary at https://github.com/phoenixframework/phoenix/blob/main/lib/phoenix/endpoint/render_errors.ex#L4-L6, the observed behaviour of Phoenix is incorrect; presently we re-raise all errors when rendering them. This has never really been visible with Cowboy since it generally does not log errors. However, Bandit's general policy is to log any errors which are raised out of a plug, which made this much more visible.

This PR corrects the behaviour of RenderErrors to match its documentation, resulting it in the following changes to Phoenix's outward behaviour:

* `NoRouteError` errors are no longer re-raised
* All other raised errors continue to be re-raised
* All client-facing logging is unchanged

In terms of broader outward behaviours:
* Cowboy's end-to-end behaviour is unchanged (it was silently eating those errors anyway)
* Bandit's behaviour changes to not log loudly on NoRouteErrors, and to continue logging loudly on all other errors

### DIFF
--- a/lib/phoenix/endpoint/render_errors.ex
+++ b/lib/phoenix/endpoint/render_errors.ex
@@ -53,16 +53,18 @@ defmodule Phoenix.Endpoint.RenderErrors do
 
   @doc false
   def __catch__(conn, kind, reason, stack, opts) do
-    receive do
-      @already_sent ->
-        send(self(), @already_sent)
-        %Plug.Conn{conn | state: :sent}
-    after
-      0 ->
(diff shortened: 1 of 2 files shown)


josevalim:

JonRowe:
FWIW this has an additional outward behaviour in that affects tests checking for 404 errors (I'm doing this to sanity check my `ErrorHTML` config as is or was possible to misconfigure it and have it fallback to the old view setup if you have the view around and components). Would you accept a PR updating the changelog to highlight this?

PR #5447, opened 2023-04-30
Improved the visual consistency of the select input core component.

krewast:
In the core components, [textarea inputs](https://github.com/phoenixframework/phoenix/blob/a310102eb0e20b918624bb2323a6afb124fcaddd/priv/templates/phx.gen.live/core_components.ex#L337) and [nearly all others](https://github.com/phoenixframework/phoenix/blob/a310102eb0e20b918624bb2323a6afb124fcaddd/priv/templates/phx.gen.live/core_components.ex#L360) have a "mt-2" class, only the [select input](https://github.com/phoenixframework/phoenix/blob/a310102eb0e20b918624bb2323a6afb124fcaddd/priv/templates/phx.gen.live/core_components.ex#L317) has a "mt-1" class. This makes things look inconsistent when, for example a text input is horizontally positioned next to a select input. This commit fixes that.

Before (Please don't mind the German :smile:):
phoenix-select-input-old

After:
phoenix-select-input-new

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -314,7 +314,7 @@ defmodule <%= @web_namespace %>.CoreComponents do
       <select
         id={@id}
         name={@name}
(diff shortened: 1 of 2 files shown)


chrismccord:
❤️❤️❤️🐥🔥

PR #5448, opened 2023-05-01
Fix typo in using ssl guide

dimamik:
This can save some time (and deploy cycles) for future folks, as we should not use `System.get_env/1` which can only be run in runtime.


Gazler:
Thanks <3

PR #5450, opened 2023-05-01
[docs] Update `Phoenix.Controller.action_fallback/1` documentation

mariusbutuc:
Maintain consistency in the documentation examples between passing the status code to `put_status/2` as either an integer or an atom. 

Give preference here to the atom format since the HTTP code is visible when passing the template name to `Phoenix.Controller.render/2`.


Gazler:
Thanks <3

PR #5454, opened 2023-05-08
Retain route order for warn on verify

### LINKED ISSUE #5451: Problem using `warn_on_verify` on a catch-all route
### Environment

* Elixir version: 1.14.3
* Phoenix version: 1.7.2
* Operating system: macOS 13.3.1

### Expected behavior

I may be using `warn_on_verify` wrong, but from what I understand, it should log a warning for any verified route URLs that would be handled by that particular route. So I _should_ be able to define a "catch-all" route to act as a custom 404 handler like so:

```elixir
scope "/", MyAppWeb do
  pipe_through :browser

  get "/", PageController, :home
  get "/*path", PageController, :not_found, warn_on_verify: true
end
```

The routes match as expected and the custom 404 page is displayed. Creating a link to `~p"/"` should not log any warnings, since it would be handled by the first route and never make it to the second.

### Actual behavior

The following warning is logged when creating a link to `~p"/"`:

```
warning: no route path for MyAppWeb.Router matches "/"
```

phoebe100:
Retain route definition order during creation of routers __verify_route__/1 functions

Closes #5451

### DIFF
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -506,10 +506,15 @@ defmodule Phoenix.Router do
     {matches, {pipelines, _}} =
       Enum.map_reduce(routes_with_exprs, {[], %{}}, &build_match/2)
 
-    verifies =
+    routes_per_path =
       routes_with_exprs
       |> Enum.group_by(&elem(&1, 1).path, &elem(&1, 0))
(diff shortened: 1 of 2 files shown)


josevalim:

PR #5473, opened 2023-05-25
Correct the use of TailwindCSS JIT class

rickclare:
A recent change was made to the root.html.heex layout - "Use tailwind JIT to accommodate CSP"

However, in that change, the syntax used was slightly incorrect (an extra space character), so TailwindCSS was not able to interpret the "arbitrary property" 

Correcting this change by omitting the space character.

(see the [TailwindCSS docs](https://tailwindcss.com/docs/adding-custom-styles#arbitrary-properties) for more information on the expected syntax for "arbitrary properties")

---
Screenshots from Chrome devtools:

**With a space:**
CleanShot 2023-05-25 at 09 49 20@2x

**Without a space:**
CleanShot 2023-05-25 at 09 50 51@2x


chrismccord:
❤️❤️❤️🐥🔥

PR #5474, opened 2023-05-28
Fix hero icons mask repeat

aiwaiwa:
Hero icons begin to repeat vertically or horizontally when calculated space shrinks by the layout rules. `mask-repeat` rule set to `no-repeat` fixes these artifacts.

I found this useful but if you think it's too restrictive and rather should be done by tweaking the layout itself, please discard.


chrismccord:
❤️❤️❤️🐥🔥

PR #5481, opened 2023-06-01
Replace 'not found' with 'internal server error' in assert_error_sent/2 examples

### LINKED ISSUE #5480: Documentation for `assert_error_sent/2` does not match actual behavior
### Environment

* Elixir version (elixir -v): Elixir 1.14.2 (compiled with Erlang/OTP 25)
* Phoenix version (mix deps): 1.7.3
* Operating system: macOS 12.6.3

### Expected behavior

Test as described in the example for [assert_error_sent/2](https://hexdocs.pm/phoenix/1.7.3/Phoenix.ConnTest.html#assert_error_sent/2) to be working

### Actual behavior

```
  1) test GET /users/not-found (VerifyAssertErrorSentWeb.ErrorHTMLTest)
     test/verify_assert_error_sent_web/controllers/error_html_test.exs:16
     expected error to be sent as 404 status, but response sent 404 without error
     code: assert_error_sent(:not_found, fn ->
     stacktrace:
       (phoenix 1.7.3) lib/phoenix/test/conn_test.ex:669: Phoenix.ConnTest.assert_error_sent/2
       test/verify_assert_error_sent_web/controllers/error_html_test.exs:17: (test)
```

### Context

This is introduced by https://github.com/phoenixframework/phoenix/pull/5445, if I revert those changes, my test succeed again. I do not know if `RenderErrors.__catch__/5`, `assert_error_sent/2` or the documentation needs to be changed.

phoebe100:
Replace 'not found' with 'internal server error' in assert_error_sent/2 code examples. 
This could reduce confusion since "not found" errors are no re-raised.
Closes #5480

### DIFF
--- a/lib/phoenix/test/conn_test.ex
+++ b/lib/phoenix/test/conn_test.ex
@@ -642,22 +642,22 @@ defmodule Phoenix.ConnTest do
   by your MyApp.ErrorView.
 
   The function accepts a status either as an integer HTTP status or
-  atom, such as `404` or `:not_found`. The list of allowed atoms is available
+  atom, such as `500` or `:internal_server_error`. The list of allowed atoms is available
   in `Plug.Conn.Status`. If an error is raised, a 3-tuple of the wrapped
   response is returned matching the status, headers, and body of the response:
 
-      {404, [{"content-type", "text/html"} | _], "Page not found"}
+      {500, [{"content-type", "text/html"} | _], "Internal Server Error"}
 
   ## Examples
 
-      assert_error_sent :not_found, fn ->
-        get(build_conn(), "/users/not-found")
+      assert_error_sent :internal_server_error, fn ->
+        get(build_conn(), "/broken/route")
       end
 
-      response = assert_error_sent 404, fn ->
-        get(build_conn(), "/users/not-found")
+      response = assert_error_sent 500, fn ->
+        get(build_conn(), "/broken/route")
       end
-      assert {404, [_h | _t], "Page not found"} = response
+      assert {500, [_h | _t], "Internal Server Error"} = response
   """
   @spec assert_error_sent(integer | atom, function) :: {integer, list, term}
   def assert_error_sent(status_int_or_atom, func) do


luka-TU:
@phoebe100 sorry, I thought, generally speaking, hitting a missing route should result in a 404 and not in `Internal Server Error`?

phoebe100:
> hitting a missing route should result in a 404

True, and a running phoenix application will do that.

But when testing controllers, you should differentiate:
a) If you want to verify, that you get back a 404 for a non-existing route, simply use:
```
conn = get(build_conn(), "/users/not-found")
assert response(conn, 404)
```
b) If you want to verify, that some **error** was raised while processing the route and that it was linked to an expected status code, then use `assert_error_sent/2`

Since requesting unknown routes do not result in (re-)raising an **error** anymore, `assert_error_sent/2`is not the right function for testing unknown routes. This commit just tries to clean up the example code in the docs, so that people are not tempted to use that function for testing unknown routes.

luka-TU:
@phoebe100 I think I get what you are trying to say. What throws me off the the name of the method: `assert_error_sent`: it suggests that we should expect and error be `sent` when in reality it is to check if an error was `raised`. In python this naming is a bit more to the point imho, with `assertRaises`.

phoebe100:
> it suggests that we should expect and error be sent when in reality it is to check if an error was raised

Actually, it's both. The intention of the function is to assert:

- that an error was raised
- that the error is wrapped in a response with a specific HTTP status 
- and a specific content (you can pattern match on the returned response tuple as mentioned in the docs) 

So it's all about asserting error translation.

luka-TU:
@phoebe100 this is just my opinion, but not getting an error raised for a 404 in combination with this method name do cause some confusion.

luka-TU:
@stefanchrobot the more detailed the docs are - the better, agree. At the same time, maybe the naming of these methods could have been a bit better, borrowing from some other languages like in the example above.

chrismccord:
❤️❤️❤️🐥🔥

Issue #5415, opened 2023-04-13
Open redirect in Phoenix.Controller


kubaugustyn:
### Environment

* Elixir version (elixir -v): 1.13.0-otp-23
* Phoenix version (mix deps): 1.5.13
* Operating system: macos

### Expected behavior
400 on redirect when passing `return_to = "/%09/evil.com"` in Phoenix.Controller:

      conn
      |> redirect(to: return_to)

### Actual behavior

Redirect to evil.com

chrismccord:
Thanks for the report! Looks like this only happens in chrome when passing something like `localhost:4000/test?redirect=/%09/evil.com`, so others browsers escape or munge it differently, which is interesting.

PR #5482, opened 2023-06-01
Allow local redirect to paths containing percent encoded characters (with few exceptions)

follmann:
With the phoenix `1.7.3` release  a few tests started failing for a project suddenly (`1.7.2` works just fine).

Digging into it, a recent fix https://github.com/phoenixframework/phoenix/commit/5d737201e09adcdfbea0b95b341aaa93e092c869 for the issue #5415 introduced a problem for us. 

We allow a wider palette of Unicode characters in our URLs and as they get percent encoded we have (for our purpose) valid URLs that can look like that in the browsers address bar:

`http://localhost:4000/Поездка-на-лыжах/:id`

And when copy & pasted from the address bar it's apparent that they are actually percent encoded where required:

`http://localhost:4000/%D0%BF%D0%BE%D0%B5%D0%B7%D0%B4%D0%BA%D0%B0-%D0%BD%D0%B0-%D0%BB%D1%8B%D0%B6%D0%B0%D1%85/:id`

In essence the provided fix deems percent encoded control characters (the range starting with %0 / %1) of the URL for a local redirect invalid, but allows to pass other percent encoded characters. 

I tried to keep the fix as minimal as possible, a regex to test for invalid characters seemed like the more concise approach. If an explicit set of characters is preferred I can adapt the PR accordingly. Do also let me know if the test coverage suffices. 

Thanks!

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -504,11 +504,11 @@ defmodule Phoenix.Controller do
     end
   end
 
-  @invalid_local_url_chars ["\\", "/%", "/\t"]
+  @invalid_local_url_regex ~r/\\|\t|\/%[0-1][0-9A-F]/
   defp validate_local_url("//" <> _ = to), do: raise_invalid_url(to)
 
   defp validate_local_url("/" <> _ = to) do
-    if String.contains?(to, @invalid_local_url_chars) do
+    if String.match?(to, @invalid_local_url_regex) do
       raise ArgumentError, "unsafe characters detected for local redirect in URL #{inspect(to)}"
     else
       to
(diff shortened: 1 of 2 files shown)


chrismccord:
We've had some churn in this area and it's not clear to me if this will be handled properly as it relates to the original vuln. For this specific case, I think you are better off building the redirect yourself, ie:

```elixir
  def my_redirect(conn, opts) when is_list(opts) do
    href = Keyword.fetch!(opts, href)
    html = Plug.HTML.html_escape(href)
    body = "<html><body>You are being <a href=\"#{html}\">redirected</a>.</body></html>"

    conn
    |> put_resp_header("location", href)
    |> send_resp(conn.status || 302, "text/html", body)
  end
```

Thanks!

PR #5485, opened 2023-06-06
Address three minor typos in the JSON and APIs guide.

jhayden:
I propose the following three changes to the JSON and APIs guide.


mcrumm:
Thanks!

Issue #5326, opened 2023-03-04
[channel_test] [upgrade 1.6 -> 1.7] unable to connect to socket within "non-test" process


verrchu:
### Environment
```

Erlang/OTP 24 [erts-12.3.2.6] Elixir (1.12.3)

# mix.lock (phoenix-related)
  "phoenix": {:hex, :phoenix, "1.7.1", "a029bde19d9c3b559e5c3d06c78b76e81396bedd456a6acedb42f9c7b2e535a9", [:mix], [{:castore, ">= 0.0.0", [hex: :castore, repo: "hexpm", optional: false]}, {:jason, "~> 1.0", [hex: :jason, repo: "hexpm", optional: true]}, {:phoenix_pubsub, "~> 2.1", [hex: :phoenix_pubsub, repo: "hexpm", optional: false]}, {:phoenix_template, "~> 1.0", [hex: :phoenix_template, repo: "hexpm", optional: false]}, {:phoenix_view, "~> 2.0", [hex: :phoenix_view, repo: "hexpm", optional: true]}, {:plug, "~> 1.14", [hex: :plug, repo: "hexpm", optional: false]}, {:plug_cowboy, "~> 2.6", [hex: :plug_cowboy, repo: "hexpm", optional: true]}, {:plug_crypto, "~> 1.2", [hex: :plug_crypto, repo: "hexpm", optional: false]}, {:telemetry, "~> 0.4 or ~> 1.0", [hex: :telemetry, repo: "hexpm", optional: false]}, {:websock_adapter, "~> 0.4", [hex: :websock_adapter, repo: "hexpm", optional: false]}], "hexpm", "ea9d4a85c3592e37efa07d0dc013254fda445885facaefddcbf646375c116457"},
  "phoenix_ecto": {:hex, :phoenix_ecto, "4.4.0", "0672ed4e4808b3fbed494dded89958e22fb882de47a97634c0b13e7b0b5f7720", [:mix], [{:ecto, "~> 3.3", [hex: :ecto, repo: "hexpm", optional: false]}, {:phoenix_html, "~> 2.14.2 or ~> 3.0", [hex: :phoenix_html, repo: "hexpm", optional: true]}, {:plug, "~> 1.9", [hex: :plug, repo: "hexpm", optional: false]}], "hexpm", "09864e558ed31ee00bd48fcc1d4fc58ae9678c9e81649075431e69dbabb43cc1"},
[...]

PR #5488, opened 2023-06-15
Allow using socket/connect in another process.

Hermanverschooten:
As proposed by @josevalim in #5326, I added an optional 4th argument to `socket/3` and `connect/3` that 
allows to give the `pid` of the test process.  This is needed when wanting to use the socket from a `Task` or 
other process that is not  your test.

### DIFF
--- a/lib/phoenix/test/channel_test.ex
+++ b/lib/phoenix/test/channel_test.ex
@@ -274,7 +274,7 @@ defmodule Phoenix.ChannelTest do
   end
 
   defp fetch_test_supervisor!(options) do
-    case ExUnit.OnExitHandler.get_supervisor(Keyword.get(options, :test_process,self())) do
+    case ExUnit.OnExitHandler.get_supervisor(Keyword.get(options, :test_process, self())) do
       {:ok, nil} ->
         opts = [strategy: :one_for_one, max_restarts: 1_000_000, max_seconds: 1]
         {:ok, sup} = Supervisor.start_link([], opts)
--- a/lib/phoenix/test/channel_test.ex
+++ b/lib/phoenix/test/channel_test.ex
@@ -197,7 +197,7 @@ defmodule Phoenix.ChannelTest do
   Use this function when you want to create a blank socket
   to pass to functions like `UserSocket.connect/3`.
 
-  Otherwise, use `socket/3` if you want to build a socket with
+  Otherwise, use `socket/4` if you want to build a socket with
   existing id and assigns.
 
   ## Examples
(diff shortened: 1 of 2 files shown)


lib/phoenix/test/channel_test.ex
```diff
@@ -257,8 +273,8 @@ defmodule Phoenix.ChannelTest do
     end
   end
 
-  defp fetch_test_supervisor!() do
-    case ExUnit.OnExitHandler.get_supervisor(self()) do
+  defp fetch_test_supervisor!(options) do
+    case ExUnit.OnExitHandler.get_supervisor(Keyword.get(options, :test_process,self())) do
```
josevalim, line 277:
```suggestion
    case ExUnit.OnExitHandler.get_supervisor(Keyword.get(options, :test_process, self())) do
```

lib/phoenix/test/channel_test.ex
```diff
@@ -276,13 +292,13 @@ defmodule Phoenix.ChannelTest do
   @doc false
   @deprecated "Phoenix.ChannelTest.socket/0 is deprecated, please call socket/1 instead"
   defmacro socket() do
-    socket(nil, nil, [], __CALLER__)
+    socket(nil, nil, [],[], __CALLER__)
```
josevalim, line 295:
```suggestion
    socket(nil, nil, [], [], __CALLER__)
```

lib/phoenix/test/channel_test.ex
```diff
@@ -276,13 +292,13 @@ defmodule Phoenix.ChannelTest do
...
   end
 
   @doc false
-  @deprecated "Phoenix.ChannelTest.socket/2 is deprecated, please call socket/3 instead"
+  @deprecated "Phoenix.ChannelTest.socket/2 is deprecated, please call socket/4 instead"
   defmacro socket(id, assigns) do
-    socket(nil, id, assigns, __CALLER__)
+    socket(nil, id, assigns,[], __CALLER__)
```
josevalim, line 301:
```suggestion
    socket(nil, id, assigns, [], __CALLER__)
```

lib/phoenix/test/channel_test.ex
```diff
@@ -291,21 +307,21 @@ defmodule Phoenix.ChannelTest do
   Useful for testing UserSocket authentication. Returns
   the result of the handler's `connect/3` callback.
   """
-  defmacro connect(handler, params, connect_info \\ quote(do: %{})) do
+  defmacro connect(handler, params, connect_info \\ quote(do: %{}), options \\ []) do
```
josevalim, line 310: I find two optional arguments a smell, so what if we do this:

```
  defmacro connect(handler, params, options \\ quote(do: [])) do
```

And then `__connect__` we do:

```
  {connect_info, options} =
    if is_map(options) do
      IO.warn("Passing \"connect_info\" directly to connect/3 is deprecated, please pass \"connect_info: ...\" as an option instead")
      {options, []}
    else
      Keyword.pop(options, :connect_info, %{})
    end
```

josevalim:

Issue #5486, opened 2023-06-07
Flash not showing correctly


moogle19:
### Environment

* Elixir version (elixir -v): 1.14.5
* Phoenix version (mix deps): 1.7.3
* Operating system: macOS

### Expected behavior
When I display an error and an info flash and dismiss the last one, the other should still be there (or maybe both should show?)

### Actual behavior
When I show both error and info and dismiss the latest one, the flash assign for the other one is still there, but the flash message isn't show. 
If I call `put_flash` again, nothing is shown.

https://github.com/phoenixframework/phoenix/assets/792046/0794d6ab-e5c1-4913-abfd-f51a25bb7530

<details>
  <summary>Example Code</summary>

  ```elixir
   defmodule PhxTestWeb.TestLive do
     use PhxTestWeb, :live_view

    @impl true
    def mount(_params, _session, socket) do
      {:ok, socket}
    end

    @impl true
    def handle_event("show-info", _params, socket) do
      {:noreply, put_flash(socket, :info, "Some info message")}
    end

    def handle_event("show-error", _params, socket) do
      {:noreply, put_flash(socket, :error, "Some error message")}
    end

    @impl true
    def render(assigns) do
      ~H"""
      <div>
        <%= inspect(@flash) %>
      </div>
      <div>
        <.button phx-click={JS.push("show-info")}>Show Info</.button>
        <.button phx-click={JS.push("show-error")}>Show Error</.button>
      </div>
      """
    end
  end

  ```
</details>

phoebe100:
The generated `flash_group` component found in `core_components.ex` uses the same id for the <div> elements containing the flash content. As a result, as soon as you click on the top flash message, all flash messages below will be hidden and you cannot recover from that.

You can fix it by adjusting the code for `flash_group/1` in `core_components.ex`:
For example, add custom IDs by replacing
```
<.flash kind={:info} title="Success!" flash={@flash} />
<.flash kind={:error} title="Error!" flash={@flash} />
```
with
```
<.flash id="flash_info" kind={:info} title="Success!" flash={@flash} />
<.flash id="flash_error" kind={:error} title="Error!" flash={@flash} />
```

michallepicki:
Hi @moogle19 , can you check if https://github.com/phoenixframework/phoenix/commit/10891447eddc023bd79e1b9f8b46f8a3344542bb resolved this issue?

PR #5490, opened 2023-06-19
Add unique ids flashes in flash_group

begedin:
This should solve #5486 

# Description

The user-provided flashes for `:info` and `:error` in the `<.flash_group />` component render using the `<flash />` component, but do not specify an id, causing the default id of `"flash"` to be used for both.

Since the logic around flashes allow for both info and error to be present, this may result in neither rendering, or one rendering over another.

Providing a unique ID results in things working in a less confusing way.

This updates the generator for core components to provide this unique id.

# Notes

- I'm not aware of a rule for specifying ids. If these do not match an expected format, I'm happy to update 
- The solution could be smaller - only one of the two flashes could be given an id, allowing the other to use the default
- It could also be solved by adding some  logic to the component to ensure only one flash could be rendered.

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -143,8 +143,8 @@ defmodule <%= @web_namespace %>.CoreComponents do
 
   def flash_group(assigns) do
     ~H"""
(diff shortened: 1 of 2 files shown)


phoebe100 (commented):
Looks good to me - thank you

PR #5495, opened 2023-06-25
Export static `at` path from Endpoint

### LINKED ISSUE #5387: `No route path` warning changing the static assets path
### Environment

* Elixir version (elixir -v):
```
Erlang/OTP 25 [erts-13.2] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [jit:ns]

Elixir 1.14.3 (compiled with Erlang/OTP 23)
```
* Phoenix version (mix deps): `{:phoenix, "~> 1.7.2"}`
* Operating system: `macOS 11.7.3`

### Expected behavior

If I want to change the path where the static assets will be requested, I want to be able to compile the project without warnings because of no route paths. Steps to reproduce in a new project with the latest phx_new version:
1. `mix archive.install hex phx_new`
2. `mix phx.new test_warning_sigil --no-ecto`
3. `mix setup`
4. Modify the `Plug.Static` in the `lib/test_warning_sigil_web/endpoint.ex` file to:
```
  plug(Plug.Static,
    at: "/custom_path",
    from: :test_warning_sigil,
    gzip: false,
    only: TestWarningSigilWeb.static_paths()
  )
```
5. Update the `lib/test_warning_sigil_web/components/layouts/root.html.heex` file with:
```
    ...
    <link phx-track-static rel="stylesheet" href={~p"/custom_path/assets/app.css"} />
    <script defer phx-track-static type="text/javascript" src={~p"/custom_path/assets/app.js"}>
   ...
```

### Actual behavior

If I follow the previous steps, I can run the application and the static files looks to be properly loaded, but I get those warnings:
```
warning: no route path for TestWarningSigilWeb.Router matches "/custom_path/assets/app.js"
  lib/test_warning_sigil_web/components/layouts/root.html.heex:11: TestWarningSigilWeb.Layouts.root/1

warning: no route path for TestWarningSigilWeb.Router matches "/custom_path/assets/app.css"
  lib/test_warning_sigil_web/components/layouts/root.html.heex:10: TestWarningSigilWeb.Layouts.root/1
```

alisinabh:
Closes #5387

Static plug has a configuration called `at` which the `VerifiedRoutes` does not take into account since it is not aware of it.

This is one possible solution that export the configuration from the `XxxWeb.Endpoint` to `XxxWeb` as a function called `static_at/0`. Later the `VerifiedRoute` module makes use of this with accepting a new config.

Also this PR can use some recommendation regarding the naming of the configuration option `static_at` if it is decided an acceptable approach.

Thanks!

### DIFF
--- a/installer/templates/phx_single/lib/app_name_web.ex
+++ b/installer/templates/phx_single/lib/app_name_web.ex
@@ -18,6 +18,7 @@ defmodule <%= @web_namespace %> do
   """
 
   def static_paths, do: ~w(assets fonts images favicon.ico robots.txt)
+  def static_at, do: "/"
 
   def router do
     quote do
@@ -100,7 +101,8 @@ defmodule <%= @web_namespace %> do
       use Phoenix.VerifiedRoutes,
         endpoint: <%= @endpoint_module %>,
         router: <%= @web_namespace %>.Router,
-        statics: <%= @web_namespace %>.static_paths()
+        statics: <%= @web_namespace %>.static_paths(),
+        statics_at: <%= @web_namespace %>.static_at()
     end
   end
 
--- a/installer/templates/phx_umbrella/apps/app_name_web/lib/app_name.ex
+++ b/installer/templates/phx_umbrella/apps/app_name_web/lib/app_name.ex
@@ -18,6 +18,7 @@ defmodule <%= @web_namespace %> do
   """
 
   def static_paths, do: ~w(assets fonts images favicon.ico robots.txt)
+  def static_at, do: "/"
 
   def router do
     quote do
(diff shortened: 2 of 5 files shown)


chrismccord:
Thanks for this! I think perhaps a doc PR is best for this one because you only need to opt-in to something like this if you change the default setup. So I don't think it's worth adding another function out of the box.

alisinabh:
Thanks for the review. Honestly, I don't understand how without changing `VerifiedRoutes` module this can be fixed by the user. If you can share with me how you imagined it I will take care of the docs.

hoarf:
Same issue here

PR #5496, opened 2023-06-26
fix auth alias order

ayrat555:
credo's readability check catches it as a warning

image


chrismccord:
I am 👎 on alpha ordering. Thanks!

PR #5498, opened 2023-06-28
Clarify CoreComponents moduledoc

Cellane:
When reading through `CoreComponents` `@moduledoc`, I thought that perhaps some of the sentences could use small adjustments – though my English is not very good so I’m not sure if I actually improved anything.

I found the same string in two different files so I made the same changes to both, but am not sure if that’s required. Please let me know your thoughts 🙏


chrismccord:
❤️❤️❤️🐥🔥

PR #5503, opened 2023-07-03
Update `Telemetry` installation template

jbcaprell:
I think this is a typo, that the actual event is `[:phoenix, :channel_joined]`.

https://github.com/phoenixframework/phoenix/blob/3f0dfced31a66edb0932c479e2da51125d1c8e30/lib/phoenix/logger.ex#L137
https://github.com/phoenixframework/phoenix/blob/3f0dfced31a66edb0932c479e2da51125d1c8e30/lib/phoenix/logger.ex#L333


chrismccord:
❤️❤️❤️🐥🔥

PR #5506, opened 2023-07-05
Adds radio input case to `input` function definition in `CoreComponents`

neurodynamic:
This code is a slightly modified version of @codeanpeace's work here: https://elixirforum.com/t/radio-buttons-using-the-corecomponents-module-in-phoenix-1-7/56856/5

That thread illustrates the pitfalls that can happen with radio inputs omitted from the `input` function. If it does not make sense to handle the radio input case this way (since it is a bit of an odd-input-out in terms of how it works), then an alternative to handling it might be to add some sort of explicit warning or error that pops up for developers who try to use the `input` function to make a radio input.

### DIFF
--- a/priv/templates/phx.gen.live/core_components.ex
+++ b/priv/templates/phx.gen.live/core_components.ex
@@ -319,6 +319,23 @@ defmodule <%= @web_namespace %>.CoreComponents do
     """
   end
 
+  def input(%{type: "radio"} = assigns) do
+    ~H"""
+      <label class="flex items-center gap-2 text-sm leading-6 text-zinc-600">
+        <input
+          type="radio"
+          id={@id}
+          name={@name}
+          value={@value}
+          checked={@checked}
+          class="rounded border-zinc-300 text-zinc-900 focus:ring-0"
+          {@rest}
+        />
+        <%= @label %>
+      </label>
+    """
+  end
+
   def input(%{type: "select"} = assigns) do
     ~H"""
     <div phx-feedback-for={@name}>


lessless:
Does this also fix the broken layout in the default component? 

image

josevalim:
An error to say it is not supported works of me, yes. Because I don't think generating a single radio will help either. :)

lessless:
Can't it be a basic, indivisible building block, as an atom in Atomic Design methodology, for example?

chrismccord:
If folks want a radio button function component in their own apps, they could do something like this:

```eliixr
  @doc """
  Provides a radio group input for a given form field.

  ## Examples

      <.radio_group field={@form[:tip]}>
        <:radio value="0">No Tip</:radio>
        <:radio value="10">10%</:radio>
        <:radio value="20">20%</:radio>
      </.radio_group>
  """
  attr :field, Phoenix.HTML.FormField, required: true

  slot :radio, required: true do
    attr :value, :string, required: true
  end

  slot :inner_block

  def radio_group(assigns) do
    ~H"""
    <div>
      <%= render_slot(@inner_block) %>
      <div :for={{%{value: value} = rad, idx} <- Enum.with_index(@radio)}>
        <label for={"#{@field.id}-#{idx}"}><%= render_slot(rad) %></label>
        <input
          type="radio"
          name={@field.name}
          id={"#{@field.id}-#{idx}"}
          value={value}
          checked={to_string(@field.value) == to_string(value)}
          class="rounded-lg text-zinc-900 focus:ring-0 sm:text-sm sm:leading-6"
        />
      </div>
    </div>
    """
  end
```

matteosantama:
I'm new to Phoenix and Elixir, so perhaps I don't fully appreciate the issues at play here, but I am quite surprised to see something as fundamental as radio buttons are not natively supported. Is it a Herculean effort to properly support them? I am happy to take a stab myself if there are suggestions

josevalim:
@matteosantama it is not complicated, adding the code above by Chris would suffice. However, we try to keep the core components file as slim as possible and, because none of Phoenix default generators use radio buttons, we decided to not include one in core components. But you are more than welcome to add the missing pieces to your app.

PR #5514, opened 2023-07-10
Add eager errors option to core component input fields

FeeJai:
I had an issue when building an application with both timezone and calendar selection, that the date could become invalid by changing the timezone without the date picker being touched. The error was not shown, since the `phx-feedback-for` attribute was doing it's job. I have added a feedback slot, that allows to either disable feedback altogether, or set it to eager so it will always be displayed when available. Default does not change current behaviour

This was the question I asked in the form ahead of writing the patch: https://elixirforum.com/t/how-does-liveview-keep-track-of-which-forms-fields-have-been-altered/56014


FeeJai:
Ping?

chrismccord:
This looks like a great option for your app, but not something I think we need to ship out of the box. The core components are meant as starting point and to be adapted to your needs. Thanks!

PR #5515, opened 2023-07-12
Fix default template path in channel guide

solar05:


chrismccord:
the current guide is correct. We generate a `controllers/page_html/home.html.heex` template as of phoenix 1.7. Thank you!

PR #5517, opened 2023-07-15
Remove error_html_test on new without html

hpiaia:
When generating projects without html in umbrella (`mix phx.new my-app --umbrella --no-html`), the file `error_html_test.exs` still get created.

If you run `mix test` on a newly created project, it fails because it expects html.

I fixed that by moving the generation of that file from the "new" template to the "html" template. I updated the tests to match the expected behaviour. All other tests are passing fine.

I also took advantage of the same commit to fix a typo of `pager_controller_test.exs`, which should be `page_controller_test.exs`.

Cheers from Brazil, @chrismccord @josevalim!   🧡


chrismccord:
❤️❤️❤️🐥🔥

PR #5521, opened 2023-07-17
Dockefile template: decouple builder and runner debian versions

bettio:
hex.pm images are not updated as frequntly as Debian official images, however for security reasons latest debian version should be used for runners since they may contain important bugfixes.

This change allows using a more recent runner version while using the same Debian release for both builder and runner.

### DIFF
--- a/priv/templates/phx.gen.release/Dockerfile.eex
+++ b/priv/templates/phx.gen.release/Dockerfile.eex
@@ -13,10 +13,12 @@
 #
 ARG ELIXIR_VERSION=<%= elixir_vsn %>
 ARG OTP_VERSION=<%= otp_vsn %>
-ARG DEBIAN_VERSION=<%= debian %>-<%= debian_vsn %>-slim
+ARG DEBIAN_RELEASE=<%= debian %>
+ARG DEBIAN_VERSION=<%= debian_vsn %>-slim
+ARG DEBIAN_RUNNER_VERSION=<%= debian_vsn %>-slim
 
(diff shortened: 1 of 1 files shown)


bettio:
Additional comment here: the are some other files that make use of the same general template (such as guides/deployment/releases.md), before updating all of them let's discuss and review this first step.

Gazler (approved):
Took me a moment to work out what's actually going on here until I ran the generator.

The output before was something like:

```
ARG DEBIAN_VERSION=bullseye-20230612-slim

ARG BUILDER_IMAGE="hexpm/elixir:${ELIXIR_VERSION}-erlang-${OTP_VERSION}-debian-${DEBIAN_VERSION}"
ARG RUNNER_IMAGE="debian:${DEBIAN_VERSION}"
```

After it will be:

```
ARG DEBIAN_RELEASE=bullseye
ARG DEBIAN_VERSION=20230612-slim
ARG DEBIAN_RUNNER_VERSION=20230612-slim

RG BUILDER_IMAGE="hexpm/elixir:${ELIXIR_VERSION}-erlang-${OTP_VERSION}-debian-${DEBIAN_RELEASE}-${DEBIAN_VERSION}"
ARG RUNNER_IMAGE="debian:${DEBIAN_RELEASE}-${DEBIAN_RUNNER_VERSION}"
```

josevalim:
We should ping the hex.pm team to update those images quickly when there is a security issue. I believe it has happened in this case, so we can close this one. Thank you!

PR #5522, opened 2023-07-17
Use correct form field name in reset_password_live_test template

jclem:
Currently, the form submits "confirmation_password", which (I believe) could lead to false positives. Instead, this template should use the expected key, "password_confirmation".


chrismccord:
❤️❤️❤️🐥🔥

PR #5523, opened 2023-07-18
Ensure sourcemaps have the .js.map extension

catalinfeier:
Normally, we'd expect sourcemaps to have the .js.map extension but the ones generated by the digester have the following format:
`"app.js-a4ef30371d563677c64cc9583ad6b305.map"`
The hash should be in included in the filename, not in the extension:
`"app-a4ef30371d563677c64cc9583ad6b305.js.map"`

While in some cases, this doesn't matter because the unminifiers would look at the sourceMappingURL in .js file, there are use cases like [datadog](https://docs.datadoghq.com/real_user_monitoring/guide/upload-javascript-source-maps/?tab=webpackjs) where the expected format is having the extension `.js.map`, otherwise they are not recognised as valid sourcemaps and some manual filename manipulation would be needed to adhere to the expected format.

### DIFF
--- a/lib/phoenix/digester.ex
+++ b/lib/phoenix/digester.ex
@@ -146,8 +146,15 @@ defmodule Phoenix.Digester do
     content = File.read!(file_path)
 
     basename = Path.basename(file_path)
-    rootname = Path.rootname(basename)
-    extension = Path.extname(basename)
+
+    rootname =
+      if String.ends_with?(basename, ".js.map"),
+        do: Path.rootname(basename) |> Path.rootname(),
+        else: Path.rootname(basename)
+
+    extension =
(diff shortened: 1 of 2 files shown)


paulgendek (approved):

jmilamwalters (approved):

Iamsomasundaram (approved):

mcrumm:
Hi @catalinfeier, thanks for the PR! Unfortunately I think this change is not enough to satisfy even datadog-ci, as it expects the [asset name to match exactly](https://github.com/DataDog/datadog-ci/blob/1e954d1955e67991af8c13bbb64fb754b06b4f95/src/commands/sourcemaps/utils.ts#L3C1-L10C1) except for the `.map` extension.

PR #5526, opened 2023-07-20
Do something with assigns.errors (otherwise it is just overwitten)

derekkraan:
Another option is to remove the `errors` attr, I'm not sure what was intended.

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -289,7 +289,17 @@ defmodule <%= @web_namespace %>.CoreComponents do
   def input(%{field: %Phoenix.HTML.FormField{} = field} = assigns) do
     assigns
     |> assign(field: nil, id: assigns.id || field.id)
-    |> assign(:errors, Enum.map(field.errors, &translate_error(&1)))
+    |> assign(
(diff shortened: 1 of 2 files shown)


chrismccord:
The idea is that the `field={}` based input populates itself, but you can still pass the manual version to roll things on your own, for example `<.input type="text" name="foobar" errors={["not valid"]} />`. So in this context I don't know what the best option is. `assign_new` could work here as well, but I hadn't considered the case where someone passes a form field (with its own errors) but also wants to pass explicitly some other errors, so I'm not sure. What scenario do you need to pass both?

derekkraan:
I don't personally have a scenario where both are needed.

If the intention is that `assigns.errors` will take precedence over `assigns.field.errors` then I can update the PR to make it work that way.

The current code just ignores `assigns.errors` altogether. (Aside: I only noticed this after another dev I work with tried to fix the component by using `assigns.errors || field.errors`, which then causes `field.errors` to be ignored)

chrismccord:
I think it makes sense for the `field` version of the input to use the errors from the form field, so I'm not convinced this change is what we want by default. That said, the components are yours to customize so this change absolutely fine as you described. Thank you!

PR #5527, opened 2023-07-21
Ensure sourcemaps are digested with expected hash/extension

mcrumm:
[Sourcemaps][0] are usually generated alongside the asset they describe, and due to this many tools assume that for an asset like app-{hash}.js the sourcemap will be found at app-{hash}.js.map. This change ensures that digested sourcemaps will have paths matching their corresponding assets.

[0]: https://web.dev/source-maps/

Closes #5523

### DIFF
--- a/lib/phoenix/digester.ex
+++ b/lib/phoenix/digester.ex
@@ -23,6 +23,7 @@ defmodule Phoenix.Digester do
       File.mkdir_p!(output_path)
 
       files = filter_files(input_path)
+      files = fixup_sourcemaps(files)
       latest = generate_latest(files)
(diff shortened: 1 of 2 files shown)


mcrumm:
The code's not very pretty but I wanted to validate the concept before I refactored the file mapping :)

chrismccord:
❤️❤️❤️🐥🔥

fishtreesugar:
Hi, we've discovered that our sourcemap upload script broke after upgrading Phoenix from 1.7.7 to 1.7.8. This issue wasn't mentioned in the changelog.

It would be beneficial to include this in the changelog for future reference.

PR #5529, opened 2023-07-25
Add empty state to table

bkilshaw:
This PR ads an `empty_message` attribute to the `table` core component making it easy to display an empty-state message when a table has no records. 

```html
 <.table
  id="posts"
  rows={@streams.posts}
  empty_message="Get started by clicking the 'New Post' button above!"
>
  <:col :let={{_id, post}} label="Name"><%= post.name %></:col>
  <:col :let={{_id, post}} label="Date"><%= post.date %></:col>
</.table>
```


chrismccord:
This is a nice addition, but we have to balance keeping the core components minimal so I don't think it quite makes the cut. Thanks!

PR #5531, opened 2023-07-27
`destroy()` method for `Presence`

VimHax:
From what I have observed so far, it seems that you cannot stop tracking the presence for a channel. This feature would be useful in the case where you manage the `Presence` independently from the `Channel` in your code because `Presence` can be created and destroyed by the code which is managing the `Presence`. Otherwise you would need to pass in the `Presence` from whatever code is managing the `Channel`.

```js
class ManagePresence {
  constructor(channel){
    this.presence = new Presence(channel)
  }
  
  destroy(){
    this.presence.destroy()
  }
}
```

This feature also allows code which tracks the client presence state in parallel to `Presence` work nicer during reconnects because, without `destroy`, the presence state would remain intact over re-connections. (Meaning if many clients had previously connected but then there was a server reboot there would be unexpected `client_leave`s in the history when the `Presence` re-syncs since the other clients won't be present anymore.

```js
class ManagePresence {
  constructor(channel){
    this.history = []
    this.presence = new Presence(channel)
    this.presence.onJoin(id => history.push({type: 'client_join', id, timestamp: new Date()}))
    this.presence.onLeave(id => history.push({type: 'client_leave', id, timestamp: new Date()}))
  }
  
  destroy(){
    this.presence.destroy()
  }
}
```

### DIFF
--- a/assets/js/phoenix/presence.js
+++ b/assets/js/phoenix/presence.js
@@ -8,6 +8,7 @@ export default class Presence {
 
   constructor(channel, opts = {}){
     let events = opts.events || {state: "presence_state", diff: "presence_diff"}
+    this.refs = []
     this.state = {}
     this.pendingDiffs = []
     this.channel = channel
@@ -18,8 +19,7 @@ export default class Presence {
       onSync: function (){ }
     }
 
-    this.refs = []
-    this.refs.push(this.channel.on(events.state, newState => {
+    this.refs.push([events.state, this.channel.on(events.state, newState => {
       let {onJoin, onLeave, onSync} = this.caller
 
       this.joinRef = this.channel.joinRef()
@@ -30,9 +30,9 @@ export default class Presence {
       })
       this.pendingDiffs = []
       onSync()
-    }))
+    })])
 
-    this.refs.push(this.channel.on(events.diff, diff => {
+    this.refs.push([events.diff, this.channel.on(events.diff, diff => {
       let {onJoin, onLeave, onSync} = this.caller
 
       if(this.inPendingSyncState()){
@@ -41,7 +41,7 @@ export default class Presence {
         this.state = Presence.syncDiff(this.state, diff, onJoin, onLeave)
         onSync()
       }
-    }))
+    })])
   }
 
   onJoin(callback){ this.caller.onJoin = callback }
@@ -57,7 +57,7 @@ export default class Presence {
   }
 
   destroy(){
-    this.refs.forEach(ref => this.channel.off(ref))
+    this.refs.forEach(([event, ref]) => this.channel.off(event, ref))
   }
 
   // lower-level public static API
--- a/assets/js/phoenix/presence.js
+++ b/assets/js/phoenix/presence.js
@@ -18,7 +18,8 @@ export default class Presence {
       onSync: function (){ }
     }
 
-    this.channel.on(events.state, newState => {
+    this.refs = []
+    this.refs.push(this.channel.on(events.state, newState => {
       let {onJoin, onLeave, onSync} = this.caller
 
       this.joinRef = this.channel.joinRef()
@@ -29,9 +30,9 @@ export default class Presence {
       })
       this.pendingDiffs = []
       onSync()
-    })
+    }))
 
-    this.channel.on(events.diff, diff => {
+    this.refs.push(this.channel.on(events.diff, diff => {
       let {onJoin, onLeave, onSync} = this.caller
 
       if(this.inPendingSyncState()){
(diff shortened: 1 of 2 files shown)


chrismccord:
In this case the server is still going to be sending you the diffs, so I'm not following completely how this solves your issue. You can filter on the client and noop if needing, or signal the channel to untrack, no? Thanks!

VimHax:
My main problem with the current implementation is that there is no way to make the `Presence`s unsubscribe from the channel to my knowledge. For my above use-case, I can perform the following solution:
```ts
class ClientHistory {
  constructor(channel) {
    this.history = []
    this.destroyed = false;
    this.presence = new Presence(channel)
    this.presence.onJoin(id => {
      if (this.destroyed) return;
      history.push({type: 'client_join', id, timestamp: new Date()});
    })
    this.presence.onLeave(id => {
      if (this.destroyed) return;
      history.push({type: 'client_leave', id, timestamp: new Date()});
    })
  }
  
  destroy() {
    this.destroyed = true;
    // this.presence.destroy() does not exist
  }
}

class Project {
  constructor(channel) {
    this.clientHistory = new ClientHistory(channel);
  }

  get history() {
    return this.clientHistory.history;
  }

  destroy() {
    this.clientHistory.destroy();
  }
}
```
And this can be used as expected:
```ts
let project = null;
const projectChannel = socket.channel('project-abc');

function destroy() {
  console.log(project.history); // Correctly logs every client that has joined and left during the session.
  project.destroy();
  project = null;
}

projectChannel.onError(destroy);
projectChannel.onClose(destroy);

projectChannel.join().receive('ok', () => {
  project = new Project(projectChannel);
});
```
Where the `Project` only exists when it's currently connected to the channel. However the problem with the above implementation is that, because there's no `destroy` or any other method which unsubscribes from the channel there is a memory leak.
```ts
// Channel#on
on(event, callback){
  let ref = this.bindingRef++
  this.bindings.push({event, ref, callback})
  return ref
}
```

chrismccord:
I'm sorry I am am still not following. The channel is subscribed on the server and sends the diffs, so you can either 1) leave the channel, or 2) if you want to stay joined and receive other messgaes but stop receiving presence, you can message the channel and call untrack on the server. You can also filter the messages on the client in `onSync` and noop based on some app state if you want to ignore the presences but still receive them at will. What are the reasons that 1 or 2 are not options for your case? Thanks!

VimHax:
> I'm sorry I am am still not following. The channel is subscribed on the server and sends the diffs, so you can either 1) leave the channel, or 2) if you want to stay joined and receive other messgaes but stop receiving presence, you can message the channel and call untrack on the server. You can also filter the messages on the client in `onSync` and noop based on some app state if you want to ignore the presences but still receive them at will. What are the reasons that 1 or 2 are not options for your case? Thanks!

Thanks for the reply and for the hard work you're doing for these projects! My main use-case here is for automatic reconnecting. (So Solution: 1 is not what I'm looking for) I basically want the presence state to be cleared when disconnected and start fresh when reconnected. The reason for that is because when the channel reconnects and the new presence state is received the presence sync will diff the old clients against the new ones (which is the behavior I don't want), instead I want the presence to consider only this new connection and the clients connected instead of the old clients (as if the presence state was cleared). The way I achieve this in my above example code is by creating a new presence object every time the channel connects, ensuring that the presence state is always fresh/empty on reconnect.

Even though Solution: 2 is not what I'm looking for, would calling untrack on the server remove the listeners from the channel on the client side? From my knowledge I believe this is not the case, so the application would need to keep re-using the same presence instead of being able to create and destroy them on the fly. (Or suffer a slight memory leak by the old lingering listeners) Adding the `destroy` method (or perhaps even a `clear` method or both) in my opinion makes achieving this behavior nicer and gives more flexibility to the API; therefore more ways it can be used.

chrismccord:
When the channel reconnects and resyncs presence state, it will be the same as if you created a new channel and presence instance, no? If you are having old presences laying around that are no longer joined then that's a bug, so what is the issue with allowing the rejoin and regular sync to happen? Your presence list will be identical to a fresh channel/presence, unless I'm missing something.

VimHax:
> When the channel reconnects and resyncs presence state, it will be the same as if you created a new channel and presence instance, no? If you are having old presences laying around that are no longer joined then that's a bug, so what is the issue with allowing the rejoin and regular sync to happen? Your presence list will be identical to a fresh channel/presence, unless I'm missing something.

It is true that the presence state will be identical if you follow the diffs, however, that's not exactly my use-case here for presences. (or at least not the entire use-case) In the example I showed, I'm keeping track of a history of events where clients joining and leaving is marked. This history is recorded for every session. (where sessions are periods of connected times, a reconnect would start a new session) When recording the history, the default behavior of the presence object when reconnecting is not desirable because in the new session old client disconnect logs would show up. Now of course it's possible to write some code to workaround this but I just thought that adding this method may perhaps be useful for others, just like it would be useful for me. (This simplifies the implementation by a lot)

PR #5533, opened 2023-07-27
Favor HelloController for showcasing layout changes, closes #5532

### LINKED ISSUE #5532: Guide - Layouts - Using suggested layouts not working as expected
### Environment

* Elixir version (elixir -v): Elixir 1.14.0 (compiled with Erlang/OTP 23)
* Phoenix version (mix deps): v1.7.7
* Operating system: Ubuntu 22.04.2 LTS

### Expected behavior
The [guide](https://hexdocs.pm/phoenix/components.html#layouts) suggest to make a new application layout `admin.html.heex` by copying `app.html.heex` and editing it to remove the header. In order to use the new layout it is suggested to use 
`def home(conn, _params) do
  conn
  |> put_layout(html: :admin)
  |> render(:home)
end`
where the only difference should be that there is no header.

### Actual behavior
What happens instead is that the div with the text and buttons is no longer spaced to the left as initially, but centered and part of it is hidden under the SVG to the right.
Imgur

### Additional comments
This part of the guide as it is written suggests that up until this step `app.html.heex` is used as layout, which is actually not true because the home page is rendered with `layout: false` 
```
def home(conn, _params) do
  # The home page is often custom made,
  # so skip the default app layout.
  render(conn, :home, layout: false)
end
```

phoebe100:
Since the component guide expects that readers have gone through the request life-cycle guide, we can favor HelloController over PageController when it comes to demonstrate some layout adjustments.

Since the PageController does not use app layout after project generation, it is more straightforward to show layout adjustments on the HelloController.

closes #5532

### DIFF
--- a/guides/components.md
+++ b/guides/components.md
@@ -183,27 +183,27 @@ In other words, after rendering your page, the result is placed in the `@inner_c
 
 Phoenix provides all kinds of conveniences to control which layout should be rendered. For example, the `Phoenix.Controller` module provides the `put_root_layout/2` function for us to switch _root layouts_. This takes `conn` as its first argument and a keyword list of formats and their layouts. You can set it to `false` to disable the layout altogether.
 
-You can edit the `home` action of `PageController` in `lib/hello_web/controllers/page_controller.ex` to look like this.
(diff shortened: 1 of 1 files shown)


chrismccord:
❤️❤️❤️🐥🔥

PR #5534, opened 2023-07-27
Change socket serializer/version logs to warning.

aselder:
Often times, random bots or scanners hit URLs with garbage params. When this happens with a Phoenix socket URL, serializer negotiaiton can fail and Phoenix will log this event.

Currently this happens at the `:error` level. In most cases, this is not an error conditions, but an expected failure due to garbage in, garbage out. As such, it's probably more appropriate for this to be at the `:warning` level.


chrismccord:
❤️❤️❤️🐥🔥

PR #5538, opened 2023-08-01
Context Guide: add `+` markers for `get_cart_by_user_uuid` in diff

JuneKelly:
I've just been working through the Contexts guide (https://hexdocs.pm/phoenix/contexts.html), and I feel like the code for `get_cart_by_user_uuid` should have `+` diff markers, like the other functions that are added in the same diff.

See screenshot:
image

The line below the diff even says `We started by implementing get_cart_by_user_uuid/1`, so I feel like this was probably the original intention.

Thanks!


chrismccord:
❤️❤️❤️🐥🔥

PR #5540, opened 2023-08-04
Favor format-specific view module over default

jswanner:
Given a controller with default and format-specific view set, need to check for format-specific first, otherwise default view is always used. That said, since the desire is to deprecate non-format views this change may not be desirable.

Thanks for the amazing framework!

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -589,13 +589,13 @@ defmodule Phoenix.Controller do
   def view_module(conn, format \\ nil) do
     format = format || get_safe_format(conn)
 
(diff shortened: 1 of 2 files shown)


josevalim:
Hi @jswanner, you are right. We want to give priority to `_` to avoid breaking changes which is why it is deprecated. Thanks!

jswanner:
Oh, I see what I was missing...

I was needing to have `use Phoenix.Controller, put_default_views: false` in order to call `put_view(conn, bar: FooBAR)`, but it's easier if I have `use Phoenix.Controller, formats: [:bar]`.

Thanks for your time and consideration, @josevalim.

PR #5542, opened 2023-08-06
Assign generated id to reuse in flash template

chulkilee:
It fixes a bug introduced in 10891447eddc023bd79e1b9f8b46f8a3344542bb, which does not allow to close flash when id is not given.

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -106,10 +106,11 @@ defmodule <%= @web_namespace %>.CoreComponents do
   slot :inner_block, doc: "the optional inner block that renders the flash message"
 
   def flash(assigns) do
+    assigns = assign(assigns, :id, assigns.id || "flash-#{assigns.kind}")
     ~H"""
     <div
       :if={msg = render_slot(@inner_block) || Phoenix.Flash.get(@flash, @kind)}
-      id={@id || "flash-#{@kind}"}
+      id={@id}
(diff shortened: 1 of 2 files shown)


installer/templates/phx_web/components/core_components.ex
```diff
@@ -106,10 +106,12 @@ defmodule <%= @web_namespace %>.CoreComponents do
   slot :inner_block, doc: "the optional inner block that renders the flash message"
 
   def flash(assigns) do
+    assigns = assign(assigns, :id, assigns.id || "flash-#{assigns.kind}")
```
josevalim, line 109: Can we use assign_new instead? :)

chulkilee: This handles assigns has `id` with `nil` value - but `assign_new` would not set the value in that case since it checks whether the key exist or not.

I intentionally used `||` here to avoid having `nil` value for id in any cases.. but should I handle that here? Or should we change `assign_new` to handle that case? hmm..

Anyway, just let me know what you think - I'll update the PR accordingly.

josevalim: I think if you set `id={nil}` I would not expect that to work. As setting it to `id={{:inside_tuple}}`  won't work either.

chulkilee: That makes sense - updated to use `assign_new`

josevalim:

PR #5543, opened 2023-08-08
[Socket Draining] Dynamic socket drainer configuration

studzien:
To configure the socket `drainer`, one must currently set it up during the compile time. I suggest adding an option to pass `{M, F, A}` to return the drainer configuration keyword list, just like the `check_origin` option. This would allow for runtime configuration of the drainer.

However, I could not find any tests related to this feature.

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -807,8 +807,12 @@ defmodule Phoenix.Endpoint do
     * `:code_reloader` - enable or disable the code reloader. Defaults to your
       endpoint configuration
 
-    * `:drainer` - a keyword list configuring how to drain sockets
-      on application shutdown. The goal is to notify all channels (and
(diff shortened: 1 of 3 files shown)


studzien:
Hi @josevalim :wave: 
Since you worked on the socket draining, would you be interested in looking at this PR?
Thank you!

josevalim:
Couldn't you configure it on config/runtime.exs with dynamic values?

studzien:
The drainer can only be configured at compile-time using `socket` in the endpoint. Is there a way to configure it dynamically? I couldn't find a way to do this, so I assumed that's why `check_origin` and `connection_info.session` can be configured via MFAs.

josevalim:
You are right. I thought this was about the endpoint drainer, not the socket one. :)

josevalim:
Can you please add a test?

josevalim:

Issue #5546, opened 2023-08-10
Freshly generated app contains syntax errors in `home.css`


angelikatyborska:
### Environment

* Elixir version (elixir -v): `1.15.0`
* Phoenix version (mix deps): `phx_new-1.7.7`
* Operating system: macOS 13.5

### Actual behavior

After generating a fresh Phoenix app (exact command: ` mix phx.new phoenix_app --no-dashboard --no-ecto --no-gettext --no-mailer --no-tailwind`), I noticed that my IDE is unhappy about the `home.css` file (lots of red squiggles). I passed the file though the [CSS validator](https://jigsaw.w3.org/css-validator/#validate_by_input). It contains roughly 50 identical syntax errors, all custom CSS properties without any value given like this:
```css
--tw-pan-x:  ;
```

<img width="1061" alt="Screenshot 2023-08-10 at 19 27 21" src="https://github.com/phoenixframework/phoenix/assets/7852553/3fd4a57f-3554-4f40-b78b-aa504a2e50b3" alt="My IDE showing red squiggles in the home css file">

### Expected behavior

A freshly generated app has no syntax errors in any of its files.

Can I open a PR that will remove those properties without values, or is that going to break something in Tailwind (whose custom properties those are, I assume)?

I know this is a low priority problem because browsers are extremely forgiving about CSS syntax errors and the file is meant to be deleted, but I still think it's worth fixing because red squiggles in a fresh project might make newcomers to the framework think that _they_ did something wrong.

chrismccord:
Does the browser complain about this file? It is a throwaway file for styling on the "welcome to phoenix" page and nothing else. I'm happy to address this if it's a real problem, but again this isn't supposed to be a file that you really customize so a css linter complaining really isn't an issue imo. If the browsers errors then we should definitely address it. Thanks!

angelikatyborska:
Browsers have no problems ignoring a single line of broken CSS, so it doesn't break anything in terms of rendering. Still, I think it should be fixed if a fix is easily achievable because it might be mistaken by beginners as something they did wrong (because IDEs tend to mark those invalid files in red). I'm also of the opinion that the default generated app, even though it's meant to be thrown away, is a sort of showcase of the framework, so it should still be of high quality, right?

chrismccord:
Yes please send a PR if this is an easy fix. Note that this only applies when you run the `--no-tailwind` flag, so it's not exactly the default generated app. Thanks!

[...]

PR #5547, opened 2023-08-11
Remove invalid CSS from generated home.css

angelikatyborska:
Resolves https://github.com/phoenixframework/phoenix/issues/5546

I checked, the valueless custom CSS properties aren't used anywhere in the `home.css`, and they are technically syntax errors, so let's remove them.


chrismccord:
❤️❤️❤️🐥🔥

Issue #4186, opened 2021-01-27
Invalid Endpoint.url() when using dynamic listening port 


dominicletz:
### Environment

* Elixir version (elixir -v): 1.11.2
* Phoenix version (mix deps): 1.5.7
* Node.js version (node -v): v12.16.1
* NPM version (npm -v): v6.14.7
* Operating system: Ubuntu 18.04.5 LTS

### Expected behavior
When using a dynamic port calling `WebApp.Endpoint.url()` should still return a valid url. 

```
config :ddrive, WebApp.Endpoint,
  http: [ip: {127, 0, 0, 1}, port: 0]
```

`WebApp.Endpoint.url()` should return e.g. `"http://localhost:46815"`

### Actual behavior
`WebApp.Endpoint.url()` does always return "http://localhost:0"

I can use `ranch:get_port(WebApp.Endpoint.HTTP)` to get the correct dynamically allocated port e.g. `46815` and confirm that the site is actually running there. But  `WebApp.Endpoint.url()` always returns the wrong url.

dominicletz:
My current ugly workaround on my desktop webapp is this:

```elixir
defmodule WebApp.Endpoint do
  def url do
    url = Phoenix.Config.cache(__MODULE__, :__phoenix_url__, &Phoenix.Endpoint.Supervisor.url/1)
    String.replace(url, ":0", ":#{:ranch.get_port(WebApp.Endpoint.HTTP)}")
  end

  use Phoenix.Endpoint, otp_app: :ddrive
```

e.g. I'm overwriting the default endpoint.url() call with my ugly patched one to inject the right url

josevalim:
We don't really support dynamic ports at the moment.

It is also worth nothing there are two distinct configurations:

1. Which URL to listen on (http/https)
2. Which URL to show to your users

If you don't configure 2, we will use 1 by default. The fix is to add this config:

    url: [host: "localhost", port: 80]

If someone wants to submit a pull request that actually supports dynamic ports efficiently, we will gladly take a look at, but not in our plans. Thanks!

PR #5553, opened 2023-08-16
Support dynamic port for `c:Endpoint.url/0`

LostKobrakai:
This is a naive implementation, as it doesn't yet support the adapter crashing and potentially getting assigned a new port. I'm not sure if cowboy or bandit for the matter crash all the way up to the endpoint in such a case, so phoenix could react to a new port or if there would need to be some other mechanism to make phoenix aware of the changed port.

That should still be better than leaving the `:0`, as it works right now, I think.

Relates to #4186

### DIFF
--- a/lib/phoenix/endpoint/cowboy2_adapter.ex
+++ b/lib/phoenix/endpoint/cowboy2_adapter.ex
@@ -80,7 +80,7 @@ defmodule Phoenix.Endpoint.Cowboy2Adapter do
       Application.ensure_all_started(:ssl)
     end
 
-    ref = Module.concat(endpoint, scheme |> Atom.to_string() |> String.upcase())
+    ref = build_ref(endpoint, scheme)
 
     plug =
       if code_reloader? do
@@ -131,6 +131,18 @@ defmodule Phoenix.Endpoint.Cowboy2Adapter do
     end
   rescue
     _ -> scheme
+  end
+
+  @doc false
+  def dynamic_port(endpoint, scheme) do
+    case :ranch.get_addr(build_ref(endpoint, scheme)) do
+      {:local, _unix_path} -> raise "Trying to determine dynamic port, but the adapter is configured to use unix sockets. Remove the configuration for `[port: 0]` on the #{scheme} scheme."
+      {_addr, port} -> port
+    end
+  end
+
+  defp build_ref(endpoint, scheme) do
+    Module.concat(endpoint, scheme |> Atom.to_string() |> String.upcase())
   end
 
   # TODO: Remove this once {:system, env_var} deprecation is removed
--- a/lib/phoenix/endpoint/supervisor.ex
+++ b/lib/phoenix/endpoint/supervisor.ex
@@ -351,8 +351,8 @@ defmodule Phoenix.Endpoint.Supervisor do
 
     {scheme, port} =
       cond do
-        https -> {"https", https[:port] || 443}
-        http -> {"http", http[:port] || 80}
+        https -> {"https", transport_port(endpoint, https, :https)}
+        http -> {"http", transport_port(endpoint, http, :http)}
         true -> {"http", 80}
       end
 
@@ -368,6 +368,38 @@ defmodule Phoenix.Endpoint.Supervisor do
 
     %URI{scheme: scheme, port: port, host: host}
   end
+
+  defp transport_port(endpoint, config, scheme) do
+    cond do
+      value = config[:port] ->
+        value
+        |> port_to_integer()
+        |> handle_dynamic_port(endpoint, scheme)
+
+      scheme == :https ->
+        443
+
+      scheme == :http ->
+        80
+    end
+  end
+
+  def handle_dynamic_port(0, endpoint, scheme) do
+    config =
+      cond do
+        server = endpoint.config(:server) -> [server: server]
+        true -> []
+      end
+
+    if server?(config) do
+      adapter = endpoint.config(:adapter) || Phoenix.Endpoint.Cowboy2Adapter
+      adapter.dynamic_port(endpoint, scheme)
+    else
+      0
+    end
+  end
+
+  def handle_dynamic_port(port, _, _), do: port
 
   defp warmup_static(endpoint, %{"latest" => latest, "digests" => digests}) do
     Phoenix.Config.put_new(endpoint, :cache_static_manifest_latest, latest)
--- a/lib/phoenix/endpoint/supervisor.ex
+++ b/lib/phoenix/endpoint/supervisor.ex
@@ -391,8 +391,9 @@ defmodule Phoenix.Endpoint.Supervisor do
         true -> []
       end
 
-    if server?(config) do
-      adapter = endpoint.config(:adapter) || Phoenix.Endpoint.Cowboy2Adapter
+    adapter = endpoint.config(:adapter) || Phoenix.Endpoint.Cowboy2Adapter
+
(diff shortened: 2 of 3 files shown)


lib/phoenix/endpoint/supervisor.ex
```diff
@@ -351,8 +351,8 @@ defmodule Phoenix.Endpoint.Supervisor do
 
     {scheme, port} =
       cond do
-        https -> {"https", https[:port] || 443}
-        http -> {"http", http[:port] || 80}
+        https -> {"https", transport_port(endpoint, https, :https)}
+        http -> {"http", transport_port(endpoint, http, :http)}
```
mtrudel, line 355: Do we *need* to be so explicit about sorting through the config ourselves? Rather than special casing the times where the port is dynamic, we could just *always* ask the adapter for the port (I'm happy to add `dynamic_port/2` to `Bandit.PhoenixAdapter`) and be done with it. Taken with the changes suggested for domain sockets above, it would strip away most of the complexity of the changes to this file, I think.

mtrudel:
>  I'm not sure if cowboy or bandit for the matter crash all the way up to the endpoint in such a case, so phoenix could react to a new port or if there would need to be some other mechanism to make phoenix aware of the changed port.

Crashes of specific handlers won't unwind all of Bandit or Cowboy, so we're fine there. What may happen though, is changes to the port number in the case of suspension (see [Thousand Island docs](https://hexdocs.pm/thousand_island/1.0.0-pre.7/ThousandIsland.html#suspend/1) and [Ranch docs, 'Suspending and resuming a listener'](https://ninenines.eu/docs/en/ranch/2.1/guide/listeners/)).

lib/phoenix/endpoint/supervisor.ex
LostKobrakai: That sounds good, but it also kinda prepends on the adapter being able to return a port / being started. That should be a simpler check though.

lib/phoenix/endpoint/cowboy2_adapter.ex
```diff
@@ -133,6 +133,18 @@ defmodule Phoenix.Endpoint.Cowboy2Adapter do
     _ -> scheme
   end
 
+  @doc false
+  def dynamic_port(endpoint, scheme) do
+    case :ranch.get_addr(build_ref(endpoint, scheme)) do
+      {:local, _unix_path} -> raise "Trying to determine dynamic port, but the adapter is configured to use unix sockets. Remove the configuration for `[port: 0]` on the #{scheme} scheme."
```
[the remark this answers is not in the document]
LostKobrakai: Hmm, I saw it the other way round. If you're using unix socket nothing should even ask for a port, as that doesn't really make sense. Given the above discussion returning `nil` might be more appropriate though.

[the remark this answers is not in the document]
mtrudel: Good point. If we step back and observe that the goal here is to build a URI for the server, we should be asking 'what does that URI even look like for a domain socket' (and secondarily, 'is that the URI that the user cares about'). 

We have answers to this already in [Bandit](https://github.com/mtrudel/bandit/blob/main/lib/bandit.ex#L324-L333) and [Cowboy](https://github.com/phoenixframework/phoenix/blob/3f0dfced31a66edb0932c479e2da51125d1c8e30/lib/phoenix/endpoint/cowboy2_adapter.ex#L124-L135). If we wanted to follow down that path we should be extracting both the address and the port from the underlying server (that makes sense to me, FWIW).

lib/phoenix/endpoint/supervisor.ex
mtrudel: That's a good point - at many of the calls we're guaranteed that the server *won't* be running. What ought to be the behaviour in those cases, do you think?

josevalim:
Yeah, I don't think we should do this, especially if url will change between compile vs runtime. We use this in Livebook and we basically compute and store our own URL based on port: 0. :)

LostKobrakai:
I've been running into this with elixir desktop, which currently integrates with cowboy specifically partly due to this. I'm fine if this might not be a responsibility for phoenix, but I'd also love to see a shared solution to the issue rather than a "figure it out on your own". Especially with the printed banner on startup it really feels strange that this would be the only place in phoenix knowing the actual port, with no other way to get to know it across all adapters.

mtrudel:
> Yeah, I don't think we should do this, especially if url will change between compile vs runtime. We use this in Livebook and we basically compute and store our own URL based on port: 0. :)

Building URLs for a server is hard.

Setting aside issues of constructing URLs that account for load balancers, domain sockets, etc, even the simple case above is more nuanced than what José mentions, since the recent 'suspend/resume' patterns that @chrismccord et al have been working on will end up changing the port between resumptions. Ultimately it's really difficult to build a URL that has guaranteed meaning beyond the scope of a specific connection and as encoded already within a `Plug.Conn`. And even *that* isn't perfect, since the port & host that *we* see is often not the one that the user sees (as I've [talked about before](https://github.com/mtrudel/bandit/issues/106#issuecomment-1458847902)). 

This isn't an isolated problem, either. There's currently a [discussion](https://github.com/ueberauth/ueberauth/pull/187) going on over on uberauth that closely relates to this, and similar issues have come up a few times in Bandit's history.  I *do* think that there's an opportunity to encapsulate all of this within the adapter layer, as e.g. LiveView having to know about ranch is a less than ideal encapsulation of the whole thing. I think we would want to:

1. Formalize *exactly* what the port, host, etc, fields in a `Plug.Conn` mean for connection information in the context of a specific connection (see previous discussion on this [here](https://github.com/mtrudel/bandit/issues/106#issuecomment-1458847902))
2. Provide a way for servers to be able to provide information about their bound host and port in a general sense (ie: not with relation to a specific connection). My thought here would be as part of Plug, so that we could implement it via the Plug.Conn.Adapter behaviour.
3. More ambitiously, encapsulate server startup & lifecycle this way as well. This would alleviate the need for server-specific phoenix adapters, and also provide a single place for URL construction both for programatically built URLs as well as user configured ones (José, we've spoked about this in the past regarding URL construction as part of the `Plug.Conn.init/5` API). 

Specifically with respect to *this* issue, it sounds like the appetite is to leave it alone for now. If we can find consensus on a broader approach that would solve this and issues such as the one on uberauth at the moment, I'm happy to drive implementation.

WDYT?

josevalim:
Make it a responsibility of the connection makes a lot of sense. For desktop apps, there aren’t proxies, so always accurate. We do have rewrite plugs for proxies too.

LostKobrakai:
> Provide a way for servers to be able to provide information about their bound host and port in a general sense

This is what I feel is the missing piece – being able to initially connect to the dynamic port without depending on implementation details about the underlying servers. Any later route building based on an actual connection could be handled with plug/on_mount tooling to make sure `url(@conn_or_socket, ~p"…")` works (or even just route with relative paths only).

Point 3. certainly sounds ambitious. I'm wondering if there's an in between solution, where the server implementation could inform phoenix when their bound interface has changed.

josevalim:
We could augment the Plug.Adapter protocol to introduce a info function with metadata about the server but I'd say this is out of scope of Phoenix (or at least until it is tackled in Plug). :) Thanks @LostKobrakai!

mtrudel:
I'm planning on tackling a PR against Plug et al. for this in the next couple of weeks!

josevalim:
This has been added to main.

PR #5556, opened 2023-08-16
fix flash function selector

jamilabreu:
currently throws an "invalid selector" error:

https://share.cleanshot.com/qySqJgHs

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -110,7 +110,7 @@ defmodule <%= @web_namespace %>.CoreComponents do
     <div
       :if={msg = render_slot(@inner_block) || Phoenix.Flash.get(@flash, @kind)}
       id={@id || "flash-#{@kind}"}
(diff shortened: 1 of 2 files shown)


Gazler:
I could be wrong, but I believe the existing code is correct and the error here is because you are missing an id? The `hide/2` function takes a selector, which is being called with "#" in this instance.

Gazler:
That being said, looking at the code, if an ID is nil, one is automatically assigned for the element based on the kind:

```
      id={@id || "flash-#{@kind}"}
      phx-click={JS.push("lv:clear-flash", value: %{key: @kind}) |> hide("#{@id}")}
 ```
 
 So perhaps the fix is actually to do:
 
 ```
       id={@id || "flash-#{@kind}"}
      phx-click={JS.push("lv:clear-flash", value: %{key: @kind}) |> hide("#" <> @id || "flash-#{@kind}")}
```

josevalim:
Maybe we should remove the `default` from :id and use `assign_new` to set it if none is set? Can you please verify and amend @jamilabreu?

josevalim:
Close in favor of #5542.

Issue #5559, opened 2023-08-21
The Phoenix.Router documentation on scopes is incorrect


stepchud:
This was originally brought up in a [Slack thread](https://elixir-lang.slack.com/archives/C03QQCV4H/p1692044593629569).
There is an issue in the documentation which could cause insecure implementations. The [How to organize my routes](https://hexdocs.pm/phoenix/Phoenix.Router.html#module-how-to-organize-my-routes) section creates a scope for authenticated routes. The order of the routes in the example means the `get /posts/new` route will not be authenticated because it matches a route in the previous scope which is not authenticated.

I think the easiest fix is to swap the order of the scopes in the example. I'll submit a PR for this.

### Environment

* Elixir version (elixir -v): 1.14.5
* Phoenix version (mix deps): 1.7.7
* Operating system: macOS Ventura 13.4.1

PR #5560, opened 2023-08-21
Update documentation around authenticated routes

stepchud:
This fixes https://github.com/phoenixframework/phoenix/issues/5559

I reversed the order of the scopes in the documentation example to match the authenticated route first and added a small blurb to recommend this approach in other apps. Feel free to remove the blurb if you don't think it's helpful.


josevalim:

PR #5577, opened 2023-09-16
Add documentation on using presence with LiveView

### LINKED ISSUE #4858: Split Presence guide into two guides – Channels + Presence, LV + Presence

Gazler:
The presence guide has been split into two sections, the existing one is now usage with Channels and JavaScript.

The new section if the guide is focused on LiveView and outlines creating an Elixir based presence client, and using a LiveView stream for rendering the list of online users.

This closes #4858

### DIFF
--- a/guides/real_time/presence.md
+++ b/guides/real_time/presence.md
@@ -56,6 +56,8 @@ children = [
 ]
 ```
 
+## Usage With Channels and JavaScript
+
 Next, we will create the channel that we'll communicate presence over. After a user joins, we can push the list of presences down the channel and then track the connection. We can also provide a map of additional information to track.
 
 ```elixir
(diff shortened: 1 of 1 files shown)


guides/real_time/presence.md
```diff
@@ -122,6 +124,120 @@ Bob (count: 1)
...
+
+If we add this route to the `lib/hello_web/router.ex`:
+
+```elixir
+    live "/online/:name", OnlineLive, :index
+```
+
+Then we can navigate to http://localhost:4000/online/Alice in one tab, and http://localhost:4000/online/Bob in another, you'll see that the presences are tracked, along with the number of presences per user. Opening and closing tabs with various users will update the presence list in real-time.
+
 ## Making it safe
```
josevalim, line 241: Should we make this a subsection of the channel one?

josevalim (approved):
Beautiful, just one tiny comment and ship it!

guides/real_time/presence.md
```diff
@@ -122,6 +124,120 @@ Bob (count: 1)
 
 If we close one of the Alice tabs, then the count should decrease to 1. If we close another tab, the user should disappear from the list entirely.
 
+## Usage With LiveView
+
+Whilst Phoenix does ship with a JavaScript API for dealing with presence, it is also possible to extend the `HelloWeb.Presence` module to support LiveView.
+
+One think to keep in mind when dealing with LiveView, is that each LiveView is a stateful process, so if we keep the oresebce state in the LiveView, each LiveView process will contain the full list of online users in memory. Instead, we can keep track of the online users within the `Presence` process, and pass separate events to the LiveView, which can use a stream to update the online list.
```
jordelver, line 131:
```suggestion
One think to keep in mind when dealing with LiveView, is that each LiveView is a stateful process, so if we keep the presence state in the LiveView, each LiveView process will contain the full list of online users in memory. Instead, we can keep track of the online users within the `Presence` process, and pass separate events to the LiveView, which can use a stream to update the online list.
```

Gazler: Also, "one thing" instead of "one think"!

guides/real_time/presence.md
Gazler: This is the channels one? Do you mean should we make the content match and just change the URLs?

https://github.com/phoenixframework/phoenix/blob/d33f7bd034d1ac34145ea1a680bb79e9e2ae0d4f/guides/real_time/presence.md?plain=1#L118

PR #5579, opened 2023-09-18
Fix filename when sending downloads with non-ascii names

### LINKED ISSUE #5478: `Phoenix.Controller.send_download/3` incorrectly escapes spaces in filenames
### Environment

* Elixir version (elixir -v): Erlang/OTP 25, Elixir 1.14.4
* Phoenix version (mix deps): 1.7.2

### Expected behavior

When hitting a controller route that looks like this:

```elixir
def download(conn, _params) do
  send_download(conn, {:binary, "test"}, filename: "name with spaces.txt")
end
```

I would expect the user to be prompted to download a file named `name with spaces.txt`.

### Actual behavior

The user is prompted to download a file named `name+with+spaces.txt`

### Proposed solution

Using RFC-3986 encoding instead of "x-www-form-urlencoded"; i.e. encoding the space character as `%20` instead of `+`. The following does preserve spaces:

```elixir
def download(conn, _params) do
  send_download(conn, {:binary, "test"},
    filename: URI.encode("name with spaces.txt", %URI.char_unreserved?/1),
    encode: false
  )
end
```

I don't believe using `URI.char_unreserved?/1` instead of the default `URI.char_unescaped?/1` is actually necessary for security reasons but it is the smallest breaking change from `URI.encode_www_form/1`.

### Previous discussions

Discussed in #3344 where the `encode: false` option was introduced. The option to use `URI.encode/2` was I believe dismissed in error.

I suspect when testing whether `URI.encode/2` would work the filename accidentally got encoded twice (i.e. `filename="some%2520file.txt"`), which is what happens if you manually `URI.encode` the filename without also disabling the automatic encoding. The browser will decode that - once - into `some%20file.txt`.

Gazler:
according to https://www.rfc-editor.org/rfc/rfc6266#appendix-D when it is not possible to express a filename with regular encoding, a `filename*` parameter can be added to the header.

This commit adds the `filename*` when the encoded filename does not match the filename.

This has been tested on chrome, safari and firefox.

This closes #5478

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -1259,16 +1259,22 @@ defmodule Phoenix.Controller do
     disposition_type = get_disposition_type(Keyword.get(opts, :disposition, :attachment))
     warn_if_ajax(conn)
 
+    disposition = ~s[#{disposition_type}; filename="#{encoded_filename}"]
+
+    disposition =
+      if encoded_filename != filename do
+        disposition <> "; filename*=utf-8''#{encoded_filename}"
+      else
+        disposition
+      end
+
     conn
     |> put_resp_content_type(content_type, opts[:charset])
-    |> put_resp_header(
-      "content-disposition",
-      ~s[#{disposition_type}; filename="#{encoded_filename}"]
-    )
+    |> put_resp_header("content-disposition", disposition)
   end
 
   defp encode_filename(filename, false), do: filename
-  defp encode_filename(filename, true), do: URI.encode_www_form(filename)
+  defp encode_filename(filename, true), do: URI.encode(filename)
 
   defp get_disposition_type(:attachment), do: "attachment"
   defp get_disposition_type(:inline), do: "inline"
--- a/test/phoenix/controller/controller_test.exs
+++ b/test/phoenix/controller/controller_test.exs
@@ -530,9 +530,20 @@ defmodule Phoenix.Controller.ControllerTest do
       conn = send_download(conn(:get, "/"), {:file, @hello_txt}, filename: "hello world.json")
       assert conn.status == 200
       assert get_resp_header(conn, "content-disposition") ==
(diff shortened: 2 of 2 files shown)


mcrumm (approved):
🍬 🍬 🍬 🍬 🍬

josevalim:
I think we should be using URI.encode for the filename as well, no? Wouldn’t that be enough?

Gazler:
@mcrumm said URI.encode on the filename failed I'm Safari

josevalim:
Do you know how it failed?

mcrumm:
I think Safari saves the file as e.g. `hello%20world.json`

josevalim:
I see. So this patch is needed anyway but I suggest using URI.encode in both, instead of URI.encode_www_form in one and URI.encode in the other, wdyt?

Gazler:
According to the spec:

> Include a "filename" parameter when [US-ASCII](https://www.rfc-editor.org/rfc/rfc6266#ref-US-ASCII) ([US-ASCII]) is
   sufficiently expressive.
      
But I don't know how strictly this is enforced by browsers. Also:

> Avoid including the percent character followed by two hexadecimal
      characters (e.g., %A9) in the filename parameter, since some
      existing implementations consider it to be an escape character,
      while others will pass it through unchanged.

This is what we currently do though with either `encode_www_form` or `encode`.

josevalim:
Yeah, using both encode or encode_www_form are wrong for the current behavior and this PR addressed that. my point is that we should just use one of them once this PR is added, especially to avoid encoding twice, so I would go with URI.encode consistently :)

Gazler:
@josevalim did you mean like this https://github.com/phoenixframework/phoenix/compare/f56fc44577d4f4f6c01b1dda5fb929b3857394df..e08fc4bcb6d48dd89012bc987ed99a4def34ffcc

josevalim:
Yes, perfect, ship it!

Issue #5463, opened 2023-05-16
Scope Log-level not working


Dowwie:
### Environment

* Elixir 1.14.4
* Phoenix 1.7.2
* Operating system: macos

### Expected behavior

`log: :debug` defined at the scope-level should disable route logging when the log-level of the application is :info. 

### Actual behavior

scope-level logging continues to emit, regardless of the `log` option

joaofcosta:
Hey @Dowwie 👋 

I am taking a look at this issue and wanted to first double-check if I understood your report correctly. I created a new phoenix app with the `mix phx.new` command and then updated the router module to use `log: :debug` on one of its scopes:

```elixir
  scope "/", LoggingWeb, log: :debug do
    pipe_through :browser

    get "/", PageController, :home 
  end
```

Without changing the default log level in development, this is what I see when making a request to `/`:

<img width="520" alt="SCR-20230909-nytc" src="https://github.com/phoenixframework/phoenix/assets/6334322/fccad89b-1740-44c5-b8ee-665a9e8286bc">

After updating the log level to `:info` on `config/dev.exs` I can see that the debug message that was previously shown is now suppressed when visiting `/`:

<img width="519" alt="SCR-20230909-nyzj" src="https://github.com/phoenixframework/phoenix/assets/6334322/754576b9-4b99-4eb5-9057-fa1f3bf262a1">

What would be, for you, the expected behaviour here? Were you expecting that both the `[info] GET/` and `[info] Sent 200 in ...` messages would be suppressed seeing as the endpoint is defined in the scope where `log: :debug` is being used? 

Thanks!

Dowwie:
Messages in Slack older than 90 days cannot be referenced in free tier solutions, so unfortunately I can't point you to the discussion I had with Chris about this subject.

> What would be, for you, the expected behaviour here? Were you expecting that both the [info] GET/ and [info] Sent 200 in ... messages would be suppressed seeing as the endpoint is defined in the scope where log: :debug is being used?

Yes, precisely.  The situation I was working with was related to health checkpoints where ALB and K8s were flooding logging with probe requests.  I wanted to quiet logging on those endpoints by setting the scope-level log-level of these health-check endpoints to `:debug`, but that wasn't achieving the desired effect even when setting application-level log-level, within `/config`, to `:info`.

dinocosta:
[...]

PR #5580, opened 2023-09-19
Update documentation for :log option on both scope/2 and match/5

dinocosta:
While looking into [this issue](https://github.com/phoenixframework/phoenix/issues/5463), it was noted that maybe the documentation for `Phoenix.Router.scope/2` did not make it clear that the `:log` option only controls the log level for the router dispatching message logged by Phoenix.

[It was suggested](https://github.com/phoenixframework/phoenix/issues/5463#issuecomment-1720947223) that updating the documentation for this option can help clarify this, while also pointing users to documentation on how they can also control plug level logging.

As such, this Pull Request does exactly that, by updating the documentation for both `Phoenix.Router.scope/2` and `Phoenix.Router.match/5` in regards to the `:log` option.


Gazler:
Perfect. Thanks! <3

PR #5581, opened 2023-09-20
Change non-returning map into forEach

ilkka:
This is an embarrassingly tiny niggle, but something my linter picked up in a recent project: since the callback passed to `map` does not return anything, `map` returns just an array of `undefined`... and in any case the result of the `map` is thrown away. Therefore it would be semantically nicer to do a `forEach` loop instead (or just a for loop, but the outer loop is already a `forEach` so this fits better stylistically).


Gazler:
Thanks <3

PR #5584, opened 2023-09-23
Fix small bug in test - reused conn from previous test case

loganmac:
I noticed a flaky test from my generated liveview auth. The confirmation liveview has a test to make sure that there's not a flash with an error when confirming as a logged in user. The problem is the connection already had an error flash from a test case that was a little higher in the same test. I was only getting the bug after adding a redirect from my home url ("/") in my pages controller, but then it would have two flashes: 
```
  1) test Confirm user confirms the given token once (MyAppWeb.UserConfirmationLiveTest)
     test/my_app_web/live/user_confirmation_live_test.exs:20
     Expected false or nil, got "User confirmation link is invalid or it has expired."
     code: refute Phoenix.Flash.get(conn.assigns.flash, :error)
     arguments:

         # 1
         %{"error" => "User confirmation link is invalid or it has expired.", "info" => "User confirmed successfully."}

         # 2
         :error
```

This fixes it so that the conn isn't reused across these two scenarios. Perhaps a better long-term solution would be to split these test cases up though.

### DIFF
--- a/priv/templates/phx.gen.auth/confirmation_live_test.exs
+++ b/priv/templates/phx.gen.auth/confirmation_live_test.exs
@@ -55,8 +55,9 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
                "<%= inspect schema.alias %> confirmation link is invalid or it has expired"
 
       # when logged in
+      conn = build_conn()
       {:ok, lv, _html} =
-        build_conn()
+        conn
         |> log_in_<%= schema.singular %>(<%= schema.singular %>)
         |> live(~p"<%= schema.route_prefix %>/confirm/#{token}")
 


Gazler:
Thanks for the PR. Can you just confirm the change that you made for the redirect after running `mix phx.gen.auth` so I can trigger the bug?

loganmac:
Sure, so if you generate a new phoenix project: 
`mix phx.new phxtest`

Then do the normal startup, generate auth:
`mix phx.gen.auth Accounts User users` (set it up the liveview way)

Generate a liveview:
`mix phx.gen.live Some Example examples name:string`

Add the routes to your `require_authenticated_user` live session, then in your PageController, set  home/1 to 
`redirect(conn, to: ~p"/examples")`

Run your test for your generated auth, and you'll see some failures. The failures for 200 response instead of redirects are expected, and you'll have to update the tests for those. The confirmation live test shouldn't be failing though, and it's only because that test case uses the same `conn` from the previous "# when not logged in" assertions:

```
mix test
......................

  1) test POST /users/log_in logs the user in (PhxtestWeb.UserSessionControllerTest)
     test/phxtest_web/controllers/user_session_controller_test.exs:11
     ** (RuntimeError) expected response with status 200, got: 302, with body:
     "<html><body>You are being <a href=\"/examples\">redirected</a>.</body></html>"
     code: response = html_response(conn, 200)
     stacktrace:
       (phoenix 1.7.7) lib/phoenix/test/conn_test.ex:373: Phoenix.ConnTest.response/2
       (phoenix 1.7.7) lib/phoenix/test/conn_test.ex:387: Phoenix.ConnTest.html_response/2
       test/phxtest_web/controllers/user_session_controller_test.exs:22: (test)

...........................................................................

  2) test Confirm user confirms the given token once (PhxtestWeb.UserConfirmationLiveTest)
     test/phxtest_web/live/user_confirmation_live_test.exs:20
     Expected false or nil, got "User confirmation link is invalid or it has expired."
     code: refute Phoenix.Flash.get(conn.assigns.flash, :error)
     arguments:

         # 1
         %{"error" => "User confirmation link is invalid or it has expired.", "info" => "User confirmed successfully."}

         # 2
         :error

     stacktrace:
       test/phxtest_web/live/user_confirmation_live_test.exs:70: (test)

.........................

  3) test register user creates account and logs the user in (PhxtestWeb.UserRegistrationLiveTest)
     test/phxtest_web/live/user_registration_live_test.exs:40
     ** (RuntimeError) expected response with status 200, got: 302, with body:
     "<html><body>You are being <a href=\"/examples\">redirected</a>.</body></html>"
     code: response = html_response(conn, 200)
     stacktrace:
       (phoenix 1.7.7) lib/phoenix/test/conn_test.ex:373: Phoenix.ConnTest.response/2
       (phoenix 1.7.7) lib/phoenix/test/conn_test.ex:387: Phoenix.ConnTest.html_response/2
       test/phxtest_web/live/user_registration_live_test.exs:52: (test)

..........

  4) test GET / (PhxtestWeb.PageControllerTest)
     test/phxtest_web/controllers/page_controller_test.exs:4
     ** (RuntimeError) expected response with status 200, got: 302, with body:
     "<html><body>You are being <a href=\"/examples\">redirected</a>.</body></html>"
     code: assert html_response(conn, 200) =~ "Peace of mind from prototype to production"
     stacktrace:
       (phoenix 1.7.7) lib/phoenix/test/conn_test.ex:373: Phoenix.ConnTest.response/2
       (phoenix 1.7.7) lib/phoenix/test/conn_test.ex:387: Phoenix.ConnTest.html_response/2
       test/phxtest_web/controllers/page_controller_test.exs:6: (test)


Finished in 1.2 seconds (0.5s async, 0.6s sync)
136 tests, 4 failures
```

Malian:
I faced the same issue a few days ago, and it took me a while to work out why the test wasn't working. I applied the same "fix".

Gazler:
Thanks for detailed instructions and thanks for the PR! <3

PR #5586, opened 2023-09-27
Allow UTC datetimes in the generators

### LINKED ISSUE #5564: Phoenix generators use naive Ecto timestamps - UTC would be much better
### Actual behavior

Currently when using the Phoenix generators the timestamps are NaiveDateTime. That means you can't compare them with other `DateTime`s and see which happened first. You can't use `DateTime.to_unix/1` to convert them to unix timestamps. When serializing them, for instance in a JSON API, the consumers of the API won't know if it's in UTC because there won't automatically be a UTC offset indicator (the Z at the end of the ISO8601 timestamp).

Another thing is that a lot of Elixir projects start with the Phoenix generators and are perhaps seen as a good example or a good starting place. I think this could be improved if UTC DateTime were used for timestamps instead of NaiveDateTime.

NaiveDateTime is meant for when you don't know the timezone. We know the Ecto `timestamps` are UTC, so it makes sense that they are UTC DateTime. It's a very good example of when to use DateTime instead of NaiveDateTime. The Ecto default to NaiveDateTime is there for historical reasons when there was less support for DateTime in the Elixir standard library. As far as I understand it, the reason UTC datetimes have not be made the default in Ecto is because it would require releasing a new major version (Ecto 4) to avoid breaking existing behaviour. The Phoenix generators don't have that same restriction since the generated schemas and tables are new.

### Expected behavior

When using the phx generators timestamps should be [UTC DateTime](http://www.creativedeletion.com/2019/06/17/utc-timestamps-in-ecto.html), not NaiveDateTime. One way to do that is to have the schema line specifying the timestamp say `timestamps([type: :utc_datetime])` instead of `timestamps()`.

Gazler:
The generators now consider the :timestamp_type value in the generator config for an application.

For backwards compatability, if this value is not set, it will default to `:naive_datetime` which will generate the timestamps without an explicit type, otherwise the type will be set explicitly, such as:

`timestamps(:type, :utc_datetime)`

This closes #5564

### DIFF
--- a/installer/lib/phx_new/project.ex
+++ b/installer/lib/phx_new/project.ex
@@ -16,7 +16,7 @@ defmodule Phx.New.Project do
             opts: :unset,
             in_umbrella?: false,
             binding: [],
-            generators: []
+            generators: [timestamp_type: :utc_datetime]
 
   def new(project_path, opts) do
     project_path = Path.expand(project_path)
--- a/installer/test/phx_new_test.exs
+++ b/installer/test/phx_new_test.exs
@@ -54,6 +54,7 @@ defmodule Mix.Tasks.Phx.NewTest do
 
       assert_file("phx_blog/config/config.exs", fn file ->
         assert file =~ "ecto_repos: [PhxBlog.Repo]"
+        assert file =~ "generators: [timestamp_type: :utc_datetime]"
         assert file =~ "config :phoenix, :json_library, Jason"
         assert file =~ ~s[cd: Path.expand("../assets", __DIR__),]
         refute file =~ "namespace: PhxBlog"
@@ -557,7 +558,7 @@ defmodule Mix.Tasks.Phx.NewTest do
   test "new with binary_id" do
     in_tmp("new with binary_id", fn ->
       Mix.Tasks.Phx.New.run([@app_name, "--binary-id"])
-      assert_file("phx_blog/config/config.exs", ~r/generators: \[binary_id: true\]/)
+      assert_file("phx_blog/config/config.exs", ~r/generators: \[.*binary_id: true\.*]/)
     end)
   end
 
--- a/lib/mix/phoenix/schema.ex
+++ b/lib/mix/phoenix/schema.ex
@@ -38,7 +38,8 @@ defmodule Mix.Phoenix.Schema do
             migration_module: nil,
             fixture_unique_functions: [],
             fixture_params: [],
-            prefix: nil
+            prefix: nil,
+            timestamp_type: :naive_datetime
 
   @valid_types [
     :integer,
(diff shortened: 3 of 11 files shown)


Gazler:
I still need to document this and add it to the installer, just want to make sure we're happy with this approach.

chrismccord:
looks good! Please proceed

josevalim (approved):

dennym:
Sorry for opening this again. This sounds nice but is kinda undocumented and leaves me rather puzzled how this works.
How is this different to the `migration_timestamps` of https://hexdocs.pm/ecto_sql/Ecto.Migration.html#module-migrations-configuration ?

Currently I try to implement this for a non-production application which was created with `phx-v1.7`. Am I supposed to change 
`generators: [binary_id: true, timestamp_type: :utc_datetime]` in my `config.exs`. If yes then I am unsure how this will affect my newly created migrations since I just create new ones and then add the `create_table/1` with `timestamps/0` manually. How would additionally help with the serialization and deserialization of the timestamps?

best

josevalim:
The migration is the same, because the database is the same. This is about how Ecto will view your database data.

dennym:
So just to follow up. If I set `timestamp_type: :utc_datetime` in the generators and manually use `timestamps(:utc_datetime)` in my migrations I would also require to set the previously linked ecto setting. Would this be the correct approach?

josevalim:
You don't need to change migrations, the default timestamp is the same for all types:

https://github.com/elixir-ecto/ecto_sql/blob/master/lib/ecto/adapters/postgres/connection.ex#L1863-L1866

dennym:
I'm propably totally dumb now, but I now know even less. I have changed the config.exs generator setting to contain `timestamp_type: :utc_datetime` which does nothing to existing or new migrations (its still naive_datetime).

So my last shot of compiling all the information together is that the generator settings is just relevant for the scaffolding scripts of phoenix and nothing to do how it compiles or handles the data internally. What I have to do is to change all my migratons `timestamp/0` to `timestamps(:utc_datetime)` and schemas to `field :foo_at, :datetime`.

Just realised that I might mistake the `timestamps/0` from the schemas with the `timestamps/0` from the migrations.

So really last shot is now: 
- set the generator for vanity in case I use scaffolds (used by phx)
- set the ecto `migration_timestamps: [type: :utc_datetime]` option in my config.exs (used by ecto package)
- leave my schemas with `timestamps()` since this is taken from the preceeding change
- not change anything in my migrations and just drop the db and migrate again

Or as alternative:
- set the ecto `migration_timestamps: [type: :utc_datetime]` option

I think I caught up and just sort of need confirmation that the really last shot is the right procedure.

PR #5589, opened 2023-09-29
Automatically migrate when starting the server using sqlite3

Gazler:
When using the sqlite3 postgres adapter, the generator for the server overlay will automatically run the migrations first.

The tests have been updated to allow the adapter to be stubbed out to allow for testing the behaviour with various ecto adapters.

### DIFF
--- a/lib/mix/tasks/phx.gen.release.ex
+++ b/lib/mix/tasks/phx.gen.release.ex
@@ -38,7 +38,7 @@ defmodule Mix.Tasks.Phx.Gen.Release do
   require Logger
 
   @doc false
-  def run(args) do
+  def run(args, test_opts \\ []) do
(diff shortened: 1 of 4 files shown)


lib/mix/tasks/phx.gen.release.ex
```diff
@@ -153,6 +160,27 @@ defmodule Mix.Tasks.Phx.Gen.Release do
     |> Map.new()
   end
 
+  defp auto_migrate?(app, test_opts) do
+    repos = Application.get_env(app, :ecto_repos) |> List.wrap()
+    Enum.any?(repos, fn repo ->
```
Gazler, line 165: Perhaps this should be `Enum.all?` so if multiple adapters are used, we don't automigrate?

PR #5592, opened 2023-10-01
Use logged-in `conn` when following redirect

bronsonholden:
Hi 👋

Ran into a small bug with the auto-generated (`mix phx.gen.auth`) test file for user confirmations. If you change the confirmations LiveView to redirect to a route that requires authentication, it fails since the redirect is followed with a `conn` that isn't logged in via (e.g. for a user schema) `log_in_user`


Gazler:
Thanks <3

PR #5595, opened 2023-10-06
Docs: Remove separate index for leftmost column in multi-index

### LINKED ISSUE #5594: Possible inconsistent usage of `unique_index` in context docs
I am going through the guides, and noticed a possible inconsistency in the [Context docs](https://hexdocs.pm/phoenix/contexts.html).

When creating the `product_categories` table, it's made a point about how because one of the columns feature as the leftmost column in a `unique_index`, it doesn't need a separate index as this would only increase the write cost without bringing any benefit for reads. [This is the relevant line in the docs.](https://github.com/phoenixframework/phoenix/blob/2fc7fe7ac3f59adb7fa004e62452cb67551c12c4/guides/contexts.md?plain=1#L379)

Later, in when modifying the generated migration file for `cart_items`, a `unique_index` is added without removing the `cart_id` index. This seems inconsistent with the previous recommendation. [That happens here.](https://github.com/phoenixframework/phoenix/blob/2fc7fe7ac3f59adb7fa004e62452cb67551c12c4/guides/contexts.md?plain=1#L626)

Should this be updated? I am more than happy to create a pull request if that is the case :)

jakobsen:
Closes #5594.

### DIFF
--- a/guides/contexts.md
+++ b/guides/contexts.md
@@ -621,12 +621,12 @@ We generated a new resource inside our `ShoppingCart` named `CartItem`. This sch
       timestamps()
     end
 
-    create index(:cart_items, [:cart_id])
+-   create index(:cart_items, [:cart_id])
     create index(:cart_items, [:product_id])
 +   create unique_index(:cart_items, [:cart_id, :product_id])
 ```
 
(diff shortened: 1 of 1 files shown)


Gazler:
Thanks <3

PR #5602, opened 2023-10-15
Fix flash without ID not being closeable

jorams:
If no `id` was specified (like in `flash_group`) it defaulted to `nil`. The call to `assign_new` (#5542) did nothing because a default of `nil` had been assigned already.


Gazler:
Great spot, thanks! <3

PR #5609, opened 2023-10-22
remove duplicate text in heroku guide

koga1020:
I found that there was some duplicate text and code block in the Heroku guide. 

ref: https://github.com/phoenixframework/phoenix/pull/5558

Thanks!


Gazler:
Whoops. Good spot, thanks!

PR #5610, opened 2023-10-23
doc update: Added Education to seeds so that later example will work

adambriand:
Adding "Education" to seeds.exs, in order for the example later in the document to work as intended:

`
Title: Elixir Flashcards
Description: Flash card set for the Elixir programming language
Price: 5.000000
Views: 0
Categories:
Education
Books
`


Gazler:
Thanks <3

PR #5616, opened 2023-10-27
bugfix: flash component not closing

ivcosla:
The flash messages from the flash core component are not closing when clicking on the "cross".

The id in the component has an extra `#` symbol, so the click was not matching the proper id, after removing it I could test locally that the flash message is now closing properly.


ivcosla:
My bad, I saw this was properly fixed in https://github.com/phoenixframework/phoenix/pull/5602

Closing

PR #5617, opened 2023-10-28
Add required option to attr/3 in Components guide

sheerlox:
While following the Phoenix guides, I tried to get the compile time warning the guide was talking about, but all I could get was a runtime error when loading the page.

After reading the [`attr/3`](https://hexdocs.pm/phoenix_live_view/Phoenix.Component.html#attr/3) documentation, it seems the warning is only emitted when setting the `required` option to `true`.

*Since I'm a complete beginner with Phoenix, don't hesitate to point out any potential mistake or misunderstanding I could have made in my edit!*

### DIFF
--- a/guides/components.md
+++ b/guides/components.md
@@ -42,7 +42,7 @@ defmodule HelloWeb.HelloHTML do
 
   embed_templates "hello_html/*"
 
-  attr :messenger, :string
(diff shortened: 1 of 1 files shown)


josevalim:

PR #5619, opened 2023-10-31
Phoenix.Token doc tweaks

rktjmp:
Adjusts comments for sign and encrypt to mention their counterparts, verify and decrypt. Clarify that the default "max age" value for verify and decrypt comes from the token creation call.


josevalim:

PR #5621, opened 2023-11-01
Upgrade to latest websock_adapter, handle upgrade failures & report 400

mtrudel:
Hot on the heels of https://github.com/phoenixframework/websock_adapter/pull/14, handle upgrade errors and return 400 errors to the client, enabling folks to more easily connect upgrade failures with the client that causes them


jeregrine:
@mtrudel does this match the behavior of cowboy? So its strictly backwards compatible?

mtrudel:
Not entirely; Cowboy [returns 426](https://github.com/ninenines/cowboy/blob/22adc4de517bf54661795f6e18b6b2e62aec4140/src/cowboy_websocket.erl#L147) for failed HTTP/1 upgrades, but this is strictly incorrect (426 indicates that the *current* protocol is unsuitable for upgrading, not that the upgrade itself is incorrect). See [RFC9110§15.5.22](https://www.rfc-editor.org/rfc/rfc9110.html#section-15.5.22). 

Bandit [returns a 400](https://github.com/mtrudel/bandit/blob/main/lib/bandit/pipeline.ex#L134) in this case.

josevalim:

PR #5626, opened 2023-11-08
Add :fun option to __deriving__ of Phoenix.Param

nallwhy:
Sometimes I want to convert a struct to a Phoenix param using a function instead of a key. For this purpose, I added the `:fun` option. arbitrary function is not allowed of an error likes below.

`error: invalid quoted expression: #Function<0.26537881 in file:test/phoenix/param_test.exs>`


josevalim:
I think you should call defimpl and implement the protocol directly in such cases. Thanks!

PR #5627, opened 2023-11-08
Default to latest Elixir version Docker image

agonzalezro:
If the combination of the current Elixir version and OTP is not found in the
Docker registry we should default to a known combination, in this case the
latest Elixir version for that OTP version.

In case that no combination of Elixir/OTP is found the message shown to the
user is also more informational.


josevalim (approved):

Gazler:
This looks good. I think we should consider logging a warning to inform the user that the versions don't match exactly if we use a fallback.

agonzalezro:
Good idea @Gazler. Done.

Gazler:
Thanks <3

PR #5629, opened 2023-11-09
mix deps.unlock --unused

PragTob:
Just came here to check if that was also in phoenix, and it is - we should be able to get rid off these.

The dependency upgrade in here is a bit weird but since the `mix.exs` specifies `{:esbuild, "~> 0.8", only: :dev}` I think this is right and was just some form of missed update.

Thanks for all your work! :green_heart:


josevalim:

PR #5634, opened 2023-11-13
Use heroicons from git, closes #5622

### LINKED ISSUE #5622: Vendored heroicons take up considerable space on new projects
### Environment

* Elixir version (elixir -v):
```
Erlang/OTP 26 [erts-14.0.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit:ns]

Elixir 1.15.7 (compiled with Erlang/OTP 25)
```

* Phoenix version (mix deps):
```
* phoenix 1.7.9 (Hex package) (mix)
  locked at 1.7.9 (phoenix) 83e32da0
```

* Operating system:
```
Linux / Debian Bookworm container
```

### Actual behavior

When I generate a new  Phoenix 1.7 app into a new Git repository, I noticed that most of the storage is for vendored Heroicons.

```console
$ yes n | mix phx.new hello
$ du -sh hello
3.7M    hello
$ du -sh hello/assets/vendor/heroicons
3.5M    hello/assets/vendor/heroicons

```

When building a Docker container using the generated release Dockerfile (`phx.gen.release --docker`), I noticed that 3.5 MB out of 4.0 MB is made of up those SVG icons, even if only a few are used.

While that's not the end of the world (extra MBs in Git, extra MBs in container image), I thought it is not ideal.

I guess there's some convenience to having them vendored as is?

### Expected behavior

I expect that only the icons that I use should be included in a production build, and for all the icons to be part of a dependency that I don't need to store in my Git repository.

josevalim:


josevalim:

PR #5639, opened 2023-11-16
Move drainer configuration documentation to "options" for `Phoenix.Endpoint.socket/3`

rohan-b99:
The docs for the drainer config on the socket macro currently imply that it goes under the `websocket` or `longpoll` keys, but its actually read from the top level - see here https://github.com/phoenixframework/phoenix/blob/main/lib/phoenix/socket.ex#L450-L466

So e.g. this works

```elixir
    drainer: [batch_size: 10, batch_interval: 100],
    websocket: [
      connect_info: [session: @session_options]
    ]
```

but this doesn't:

```elixir
    websocket: [
      drainer: [batch_size: 10, batch_interval: 100],
      connect_info: [session: @session_options]
    ]
```


josevalim:
Hrm, I am almost that inside Phoenix.Socket the top-level options have been merged with the transport specific ones. So I would expect both to work.

josevalim:
Actually, looking at the code, we never merge them. We always read it from either :websockets or :longpoll, so the current version seems to correct to me (I verified it after some testing). :)

josevalim:

PR #5641, opened 2023-11-20
docs: update verified_routes.ex <.link/> examples

ruslandoga:
This PR updates `<.link>` components to use `href=` instead of `to=` as the latter doesn't seem to be supported anymore.


josevalim:

PR #5644, opened 2023-11-20
phx.gen.release: Do not show IPv6 warning for SQLite3

rhcarvalho:
The warning to update `config/runtime.exs` only applies to other supported databases.


josevalim:

PR #5648, opened 2023-11-22
Document using Mix to track JS dependencies from source

rhcarvalho:
This can be valuable to people who want to avoid Node.js in their
projects, while having a simpler way to track and update dependencies
straight from Git or path, etc.


josevalim:

PR #5655, opened 2023-12-01
chore: add `layout` to `put_layout*/2 specs

Gladear:
dialyzer shows an issue when using `plug :put_layout, {Layouts, :app}` whilst the syntax is valid (although, as far as I understood, not favoured).

Update the `@spec` of `put_layout/2` and `put_root_layout/2` to match this behaviour.

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -637,7 +637,7 @@ defmodule Phoenix.Controller do
 
   Raises `Plug.Conn.AlreadySentError` if `conn` is already sent.
   """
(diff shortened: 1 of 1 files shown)


Gladear:
Thanks for the explanation.

If anyone with the same problem as me sees this PR:
I did have a problem with the new syntax - the layout wasn't used. The was due to the use of the old syntax in the router, which takes precedence and there cannot be overriden by using the new syntax (in controllers or in other pipelines for example). You'll need to update the syntax in the router, or wherever you used it first to make it work correctly.

PR #5658, opened 2023-12-04
Add missing gettext calls in core_components.ex

### LINKED ISSUE #5649: Strings in generated apps are not fully covered by gettext
### Environment

* Elixir version (elixir -v): `any`
* Phoenix version (mix deps): `1.7.10`
* Operating system: `any`

### Actual behavior

1. Create a new app and extract strings for translation:

   ```shell
   mix phx.new  hello
   mix gettext.extract
   ```

2. App comes with `gettext` support, but many strings are not marked for translation. For example, out-of-the-box flash messages are only in English (`"We can't find the internet"`, `"Something went wrong!"`, ...)

### Expected behavior

Given the existing effort to ship `gettext` integration with new apps, it would be useful if all strings would come wrapped for translation, both to serve as an example and to reduce the toll on developers. While many of the strings in the generated code will be eventually replaced with something else, some likely stick around for the lifetime of the project.

I am extracting those strings for a new app, so I might as well send a PR for this if that would be welcome.

rhcarvalho:
Closes #5649.

Note: strings in `app.html.heex` and `home.html.heex` in the generated app are not "meant to stay", so, after ignoring them, the only thing left in a freshly generated app was `core_components.ex`, which already had 2 strings marked for translation -- the rest is updated in this PR.

---

Depending on the context of the template file, we need to output code with different syntax. To make maintenance easier and keep the with/without Gettext strings consistent, we add a helpers `maybe_*_gettext/2` to contextually call gettext or skip it, while outputting correct syntax. The helpers are internal and are never exposed to the generated code.

Both installer and Phoenix (through phx.gen.live) have a core_components.ex template file, and those must stay in sync. We don't want to create a dependency between Phoenix and the installer, so the helpers are implemented in both places.


installer/lib/phx_new/gettext_support.ex
```diff
@@ -0,0 +1,44 @@
...
+      iex> ~s|<tag>#{maybe_gettext("Hello", :eex, true)}</tag>|
+      ~S|<tag><%= gettext("Hello") %></tag>|
+
+      iex> ~s|<tag>#{maybe_gettext("Hello", :eex, false)}</tag>|
+      ~S|<tag>Hello</tag>|
+  """
+  @spec maybe_gettext(binary(), :heex_attr | :eex | :ex, boolean()) :: binary()
+  def maybe_gettext(message, role, gettext?)
+
+  def maybe_gettext(message, :heex_attr, gettext?) do
```
josevalim, line 29: Instead of the attribute, please define two functions? `attr_gettext` and `eex_gettext`? Also, we probably already have a module where we define helpers used across templates. Could these functions be moved there? I would try to avoid defining a module only for this purpose.

rhcarvalho: Hi José, thank you for the review!

I could not find a common helper module that both the installer and
Phoenix Mix tasks can access. My understanding from getting familiar with the
code was that dependencies are kept to a minimum, which I appreciate, and one
example of that being that there are two copies of `core_components.ex` kept in
sync by a test that ensures they match byte-for-byte.

A separate Gettext helper file/module seemed like the best option, as that can
be easily kept in sync with the same testing approach of reading from disk and
comparing the two copies.

If you have a different suggestion or know of a common module we could use,
please let me know and I'll be happy to update the patch.

> Instead of the attribute, please define two functions? `attr_gettext` and
`eex_gettext`?

I'm open to making that change. However, I want to ensure that the function
names accurately reflect their behavior. The current approach with
`maybe_gettext` indicates that the output may or may not contain `gettext`
calls. If we use names like `attr_gettext` and `eex_gettext`, it might suggest
that the output always contains `gettext` calls. Could you please clarify what
function signatures you had in mind, I think I might have misunderstood your
suggestion.

Once again, thank you for your feedback!

josevalim: To be clear, I didn't meant a common module for both phoenix and phx_new, but rather a common module within each of those packages instead of a new one in both.

I think `maybe_attr_gettext` and `maybe_eex_gettext` are also fine as function names. :)

rhcarvalho: > To be clear, I didn't meant a common module for both phoenix and phx_new, but rather a common module within each of those packages instead of a new one in both.

Understood. There will be two copies.

On the installer side, an obvious choice would be `installer/lib/phx_new/generator.ex`, because that is where we make the helper available for the template. The other files in `phx_new` are themed around the code they generate.

For Phoenix, we could go with `lib/mix/phoenix.ex`.

In that case, how would you suggest ensuring the helpers stay in sync? I believe the `File.read!` approach would not work as smoothly.

josevalim: I think we don't need them fully in sync, they are small functions anyway. The modules you suggested are good. :)

rhcarvalho: Done! I applied the suggested changes and rebased on top of the latest main branch.

installer/lib/phx_new/generator.ex
```diff
@@ -37,8 +37,50 @@ defmodule Phx.New.Generator do
...
+              #
+              #     iex> ~s|<tag>#{maybe_eex_gettext("Hello", false)}</tag>|
+              #     ~S|<tag>Hello</tag>|
+              defp maybe_eex_gettext(message, gettext?) do
+                if gettext? do
+                  ~s|<%= gettext(#{inspect(message)}) %>|
+                else
+                  message
+                end
+              end
```
josevalim, line 83: Instead of defining it inside the quote, define them as regular functions inside `PhxNew.Generator` which will be imported.

rhcarvalho: Done.

josevalim:

josevalim:

PR #5659, opened 2023-12-04
Add missing . to router cheat sheet filename

Gazwai:
### Description

This PR add a missing `.` in the router cheatsheet file name. So now it is `router.cheatmd` => `router.cheat.md`.


Gazler:
I believe `.cheatmd` is correct in this case. https://hexdocs.pm/ex_doc/cheatsheet.html

Issue #5660, opened 2023-12-09
Modal component in CoreComponents closes when clicked outside, may result in data loss


ArthurClemens:
For simple messages, closing the dialog when clicking the backdrop is fine. However, when used with forms, an accidental click can result in data loss. For this reason, dialogs often employ 'modal' behaviour, requiring an explicit action - typically via the Save or Cancel buttons.

The `modal` component can easily be improved by introducing a `modal` boolean and using the value in the `phx-click-away` attribute:

```html
phx-click-away={not @modal and JS.exec("data-cancel", to: "##{@id}")}
```

rhcarvalho:
Data loss could also be caused by the Esc key, or accidentally dismissing the changed form in any other way.

I think a key distinction is whether or not there's any unsaved changes in the form. If you click "New Item" and immediately dismiss the modal, nothing is lost. Similarly, clicking "Edit Item" and not changing anything, the modal can be dismissed without data loss.

The problem is when the form state is changed from its original state, do you agree?

If that's the case, then probably a better solution would consider the form state, instead of being statically baked into the modal markup.

What comes to mind are confirmations like "are you sure ... unsaved changes will be lost...", though design-wise a dialog-in-a-dialog is not great UX either, so I don't have a clear proposal here, sorry.

ArthurClemens:
I agree that the "unsaved changes" warning is a better fit for forms. 

To get this warning automatically, simply return `true` from the `beforeunload` callback.
The options for displaying this prompt are non-existent so we don't have to concern ourselves with it. Wiring up a custom dialog *could* be done but is not worth the effort. The prompt text can't be set: you get what the browser makers have decided.

```js
const handleBeforeUnload = (e) => {
  e.preventDefault();
  // return true; 
  e.returnValue = true; // for browser compatibility
};

window.addEventListener("beforeunload", handleBeforeUnload);
// and also: window.removeEventListener("beforeunload", handleBeforeUnload);
```

This can be implemented with a hook, using Ecto's changeset to conditionally set the return value , as documented in [this blog post by Patrick Thompson](http://blog.pthompson.org/alpine-js-and-liveview#beforeunload-hook_2).

But the big caveat is that `push_patch ` won't trigger `beforeunload` (nor `popstate`), so this method won't fly. And `destroy` can't be prevented. So I don't have a solution yet.

rhcarvalho:
Connecting the dots, @mxgrn sent this on the Elixir Forum:

[...]

PR #5661, opened 2023-12-09
Add opt-in modal behaviour to dialogs

ArthurClemens:
This PR resolves issue #5660

Changes:
- Added attribute `modal` to `CoreComponents`' `modal` component, which if true prevents the dialog from closing when clicking outside.
- Updated the form templates to use the `modal` attribute.

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -35,9 +35,11 @@ defmodule <%= @web_namespace %>.CoreComponents do
         This is another modal.
       </.modal>
 
+  To prevent the modal from accidentally closing when clicking on the backdrop, pass the 'modal' attr.
   """
   attr :id, :string, required: true
   attr :show, :boolean, default: false
+  attr :modal, :boolean, default: false
   attr :on_cancel, JS, default: %JS{}
   slot :inner_block, required: true
 
@@ -65,7 +67,7 @@ defmodule <%= @web_namespace %>.CoreComponents do
               id={"#{@id}-container"}
               phx-window-keydown={JS.exec("data-cancel", to: "##{@id}")}
               phx-key="escape"
-              phx-click-away={JS.exec("data-cancel", to: "##{@id}")}
+              phx-click-away={not @modal and JS.exec("data-cancel", to: "##{@id}")}
               class="shadow-zinc-700/10 ring-zinc-700/10 relative hidden rounded-2xl bg-white p-14 shadow-lg ring-1 transition"
             >
               <div class="absolute top-6 right-5">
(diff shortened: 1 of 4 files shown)


installer/templates/phx_web/components/core_components.ex
```diff
@@ -35,9 +35,11 @@ defmodule <%= @web_namespace %>.CoreComponents do
         This is another modal.
       </.modal>
 
+  To prevent the modal from accidentally closing when clicking on the backdrop, pass the 'modal' attr.
   """
   attr :id, :string, required: true
   attr :show, :boolean, default: false
+  attr :modal, :boolean, default: false
```
rhcarvalho, line 42: I like the idea of conditionally disabling the `phx-click-away`, but I feel the name of the attribute is a bit strange since the component is already called "modal".

Perhaps this was something you struggled with as well?

Brainstorming a bit:

- persistent
- sticky
- noBackdropDismiss
- preventBackdropClose

ArthurClemens: Well, I wouldn't have called it a `modal` of course :-)

Perhaps `blocking` has a similar ring.

NduatiK: `confirmClose` maybe?

ArthurClemens: These are separate concerns. Modal behaviour means that interacting with the main content is disabled until the user interacts with the dialog content. That means that clicking on the backdrop does nothing, not even asking for a confirmation.
This type of interaction is common for confirmation dialogs: they're not meant to be closed without clicking OK or Cancel.
In #5660 we agreed that asking for a confirmation provides a better UX for forms in a dialog. Nevertheless there will still be a need to be able to create actual modal dialogs.

mxgrn:
I would instead suggest to add an option for the modal to confirm on _any_ close action (clicking away, hitting 'Esc', pressing cancel etc).

josevalim:
The core components is a starting point, you are welcome to add and remove any functionality that you want once it is in your app. You don't have to stay with Phoenix recommendations here. :)

josevalim:
Actually, I have read the original issue and I understand it better. I am in favor of this but I would like a second pair of eyes. Apologies.

installer/templates/phx_web/components/core_components.ex
ArthurClemens: For accessibility, Escape to close should always be possible.

josevalim: That's the one that trips me the most. So perhaps the best solution is to ask for confirmation indeed. Let's see what Chris has to say after the holidays. :)

SteffenDE:
@ArthurClemens I'm sorry that we kept this open so long that now, since we simplified the live generators a while back and removed the modal component, it's become obsolete...

Therefore, I'm closing this. I do want to thank you very much for the contribution. If anyone comes across this and looks for a solution in their app, they can definitely use this as an example :)

PR #5665, opened 2023-12-12
Add missing gettext invocations in core components

sheerlox:
Added `gettext` to `flash_group` default error flashes.

Manually tested by using the modified version of the generator and running `mix gettext.extract`:

image


sheerlox:
I tried to address the issue with formatting (which depends on whether `gettext` is used or not) in the last commit. Please let me know if you can think of a more elegant solution!

rhcarvalho:
There's some overlap with #5658 I sent last week :)

sheerlox:
> There's some overlap with #5658 I sent last week :)

Oops, I should have checked! Closing in favor of yours :smile:

PR #5666, opened 2023-12-15
Fix flaky test for `Phoenix.Controller.current_path/2`

tanguilp:
With Elixir 1.15.7 and OTP26.2, this test randomly fails:

```
  1) test path and url generation current_path/2 allows custom nested query params (Phoenix.Controller.ControllerTest)
     test/phoenix/controller/controller_test.exs:789
     Assertion with == failed
     code:  assert current_path(conn, %{foo: %{bar: [:baz], baz: :qux}}) == "/?foo[bar][]=baz&foo[baz]=qux"
     left:  "/?foo[baz]=qux&foo[bar][]=baz"
     right: "/?foo[bar][]=baz&foo[baz]=qux"
     stacktrace:
       test/phoenix/controller/controller_test.exs:791: (test)
```

It relied on ordering of a map. Switching to a keyword list, we avoid this pitfall.

### DIFF
--- a/test/phoenix/controller/controller_test.exs
+++ b/test/phoenix/controller/controller_test.exs
@@ -788,7 +788,7 @@ defmodule Phoenix.Controller.ControllerTest do
 
     test "current_path/2 allows custom nested query params" do
       conn = build_conn_for_path("/")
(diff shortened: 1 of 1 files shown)


josevalim:

PR #5667, opened 2023-12-17
Use `origin` to protect against CSWSH instead of CSRF token

tanguilp:
This is a draft PR following discussion on https://elixirforum.com/t/http-caching-liveviews-first-render/59180

## Initial use-case

The use-case is HTTP caching of liveviews. Some liveviews contain none or almost no private data, and HTTP caching (with [`plug_http_cache`](https://github.com/tanguilp/plug_http_cache) for instance - I'm the author) could help speeding up first paint for the client. This was also requested to me to improve SEO ranking.

## The problem

Websockets are not constrained by same-origin policy, and are thus subject to [Cross-Site WebSocket hijacking](https://christian-schneider.net/CrossSiteWebSocketHijacking.html). In short, `evil.com` can open a WS to `example.com` and session data (cookies) are sent along. There are two ways of dealing with that:
- use session's CSRF token, and check it when opening the WS
- check the `origin` header

Phoenix uses CSRF tokens to mitigate against CSWSH. Therefore a LiveView, even a public one, cannot be cached by a public cache by default because the initial rendered HTML page contains a CSRF token that is unique to a session.

It is possible to disable session for a liveview: after all if a session is used, it means that there's some private data on that page. For example by configuring the socket this way:

```elixir
socket "/live", Phoenix.LiveView.Socket, websocket: [connect_info: []]**
```

instead of:

```elixir
socket "/live", Phoenix.LiveView.Socket, websocket: [connect_info: [session: @session_options]]
```

However, one might want to still have a session configured on a socket because:
- there are public and private pages and, for instance, `live_session` should still be enabled between these pages
- it allows publicly-cacheable private LiveViews

## Additional use-case

I've been thinking of *publicly-cacheable private LiveViews* lately (if you find a better name, you're welcome!). The idea is the following:
- on initial render, only public data is rendered. This HTML page can be cached
- when the liveview is mounted, it updates (hydrates?) the initially rendered page with private information

This pattern would not be suited for highly private pages. However some types of pages could benefit from it. Take for instance a typical catalog page of an ecommerce platform:
- most of the items of the page is public, cacheable and potentially costly to read in the DB: article title, price, rating, number of reviews, potential discount...
- top-right of the screen, typically, is the number of items in your basket, number of unread messages and so on, related to the user's account

Currently the CSRF check performed prevents from caching it.

I've seen some advice on elixirforums on doing the heavily loading not on initial render but after the liveview is mounted, so the pattern already exist, although the use-case is different.

Undoubtedly, this new pattern would offer new ways of shooting oneself in the foot. Some can be mitigated (a plug to prevent setting cookies on such liveviews), others probably not.

This is not a totally new idea in the world of HTTP caching: some have used ESI includes to do exactly that. The difference is that with the *publicly-cacheable private LiveViews*, the cached LiveView remains cacheable with uniquely public information. With the ESI approach, the generated page from cache becomes private and so another cache on the path to the browser cannot cache it.

## Proposed changes

I've opened this draft PR so as to discuss it.

- Is that a good idea in the first place? Are they some technical or security concerns that would advocate against doing that? I've tested caching a page on disk and it's still working after 2 days. Only top-level LiveViews can be cached.
- The PR misses tests and potentially other changes (`:cswsh_check` as a transport option like `:check_origin?`)
- Although it works well when a cookie is set, it doesn't when there's no cookie because of https://github.com/phoenixframework/phoenix_live_view/blob/main/lib/phoenix_live_view/channel.ex#L988: when `:session` is passed in `connect_info`, a session is mandatory (usually set by the `protect_from_forgery` plug). We might want to reconsider this as a user could land on such a liveview without any cookie (from direct link for instance). In this case we can still redirect to a page setting an cookie and redirecting back to the initial page though
- We'd need to update LiveView's javascript from:

```javascript
let csrfToken = document.querySelector("meta[name='csrf-token']").getAttribute("content")
let liveSocket = new LiveSocket("/live", Socket, {params: {_csrf_token: csrfToken}})    
```

to something like:

```javascript
let params = {}
if (document.querySelector("meta[name='csrf-token']")) {
  params = {_csrf_token: document.querySelector("meta[name='csrf-token']").getAttribute("content")}
}                                                                                                    
                                                                                                    
let liveSocket = new LiveSocket("/live", Socket, {params: params})
```

I'm very curious what you think about this proposal ☺ Please let me know if you consider I should go on with this idea and continue improving the PR, or if on the contrary there are some fundamental reasons not going further in this direction. Cheers!

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -213,6 +213,13 @@ defmodule Phoenix.Endpoint do
 
     * `:log_access_url` - log the access url once the server boots
 
+    * `:cswsh_check` - configure the CSWSH (Cross-Site WebSocket Hijacking) check
+    mechanism for the endpoint. When `:crsf_token` (the default), a CRSF token is
+    used. This mechanism uses the session and is stateful. When `:origin`, the
+    `origin` header is checked according to the rule set by the `:origin` option,
+    without requiring the use of cookies, but at the cost of not supporting older
+    browsers that don't send this header or other non-browser-based clients.
+
   Note that you can also store your own configurations in the Phoenix.Endpoint.
   For example, [Phoenix LiveView](https://hexdocs.pm/phoenix_live_view) expects
   its own configuration under the `:live_view` key. In such cases, you should
--- a/lib/phoenix/socket/transport.ex
+++ b/lib/phoenix/socket/transport.ex
@@ -338,7 +338,7 @@ defmodule Phoenix.Socket.Transport do
         conn
 
       origin_allowed?(check_origin, URI.parse(origin), endpoint, conn) ->
-        conn
+        put_private(conn, :phoenix_origin_checked, true)
 
       true ->
         Logger.error("""
@@ -488,14 +488,12 @@ defmodule Phoenix.Socket.Transport do
 
   defp connect_session(conn, endpoint, {key, store, {csrf_token_key, init}}) do
     conn = Plug.Conn.fetch_cookies(conn)
+    cswsh_config = endpoint.config(:cswsh_check, :csrf_token)
 
-    with csrf_token when is_binary(csrf_token) <- conn.params["_csrf_token"],
-         cookie when is_binary(cookie) <- conn.cookies[key],
+    with cookie when is_binary(cookie) <- conn.cookies[key],
          conn = put_in(conn.secret_key_base, endpoint.config(:secret_key_base)),
          {_, session} <- store.get(conn, cookie, init),
-         csrf_state when is_binary(csrf_state) <-
-           Plug.CSRFProtection.dump_state_from_session(session[csrf_token_key]),
-         true <- Plug.CSRFProtection.valid_state_and_csrf_token?(csrf_state, csrf_token) do
+         true <- check_cswsh(conn, session, csrf_token_key, cswsh_config) do
       session
     else
       _ -> nil
@@ -511,6 +509,18 @@ defmodule Phoenix.Socket.Transport do
         raise ArgumentError,
               "the MFA given to `session_config` must return a keyword list, got: #{inspect(other)}"
     end
+  end
+
+  defp check_cswsh(conn, session, csrf_token_key, :csrf_token) do
+    with csrf_token when is_binary(csrf_token) <- conn.params["_csrf_token"],
+         csrf_state when is_binary(csrf_state) <-
+           Plug.CSRFProtection.dump_state_from_session(session[csrf_token_key]) do
+      Plug.CSRFProtection.valid_state_and_csrf_token?(csrf_state, csrf_token)
+    end
+  end
+
+  defp check_cswsh(conn, _session, _csrf_token, :origin) do
+    conn.private[:phoenix_origin_checked]
   end
 
   defp fetch_x_headers(conn) do


lib/phoenix/socket/transport.ex
```diff
@@ -338,7 +338,7 @@ defmodule Phoenix.Socket.Transport do
         conn
 
       origin_allowed?(check_origin, URI.parse(origin), endpoint, conn) ->
-        conn
+        put_private(conn, :phoenix_origin_checked, true)
```
josevalim, line 341: Instead of passing this env var, what if we check `cswsh_check` in this function? If there is no origin or no origin check and `cswsh_check` is set to `:origin`, then the first clause must fail too.

lib/phoenix/endpoint.ex
```diff
@@ -213,6 +213,13 @@ defmodule Phoenix.Endpoint do
 
     * `:log_access_url` - log the access url once the server boots
 
+    * `:cswsh_check` - configure the CSWSH (Cross-Site WebSocket Hijacking) check
+    mechanism for the endpoint. When `:crsf_token` (the default), a CRSF token is
+    used. This mechanism uses the session and is stateful. When `:origin`, the
```
josevalim, line 218: I am not 100% sure I agree with this definition. We actually check both `:csrf_token` and `:origin`, it depends on which are provided. The thing is, because Phoenix does not allow you to read anything from headers, we actually don't need to check anything, unless you want to read the session. So maybe we call this `session_validator: :csrf_token | :origin`? Although it is weird to restrict it exclusively to session. 🤔  Thoughts?

tanguilp: > We actually check both :csrf_token and :origin, it depends on which are provided.

We check the `origin` header unless there's no `origin` header in the request, that is if:
- the client is not a browser
- the browser is too old and doesn't support this header (which is rarely the case nowadays)
- for some reason the browser didn't send the header (very unlikely example: browser has a vulnerability which is used by a malicious actor)

CSRF token is always verified as soon as `:session` is set in `:connect_info`. Which makes this mechanism intrusive by setting the CSRF token in the page, turning Liveviews into non-cacheable pages.

> The thing is, because Phoenix does not allow you to read anything from headers, we actually don't need to check anything, unless you want to read the session.

Indeed, in this case only the `origin` header is checked, if it is configured to do so *and* is present in the request. I'm not sure it's even necessary for public Liveviews. Might be some other authentication mechanism (not Phoenix session) that benefit this though.

> So maybe we call this session_validator: :csrf_token | :origin?

I don't have a strong opinion on this. I feel like it's slightly different of session validation (which, we could argue, is about checking the cookie has not been tampered with or has not expired for example). Basically it's about checking the cookie comes from a domain we control. `Phoenix.Transports.LongPoll` uses the same mechanism, but I haven't double check yet that checking the origin is sufficient from a security perspective. But in this case indeed `cswsh` makes no sense.

 It makes me think a bit about [RFC8471 - The Token Binding Protocol Version 1.0](https://datatracker.ietf.org/doc/html/rfc8471). Maybe `:session_binding`?

> Although it is weird to restrict it exclusively to session.

Probably not, unless another custom session mechanism is in use.

lib/phoenix/socket/transport.ex
tanguilp: Yes this is a good option as well.

We would still need some code in the other module that makes the difference between the 2 cases (CSRF check or no check because it's done here).

A third option if we want to avoid putting a value in the `conn` is to simply call again `Phoenix.Socket.Transport.check_origin/3` in the other module.

lib/phoenix/socket/transport.ex
```diff
@@ -513,6 +511,18 @@ defmodule Phoenix.Socket.Transport do
...
+  defp check_cswsh(conn, session, csrf_token_key, :csrf_token) do
+    with csrf_token when is_binary(csrf_token) <- conn.params["_csrf_token"],
+         csrf_state when is_binary(csrf_state) <-
+           Plug.CSRFProtection.dump_state_from_session(session[csrf_token_key]) do
+      Plug.CSRFProtection.valid_state_and_csrf_token?(csrf_state, csrf_token)
+    end
+  end
+
+  defp check_cswsh(conn, _session, _csrf_token, :origin) do
+    conn.private[:phoenix_origin_checked]
```
josevalim, line 523: Now that I think about it, is this the wrong solution to the problem? This session is not valid and we should not assume it is. All this does is to set the `session` to `nil`. So what if instead of change Phoenix.LiveView instead so, if the LiveView explicitly allows such, it does not require the session to be valid (i.e. it accepts nil and converts it to an empty map)?

tanguilp: I'm replying assuming you refer to make changes to LiveView as well. I mixed up goals in the initial explanation, so I'm going try to be clearer :sweat_smile: 

The goals are:
1. remove session data from Liveview's HTML to make them cacheable
2. remove the requirement of Liveview to have session data to enable visiting liveviews with or without cookie. This is a secondary goal and 1. is a prerequisite for it

To do that we would need:
a. to protect against websocket hijacking using `origin`, instead of session-based CSRF → this PR
b. make Liveview accept "empty" or `nil` session one way or another → not this PR

Now, because today we always check sessionfull liveviews with a CSRF token, there's always a session setup the:
```
            LiveView session was misconfigured or the user token is outdated.

            1) Ensure your session configuration in your endpoint is in a module attribute:

                @session_options [
                  ...
```
is displayed.

b. is not so important for now because there's always the possibility set a cookie before opening a liveview (via redirection).

> All this does is to set the session to nil. So what if instead of change Phoenix.LiveView so, if the LiveView explicitly allows such, it does not require the session to be valid (i.e. it accepts nil and converts it to an empty map)?

At first I changed the returned value here from `nil` to `%{}` when there's no session, but in that case LiveView will never show the above error message because there will always be an fake empty session. I'll make a separate PR with your proposed solution for `phoenix_live_view`. And I'll rename this PR as well.

josevalim: What I am saying is that I don’t think we need “a” (this PR). Browsers always send the origin and we always validate it if it is sent. Phoenix is already safe from CSWHJ. The session code is only validating if the session is valid, nothing more.

tanguilp: > The session code is only validating if the session is valid, nothing more.

Might be wrong but I think the session code in socket/transport performs CSRF check to prevent websocket hjiacking (CSWSH) specifically. There's no need to check CSRF for GET request for instance, right?

> Browsers always send the origin and we always validate it if it is sent.

If we assume that (which is true in 99.99% of cases I guess), then we can:
- let `origin` check as it is today
- remove the CSRF token check in the session code (if my assumption about the initial goal of its implementation is true)

What do you think?

josevalim: > Might be wrong but I think the session code in socket/transport performs CSRF check to prevent websocket hjiacking (CSWSH) specifically. There's no need to check CSRF for GET request for instance, right?

The session+csrf code checks if the session is valid/invalid. You can use the fact if the session is valid or invalid to protect against CSWSH by failing to accept connections with a nil session. My point is that the CSRF code here does not cancel and abort the request in any way, it leaves this decision to a future layer.

tanguilp: Yes, but my point is that you can't fetch the session without CSRF token, and checking against CSWSH with a CSRF token makes the liveview non-cacheable (in a shared = public cache), as the token is linked to the user session.

This is why I'm looking to implement a mechanism that:
- protects against CSWSH → using the `origin` header instead
- allows fetching the session → even in the absence of a CSRF token
- doesn't set private data in Liveview's initial render HTML → no use of a CSRF token embedded in the HTML

> My point is that the CSRF code here does not cancel and abort the request in any way

It does prevent the Liveview from mounting if user A request first the page which is cached and then user B fetches the cached page: CSRF fails to validate.

> The session+csrf code checks if the session is valid/invalid.

I'd say the plug session store checks the validity of the cookie (signature, expiration) and I believe CSRF in this case specifically protects against CSWSH. [
Plug.CSRFProtection ](https://hexdocs.pm/plug/Plug.CSRFProtection.html) says:

> GET requests are not protected, as they should not have any side-effect or change your application state. JavaScript requests are an exception: by using a script tag, external websites can embed server-side generated JavaScript, which can leak information. For this reason, this plug also forbids any GET JavaScript request that is not XHR (or AJAX)

which makes me think that since Liveview's request is a `GET` request.

josevalim: Ok, so the part we are disagreeing is this:

> allows fetching the session → even in the absence of a CSRF token

Why would want to fetch the session in a LiveView that was publicly cached?

This seems like a very dangerous place to be because, if you are relying on a session existing in the first place, then the page you are caching (which was most likely generated by the same LiveView) is using parts of the session.

So my point is: if you want to cache it, then the safest thing for us to do is to make sure you can't access the session at all. Not in the dead render, nor the live render.

> It does prevent the Liveview from mounting if user A request first the page which is cached and then user B fetches the cached page: CSRF fails to validate.

Yes, but this is LiveView logic, no Phoenix logic. :)

tanguilp: > Why would want to fetch the session in a LiveView that was publicly cached?

That's the *publicly-cacheable private LiveViews* use-case: in many cases only a few few elements are "private" (session-based). For example on my online shop's landing page the only session-based elements are: my name, the number of liked products and the number of products in the basket. The idea is to:
- cache the initial Liveview render - the page only contains public data
- let Liveview update the page with public & private elements through websocket

> This seems like a very dangerous place to be

Definitely! So many ways to shoot yourself in the foot. However this is a pattern that's already being used:
- varnish + ESI (Edge Side Includes) that do just that
- some commercial products that seem to use either that or ws

Since I published [Introducing `http_cache`, a BEAM-native standard-compliant HTTP caching library](http://svground.fr/blog/posts/introducing-http-cache/), I received a few messages from people wanting to do exactly this.

For information I published this example on elixirforums:

```elixir
def mount(params, session, socket) do
  if not connected?(socket) do
    # Initial request: here we render only **public** data.
    # The response can be cached by shared caches
    products = Products.list(params)

    {:ok, assign(socket, products: products)}
  else
    # Check user authentication
    socket =
      case session do
        %{"user_id" => user_id} ->
          assign(socket, user: Accounts.get_user!(user_id))

        _ ->
          socket
      end

    # Perform stateful stuff that don't depend on the user
    Phoenix.PubSub.subscribe(MyApp.PubSub, "product-updates")

    if socket.assigns[:user] do
      # Prepare rendering for authenticated users
      socket =
        socket
        |> assign(:pending_orders, User.get_pending_orders!(socket.assigns.user))
        |> assign(:pending_notifications, User.get_pending_notifications(socket.assigns.user))

      {:ok, socket}
    else
      # Prepare rendering for anonymous users
      {:ok, socket}
    end
  end
end
```

One of the goal of this PR is to determine if there's a chance to have it in Phoenix, given the security risks it indeed poses.

> I understand that you could use Vary: Cookie as for the HTTP Caching but then I'm not sure what you are meant to gain from HTTP caching with Vary: Cookie on a LiveView.

I think the most common case would be to not set cookie from Liveviews. I just updated `http_cache` yesterday to prevent setting cookies on shared caches.

josevalim:
> Definitely! So many ways to shoot yourself in the foot. 

What if we allow you putting a specific value in the session that says you don't want it to be validated with CSRF? The benefit is that it could be granular instead for the whole app all at once. This way we could hook everything together:

1. When doing the dead render, you clear up the session and only set this field, to make sure the LiveView does not actually use any session information on its cached render.
2. On the live render, this field will allow the session to continue.
3. We should document that you MUST have `check_origin` enabled if you want to use this feature.

niccolox:
> > Definitely! So many ways to shoot yourself in the foot.
> 
> What if we allow you putting a specific value in the session that says you don't want it to be validated with CSRF? The benefit is that it could be granular instead for the whole app all at once. This way we could hook everything together:
> 
> 1. When doing the dead render, you clear up the session and only set this field, to make sure the LiveView does not actually use any session information on its cached render.
> 2. On the live render, this field will allow the session to continue.
> 3. We should document that you MUST have `check_origin` enabled if you want to use this feature.

I think the way to do this is cache the DOM and not cache the DOM patch

tanguilp:
> Definitely! So many ways to shoot yourself in the foot.

I might have overstated the risk here. As soon as one does HTTP caching, there's a risk of caching private data or `set-cookie` header. I've updated `http_cache` to raise when setting cookies in shared caches by default.

We might limit the risk as well with best practice, for instance removing any user ID from the session before deadrendering:

```elixir
def mount(params, session, socket) do
  if not connected?(socket) do
    session = Map.delete(session, "user_id")

    # Initial request: here we render only **public** data.
    # The response can be cached by shared caches
    products = Products.list(params)

    {:ok, assign(socket, products: products)}
  else
  #...
```

> What if we allow you putting a specific value in the session that says you don't want it to be validated with CSRF? The benefit is that it could be granular instead for the whole app all at once.

Not sure why we would want it to be granular: for a given page, caching is either enabled or disabled for all users. Using the session also doesn't allow to differentiate between different paths.

> This way we could hook everything together:
> 
>     1. When doing the dead render, you clear up the session and only set this field, to make sure the LiveView does not actually use any session information on its cached render.
>     2. On the live render, this field will allow the session to continue.
>     3. We should document that you MUST have `check_origin` enabled if you want to use this feature.

How about:
- having an option to disable CRSF check (enabled by default) in `phoenix`. In that case `check_origin` must be set
- optional: having an option in `phoenix_live_view` to allow `nil` session (and turning it into `%{}`). Disabled by default. It's always possible to redirect back and forth to a set-cookie page, which is why this is optional
- possibly: allow setting the `phoenix` CSRF-disabling option at the live route level, to avoid setting it globally

josevalim:
> Not sure why we would want it to be granular: for a given page, caching is either enabled or disabled for all users. Using the session also doesn't allow to differentiate between different paths.

What I am saying is that it is granular per page. You don't disable session validation or change the CSWSH for the whole app, only on the pages you want to cache.

And it would be per page, because we would store this information on the data-phx-session metadata that is rendered in the page (and validated on the server).

josevalim:
I realize now that I only said "session" in my earlier comment and that was ambiguous. There are two sessions, the connection one and LiveView one. I meant the latter. :D In any case, here is what I propose more explicitly:

1. Disable the session in certain routes:

    live_session :disabled do
      live "/foo", FooLive
    end

2. Now, on dead render, we won't use the connection session will be empty when we call the `mount`. We will also store some metadata on data-phx-session (or whatever it is called) saying the session must be discarded

3. On the live render, if there is no session information from the socket AND the session was disabled, we don't log the current error, we continue as usual

I think strictly speaking we don't even need to tell people to use check origin because these pages should be public?

tanguilp:
> I think strictly speaking we don't even need to tell people to use check origin because these pages should be public?

Nope, because these pages are not public in the *publicly-cacheable private LiveViews* use-case (when private data is only processed during live render), and therefore we need to protect against CSWSH with origin-based check.

Sockets without the `:session` option are already cacheable, since we don't check the CSRF token in this case (and we can even omit it in the HTML).

> And it would be per page, because we would store this information on the data-phx-session metadata that is rendered in the page (and validated on the server).

Talking about `data-phx-session`:
- I haven't tried with `live_session` yet, only with plain liveviews → unsure how it would play in this case
- there is some data in `data-phx-session` that I'm not sure if it's really cacheable: `:id`, `:live_session` and token expiration set to 2 weeks:

```elixir
iex> phx_data_session |> String.split(".") |> Enum.at(1) |> Base.url_decode64!(padding: false) |> :erlang.binary_to_term()
{{5,
  %{
    id: "phx-F6RjBqqLbtyHxAAh",
    session: %{},
    parent_pid: nil,
    view: HttpCacheWithLiveviewWeb.UserLive.Show,
    root_pid: nil,
    router: HttpCacheWithLiveviewWeb.Router,
    live_session: {:default, 1703259653237121500},
    root_view: HttpCacheWithLiveviewWeb.UserLive.Show
  }}, 1703595439340, 86400}
```

For the rest of your comments (and thanks for the swift answers!), I need a bit more time to process them :sweat_smile:

tanguilp:
First of all, happy new year all :christmas_tree: 

> 1. Disable the session in certain routes:
>        ```
>         live_session :disabled do
>           live "/foo", FooLive
>         end
>        ```

Not sure if you mean that the `:disabled` name here has a special meaning?

Caching should work even if the liveview is not in a `live_session`,and it does with the current modifications. I've tested caching liveviews in live sessions, seems to not interfere with them.



> 2\. Now, on dead render, we won't use the connection session (it will be empty) when we call `mount`. We will also store some metadata on data-phx-session (or whatever it is called) saying the socket session must be discarded

I understand the idea to protect the user at a maximum, but I'm unsure if we want to do it to that extent. After all:
- one can always erroneously cache private data in shared caches. It's not a intrinsic feature of liveviews: deadviews are also prone to this
- it's not possible to cache `set-cookies` header because of how phoenix sets it just before calling the adapter - registered callbacks on `conns` cannot see it. That would be the most critical vulnerability and we already handle this case

> I think strictly speaking we don't even need to tell people to use check origin because these pages should be public?

As I replied before NY this mainly applies to *publicly-cacheable private LiveViews*, so we need the session. Only the deadrender template is public.

However maybe it can apply to public liveviews as well. Even if we don't use the `websocket: [connect_info: [session: @session_options]]`, the CSRF token might still be set if other liveviews are private (using session) and I don't think we want to store and serve public HTML pages with session-bound CSRF tokens in it.

josevalim:
Happy new year!

I have started writing a reply and it was clear we were going in loops, so I discarded that and decided to draw a line in the sand: I will be glad to review and accept a pull request to LiveView that implements my earlier comment. I will repeat them here for convenience (and also answering some of your questions):

1. Allow disabling the session in certain routes:

        live_session session: false do
          live "/foo", FooLive
        end

2. When session: false, on dead render, we won't use the connection session (it will be empty) when we call `mount`. We will also store some metadata on data-phx-session (or whatever a similar attribute) saying the socket session must be discarded

3. On the live render, if there is no session information from the socket AND the session was disabled (this is in phoenix_live_view/channel.ex), we don't log the current error, we continue as usual

This may not solve all of your cases but it should address some cases, including the ones described in the initial problem statement:

> Some liveviews contain none or almost no private data

Once this is working, I will be glad to discuss how we can move forward in a way that's safe to users. I ultimately don't think the solution to this problem belongs in Phoenix but I think both of us will better understand each other once the solution above is in place.

Thank you!

tanguilp:
Indeed we were going in circles :sweat_smile: Thanks for the feedback and the time you took to respond.

I'm still a bit reluctant with the solution the proposed solution. My main fear is that it'll bloat `phoenix` with non-generic code very specific to this unique use-case. I also feel we're trying to protect users to a too great extent here, because 1) you need to set this feature + `cache-control` header in the liveview pipeline + `plug_http_cache` to enable caching 2) caching private data in shared cache has always been an issue to deal with.

I think the best approach for now is to develop in a fork and iron out the concept with interested users.

Then I'll come back with feedback and better proposals. Again, thanks!

tanguilp:
We went full circle then :smile: I'm convinced changes are also need in `phoenix` (otherwise we can't use cookie session data since CSRF check fails).

I'll dig further, maintain a fork with these changes, write a blog post, collect feedback from users and in a few months we'll be able to assess if it's worth investing effort on it. Who knows, maybe *publicly-cacheable private LiveViews* is a bad idea for some reason after all :)

Cheers!

PR #5678, opened 2023-12-23
Pass classes into inputs in liveview core components

peteyycz:


josevalim:
You are welcome to make these changes into your app if desired. But we try to avoid adding functionality that we don't immediately use to core in our generated, otherwise it is easy to bloat them.

PR #5679, opened 2023-12-27
Set Heroicon default size to native icon size

rhcarvalho:
We recently added support for the new 'micro' variant with proper sizing (16x16 pixels).

This change updates the default size of the original icons (no suffix) and their 'solid' variant to be their native size (24x24 pixels).

The 'mini' variant remains at their native size (20x20 pixels).


josevalim:

PR #5683, opened 2023-12-28
Use umbrella apps names instead of "default"

### LINKED ISSUE #5631: Multiple web apps in umbrella project generate conflicting esbuild and tailwind configs
### Environment

* Elixir version (elixir -v): 1.15.6 (compiled with Erlang/OTP 26)
* Phoenix version (mix deps): 1.7.10
* Operating system: macOS Ventura 13.6

### Actual behavior

I've created a Phoenix umbrella app and generated a second web application into it.

This resulted in duplicated configs generated for `:esbuild` and `:tailwind` in `config/config.exs`:

```
config :esbuild,
  version: "0.17.11",
  default: [
    args:
      ~w(js/app.js --bundle --target=es2017 --outdir=../priv/static/assets --external:/fonts/* --external:/images/*),
    cd: Path.expand("../apps/hello_admin_web/assets", __DIR__),
  ]

config :esbuild,
  version: "0.17.11",
  default: [
    args:
      ~w(js/app.js --bundle --target=es2017 --outdir=../priv/static/assets --external:/fonts/* --external:/images/*),
    cd: Path.expand("../apps/hello_web/assets", __DIR__),
  ]
```

```
config :tailwind,
  version: "3.3.2",
  default: [
    args: ~w(
      --config=tailwind.config.js
      --input=css/app.css
      --output=../priv/static/assets/app.css
    ),
    cd: Path.expand("../apps/hello_admin_web/assets", __DIR__)
  ]

config :tailwind,
  version: "3.3.2",
  default: [
    args: ~w(
      --config=tailwind.config.js
      --input=css/app.css
      --output=../priv/static/assets/app.css
    ),
    cd: Path.expand("../apps/hello_web/assets", __DIR__)
  ]
```

When I run either the umbrella or a single app directly with `mix phx.server`, the `priv/static/assets/app.[css|js]` files will be generated only for one of the applications. It seems that the last definition of `:tailwind` and `:esbuild` will be the effective config.

### Expected behavior

`priv/static/assets/app.[css|js]` should be generated for all web apps in an umbrella.

agonzalezro:
When adding more than one app to your umbrella project you can't use default for tailwind and esbuild configurations or they will clash between them. By using their names instead of "default" the issue is fixed.

Closes #5631

Steps to test it (assuming you are [using phoenix mix from source code](https://github.com/phoenixframework/phoenix#generating-a-phoenix-project-from-unreleased-versions)):

1. `mix phx.new --umbrella /tmp/a`
2. `mix phx.new.web /tmp/a_umbrella/apps/b`
3. Change `config/runtime.exs` &  `config/dev.exs` to change one of the apps to listen in something that isn't `4000`.
4. `cd /tmp/a_umbrella && mix ecto.migrate`

### DIFF
--- a/installer/templates/phx_umbrella/apps/app_name_web/config/config.exs
+++ b/installer/templates/phx_umbrella/apps/app_name_web/config/config.exs
@@ -20,7 +20,7 @@ config :<%= @web_app_name %>, <%= @endpoint_module %>,
 # Configure esbuild (the version is required)
 config :esbuild,
   version: "0.17.11",
-  default: [
+  <%= @web_app_name %>: [
     args:
       ~w(js/app.js --bundle --target=es2017 --outdir=../priv/static/assets --external:/fonts/* --external:/images/*),
     cd: Path.expand("../apps/<%= @web_app_name %>/assets", __DIR__),
@@ -30,7 +30,7 @@ config :esbuild,
 # Configure tailwind (the version is required)
 config :tailwind,
   version: "3.3.2",
-  default: [
+  <%= @web_app_name %>: [
     args: ~w(
       --config=tailwind.config.js
       --input=css/app.css
--- a/installer/templates/phx_single/config/config.exs
+++ b/installer/templates/phx_single/config/config.exs
@@ -36,7 +36,7 @@ config :<%= @app_name %>, <%= @app_module %>.Mailer, adapter: Swoosh.Adapters.Lo
 # Configure esbuild (the version is required)
 config :esbuild,
   version: "0.17.11",
-  default: [
+  <%= @app_name %>: [
     args:
       ~w(js/app.js --bundle --target=es2017 --outdir=../priv/static/assets --external:/fonts/* --external:/images/*),
     cd: Path.expand("..<%= if @in_umbrella, do: "/apps/#{@app_name}" %>/assets", __DIR__),
@@ -46,7 +46,7 @@ config :esbuild,
 # Configure tailwind (the version is required)
 config :tailwind,
   version: "3.3.2",
-  default: [
+  <%= @app_name %>: [
     args: ~w(
       --config=tailwind.config.js
       --input=css/app.css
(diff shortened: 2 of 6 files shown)


josevalim:
We also need to change `mix.exs`. I would actually go ahead and change this for regular apps as well. :)

installer/templates/phx_umbrella/apps/app_name_web/config/config.exs
```diff
@@ -20,7 +20,7 @@ config :<%= @web_app_name %>, <%= @endpoint_module %>,
 # Configure esbuild (the version is required)
 config :esbuild,
   version: "0.17.11",
-  default: [
+  "<%= @web_app_name %>": [
```
josevalim, line 23: I think we don't need the quotes, the web app name should be a valid identifier.

agonzalezro:
@josevalim I've changed it for single apps as well and remove the extra quotes. However, I am not sure what needs to be done in `mix.exs`.

josevalim:
The aliases like assets.setup need to pass the namespace now, otherwise they default to “default”, no? You can generate the app and give it a try!

agonzalezro:
Understood @josevalim, fixed with the last commit. Thanks for the heads up!

agonzalezro:
The tests are failing because of the formatter, example:

```
❯ mix do format --check-formatted
...
** (Mix) mix format failed due to --check-formatted.
The following files are not formatted:

/private/tmp/w/mix.exs

       |
75 75  |      "assets.setup": ["tailwind.install --if-missing", "esbuild.install --if-missing"],
76 76  |      "assets.build": ["tailwind w", "esbuild w"],
77    -|      "assets.deploy":
78    -|        ["tailwind w --minify", "esbuild w --minify", "phx.digest"]
   77 +|      "assets.deploy": ["tailwind w --minify", "esbuild w --minify", "phx.digest"]
79 78  |    ]
80 79  |  end
       |
```

The problem I see now is that either with `inspect` or `IO.puts` I would need to mimic what the formatter does. Is there any reason why the output of the template generation isn't being formatted as another step?

josevalim:
You can always break them over multiple lines and the formatter will always be happy with it. :)

agonzalezro:
Could the faulty test be flaky? I can't rerun, maybe it would fix it?

Also, I am not extremely happy about how I solved the multi-line, any suggestion would be more than welcomed!

Thanks 🙏🏽

josevalim:

PR #5687, opened 2024-01-01
Update copyright year to 2014 - Present

devcamke:


chrismccord:
Present is assumed. Thanks!

PR #5689, opened 2024-01-06
Make test suites generated by phx.gen.auth use async: true

goncalotomas:
Test suites generated by `mix phx.gen.auth` conditionally set `async: true` if the database adapter is Postgres. Some of those test suites are missing that condition despite no apparent issues running tests concurrently.

I used these steps to test this out:
- Created a simple demo app using `mix phx.new`
- Add authentication code using `mix phx.gen.auth Accounts User users` and run migrations
- Run `for X in 1 2 3 4 5 6 7 8 9 10; do mix test; done`

I've ran these tests many times and I did not see any failures.
Originally posed as a question [here](https://elixirforum.com/t/why-arent-all-test-suites-generated-by-mix-phx-gen-auth-tagged-with-async-true/60812), pinging @josevalim as requested :)


josevalim:

PR #5693, opened 2024-01-11
using map.field notation (without parentheses) to invoke function Phoenix.Digester.Gzip.file_extensions() is deprecated, you must add parentheses instead: remote.function()

bradhanks:


chrismccord:
❤️❤️❤️🐥🔥

PR #5696, opened 2024-01-15
Add assert redirect in auth tests

IgnacioGoldchluk:
The `phx.auth` tests for `on_mount/4`  do not verify the `updated_socket.redirected` attribute against the expected values.

Live view tests and connection tests do verify redirects via `assert_redirected` and `redirected_to` function helpers respectively, but I could not find any helper for `%Phoenix.LiveView.Socket{}` structs, therefore the PR adds test to match against the `redirected` attribute. Not sure if this is the ideal way to test redirects for `on_mount/4`?

Additionally, the PR removes some extra whitespaces from a test name.


chrismccord:
The redirected type is internal and this kind of test is better suited to the integration level LoginLiveTest that we currently generated and assert the redirected URLs, so I don't think adding this at the unit level is necessary. Halting is what we really care about here. Thanks!

PR #5699, opened 2024-01-17
input validation errors must disappear when resolved

isimluk:
I came across an interesting a corner case.

Validation errors for an input field would never disappear, even after user amends their input to be valid. I noticed that heex template would be re-rendered on the server (`input()` method was called), but UI would never remove the last errors from the view.
   
Getting rid of `defaults: []` fixed this. It seems this is safe to do as errors are always assigned with 

```
    |> assign(:errors, Enum.map(field.errors, &translate_error(&1)))
````

Critical look at this pr is advised as I am newbie here.

Thank You for your work on this amazing work, I enjoyed myself exploring it during the last 14 days.


SteffenDE:
Hey @isimluk,

can you provide a sample project that reproduces your issue? I don't think that this is the correct solution, since I'm using the components as is without ever coming across this.

PR #5700, opened 2024-01-19
Omit meaningless markup from hidden input in core components

bopm:
## Motivation

The current version of core components given 
```
<.input type="hidden" field={@form[:user_id]} value={@current_user.id} />
```
produces next markup
```
<div phx-feedback-for="message[user_id]" data-phx-id="..." class="phx-no-feedback">
  <label for="message_user_id" class="block text-sm font-semibold leading-6 text-zinc-800" data-phx-id="..."></label>
  <input type="hidden" name="message[user_id]" id="message_user_id" value="2" class="mt-2 block w-full rounded-lg text-zinc-900 focus:ring-0 sm:text-sm sm:leading-6 phx-no-feedback:border-zinc-300 phx-no-feedback:focus:border-zinc-400 border-zinc-300 focus:border-zinc-400">
</div>
```

Most of that is completely meaningless for hidden fields.

## Change

After implementing these changes, it'll generate next markup:

```
<input type="hidden" name="message[user_id]" id="message_user_id" value="2" data-phx-id="...">
```

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -366,6 +366,17 @@ defmodule <%= @web_namespace %>.CoreComponents do
     """
   end
 
+  def input(%{type: "hidden"} = assigns) do
+    ~H"""
+    <input
+      type="hidden"
+      name={@name}
+      id={@id}
+      value={Phoenix.HTML.Form.normalize_value(@type, @value)}
+    />
+    """
+  end
+
   # All other inputs text, datetime-local, url, password, etc. are handled here...
   def input(assigns) do
     ~H"""
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -373,6 +373,7 @@ defmodule <%= @web_namespace %>.CoreComponents do
       name={@name}
       id={@id}
       value={Phoenix.HTML.Form.normalize_value(@type, @value)}
+      {@rest}
     />
     """
   end
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -368,13 +368,7 @@ defmodule <%= @web_namespace %>.CoreComponents do
 
   def input(%{type: "hidden"} = assigns) do
     ~H"""
-    <input
-      type="hidden"
-      name={@name}
-      id={@id}
-      value={Phoenix.HTML.Form.normalize_value(@type, @value)}
-      {@rest}
-    />
+    <input type="hidden" name={@name} id={@id} value={@value} {@rest} />
     """
   end
 
--- a/priv/templates/phx.gen.live/core_components.ex
+++ b/priv/templates/phx.gen.live/core_components.ex
@@ -366,6 +366,17 @@ defmodule <%= @web_namespace %>.CoreComponents do
     """
   end
 
+  def input(%{type: "hidden"} = assigns) do
+    ~H"""
+    <input
+      type="hidden"
+      name={@name}
+      id={@id}
+      value={Phoenix.HTML.Form.normalize_value(@type, @value)}
+    />
+    """
+  end
+
   # All other inputs text, datetime-local, url, password, etc. are handled here...
   def input(assigns) do
     ~H"""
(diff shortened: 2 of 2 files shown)


installer/templates/phx_web/components/core_components.ex
```diff
@@ -366,6 +366,17 @@ defmodule <%= @web_namespace %>.CoreComponents do
...
   end
 
+  def input(%{type: "hidden"} = assigns) do
+    ~H"""
+    <input
+      type="hidden"
+      name={@name}
+      id={@id}
+      value={Phoenix.HTML.Form.normalize_value(@type, @value)}
+    />
```
Gazler, line 376: We probably want a `{@rest}` in here to capture any additional provided attributes.

bopm: Rest consists of things that should not be in hidden input as well. At least to my understanding. https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/hidden

SteffenDE: Passing something like `form` can still be relevant for hidden inputs, therefore rest should be included in my opinion.

bopm: Added.

josevalim:
In other words, I would remove hidden from here: https://github.com/phoenixframework/phoenix/blob/main/priv/templates/phx.gen.live/core_components.ex#L277

josevalim:
We should remove radio as well to solve #5506 .

bopm:
> For what is worth, I don’t think we should support :hidden through `<.input`. Instead, I would just write the input directly.

I tried it too, leading to an unsafe input error. I am AFK to reproduce that error exactly, but it is something like that.

bopm:
Yep, the exact error is 
```
Phoenix.HTML.Safe not implemented for %Phoenix.HTML.FormField
```

SteffenDE:
@bopm you will have to do something like this when using `<input>` directly:
```
<input type="hidden" id={@form[:field].id} name={@form[:field].name} value={@form[:field}.value} />
```

You cannot pass the form field to a plain input tag, that's where the error comes from.

bopm:
@SteffenDE That works, but do I understand correctly that this way I am losing things like `Phoenix.HTML.Form.normalize_value`?

SteffenDE:
That's right, but normalize_value does nothing for "hidden" inputs anyway. As this PR is only about what is generated in new applications, you can of course keep the "hidden" case for the `<.input>` component in your app. That's what I do as well, as I see a benefit in not always writing out the attributes.

bopm:
That makes sense too, my other concern is that this general approach that @josevalim is offering will push developers to remember that some field types require different notation than others. For me, `don't make me think` feels like a better approach, as Phoenix is already complex enough, but I am unsure if I am getting it all right.

installer/templates/phx_web/components/core_components.ex
```diff
@@ -366,6 +366,18 @@ defmodule <%= @web_namespace %>.CoreComponents do
...
 
+  def input(%{type: "hidden"} = assigns) do
+    ~H"""
+    <input
+      type="hidden"
+      name={@name}
+      id={@id}
+      value={Phoenix.HTML.Form.normalize_value(@type, @value)}
+      {@rest}
+    />
```
josevalim, line 377:
```suggestion
    <input type="hidden" name={@name} id={@id} value={@value} {@rest} />
```

priv/templates/phx.gen.live/core_components.ex
```diff
@@ -366,6 +366,18 @@ defmodule <%= @web_namespace %>.CoreComponents do
...
 
+  def input(%{type: "hidden"} = assigns) do
+    ~H"""
+    <input
+      type="hidden"
+      name={@name}
+      id={@id}
+      value={Phoenix.HTML.Form.normalize_value(@type, @value)}
+      {@rest}
+    />
```
josevalim, line 377:
```suggestion
    <input type="hidden" name={@name} id={@id} value={@value} {@rest} />
```

josevalim:
I squeezed it all into a one liner, so maybe it is more amenable for merging.

PR #5703, opened 2024-01-25
Change example filename to fit the scenario ✅

dewetblomerus:
This is a very small nitpick, but the sentence in the docs reads clearer with this change.


Gazler:
THanks <3

PR #5706, opened 2024-01-30
Make Bandit the default web server for new Phoenix apps

mtrudel:
This PR:

* Makes Bandit the default web server for newly generated Phoenix apps
* Updates the Phoenix docs to reflect both Bandit and Cowboy as first-class supported adapters
* Keeps the default adapter as Cowboy if not explicitly specified for backwards compatibility
* Since Phoenix still supports (and runs CI) back to 1.11, Phoenix's internal tests are NOT updated to use Bandit here. There's a separate PR coming to add this, which will gate on Phoenix bumping to 1.12+

PR #5707, opened 2024-01-30
Add Bandit to tests in /integration

mtrudel:
Intended to follow #5706. Depends on Phoenix moving its minimum Elixir version to 1.12, since Bandit has hard test-dependencies thereon.


mtrudel:
Finally had a chance to look at the Cowboy slowness. Didn't / couldn't solve the underlying issue with Cowboy taking 5 seconds to terminate at the end of each test set. Moreover, we need to keep the Cowboy startup in a `setup_all` call (otherwise it gets started/stopped on every test, which adds 5s to every test instead of just to the entire test suite). 

As a result, the structure of these tests is a bit oddball, with the tests themselves macro'd into discrete Bandit/Cowboy test modules. The *vast* majority of the noise in this PR is just indenting to accommodate the macro-ing of the tests. 

Would love a better way to do this; if anyone has ideas I'm all ears.

Also, there's one failing setup in `test/phoenix/integration/websocket_channels_test.exs` that I can't figure out how to resolve. Hopefully more eyeballs make it transparent.

PR #5708, opened 2024-01-31
Core components: Fix flash messages obscured by dialog backdrop

ArthurClemens:
Use a z-index that is lower than the flash component's.


SteffenDE:
Modals are gone from the generated CoreComponents now, so this has become obsolete too... Thank you again for your contribution!

PR #5710, opened 2024-02-03
Fix references to 'Sign in' button rather than nav bar 'Log in' link

phinnaeus:
I noticed these tests start failing in my project when I removed the nav bar links. I think the intention was to detect the primary 'Sign in' button, but this has been detecting the 'Log in' link in the nav bar.

### DIFF
--- a/priv/templates/phx.gen.auth/login_live_test.exs
+++ b/priv/templates/phx.gen.auth/login_live_test.exs
@@ -8,7 +8,7 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
     test "renders log in page", %{conn: conn} do
       {:ok, _lv, html} = live(conn, ~p"<%= schema.route_prefix %>/log_in")
 
-      assert html =~ "Log in"
+      assert html =~ "Sign in"
       assert html =~ "Register"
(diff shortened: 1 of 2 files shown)


josevalim:
The navbar generates "Log in", so this should probably stay as is?

phinnaeus:
Thanks for taking a look! I assumed the tests in question were attempting to detect the content of the page (https://github.com/phoenixframework/phoenix/blob/main/priv/templates/phx.gen.auth/login_live.ex#L30) rather than the content of the navbar (https://github.com/phoenixframework/phoenix/blob/main/lib/mix/tasks/phx.gen.auth/injector.ex#L206) which, as it is present on all pages by default, would be the purview of more "generic" tests (which don't seem to exist right now).

That assumption might be wrong, as you say. Thinking about it some more, maybe its more appropriate to make one or more different changes instead:
* check for both the navbar and the main content elements intentionally through more specific selectors (of course they should stay as generic as possible, but distinguished in some way) either as new assertions in these tests or as separate tests
* go through the different "Log in"/"Login"/"Sign in"/"Signin" string usages and only use one of the consistently across the templates

josevalim:
Good find. I pushed consistent naming to main. :)

PR #5711, opened 2024-02-03
Fix typo in the contexts guide

rubynho:
There is a typo on the contexts guide where the field `cart_id` is mentioned as `cart-id`.

### DIFF
--- a/guides/contexts.md
+++ b/guides/contexts.md
@@ -626,7 +626,7 @@ We generated a new resource inside our `ShoppingCart` named `CartItem`. This sch
 +   create unique_index(:cart_items, [:cart_id, :product_id])
 ```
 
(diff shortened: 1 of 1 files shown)


guides/contexts.md
```diff
@@ -626,7 +626,7 @@ We generated a new resource inside our `ShoppingCart` named `CartItem`. This sch
...
+fWe used the `:delete_all` strategy again to enforce data integrity. This way, when a cart or product is deleted from the application, we don't have to rely on application code in our `ShoppingCart` or `Catalog` contexts to worry about cleaning up the records. This keeps our application code decoupled and the data integrity enforcement where it belongs – in the database. We also added a unique constraint to ensure a duplicate product is not allowed to be added to a cart. As with the `product_categories` table, using a multi-column index lets us remove the separate index for the leftmost field (`cart_id`). With our database tables in place, we can now migrate up:
```
rubynho, line 629:
```suggestion
We used the `:delete_all` strategy again to enforce data integrity. This way, when a cart or product is deleted from the application, we don't have to rely on application code in our `ShoppingCart` or `Catalog` contexts to worry about cleaning up the records. This keeps our application code decoupled and the data integrity enforcement where it belongs – in the database. We also added a unique constraint to ensure a duplicate product is not allowed to be added to a cart. As with the `product_categories` table, using a multi-column index lets us remove the separate index for the leftmost field (`cart_id`). With our database tables in place, we can now migrate up:
```

josevalim:

PR #5715, opened 2024-02-07
Add nodePaths to the example code

kuon:
When `esbuild` is called by elixir in the default template (`dev.exs`), it has the the following environment set:

```
```

This patch suggest a similar option using a `build.js` file. This helps the developer discover the option.


josevalim:

PR #5721, opened 2024-02-14
Add default argument to changeset generator in schema

srcrip:
Just a little proposal: I have seen this pattern a lot working with Phoenix and I think it's a better default, but if there's a great reason not to do it let me know. I think the default generators should really produce changesets with a default `%{}` parameter for the attrs parameter.

### DIFF
--- a/priv/templates/phx.gen.schema/schema.ex
+++ b/priv/templates/phx.gen.schema/schema.ex
@@ -13,7 +13,7 @@ defmodule <%= inspect schema.module %> do
   end
 
   @doc false
-  def changeset(<%= schema.singular %>, attrs) do
+  def changeset(<%= schema.singular %>, attrs \\ %{}) do
     <%= schema.singular %>
     |> cast(attrs, [<%= Enum.map_join(schema.attrs, ", ", &inspect(elem(&1, 0))) %>])
     |> validate_required([<%= Enum.map_join(Mix.Phoenix.Schema.required_fields(schema), ", ", &inspect(elem(&1, 0))) %>])


josevalim:
Please explain why you believe this is better. For instance, I can't think when/why you would call this function with no parameters to change it.

srcrip:
I mean I do this kind of thing in LiveView all the time:

```elixir
def mount(_params, _session, socket) do
  form =
    %NewThing{}
    |> NewThing.changeset()
    |> to_form()

  socket = assign(socket, :form, form)

  {:ok, socket}
end
```

Since you gotta do this for every 'initialize a new form for a schema' kind of situation, if I don't default my changeset params to `%{}` I'm always passing `%{}` around the call sites. It's also a pattern I've seen in a lot of other peoples apps, both professionally and in open source. when I search for:

https://github.com/search?q=language%3AElixir+%2Fdef+changeset.*%25%5C%7B%5C%7D%5C%29+do%2F&type=code

(hopefully that link works for other people)

I get 9k responses back on GH. So I think its a pretty popular pattern! It'd be awesome to make it the default. I also don't see a lot of downside too it, unless you think this encourages some kind of anti pattern. It's really just a convenience.

srcrip:
Thats a fine pattern and all, but maybe you're going to use this changeset somewhere else in which case its not going to harm anything to have the default applied in the schema module.

srcrip:
Also I'm partial to the idea that you want to push people in a certain direction, but the fact that this pattern is so common on GH I think points towards this change probably being beneficial

chrismccord:
Since we expose a default arg on the changeset one, I'm not convinced it is necessary here, but of course it's fine for you to also default it if you'd like in your apps. Thanks!

josevalim:
@srcrip `def changeset` is indeed very popular, because that's what Phoenix generates. The question is if people are calling directly from their LiveViews and so on. Phoenix already provides defaults at the context level and we can provide more if necessary, but not at the schema level (IMO).

srcrip:
I understand if you guys don't want to hash it out in GitHub Pull Requests, I can make an Elixir Forums post if that would be better?

josevalim:
> i understand what you're saying about do it in the context, but what if your app grows and you start to share common schemas across a few contexts? it would at that point become simpler to set it in the schema.

Then you add it to the schemas that need it. When working on the generators, we very intentionally avoid “future coding” or “what if” scenarios. Otherwise it is easy to end-up with bloat or APIs that nobody use (just because someone might).

PR #5722, opened 2024-02-14
Use `get_env/3` rather than `fetch_env!/2` in conditional check

JuneKelly:
[`Application.fetch_env!/2`](https://hexdocs.pm/elixir/1.16.1/Application.html#fetch_env!/2) will raise if the key is not found, crashing the Phoenix application. By using [`Application.get_env/3`](https://hexdocs.pm/elixir/1.16.1/Application.html#get_env/3) instead, this conditional check works as intended, calling `Phoenix.Logger.install()` only when the logger is configured and the value of the configuration is truthy.

## Context

While reading the phoenix source code, I noticed what I _think_ is a (harmless) error in `Phoenix.start/2`. We're checking for the value of the `:logger` configuration in an `if` conditional, but we're getting the value by calling `Application.fetch_env!/2`, which raises if the value is not found. It looks like we should be using `Application.get_env/3` instead, which returns `nil` in case the value is not found.

Now, it's possible that I've misread this situation, and we expect `:logger` configuration to always be present, in which case we don't need this change 😄

### DIFF
--- a/lib/phoenix.ex
+++ b/lib/phoenix.ex
@@ -18,7 +18,7 @@ defmodule Phoenix do
       :erlang.system_flag(:backtrace_depth, stacktrace_depth)
     end
 
-    if Application.fetch_env!(:phoenix, :logger) do
+    if Application.get_env(:phoenix, :logger) do
       Phoenix.Logger.install()
     end
 


chrismccord:
You are right that this kind of check could fail if application env is missing, but in Phoenix's case we provide the default config in Phoenix's mix file:
https://github.com/phoenixframework/phoenix/blob/main/mix.exs#L60-L61

So this check should always return a value and will not raise. Thanks!

PR #5724, opened 2024-02-14
Fix bug regarding token deletions upon successful password resets

roberto-aguilar:
In the default authentication system created by the `mix phx.gen.auth`, when a password has been successfully reset, it deletes all of the tokens regardless of their context, however, this is problematic in the following scenario:

- A user has been registered, which creates a token with the `confirm` context and account confirmation instructions delivered via email.
- The user has not clicked on the confirmation email message yet.
- The user requests password reset instructions and gets them via email.
- The user successfully follows the password reset instructions.
- The user tries to click on the confirmation email, but it is no longer valid.

By scoping the deletion to only `reset_password` tokens, the bug is gone and the `confirm` token will still be valid regardless of the abovementioned process.

### DIFF
--- a/priv/templates/phx.gen.auth/context_functions.ex
+++ b/priv/templates/phx.gen.auth/context_functions.ex
@@ -335,7 +335,7 @@
   def reset_<%= schema.singular %>_password(<%= schema.singular %>, attrs) do
     Ecto.Multi.new()
     |> Ecto.Multi.update(:<%= schema.singular %>, <%= inspect schema.alias %>.password_changeset(<%= schema.singular %>, attrs))
-    |> Ecto.Multi.delete_all(:tokens, <%= inspect schema.alias %>Token.by_<%= schema.singular %>_and_contexts_query(<%= schema.singular %>, :all))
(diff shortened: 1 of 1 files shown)


josevalim:
Resetting the password should typically invalidate all other sessions, requests to change emails, etc. It doesn't need to reset the confirmation token indeed, but almost everything else is a must.

ShPakvel:
> Resetting the password should typically invalidate all other sessions, requests to change emails, etc. It doesn't need to reset the confirmation token indeed, but almost everything else is a must.

@josevalim, this sounds reasonable for me as well. I've added necessary changes to keep confirmation token on this case. There is also other case (email change) that I think should be improved from security point of view. Some corrections. And optional idea described to consider as well. Could you please review it? Thank you in advance. 🙇🏼  https://github.com/phoenixframework/phoenix/pull/5951

SteffenDE:
Thank you for your contribution, but I'm afraid that we don't need this change any more with the new magic link generators in Phoenix 1.8. Authenticating by magic link will always confirm and setting the password is not possible without confirming first.

For more details, see https://github.com/phoenixframework/phoenix/issues/6041 and the related PR :)

PR #5727, opened 2024-02-19
Correctly describe Head Plug

ritchiey:
The [Head Plug](https://github.com/elixir-plug/plug/blob/85f03493afd78441b1ee2692b198641ccbee6bdc/lib/plug/head.ex#L18) doesn't strip the body from the response, it just converts `HEAD` requests to `GET` requests.


josevalim:

PR #5729, opened 2024-02-21
Updates the release guide Docker example

crenwick:
The release guide provides an example Docker image that is generated from running `mix phx.gen.release --docker`. 

However, the example in the guide out of date and includes a bug 🐛 (the location of `COPY lib lib`).

This simply updates the example in the guide with the updated Docker image file.


Gazler:
Thanks <3

PR #5733, opened 2024-02-25
Fix CSS transition duration in generators

balexand:
The value of the Tailwind CSS transition duration class should match the value of the `:time` option passed to [`Phoenix.LiveView.JS.show/1`](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.JS.html#show/1) and [`Phoenix.LiveView.JS.hide/1`](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.JS.html#hide/1).

This PR fixes cases where the Tailwind `duration-300` class was used when the `:time` option was left at the default of `200`.


SteffenDE:
Thank you! Closing this in favor of #5836.

PR #5736, opened 2024-02-29
Generate certificate with a valid end date on a February 29

quitejonny:
### Problem

We could not generate a certificate with phx.gen.cert today because of an invalid end date.
With a starting date of 2024-02-29 the task would generate an invalid end date of 2025-02-29.

### Solution

This PR fixes this bug by calculating an end 365 days in the future.

### DIFF
--- a/lib/mix/tasks/phx.gen.cert.ex
+++ b/lib/mix/tasks/phx.gen.cert.ex
@@ -235,20 +235,9 @@ defmodule Mix.Tasks.Phx.Gen.Cert do
   defp new_cert(public_key, common_name, hostnames) do
     <<serial::unsigned-64>> = :crypto.strong_rand_bytes(8)
 
-    # Dates must be in 'YYMMDD' format
-    {{year, month, day}, _} =
-      :erlang.timestamp()
-      |> :calendar.now_to_datetime()
-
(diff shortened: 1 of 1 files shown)


chrismccord:
❤️❤️❤️🐥🔥

lib/mix/tasks/phx.gen.cert.ex
```diff
@@ -270,6 +259,19 @@ defmodule Mix.Tasks.Phx.Gen.Cert do
...
+    {year, month, day} =
+      date
+      |> Date.add(days_to_add)
+      |> Date.to_erl()
+
+    yy = year |> Integer.to_string() |> String.slice(2, 2)
+    mm = month |> Integer.to_string() |> String.pad_leading(2, "0")
+    dd = day |> Integer.to_string() |> String.pad_leading(2, "0")
+
+    yy <> mm <> dd
```
wojtekmach, line 272: @quitejonny you can also use Date.to_iso8601(date, :basic) which does all of the above. The I don’t think we need the helper function. :) could you send another patch?

quitejonny: I opened a new PR #5738 which uses the Date.to_iso8601 function instead of a custom implementation.

PR #5738, opened 2024-03-01
Use Date.iso8601 function to calculate start and end date for the generated certificate

quitejonny:
Use Date.iso8601 function instead of custom implementation to calculate start and end date for the generated certificate


rud (approved):
Good plan

PR #5742, opened 2024-03-04
remove `view` from doc

ruslandoga:
As "view" mentions have been removed from the `moduledoc` and were replaced by "component" I thought it might make sense to remove it from `__using__` doc as well.


josevalim:

PR #5745, opened 2024-03-07
Clarify force_ssl in Heroku deployment guide

jc00ke:
According to the Using SSL guide, `force_ssl` is compile time to should be set in a file other than `runtime.exs`.

I also took the opportunity to clarify Heroku's "best practices" as being the 12-Factor App methodology and to improve some of the working.

This config works for me, but if I missed something, please let me know!


josevalim:

PR #5746, opened 2024-03-08
enable runtime checks for new lv projects in dev env

SteffenDE:
Waiting for https://github.com/phoenixframework/phoenix_live_view/pull/3161


josevalim:
I wonder if we should enable this one also for tests. Sometimes it is really easy to miss something during dev... :)

chrismccord:
I'll enable in test as well. Thanks! ❤️❤️❤️🐥🔥

PR #5747, opened 2024-03-09
Update comment in socket.ex

thoughtarray:
Very small docs correction. Saw it as I was reading the docs myself. Incorrect use of possessive as plural. I picked upcase for acronym "ID". Downcase would be fine too.


josevalim:

PR #5750, opened 2024-03-11
Extending phx.gen.* functions to support custom generation type handlers.

noizu:
Custom generators provide an appropriate migration field type, schema field type, live view form input and input hydration hooks. This allows projects to build out custom/house type live view form input handlers and use of engine specific type mapping such as postgres user types, jsonb, etc. without the need to manually edit generated live views.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -1,3 +1,13 @@
+# Intellij
+.idea/
+installer/phx_new.iml
+installer/templates/phx_single/phx_single.iml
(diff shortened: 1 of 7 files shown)


josevalim:
We don't plan to go down this path at the moment, thank you. The generators are meant to be a starting point, so it is expected that you have to tweak them, especially when it comes to more complex types.

PR #5766, opened 2024-03-29
Add ability to customize the primary key name

saleyn:
This change allows to use a custom primary key field name in projects where the `id` is [not a conventional option](https://github.com/jarulraj/sqlcheck/blob/master/docs/logical/1004.md).

## Example
```
$ mix phx.gen.live Schema User user name:string:unique --table user --binary-id --primary-key user_id
```

### DIFF
--- a/priv/templates/phx.gen.live/show.html.heex
+++ b/priv/templates/phx.gen.live/show.html.heex
@@ -1,5 +1,5 @@
 <.header>
-  <%= schema.human_singular %> <%%= @<%= schema.singular %>.id %>
+  <%= schema.human_singular %> <%%= @<%= schema.singular %>.<%= schema.opts[:primary_key] || :id %> %>
   <:subtitle>This is a <%= schema.singular %> record from your database.</:subtitle>
   <:actions>
     <.link patch={~p"<%= schema.route_prefix %>/#{@<%= schema.singular %>}/show/edit"} phx-click={JS.push_focus()}>
@@ -17,7 +17,7 @@
 <.modal :if={@live_action == :edit} id="<%= schema.singular %>-modal" show on_cancel={JS.patch(~p"<%= schema.route_prefix %>/#{@<%= schema.singular %>}")}>
   <.live_component
     module={<%= inspect context.web_module %>.<%= inspect Module.concat(schema.web_namespace, schema.alias) %>Live.FormComponent}
-    id={@<%= schema.singular %>.id}
+    id={@<%= schema.singular %>.<%= schema.opts[:primary_key] || :id %>}
     title={@page_title}
     action={@live_action}
     <%= schema.singular %>={@<%= schema.singular %>}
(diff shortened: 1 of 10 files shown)


saleyn:
It's been several months since this PR was submitted, and there's no indication it was reviewed.  Please provide feedback.

priv/templates/phx.gen.live/show.html.heex
```diff
@@ -17,7 +17,7 @@
 <.modal :if={@live_action == :edit} id="<%= schema.singular %>-modal" show on_cancel={JS.patch(~p"<%= schema.route_prefix %>/#{@<%= schema.singular %>}")}>
   <.live_component
     module={<%= inspect context.web_module %>.<%= inspect Module.concat(schema.web_namespace, schema.alias) %>Live.FormComponent}
-    id={@<%= schema.singular %>.id}
```
haljpg, line 20: 13085

haljpg: Any good ?

haljpg (changes requested):

priv/templates/phx.gen.schema/schema.ex
```diff
@@ -2,8 +2,9 @@ defmodule <%= inspect schema.module %> do
   use Ecto.Schema
   import Ecto.Changeset
 <%= if schema.prefix do %>
-  @schema_prefix :<%= schema.prefix %><% end %><%= if schema.binary_id do %>
-  @primary_key {:id, :binary_id, autogenerate: true}
+  @schema_prefix :<%= schema.prefix %><% end %><%= if schema.opts[:primary_key] do %>
+  @derive {Phoenix.Param, key: :<%= schema.opts[:primary_key] %>}<% end %><%= if schema.binary_id do %>
+  @primary_key {:<%= schema.opts[:primary_key] || :id %>, :binary_id, autogenerate: true}
```
SteffenDE, line 7: I think we also need an else case if we have a custom primary key name, but not using binary ids?

SteffenDE (commented):
One comment from my side. Apart from that lgtm.

priv/templates/phx.gen.live/show.html.heex
saleyn: I am a bit unclear about your comment.  The suggested change conditionally adds a prefix using either `:id` or the name of the `primary_key` if one is defined:
```elixir
-    id={@<%= schema.singular %>.id}
+    id={@<%= schema.singular %>.<%= schema.opts[:primary_key] || :id %>}
```

SteffenDE: I also don't see anything wrong with that line.

priv/templates/phx.gen.schema/schema.ex
saleyn: You are right.  I'll update and test in a little bit.

saleyn:
@SteffenDE, I addressed your concern, please approve.

SteffenDE:
@saleyn sorry for letting this sit here so long! I created a new commit because of merge conflicts and added you as co author. Thank you!

saleyn:
@SteffenDE, will you address the integration test issues:
https://github.com/phoenixframework/phoenix/actions/runs/12037126907/job/33559880380

SteffenDE:
This is already merged on main and the pipeline is happy. I’ll need to follow up though and make the same changes to phx.gen.html and similar as well for Phoenix 1.8. You can submit a PR if you want.

PR #5773, opened 2024-04-05
replace deprecated live_flash for Phoenix.Flash.get in phx.gen.auth

guissalustiano:
The [Phoenix docs say](https://hexdocs.pm/phoenix_live_view/0.20.12/Phoenix.Component.html#live_flash/2):
> `live_flash(other, key)`
> This function is deprecated. Use Phoenix.Flash.get/2 in Phoenix v1.7+. 

But the live_flash is still being used on generated auth, this PR fixes it.
I didn't find any other place that uses the `live_flash/2`.

Thanks!


chrismccord:
❤️❤️❤️🐥🔥

guissalustiano:
Wow, that was very fast.
Thanks!

PR #5775, opened 2024-04-07
Remove unnecessary control-flow from registration_live.ex

alexandremcosta:
It replaces the `if` branches by the `changeset.valid?` boolean.
Just a very small change, please dismiss if you find it less clear.

Also a question, does the `to_form` call right above really needs the `as:` option?

Thanks!


alexandremcosta:
Nevermind I checked the UI, now I understand why you don't want to always pass this option.

PR #5776, opened 2024-04-08
Dark mode support for code reloader error pages

jonnytest1:
could use some thoughts on which colors to pick but hey its dark ^^

image

### DIFF
--- a/lib/phoenix/code_reloader.ex
+++ b/lib/phoenix/code_reloader.ex
@@ -121,6 +121,10 @@ defmodule Phoenix.CodeReloader do
 
 
         @media (prefers-color-scheme: dark) {
+            body{
+                background: #000;
+            }
+
             html .top-details {
                 background: #000;
                 color: #fff;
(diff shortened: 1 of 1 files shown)


ryanwinchester:
The text needs some more contrast but this would be very helpful while streaming programming so as to not flashbang viewers

03juan:
I think @style.accent also should be tweaked, but that's in a lot of different selectors

03juan:
Alternatively, also provide assigns for background colors, and let users override it by implementing their own debug handler and passing custom fg, bg, accent, etc. colors

jonnytest1:
yeah should definitely be configurable since the other colors are , it would be very counterintuitive if the defined color isnt the one displayed

alternative 

    filter: invert(1);
   
can also be considered
   
image

rhcarvalho:
Hi @jonnytest1 :purple_heart:

Considering the code changes, I'd suggest updating the title of the PR to help the Phoenix team more quickly understand the scope when triaging the open PRs. E.g.: "Dark mode support for code reloader error pages".

SteffenDE:
Closing in favor of https://github.com/phoenixframework/phoenix/pull/6093. Thank you!

PR #5784, opened 2024-04-17
Generate text data type as textarea

acangiano:
Currently, `mix phx.gen.html` (and `mix phx.gen.live`) generate a single-line input field even if the user specifies `text` as the data type. For example:

`mix phx.gen.html Blog Post posts title:string body:text`

will generate the following form which is not what most users would expect.

Posts

In essence, there is no distinction between `string` and `text` types as far as the generated HTML goes (though, the distinction exists in the generated migration file). 

This small patch allows for fields declared as `text` to be generated as `<textarea>`.

improved-posts

### DIFF
--- a/lib/mix/tasks/phx.gen.html.ex
+++ b/lib/mix/tasks/phx.gen.html.ex
@@ -4,7 +4,7 @@ defmodule Mix.Tasks.Phx.Gen.Html do
   @moduledoc """
   Generates controller with view, templates, schema and context for an HTML resource.
 
-      mix phx.gen.html Accounts User users name:string age:integer
+      mix phx.gen.html Accounts User users name:string age:integer about:text
 
   The first argument, `Accounts`, is the resource's context.
   A context is an Elixir module that serves as an API boundary for closely related resources.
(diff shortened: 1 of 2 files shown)


lib/mix/tasks/phx.gen.html.ex
```diff
@@ -4,15 +4,16 @@ defmodule Mix.Tasks.Phx.Gen.Html do
   @moduledoc """
   Generates controller with view, templates, schema and context for an HTML resource.
 
-      mix phx.gen.html Accounts User users name:string age:integer
+      mix phx.gen.html Accounts User users name:string age:integer about:text
```
rhcarvalho, line 7: I find that using a `textarea` for `text`-type fields is a positive change, but the current docs changes do not help convey that distinction. It is probably not worth documenting anyway, as people are free and encouraged to change the generated code.

IMHO I'd keep the example `mix` call shorter as it was.

acangiano: That's fair enough. :)

acangiano:
I removed the text example from the docs.

SteffenDE:
Lovely! Thank you very much! 🙌🏻

PR #5785, opened 2024-04-17
Clarify log level for Phoenix.Channel

stilist:
The previous language wasn't explicit enough that setting the log level changes which level is used, not which logs are emitted — I thought `log_join: :debug` would add additional messages, not make `"join"` events be debug-level.

### DIFF
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -335,13 +335,16 @@ defmodule Phoenix.Channel do
   ## Logging
 
   By default, channel `"join"` and `"handle_in"` events are logged, using
-  the level `:info` and `:debug`, respectively. Logs can be customized per
-  event type or disabled by setting the `:log_join` and `:log_handle_in`
-  options when using `Phoenix.Channel`. For example, the following
-  configuration logs join events as `:info`, but disables logging for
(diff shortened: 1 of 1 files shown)


lib/phoenix/channel.ex
```diff
@@ -335,14 +335,17 @@ defmodule Phoenix.Channel do
...
+  the level `:info` and `:debug`, respectively. You can change the level used
+  for each event, or disable logs, per event type by setting the `:log_join`
+  and `:log_handle_in` options when using `Phoenix.Channel`. For example, the
+  following configuration logs join events as `:info`, but disables logging for
   incoming events:
 
       use Phoenix.Channel, log_join: :info, log_handle_in: false
 
+  Note that changing an event type's level doesn't affect what is logged,
+  unless you set it to `false`; it affects the associated level.
```
josevalim, line 347:
```suggestion
  unless you set it to `false`, it affects the associated level.
```

lib/phoenix/channel.ex
```diff
@@ -335,14 +335,17 @@ defmodule Phoenix.Channel do
...
+
```
josevalim, line 348:
```suggestion
```

josevalim:

PR #5788, opened 2024-04-21
fix code comment.

simonelnahas:


SteffenDE:
I'm confused. This looks exactly like https://github.com/phoenixframework/phoenix/pull/5780 and on main the backtick is there.

Issue #5786, opened 2024-04-21
Swoosh crashes if phx.new.ecto does not get run with --no-mailer


adonig:
### Environment

* Elixir version (elixir -v): 1.16.2
* Phoenix version (mix deps): 1.7.12
* Operating system: Fedora Linux 39 (Workstation Edition)

### Actual behavior

```
$ mix phx.new.ecto domain
* creating domain/lib/domain/application.ex
* creating domain/lib/domain.ex
* creating domain/test/test_helper.exs
* creating domain/README.md
* creating domain/mix.exs
* creating domain/.gitignore
* creating domain/.formatter.exs
* creating domain/lib/domain/repo.ex
* creating domain/priv/repo/migrations/.formatter.exs
* creating domain/priv/repo/seeds.exs
* creating domain/test/support/data_case.ex

Fetch and install dependencies? [Yn] 
* running mix deps.get
* running mix deps.compile

We are almost there! The following steps are missing:

    $ cd domain

Then configure your database in config/dev.exs and run:

    $ mix ecto.create

You can run your app inside IEx (Interactive Elixir) as:

    $ iex -S mix

$ cd domain
$ mix ecto.create
Compiling 3 files (.ex)
Generated domain app
The database for Domain.Repo has been created
$ iex -S mix
10:40:25.899 [error] Could not find hackney dependency.

Please add :hackney to your dependencies:

    {:hackney, "~> 1.9"}

Or set your own Swoosh.ApiClient:

    config :swoosh, :api_client, MyAPIClient

10:40:25.907 [notice] Application swoosh exited: exited in: Swoosh.Application.start(:normal, [])
    ** (EXIT) an exception was raised:
        ** (RuntimeError) missing hackney dependency
            (swoosh 1.16.3) lib/swoosh/api_client/hackney.ex:26: Swoosh.ApiClient.Hackney.init/0
            (swoosh 1.16.3) lib/swoosh/api_client.ex:43: Swoosh.ApiClient.init/0
            (swoosh 1.16.3) lib/swoosh/application.ex:7: Swoosh.Application.start/2
            (kernel 9.2.2) application_master.erl:293: :application_master.start_it_old/4
10:40:25.932 [notice] Application finch exited: :stopped
10:40:25.932 [notice] Application nimble_options exited: :stopped
10:40:25.934 [notice] Application nimble_pool exited: :stopped
10:40:25.936 [notice] Application mint exited: :stopped
[...]

PR #5768, opened 2024-03-31
Clarify Token secrecy in docs

ollien:
Closes #5757

PR #5789, opened 2024-04-22
Set the default for mailer depending on the value of ecto

adonig:
Sets the default value for the mailer binding (previously true) to the complement of the value of the ecto binding. That way the mailer gets disabled for Ecto projects by default but it is still possible to enable it using the `--mailer` command line flag. Fixes issue #5786.


josevalim:
Can you please look at the integration tests? thank you!

adonig:
I can fix the failing tests by explicitly enabling the mailer but I think it is a bad idea. It fixes the problem with issue #5768 but it might break for people who rely on the mailer just always being enabled by default. I believe this is a wont-fix.

PR #5792, opened 2024-04-22
Persist `remote_ip` during recycling of a conn

JonRowe:
Currently, if you set `remote_ip` during a test when building a conn, calling a `dispatch` call e.g `get` will run the conn through `recycle` which will lose this information. This PR changes that so it is maintained similar to `host`.

### DIFF
--- a/test/phoenix/test/conn_test.exs
+++ b/test/phoenix/test/conn_test.exs
@@ -180,6 +180,14 @@ defmodule Phoenix.Test.ConnTest do
       assert conn.host == "localhost"
     end
 
+    test "remove_ip is persisted" do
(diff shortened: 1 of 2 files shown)


test/phoenix/test/conn_test.exs
```diff
@@ -180,6 +180,14 @@ defmodule Phoenix.Test.ConnTest do
       assert conn.host == "localhost"
     end
 
+    test "remove_ip is persisted" do
```
nathanl, line 183:
```suggestion
    test "remote_ip is persisted" do
```

josevalim:

PR #5794, opened 2024-04-28
Update phx.gen.auth timestamps to match config

rhcarvalho:
Follows up on #5586 that introduced the concept but missed a few places that set timestamps.

### DIFF
--- a/priv/templates/phx.gen.auth/migration.ex
+++ b/priv/templates/phx.gen.auth/migration.ex
@@ -8,7 +8,7 @@ defmodule <%= inspect schema.repo %>.Migrations.Create<%= Macro.camelize(schema.
 <%= if schema.binary_id do %>      add :id, :binary_id, primary_key: true
 <% end %>      <%= migration.column_definitions[:email] %>
       add :hashed_password, :string, null: false
-      add :confirmed_at, :naive_datetime
+      add :confirmed_at, <%= schema.timestamp_type %>
       timestamps(<%= if schema.timestamp_type != :naive_datetime, do: "type: #{inspect schema.timestamp_type}" %>)
     end
 
@@ -20,7 +20,7 @@ defmodule <%= inspect schema.repo %>.Migrations.Create<%= Macro.camelize(schema.
       <%= migration.column_definitions[:token] %>
       add :context, :string, null: false
       add :sent_to, :string
-      timestamps(updated_at: false)
+      timestamps(<%= if schema.timestamp_type != :naive_datetime, do: "type: #{inspect schema.timestamp_type}, " %>updated_at: false)
     end
 
     create index(:<%= schema.table %>_tokens, [:<%= schema.singular %>_id])


rhcarvalho:
Need to test this better, perhaps the change of data type has unintented impact in other places.

priv/templates/phx.gen.auth/migration.ex
```diff
@@ -8,7 +8,8 @@ defmodule <%= inspect schema.repo %>.Migrations.Create<%= Macro.camelize(schema.
 <%= if schema.binary_id do %>      add :id, :binary_id, primary_key: true
 <% end %>      <%= migration.column_definitions[:email] %>
       add :hashed_password, :string, null: false
-      add :confirmed_at, :naive_datetime
+      add :confirmed_at, <%= inspect schema.timestamp_type %>
```
rhcarvalho, line 11: Would output one of `:naive_datetime`, `:utc_datetime` or `:utc_datetime_usec`.

I understand from the first note in https://hexdocs.pm/ecto/Ecto.Schema.html#module-primitive-types that those are valid values both in the schema definition and in the migration file.

> When using database migrations provided by "Ecto SQL", you can pass your Ecto type as the column type. However, note the same Ecto type may support multiple database types. For example, all of :varchar, :text, :bytea, etc. translate to Ecto's :string. Similarly, Ecto's :decimal can be used for :numeric and other database types. For more information, see [all migration types](https://hexdocs.pm/ecto_sql/Ecto.Migration.html#module-field-types).

priv/templates/phx.gen.auth/schema.ex
```diff
@@ -126,8 +126,11 @@ defmodule <%= inspect schema.module %> do
   Confirms the account by setting `confirmed_at`.
   """
   def confirm_changeset(<%= schema.singular %>) do
-    now = NaiveDateTime.utc_now() |> NaiveDateTime.truncate(:second)
-    change(<%= schema.singular %>, confirmed_at: now)
+    <%= case schema.timestamp_type do %>
+    <% :naive_datetime -> %>now = NaiveDateTime.utc_now() |> NaiveDateTime.truncate(:second)
+    <% :utc_datetime -> %>now = DateTime.utc_now() |> DateTime.truncate(:second)
+    <% :utc_datetime_usec -> %>now = DateTime.utc_now() |> DateTime.truncate(:microsecond)
+    <% end %>change(<%= schema.singular %>, confirmed_at: now)
```
rhcarvalho, line 133: Only handling the 3 documented values; formatted so as to not produce extraneous empty lines when rendered.

I noticed Elixir 1.15 supports the shorter `DateTime.utc_now(:second)`, but I assume that since `phx.new` generates a `mix.exs` requiring Elixir 1.14 then there's no reason to break compatibility right now.

rhcarvalho (commented):
Manually tested that generated app with these changes passes it's own generated tests. In particular, setting `:confirmed_at` using the correct type.

PR #5796, opened 2024-04-29
Re-order doc to better match example

jwbaldwin:
Just a small doc change to reduce confusion. It now matches the order of the list on line 34, which is the expected behavior.  👍


jwbaldwin:
Noticed by @tjdevries

josevalim:

PR #5802, opened 2024-05-06
[Fix] Using `--no-html` flag test config doesn't generate live_view config

### LINKED ISSUE #5790: Using `mix phx.new` with `--no-html` still configures live_view in test config
### Environment

* Elixir version (elixir -v): 
    Erlang/OTP 26 [erts-14.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]

    Elixir 1.16.0 (compiled with Erlang/OTP 26)
* Phoenix version (mix deps):
    {:phoenix, "~> 1.7.12"}
* Operating system:
    MacOS Sanoma 14.4.1
    

### Actual behavior

Running `mix test` fails because `config/test.exs` contains a configuration for `:phoenix_live_view` which isn't installed under the `--no-html` flag. The offending block is:

``` elixir
config :phoenix_live_view,
  # Enable helpful, but potentially expensive runtime checks
  enable_expensive_runtime_checks: true
```

### Expected behavior
Running `mix test` passes all pre-configured tests immediately after project generation.

pcapel:
Fixes #5790

### DIFF
--- a/installer/templates/phx_single/config/test.exs
+++ b/installer/templates/phx_single/config/test.exs
@@ -20,6 +20,8 @@ config :logger, level: :warning
 # Initialize plugs at runtime for faster test compilation
 config :phoenix, :plug_init_mode, :runtime
 
+<%= if @live do %>
 config :phoenix_live_view,
   # Enable helpful, but potentially expensive runtime checks
   enable_expensive_runtime_checks: true
+<% end %>
--- a/installer/test/phx_new_test.exs
+++ b/installer/test/phx_new_test.exs
@@ -474,6 +474,10 @@ defmodule Mix.Tasks.Phx.NewTest do
         assert file =~ ~s|# socket "/live"|
         refute file =~ ~s|plug Phoenix.LiveDashboard.RequestLogger|
       end)
+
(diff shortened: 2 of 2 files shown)


installer/templates/phx_single/config/test.exs
```diff
@@ -20,6 +20,8 @@ config :logger, level: :warning
 # Initialize plugs at runtime for faster test compilation
 config :phoenix, :plug_init_mode, :runtime
 
+<%= if @live do %>
 config :phoenix_live_view,
   # Enable helpful, but potentially expensive runtime checks
   enable_expensive_runtime_checks: true
+<% end %>
```
josevalim, line 27:
```suggestion
config :phoenix, :plug_init_mode, :runtime<%= if @live do %>

config :phoenix_live_view,
  # Enable helpful, but potentially expensive runtime checks
  enable_expensive_runtime_checks: true<% end %>
```

josevalim: We need to rewrite it like this to avoid empty lines. :)

PR #5803, opened 2024-05-06
Link module documentation of `leave` and `close` crashing the test process

LostKobrakai:
This at least makes this more discoverable. I still don't really understand why the API wouldn't handle this for the user though.


SteffenDE:
Thank you! 🙌🏻

PR #5804, opened 2024-05-07
[Fix] Fix phx.auth's context functions ordering

ayrat555:
I know that some people are against of alpha ordering, but it's a default check in `credo` so it would be nice to follow it here

Screenshot from 2024-05-07 17-21-20


chrismccord:
I would never include such a rule. Thanks!

SteffenDE:
@ayrat555 you can look at using styler to auto format many of the credo rules, including this one:
https://github.com/adobe/elixir-styler

PR #5805, opened 2024-05-08
Add HTTP verb to VerifiedRoutes

Gladear:
This PR is a follow-up of the proposal on the Elixir Forum: [Proposal for Phoenix: add HTTP verb to VerifiedRoutes](https://elixirforum.com/t/proposal-for-phoenix-add-http-verb-to-verifiedroutes/63319).

*tldr;* I propose to add a way for VerifiedRoutes to know if they’re used for get, post, etc to make them even more robust, and safer to refactor. This solution could be to add a “tag” to verified routes, like so: `~p"/user/#{user}"get`. More context and details can be found on the original post.

Please tell me what you think about this proposition. The code in the PR is functional, although I'd rather have more people re-read the test cases carefully to ensure the behaviour is the one expected. Also, I didn't update the documentation yet.
Here's the behaviour I expected to introduce (this is a copy-paste of the "add tests explaining behaviour" commit description):

> Add tests to VerifiedRoutes ensuring the correctness
of the introduced behaviour.
We want to ensure the following is respected:
> * runtime behaviour isn't altered at all: a ~p
  with http verb modifier emits the path as a
  string,
> * existing warning behaviour isn't altered:
  a ~p - with or without an http verb - not
  matching a route emits a warning,
> * (new) a ~p matching an existing route without
  a matching http verb emits a warning too,
> * (new) a ~p matching an existing route with
  a matching http verb does not emit a warning.

PS: I know the proposal wasn't validated or anything, but I figured this wouldn't take me too long to make a POC anyway, so I don't mind reworking the solution if this one isn't the one we ought to keep for any reason.

### DIFF
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -573,7 +573,7 @@ defmodule Phoenix.Router do
     verify_catch_all =
       quote generated: true do
         @doc false
-        def __verify_route__(_path_info) do
+        def __verify_route__(_path_info, _verb) do
           :error
         end
       end
@@ -635,6 +635,7 @@ defmodule Phoenix.Router do
 
   defp build_verify(path, routes_per_path) do
     routes = Map.get(routes_per_path, path)
+    verbs = Enum.map(routes, &Atom.to_string(&1.verb))
 
     forward_plug =
       Enum.find_value(routes, fn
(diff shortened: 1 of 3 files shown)


Gladear:
As proposed on the forum, I updated to PR to use a "prefix" syntax : `~p"get /posts/1"`.
The "flag syntax" is still in the commit history, but I'll remove the commits if the new syntax is kept

chrismccord:
Thanks for this, but I am not convinced it's what we want. There are many cases where use of a route is determined at runtime, such as a form's HTTP verb based on a new changest for record insert or existing changeset for a record update. Ultimately you should have a test hitting these verbs so I'm not convinced the value is worth complexity and extended usage. Thanks!

PR #5808, opened 2024-05-10
Update mix.exs

cvkmohan:
Yes, it is nitpicking. Changing a todo typo is ridiculous. But still...


chrismccord:
In this case it can stay as 0.0.0 as we just want to make sure it's there. Thanks!

PR #5812, opened 2024-05-16
CoreComponents: Add Levels for "headers"

krewast:
Using the "headers" function of the CoreComponents always creates h1 headers. That's a bit limiting.

This pull request introduces the new attribute "level" for the header function which is used to determine what heading level should be shown. There are officially 6 levels, the default is always h1 (https://developer.mozilla.org/en-US/docs/Web/HTML/Element/Heading_Elements).

Here is how it looks:

header-levels

And here is the code to check it out yourself:

```heex
<.header level="1" class="mt-[1em]">
  Header Level 1 (text-3xl)
</.header>

<.header level="2" class="mt-[1em]">
  Header Level 2 (text-2xl)
</.header>

<.header level="3" class="mt-[1em]">
  Header Level 3 (text-xl)
</.header>

<.header level="4" class="mt-[1em]">
  Header Level 4 (text-lg)
</.header>

<.header level="5" class="mt-[1em]">
  Header Level 5 (text-md)
</.header>

<.header level="6" class="mt-[1em]">
  Header Level 6 (text-sm)
</.header>
```

At the momemt, running `mix test` results in serveral warnings that all look like this:

> .warning: assign @level not available in EEx template. Please ensure all assigns are given as options.

It would be great if someone with more Elixir/Phoenix knowledge could have a look at this, because I don't really understand the issue. The attribute "level" has been defined, the type info and default value have been added. Accessing it via "@level" doesn't differ from accessing @class, for example. A pointer into the right direction would be helpful, I'm of course willing to add the necessary changes to fix these warnings.

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -417,8 +417,25 @@ defmodule <%= @web_namespace %>.CoreComponents do
   end
 
   @doc """
-  Renders a header with title.
+  Renders a header with a title, optional subtitle and actions.
+
+  ## Examples
+
+      <.header>
+        Title
+        <:actions>
+          Subtitle
+        </:actions>
+        <:actions>
+          ...
+        </:actions>
+      </.header>
+
+      <.header level="2" class="mt-10">
+        Title
+      </.header>
   """
+  attr :level, :string, default: "1"
   attr :class, :string, default: nil
 
   slot :inner_block, required: true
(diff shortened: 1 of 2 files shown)


krewast:
Alright, I get it, thanks for the feedback.

I just thought it would be cool to have as a default, because creating headings with different levels is such a standard thing to do. And that probably results in many reimplementations of this (or similar) logic where it could be solved once.

PR #5814, opened 2024-05-17
Remove invalid nested <button> inside <a> element in heex templates

### LINKED ISSUE #5770: HTML and Live CRUD generators output invalid HTML (buttons nested in links)
### Environment

### Actual behavior

Those are the affected files:

#### Double focus problem gif

#### HTML validation error

### Expected behavior

phoebe100:
Fixes #5770

### DIFF
--- a/priv/templates/phx.gen.html/index.html.heex
+++ b/priv/templates/phx.gen.html/index.html.heex
@@ -1,8 +1,8 @@
 <.header>
   Listing <%= schema.human_plural %>
   <:actions>
-    <.link href={~p"<%= schema.route_prefix %>/new"}>
-      <.button>New <%= schema.human_singular %></.button>
+    <.link href={~p"<%= schema.route_prefix %>/new"} class="py-2 px-3 text-sm font-semibold rounded-lg bg-zinc-100 hover:bg-zinc-200/80">
+      New <%= schema.human_singular %>
     </.link>
   </:actions>
 </.header>
--- a/priv/templates/phx.gen.html/show.html.heex
+++ b/priv/templates/phx.gen.html/show.html.heex
@@ -2,8 +2,8 @@
   <%= schema.human_singular %> <%%= @<%= schema.singular %>.id %>
   <:subtitle>This is a <%= schema.singular %> record from your database.</:subtitle>
   <:actions>
-    <.link href={~p"<%= schema.route_prefix %>/#{@<%= schema.singular %>}/edit"}>
-      <.button>Edit <%= schema.singular %></.button>
+    <.link href={~p"<%= schema.route_prefix %>/#{@<%= schema.singular %>}/edit"} class="py-2 px-3 text-sm font-semibold rounded-lg bg-zinc-100 hover:bg-zinc-200/80">
+      Edit <%= schema.singular %>
     </.link>
   </:actions>
 </.header>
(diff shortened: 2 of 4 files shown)


priv/templates/phx.gen.html/index.html.heex
```diff
@@ -1,8 +1,8 @@
 <.header>
   Listing <%= schema.human_plural %>
   <:actions>
-    <.link href={~p"<%= schema.route_prefix %>/new"}>
-      <.button>New <%= schema.human_singular %></.button>
+    <.link href={~p"<%= schema.route_prefix %>/new"} class="py-2 px-3 text-sm font-semibold rounded-lg bg-zinc-100 hover:bg-zinc-200/80">
```
phoebe100, line 4: I suggest this style since a very similar style is already used for the button-alike link "Get Started" in app.html.heex

josevalim: We should avoid using tailwind styles in generators. The home page is fine, because we discard it. Could we use <.button> only instead?

SteffenDE: In my projects I have a `<.button link={%{patch: ...}}` that then has a properly styled `<.link` inside. Maybe that's the way to go?

phoebe100: What about plain old link by using simply <.link> without any fancy style?
Buttons would always require javascript and some users might create their projects with --no-live flag.

Or the generators could produce different elements:
-  mix phx.gen.live could use <.button> with phx-click only
- while mix phx.gen.html could stick to simple <.link> (which could still be styled/replaced further if CoreComponents gets some button-alike link component in the future)

lifeiscontent:
👋🏻  new to Elixir & Phoenix, but the way the industry has been solving the (I need button styles for my link) problem is by using a prop called `asChild` in React. e.g. <Button asChild><a /></Button> where the `<Button>` would forward its props onto the `a`, is this something thats possible for Phoenix components? if yes, this might be a good solution to the problem.

SteffenDE:
Closed via https://github.com/phoenixframework/phoenix/commit/1f4be18bdd14f025669467ecaa9786ed90447e31. Thank you!

lifeiscontent:
@SteffenDE I've come up with an idea I think might be a good fit for phoenix defaults, in my own codebase I've created a `link_button` which is the exact same thing as `button` but just uses `link` under the hood instead. for all the styling I've abstracted the classes out into private helpers, if you think this might be a reasonable approach for phoenix defaults, I can open a PR.

SteffenDE:
@lifeiscontent This sounds similar to what I proposed initially in #5820 (I added a link attribute to the button https://github.com/phoenixframework/phoenix/commit/89c76dfa8004ce665d47b4fe2fea57857f3a000e, but another idea was a `button_link`). We ultimately decided against it because we didn't want to mirror the link component's API.

lifeiscontent:
@SteffenDE the current solution is still producing invalid HTML though, you can't nest `button -> a` or `a -> button` and doing either may cause some bizarre bugs

SteffenDE:
@lifeiscontent you‘re absolutely right and now I feel stupid 😃
I also somehow missed the part of your comment where you already mentioned the commit. I’m not sure why I thought only nesting the one way wasn’t valid, so back to square one…

josevalim:
@SteffenDE should we revert this? Or this is an improvement anyway?

SteffenDE:
No need to revert, it’s basically the same, but I plan to send a new PR

PR #5816, opened 2024-05-18
fix typo in verified routes docs

IgnacioGoldchluk:
I believe it should be `sigil_p/2` like in the docs for `url/1` instead of `sigil_p/1`, which does not exist.


SteffenDE:
Thank you! 🙌

PR #5823, opened 2024-05-24
Correct documentation about route helpers generation

L4vlet:
I was trying to skip helpers generation for some of my routes.

Contrary to what is stated in the documentation, you need to pass `as: nil` option instead of `as: false` to achieve this behaviour.


josevalim:

PR #5826, opened 2024-05-27
Add class attr to input component.

bratsche:
I'd like to propose adding a `class` attr for CoreComponents input component. As an example, I'm trying to make a form with a username field with an attached '@' span. I'd like to drop the round corners on the left side of the input by doing `<.input class="rounded-none rounded-r-lg" />` but it currently doesn't work and looks like this:

Screenshot 2024-05-26 at 11 55 55 PM

With my proposed PR I can pass in those classes and it looks like this:

Screenshot 2024-05-26 at 11 57 12 PM

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -269,6 +269,7 @@ defmodule <%= @web_namespace %>.CoreComponents do
       <.input name="my-input" errors={["oh no!"]} />
   """
   attr :id, :any, default: nil
+  attr :class, :string, default: nil
   attr :name, :any
   attr :label, :string, default: nil
   attr :value, :any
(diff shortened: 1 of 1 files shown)


SteffenDE:
Hi @bratsche, thank you for the contribution. The CoreComponents are built in the spirit that users are encouraged to freely modify them in their projects. So while the generated CoreComponents should have the basic functionality to get started, I am not sure if adding a `:class` by default is the right choice. Since the input component actually renders multiple HTML elements, someone else might want the class on the `<div>` that wraps the input instead, or another attribute `:wrapper_class`, etc.

As you've seen it's not a big change to do yourself, so I'm not sure if it's worth to include by default. That's just my opinion though. Let's see what others think.

bratsche:
Hi @SteffenDE, thanks for the response! Yeah, I know CoreComponents are intended to be customizable. But I feel like `class` is one of those things that should probably belong to almost all the components. It's already on some of them, for example `button`, `header`, `icon`. It seems like if nothing else then it seems like it's useful for adding margins and stuff like that without having to add additional div wrappers to pollute the markup.

If you guys disagree, no worries of course and we can close this. If you think other components should work similarly I'd be more than happy to either file another PR to fix the inconsistencies, or add another commit to this PR.

josevalim:
Agreed with @SteffenDE. It is very easy to add a bunch of small features, one by one, and then they become larger and larger. :) Thanks!

PR #5827, opened 2024-05-29
Remove 'antialiased' CSS class from root template

rhcarvalho:
One day, perhaps over a decade ago, it used to bring visual improvement and consistency to font rendering across browsers and operating systems.

But, over the years, the default subpixel-antialising in browsers/OSes produces better results than 'antialiased'.

There are many discussions about this on the web, I mention a few:
- https://github.com/google/fonts/issues/1170
- https://developer.mozilla.org/en-US/docs/Web/CSS/font-smooth
- https://web.dev/articles/antialiasing-101
- https://en.wikipedia.org/wiki/Subpixel_rendering

### DIFF
--- a/installer/templates/phx_static/home.css
+++ b/installer/templates/phx_static/home.css
@@ -940,11 +940,6 @@ select {
   color: rgb(63 63 70 / var(--tw-text-opacity));
 }
 
-.antialiased {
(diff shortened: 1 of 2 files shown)


SteffenDE (approved):

SteffenDE:
Thank you! 🙌

PR #5828, opened 2024-05-29
Update home.css for phx.new --no-tailwind

rhcarvalho:
Pick up template changes (including core components) and Tailwind upstream changes.

The majority of additions are due to picking up classes referenced in core_components.ex which were previously not part of the original home.css installer template.

Used the same Tailwind version as new Phoenix apps to re-generate the CSS file.

Steps to reproduce:

```
mix phx.new my_app
mix assets.build
cp priv/static/assets/app.css installer/templates/phx_static/home.css (retaining the note at the top of the file)
```

Steps to test the changes:

```
mix phx.new my_app2 --no-assets
mix phx.server
(verify that app renders as expected)
```

### DIFF
--- a/installer/templates/phx_static/home.css
+++ b/installer/templates/phx_static/home.css
@@ -1,7 +1,7 @@
 /* Default styling for the home page, this file can be deleted safely */
 
 /*
-! tailwindcss v3.2.4 | MIT License | https://tailwindcss.com
+! tailwindcss v3.4.4 | MIT License | https://tailwindcss.com
 */
 
 /*
@@ -33,9 +33,12 @@
 3. Use a more readable tab size.
 4. Use the user's configured `sans` font-family by default.
 5. Use the user's configured `sans` font-feature-settings by default.
+6. Use the user's configured `sans` font-variation-settings by default.
+7. Disable tap highlights on iOS
 */
 
-html {
+html,
+:host {
   line-height: 1.5;
   /* 1 */
   -webkit-text-size-adjust: 100%;
@@ -45,10 +48,14 @@ html {
   -o-tab-size: 4;
      tab-size: 4;
   /* 3 */
-  font-family: ui-sans-serif, system-ui, -apple-system, BlinkMacSystemFont, "Segoe UI", Roboto, "Helvetica Neue", Arial, "Noto Sans", sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
+  font-family: ui-sans-serif, system-ui, sans-serif, "Apple Color Emoji", "Segoe UI Emoji", "Segoe UI Symbol", "Noto Color Emoji";
   /* 4 */
   font-feature-settings: normal;
   /* 5 */
+  font-variation-settings: normal;
+  /* 6 */
+  -webkit-tap-highlight-color: transparent;
+  /* 7 */
 }
 
 /*
@@ -120,8 +127,10 @@ strong {
 }
 
 /*
-1. Use the user's configured `mono` font family by default.
-2. Correct the odd `em` font sizing in all browsers.
+1. Use the user's configured `mono` font-family by default.
+2. Use the user's configured `mono` font-feature-settings by default.
+3. Use the user's configured `mono` font-variation-settings by default.
+4. Correct the odd `em` font sizing in all browsers.
 */
 
 code,
@@ -130,8 +139,12 @@ samp,
 pre {
   font-family: ui-monospace, SFMono-Regular, Menlo, Monaco, Consolas, "Liberation Mono", "Courier New", monospace;
   /* 1 */
-  font-size: 1em;
+  font-feature-settings: normal;
   /* 2 */
+  font-variation-settings: normal;
+  /* 3 */
+  font-size: 1em;
+  /* 4 */
 }
 
 /*
@@ -190,11 +203,17 @@ select,
 textarea {
   font-family: inherit;
   /* 1 */
+  font-feature-settings: inherit;
+  /* 1 */
+  font-variation-settings: inherit;
+  /* 1 */
   font-size: 100%;
   /* 1 */
   font-weight: inherit;
   /* 1 */
   line-height: inherit;
+  /* 1 */
+  letter-spacing: inherit;
   /* 1 */
   color: inherit;
   /* 1 */
@@ -219,9 +238,9 @@ select {
 */
 
 button,
-[type='button'],
-[type='reset'],
-[type='submit'] {
+input:where([type='button']),
+input:where([type='reset']),
+input:where([type='submit']) {
   -webkit-appearance: button;
   /* 1 */
   background-color: transparent;
@@ -337,6 +356,14 @@ ul,
 menu {
   list-style: none;
   margin: 0;
+  padding: 0;
+}
+
+/*
+Reset default styling for dialogs.
+*/
+
+dialog {
   padding: 0;
 }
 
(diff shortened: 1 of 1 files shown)


installer/templates/phx_static/home.css
```diff
@@ -597,14 +624,48 @@ select {
   --tw-skew-y: 0;
   --tw-scale-x: 1;
   --tw-scale-y: 1;
+  --tw-pan-x:  ;
```
rhcarvalho, line 627: See https://github.com/phoenixframework/phoenix/issues/5546#issuecomment-2136931546

rhcarvalho:
(I'm on mobile)

Hi! From memory, the changes comprise:

1. Bringing back deleted but valid parts of Tailwind.
2. Changes between the Tailwind versions, like different preflight styles.
3. I think "container" appears in a comment in core_components.ex but is not used as a class, and that pulls some dependencies (this one I remember vividly because I was not super happy introducing so many lines).
4. Existing CSS could to be missing classes (not sure)

I documented in the commit message (IIRC) the steps I did, and indeed it is not only the home.html.ex we have to look at.

I thought not making manual exceptions and edits would make maintenance easier, perhaps eventually automating generating this file.

Okay from my side if you prefer a different direction! Thanks for having a look and all the work you've been doing 💜

rhcarvalho:
@SteffenDE I'm taking another look at this.

Your suggestion of using the Tailwind Playground was nice, thanks! I don't recall using that tool and it's good to know it's there.

But the approach of pasting code in the playground missed a few things, even when I pasted the full static HTML to render the home page of a new app:

- Custom tailwind config which is hard to reproduce on the playground
  - In particular, `hero-*` icons are missing (there are 6 `hero-*` occurrences in the HTML)
  - Misses `phx-*` variants like `phx-submit-loading`
  - Misses everything from `@tailwindcss/forms` like CSS for `button`
- Dynamic classes coming in from core components like classes used in `show/2` and `hide/2`
  Even after copying the full static HTML source into the playground, Tailwind cannot detect some class names because of how they end up in the HTML, example:

  ```html
  phx-connected="[[&quot;hide&quot;,{&quot;time&quot;:200,&quot;to&quot;:&quot;#client-error&quot;,&quot;transition&quot;:[[&quot;transition-all&quot;,&quot;transform&quot;,&quot;ease-in&quot;,&quot;duration-200&quot;],[&quot;opacity-100&quot;,&quot;translate-y-0&quot;,&quot;sm:scale-100&quot;],[&quot;opacity-0&quot;,&quot;translate-y-4&quot;,&quot;sm:translate-y-0&quot;,&quot;sm:scale-95&quot;]]}]]"
  ```
  
  The `transition-all` etc are all missing from the output CSS.

rhcarvalho:
Rebased on top of latest main to fix conflict (caused by removing the `antialiased` class in #5827) and updated Tailwind version to 3.4.4 (as in #5858; note the Tailwind version caused no change to the generated CSS).

<details>
<summary>diff</summary>

```diff
diff --git installer/templates/phx_static/home.css installer/templates/phx_static/home.css
index 363f2522..137c999a 100644
--- installer/templates/phx_static/home.css
+++ installer/templates/phx_static/home.css
@@ -1,7 +1,7 @@
 /* Default styling for the home page, this file can be deleted safely */
 
 /*
-! tailwindcss v3.4.3 | MIT License | https://tailwindcss.com
+! tailwindcss v3.4.4 | MIT License | https://tailwindcss.com
 */
 
 /*
@@ -1526,11 +1526,6 @@ select {
   color: rgb(24 24 27 / var(--tw-text-opacity));
 }
 
-.antialiased {
-  -webkit-font-smoothing: antialiased;
-  -moz-osx-font-smoothing: grayscale;
-}
-
 .opacity-0 {
   opacity: 0;
 }
```
</details>

installer/templates/phx_static/home.css
```diff
@@ -617,14 +678,159 @@ select {
...
+  .container {
+    max-width: 1280px;
+  }
+}
+
+@media (min-width: 1536px) {
+  .container {
+    max-width: 1536px;
+  }
+}
```
rhcarvalho, line 757: These are notably accidental, as the word "container" appears in `attr` docs in `core_components.ex`.

Considering how Tailwind works with heuristics to eliminate unused classes, I'd rather follow an easy to reproduce procedure that produces all necessary classes, than cherry pick and risk missing classes.

installer/templates/phx_static/home.css
```diff
@@ -617,14 +678,159 @@ select {
...
+  -webkit-mask: var(--hero-arrow-left-solid);
+  mask: var(--hero-arrow-left-solid);
+  -webkit-mask-repeat: no-repeat;
+          mask-repeat: no-repeat;
+  background-color: currentColor;
+  vertical-align: middle;
+  display: inline-block;
+  width: 1.5rem;
+  height: 1.5rem;
+}
```
rhcarvalho, line 770: Heroicons were all originally missing and are not picked up by the Tailwind Playground because we embed them using a custom plugin.

Some icons do get used in the static HTML for the home page:

image

installer/templates/phx_static/home.css
```diff
@@ -617,14 +678,159 @@ select {
...
+  position: absolute;
+  width: 1px;
+  height: 1px;
+  padding: 0;
+  margin: -1px;
+  overflow: hidden;
+  clip: rect(0, 0, 0, 0);
+  white-space: nowrap;
+  border-width: 0;
 }
```
rhcarvalho, line 834: Another "accidental" addition that is part of `core_components.ex` but not used in `home.html.heex` in particular. This one is used in the `table` component.

installer/templates/phx_static/home.css
```diff
@@ -644,61 +850,160 @@ select {
...
-  right: 0px;
-  bottom: 0px;
-  left: 0px;
+  inset: 0px;
+}
+
+.-inset-y-px {
+  top: -1px;
+  bottom: -1px;
 }
```
rhcarvalho, line 859: Another one used in the `table` component, not in home... and so will be the case for some other classes.

There's a decision whether it's more important to have all necessary classes and perhaps a few false positives (Tailwind's approach), or trying to have a minimal CSS file and perhaps missing a few classes and having to keep the diligence when making future updates.

Considering we've missed classes for a long time and nobody complained, I think either way doesn't really matter. This CSS file is only used for projects without the assets pipeline, and the very next thing after creating the app is changing the HTML template and CSS anyway -- it possibly doesn't matter that icons are missing or small things are slightly off.

josevalim (approved):

SteffenDE:
@rhcarvalho I looked at this again, and I do think that we don't need any styles from core components at all for the home.css. It is used for the home page when `--no-tailwind` is used and while the flash component is included on that page, I don't think it should be styled there, as it also won't be styled on any other page. You can see the diff here: https://github.com/phoenixframework/phoenix/pull/6099 and looking at it, I don't think there's a reason to update at all?

SteffenDE:
> When you say that the flash component should not be styled, then maybe the best code is no code -- generate no CSS for the --no-tailwind case?

Well I guess what I'm trying to say is: there's not a real need to update the css at all, since the page is self-contained and we don't use any components in it.

> It is not clear to me how that diff was produced, so it's hard to compare.

I cloned phoenix, created a fresh project with phx.new --dev and then deleted the core components (as we don't need them), then regenerated styles with `mix tailwind`.

> Open to help whichever way you think is best -- either by updating this or closing it in favor of other changes or no changes.

At the moment I think we don't need any changes. We can revisit if for whatever reason the page starts rendering wrong in a future browser release.

Thank you very much for contributing!

PR #5842, opened 2024-06-15
Add PHX_HOST variable guideline

bohdan-shulha:
As the experiments shown, one needs to specify PHX_HOST variable to serve the web requests in production.

### DIFF
--- a/guides/deployment/deployment.md
+++ b/guides/deployment/deployment.md
@@ -65,17 +65,17 @@ The error message is quite clear: it says Phoenix could not find a static manife
 
 ## Starting your server in production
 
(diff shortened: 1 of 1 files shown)


SteffenDE:
It might be worth mentioning that the generated code assumes that you're going to use a reverse proxy to forward traffic to the `PORT`, as we're also hardcoding the scheme to https and port to 443 for generating URLs by default:

https://github.com/phoenixframework/phoenix/blob/be4eb9aea66078f647d33508f2bb5e1a7fa395df/installer/templates/phx_single/config/runtime.exs#L36-L42

And guide the user towards the `runtime.exs` config for customization.

bohdan-shulha:
Nice point, thanks. I'll write more text then. :)

PR #5623, opened 2023-11-02
Removing tmp files

mustela:
While working on my own generators and utilizing the mix_helper module, I noticed that tmp files were not being removed; only the contents within them were. This makes things difficult to track, especially when trying to see what files are being generated, due to the numerous random directories being created and not being cleaned between runs. Here you can see the output of one run:
 

```bash
❯ cd tmp
❯ ls
+y5E0V98wE/                FWssDkXKgo/                TK5FAP/                    dwvhd9RNA3/                olnF+LTFg4/
0r5AJLwI1d/                G2qEVW9bWD/                TOvg5FCelf/                fb/                        pRHHd1rZnn/
1BLZ5enP/                  GB4Xr5gQ/                  TXkqg956we/                fjYfzH3Axb/                pTW/
1CBqNafDs6/                GSADpO41Ar/                U2EuGvKAfN/                fucoCI3jK9/                pV6UDHRiGK/
1I6ijIazqW/                GoKf4z6CxU/                U82nwLCeYf/                g3jj18159L/                phoenix_digest/
1cRAdX7Jfs/                HJkc8RGytH/                VG5x4cvZE4/                hFnIKQA/                   phoenix_digest_twice/
3YxILcqWLW/                K1fIB4mlFA/                VesruofdWF/                hG6L2SWwd/                 pkvCH7OOu2/
4HTyA0hs83/                K4nQQkkRKQ/                VrsJyjvJqY/                i3P/                       ppe9nEWgXu/
5Xn0ALGBqH/                KTFJdIf98F/                VsAwbufkny/                iSs3TmqyFz/                r8snnQ9ZpI/
6b0SfbYf5L/                KVdyRu0JuU/                Vt0AKtBHOT/                iautlemqiI/                rQ5z8q07y1/
7Ha702VA8u/                KsUo9kYFmh/                VvAiiRfqaN/                izef+BTlWg/                rhXDbMwdeU/
8+Vig1ZSdF/                L2s9mXdAse/                X1Xah5dZgQ/                jAMYpwQiO5/                rpfVnq6kzs/
8NDW2D9VpD/                LAdHqo/                    XobGK0qI6S/                jXqVyHRu+4/                s6zVju6X3Z/
9BxY6TLfIs/                LHjoUroWS6/                XwdMGp6mdP/                jbIjgAqcFP/                t6b6NQRnNk/
9KRv7zRYPS/                LQfDEg+JQf/                Y+RrrCp0uD/                jgkA0A5Bib/                test removes old versions/
[...]

PR #5843, opened 2024-06-16
make tests repeatable (--repeat-until-failure)

SteffenDE:
This allows to use the new `mix test --repeat-until-failure` feature to run the unit tests.

I also found some more issues where mix compile was called in tests leading to missing modules if you had bad luck. This is now fixed by always passing `--no-compile`.

Furthermore, all directories in the `tmp` folder are now properly cleaned up. https://github.com/phoenixframework/phoenix/pull/5623 already tried to do this, but this still left directories if the generated test name or random string contained a `/` character.

### DIFF
--- a/installer/test/mix_helper.exs
+++ b/installer/test/mix_helper.exs
@@ -11,24 +11,26 @@ defmodule MixHelper do
   end
 
   defp random_string(len) do
-    len |> :crypto.strong_rand_bytes() |> Base.encode64() |> binary_part(0, len)
+    len |> :crypto.strong_rand_bytes() |> Base.url_encode64() |> binary_part(0, len)
   end
 
   def in_tmp(which, function) do
-    path = Path.join([tmp_path(), random_string(10) <> to_string(which)])
+    base = Path.join([tmp_path(), random_string(10)])
+    path = Path.join([base, to_string(which)])
 
     try do
       File.rm_rf!(path)
       File.mkdir_p!(path)
       File.cd!(path, function)
     after
-      File.rm_rf!(path)
+      File.rm_rf!(base)
     end
   end
 
   def in_tmp_project(which, function) do
     conf_before = Application.get_env(:phoenix, :generators) || []
-    path = Path.join([tmp_path(), random_string(10) <> to_string(which)])
+    base = Path.join([tmp_path(), random_string(10)])
+    path = Path.join([base, to_string(which)])
 
     try do
       File.rm_rf!(path)
@@ -47,14 +49,15 @@ defmodule MixHelper do
         function.()
       end)
     after
-      File.rm_rf!(path)
+      File.rm_rf!(base)
       Application.put_env(:phoenix, :generators, conf_before)
     end
   end
 
   def in_tmp_umbrella_project(which, function) do
     conf_before = Application.get_env(:phoenix, :generators) || []
-    path = Path.join([tmp_path(), random_string(10) <> to_string(which)])
+    base = Path.join([tmp_path(), random_string(10)])
+    path = Path.join([base, to_string(which)])
 
     try do
       apps_path = Path.join(path, "apps")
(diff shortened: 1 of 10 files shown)


installer/test/mix_helper.exs
```diff
@@ -81,7 +84,10 @@ defmodule MixHelper do
 
     try do
       capture_io(:stderr, fn ->
-        Mix.Project.in_project(app, path, [prune_code_paths: false], fun)
+        Mix.Project.in_project(app, path, [prune_code_paths: false], fn mod ->
+          fun.(mod)
+          Mix.Project.clear_deps_cache()
```
SteffenDE, line 89: this was a tricky one 😃

installer/test/mix_helper.exs
```diff
@@ -11,24 +11,26 @@ defmodule MixHelper do
   end
 
   defp random_string(len) do
-    len |> :crypto.strong_rand_bytes() |> Base.encode64() |> binary_part(0, len)
+    len |> :crypto.strong_rand_bytes() |> Base.url_encode64() |> binary_part(0, len)
   end
 
   def in_tmp(which, function) do
```
josevalim, line 17: ExUnit now has a feature for creating tmp paths for tests, but this is fine for now. :)

test/phoenix/digester_test.exs
```diff
@@ -172,38 +179,46 @@ defmodule Phoenix.DigesterTest do
...
+
+        assert :ok = Phoenix.Digester.compile(input_path, input_path, true)
+        assert :ok = Phoenix.Digester.compile(input_path, input_path, true)
+
+        output_files = assets_files(input_path)
+        refute "file.js.gz.gz" in output_files
+        refute "cache_manifest.json.gz" in output_files
+        refute Enum.any?(output_files, &(&1 =~ ~r/file-#{@hash_regex}.[\w|\d]*.[-#{@hash_regex}/))
+      after
+        File.rm_rf!(input_path)
```
josevalim, line 220: Do we need to clean up before and after? Only before should be enough, no? It doesn’t hurt to do both, so this is fine, but maybe doing less would work too!

test/phoenix/verified_routes_test.exs
```diff
@@ -553,6 +584,9 @@ defmodule Phoenix.VerifiedRoutesTest do
           end)
 
         assert warnings == ""
+      after
+        :code.purge(__MODULE__.VerifyFalseTrueMatchesFirst)
+        :code.delete(__MODULE__.VerifyFalseTrueMatchesFirst)
```
josevalim, line 589: You could also use tags for making this cleaner.

    @tag cleanup: SomeMod

and then on setup you have on_exit that purges it if the cleanup is set.

josevalim (approved):
Minor bits to improve, but your call! Ship it!!!

test/phoenix/verified_routes_test.exs
SteffenDE: Is there a way to get a list of modules defined by the current process? Pseudo-code

```elixir
setup do
  on_exit(fn -> Enum.each(Process.defined_modules(), fn mod ->
    :code.purge(mod)
    :code.delete(mod)
  end)
end
```

That would also help in some Elixir tests where we found some missing purges recently

installer/test/mix_helper.exs
SteffenDE: I actually wanted to use it, but ExUnit doesn't cleanup the tmp_dir afterwards, maybe that could be an (optional) improvement for ExUnit as well?

josevalim: We don’t clean up on purpose, to help debugging failed tests, but we should always clean up before the new run!

test/phoenix/verified_routes_test.exs
josevalim: There isn’t, unfortunately

PR #5844, opened 2024-06-16
Make the default :info flash display "Info" instead, and add a :success flash with the old behaviour

### LINKED ISSUE #5795: Default :info flash is not an Info flash, it is a Success flash
tl;dr `:info` flashes are not really `:info`, they're `:success`. 

When I `put_flash(conn, :info, message)`, I don't expect the flash will have green success classes and a "Success!" header.

## Example
This code:

Screenshot 2024-04-28 at 3 59 14 PM

Results in this flash:

Screenshot 2024-04-28 at 4 01 34 PM

Whether we want to say the `Reset Password` operation not erroring out is a success is a matter of professional opinion, but from a pure code perspective, my expectation is `put_flash(:info, message)` would pop a generic informational flash to the user and `:success` would pop the "Success!" toast.

## Potential improvement
In my own projects I've wired `@kind == :success` to the "Success!" header and `:info` to a more vanilla "Info" flash styling/header in the `flash/1` core component. 

I can attempt to open a PR to add this to the core components and to the generators if desirable. I realize this is a departure from how these flashes have been by default for some time, but IMHO it'll reduce confusion.

DitherWither:
Fixes #5795

Also use blue for the info flash, as that is what info dialogues tend to use.

Image of the info flash

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -100,7 +100,7 @@ defmodule <%= @web_namespace %>.CoreComponents do
   attr :id, :string, doc: "the optional id of flash container"
   attr :flash, :map, default: %{}, doc: "the map of flash messages to display"
   attr :title, :string, default: nil
-  attr :kind, :atom, values: [:info, :error], doc: "used for styling and flash lookup"
+  attr :kind, :atom, values: [:success, :info, :error], doc: "used for styling and flash lookup"
   attr :rest, :global, doc: "the arbitrary HTML attributes to add to the flash container"
 
   slot :inner_block, doc: "the optional inner block that renders the flash message"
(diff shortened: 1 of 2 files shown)


josevalim:
Thank you for the PR. However, I think we should keep the generated code simple and not add a new style. We should either rename :info to :success or change the style of `:info` to something slightly more neutral (or keep it all as is).

mattmatters:
We've done something similar on a couple projects. I think the defaults are fine as they are, but some of the documentation doesn't actually make it very clear that flash types can be whatever (at least in the liveview documentation). Maybe we start with just adding a little nice clarification [here](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#put_flash/3) for now.

SteffenDE:
@mattmatters the docs were already adjusted in response to https://github.com/phoenixframework/phoenix_live_view/issues/3342, but that’s not released yet :)

SteffenDE:
We decided to change the color to blue and remove the title by default https://github.com/phoenixframework/phoenix/pull/6065.

PR #5847, opened 2024-06-18
Update error component spacing

osbre:
IMO, there's too much of a gap between the icon and text.

image


josevalim (approved):

SteffenDE:
I assume the intent of the big spacing is to have the same gap between the top and right of the icon. So yeah while I agree that it is probably a bit much, I personally prefer the consistency 😃

osbre:
@SteffenDE Understood, then how about a consistent 1.5? Here's a preview:

image

SteffenDE:
LGTM! Thank you for including the screenshot :)

PR #5851, opened 2024-06-20
Use Mix.Phoenix.generator_paths in phx.gen.auth.ex

Zurga:
Replace the local private function with the public one in Mix.Phoenix used by all the other generators.


SteffenDE:
Thank you! 🙌🏻

PR #5856, opened 2024-06-25
Pass all_args to compile task in digest tasks

greg-freewave:
Args like "--no-deps-check" do not get passed through causing `compile` to fail in instances where it's important to not check deps. For example this broke our nix build which handles deps in its own way.

It breaks these instructions: https://github.com/NixOS/nixpkgs/blob/master/doc/languages-frameworks/beam.section.md?plain=1#L215

Also adds command line docs and removes `--no-deps-check` flag on tests that use `--no-compile`


josevalim:

PR #5857, opened 2024-06-28
Remove noop Tailwind class 'transform'

rhcarvalho:
Not needed since Tailwind v2.2, see:
- https://github.com/tailwindlabs/tailwindcss/blob/3.4/CHANGELOG.md#220---2021-06-17


rhcarvalho:
Nice catch from integration tests wrt `mix format` on the generate code from templates!

Gazler:
Great spot. Thanks <3

PR #5858, opened 2024-06-28
Upgrade Tailwind to 3.4.4

rhcarvalho:


rhcarvalho:
I have a feeling the test failure is a flake, failed to listen on a used port -- seems unrelated to the change of Tailwind version.

josevalim:
I'd say it is fine. We don't need to update on every patch version to avoid churn. But we should update if there is a new major or minor. Thanks!

PR #5861, opened 2024-07-06
Fix live_reload patterns

azizk:
Match the dot of the file extension and at least one basename character.


josevalim:
Do we really need to match on `.`? Even if we have false positives, there isn't a negative consequence. I'd rather keep the patterns simple.

josevalim:
No, it doesn't recompile.

josevalim:
Closing per the above.

PR #5863, opened 2024-07-08
Set default generator timestamp to be :utc_datetime_usec

crertel:
Exactly what it says on the tin. Every project I've ever done this is one of the first things that needs doing. I wager it's fairly standard practice at this point, so I wanted to open a PR to see if folks would be okay switching.


josevalim:
My experience with Ecto is that the microsecond precision support across databases is spotty and not consistent. IIRC, SQlite3 does not support microseconds in its lightweight representation (integers) only in the text version. So in order to provide a default and consistent experience to everyone starting out, the usec migration should be opt-in. Thanks.

PR #5864, opened 2024-07-09
templates: use inspect/1 instead of to_string/1 in errors_on/1

v0idpwn:
Per https://github.com/elixir-ecto/ecto/issues/4445

I don't see a reason to prefer `to_string` over `inspect` here, maybe I'm missing something :)


viniciussbs:
So, I've tried `inspect/1` in a custom `errors_on/1` implementation in my code base. I gave up because decimal values get weird representation when interpolated using `inspect/1`:

```
iex(1)> to_string(Decimal.new(0))
"0"
iex(2)> inspect(Decimal.new(0))
"Decimal.new(\"0\")"
```

v0idpwn:
Considering this function is built for testing, one could would argue that `Decimal.new("0")` is a more precise depiction of the data for the error message than `0`.

PR #5865, opened 2024-07-10
phx.gen.release: Add explicit mix assets.setup step

### LINKED ISSUE #5833: In the default Dockerfile, `mix assets.deploy` downloads tailwind & esbuild every time assets change
Thanks for making Phoenix! I noticed that in the default Dockerfile `mix assets.deploy` downloads tailwind & esbuild every time assets change:

https://github.com/phoenixframework/phoenix/blob/69685f783bc60d2a124fc2fb48028b065bc24b22/priv/templates/phx.gen.release/Dockerfile.eex#L46-L55

Do you think the default could be changed to run

```
mix tailwind.install
mix esbuild.install
```

... after `mix deps.compile` and before `mix assets.deploy`? This should make most Docker image builds faster because they won't have to download tailwind/esbuild every time anything in priv, lib, or assets changes.

I'd be happy to open a small PR to do this if it's alright with you!

The only challenge is that it'd have to handle `--no-tailwind` and `--no-esbuild`. Happy to take a stab at handling that too with some pointers; a cursory search for `no-tailwind` didn't turn up anything: https://github.com/search?q=repo%3Aphoenixframework%2Fphoenix+no-tailwind&type=code

rhcarvalho:
This step will install Tailwind and esbuild earlier and the container layer shall be cached depending only on dependencies and config (where Tailwind and esbuild tool versions are actually defined).

Code changes in priv, lib and assets shall not cause those tools to be re-downloaded and installed.

Closes #5833.

### DIFF
--- a/priv/templates/phx.gen.release/Dockerfile.eex
+++ b/priv/templates/phx.gen.release/Dockerfile.eex
@@ -44,7 +44,9 @@ RUN mkdir config
 # to be re-compiled.
 COPY config/config.exs config/${MIX_ENV}.exs config/
 RUN mix deps.compile
-
+<%= if assets_dir_exists? do %>
+RUN mix assets.setup
+<% end %>
 COPY priv priv
 
 COPY lib lib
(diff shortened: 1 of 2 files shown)


rhcarvalho:
[Tests failed fetching from Docker Hub](https://github.com/phoenixframework/phoenix/actions/runs/9874184269/job/27268048838?pr=5865); I think they're having a bad time as I had `mix phx.gen.release --docker` fail a couple of times locally, needed a couple retries.

Someone would need to retry the job.

josevalim (approved):

josevalim:

PR #5866, opened 2024-07-11
update debian to bookworm

SteffenDE:


josevalim (approved):
LGTM! However, keep in mind that OTP 24 is no longer officially supported, so we could push for bookworm only and update CI if you want to.

josevalim (approved):

Issue #5867, opened 2024-07-12
Mix phx.server with -- halts immediately


jerko-culina:
### Environment

* Elixir version (elixir -v): 1.17.2
* Phoenix version (mix deps): 1.7.14
* Operating system: MacOS

### Actual behavior

```elixir
$ mix phx.server 
```
Starts server normally.

```elixir
$ mix phx.server -- 
```
```elixir
$ mix phx.server -- arg1 arg2
```
Starts server and exists (halts) immediately.

### Expected behaviour

```elixir
$ mix phx.server
```
```elixir
$ mix phx.server -- 
```
```elixir
$ mix phx.server -- arg1 arg2
```
All three commands should start server normally without halt.

### Potential root of issue
As per [OptionParser documentation](https://github.com/elixir-lang/elixir/blob/main/lib/elixir/lib/option_parser.ex#L246), the separator `--` implies options should no longer be processed (and as far as i see they are not).

Problem is in `phx.server` command that adds `--no-halt` flag after separator `--` so `--no-halt` flag has been ignored by `OptionParser`.

Here is copy paste of `phx.server` implementation of `run` command [link to command](https://github.com/phoenixframework/phoenix/blob/main/lib/mix/tasks/phx.server.ex#L34-L37)

```elixir
  def run(args) do
    Application.put_env(:phoenix, :serve_endpoints, true, persistent: true)
    Mix.Tasks.Run.run(open_args(args) ++ run_args())
  end
```
if I run next command:
```
$ mix phx.server -- arg
```
arguments passed to `Mix.Tasks.Run.run`  will look like 
```elixir
["--", "arg", "--no-halt"]
```
and `OptionParser` will ignore everything after `--` including `--no-halt`

### Potential solution
Just replace order of args:

```diff
  def run(args) do
    Application.put_env(:phoenix, :serve_endpoints, true, persistent: true)
-    Mix.Tasks.Run.run(open_args(args) ++ run_args())
+    Mix.Tasks.Run.run(run_args() ++ open_args(args))
  end
```

So, am I missing something or not?

Gazler:
Changing the order works. If you could open a PR, that would be most appreciated.

jerko-culina:
Link to PR https://github.com/phoenixframework/phoenix/pull/5868

PR #5868, opened 2024-07-12
Fix mix phx.server command bug - it halts if command line args separator -- has been used

jerko-culina:
When `mix phx.server -- arg1` command has been run, it starts server, then halts it immediately.

This PR fixes that bug, so once you run command it does not halt server.

More details on [issue tracker](https://github.com/phoenixframework/phoenix/issues/5867#issue-2406001001)


Gazler:
Thanks <3

PR #5876, opened 2024-07-22
Fix linting on README templates

georgeguimaraes:
Markdown linters hate indented bullet points.

This PR fixes the README that Phoenix generates when running `mix phx.new`


SteffenDE:
Thank you! 🙌

PR #5878, opened 2024-07-24
Update Heroku deployment documentation

kimihito:
## Problem

The current Heroku deployment instructions are incompatible with newer versions of Node.js, causing deployment failures and confusion for developers using up-to-date environments. See https://github.com/gjaldon/heroku-buildpack-phoenix-static/issues/112

## Changes

We have replaced the current buildpack with https://github.com/gigalixir/gigalixir-buildpack-phoenix-static, which supports installation of current Node.js versions. This change allows us to use more recent Node.js versions in our Heroku deployments, resolving the compatibility issues we were facing.

### DIFF
--- a/guides/deployment/heroku.md
+++ b/guides/deployment/heroku.md
@@ -116,13 +116,13 @@ web: mix phx.server
 
 ### Optional: Node, npm, and the Phoenix Static buildpack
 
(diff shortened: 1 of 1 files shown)


Gazler:
Thanks <3

PR #5879, opened 2024-07-25
docs: minor grammar etc improvements.

josh-nz:
As I read through the docs, there are some minor improvements that can be made to help improve readability. These are not documentation bugs.

Most of these changes fall into categories like plurality correction, correct tense usage within a paragraph, grammar adjustments, etc.

I've deliberately tried not to rewrite or reword things except where I feel they read better, and in most cases you'll find these changes to be very minor. I'm cautious to avoid changing any meaning.

This is my first contribution, and while it's low hanging fruit, I appreciate the focus the community has on good documentation, so helping to make this even better is a great place to start.

I have only read a few guides so there is not a lot to this commit but I wanted to submit early in part as a check to ensure that the types of changes I'm making are wanted and considered improvements, before I spend too much time on it. I'm also currently in the LiveView docs (and making adjusts there as I go), and are unsure when I'll return to the Phoenix docs.

I know that there are many different styles to writing (which is partially why this is not an attempt to rewrite things), and I'm not an editor, technical writer, or otherwise of note for this sort of activity. Simply a dev who read some guides and thought, I can make this just that little bit better for the community.

If there is anything in particular you'd like me to undo, just let me know. And if this contribution is beneficial, I will continue to contribute as I continue to read the docs.

### DIFF
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -23,7 +23,7 @@ making it easier for developers to [contribute to Phoenix](#pull-requests).
 ## Bug reports
 
 A bug is either a _demonstrable problem_ that is caused by the code in the repository,
-or indicate missing, unclear, or misleading documentation. Good bug reports are extremely
+or indicates missing, unclear, or misleading documentation. Good bug reports are extremely
 helpful - thank you!
 
 Guidelines for bug reports:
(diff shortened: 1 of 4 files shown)


Gazler:
Thanks <3

PR #5880, opened 2024-07-26
Fix typos and improve language

preciz:
Thank you for reviewing

### DIFF
--- a/lib/phoenix/code_reloader/server.ex
+++ b/lib/phoenix/code_reloader/server.ex
@@ -48,7 +48,7 @@ defmodule Phoenix.CodeReloader.Server do
             Mix.Project.build_structure()
           else
             Logger.warning(
-              "Phoenix is unable to create symlinks. Phoenix' code reloader will run " <>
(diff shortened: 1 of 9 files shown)


lib/phoenix/code_reloader/server.ex
```diff
@@ -48,7 +48,7 @@ defmodule Phoenix.CodeReloader.Server do
             Mix.Project.build_structure()
           else
             Logger.warning(
-              "Phoenix is unable to create symlinks. Phoenix' code reloader will run " <>
+              "Phoenix is unable to create symlinks. Phoenix's code reloader will run " <>
```
SteffenDE, line 51: I'm not a native speaker, but do you really add an `s` if the word already ends in an s-like sound, which Phoenix does? `ˈfiːnɪks`

SteffenDE (approved):

lib/phoenix/code_reloader/server.ex
preciz: The internet says yes.

SteffenDE:
Thank you! 🙌🏻

PR #5884, opened 2024-07-28
Fix `CoreComponents` margin between inputs and labels

### LINKED ISSUE #5832: Remove margin top on the input component if no label is provided in `core_components.ex`
### Environment

* Elixir version (elixir -v): 
```
Erlang/OTP 25 [erts-13.2.2.6] [source] [64-bit] [smp:10:10] [ds:10:10:10] [async-threads:1]

Elixir 1.16.1 (compiled with Erlang/OTP 25)
```
* Phoenix version (mix deps): `1.7.12`
* Operating system: MacOS

### Actual behavior

Hello Phoenix Team 👋 
This is a minor issue with the input component from the `core_components.ex` file (used in `phx.gen.live`)
A margin top is added by default for the spacing between the label and the input. 

https://github.com/phoenixframework/phoenix/blob/69685f783bc60d2a124fc2fb48028b065bc24b22/priv/templates/phx.gen.live/core_components.ex#L357-L361

However it is not removed if a label is not provided. Which causes alignment issue when combining the input component with other elements on our UI. 

mt-2-top-screen

Of course this can be fixed easily by modifying the `core_component` in my project but I guess this can be a quick win improvement to update the `core_components.ex` directly in Phoenix. I can propose a PR if you think it's worth making a small update 🙏 

### Expected behavior

Avoid adding a margin to the input if a `label` is not provided to the input component.

Fix: one idea could be to put the `mt-2` class behind a condition, e.g: 

```elixir
class={[ 
    "block w-full rounded-lg text-zinc-900 focus:ring-0 sm:text-sm sm:leading-6 min-h-[6rem]", 
    @errors == [] && "border-zinc-300 focus:border-zinc-400", 
    @errors != [] && "border-rose-400 focus:border-rose-400",
    # new condition
    label != nil && "mt-2",
]} 
```

NTurchi:
## Description 

This is a minor issue with the input component from the `core_components.ex` file (used in phx.gen.live)
A margin top is added by default on inputs for the spacing between the label and the input.

However it is not removed if a label is not provided. Which causes alignment issue when combining the input component with other elements.

[As suggested by José Valim](https://github.com/phoenixframework/phoenix/issues/5832#issuecomment-2218486504), putting the margin on the label itself is more appropriate. 

## Changes

- Added a class `attr` to the `label` component
- Remove `mt-2` from inputs in `CoreComponents`
- Add `mb-2` to `label` with `input` 

Closes #5832 

| **before:** margin on the input | **after:** margin on the label |
|-------|------|
| image |  demo |

### DIFF
--- a/priv/templates/phx.gen.live/core_components.ex
+++ b/priv/templates/phx.gen.live/core_components.ex
@@ -331,11 +331,11 @@ defmodule <%= @web_namespace %>.CoreComponents do
   def input(%{type: "select"} = assigns) do
     ~H"""
     <div>
-      <.label for={@id}><%%= @label %></.label>
+      <.label class="mb-2" for={@id}><%%= @label %></.label>
       <select
         id={@id}
         name={@name}
-        class="mt-2 block w-full rounded-md border border-gray-300 bg-white shadow-sm focus:border-zinc-400 focus:ring-0 sm:text-sm"
+        class="block w-full rounded-md border border-gray-300 bg-white shadow-sm focus:border-zinc-400 focus:ring-0 sm:text-sm"
         multiple={@multiple}
         {@rest}
       >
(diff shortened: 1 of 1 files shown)


priv/templates/phx.gen.live/core_components.ex
```diff
@@ -370,14 +370,14 @@ defmodule <%= @web_namespace %>.CoreComponents do
   def input(assigns) do
     ~H"""
     <div>
-      <.label for={@id}><%%= @label %></.label>
+      <.label for={@id} class="mb-2"><%%= @label %></.label>
```
lifeiscontent, line 373: https://mxstbr.com/thoughts/margin
```suggestion
    <div class="flex flex-col gap-2">
      <.label for={@id}><%%= @label %></.label>
      <.label for={@id}><%%= @label %></.label>
```

Issue #5886, opened 2024-07-30
Some tags of core components don't accept custom classes


ThiagoMonaco:
### Environment

* Elixir version (elixir -v): 1.16.3
* Phoenix version (mix deps): 1.7.14
* Operating system: MacOS Ventura 13.4.1

### Actual behavior

When using components from core components, some of them don't accept custom classes.
(in this example, the `.form`accepts new classes but `.label` and `.input` no)
![CleanShot 2024-07-30 at 19 50 13](https://github.com/user-attachments/assets/0949b89e-a81d-4877-8ece-a88476cfef58)
```
defmodule ChatWeb.CreateAccount.CreateAccountLive do
  alias Chat.Accounts
  alias Chat.Accounts.Account
  use ChatWeb, :live_view 
  
  def mount(_params, _session, socket) do
    changeset = Accounts.change_account(%Account{})
    {:ok, assign(socket, :form, to_form(changeset))}
  end

  def render(assigns) do
    ~H"""
      <div class="container">
        <h1 class="text-center text-4xl font-bold tracking-tight text-gray-900 sm:text-6xl">
          Account Creation
        </h1>
        <.form 
          class="flex w-full items-center flex-col mt-10 text-xl"
          for={@form}>
            <div class="flex w-1/2 justify-center flex-col ">
              <.label class="bg-red-800" for="name">Name</.label>
              <.input class="mt-3 bg-blue-600" field={@form[:name]} placeholder="Name" />
            </div>
        </.form>
      </div>
    """
  end
end
```

### Expected behavior
The class "bg-red-800" to be appended with default classes of `.label`
The class "mt-3 bg-blue-600" to be appended with default classes of `.input`

chrismccord:
https://github.com/phoenixframework/phoenix/pull/5887#issuecomment-2259426438

PR #5887, opened 2024-07-30
Allowing custom classes on some core components

ThiagoMonaco:
This changes allow custom classes in some core components provided by Phoenix (#5886)


chrismccord:
This is actually not something we want to promote as a general pattern. It's fine in your app here and there, but the issue is the precedence of the classes passed by the caller cannot be trusted. If the caller is trying to override or accidentally shadows an existing class, it's mostly a complete guess on which classes apply. This is because the css precedence actually users the order the class appears in the stylesheet. For customizing higher level components, what you really want is something like a variant system. Thanks!

PR #5891, opened 2024-08-01
Respect MIX_DEPS_PATH in generated config. Fix #5890. Fix #5414.

### LINKED ISSUE #5414: use env variable MIX_DEPS_PATH in config.exs templates
### Environment

* Phoenix version: 1.7.2

### Expected behavior

When using the MIX_DEPS_PATH env variable, the NODE_PATH (in config.exs) should be taken from that dependencies path.

### Actual behavior

Config templates call `%{"NODE_PATH" => Path.expand("../deps", __DIR__)}` so it fails to get the phoenix js library.
Templates may be defined as` %{"NODE_PATH" => Path.expand(System.get_env("MIX_DEPS_PATH", "../deps"), __DIR__)}`

As it's a template, we shouldn't expect to handle these cases, but in my case it would have saved me a lot of debugging time (I had an old deps folder in the default folder with an old phoenix version. So it was running, but with a lot of js errors).

Thanks!

### LINKED ISSUE #5890: tailwind config doesn't respect MIX_DEPS_PATH
### Environment

* Elixir version (elixir -v): 1.17.2
* Phoenix version (mix deps): 1.7.14
* Operating system: alpine linux (the `elixir:1.17.2-alpine` docker container)

### Context

I'm building a docker container to run my phoenix app in a dev environment. In the container I install the code at `/app`, which is a volume shared with my laptop so I can edit the code there. Also, I have a volume at `/deps` where I want the deps installed, so I used `MIX_DEPS_PATH` to accomplish this. I want the deps to be in an isolated volume to prevent conflicts on the laptop.

If I just put the volume at `/app/deps`, then if the laptop has installed deps those will get imported from the laptop into the volume just once at creation, and you get into a world of weird errors because they don't always update.

So I use `MIX_DEPS_PATH` to make it work.

### Actual behavior

When running `mix assets.build` while `MIX_DEPS_PATH` is set outside of the app, you get this error:

```
$ mix assets.build

Rebuilding...
Error: ENOENT: no such file or directory, scandir '/app/deps/heroicons/optimized/24/outline'
    at Object.readdirSync (node:fs:1405:3)
    at Object.readdirSync (pkg/prelude/bootstrap.js:1309:35)
    at /app/assets/tailwind.config.js:45:12
    at Array.forEach (<anonymous>)
    at /app/assets/tailwind.config.js:44:13
    at registerPlugins (/snapshot/tailwindcss/lib/lib/setupContextUtils.js:815:61)
    at createContext (/snapshot/tailwindcss/lib/lib/setupContextUtils.js:1221:5)
    at createContext (/snapshot/tailwindcss/lib/processTailwindFeatures.js:44:61)
    at Object.getContext (/snapshot/tailwindcss/lib/cli/build/plugin.js:200:24)
    at /snapshot/tailwindcss/lib/cli/build/plugin.js:247:38 {
  errno: -2,
  syscall: 'scandir',
  code: 'ENOENT',
  path: '/app/deps/heroicons/optimized/24/outline'
}
** (Mix) `mix tailwind my_app` exited with 1
```

This is probably from this line:
https://github.com/phoenixframework/phoenix/blob/2920eb281ad78f06c3c5f3fd5afb02f175562172/installer/templates/phx_assets/tailwind.config.js#L36
which doesn't seem to respect `MIX_ENV_PATH`.

A similar issue was reported in #5414.

### Expected behavior

I expect `mix assets.build` to work when `MIX_ENV_PATH` is set to a directory other than the usual path.

ericlathrop:
Make the tailwind config use `MIX_DEPS_PATH` if set to determine path to the assets.

### DIFF
--- a/config/config.exs
+++ b/config/config.exs
@@ -14,7 +14,7 @@ if Mix.env() == :dev do
     [
       args: ~w(./js/phoenix --bundle) ++ args,
       cd: Path.expand("../assets", __DIR__),
-      env: %{"NODE_PATH" => Path.expand("../deps", __DIR__)}
+      env: %{"NODE_PATH" => System.get_env("MIX_DEPS_PATH", Path.expand("../deps", __DIR__))}
     ]
   end
 
--- a/installer/templates/phx_assets/tailwind.config.js
+++ b/installer/templates/phx_assets/tailwind.config.js
@@ -33,7 +33,11 @@ module.exports = {
     // See your `CoreComponents.icon/1` for more information.
     //
     plugin(function({matchComponents, theme}) {
-      let iconsDir = path.join(__dirname, "..<%= if @in_umbrella, do: "/../.." %>/deps/heroicons/optimized")
+      let deps_path = path.join(__dirname, "..<%= if @in_umbrella, do: "/../.." %>/deps")
+      if (typeof process.env.MIX_DEPS_PATH === "string") {
+        deps_path = process.env.MIX_DEPS_PATH
+      }
+      let iconsDir = path.join(deps_path, "heroicons/optimized")
       let values = {}
       let icons = [
         ["", "/24/outline"],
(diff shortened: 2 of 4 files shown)


ericlathrop:
Also fixed #5414, which was closed, but still exists.

josevalim:
As per the issue, we don't plan to add more to generated templates. Thank you for the PR and documenting the steps though!

PR #5894, opened 2024-08-09
Use ensure_loaded to avoid concern from already loaded errors

fhunleth:
The return value from `Application.load/1` isn't checked so this doesn't
functionally change anything. However, when these functions are called
in a release, the loads return already loaded errors. The already loaded
error is completely harmless and switching to
`Application.ensure_loaded/1` makes go away.
`Application.ensure_loaded/1` has been available since Elixir 1.10.


Gazler:
Thanks <3

PR #5896, opened 2024-08-10
Minor improvements to guides

bkircher:
Fix small typo in mix phx.gen.auth guide and uniformly use Node.js (instead of Node.JS).


SteffenDE:
Thank you! 🙌🏻

PR #5898, opened 2024-08-12
Improve channel docs and example code

alexcybernetic:
Updating filename of heex template to hold token script tags
`.../templates/layout/app.html.heex` > `.../templates/layout/root.html.heex`


SteffenDE:
Thank you! 🙌🏻

PR #5899, opened 2024-08-12
Add missing ~p in phx.gen.auth templates

rhcarvalho:
For consistency with other URLs that use VerifiedRoutes.


SteffenDE:
Thank you! 🙌🏻

PR #5900, opened 2024-08-14
Simplified live generator

mcrumm:
Simplifies the `phx.gen.live` generator with the following changes:

* Replaces FormComponent with FormLive to handle `/new` and `/:id/edit` routes.
* Removes the `:id/show/edit` route. FormLive accepts a `?return_to=index|show` param to negotiate live navigation after successful form submit.
* Removes the `<.modal>` component and its helper functions as they are no longer used.


priv/templates/phx.gen.live/form.ex
```diff
@@ -0,0 +1,99 @@
...
+    <%= schema.singular %> = %<%= inspect schema.alias %>{}
+
+    socket
+    |> assign(:page_title, "New <%= schema.human_singular %>")
+    |> assign(:<%= schema.singular %>, <%= schema.singular %>)
+    |> assign(:form, to_form(<%= inspect context.alias %>.change_<%= schema.singular %>(<%= schema.singular %>)))
+  end
+
+  defp return_to("index"), do: :index
+  defp return_to("show"), do: :show
```
SteffenDE, line 92: I don't see `return_to=show` actually used anywhere, so we currently don't need that, right?

SteffenDE (approved):
Only one minor comment, lgtm!

priv/templates/phx.gen.live/form.ex
josevalim: Should the show page have a link to the edit one?

josevalim: Btw, I would skip the conversion to atom here. And instead have return_path match on strings directly. WDYT?

priv/templates/phx.gen.live/form.ex
```diff
@@ -0,0 +1,99 @@
...
+  defp return_to(_), do: :show
+
+  @impl true
+  def mount(_params, _session, socket) do
+    {:ok, socket}
+  end
```
josevalim, line 98: I think we typically have `mount` on top, right after render. We should probably set the `page_title` here too, wdyt?

priv/templates/phx.gen.live/index.ex
```diff
@@ -2,39 +2,15 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
...
-    socket
-    |> assign(:page_title, "Listing <%= schema.human_plural %>")
-    |> assign(:<%= schema.singular %>, nil)
-  end
-
-  @impl true
-  def handle_info({<%= inspect context.web_module %>.<%= inspect Module.concat(schema.web_namespace, schema.alias) %>Live.FormComponent, {:saved, <%= schema.singular %>}}, socket) do
-    {:noreply, stream_insert(socket, :<%= schema.collection %>, <%= schema.singular %>)}
+  def handle_params(_params, _url, socket) do
+    {:noreply, socket |> assign(:page_title, "Listing <%= schema.human_plural %>")}
```
josevalim, line 13: We can move the setting of page_title to mount, right?

priv/templates/phx.gen.live/form.ex
mcrumm: done!

priv/templates/phx.gen.live/form.ex
mcrumm: +1 for moving mount, but page_title is set in handle_params (actually apply_action) since it differs on new/edit :)

mcrumm: done!

priv/templates/phx.gen.live/index.ex
mcrumm: done in 7adef96

priv/templates/phx.gen.live/index.html.heex
```diff
@@ -1,7 +1,7 @@
 <.header>
   Listing <%= schema.human_plural %>
   <:actions>
-    <.link patch={~p"<%= schema.route_prefix %>/new"}>
+    <.link navigate={~p"<%= schema.route_prefix %>/new?return_to=index"}>
```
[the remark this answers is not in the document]
mcrumm: done in 432ae36

mcrumm:
@josevalim ready for another look!

priv/templates/phx.gen.live/form.ex
```diff
@@ -0,0 +1,98 @@
+defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web_namespace, schema.alias) %>Live.Form do
+  use <%= inspect context.web_module %>, :live_view
+
+  alias <%= inspect context.module %>
+  alias <%= inspect schema.module %>
+
+  @impl true
+  def render(assigns) do
+    ~H"""
+    <div>
```
josevalim, line 10: Sorry, small nit. We probably don't need this `div`? It was needed before because it was a function component. We also need to decide if we want a `render` function or to define `form.html.heex`. The file would be more consistent as that's also what we use for index/show. @chrismccord, any preferences?

priv/templates/phx.gen.live/form.ex
```diff
@@ -0,0 +1,98 @@
...
+
+  @impl true
+  def mount(_params, _session, socket) do
+    {:ok, socket}
+  end
+
+  @impl true
+  def handle_params(params, _url, socket) do
+    socket = assign(socket, :return_to, return_to(params["return_to"]))
+    {:noreply, apply_action(socket, socket.assigns.live_action, params)}
```
josevalim, line 36: I think this can all be done on mount, since it is based on params, and we don't expect the same LiveView to migrate from new/edit. Or `live_action` only works on `handle_params`?

priv/templates/phx.gen.live/form.ex
```diff
@@ -0,0 +1,98 @@
...
+      </.header>
+
+      <.simple_form for={@form} id="<%= schema.singular %>-form" phx-change="validate" phx-submit="save">
+<%= Mix.Tasks.Phx.Gen.Html.indent_inputs(inputs, 8) %>
+        <:actions>
+          <.button phx-disable-with="Saving...">Save <%= schema.human_singular %></.button>
+        </:actions>
+      </.simple_form>
+
+      <.back navigate={return_path(@return_to.key, @<%= schema.singular %>)}>Back to <%%= @return_to.name %></.back>
```
josevalim, line 23:
```suggestion
      <.back navigate={return_path(@return_to.key, @<%= schema.singular %>)}>Back</.back>
```

And then we skip the whole plural/singular? Or we make this generic and not related to the back functionality.


```suggestion      
      <%= if @<%= schema.singular %>.id %>
        <.link navigate={~p"<%= schema.route_prefix %>/#{@<%= schema.singular %>}"}>See <%= schema.singular %></.link> |
      <% end %>
      <.link navigate={~p"<%= schema.route_prefix %>"}>All <%= schema.plural %></.link>
```

Then we kill the <.back> in show as well and get rid of another component. WDYT @chrismccord?

josevalim: Actually, if we kill `.back`, we can do it another PR, to keep this one focused.

priv/templates/phx.gen.live/form.ex
josevalim: @chrismccord prefers to have all of them colocated, so perhaps we do that in another PR?

chrismccord: We can kill the div because it was only there for the LC which requires a root, so ✂️ away

mcrumm: Done in e52c465, good catch :)

priv/templates/phx.gen.live/form.ex
mcrumm: Great catch! Done in e52c465

priv/templates/phx.gen.live/form.ex
SteffenDE: > Colocated the templates in https://github.com/phoenixframework/phoenix/commit/e21c02560eb0cc963d1ab4d416a7a898015dcd18 :)

Love it!

josevalim (approved):
Beautiful!

TheArrowsmith:
Why remove `modal` from corecomponents? Just because it's not used in the generators doesn't mean it's not useful elsewhere, e.g. to include it in one of my own liveviews/components.

josevalim:
The rule has always been to ship with the components used by the generators, otherwise we could easily argue that core components should be a 1000LOC file with accordions, stats bars, breadcrumbs, etc. People can always add the features they need.

TheArrowsmith:
Ah okay, that's totally fair. Thanks for the reply!

clarkware:
Great work, Michael! This will definitely help folks new to Phoenix. Can't wait to have it in an official release! 😀

josevalim:
Of course, as any suggestion we receive, from newcomers and experienced developers alike, we discuss, interpret and filter it accordingly. We aren’t doing changes because people ask them. We do them because they ask, we evaluate and then proceed accordingly.

In any case, the CRUD capabilities are all still there. Only the modals are gone.

kuatroka:
Thanks @josevalim 
@mcrumm - would it be possible to share a screenshot of the new UI of :edit and :show will look like? Thanks

clarkware:
Is the plan to hold this back until the 1.8 release? I was surprised it wasn't part of recent 1.7 releases.

josevalim:
Yes, on v1.8. it would be a breaking change to change the generators this considerably in a patch release. :)

TheArrowsmith:
@kuatroka you can see the new UI yourself - just update your `mix.exs` to get the Phoenix dep from Github main instead of from hex:

```elixir
{:phoenix, github: "phoenixframework/phoenix"},
```

You might need to pull some other deps from GitHub too e.g. `:phoenix_live_view`. I forget the details but it's definitely possible to get Phoenix working locally with latest `main` to preview the new generators, I've done it myself.

SteffenDE:
The easiest way to test the new generators is by cloning the repo, fetching deps and then going into the installer folder, from there:

```
mix phx.new my_app --dev
```

note the `--dev`. I think the steps are also written in the README.

kuatroka:
I've tried @SteffenDE 's approach as I wanted to test in a totally new app before applying it to mine, but faces too many dependency issues when standing up the up and trying to run generators. So I have just left this topic for now. I guess I'll test it when it's all ready and polished. Thanks all

PR #5904, opened 2024-08-20
WebSockets: support passing through sec-websocket-protocol header

feld:
Opening this MR to begin a wider discussion on this.

Let's start with the current state of things: in Phoenix it's not possible to access any headers of Websocket connections except the X-Foo headers if you explicitly enable it with `connect_info: [:x_headers]`. The [docs](https://hexdocs.pm/phoenix/Phoenix.Endpoint.html#socket/3-connect-info) are very clear about this limitation:

> Phoenix only gives you limited access to the connection headers for security reasons...

However, this severely limits the capabilities of Phoenix. We converted Pleroma from Phoenix+Cowboy for websockets to `Phoenix.Socket.Transport` so we could support both Cowboy and Bandit. In the process I encountered a [weird test case](https://git.pleroma.social/pleroma/pleroma/-/merge_requests/4064/diffs#bef485859b4b4e2501c0f84a2432a35b7c01f53d_271_271) we had written that validated you could pass through an auth token in the `Sec-WebSocket-Protocol` header.

Nobody could remember at the time why this existed.

It turns out, this is actually how the Mastodon client protocol authenticates over websockets!

At first I thought this was a case of abusing this header for something it wasn't meant for. The RFC doesn't make it sound like this is the purpose of the header. So why were they doing it this way?

As it turns out, [lots of applications do this](https://github.com/search?q=Sec-WebSocket-Protocol+auth&type=code). [Here's one](https://github.com/zerobugdebug/websocket-authorizer) from 3 weeks ago doing it in AWS Lambda.

So how can we make sure Phoenix is a suitable framework for all types of Websocket applications/protocols as the current limitations are very narrow-sighted? I do not want to have to [maintain a fork](https://git.pleroma.social/pleroma/pleroma/-/merge_requests/4206) of Phoenix to support this.

### DIFF
--- a/lib/phoenix/socket/transport.ex
+++ b/lib/phoenix/socket/transport.ex
@@ -257,7 +257,7 @@ defmodule Phoenix.Socket.Transport do
 
     connect_info =
       Enum.map(connect_info, fn
-        key when key in [:peer_data, :trace_context_headers, :uri, :user_agent, :x_headers] ->
+        key when key in [:peer_data, :trace_context_headers, :uri, :user_agent, :x_headers, :sec_websocket_protocol] ->
           key
 
         {:session, session} ->
@@ -268,7 +268,7 @@ defmodule Phoenix.Socket.Transport do
 
         other ->
           raise ArgumentError,
-                ":connect_info keys are expected to be one of :peer_data, :trace_context_headers, :x_headers, :uri, or {:session, config}, " <>
+                ":connect_info keys are expected to be one of :peer_data, :trace_context_headers, :x_headers, :user_agent, :sec_websocket_protocol, :uri, or {:session, config}, " <>
                   "optionally followed by custom keyword pairs, got: #{inspect(other)}"
       end)
 
@@ -458,6 +458,8 @@ defmodule Phoenix.Socket.Transport do
 
     * `:user_agent` - the value of the "user-agent" request header
 
+    * `:sec_websocket_protocol` - the value of the "sec-websocket-protocol" header
+
   """
   def connect_info(conn, endpoint, keys) do
     for key <- keys, into: %{} do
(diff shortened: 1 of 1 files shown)


lambadalambda:
I second this, the current behavior is great for new applications, but if you want to re-implement somebody else's API the only way to do it seems to be to run a fork. 

Maybe it's possible to expose these headers after some config setting has been set that makes it explicit that this is not something you should do if it can be avoided in any way?

feld:
What is needed from me to move this along? If I can get an idea of what solution will be accepted I will gladly update the branch to include the required changes to docs, tests, etc.

mtrudel:
If you want to write a handler for a 'raw' websocket (ie: without channels), you can trivially do this by calling `Plug.Conn.upgrade` in your controller and passing in a WebSock implementation. I gave a talk at ElixirConf EU a few years ago outlining the process: https://www.youtube.com/watch?v=usKLrYl4zlY

josevalim:
My understanding is that the subprotocol is set to a static field, such as "bearer", so we do support setting subprotocols on the websocket configuration: https://hexdocs.pm/phoenix/Phoenix.Endpoint.html#socket/3-websocket-configuration

If that is not enough, we would gladly merge a PR that adds `:sec_websocket_headers` to `connect_info` and we would pass all headers starting with "sec-websocket" (as we pass x-headers). If that's really a requirement, we may even consider deprecating the subprotocols feature in the future (as folks can now easily check that themselves).

feld:
I also thought that the subprotocol was meant to be a static field, but that's not how it's being used in practice for some existing services/protocols. It is indeed strange that this is being treated as a header to hold authentication tokens.

I will rework this branch to support the `:sec_websocket_headers` option to `connect_info`.

PR #5916, opened 2024-08-28
phx.new regex matches regex from mix new

### LINKED ISSUE #5915: `mix phx.new` does not comply with `mix new` Application Name validation and fails later in mix release
### Environment

* Elixir version (elixir -v): 
Erlang/OTP 26 [erts-14.2.5] [source] [64-bit]
Elixir 1.16.2 (compiled with Erlang/OTP 26)

* Phoenix version (mix deps):  1.7.14
* Operating system: MacOS (Darwin Kernel Version 23.6.0)

### Actual behavior

In the current implementation, I can create an app with upper case letters, although according t[o the error message](https://github.com/phoenixframework/phoenix/blob/8fd768355e665f54955af4bcda75741ffa1f7d46/installer/lib/mix/tasks/phx.new.ex#L353) only lower case letters should be used for an app name. This is a different behaviour than with `mix new` - which catches us later with a `mix release` (in my example in the generated Docker file in line 64)  

```
mix phx.new exAppName
cd exAppName
mix phx.gen.release --docker
docker build .

 > [builder 17/17] RUN mix release:                                                                                                                                                                
0.523 ** (Mix) Invalid release name. A release name must start with a lowercase ASCII letter, followed by lowercase ASCII letters, numbers, or underscores, got: :exAppName
------
Dockerfile:64
--------------------
  62 |     
  63 |     COPY rel rel
  64 | >>> RUN mix release
  65 |     
  66 |     # start a new build stage so that the final image will only contain

```

### Expected behavior
```
mix phx.new exAppName
** (Mix) Application name must start with a letter and have only lowercase letters, numbers and underscore, got: "exAppName". The application name is inferred from the path, if you'd like to explicitly name the application then use the `--app APP` option.
```

bmalum:
fixes #5915


SteffenDE (approved):

SteffenDE:
LGTM, especially since it not actually matches the warning

> Application name must start with a letter and have only lowercase letters, numbers and underscore

Interestingly there's actually a test trying to use an uppercase name:

https://github.com/phoenixframework/phoenix/blob/8fd768355e665f54955af4bcda75741ffa1f7d46/installer/test/phx_new_umbrella_test.exs#L528

🤔

PR #5919, opened 2024-09-01
Only spin loading spinners if user does not prefer reduced motion

mattmatters:
This disables some animation if a user prefers reduced motion using one of those tailwind modifiers described [here](https://tailwindcss.com/docs/animation#prefers-reduced-motion).


SteffenDE:
Thank you! 🙌🏻

PR #5920, opened 2024-09-01
Use aria-live="polite" for flash_groups

mattmatters:
This nicely notifies screen readers and such of changed content/alerts.

See [aria-live](https://developer.mozilla.org/en-US/docs/Web/Accessibility/ARIA/Attributes/aria-live) for more info.

I've seen some folks use `assertive` instead of `polite` for the aria-live setting in a flash/toast context. I'm unsure if there is a defacto one to use in this case though. Open to feedback!


ponychicken:
I think for most cases "polite" seems like a reasonable choice. In any case, its part of CoreComponents, so if users desire they can easily alter the default.

SteffenDE:
Thank you! 🙌

PR #5921, opened 2024-09-02
Correct example router.ex code in Plug guide

jrko:
The first example (adding locale) of router.ex doesn't align with generated code or subsequent examples. There are two issues:

1.  `plug :fetch_flash` should be `plug :fetch_live_flash` 
2. Missing line: `plug :put_root_layout ...`


Gazler:
Thanks <3

PR #5923, opened 2024-09-03
Update docs for `Phoenix.Param`

kevinschweikert:
Updates the documentation to show usage in `VerifiedRoutes` instead of Route helpers and shows an example to directly implement the protocol for an internal Elixir `Date` struct instead of deriving an implementation, when you don't define the struct yourself


SteffenDE (approved):

SteffenDE:
Thank you! 🙌🏻

PR #5928, opened 2024-09-06
Send download encodes reserved chars

### LINKED ISSUE #5926: Error when using send_download/3 with a filename that contains spaces and commas
### Environment

* Elixir version (elixir -v):
Erlang/OTP 26 [erts-14.2.5] [source] [64-bit] [smp:6:6] [ds:6:6:10] [async-threads:1]
Elixir 1.16.3 (compiled with Erlang/OTP 26)
* Phoenix version (mix deps):
phoenix 1.7.12 (Hex package) (mix)
* Operating system:
MacOS

### Actual behaviour
Chrome shows an error page stating that the server sent an invalid response with the error `ERR_RESPONSE_HEADERS_MULTIPLE_CONTENT_DISPOSITION`. Sending a file with filename `datei, mit beistrich.pdf` will send the following header.
```
content-disposition: attachment; filename="datei,%20mit%20beistrich.pdf"; filename*=utf-8''datei,%20mit%20beistrich.pdf
```
Notably, the comma has not been encoded. 

### Expected behaviour
For context, the encoding has been adjusted in as part of #5478. Looking at [RFC5987](https://www.rfc-editor.org/rfc/rfc5987), and it's definition of a token in [RTC2616](https://www.rfc-editor.org/rfc/rfc2616#section-2.2), I believe the comma and other characters should have been encoded, since they are excluded. A possible fix would be to change the function call in [Phoenix.Controller.encode_filename/2](https://github.com/phoenixframework/phoenix/blob/v1.7.14/lib/phoenix/controller.ex#L1277) to `URI.encode(filename, &URI.char_unreserved?/1)`

mcrumm:
- Closes #5926

### DIFF
--- a/test/phoenix/controller/controller_test.exs
+++ b/test/phoenix/controller/controller_test.exs
@@ -537,6 +537,17 @@ defmodule Phoenix.Controller.ControllerTest do
              "world"
     end
 
+    test "issue #5926, sends file for download for filename with unreserved characters" do
+      conn = send_download(conn(:get, "/"), {:file, @hello_txt}, filename: "hello, world.json")
+      assert conn.status == 200
+      assert get_resp_header(conn, "content-disposition") ==
+             ["attachment; filename=\"hello%2C%20world.json\"; filename*=utf-8''hello%2C%20world.json"]
+      assert get_resp_header(conn, "content-type") ==
+             ["application/json"]
+      assert conn.resp_body ==
+             "world"
+    end
+
     test "sends file supports UTF-8" do
       conn = send_download(conn(:get, "/"), {:file, @hello_txt}, filename: "测 试.txt")
       assert conn.status == 200
--- a/test/phoenix/controller/controller_test.exs
+++ b/test/phoenix/controller/controller_test.exs
@@ -537,7 +537,7 @@ defmodule Phoenix.Controller.ControllerTest do
              "world"
     end
 
-    test "issue #5926, sends file for download for filename with unreserved characters" do
+    test "sends file for download for filename with unreserved characters" do
       conn = send_download(conn(:get, "/"), {:file, @hello_txt}, filename: "hello, world.json")
       assert conn.status == 200
       assert get_resp_header(conn, "content-disposition") ==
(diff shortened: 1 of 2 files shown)


test/phoenix/controller/controller_test.exs
```diff
@@ -537,6 +537,17 @@ defmodule Phoenix.Controller.ControllerTest do
              "world"
     end
 
+    test "issue #5926, sends file for download for filename with unreserved characters" do
```
Gazler, line 540: I don't think we usually include the issue # in the test name.

Gazler (approved):
LGTM!

test/phoenix/controller/controller_test.exs
mcrumm, line 540:
```suggestion
    test "sends file for download for filename with unreserved characters" do
```

josevalim:

ftes:
It looks like this still hasn't been released. Is there anything blocking this from being released?

SteffenDE:
@ftes there are a lot of changes on main that are waiting for 1.8. We're planning to focus on the missing pieces in the next couple of weeks. For now, we haven't planned to do another 1.7 release, but maybe we should? @josevalim

josevalim:
I didn't backport this on purpose to v1.7 because people could be relying on the old behaviour for some reason, even if buggy. So you can either prune it yourself, use main, or wait until v1.8 is out :)

ftes:
Thanks for the update!

Workaround in user land:

```ex
filename = URI.encode(filename, &URI.char_unreserved?/1)
send_download(conn, {:binary, body}, filename: filename, encode: false)
```

artonragsdale:
We ran into this issue in production, where some proxies fail the download completely if commas are present in the filename. I'm not sure why anyone would depend on the incorrect behavior. Chrome-based browsers also trims the end of the filename after a comma, resulting in no file extension. I feel like it's worth a patch, but we rolled back to phoenix 1.6 for a quick fix and will probably try the workaround mentioned above until 1.8 is released I guess.

PR #5930, opened 2024-09-10
fix broken link

lbellomo:
I am trying to fix a broken link [Working with Fly.io applications](https://fly.io/docs/getting-started/working-with-fly-apps/). 
According to what I see from [IA](https://web.archive.org/web/20200812153205/https://fly.io/docs/getting-started/working-with-fly-apps/#viewing-applications), the page doesn't exist anymore and they separated it into different parts.


Gazler:
Thanks <3

PR #5934, opened 2024-09-20
Clarify instructions for using npm install method for third-party JS packages

connorjacobsen:
In Phoenix `1.7.14`, using `npm install some-package --save` from inside the `assets/` directory doesn't actually appear to do anything while `npm install some-package --prefix assets` creates `package.json` and `package-lock.json` files and is what was suggested in the auto-generated `app.js` that you get with a new Phoenix app. So I attempted to tweak the guide to clarify this + make it match the comments in `app.js`.

Feel free to tweak / ignore as desired.

### DIFF
--- a/guides/asset_management.md
+++ b/guides/asset_management.md
@@ -20,7 +20,7 @@ If you want to import JavaScript dependencies, you have at least three options t
    import topbar from "../vendor/topbar"
    ```
 
-2. Call `npm install topbar --save` inside your assets directory and `esbuild` will be able to automatically pick them up:
+2. Call `npm install topbar --prefix assets` will create `package.json` and `package-lock.json` inside your assets directory and `esbuild` will be able to automatically pick them up:
 
    ```js
    import topbar from "topbar"


SteffenDE:
Do you have a package.json file in the root of your repo/project?

connorjacobsen:
Not at the root, no. But the docs currently say to run the command from the `assets/` dir. Would it work that way if you did?

Running the command specified in the generated code comments from the project root made one in the `assets/` dir and worked properly, though.

SteffenDE:
Weird. From my experience npm looks at the parent directories for a package.json file. If it doesn't fine one it creates a new package.json in the current working directory. I just tested this in a new (mix phx.new) project.

```
mix phx.new assets_test
cd assets_test/assets
npm install svelte --save
```

And it correctly created the files. So you might want to check you parent folders (multiple levels) for any stale package.json files that you might have accidentally created.

That being said, explicitly setting the prefix isn't wrong either and it will help in such cases, so I'll merge this :)

Thank you!

PR #5935, opened 2024-09-20
change es2017 to es2022 in asset management guide

srcrip:
es2022 is pretty safe now imo


SteffenDE:
I’d say the docs should follow what the generator generates:

https://github.com/phoenixframework/phoenix/blob/48ee8cc407f060eee1152f2c8f64bff49a6b4f15/installer/templates/phx_single/config/config.exs#L41

So, either we change it in the generator templates as well, or we’ll keep it as is. Is there any notable improvement using the 2022 target instead of 2017, e.g. smaller bundle size?

srcrip:
Yes, I'd reccomend changing the generator too. I just missed that. I'll change my PR. As for improvements yes, there's a number of features that esbuild will convert to some kind of alternative if you specify a lower level, so the improvement would be using those native features + the bundle size would be slightly smaller cause esbuild doesn't need to provide it's own polyfills

SteffenDE (approved):

SteffenDE:
Thank you! 🙌🏻

PR #5937, opened 2024-09-24
Correct migration step in phx.gen.auth docs

jrko:
After running `mix phx.gen.auth` you are prompted to run `mix deps.get` followed by `mix ecto.migrate`.

The docs reference `mix ecto.setup` however, which seems to be deprecated in favor of `mix ecto.create`, neither of which are correct in this context, assuming a basic Phoenix app is up and running prior to adding auth.

This PR corrects the docs, replacing `mix ecto.setup` with `mix ecto.migrate` along with an appropriate description of the step.

It also adds the assumption that the reader has gone through the introductory guides and has a running Phoenix app, as well as calling out the addition of the `Register` and `Log in` links added to the default page. 

I also took the liberty to remove some extraneous language about umbrella app root locations, which seemed unnecessary in the context of learning auth basics.

### DIFF
--- a/guides/authentication/mix_phx_gen_auth.md
+++ b/guides/authentication/mix_phx_gen_auth.md
@@ -1,12 +1,14 @@
 # mix phx.gen.auth
 
+> This guide assumes that you have gone through the [introductory guides](overview.html) and have a Phoenix application [up and running](up_and_running.html).
+
 The `mix phx.gen.auth` command generates a flexible, pre-built authentication system into your Phoenix app. This generator allows you to quickly move past the task of adding authentication to your codebase and stay focused on the real-world problem your application is trying to solve.
 
 ## Getting started
 
 > Before running this command, consider committing your work as it generates multiple files.
 
-Let's start by running the following command from the root of our app (or `apps/my_app_web` in an umbrella app):
+Let's start by running the following command from the root of our app:
 
 ```console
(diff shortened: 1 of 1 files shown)


josevalim:
`ecto.setup` is generated as part of new applications, so we should continue to refer to it: https://github.com/phoenixframework/phoenix/blob/a2e7de1058a0972447d8a0e362166cc6b6911f57/installer/templates/phx_single/mix.exs#L79.

If your app does not have it, maybe it was generated in older versions or removed?

jrko:
Aah, I do indeed have `ecto.setup` (using Elixir 1.14.0, Erlang/OTP 24, Phoenix 1.7.14). I think my confusion came from a couple places:

1. The console output of `mix phx.new` and the `Up and Running` docs both reference `ecto.create`.
2. Searching [Ecto HexDocs](https://hexdocs.pm/ecto/search.html?q=ecto.setup) for `ecto.setup` returns no results, but `ecto.create` does.

I see now that `ecto.setup` is a valid alias in Phoenix, though it still seems a bit awkward to have the console output of `phx.gen.auth` reference `ecto.migrate` as a next step, while the auth docs reference `ecto.setup`.

Anyway, feel free to close the PR if you like. I was able to work through the confusion without too much fuss, but thought I might be able to streamline the docs a bit for future newcomers, like myself!

josevalim:

PR #5938, opened 2024-09-24
Remove an old and unused PubSub subscribe option

joshk:
The `link: true` option was supported with versions of `phoenix_pubsub` earlier than 2.0.0, and was removed in this commit https://github.com/phoenixframework/phoenix_pubsub/commit/c05d47f6411d81f3cd94d53607848b5195239c4f

As Phoenix requires `phoenix_pubsub` `~> 2.1`, I believe removing these options is safe.


josevalim:

PR #5939, opened 2024-09-25
Remove an old test file from the installer

JonRowe:
👋 I found this file whilst looking through the installer templates, it appears not to be referenced anywhere and would never pass unless I'm missing something in a freshly generated app, was it left behind from #4431?


SteffenDE:
I think you're completely right :)
Good catch, thank you! 🙌🏻

PR #5940, opened 2024-09-26
phx.new - Fix deprecation warning when using Elixir ~> 1.18

psantos10:
When generating a new application using Elixir 1.18.0-dev, this warning is raised:

> Compiling 1 file (.ex)
>      warning: Mix.Rebar.rebar_cmd/1 is deprecated. Use rebar_args/2 or available?/1 instead
>      │
>  275 │       Mix.Rebar.rebar_cmd(:rebar3)
>      │                 ~
>      │
>      └─ lib/mix/tasks/phx.new.ex:275:17: Mix.Tasks.Phx.New.rebar_available?/0

Not sure if this is a good practice, but that's I come up with.

### DIFF
--- a/installer/lib/mix/tasks/phx.new.ex
+++ b/installer/lib/mix/tasks/phx.new.ex
@@ -268,7 +268,7 @@ defmodule Mix.Tasks.Phx.New do
 
   if Version.match?(System.version(), "~> 1.18") do
     defp rebar_available? do
-      Mix.Rebar.rebar_args(:rebar3, [])
+      true
     end
   else
     defp rebar_available? do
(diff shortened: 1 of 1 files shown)


installer/lib/mix/tasks/phx.new.ex
```diff
@@ -266,8 +266,14 @@ defmodule Mix.Tasks.Phx.New do
     Code.ensure_loaded?(Hex)
   end
 
-  defp rebar_available? do
-    Mix.Rebar.rebar_cmd(:rebar3)
+  if Version.match?(System.version(), "~> 1.18") do
+    defp rebar_available? do
+      Mix.Rebar.rebar_args(:rebar3, [])
```
josevalim, line 271: Rebar is automatically installed on more recent Elixir versions, so we are good. :)

```suggestion
      true
```

josevalim:

PR #5941, opened 2024-09-28
 Open default Phoenix app documentation links in a new tab

daniellionel01:
Usually you want these links to open in a new tab, so that you dont navigate away from localhost when starting up the project.

So I added `target="_blank"` to all ankers on the starting page.


josevalim:
Thank you but I believe this is fine as is. Someone can equally dislike when something opens up in a new tab and they didn't want to. I can press cmd if I want to get a new tab. Thank you.

daniellionel01:
Ok :)

PR #5943, opened 2024-10-03
allow users to filter out certain `handle_in` messages

cocoa-xu:
When running in dev mode there can be quite a lot `handle_in` messages being logged, and it would be better if we can filter out certain messages, e.g., filter out the ones that have already been verified to work and just log the ones I'm interested in or currently debugging.


tw00 (approved):

josevalim:
Instead of making this a compile-time configuration, perhaps we can make it a socket one? Perhaps an API that adds a filter to the `socket`? For example, on join, we could do:

```elixir
log_filter(socket, &some_function/1)
```

SteffenDE:
Is there still interest to move this forward? :)

PR #5948, opened 2024-10-09
Fix action rename example in controllers guides

ShPakvel:
Only action name should be changed to match route renaming.
Change of passing template's name leads to bug, as the template is not renamed.
And description above clearly talks only about action.


josevalim:

PR #5950, opened 2024-10-10
Correct query result in doc to verify_change_email_token_query.

ShPakvel:
Simple typo in doc.
Query build in method `verify_change_email_token_query` differs from `verify_email_token_query`. It returns `user_token` instead of `user`.
For quick presentation of proof, here is reference to invocation code:
```elixir
    with {:ok, query} <- UserToken.verify_change_email_token_query(token, context),
         %UserToken{sent_to: email} <- Repo.one(query),
```


josevalim:

PR #5952, opened 2024-10-13
Add `check_csrf` option to socket transport options

tanguilp:
This PR proposes adding a `check_csrf` option to Phoenix's socket transport, so as to enable caching Liveview's static rendering.

## Context

After releasing [`plug_http_cache`](https://github.com/tanguilp/plug_http_cache), I was asked if we could use it to cache Liveviews.

After a few experiment, I came with the following PR last year: https://github.com/phoenixframework/phoenix/pull/5667. Note that in this PR, I **didn't** follow the proposed approach in the discussion there (prefer being honest here :see_no_evil:).

I took more time to study the different approaches how to implement it this year. I described it in 2 blog posts:
- [Caching Liveviews - Part 1: The road to HTTP-caching Liveviews](https://svground.fr/blog/posts/caching-liveviews-part-1/)
- [Caching Liveviews - Part 2: Publicly caching private Liveviews](https://svground.fr/blog/posts/caching-liveviews-part-2/)

We're talking about caching Liveview's with *session enabled*. When disabled, caching works OOTB.

Key takeaways are:
- there are 2 mechanisms to prevent CSWSH: checking `origin` and checking a CSRF token
- both mechanisms protect against CSWSH. As a consequence using only one is safe
- by default Phoenix uses both for websocket & longpoll transports. `origin` check can be disabled, CSRF check cannot
- we can't cache liveviews because the generated pages contain a CSRF token, which is user-specific
- we propose instead to enable `origin` check, and disable the CSRF check
- the second article also discusses:
  - how to selectively caching some liveviews (opt-in caching)
  - a pattern to avoid accidentally cache private data (`assign_private`, similar to `assign_async`)

I believe this pattern (*Publicly caching private Liveviews*) is actually another use-case for Liveview, separate but not exclusive to server-rendered interactivity. Usually you would make heavy use of javascript, requesting a separate APIs to get the little bits of private user data. The proposed pattern greatly simplifies that, and is compatible with any caching system (`plug_http_cache`, CDN, any shared cache between the Phoenix server and the user browser). But only time will tell :grin: 

Screenshot from 2024-09-04 16-45-11

## Implementation

To disable CSRF check, you would do:

```elixir
socket "/live", Phoenix.LiveView.Socket, websocket: [connect_info: [session: @session_options], check_csrf: false]
```

The `check_csrf` option is transport-specific, contrary to `check_origin` that can be configured at the endpoint level as well. This is because:
- we could have non-web transports for which checking CSRF makes no sense
- this could confuse users, because you still need to check CSRF for Phoenix views' forms for instance

We change the signature of `Phoenix.Socket.Transport.connect_info/4` to add an `opts` parameter that defaults to `[check_csrf: true]`.

When disabling CSRF check for Phoenix socket transports, we no longer need to 1) add a CSRF token in the HTML page 2) send the token as a parameter to the JS `LiveSocket` call.

A test was added to check that when this option is set to `true` (default value), including an invalid CSRF token fails to retrieve session data.

From a security perspective, I think that with the default values (secure by default) and the current warnings, it's hard to make Liveview insecure without noticing it.

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -859,8 +859,8 @@ defmodule Phoenix.Endpoint do
       The MFA is invoked with the request `%URI{}` as the first argument,
       followed by arguments in the MFA list, and must return a boolean.
 
-    * `:check_csrf` - if the transport should perform CSRF check. If `origin` check is disabled as
-    well as CSRF check, your app is vulnerable to Cross-Site WebSocket Hijacking (CSWSH) attacks.
+    * `:check_csrf` - if the transport should perform CSRF check. Note that disabling
+    both CSRF and origin checks at the same time is not allowed and will raise.
     Defaults to `true`
 
     * `:code_reloader` - enable or disable the code reloader. Defaults to your
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -860,8 +860,8 @@ defmodule Phoenix.Endpoint do
       followed by arguments in the MFA list, and must return a boolean.
 
     * `:check_csrf` - if the transport should perform CSRF check. Note that disabling
-    both CSRF and origin checks at the same time is not allowed and will raise.
-    Defaults to `true`
+      both CSRF and origin checks at the same time is not allowed and will raise.
+      Defaults to `true`
 
     * `:code_reloader` - enable or disable the code reloader. Defaults to your
       endpoint configuration
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -859,9 +859,12 @@ defmodule Phoenix.Endpoint do
       The MFA is invoked with the request `%URI{}` as the first argument,
       followed by arguments in the MFA list, and must return a boolean.
 
-    * `:check_csrf` - if the transport should perform CSRF check. Note that disabling
-      both CSRF and origin checks at the same time is not allowed and will raise.
-      Defaults to `true`
+    * `:check_csrf` - if the transport should perform CSRF check. To avoid
+      "Cross-Site WebSocket Hijacking", you must have at least one of
+      `check_origin` and `check_csrf` enabled. If you set both to `false`,
+      Phoenix will raise, but it is still possible to disable both by passing
+      a custom MFA to `check_origin`. In such cases, it is your responsibility
+      to ensure at least one of them is enabled. Defaults to `true`
 
     * `:code_reloader` - enable or disable the code reloader. Defaults to your
       endpoint configuration
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -859,6 +859,10 @@ defmodule Phoenix.Endpoint do
       The MFA is invoked with the request `%URI{}` as the first argument,
       followed by arguments in the MFA list, and must return a boolean.
 
+    * `:check_csrf` - if the transport should perform CSRF check. If `origin` check is disabled as
+    well as CSRF check, your app is vulnerable to Cross-Site WebSocket Hijacking (CSWSH) attacks.
+    Defaults to `true`
+
     * `:code_reloader` - enable or disable the code reloader. Defaults to your
       endpoint configuration
 
--- a/lib/phoenix/endpoint/supervisor.ex
+++ b/lib/phoenix/endpoint/supervisor.ex
@@ -84,11 +84,10 @@ defmodule Phoenix.Endpoint.Supervisor do
     children =
       config_children(mod, secret_conf, default_conf) ++
         pubsub_children(mod, conf) ++
-        socket_children(otp_app, mod, :child_spec) ++
+        socket_children(mod, conf, :child_spec) ++
         server_children(mod, conf, server?) ++
-        socket_children(otp_app, mod, :drainer_spec) ++
+        socket_children(mod, conf, :drainer_spec) ++
         watcher_children(mod, conf, server?)
-
     Supervisor.init(children, strategy: :one_for_one)
   end
 
@@ -118,12 +117,11 @@ defmodule Phoenix.Endpoint.Supervisor do
     end
   end
 
-  defp socket_children(otp_app, endpoint, fun) do
+  defp socket_children(endpoint, conf, fun) do
     for {_, socket, opts} <- Enum.uniq_by(endpoint.__sockets__(), &elem(&1, 1)),
+        _ = check_origin_or_csrf_checked!(conf, opts),
         spec = apply_or_ignore(socket, fun, [[endpoint: endpoint] ++ opts]),
         spec != :ignore do
-      check_origin_or_csrf_checked!(otp_app, endpoint, opts)
-
       spec
     end
   end
@@ -137,12 +135,10 @@ defmodule Phoenix.Endpoint.Supervisor do
     end
   end
 
-  defp check_origin_or_csrf_checked!(otp_app, endpoint, socket_opts) do
-    endpoint_check_origin = config(otp_app, endpoint)[:check_origin]
-
-    for {transport, transport_opts} <- socket_opts do
+  defp check_origin_or_csrf_checked!(endpoint_conf, socket_opts) do
+    for {transport, transport_opts} <- socket_opts, is_list(transport_opts) do
       check_origin =
-        Keyword.get(transport_opts, :check_origin, endpoint_check_origin)
+        Keyword.get(transport_opts, :check_origin, endpoint_conf[:check_origin])
 
       check_csrf = transport_opts[:check_csrf]
 
--- a/lib/phoenix/endpoint/supervisor.ex
+++ b/lib/phoenix/endpoint/supervisor.ex
@@ -136,9 +136,10 @@ defmodule Phoenix.Endpoint.Supervisor do
   end
 
   defp check_origin_or_csrf_checked!(endpoint_conf, socket_opts) do
+    check_origin = endpoint_conf[:check_origin]
+
     for {transport, transport_opts} <- socket_opts, is_list(transport_opts) do
-      check_origin =
-        Keyword.get(transport_opts, :check_origin, endpoint_conf[:check_origin])
+      check_origin = Keyword.get(transport_opts, :check_origin, check_origin)
 
       check_csrf = transport_opts[:check_csrf]
 
--- a/lib/phoenix/endpoint/supervisor.ex
+++ b/lib/phoenix/endpoint/supervisor.ex
@@ -142,11 +142,7 @@ defmodule Phoenix.Endpoint.Supervisor do
 
     for {transport, transport_opts} <- socket_opts do
       check_origin =
-        if is_boolean(transport_opts[:check_origin]) do
-          transport_opts[:check_origin]
-        else
-          endpoint_check_origin
-        end
+        Keyword.get(transport_opts, :check_origin, endpoint_check_origin)
 
       check_csrf = transport_opts[:check_csrf]
 
--- a/lib/phoenix/endpoint/supervisor.ex
+++ b/lib/phoenix/endpoint/supervisor.ex
@@ -152,7 +152,7 @@ defmodule Phoenix.Endpoint.Supervisor do
 
       if check_origin == false and check_csrf == false do
         raise ArgumentError,
-              "One of :check_origin and :check_csrf must be set to non-false value for " <>
+              "one of :check_origin and :check_csrf must be set to non-false value for " <>
                 "transport #{inspect(transport)}"
       end
     end
--- a/lib/phoenix/endpoint/supervisor.ex
+++ b/lib/phoenix/endpoint/supervisor.ex
@@ -84,9 +84,9 @@ defmodule Phoenix.Endpoint.Supervisor do
     children =
       config_children(mod, secret_conf, default_conf) ++
         pubsub_children(mod, conf) ++
-        socket_children(mod, :child_spec) ++
+        socket_children(otp_app, mod, :child_spec) ++
         server_children(mod, conf, server?) ++
-        socket_children(mod, :drainer_spec) ++
+        socket_children(otp_app, mod, :drainer_spec) ++
         watcher_children(mod, conf, server?)
 
     Supervisor.init(children, strategy: :one_for_one)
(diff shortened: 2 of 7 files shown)


josevalim:
Hi @tanguilp, I believe we had previous discussions about this. Do you remember where they happened? I can't recall all of the details and linking back to them could help speed up the process. Meanwhile, thank you for the PR. :heart:

rhcarvalho:
More specific links:

- https://elixirforum.com/t/http-caching-liveviews-first-render/59180/16?u=rhcarvalho
- https://github.com/phoenixframework/phoenix/pull/5667#issuecomment-1880133966

(Full disclosure I'm not associated with the changes, but interested to learn from it and trying to help)

josevalim:
Thank you @rhcarvalho, the Elixir Forum was the additional discussion that I was missing. Let me get up to date.

tanguilp:
> my understanding is that we only require the CSRF tokens if they are sent by the client, no?

No, we always check them if the session is configured in the transport.

> So why not send the CSRF token from the client in the first place an option? I believe the answer is because LiveView then fails... but that would be something to address in LiveView, not here. :)

It's not failing in Liveview, but in Phoenix's socket transport. There was a problem with LiveView when no session was set but it turned out you can, well, just set a random value in the session.

They're two cases:
- `connect_info: [session: @session_options]` is **not** set → then no problem, caching works OOTB
- `connect_info: [session: @session_options]` is set → then [`Phoenix.Socket.Transport.connect_info/3`](https://github.com/tanguilp/phoenix/blob/main/lib/phoenix/socket/transport.ex#L462-L503) is called and CSRF checki is always performed. This PR makes this CSRF check it optional (enabled by default) so that caching can be enabled (and maybe for other use-cases - CSRF check is not needed as long as we have origin check enabled)

> I believe we had previous discussions about this. Do you remember where they happened? I can't recall all of the details and linking back to them could help speed up the process.

(Thanks @rhcarvalho for posting the links!)

Last time we [stopped](https://github.com/phoenixframework/phoenix/pull/5667) while discussing the following points:
- the need to allow / encourage granular caching → this is discussed in Part 1 of the blog post (`live "/", MainLive.Index, private: %{cache: true}` for example for opt-in caching)
- "We should document that you MUST have check_origin enabled if you want to use this feature." → I think this is sufficiently documented, although we could add a more visible warning
- "What I am saying is that it is granular per page. You don't disable session validation or change the CSWSH for the whole app, only on the pages you want to cache." → this is clearly not the direction taken by this PR. The rationale is that 1) the origin check does protect against CSWSH 2) disabling both checks by accident is hard unless you really don't read the doc. I think allowing to disable by route / page would make Phoenix's code more complex for a feature that might not have huge popularity
- by the way this option can only be configured at the transport level - not at the endpoint level
- there's still the problem of `live_session`'s `:session` parameter that adds some values (potentially private data!) into the HTML directly as discussed in Part 2 of the blog post. Not sure how to handle this point. Document it in the [option](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.Router.html#live_session/3-options)?
- Part 2 discusses how to avoid accidentally caching private user data (`assign_private` on the model of `assign_async`)

Hopefully it answers your questions! I think the core issue is that this PR allows developers to shoot in their feet if they disable both origin and CSRF check, which is not possible now. The alternatives being:
- bake in this feature deeper into Phoenix
- maybe a custom `Phoenix.Socket.Transport` could help? Unsure about it, seems like it would require copy/pasting current websocket & longpolling code without the CSRF check
- not allowing caching LiveViews to avoid possible critical security failures

Cheers!

josevalim:
> It's not failing in Liveview, but in Phoenix's socket transport. There was a problem with LiveView when no session was set but it turned out you can, well, just set a random value in the session.

Can you please explain how it is failing? Because, [looking at this code](https://github.com/phoenixframework/phoenix/blob/main/lib/phoenix/socket/transport.ex#L490-L502), we simply set the session to nil, which will then fail in LiveView, but not in Phoenix. In that case, I'd prefer to deal with this in LiveView.

tanguilp:
The point is that we *want* the session but *without* the CSRF check so as to implement the [Publicly caching private Liveviews](https://svground.fr/blog/posts/caching-liveviews-part-2/) pattern.

josevalim:
Doesn’t it mean then that we can cache user specific information, if we are not careful about it? So you need to write the page carefully to not expose any of that? How often is that used and why not cache parts of the page at the Phoenix layer instead of the HTTP layer?

josevalim:
In any case, this is good to me as long as we raise if both check_origin and check_csrf is false. You need to have at least one of them enabled. :)

tanguilp:
> Doesn’t it mean then that we can cache user specific information, if we are not careful about it? So you need to write the page carefully to not expose any of that?

Yes, but the risk is totally mitigated if the approach described in the article is used.

> How often is that used

Now nobody uses it but there's interest for sure!

> and why not cache parts of the page at the Phoenix layer instead of the HTTP layer?

It's just more complicated to reassemble the parts of the page, and it's not compatible with caching with external systems (including CDNs).

> In any case, this is good to me as long as we raise if both check_origin and check_csrf is false. You need to have at least one of them enabled. :)

Alright! Just need to check how to handle this in dev, where `check_origin` is set to `false`. Should we raise in dev as well when both checks are disabled?

josevalim:
We should raise in dev as well, yes!

tanguilp:
> We should raise in dev as well, yes!

Done in 5f2bf63ffed9af6dc2093975200f2b54902335a8 and d80a5352dda060768f4335edb6debbc30b45b20f

lib/phoenix/socket/transport.ex
```diff
@@ -332,8 +332,13 @@ defmodule Phoenix.Socket.Transport do
     import Plug.Conn
     origin = conn |> get_req_header("origin") |> List.first()
     check_origin = check_origin_config(handler, endpoint, opts)
+    check_csrf = opts[:check_csrf]
 
     cond do
+      check_origin == false and check_csrf == false ->
```
josevalim, line 338: Is there any chance we can detect and raise this when the endpoint is being configured, rather than at request time?

lib/phoenix/socket/transport.ex
```diff
@@ -458,8 +463,11 @@ defmodule Phoenix.Socket.Transport do
 
     * `:user_agent` - the value of the "user-agent" request header
 
+  The CSRF check can be disabled with the option `:check_csrf`. Beware: if
+  the `origin` header check is disabled as well, then your app is vulnerable
+  to Cross-Site WebSocket Hijacking (CSWSH) attacks.
   """
-  def connect_info(conn, endpoint, keys) do
+  def connect_info(conn, endpoint, keys, opts \\ [check_csrf: true]) do
```
josevalim, line 470:
```suggestion
  def connect_info(conn, endpoint, keys, opts \\ []) do
```

lib/phoenix/endpoint.ex
```diff
@@ -859,6 +859,10 @@ defmodule Phoenix.Endpoint do
       The MFA is invoked with the request `%URI{}` as the first argument,
       followed by arguments in the MFA list, and must return a boolean.
 
+    * `:check_csrf` - if the transport should perform CSRF check. Note that disabling
+    both CSRF and origin checks at the same time is not allowed and will raise.
+    Defaults to `true`
```
josevalim, line 864:
```suggestion
      both CSRF and origin checks at the same time is not allowed and will raise.
      Defaults to `true`
```

Schultzer:
First of all I like this is going to be configurable, I wanted to add this SO comment as a good overview of when to use CSRF token vs Origin headers https://stackoverflow.com/questions/24680302/csrf-protection-with-cors-origin-header-vs-csrf-token#24692474

Origin headers is not always are guaranteed to be there, but I don’t believe this is a concern at all, since we have all the control and know which routes and methods are allowed in the first place at compile time, so there is potential for improving security / education even further for a developer.

lib/phoenix/socket/transport.ex
tanguilp: Technically we could do it at compile time somewhere like in https://github.com/phoenixframework/phoenix/blob/main/lib/phoenix/endpoint.ex#L637

However, `check_origin` is a runtime configuration option so there's not really to ensure it will no be changed at runtime. We could also check it when starting the endpoint, but as far as I can see we don't do this kind of checks in the current code.

This also raises the question of the `check_origin: {m, f, a}` configuration: should we raise if the MFA call returns `false` and `check_csrf` is set to `false` as well (not currently done)?

josevalim: Can we check it when the supervision starts or the transports start? if it is a MFA, it is fine to not check, then the user is really up their own devices after manually bypassing all layers.

tanguilp: Done in d339626bd2e7be2198bda6bb34aeb7162453b227. Not sure it's the best place to do this, but I've pushed to discuss this :) There are no tests yet, hence I've set this PR to draft.

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -84,9 +84,9 @@ defmodule Phoenix.Endpoint.Supervisor do
     children =
       config_children(mod, secret_conf, default_conf) ++
         pubsub_children(mod, conf) ++
-        socket_children(mod, :child_spec) ++
+        socket_children(otp_app, mod, :child_spec) ++
```
josevalim, line 87: Instead of passing the `otp_app`, please pass the `conf` instead. This way, you don't have to read it again.

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -135,6 +137,27 @@ defmodule Phoenix.Endpoint.Supervisor do
...
+          transport_opts[:check_origin]
+        else
+          endpoint_check_origin
+        end
+
+      check_csrf = transport_opts[:check_csrf]
+
+      if check_origin == false and check_csrf == false do
+        raise ArgumentError,
+              "One of :check_origin and :check_csrf must be set to non-false value for " <>
```
josevalim, line 155:
```suggestion
              "one of :check_origin and :check_csrf must be set to non-false value for " <>
```

josevalim (commented):
I dropped one nitpick and we can ship it!

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -135,6 +137,27 @@ defmodule Phoenix.Endpoint.Supervisor do
...
+  defp check_origin_or_csrf_checked!(otp_app, endpoint, socket_opts) do
+    endpoint_check_origin = config(otp_app, endpoint)[:check_origin]
+
+    for {transport, transport_opts} <- socket_opts do
+      check_origin =
+        if is_boolean(transport_opts[:check_origin]) do
+          transport_opts[:check_origin]
+        else
+          endpoint_check_origin
+        end
```
josevalim, line 149:
```suggestion
        Keyword.get(transport_opts, :check_origin, endpoint_check_origin)
```

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -135,6 +137,27 @@ defmodule Phoenix.Endpoint.Supervisor do
...
+
+      check_csrf = transport_opts[:check_csrf]
+
+      if check_origin == false and check_csrf == false do
+        raise ArgumentError,
+              "One of :check_origin and :check_csrf must be set to non-false value for " <>
+                "transport #{inspect(transport)}"
+      end
```
josevalim, line 157: Can you please add a test? Look at endpoint_test and you should be able to verify that `Endpoint.start_link` fails if these are disabled.

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -118,10 +118,12 @@ defmodule Phoenix.Endpoint.Supervisor do
     end
   end
 
-  defp socket_children(endpoint, fun) do
+  defp socket_children(otp_app, endpoint, fun) do
     for {_, socket, opts} <- Enum.uniq_by(endpoint.__sockets__(), &elem(&1, 1)),
         spec = apply_or_ignore(socket, fun, [[endpoint: endpoint] ++ opts]),
```
josevalim, line 123: Since this is calling `start_link` in the transport, can't you move the check to the transport's `start_link` instead? This way we only check it for the transports that make sense.

josevalim: Ah, ignore me, the transport is not defined in Phoenix anymore. :)

lib/phoenix/endpoint/supervisor.ex
tanguilp: Thanks for your feedback! Done in https://github.com/phoenixframework/phoenix/pull/5952/commits/7018509e66a5b3e7bded506b46eebf53463e124b:
- in tests we check both when `check_origin` is set as endpoint option or transport option
- we do the security check even if the transport doesn't have a `child_spec/1`

lib/phoenix/endpoint/supervisor.ex
```diff
@@ -135,6 +135,21 @@ defmodule Phoenix.Endpoint.Supervisor do
     end
   end
 
+  defp check_origin_or_csrf_checked!(endpoint_conf, socket_opts) do
+    for {transport, transport_opts} <- socket_opts, is_list(transport_opts) do
+      check_origin =
+        Keyword.get(transport_opts, :check_origin, endpoint_conf[:check_origin])
```
josevalim, line 141:
```suggestion
    check_origin = endpoint_conf[:check_origin]

    for {transport, transport_opts} <- socket_opts, is_list(transport_opts) do
      check_origin = Keyword.get(transport_opts, :check_origin, check_origin)
```

lib/phoenix/endpoint.ex
```diff
@@ -859,6 +859,10 @@ defmodule Phoenix.Endpoint do
       The MFA is invoked with the request `%URI{}` as the first argument,
       followed by arguments in the MFA list, and must return a boolean.
 
+    * `:check_csrf` - if the transport should perform CSRF check. Note that disabling
+      both CSRF and origin checks at the same time is not allowed and will raise.
+      Defaults to `true`
```
josevalim, line 864:
```suggestion
    * `:check_csrf` - if the transport should perform CSRF check. To avoid
      "Cross-Site WebSocket Hijacking", you must have at least one of
      `check_origin` and `check_csrf` enabled. If you set both to `false`,
      Phoenix will raise, but it is still possible to disable both by passing
      a custom MFA to `check_origin`. In such cases, it is your responsibility
      to ensure at least one of them is enabled. Defaults to `true`
```

josevalim:

PR #5955, opened 2024-10-21
Implement mix listener to respect concurrent compilations

jonatanklosko:
This improves the code reloader to better handle concurrent compilations using Mix listeners API coming in Elixir v1.18.

The code reloader already handles concurrent compilation by checking the Elixir compiler manifest. If the manifest appears stale, we prune all modules from the given application. However, always pruning all modules can be expensive if the number of loaded modules is large.

This PR introduces a Mix listener that listens to concurrent compilations and keeps track of the changed modules. Then, before compiling, we prune only the modules that actually changed.

On Elixir < 1.18 we fall back to the current approach. On Elixir >= 1.18 we warn if the listener is not configured (and also use the fallback).

cc @josevalim


josevalim:

PR #5961, opened 2024-10-29
Auth links consistency on email confirmation pages.

ShPakvel:
Email confirmation pages can be open by logged in user. In such case we have following inconsistency.
- Below confirmation form we have "Register" and "Log in" links.
- While at the top menu we have "Settings" and "Log out".

This change fixing inconsistency by presenting links below confirmation form only if user is not logged in.
Note: I used `if !condition` because `unless` is going to be deprecated.


SteffenDE (approved):

josevalim:

PR #5964, opened 2024-11-04
docs: add missing console output

darricheng:
Hello! I came across missing docs at the following link: https://hexdocs.pm/phoenix/routing.html

Not sure if making this change will affect the docs I linked, because I'm new to the Elixir ecosystem. I would appreciate any pointing in the right direction.

### DIFF
--- a/guides/routing.md
+++ b/guides/routing.md
@@ -410,6 +410,33 @@ end
 
 You can run `mix phx.routes` to see how these definitions will look like.
 
+```console
+GET     /api/v1/images           HelloWeb.Api.V1.ImageController :index
+GET     /api/v1/images/:id/edit  HelloWeb.Api.V1.ImageController :edit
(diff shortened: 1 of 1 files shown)


glhrmv:
You added the text in the correct file, but I'm not sure this contribution is needed. Very shortly before the paragraph that precedes the line you started from, there is already a console output for the resource routes under the parent `/admin` route. I don't think the change mentioned in the docs warrants a whole new copy-paste of the console output, which is quite lengthy. Maybe a truncated version, but even then I feel it's pretty intuitive to read as it is.

darricheng:
@glhrmv That's a perspective that I didn't consider at all, and in thinking about it again I do agree with you guys. 

@SteffenDE Thank you for taking the time to review this PR! I understand the reasons for closing this, so it won't refrain me from potentially contributing more in the future 😃

PR #5965, opened 2024-11-06
Remove redundant validate_attr case.

ShPakvel:
- After change from [this PR](https://github.com/phoenixframework/phoenix/pull/4601/files#diff-4307581434ae66da4750cde438c2b192c4a11c8b71ba32dbf78aea36daa0ee3eR368), `:enum` valid case becomes redundant. Next line has general check that includes valid enum case.
- Also removed redundant passing throw for `:datetime` case. Instead, return already prepared result right away.

### DIFF
--- a/lib/mix/phoenix/schema.ex
+++ b/lib/mix/phoenix/schema.ex
@@ -425,7 +425,7 @@ defmodule Mix.Phoenix.Schema do
       mix phx.gen.schema Comment comments body:text status:enum:published:unpublished
   """
 
-  defp validate_attr!({name, :datetime}), do: validate_attr!({name, :naive_datetime})
+  defp validate_attr!({name, :datetime}), do: {name, :naive_datetime}
 
   defp validate_attr!({name, :array}) do
     Mix.raise("""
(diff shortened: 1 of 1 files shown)


lib/mix/phoenix/schema.ex
```diff
@@ -425,7 +425,7 @@ defmodule Mix.Phoenix.Schema do
       mix phx.gen.schema Comment comments body:text status:enum:published:unpublished
   """
 
-  defp validate_attr!({name, :datetime}), do: validate_attr!({name, :naive_datetime})
+  defp validate_attr!({name, :datetime}), do: {name, :naive_datetime}
```
SteffenDE, line 428: I think the intent with the datetime case may be to throw if (for whatever reason) the naive_datetime type was removed from the `@valid_types`.

SteffenDE (approved):

lib/mix/phoenix/schema.ex
ShPakvel: @SteffenDE, Thank you for the review! 🙇🏼 I thought the same. Though I am not fun of extra app code processing to catch possible future bugs. Tests are better approach for it. And conveniently there is test that will fail if somebody not fully clean up places with `naive_datetime` usage. This is why I think it is safe to remove extra protection in code.
[Existing test for example](https://github.com/phoenixframework/phoenix/blob/main/test/mix/tasks/phx.gen.schema_test.exs#L252)
```elixir
  test "generates schema with proper datetime types", config do
    in_tmp_project config.test, fn ->
      Gen.Schema.run(~w(Blog.Comment comments title:string drafted_at:datetime published_at:naive_datetime edited_at:utc_datetime locked_at:naive_datetime_usec))

      assert_file "lib/phoenix/blog/comment.ex", fn file ->
        assert file =~ "field :drafted_at, :naive_datetime"
        assert file =~ "field :published_at, :naive_datetime"
```

josevalim:

PR #5970, opened 2024-11-09
Add Oban generator to phx.new 

sorentwo:
This adds Oban by default to `phx.new` with the following considerations:

* It uses the [facade approach](https://hexdocs.pm/oban/Oban.html#__using__/1), e.g. `use Oban`, to generate a dedicated `MyApp.Oban` module for the application. This makes configuration in the application clearer and more familiar, because it matches `MyApp.Repo`. This would also allow multiple apps in an umbrella to have their own Oban instances without any name conflicts.
* It automatically selects the correct engine for `Postgres` or `SQLite3`. Creating a new project with other, currently unsupported, databases won't include Oban. This is documented in the task help.
* Configured with a single `default` queue, a pruning plugin set to retain jobs for one hour, and an empty crontab to indicate that periodic jobs are available.
* A default migration that will create the necessary oban tables.
* Configuration for testing as well as test helpers included in `DataCase`.

### Questions

* Is the facade (`MyApp.Oban`) approach alright with others? It's newer, and officially supported for several versions now, but it isn't what the installation docs prescribe.
* The migration is generated with the number `0`, rather than a timestamp just to get it working. Any thoughts or suggestions on how to get the correctly named file, or whether a different hard-coded number may work?
* This only adds minimal documentation, and it feels sparse, esp. without some concrete examples. Thoughts?
* There isn't support for the umbrella generator yet. I can add that once we've decided on a direction here.

### It Works!

Here's a screenshot after running `mix phx.new my_app; mix ecto.reset; iex -S mix phx.server` and checking that Oban is up and running:

Screenshot 2024-11-09 at 17 03 56


SteffenDE:
I am a big fan of Oban (also using Oban Pro at work), but I’m not sure how far we should go with including external libraries in the official generators, as those integrations need to be maintained. Maybe it would be better to improve extending the generators from external packages? There could be a `mix oban.phx.new` or `mix phx.new --plugin oban` (where the oban hex package would need to provide a specific module).

sorentwo:
After some reflection, and some outside advice, I don't think this is a good fit. Oban shouldn't be in every Phoenix app, especially not by default. Aside from the slippery slope of "what belongs in Phoenix by default", there's too much separate education necessary to use it and understand what it is doing.

Now that I've gone through the exercise of seeing all the touch points, I'll revisit a simplified igniter based installer instead.

sorentwo:
> Maybe it would be better to improve extending the generators from external packages? There could be a mix oban.phx.new or mix phx.new --plugin oban (where the oban hex package would need to provide a specific module).

That's a nice idea. Making it possibly to auto-install commonly used packages seems helpful.

PR #5979, opened 2024-11-19
Change snake_case auth URLs to kebab-case

### LINKED ISSUE #5917: URLs generated by `phx.gen.auth` should be in kebab-case and not snake_case
Hi everyone, keeping this short and simple: Phoenix's (amazing) `phx.gen.auth` generator results in changing the router like so:

```ex
# ...
live "/users/register", UserRegistrationLive, :new
live "/users/log_in", UserLoginLive, :new
live "/users/reset_password", UserForgotPasswordLive, :new
live "/users/reset_password/:token", UserResetPasswordLive, :edit
```

I was taken aback by the use of snake_case in the URL paths here. 

There is no authoritative standard on URL naming conventions, and I know the developer can very easily change these URLs manually, but we all know that kebab-case is far more widespread on the web, and I have reason to believe this was an oversight: every single "official" Elixir-related webpage (such as `elixir-lang.org`, `phoenixframework.org`, `elixirforum.com`, `hex.pm`, etc.) uses hyphen and not underscore as the word separator in multi-word URL paths. 

In the interest of maintaining consistency and cohesive alignment with the rest of the ecosystem (and the wider web), I believe we should change the casing of these URLs in the source (the generator). I can open a PR to address this, just let me know.

Additionally - and this is related but separate from what I just wrote above - I believe the `/log_in` and `/register` URLs should be changed to `/login` and `/signup`, respectively. Again, this aligns better with established practice in most popular websites and also with the rest of the Elixir ecosystem (e.g. `hex.pm`).

glhrmv:
Fixes #5917

### DIFF
--- a/lib/mix/tasks/phx.gen.auth/injector.ex
+++ b/lib/mix/tasks/phx.gen.auth/injector.ex
@@ -161,7 +161,7 @@ defmodule Mix.Tasks.Phx.Gen.Auth.Injector do
   Menu code to inject into the application layout template.
   """
   def app_layout_menu_code_to_inject(%Schema{} = schema, padding \\ 4, newline \\ "\n") do
-    already_injected_str = "#{schema.route_prefix}/log_in"
+    already_injected_str = "#{schema.route_prefix}/log-in"
 
     base_tailwind_classes = "text-[0.8125rem] leading-6 text-zinc-900"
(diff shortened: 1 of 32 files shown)


josevalim (approved):
LGTM! If CI passes, please merge!

josevalim:
CI is failing but for other reasons. But this looks good to me!

SteffenDE (approved):

PR #5990, opened 2024-11-29
WIP: Update tailwind to version 4.0.0

leifmetcalf:
This is waiting on the elixir library being updated to support tailwind v4 in https://github.com/phoenixframework/tailwind/pull/103.

Things I'm not sure about:
* Where the heroicons library should live. For the meantime I've put it in `assets/css/tailwind_heroicons.js`, but I'm not sure about the name and I don't think it belongs in `assets/css`. Also, for now, the tailwind standalone cli doesn't bundle the javascript library, so the heroicons plugin fails to load without running `npm install tailwindcss` or similar. See https://github.com/tailwindlabs/tailwindcss/issues/15235.


installer/templates/phx_assets/app.css
```diff
@@ -1,5 +1,14 @@
...
+
+@import "tailwindcss";
+@plugin "@tailwindcss/forms";
+@plugin "./tailwind_heroicons.js";
+@variant phx-click-loading ([".phx-click-loading&", ".phx-click-loading &"]);
+@variant phx-submit-loading ([".phx-submit-loading&", ".phx-submit-loading &"]);
+@variant phx-change-loading ([".phx-change-loading&", ".phx-change-loading &"]);
+@theme {
+  --color-brand: "#FD4F00",
+  }
```
josevalim, line 12:
```suggestion
}
```

josevalim: Also, should we add `;` for consistency?

installer/templates/phx_assets/tailwind_heroicons.js
```diff
@@ -0,0 +1,42 @@
...
+        "mask-repeat": "no-repeat",
+        "background-color": "currentColor",
+        "vertical-align": "middle",
+        "display": "inline-block",
+        "width": size,
+        "height": size
+      }
+    }
+  }, {values})
+}
```
aptinio, line 42:
```suggestion
})
```

SteffenDE:
@Flo0807 mentioned in the related issue that the plugin works if it is directly exported as function: https://github.com/phoenixframework/tailwind/pull/107#issuecomment-2629067324

I verified that it works, but I'm hesitant to rely on this. v4 doesn't really document how plugins should look like so that could change. I think if we wait for https://github.com/tailwindlabs/tailwindcss/pull/15934, we should be good!

installer/templates/phx_assets/app.css
leifmetcalf: Changed both, thanks

installer/templates/phx_assets/tailwind_heroicons.js
leifmetcalf: Done, thanks.

installer/templates/phx_assets/app.css
```diff
@@ -1,6 +1,15 @@
...
-@import "tailwindcss/utilities";
+/* See the Tailwind configuration guide for advanced usage
+   https://tailwindcss.com/docs/configuration */
+
+@import "tailwindcss";
+@plugin "@tailwindcss/forms";
+@plugin "./tailwind_heroicons.js";
+@variant phx-click-loading ([".phx-click-loading&", ".phx-click-loading &"]);
+@variant phx-submit-loading ([".phx-submit-loading&", ".phx-submit-loading &"]);
+@variant phx-change-loading ([".phx-change-loading&", ".phx-change-loading &"]);
```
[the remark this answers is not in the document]
leifmetcalf: Thank you for fixing this? Looking at the updated tailwind docs, we might also want to use `@custom-variant` instead of `@variant`? https://tailwindcss.com/docs/functions-and-directives#variant-directive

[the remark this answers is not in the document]
aptinio: I think so. Here's are more details about it:
https://tailwindcss.com/docs/hover-focus-and-other-states#registering-a-custom-variant

installer/templates/phx_assets/app.css
```diff
@@ -1,6 +1,15 @@
...
+@theme {
+  --color-brand: "#FD4F00",
+};
```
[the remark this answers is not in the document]
SteffenDE: sorry, false alarm. I still had a node_modules folder lying around in a parent directory. So the plugin is still broken, but we need to fix the semicolon nonetheless.

SteffenDE:
Thank you! @leifmetcalf

PR #5991, opened 2024-11-29
Add LiveViewTest.assert_* macros to locals_without_parens config

kelvinst:
So one can simply add `:phoenix` to their `:plugins` option in `.formatter.exs` and it will not automatically include parens to those functions when `mix format`ing

Original PR https://github.com/phoenixframework/phoenix_live_view/pull/3535

### DIFF
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -76,7 +76,17 @@ locals_without_parens = [
   on_mount: 1,
   slot: 1,
   slot: 2,
-  slot: 3
+  slot: 3,
+
+  # Phoenix.LiveViewTest
+  assert_patch: 2,
+  assert_patch: 3,
+  assert_patched: :2,
+  assert_push_event: 4,
+  assert_redirect: 2,
+  assert_redirect: 3,
+  assert_redirected: 2,
+  assert_reply: 3
 ]
 
 [
--- a/.formatter.exs
+++ b/.formatter.exs
@@ -79,14 +79,21 @@ locals_without_parens = [
   slot: 3,
 
   # Phoenix.LiveViewTest
+  assert_patch: 1,
   assert_patch: 2,
   assert_patch: 3,
-  assert_patched: :2,
+  assert_patched: 2,
+  assert_push_event: 3,
   assert_push_event: 4,
+  assert_redirect: 1,
   assert_redirect: 2,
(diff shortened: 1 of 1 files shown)


.formatter.exs
```diff
@@ -76,7 +76,17 @@ locals_without_parens = [
...
+
+  # Phoenix.LiveViewTest
+  assert_patch: 2,
+  assert_patch: 3,
+  assert_patched: :2,
+  assert_push_event: 4,
+  assert_redirect: 2,
+  assert_redirect: 3,
+  assert_redirected: 2,
+  assert_reply: 3
```
SteffenDE, line 89: We already have `assert_reply/3` in line 53. I don't know if duplicates are a problem though.

.formatter.exs
```diff
@@ -76,7 +76,17 @@ locals_without_parens = [
   on_mount: 1,
   slot: 1,
   slot: 2,
-  slot: 3
+  slot: 3,
+
+  # Phoenix.LiveViewTest
+  assert_patch: 2,
```
SteffenDE, line 82: Do we also need `assert_patch/1` and `assert_redirect/1`? (@josevalim)

The second parameter can be an optional timeout:

https://hexdocs.pm/phoenix_live_view/Phoenix.LiveViewTest.html#assert_patch/2
https://hexdocs.pm/phoenix_live_view/Phoenix.LiveViewTest.html#assert_redirect/2

```elixir
assert_patch view
```

(the same might apply with assert_push_event/3 so if we need the ones above, please double check)

kelvinst: Yep, we need to put them there, doing it!

.formatter.exs
kelvinst: As for duplicates, they shouldn't affect the code, as the function that checks them will work the same (it basically just `Enum.any?` this list here: https://github.com/elixir-lang/elixir/blob/44107f1010229f58d899177e34eb04357a075ab3/lib/elixir/lib/code/formatter.ex#L180-L185

IMO we should leave it here, just in case someday someone removes one of these sections.

kelvinst:
Not sure if the failing checks have something to do with my change. Anything else I need to do here?

SteffenDE (approved):

josevalim:

PR #5994, opened 2024-12-02
update dependencies

SteffenDE:
I only wanted to update makeup_diff at first to fix diff highlighting, but I did a `mix hex.outdated` and I think we can bump all the others as well.


chrismccord:
❤️❤️❤️🐥🔥

PR #6000, opened 2024-12-05
Include more information on AlreadySentError errors

Adzz:
We have an app that is getting a lot of AlreadySentErrors. Likely somewhere we have missed an `halt()`, but tracking that down is tricky. I wonder if we could supply some more information in the various places the error gets raised to help someone debug the issue? I have supplied a small example, but the exact wording and contents can be discussed.

I'm happy to roll this out to more places if we agree it's a good idea.

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -541,7 +541,16 @@ defmodule Phoenix.Controller do
     put_private_view(conn, :phoenix_view, :replace, formats)
   end
 
-  def put_view(%Plug.Conn{}, _module), do: raise(AlreadySentError)
+  def put_view(%Plug.Conn{} = conn, module) do
+    raise(AlreadySentError, """
+      The response was already sent.
+
+          Status code: `#{conn.status}`
+          Request path: `#{conn.request_path}`
+          Method: `#{conn.method}`
+          View module: `#{inspect(module)}`
+      """)
+  end
 
   defp put_private_view(conn, priv_key, kind, formats) when is_list(formats) do
     formats = Enum.into(formats, %{}, fn {format, value} -> {to_string(format), value} end)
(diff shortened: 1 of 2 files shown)


lib/phoenix/controller.ex
```diff
@@ -642,7 +642,9 @@ defmodule Phoenix.Controller do
     if state in @unsent do
       put_private_layout(conn, :phoenix_layout, :replace, layout)
     else
-      raise AlreadySentError
+      raise AlreadySentError, """
+      The response was already sent. Request was to `#{conn.request_path}` using #{inspect(layout)}\
```
josevalim, line 646: the request path should be in your logs/error handler anyway. Suggestion is to add information like the status code instead and response information.

josevalim: Ping :)

Adzz: Ah sorry, thanks for the reminder! I have pushed a thing see what you think :)

lib/phoenix/controller.ex
```diff
@@ -578,7 +587,16 @@ defmodule Phoenix.Controller do
...
 
-  def put_new_view(%Plug.Conn{}, _module), do: raise(AlreadySentError)
+  def put_new_view(%Plug.Conn{} = conn, module) do
+    raise(AlreadySentError, """
+      The response was already sent.
+
+          Status code: `#{conn.status}`
+          Request path: `#{conn.request_path}`
+          Method: `#{conn.method}`
+          View module: `#{inspect(module)}`
```
josevalim, line 597: Error messages start in lowercase and I think the backticks would make things harder to read, so I suggest removing them:

```suggestion
      the response was already sent.

          Status code: #{conn.status}
          Request path: #{conn.request_path}
          Method: #{conn.method}
          View module: #{inspect(module)}
```

josevalim (commented):
One comment and we should be good to go!

josevalim:

PR #6004, opened 2024-12-07
Format generators related files to eliminate future changes interference.

ShPakvel:
It is auto-formatting changes. Applied according to formatting configuration of this project. To eliminate interference in future specific logic changes.


ShPakvel:
@josevalim I gonna try to do fixes and improvements for generators gradually. For the first step/PR I did this `mix format` for related file. To eliminate unrelated changes from future changes. This should make easier following reviews and contributions.

josevalim:
One reason why we have postponed formatting files is because it can break the other open pull requests, make backporting hard, etc. I understand it is annoying though, the Phoenix Team has to deal with the same issues, but it is best to still avoid it for now. Thanks!

PR #6009, opened 2024-12-09
Fix typo

aptinio:


SteffenDE:
That’s not a typo but intentional, the umbrella root does not care about the exact LiveView version.

PR #6010, opened 2024-12-10
Relocate check for context files generation into context generator.

ShPakvel:
- Relocate check for context files generation into context generator.

- Invoke context files copy before parent generator files, to skip extra binding creation. It is not only skipping not used data passing in arguments, but also improve work with code as well (no need to double check if it is used data in context or not).

- Apply the same improvement to schema generation check.

- There were auto-formatting triggered on save for `phx.gen.schema.ex` file.


josevalim (approved):

rhcarvalho (commented):

I noticed that many of the hunks in the patch are related to `mix format` changes, probably unintentional/accidental while editing the files with an editor configured to auto-format on save.

José has asked to avoid this in https://github.com/phoenixframework/phoenix/pull/6015#issuecomment-2537485802.

Overall I don't see what the refactor here is gaining us (most likely the unrelated changes are not helping me see it, probably the meaningful patch should be more like ~2-4 hunks).

PR #6012, opened 2024-12-11
check for new phx.new version

SteffenDE:
Adds a warning message at the end of the phx.new output if a newer version of the installer is available:

image


installer/mix.exs
```diff
@@ -61,6 +61,7 @@ defmodule Phx.New.MixProject do
 
   def deps do
     [
+      {:hex_core, "~> 0.11.0"},
```
josevalim, line 64: Archives cannot include dependencies afaik :(

We will need to request against Hex API.

josevalim:
@SteffenDE should we check before we do anything, to make sure they are generating an app in the latest version?

installer/mix.exs
SteffenDE: reworked to use https://hex.pm/api/packages/phoenix instead. But as we need to parse JSON without a dependency, the whole check is gated behind an Elixir version check now.

SteffenDE:
> @SteffenDE should we check before we do anything, to make sure they are generating an app in the latest version?

I don't know who said it, maybe Chris, but initial feedback was that it should not block the task from proceeding.

josevalim (approved):

installer/lib/mix/tasks/phx.new.ex
```diff
@@ -404,4 +430,88 @@ defmodule Mix.Tasks.Phx.New do
...
+  end
+
+  # we need to parse JSON, so we only check for new versions on Elixir 1.18+
+  if Version.match?(System.version(), "~> 1.18") do
+    defp get_latest_version(package) do
+      Task.async(fn ->
+        # ignore any errors to not prevent the generators from running
+        # due to any issues while checking the version
+        try do
+          with {:ok, package} <- get_package(package) |> dbg do
```
josevalim, line 462:
```suggestion
          with {:ok, package} <- get_package(package) do
```

Issue #5685, opened 2023-12-29
Use mix format?


PragTob:
(this is neither a bug report nor a feature request so I was unsure where to put it and decided for here, if that's wrong please point me to where to best post it, thanks!)

I noticed the code base doesn't use `mix format` - as in, it's not configured for use in `.formatter.exs`, & code formatting isn't checked in CI. As one of the premiere elixir projects I think it'd be good for phoenix to use the formatter. That said, I'm of course blissfully unaware of any discussion that happened around this before and so am missing a lot of context - links would be appreciated :)

Running the formatter on lib and test (as is) results in: `65 files changed, 2558 insertions(+), 1641 deletions(-)` and it'd probably be pretty disruptive to the open PRs.

Thanks for all your work!

chrismccord:
We format as we go. Thanks!

PR #6015, opened 2024-12-12
Run mix format on generated files by tasks phx.gen.*

ShPakvel:
## Issue

There is `assert_passes_formatter_check(app_root_path)` in `integrated_test/` tests. Which was pseudo-truthy, because it passes only with specific short input for `phx.gen.*` tasks. Simply adding several more attributes for a `phx.gen.*` task run, we witnessing mentioned assertion to fail.

User-developer can pass any number of attributes or long names for modules. It is hard to achieve correctly formatted code in generated files. And even harder to supported it during changes.

This why I propose to apply `mix format` for `phx.gen.*` tasks.

## Solution

I implemented it with this considerations:

**Formatting is [recommended](https://hexdocs.pm/mix/Mix.Tasks.Format.html#module-when-to-format-code) and [implemented in the very basis of elixir](https://hexdocs.pm/elixir/Code.html#format_string!/2). But there are still projects which do not use it. And we need to respect that decision and provide mechanism to configure or even turn off formatting during files generation.**

Based on this, following conditions are reasonable for me:
- By default format is turned on for files with extensions `[".ex", ".exs", ".heex"]`.
- Run format once for all files related to generation and print notice in console about it, with instruction how to configure or turn it off.
- Format all files for just generated content: new and modified. If developers use formatting, then it is reasonable to do. If they do not use formatting, then they will turned it off in general.

Applied to live, html, json, context, schema, embedded. Can be applied to others later, if this will be approved.
@josevalim, @chrismccord, I believe this is pretty useful feature, that bring flexibility on future generators updates and decrease manual work for adjusting formatting complexity, simplifying code as well.

### Notes about implementation.

UPD: Auto-formatted code is reverted.
~~Because code formatting on Phoenix project is not fully applied and protected by CI, some files were auto-formatted on save. I expect it is ok based on [response to this issue](https://github.com/phoenixframework/phoenix/issues/5685#issuecomment-1873459840).~~

For quick introduction, here is the feature in essence.

Gathering files with generated content done this way (example from `phx.gen.html`):
```elixir
  defp format_files(%Context{} = context) do
    files_to_format(context) |> Mix.Phoenix.maybe_format()
    context
  end

  defp files_to_format(%Context{} = context) do
    (files_to_be_generated(context) |> Enum.map(fn {_, _, file} -> file end)) ++
      Gen.Context.files_to_format(context)
  end
```

It is invoked after copy new files and modify existing.
```diff
    |> copy_new_files(paths, binding)
+   |> format_files()
    |> print_shell_instructions()
```

And format executed with this code:
```elixir
  @default_format_extensions [".ex", ".exs", ".heex"]
  @doc """
  Conditionally run `mix format` for generated files.
  By default, unless `format_extensions` is set in generators config,
  files with extensions `#{inspect(@default_format_extensions)}` are formatted and
  override format instruction is printed into console.
  `format_extensions` can be set to `[]` to turn off formatting.
  If no files pass condition, formatting is not performed.
  """
  def maybe_format(files) when is_list(files) do
    config = Application.get_env(otp_app(), :generators, []) |> Keyword.get(:format_extensions)
    format_extensions = config || @default_format_extensions
    files = Enum.filter(files, &String.ends_with?(&1, format_extensions))

    if files != [] do
      Mix.Task.run("format", files)
      if !config, do: Mix.shell().info(override_format_instruction())
    end
  end

  @doc """
  Override format instruction for console and docs.
  """
  def override_format_instruction do
    """

    Files with generated content (new and modified) are formatted with `mix format`.

    By default files with extensions `#{inspect(@default_format_extensions)}` are formatted.
    List of extensions can be changed via generators config:

        config :your_app, :generators,
          format_extensions: [".ex", ".exs"]

    Formatting can be turned off by setting empty list:

        config :your_app, :generators,
          format_extensions: []
    """
  end
```

For `integration_test/` I added extra attributes in `mix_run!` for `phx.gen.*` tasks, to truthfully test formatting. There is issue for mysql case (some files from `phx.new`is not formatted). So for mysql I left TODO for future separate investigation, as it is not related to `phx.gen.*` files formatting.

For `test/mix/tasks/` there is no deps `[:phoenix, :ecto, :ecto_sql]` we can format code according to. So for now I disabled formatting via new config `with_generator_env([format_extensions: []], function)` applied in `in_tmp_project` and `in_tmp_umbrella_project`. The `with_generator_env` function itself has minor update as well, to merge passing generators config instead of fully replacing it.
It is pretty simple and generally applied change I came up with as first idea. Let me know if you want to resolve it in some better way.

---

Small improvements that were done in the same gen files:
- Simplify conditions for context and schema files by relocating logic from live, html, json into context and schema itself, where it belongs.
- Invoke context files copy before parent generator files, to skip extra binding creation. It is not only skipping unused data passing in arguments, but also improve work with code (no need to double check if it is used in context or not).

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ erl_crash.dump
 phoenix-*.ez
 
 .DS_Store
+
+/.vscode/
--- a/.gitignore
+++ b/.gitignore
@@ -21,5 +21,3 @@ erl_crash.dump
 phoenix-*.ez
 
 .DS_Store
-
-/.vscode/
--- a/lib/mix/phoenix.ex
+++ b/lib/mix/phoenix.ex
@@ -68,21 +68,17 @@ defmodule Mix.Phoenix do
   @default_format_extensions [".ex", ".exs", ".heex"]
   @doc """
   Conditionally run `mix format` for generated files.
-  By default, unless `format_extensions` is set in generators config,
-  files with extensions `#{inspect(@default_format_extensions)}` are formatted and
-  override format instruction is printed into console.
-  `format_extensions` can be set to `[]` to turn off formatting.
+  By default files with extensions `#{inspect(@default_format_extensions)}` are formatted.
+  It can be adjusted with generators config `format_extensions`, and turned off with value `[]`.
   If no files pass condition, formatting is not performed.
   """
   def maybe_format(files) when is_list(files) do
-    config = Application.get_env(otp_app(), :generators, []) |> Keyword.get(:format_extensions)
-    format_extensions = config || @default_format_extensions
-    files = Enum.filter(files, &String.ends_with?(&1, format_extensions))
+    format_extensions =
+      Application.get_env(otp_app(), :generators, [])
+      |> Keyword.get(:format_extensions, @default_format_extensions)
 
-    if files != [] do
-      Mix.Task.run("format", files)
-      if !config, do: Mix.shell().info(override_format_instruction())
-    end
+    files = Enum.filter(files, &String.ends_with?(&1, format_extensions))
+    if files != [], do: Mix.Task.run("format", files)
   end
 
   @doc """
--- a/lib/mix/phoenix.ex
+++ b/lib/mix/phoenix.ex
@@ -65,6 +65,47 @@ defmodule Mix.Phoenix do
   defp to_app_source(app, source_dir) when is_atom(app),
     do: Application.app_dir(app, source_dir)
 
+  @default_format_extensions [".ex", ".exs", ".heex"]
+  @doc """
+  Conditionally run `mix format` for generated files.
+  By default, unless `format_extensions` is set in generators config,
+  files with extensions `#{inspect(@default_format_extensions)}` are formatted and
+  override format instruction is printed into console.
+  `format_extensions` can be set to `[]` to turn off formatting.
+  If no files pass condition, formatting is not performed.
+  """
+  def maybe_format(files) when is_list(files) do
+    config = Application.get_env(otp_app(), :generators, []) |> Keyword.get(:format_extensions)
+    format_extensions = config || @default_format_extensions
+    files = Enum.filter(files, &String.ends_with?(&1, format_extensions))
+
+    if files != [] do
+      Mix.Task.run("format", files)
+      if !config, do: Mix.shell().info(override_format_instruction())
+    end
+  end
+
+  @doc """
+  Override format instruction for console and docs.
+  """
+  def override_format_instruction do
+    """
+
+    Files with generated content (new and modified) are formatted with `mix format`.
+
+    By default files with extensions `#{inspect(@default_format_extensions)}` are formatted.
+    List of extensions can be changed via generators config:
+
+        config :your_app, :generators,
+          format_extensions: [".ex", ".exs"]
+
+    Formatting can be turned off by setting empty list:
+
+        config :your_app, :generators,
+          format_extensions: []
+    """
+  end
+
   @doc """
   Inflects path, scope, alias and more from the given name.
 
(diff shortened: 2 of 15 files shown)


josevalim:
Hi @ShPakvel! Thank you for the PR. As I mentioned in the other issue, please don’t format the files. It makes our work in reviewing the PRs much harder. Generally speaking, a PR should only do one thing. :)

ShPakvel:
@josevalim Auto-formatting is reverted. I disabled formatting locally for phoenix project. Auto-formatting won't happen again from my side.
CI has one random fail, not related to changes (the same commit CI has passed).

ShPakvel:
CI failed one job with random reason. Not related to changes. The same CI passed on referenced branch in my fork.
image

.gitignore
```diff
@@ -21,3 +21,5 @@ erl_crash.dump
 phoenix-*.ez
 
 .DS_Store
+
+/.vscode/
```
josevalim, line 25: These changes should go to your global gitignore: https://sebastiandedeyne.com/setting-up-a-global-gitignore-file/

Allegedly the .DS_Store one too :D

lib/mix/phoenix.ex
```diff
@@ -65,6 +65,47 @@ defmodule Mix.Phoenix do
...
+  If no files pass condition, formatting is not performed.
+  """
+  def maybe_format(files) when is_list(files) do
+    config = Application.get_env(otp_app(), :generators, []) |> Keyword.get(:format_extensions)
+    format_extensions = config || @default_format_extensions
+    files = Enum.filter(files, &String.ends_with?(&1, format_extensions))
+
+    if files != [] do
+      Mix.Task.run("format", files)
+      if !config, do: Mix.shell().info(override_format_instruction())
```
josevalim, line 84: We don't need to print this every time we run. Users will read the docs if they need it. Especially because we only format the files we only generate.

ShPakvel: Sure, deleted console notification.

ShPakvel:
Addressed review comments. ✅

lib/mix/phoenix.ex
```diff
@@ -65,6 +65,43 @@ defmodule Mix.Phoenix do
   defp to_app_source(app, source_dir) when is_atom(app),
     do: Application.app_dir(app, source_dir)
 
+  @default_format_extensions [".ex", ".exs", ".heex"]
```
rhcarvalho, line 68: I would respect the target project's `.formatter.exs` configuration if possible, similar to how running `mix format` in the project context would behave.

That would mean we don't need additional config in `:generators`. Note that projects enforcing `mix format --check-formatted` in CI would need to keep the two configs in sync or manually run `mix format` after generating code.

For new projects that means respecting 

https://github.com/phoenixframework/phoenix/blob/7d49cdd50273461a9d8325d66c5ee6c17061e7a1/installer/templates/phx_single/formatter.exs.eex#L5

rhcarvalho (commented):

I came to this thinking Phoenix v1.9 would be a good time to introduce such a change, and that in theory it should be a small addition to the generators pipeline, plus a `--format` / `--no-format` arg to related Mix tasks.

On second thought, though, the very same result could be achieved on userland by simply running `mix format` explicitly:

```
mix phx.* [...]; mix format
```

Mix even supports it as a single invocation:

```
mix do phx.gen.presence + format
```

In fact, **a fine alternative to `--format/--no-format` flags in all generators might be adding a line like above to the help output** of all generators.

Also considering that when coding with modern AI agents they will typically run `mix format` anyway (to cover arbitrary edits), having the generators produce always perfectly formatted code (in the face of long names, etc) seems less useful to me today.

Not formatting files by default also plays nicer with existing code bases, leaving no room for existing users to complain of a default behavior change triggering unexpected code churn on their projects.

PR #6018, opened 2024-12-16
Prevent multiple emails separated by commas in email field

tsubery:
I know the conventional wisdom is there's no reason to over complicate email validation because it is implicitly validated by  sending a confirmation email.
The current regex accepts multiple emails separated by commas or semicolons. For example: 
- `email1@example.com,email2@example.com`
- `email1@example.com;email2@example.com`

Some email servers accept a comma or semicolon seperated list of emails as valid and would send the content to each recipient. For example, [Mailgun](https://mailgun-docs.redoc.ly/docs/mailgun/api-reference/openapi-final/tag/Messages/). Documentation mentions you can use comma to separate multiple recipients. (I haven't tested it)

That means the implicit validation of the email is not enough. It could allows user with one email to sign up multiple times by using a suffix of random emails or enable abuse by submitting registrations with more than 20 emails at once.

### DIFF
--- a/priv/templates/phx.gen.auth/schema.ex
+++ b/priv/templates/phx.gen.auth/schema.ex
@@ -46,7 +46,9 @@ defmodule <%= inspect schema.module %> do
   defp validate_email(changeset, opts) do
     changeset
     |> validate_required([:email])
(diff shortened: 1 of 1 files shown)


josevalim:

PR #6027, opened 2024-12-21
Update longpoll.js

yuvashrikarunakaran:
Adding feedback to the LongPoll implementation ensures better observability, easier debugging, and improved developer experience. Here’s an updated version with feedback mechanisms included


josevalim:
Please keep changes to a minimum, avoid changing only styles. Thank you.

PR #6028, opened 2024-12-21
Update presence.js

yuvashrikarunakaran:
Here's the updated syncState and syncDiff methods, ensuring proper metadata is included in the onJoin callback


josevalim:
Can you please explain your changes in detail? It is discarding comments and other important into.

PR #6029, opened 2024-12-21
Update push.js

yuvashrikarunakaran:
Here’s the updated Push class with an ensured newline at the end of the file for better formatting and adherence to coding standards


josevalim:
We don't accept format changes, as it messes up the git history and make order PRs harder to merge. Thank you.

PR #6032, opened 2024-12-30
Require Elixir v1.15 for Phoenix

josevalim:
v1.13 and earlier are no longer supported and, given we will probably stay on v1.8 for a while, it makes sense for us to push to at least v1.15.


SteffenDE:
We should bump the `@elixir_requirement` in the mix.exs as well, or did you keep it intentionally?

PR #6035, opened 2025-01-01
Update LICENSE.md, fix copyright license year

JasonnnW3000:
Yearly copyright bump. Padding my changed LOC! 😛


SteffenDE:
As with your other PR, we did never bump those yet, so I think we’re fine as is. Thanks!

PR #6039, opened 2025-01-07
Dockerfile: use qualified image names

leifmetcalf:
Before this change, Podman would fail to build the image as it doesn't use dockerhub as the image repository for unqualified names.


SteffenDE:
Thank you! 🙌🏻

Issue #5759, opened 2024-03-22
`Phoenix.Presence.get_by_key/2` typespec, return values, and documentation are inconsistent


jaminthorns:
### Environment

* Elixir version (elixir -v): 1.16.1
* Phoenix version (mix deps): 1.7.11
* Operating system: macOS 14.4 (23E214)

### Actual behavior

Issue #4514 was never properly addressed. The commit 3c7f494d7f7071becd1a14b011bbe3584fcdf2b1 says that it addressed the issue, but it only updated the typespec of `get_by_key/2` to something that is still inconsistent with the implementation and the example.

### Expected behavior

I'll defer to the referenced issue for full explanation and possible solutions, since the author did a great job of breaking down the issue, and I agree with their proposed solutions.

chrismccord:
If you think the docs are unclear, please send a PR, but this looks correct to me. The value returned matches the shape of `presence()` (note singular presence), and the function returns a list of those for that given id. Thanks!

chrismccord:
Sorry, you're right the example just wrong :)

Issue #4514, opened 2021-09-29
Phoenix.Presence.get_by_key/2 return values and documentation are inconsistent


sokolnickim:
## Environment
Elixir: 1.12
Phoenix: 1.5.8, it looks like it's applicable to all versions since the function was introduced.

## Problem
According to documentation, `get_by_key/2` in Phoenix.Presence should return "the map of presence metadata for a socket/topic-key pair", and the further Examples section says it "Uses the same data format as `list/1`. This would be congruent with the specs, which state both `get_by_key/2` and `list/1` return the type `presences()`, defined as:

```elixir
presences() :: %{required(String.t()) => %{metas: [map()]}}
```

On the other hand, the actual code in the example shows the function returning a list:

```elixir
iex> MyPresence.get_by_key("room:1", "user1")
[%{name: "User 1", metas: [%{device: "Desktop"}, %{device: "Mobile"}]}]
```

Which looks more like what we would get if we took the `presences()` format and applied `Map.values/1` to it.

Finally, the actual behavior of the function seems to be a cross between both of these conflicting formats. For existing entries, it returns the map containing the `metas` key, in other words what the example code shows, without the surrounding list (which is superfluous anyway, since there will only ever be one entry for one user). For missing entries, it returns the empty list.

Among all these conflicting sources of information, it isn't clear what should even be the correct behavior, that's why I didn't even attempt to create a pull request. I can see three reasonable solutions.

## Solutions

The first is to be consistent with `list/1` and to follow the spec. The example code should then read `%{"123" => %{name: "User 1", metas: [%{device: "Desktop"}, %{device: "Mobile"}]}}`, and non-existent entries should return an empty map. The drawbacks though are that it returns useless data (the outer map), empty maps are unpleasant to use in pattern matches and most importantly, it would break all existing code, since all returned values would change.

The second would be to stick to the current behavior and rewrite the spec and docs. The only quirk here is that an empty list is still returned for non-existent entries, which would have to be documented and justified. It also has the disadvantage of being a truthy value, so if all we are interested in is whether the user is present, we can't do `if get_by_key(topic, id) ...`.

The third solution would be my favorite, and it would be to just change the return for non-existent entries from an empty list to a `nil`. That brings it in line with the convention of the rest of the language and should bring minimum breakage. Both docs and specs would need to be changed, but we should gain clarity in the process.

PR #6042, opened 2025-01-08
Improve Phoenix.Presence.get_by_key/2 documentation, return value, and typespec.

jaspervanbrian:
## Issue
#5759 states that issue #4514 wasn't addressed properly. Checking the [docs for `Phoenix.Presence.get_by_key/2`](https://hexdocs.pm/phoenix/Phoenix.Presence.html#c:get_by_key/2), it does have inconsistencies between its return value and the typespec it's being referred to.

## Improvements and Changes
Going by the given problem and proposed solution from #4514, I have made the changes:

- Improve and fix the `presence` typespec that returns a map with the `:metas`.
- If no `presence` found under the supplied `key` in `socket_or_topic` scope, return `nil`. Pros of having a `nil` return value instead of `[]` is that it's easier to evaluate not being a truthy value.
- Change the documented return value of `Phoenix.Presence.get_by_key/2` to return a map instead of a list that contains the `metas` information, along with some additional info that might be returned on the `Phoenix.Presence.fetch/2` callback.
- Change test specs on `test/phoenix/presence_test.exs` to check with `is_nil` instead of `[]` if the returned value on untracked `key` is empty.

### DIFF
--- a/lib/phoenix/presence.ex
+++ b/lib/phoenix/presence.ex
@@ -189,8 +189,8 @@ defmodule Phoenix.Presence do
 
   """
 
-  @type presences :: %{String.t() => %{metas: [map()]}}
-  @type presence :: %{key: String.t(), meta: map()}
+  @type presence :: %{metas: [map()]}
+  @type presences :: %{String.t() => presence}
   @type topic :: String.t()
 
   @doc """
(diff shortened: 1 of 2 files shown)


bkilshaw:
Just ran into this; it feels bit odd to get back either a map or an empty list, 'nil` feels like a good solution.

SteffenDE:
Changing the return value is a breaking change, so I'm not sure we can do this (cc @josevalim).

PR #6043, opened 2025-01-09
Encode SVG icons data

### LINKED ISSUE #6038: Use encodeURIComponent instead of replace in tailwind config
Is there any reason to use replace instead of encodeURIComponent for svg url in tailwind config?

https://github.com/phoenixframework/phoenix/blob/06e3525edf9bb5918785ce929433761c11935d70/installer/templates/phx_assets/tailwind.config.js#L52

Isn't 

```js
                        let content = fs.readFileSync(fullPath).toString()
                        content = encodeURIComponent(content)
```

Better?

kuon:
Ensure SVG icon data is properly encoded as data URI, fixes #6038.


SteffenDE:
Thank you! 🙌🏻

PR #6051, opened 2025-01-13
Update topbar.js to v3.0.0

rayere:
Changes:
- Removes polyfill for requestAnimationFrame as it's no longer needed for modern browsers
- Improves DOM manipulation by only appending canvas when showing
- Updates copyright year and version numbers


SteffenDE (approved):

SteffenDE:
Thank you! 🙌🏻

PR #6056, opened 2025-01-17
fix: ReferenceError: Cannot access 'global' before initialization

satoren:
Here, global refers to an uninitialized global, so if the error is ignored and executed, it will result in undefined.


SteffenDE:
Thank you! 🙌🏻

PR #6062, opened 2025-01-23
Improve `:check_origin` docs.

leandrocp:
Realized that passing only a host to `:check_origin` is not enough as the doc suggest, since it might also check scheme, host, and port against a URI.

So this was not really working: `check_origin: ["my-app.fly.dev"]`
I had to include the scheme as well: `check_origin: ["https://my-app.fly.dev"]`

The examples are clear but the rest of the doc is a bit misleading.

See https://github.com/phoenixframework/phoenix/blob/5dba0bb6216e26fe5c48aafebad64d5f96da8868/lib/phoenix/socket/transport.ex#L616-L618

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -828,7 +828,7 @@ defmodule Phoenix.Endpoint do
       if so, the level
 
     * `:check_origin` - if the transport should check the origin of requests when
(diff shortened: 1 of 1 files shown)


SteffenDE:
Thank you! 🙌🏻

PR #6065, opened 2025-01-27
Make title optional in generated flash component

SteffenDE:
Use sky as color to be more info-like for type :info.

Relates to: #5969

image

Multi-line messages
image

Info
image

I increased the icon size to match the text's line-height, otherwise it would not look aligned:

image

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -29,6 +29,7 @@ defmodule <%= @web_namespace %>.CoreComponents do
   """
   attr :id, :string, doc: "the optional id of flash container"
   attr :flash, :map, default: %{}, doc: "the map of flash messages to display"
+  attr :title, :string, default: nil
   attr :kind, :atom, values: [:info, :error], doc: "used for styling and flash lookup"
   attr :rest, :global, doc: "the arbitrary HTML attributes to add to the flash container"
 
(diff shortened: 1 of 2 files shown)


priv/templates/phx.gen.live/core_components.ex
```diff
@@ -77,12 +76,11 @@ defmodule <%= @web_namespace %>.CoreComponents do
...
     ~H"""
     <div id={@id} aria-live="polite">
-      <.flash kind={:info} title=<%= maybe_heex_attr_gettext.("Success!", @gettext) %> flash={@flash} />
-      <.flash kind={:error} title=<%= maybe_heex_attr_gettext.("Error!", @gettext) %> flash={@flash} />
+      <.flash kind={:info} flash={@flash} />
+      <.flash kind={:error} flash={@flash} />
       <.flash
         id="client-error"
         kind={:error}
-        title=<%= maybe_heex_attr_gettext.("We can't find the internet", @gettext) %>
```
SteffenDE, line 85: well maybe we do still want the title for those? @josevalim

SteffenDE: or we move the "Something went wrong!" part into the message

installer/templates/phx_web/components/core_components.ex
```diff
@@ -77,12 +76,11 @@ defmodule <%= @web_namespace %>.CoreComponents do
...
     ~H"""
     <div id={@id} aria-live="polite">
-      <.flash kind={:info} title=<%= maybe_heex_attr_gettext.("Success!", @gettext) %> flash={@flash} />
-      <.flash kind={:error} title=<%= maybe_heex_attr_gettext.("Error!", @gettext) %> flash={@flash} />
+      <.flash kind={:info} flash={@flash} />
+      <.flash kind={:error} flash={@flash} />
       <.flash
         id="client-error"
         kind={:error}
-        title=<%= maybe_heex_attr_gettext.("We can't find the internet", @gettext) %>
```
josevalim, line 85: Yes, I agree about moving this to the message. Feel free to even use bold or something within the message below?

installer/templates/phx_web/components/core_components.ex
```diff
@@ -94,7 +92,6 @@ defmodule <%= @web_namespace %>.CoreComponents do
       <.flash
         id="server-error"
         kind={:error}
-        title=<%= maybe_heex_attr_gettext.("Something went wrong!", @gettext) %>
```
josevalim, line 97: Same for this one.

SteffenDE:
image
image

SteffenDE:
image

Slightly less bold

josevalim:
Looks good! Do we want to add some breaks/vertical spacing? Maybe we render the same as before but without forcing a title on everyone? Alternatively we can make the title optional inside `flash` itself. Thoughts?

josevalim:
Nah, ignore me. No need for spacing. Ship it!

josevalim:
Yeah, whatever is less code then. :)

josevalim (approved):

PR #6067, opened 2025-01-31
Use :public_key.cacerts_get, closes #6066

### LINKED ISSUE #6066: CA Certificate Resolution in mix phx.gen.release --docker doesn't support custom CA Certs
I will be opening a PR but wanted to open an issue and make sure I can do this at all.

Referencing Jose's PR here - https://github.com/erlang/otp/pull/8920 about CA certs, we need to update these http calls to respect users who need to provide custom CA certs like those behind corporate SSL inspection. 

I'll be updating this to use the underlying erlang method of loading certificates if that's ok?

josevalim:


SteffenDE:
@josevalim what about
https://github.com/phoenixframework/phoenix/blob/2614f2a0d95a3b4b745bdf88ccd9f3b7f6d5966a/installer/lib/mix/tasks/phx.new.ex#L220-L221

josevalim:
That's need for esbuild and tailwind projects, as they depend on castore still for now.

SteffenDE:
Thank you! 🙌🏻

PR #6072, opened 2025-02-05
validate options passed to socket longpoll / websocket options

SteffenDE:
We cannot validate options other than websocket / longpoll, because they might be used by custom sockets. See https://github.com/phoenixframework/phoenix/commit/ab2ff9ede9f20720412f7e26f9789c86fbe7ec3d#diff-5ab73ffaee87d2c9de2a1455f11157d5d3c0dd01157ee019d58d37b19dafd68eR29.

@josevalim please double check if validating the websocket / longpoll keys is actually fine.

### DIFF
--- a/lib/phoenix/endpoint.ex
+++ b/lib/phoenix/endpoint.ex
@@ -690,8 +690,43 @@ defmodule Phoenix.Endpoint do
 
   defp socket_paths(endpoint, path, socket, opts) do
     paths = []
-    websocket = Keyword.get(opts, :websocket, true)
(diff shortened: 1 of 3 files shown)


josevalim (approved):
LGTM but let's not backport this one, as it may crash!

giddie:
Just wondering whether it was intentional to include only a subset of the websock_adapter opts in the list of keys? In particular we're relying on `deflate_options` to reduce scheduler usage.

edgurgel:
https://github.com/phoenixframework/phoenix/pull/6758 Opened this one to add missing config keys that websock_adapter supports

PR #6081, opened 2025-02-11
Update phx.gen.auth with sudo mode and magic links

### LINKED ISSUE #6041: Update phx.gen.auth with sudo mode and magic links
Today phx.gen.auth generates the following functionality:

1. Registration
2. Login / logout
3. Confirmation
4. Forgot password
5. Change email/password

I believe we can improve both the user experience and reduce the amount of generated code with a few changes. Behind the scenes, forgot password already behaves as magic links with extra steps. My proposal is:

1. Introduce magic links as an official mechanism and remove "Forgot password"
2. For users who forget their password, they can use a magic link and then change their password in Settings
3. In order to keep changing email and passwords safe, we will introduce "sudo mode" that expects users to have authenticated within the last 15 minutes

This keeps the threat model roughly the same given anyone with access to the user email _today_ is already capable of resetting and changing the user password. Anyone with access to the email can also change the email used in the account too. 

Furthermore, requiring the user to enter "sudo mode" upfront actually improves the user experience as changes to email/password will immediately take effect instead of requiring an email confirmation _later_, which the user may forget. It also simplifies the code considerably and provides a more reusable functionality (as it can be used to authenticate other features as done by GitHub).

My only question is if we should change the confirmation feature to be implemented on top of magic links too. Today, confirming an account does not sign the user in, as we want to avoid leaked tokens from giving user access. However, password recovery (and magic links) give a leaked token access. There are two possible arguments here:

1. Keep confirmation separate, as such emails are sent by default after sign in, and theoretically certain users may not even use password recovery or magic links (therefore they would never be vulnerable to token exposure)

2. Realistically most users will use magic links and password recovery. Furthermore a user who is being targeted will be vulnerable to those attacks anyway. Therefore there is little benefit in not giving users access after confirmation

In any case, using a magic link should automatically confirm the account anyway. Thoughts?

SteffenDE:
Closes #6041.

If you want to test this:

```bash
git clone https://github.com/phoenixframework/phoenix.git
cd phoenix
git checkout sd-phx-gen-auth-magic
mix deps.get
cd installer
mix phx.new auth_app --dev --database sqlite3
cd auth_app
mix phx.gen.auth Accounts User users
mix deps.get
mix ecto.migrate
mix phx.server
# visit http://localhost:4000/users/register
```


lib/mix/tasks/phx.gen.auth.ex
```diff
@@ -26,6 +26,32 @@ defmodule Mix.Tasks.Phx.Gen.Auth do
...
+
+  This is why the default implementation raises whenever a user tries to log in by magic link
+  when there is already a password set. You can safely remove this check if you do not allow
+  unconfirmed users to log in. In that case, registering with email and password is still secure.
+
+  If you want to allow unconfirmed users to log in, you need to ensure that they need to reset
+  their password when they first sign in by magic link. To do this, you can add a required password
+  field to the confirmation LiveView / template and re-use the password change functionality from
+  the settings page.
+
```
josevalim, line 54: Notes:

1. We don't even need to log them in after their password, as long as the attacker knows the password, the attack is enough is enough

2. We already avoid session fixation by clearing up all sessions

Todos:

1. We should not allow someone to set a password if the account is not confirmed yet.
   This way someone can choose to login after sign up if they want to (although still discouraged)

```suggestion
  `mix phx.gen.auth` generates email based authentication, which assumes the user who
  owns the email address has control over the account. Therefore, it is extremely
  important to void all accesses once the user confirms their account for the first
  time, to avoid credential stuffing. This is done by revoking all tokens on
  confirmation.
  
  However, if you allow users to create an account with password, you must also
  require them to be logged in by the time of confirmation, otherwise the following
  attack is possible:

  1. An attacker registers a new account with the email address of their target, anticipating
     that the target creates an account at a later point in time.
  2. The attacker sets a password when registering.
  3. The target registers an account and sees that their email address is already in use.
  4. The target logs in by magic link, but does not change the existing password.
  5. The attacker maintains access using the password they previously set.

  This is why the default implementation raises whenever a user tries to log in for the first
  time by magic link and there is already a password set. If you add registration with email
  and password, then you must require the user to be logged in to confirm their account.
  If they don't have a password (because it was set by the attacker), then they can set one
  via a "Forgot your password?"-like workflow.

```

josevalim (approved):

Issue #5778, opened 2024-04-11
Current approach to passing token in channels is unsafe


mspanc:
### Actual behavior

As of Phoenix 1.7 the [recommended approach](https://hexdocs.pm/phoenix/channels.html#step-3-pass-the-token-to-the-socket-constructor-and-verify) to passing tokens while authenticating in channels results in appending the token to the request URL.

Given that such a request is GET in many server environments such URL, including the token will be exposed in the logs. Even if the app itself won't log the token, it can be done on some ingress/routing layer that is handling incoming requests. This is clearly leading to the situation where access tokens can be easily leaked. It can be used to impersonating users by some malicious actor who will get access to the logs.
 
In the ideal world, the access to the logs is very limited to the trusted personnel, in practice you often need to give wide access to many developers for sake of debugging. Moreover, the logs are often passed to some third party log processing services where you lose control over security and can be leaked.
  
### Expected behavior

The authentication should happen in the WebSocket layer. For example, the initial connection params should be sent as a first message after establishing connection. If such message does not happen within certain time window, connection should be dropped. From the API perspective it can appear to be identical synchronous `connect` callback identical to the current one.

I understand that requires modyfying the protocol, but current approach is unsafe.

Exadra37:
As someone who has worked previously on API security, I can confirm that this approach of using a token in a URL is unsafe, and its use may induce others to consider it a best practice.

I would recommend addressing this security issue as soon as possible, despite it not being a high-severity one.

@chrismccord @josevalim Can I help with addressing this issue?

josevalim:
I don’t believe Phoenix uses the token by default. So It is a matter of updating the guides and docs.

In particular, Phoenix supports session in websockets, which has been the encouraged approach for a while. Then we should mention the token approach, as long as it is short lived, and then also passing the token on join, which is what this issue recommends. PRs to docs are welcome.

josevalim:
Ping if you folks are still interested in a PR to the docs. Ping me if you do!

satoren:
The [template](https://github.com/phoenixframework/phoenix/blob/e0fa893af6122bd89eadd379ae9b450bee34cc7b/priv/templates/phx.gen.socket/socket.js#L9) uses token. In other words, it can be said that token is used by default.

mspanc:
There is no method in the JavaScript [WebSockets API](https://html.spec.whatwg.org/multipage/comms.html#network) for specifying additional headers for the client/browser to send.

[...]

PR #6086, opened 2025-02-14
Support transport agnostic token passing in channels

SteffenDE:
For WebSocket, the `Sec-WebSocket-Protocol` header is used. For LongPoll, an `Authorization` header is passed instead.
    
Fixes https://github.com/phoenixframework/phoenix/issues/5778.

### DIFF
--- a/assets/js/phoenix/longpoll.js
+++ b/assets/js/phoenix/longpoll.js
@@ -15,7 +15,12 @@ let arrayBufferToBase64 = (buffer) => {
 
 export default class LongPoll {
 
-  constructor(endPoint){
+  constructor(endPoint, protocols){
+    // we only support subprotocols for authToken
+    // ["phoenix", "base64url.bearer.authorization.phx.BASE64_ENCODED_TOKEN"]
+    if (protocols.length === 2 && protocols[1].startsWith("base64url.bearer.authorization.phx.")) {
+      this.authToken = atob(protocols[1].slice("base64url.bearer.authorization.phx.".length))
+    }
     this.endPoint = null
     this.token = null
     this.skipHeartbeat = true
@@ -58,7 +63,11 @@ export default class LongPoll {
   isActive(){ return this.readyState === SOCKET_STATES.open || this.readyState === SOCKET_STATES.connecting }
 
   poll(){
-    this.ajax("GET", "application/json", null, () => this.ontimeout(), resp => {
+    const headers = {"Accept": "application/json"}
+    if (this.authToken) {
+      headers["Authorization"] = `Bearer ${this.authToken}`
+    }
+    this.ajax("GET", headers, null, () => this.ontimeout(), resp => {
       if(resp){
         var {status, token, messages} = resp
         this.token = token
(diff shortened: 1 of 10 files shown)


assets/js/phoenix/longpoll.js
```diff
@@ -58,7 +63,11 @@ export default class LongPoll {
   isActive(){ return this.readyState === SOCKET_STATES.open || this.readyState === SOCKET_STATES.connecting }
 
   poll(){
-    this.ajax("GET", "application/json", null, () => this.ontimeout(), resp => {
+    const headers = {"Accept": "application/json"}
+    if (this.authToken) {
+      headers["Authorization"] = `Bearer ${this.authToken}`
```
josevalim, line 68: Let's use our own header.

guides/real_time/channels.md
```diff
@@ -385,7 +385,24 @@ That's all there is to our basic chat app. Fire up multiple browser tabs and you
...
+### Step 1 - Enable the `auth_token` functionality in the socket
+
+Phoenix supports a transport agnostic way to pass an authentication token to the server. To enable this, we need to pass the `:auth_token` option to the socket declaration in our `Endpoint` module and configure the `connect_info` to include the `:auth_token` key.
+
+```elixir
+defmodule HelloWeb.Endpoint do
+  use Phoenix.Endpoint, otp_app: :hello
+
+  socket "/socket", HelloWeb.UserSocket,
+    websocket: [connect_info: [:auth_token]],
```
josevalim, line 397: Let's have it on by default. nil if nothing is given.

lib/phoenix/transports/websocket.ex
```diff
@@ -82,4 +95,36 @@ defmodule Phoenix.Transports.WebSocket do
...
+          subprotocols_header
+          |> Plug.Conn.Utils.list()
+          |> Enum.split_with(&String.starts_with?(&1, @auth_token_prefix))
+
+        case request_subprotocols do
+          {[@auth_token_prefix <> encoded_token], actual_subprotocols} ->
+            token = Base.decode64!(encoded_token, padding: false)
+
+            conn
+            |> put_private(:__phoenix_transport_auth_token, token)
```
josevalim, line 115:
```suggestion
            |> put_private(:phoenix_transport_auth_token, token)
```

lib/phoenix/transports/websocket.ex
```diff
@@ -35,12 +37,23 @@ defmodule Phoenix.Transports.WebSocket do
...
+      end
+
     conn
     |> fetch_query_params()
     |> Transport.code_reload(endpoint, opts)
     |> Transport.transport_log(opts[:transport_log])
     |> Transport.check_origin(handler, endpoint, opts)
-    |> Transport.check_subprotocols(opts[:subprotocols])
+    |> maybe_auth_token_from_header(opts[:auth_token])
+    |> Transport.check_subprotocols(subprotocols)
```
josevalim, line 56: Honestly, it feels we should just bring this function to this module and have it deal with both auth_token and subprotocols at once. `Transport` is meant to have transport agnostic functions, and that one is clearly websockets related.

josevalim (approved):
LGTM. I have dropped a comment about subprotocol handling but it can be handled in a future PR!

lib/phoenix/transports/websocket.ex
SteffenDE: It is part of the public API, so we can't really change it, can we?
https://hexdocs.pm/phoenix/Phoenix.Socket.Transport.html#check_subprotocols/2

josevalim: Yes, we cannot, but maybe we can move the auth token handling there anyway?

PR #6089, opened 2025-02-17
add button_classes to CoreComponents

### LINKED ISSUE #5770: HTML and Live CRUD generators output invalid HTML (buttons nested in links)

SteffenDE:
Buttons and Links must not be nested inside each other, otherwise there are accessibility issues. A previous PR did not fix the problem, but just inverted the problem.

Adds a helper function to return button_classes for styling other components, like links.

Fixes #5770.
Relates to: #5820

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -178,6 +178,41 @@ defmodule <%= @web_namespace %>.CoreComponents do
   end
 
   @doc """
+  Renders a link styled as a button.
+
+  ## Examples
+
+      <.link_button href={~p"/items/new"}>
+        New Item
+      </.link_button>
+  """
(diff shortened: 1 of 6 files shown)


josevalim:
I would personally prefer to avoid new function components. I wonder if we could instead have a function called `button_classes()` that we use in our button component and elsewhere?

josevalim:
Yes, exactly. If we can share that across places, then it looks good to me! We may need appearance-none.

SteffenDE:
Changed! I don't think we need appearance-none there, but people may add it if they use it in weird places 😃

josevalim (approved):
!!!!!

PR #6094, opened 2025-02-17
WebSockets: support passing through sec-websocket headers

SteffenDE:
Supersedes https://github.com/phoenixframework/phoenix/pull/5904.

### DIFF
--- a/lib/phoenix/socket/transport.ex
+++ b/lib/phoenix/socket/transport.ex
@@ -269,7 +269,7 @@ defmodule Phoenix.Socket.Transport do
 
     connect_info =
       Enum.map(connect_info, fn
-        key when key in [:peer_data, :trace_context_headers, :uri, :user_agent, :x_headers, :auth_token] ->
+        key when key in [:peer_data, :trace_context_headers, :uri, :user_agent, :x_headers, :sec_websocket_protocol, :auth_token] ->
           key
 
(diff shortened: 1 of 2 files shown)


josevalim:
Ideally `connect_info` is not specific to any transport. What if we just include all websocket headers alongside x-headers?

SteffenDE:
@josevalim well, `:x_headers` is also not transport agnostic. If I build a transport based on telnet, there are no x_headers. I don't think it makes sense to mix x_headers + sec_websocket. If we didn't have `x_headers` yet, I'd propose a generic `:headers` key that one would configure with a list of headers / prefixes:

```elixir
socket "/socket", AppWeb.UserSocket,
  websocket: [
    connect_info: [:uri, headers: [prefix: ["x-", "sec-websocket-"]]]
  ]
```

But since we already have x_headers 🤷🏻‍♂️

haljpg (approved):

josevalim:
@SteffenDE x-headers work for websockets too, except in the browser. Another interpretation is to say that `x_headers` means "extra headers", and that would fit both x- and websocket prefixes. However, if you prefer to go with this PR, I am ok with it. Your call.

josevalim (approved):

SteffenDE:
Chris approves, let's go!

PR #6096, opened 2025-02-18
support --primary-key in phx.gen.{html,json,live}

SteffenDE:
Follow-up for https://github.com/phoenixframework/phoenix/pull/5766.


haljpg (approved):

haljpg (approved):

haljpg:
Session identifier 6069

josevalim (approved):
This is good to me but it might complicate scopes. So there is a chance we will need to revert it later on. Your call.

SteffenDE:
At the moment, I don't think it will complicate scopes much. But we'll see :)

PR #6101, opened 2025-02-19
align sqlite migration code with comment

### LINKED ISSUE #5945: sqlite generated code - wrong wording?
### Environment

* Elixir version (elixir -v): 1.17.3
* Phoenix version (mix deps): 1.7.14
* Operating system: MacOS 15.0

### Actual behavior

When generating a new Phoenix app with `--database sqlite3`, a call to `Ecto.Migrator` is added to the list of children.
This has a `:skip?` option that uses this generated function:

```elixir
  defp skip_migrations?() do
    # By default, sqlite migrations are run when using a release
    System.get_env("RELEASE_NAME") != nil
  end
```

The comment and the result are different, by default this returns `true` in a release, which causes the migrations to be skipped.

rmand97:
fixes #5945 

The wording of the code comment implies the opposite of what the function actually does.
I aligned the code with the comment. Can also do it the other way around if needed.

### DIFF
--- a/installer/templates/phx_single/lib/app_name/application.ex
+++ b/installer/templates/phx_single/lib/app_name/application.ex
@@ -37,6 +37,6 @@ defmodule <%= @app_module %>.Application do
 
   defp skip_migrations?() do
     # By default, sqlite migrations are run when using a release
-    System.get_env("RELEASE_NAME") != nil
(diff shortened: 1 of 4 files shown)


Gazler:
Thanks <3

PR #6102, opened 2025-02-20
add scopes to generators

SteffenDE:
Adds a new concept called "scopes" to the Phoenix generators, as introduced by @chrismccord a while ago:

https://youtu.be/Ckgl9KO4E4M?si=TKgdWCQ87OIPzW_A&t=307
https://github.com/chrismccord/todo_trek/blob/main/lib/todo_trek/scope.ex

### DIFF
--- a/lib/mix/tasks/phx.gen.auth.ex
+++ b/lib/mix/tasks/phx.gen.auth.ex
@@ -391,7 +391,7 @@ defmodule Mix.Tasks.Phx.Gen.Auth do
     %{scope: scope, default_scope: default_scope, create_new?: new?} = binding[:scope_config]
 
     cond do
-      # TODO: check if we want to include this or only ask for the existing default case
+      # this can only happen if --scope is used and the user explicitly asked for the scope name
       scope && not new? ->
         Mix.shell().yes?("""
         The scope #{scope.name} is already configured.


integration_test/test/code_generation/app_with_no_options_test.exs
```diff
@@ -59,7 +59,7 @@ defmodule Phoenix.Integration.CodeGeneration.AppWithNoOptionsTest do
         [
           "--no-halt",
           "-e",
-          "spawn fn -> IO.gets('') && System.halt(0) end",
+          "spawn fn -> IO.gets(~c[]) && System.halt(0) end",
```
josevalim, line 62:
```suggestion
          "spawn fn -> IO.gets([]) && System.halt(0) end",
```

lib/mix/tasks/phx.gen.context.ex
```diff
@@ -93,7 +93,9 @@ defmodule Mix.Tasks.Phx.Gen.Context do
     live: :boolean,
     compile: :boolean,
     primary_key: :string,
-    migration: :boolean
+    migration: :boolean,
+    scope: :string,
+    no_scope: :boolean
```
josevalim, line 98: I believe you don't need `no_scope`. Since you have `--scope`, you can pass `--no-scope`, and it will come with `scope: nil`. You can distinguish between them verifying if the option exists or not.

priv/templates/phx.gen.auth/auth.ex
```diff
@@ -121,28 +122,28 @@ defmodule <%= inspect auth_module %> do
...
-      and assigns the current_<%= schema.singular %> to socket assigns based
+      and assigns the current_scope to socket assigns based
       on <%= schema.singular %>_token.
       Redirects to login page if there's no logged <%= schema.singular %>.
 
   ## Examples
 
   Use the `on_mount` lifecycle macro in LiveViews to mount or authenticate
-  the current_<%= schema.singular %>:
+  the current_scope:
```
josevalim, line 141:
```suggestion
  the `current_scope`:
```

priv/templates/phx.gen.auth/scope.ex
```diff
@@ -0,0 +1,31 @@
...
+  """
+
+  alias <%= inspect schema.module %>
+
+  defstruct user: nil
+
+  @doc """
+  Creates a scope for the given <%= schema.singular %>.
+  """
+  def for_<%= schema.singular %>(nil), do: %__MODULE__{user: nil}
```
josevalim, line 26: Honestly, I would prefer to not have this pattern. If we are thinking about the type system, it means that now we need to handle about the possibility of user being nil everywhere. Or we have a scope with user, or we have no scope.

priv/templates/phx.gen.auth/scope.ex
```diff
@@ -0,0 +1,31 @@
...
+
+  def for_<%= schema.singular %>(%<%= inspect schema.alias %>{} = <%= schema.singular %>) do
```
josevalim, line 28: Does this mean we will have `UserScope.for_user`? Maybe we should have `AccountsScope` or just `AuthScope`? I also don't like calling it `UserScope` because in many cases you will have an org field too, and you will likely want to customize it to be per org. We should probably have some docs/guides on that. Perhaps under the `mix phx.gen.auth` section of the guides.

priv/templates/phx.gen.context/schema_access_scope.ex
```diff
@@ -0,0 +1,139 @@
...
+
+    * {:created, %<%= inspect schema.alias %>{}}
+    * {:updated, %<%= inspect schema.alias %>{}}
+    * {:deleted, %<%= inspect schema.alias %>{}}
+
+  """
+  def subscribe_<%= schema.plural %>(%<%= inspect scope.alias %>{} = <%= scope.name %>_scope) do
+    key = <%= scope.name %>_scope.<%= Enum.join(scope.access_path, ".") %>
+
+    Phoenix.PubSub.subscribe(<%= inspect context.base_module %>.PubSub, "<%= scope.name %>:#{key}:<%= schema.plural %>")
```
josevalim, line 18: I was wondering if we should have this under the scope somehow. `subscribe(current_scope, topic)` and `broadcast(current_scope, topic)`. But this also means a basic scope will need additional functions in order to work with generators.

lib/mix/tasks/phx.gen.context.ex
SteffenDE: Hmm
```
iex(1)> OptionParser.parse(["--no-scope"], switches: [scope: :string])
{[], [], []}
iex(2)> OptionParser.parse(["--scope"], switches: [scope: :string])
{[], [], [{"--scope", nil}]}
iex(3)> OptionParser.parse(["--scope", "foo"], switches: [scope: :string])
{[scope: "foo"], [], []}
```

priv/templates/phx.gen.auth/scope.ex
SteffenDE: Well yes, we could have the module as MyApp.Accounts.AuthScope.for_user, but I think the scope itself should still be called `:user`, because if I create a new application, resources are scoped **by user**. If I then add a tenant / org, I'd do

```
config :my_app, :scopes,
  user: [
    default: false,
    module: MyApp.Accounts.AuthScope,
    assign_key: :current_scope,
    access_path: [:user, :id],
    schema_key: :user_id,
    schema_type: :id,
    schema_table: :users
  ],
  org: [
    default: true,
    module: MyApp.Accounts.AuthScope,
    assign_key: :current_scope,
    access_path: [:user, :org_id],
    schema_key: :org_id,
    schema_type: :id,
    schema_table: :organization
  ]
```

And then I can decide with a single scope module if I want to scope a new resource by user or organization.

priv/templates/phx.gen.auth/scope.ex
SteffenDE: I guess it depends if we think that it is common to want scoped functionality for unauthenticated requests. I think it would be common to have one scope struct with fields that can either be nil or not instead of different scope structs.

The alternative is to have a `for_user(nil), do: nil` or the need to change all UserAuth functions to something like

```elixir
defp mount_current_scope(socket, session) do
    Phoenix.Component.assign_new(socket, :current_scope, fn ->
      with user_token when is_binary(user_token) <- session["user_token"],
           user when not is_nil(user) <- Accounts.get_user_by_session_token(user_token) do
          AuthScope.for_user(user)
       end
    end)
  end
```

and only call the scope function if we have a user in the first place.

lib/mix/tasks/phx.gen.context.ex
josevalim: Sad trombone. You are right. You could explicitly check for `"--no-scope" in args` and act accordingly, but then I guess it would end-up the same.

priv/templates/phx.gen.auth/scope.ex
josevalim: > I think it would be common to have one scope struct with fields that can either be nil or not instead of different scope structs.

I agree, but I would prefer to not assume and start with the most streamlined version. For example, in teams, I believe we always have org and user.

priv/templates/phx.gen.auth/scope.ex
josevalim: Agreed, which is why I think it should be called AuthScope or something more generic. :+1:

priv/templates/phx.gen.context/schema_access_scope.ex
```diff
@@ -0,0 +1,139 @@
...
+  Returns the list of <%= schema.plural %>.
+
+  ## Examples
+
+      iex> list_<%= schema.plural %>(scope)
+      [%<%= inspect schema.alias %>{}, ...]
+
+  """
+  def list_<%= schema.plural %>(%<%= inspect scope.alias %>{} = <%= scope.name %>_scope) do
+    Repo.all(from <%= schema.singular %> in <%= inspect schema.alias %>, where: <%= schema.singular %>.<%= scope.schema_key %> == ^<%= scope.name %>_scope.<%= Enum.join(scope.access_path, ".") %>)
```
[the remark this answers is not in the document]
josevalim: Anyway, not the topic of this PR, we can update this and the auth generators to use Repo.transact once Ecto is actually out. I will start working on that front.

priv/templates/phx.gen.auth/scope.ex
SteffenDE: now called AuthScope

priv/templates/phx.gen.context/schema_access_scope.ex
SteffenDE: I think it's fine to have those in the context. People can extract to their scope if they want.

guides/scopes.md
```diff
@@ -0,0 +1,261 @@
+# Scopes
+
+> **Requirement**: This guide expects that you have gone through the [introductory guides](installation.html) and got a Phoenix application [up and running](up_and_running.html).
+
+> **Requirement**: This guide expects that you have gone through the [Contexts guide](contexts.html).
```
josevalim, line 5: Btw, I'd remove those requirements, once we make this more of a reference.

lib/mix/phoenix/scope.ex
```diff
@@ -0,0 +1,93 @@
...
+  @doc """
+  Returns the default scope.
+  """
+  def default_scope(otp_app) do
+    case Enum.filter(scopes_from_config(otp_app), fn {_, scope} -> scope.default end) do
+      [{_name, scope}] ->
+        scope
+
+      [_ | _] = scopes ->
+        raise """
```
josevalim, line 49: We want to use `Mix.raise` in this file. It is specific to Mix and the main difference is that it doesn't include a stacktrace. Those error messages also usually start with uppercase.

lib/mix/phoenix/scope.ex
```diff
@@ -0,0 +1,93 @@
...
+  def scope_from_opts(_otp_app, _name, true), do: nil
+
+  def scope_from_opts(otp_app, nil, _), do: default_scope(otp_app)
+
+  def scope_from_opts(otp_app, name, _) do
+    key = String.to_atom(name)
+    scopes = scopes_from_config(otp_app)
+
+    Map.get_lazy(scopes, key, fn ->
+      raise """
```
josevalim, line 78: Same here and above.

lib/mix/tasks/phx.gen.auth.ex
```diff
@@ -268,35 +273,146 @@ defmodule Mix.Tasks.Phx.Gen.Auth do
...
   end
 
-  defp files_to_be_generated(%Context{schema: schema, context_app: context_app} = context) do
+  defp prompt_for_scope_conflicts(binding) do
+    schema = binding[:schema]
+    %{scope: scope, default_scope: default_scope, create_new?: new?} = binding[:scope_config]
+
+    cond do
+      # TODO: check if we want to include this or only ask for the existing default case
+      scope && not new? ->
```
josevalim, line 350: I would try to keep it simpler and just pick a different name?

lib/mix/tasks/phx.gen.auth/injector.ex
```diff
@@ -128,8 +135,8 @@ defmodule Mix.Tasks.Phx.Gen.Auth.Injector do
     "plug " <> router_plug_name(schema)
   end
 
-  defp router_plug_name(%Schema{} = schema) do
-    ":fetch_current_#{schema.singular}"
+  defp router_plug_name(_schema) do
+    ":fetch_current_scope"
```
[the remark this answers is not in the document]
josevalim: As discussed, maybe `fetch_current_scope_for_user`?

lib/mix/tasks/phx.gen.context.ex
```diff
@@ -336,6 +358,8 @@ defmodule Mix.Tasks.Phx.Gen.Context do
           "Cannot generate schema #{schema} because it has the same name as the application"
         )
 
+
+
```
josevalim, line 362:
```suggestion
```

:D

SteffenDE:
Scopes guide updated!

lib/mix/tasks/phx.gen.auth.ex
josevalim: I would try user, then accounts_user, then my_app_accounts_user. If all of these are taken, then give up :P

SteffenDE: alright!

priv/templates/phx.gen.auth/auth.ex
```diff
@@ -121,28 +122,28 @@ defmodule <%= inspect auth_module %> do
   end
 
   <%= if live? do %>@doc """
-  Handles mounting and authenticating the current_<%= schema.singular %> in LiveViews.
+  Handles mounting and authenticating the current_scope in LiveViews.
 
   ## `on_mount` arguments
 
-    * `:mount_current_<%= schema.singular %>` - Assigns current_<%= schema.singular %>
+    * `:mount_current_scope` - Assigns current_scope
```
josevalim, line 129: Shouldn't this mirror `fetch_current_scope_for_...` and become `mount_current_scope_for_...`?

josevalim (approved):
I have added one comment and we can ship it! I will work on the docs this or the next week, so we can ship the RC.

priv/templates/phx.gen.auth/auth.ex
SteffenDE: we can change it, but since LV hooks don't have naming conflicts (you always need to pass the module), it's not really an issue

josevalim: Ah, ok, sounds good to me!

priv/templates/phx.gen.auth/auth.ex
```diff
@@ -229,7 +233,7 @@ defmodule <%= inspect auth_module %> do
   they use the application at all, here would be a good place.
   """
   def require_authenticated_<%= schema.singular %>(conn, _opts) do
-    if conn.assigns[:current_<%= schema.singular %>] do
+    if conn.assigns.current_scope do
```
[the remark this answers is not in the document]
josevalim: Agreed on being more precise here in case someone adds more fields to the scope in the future. It doesn’t hurt.

josevalim:

PR #6107, opened 2025-02-23
Fix ex_cldr_routes URL

nwjlyons:
https://hex.pm/packages/cldr_routes is a 404. Should be https://hex.pm/packages/ex_cldr_routes


SteffenDE:
Thank you! 🙌🏻

PR #6111, opened 2025-03-04
Update core components to support multiple=true on checkbox inputs

andrewtimberlake:
Currently, if you want to use the core components for multi value checkboxes, you can’t.
This change maintains backwards compatibility with standard on/off checkboxes and adds the ability to do this:
```heex
<form>
  <.input type="checkbox" name="field[active][]" value="true" label="Yes" multiple={true} />
  <.input type="checkbox" name="field[active][]" value="false" label="No" multiple={true} />
</form>
```

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -225,7 +225,7 @@ defmodule <%= @web_namespace %>.CoreComponents do
   attr :checked, :boolean, doc: "the checked flag for checkbox inputs"
   attr :prompt, :string, default: nil, doc: "the prompt for select inputs"
   attr :options, :list, doc: "the options to pass to Phoenix.HTML.Form.options_for_select/2"
-  attr :multiple, :boolean, default: false, doc: "the multiple flag for select inputs"
+  attr :multiple, :boolean, default: false, doc: "the multiple flag for select and checkbox inputs"
 
   attr :rest, :global,
     include: ~w(accept autocomplete capture cols disabled form list max maxlength min minlength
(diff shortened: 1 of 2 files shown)


SteffenDE:
What is a "multi-value checkbox"? A checkbox is either on or off (or maybe indeterminate) and even in your example, how does it make sense to have an active field that can be both yes and no at the same time?

CoreComponents are meant to be modified if your app needs it. To me, this looks like such a thing that may be needed in some rare circumstances, but not worth being included by default.

andrewtimberlake:
That’s a simple example, but a practical one.
CleanShot 2025-03-04 at 13 13 10@2x

A multi-value checkbox is a collection of checkboxes that produce an array of results. It is similar to using a multi value select but sometimes it’s a much nicer UI.

I have made this change in a number of different projects:
- check off from a list of possible files to attach to a message
- check off from a list of fields to include when sending a webhook
- check off all the regions this business belongs to
- check off all the categories this project belongs to

andrewtimberlake:
And I’m about to implement another: checking off rows of a table (edit multiple, delete multiple etc).
Similar to this:
CleanShot 2025-03-04 at 13 21 13@2x

SteffenDE:
Those examples make more sense :)

I still think core components should focus on the basic functionality. Since the generators don't use this, I'm not sure that we should add it. Other opinions? @josevalim @chrismccord

andrewtimberlake:
Thanks. One more argument for this change.
If it’s not made, then this code:
```heex
<form>
  <.input type="checkbox" name="select[id][]" value="1" label="First" />
  <.input type="checkbox" name="select[id][]" value="2" label="Second" />
</form>
```
Will produce a params of `%{"select": %{"id" => ["false", "1", "false", "2"]}}` which is not expected.

josevalim:
Right, when I have to implement multi-value checkboxes, I just use the built-in HTML checkboxes instead, because a large part of the structure is different:

1. You don't want one error per checkbox
2. You don't want a hidden input
3. IDs and names are different

Your approach is fine but I wanted to point out there are other options. :) In any case, CoreComponents goal is to serve the built-in generators and all cases, so you are welcome to augment it as needed, when needed. Thanks for the PR!

PR #6118, opened 2025-03-12
add route_prefix option to scopes

SteffenDE:


guides/authn_authz/scopes.md
```diff
@@ -202,6 +202,10 @@ In this example, the scope is called `user` and it is the `default` scope that i
 
 * `access_path` - a list of keys that define the path to the identifying field in the scope struct. The generators generate code like `where: schema_key == ^scope.user.id`.
 
+* `route_prefix` - a path template string for how resources should be nested. For example, `/orgs/:org` would generate routes like `/orgs/:org/posts`. The parameter segment (`:org`) will be replaced with the appropriate scope access value in templates and LiveViews.
+
+* `route_access_path` - a list of keys that define the path to the field used in route generation. This is particularly useful for user-friendly URLs where you might want to use a slug instead of an ID. If not specified, it defaults to `access_path`. For example, `[:organization, :slug]` would generate URLs using organization slugs instead of IDs.
```
josevalim, line 207: Do we need this? Can't we rely on Phoenix.Param?

SteffenDE: If someone has phoenix param implemented, they can just use `[:organization]` as parameter. But we must have some path, we cannot assume how the scope struct looks.

lib/mix/tasks/phx.gen.json.ex
```diff
@@ -175,26 +178,39 @@ defmodule Mix.Tasks.Phx.Gen.Json do
 
   @doc false
   def print_shell_instructions(%Context{schema: schema, context_app: ctx_app} = context) do
+    scope_message = if schema.scope && schema.scope.route_prefix do
+      "\nPlease ensure that the scope is assigned in your pipeline hooks before accessing these routes."
```
josevalim, line 182:
```suggestion
      "\nEnsure the routes are defined in a block that sets the `current_scope` assign."
```

?

lib/mix/tasks/phx.gen.live.ex
```diff
@@ -260,10 +269,16 @@ defmodule Mix.Tasks.Phx.Gen.Live do
     prefix = Module.concat(context.web_module, schema.web_namespace)
     web_path = Mix.Phoenix.web_path(ctx_app)
 
+    scope_message = if schema.scope && schema.scope.route_prefix do
+      "\nPlease ensure that the scope is assigned in your pipeline / on_mount hooks before accessing these routes."
```
josevalim, line 273: Similar here.

guides/authn_authz/scopes.md
```diff
@@ -419,6 +423,8 @@ config :my_app, :scopes,
     module: MyApp.Accounts.Scope,
     assign_key: :current_scope,
     access_path: [:organization, :id],
+    route_prefix: "/orgs/:org",
+    route_access_path: [:organization, :slug],
```
josevalim, line 427: In this example, do we want to implement Phoenix.Param or use `route_access_path`? I would prefer Phoenix.Param, it feels more idiomatic.

lib/mix/tasks/phx.gen.json.ex
josevalim: I guess we need to get the assign name from the scope config.

lib/mix/phoenix/scope.ex
```diff
@@ -25,6 +27,7 @@ defmodule Mix.Phoenix.Scope do
       scope
       | name: name,
         alias: alias,
+        route_access_path: scope.route_access_path || scope.access_path,
```
josevalim, line 30: I am thinking it should default to either `Enum.drop(scope.access_path, -1)` or simply `[name]`. Thoughts? I really don't think we should be generated .id or .slug by default.

SteffenDE: `Enum.drop(scope.access_path, -1)` is good 👍

josevalim: I guess it could be empty in some cases, then maybe we fallback to scope.access_path as a whole. Unsure.

guides/authn_authz/scopes.md
josevalim: We can remove this if we use Enum.drop, so that gets my vote :)

josevalim (approved):
Ship it after comments!!!! :)

PR #6122, opened 2025-03-15
Add `Phoenix.VerifiedRoutes` behaviour for custom routers

zachdaniel:
This comes with two parts:

- These routes are expanded when printing routes using `mix phx.routes`. I've added an additional `label` metadata which is extremely useful for things like `AshJsonApi` to display the resource & action being routed to
- verified routes logic has been modified to match these expanded routes as well. 

I've tested manually and it works well, but I will of course add tests for this logic if this kind of thing seems like a good idea to the team.

Original ElixirForum discussion: https://elixirforum.com/t/allow-forwarded-to-plugs-to-have-a-callback-that-returns-additional-routes/65365/2

I had to use `Code.ensure_compiled/1` on the nested plug, which I couldn't find a good way around but I'm open to ideas on (open to ideas on the whole implementation).

One note is that part of the reason for a lot of these forwarded-to plugs to exist is to do dynamic routing that can't be done at compile time, so it would negate most of the benefit if we expanded these routes at compile time directly into the router (verified routes being different because it happens in after_verify).

Thoughts?

### DIFF
--- a/lib/phoenix/router.ex
+++ b/lib/phoenix/router.ex
@@ -358,6 +358,7 @@ defmodule Phoenix.Router do
       unquote(prelude(opts))
       unquote(defs())
       unquote(match_dispatch())
+      unquote(verified_routes())
     end
   end
 
@@ -537,6 +538,20 @@ defmodule Phoenix.Router do
     end
   end
 
+  defp verified_routes() do
+    quote location: :keep, generated: true do
+      @behaviour Phoenix.VerifiedRoutes
+
+      def formatted_routes(_) do
+        Phoenix.Router.__formatted_routes__(__MODULE__)
+      end
+
+      def verified_route?(_, split_path) do
+        Phoenix.Router.__verified_route__?(__MODULE__, split_path)
+      end
+    end
+  end
+
   @doc false
   defmacro __before_compile__(env) do
     routes = env.module |> Module.get_attribute(:phoenix_routes) |> Enum.reverse()
@@ -623,14 +638,15 @@ defmodule Phoenix.Router do
     warn_on_verify? = Enum.all?(routes, & &1.warn_on_verify?)
 
     case Enum.find(routes, &(&1.kind == :forward)) do
-      %{metadata: %{forward: forward}, plug: plug} ->
+      %{metadata: %{forward: forward}, plug: plug, plug_opts: plug_opts} ->
         quote generated: true do
           def __forward__(unquote(plug)) do
             unquote(forward)
           end
 
           def __verify_route__(unquote(path)) do
-            {{unquote(plug), unquote(forward)}, unquote(warn_on_verify?)}
+            {{unquote(plug), unquote(forward), unquote(Macro.escape(plug_opts))},
+             unquote(warn_on_verify?)}
           end
         end
 
@@ -1280,6 +1296,66 @@ defmodule Phoenix.Router do
     with {metadata, _prepare, _pipeline, {_plug, _opts}} <-
            router.__match_route__(split_path, method, host) do
       Map.delete(metadata, :conn)
+    end
+  end
+
+  @doc false
+  def __formatted_routes__(router) do
+    Enum.flat_map(router.__routes__(), fn route ->
+      Code.ensure_loaded(route.plug)
+
+      if function_exported?(route.plug, :formatted_routes, 1) do
+        route.plug_opts
+        |> route.plug.formatted_routes()
+        |> Enum.map(fn nested_route ->
+          %{
+            route
+            | path: Path.join(route.path, nested_route.path),
+              verb: nested_route.verb,
+              plug_opts: nested_route.plug_opts
+          }
+        end)
+      else
+        plug =
+          case route.metadata[:mfa] do
+            {module, _, _} -> module
+            _ -> route.plug
+          end
+
+        label = "#{inspect(plug)} #{inspect(route.plug_opts)}"
+
+        [
+          %{
+            helper: route.helper,
+            verb: route.verb,
+            path: route.path,
+            label: label
+          }
+        ]
+      end
+    end)
+  end
+
+  @doc false
+  def __verified_route__?(router, split_path) do
+    case router.__verify_route__(split_path) do
+      {_forward_plug, true = _warn_on_verify?} ->
+        false
+
+      {nil = _forward_plug, false = _warn_on_verify?} ->
+        true
+
+      {{router, script_name, plug_opts}, false = _warn_on_verify?} ->
+        Code.ensure_loaded(router)
+
+        if function_exported?(router, :verified_route?, 2) do
+          router.verified_route?(plug_opts, split_path -- script_name)
+        else
+          true
+        end
+
+      :error ->
+        false
     end
   end
 end
--- a/lib/phoenix/router/console_formatter.ex
+++ b/lib/phoenix/router/console_formatter.ex
@@ -10,7 +10,8 @@ defmodule Phoenix.Router.ConsoleFormatter do
   @longpoll_verbs ["GET", "POST"]
 
   def format(router, endpoint \\ nil) do
-    routes = Phoenix.Router.routes(router)
+    routes = router.formatted_routes([])
+
     column_widths = calculate_column_widths(router, routes, endpoint)
 
     IO.iodata_to_binary([
@@ -19,7 +20,7 @@ defmodule Phoenix.Router.ConsoleFormatter do
     ])
   end
 
-  defp format_endpoint(nil, _), do: ""
+  defp format_endpoint(nil, _router), do: ""
 
   defp format_endpoint(endpoint, widths) do
     case endpoint.__sockets__() do
@@ -104,21 +105,12 @@ defmodule Phoenix.Router.ConsoleFormatter do
     %{
       verb: verb,
       path: path,
-      plug: plug,
-      metadata: metadata,
-      plug_opts: plug_opts,
-      helper: helper
+      label: label
     } = route
 
     verb = verb_name(verb)
-    route_name = route_name(router, helper)
+    route_name = route_name(router, Map.get(route, :helper))
     {verb_len, path_len, route_name_len} = column_widths
-
-    log_module =
-      case metadata[:mfa] do
-        {mod, _fun, _arity} -> mod
-        _ -> plug
-      end
 
     String.pad_leading(route_name, route_name_len) <>
       "  " <>
@@ -126,7 +118,7 @@ defmodule Phoenix.Router.ConsoleFormatter do
       "  " <>
       String.pad_trailing(path, path_len) <>
       "  " <>
-      "#{inspect(log_module)} #{inspect(plug_opts)}\n"
+      label <> "\n"
   end
 
   defp route_name(_router, nil), do: ""
--- a/lib/phoenix/verified_routes.ex
+++ b/lib/phoenix/verified_routes.ex
@@ -181,6 +181,35 @@ defmodule Phoenix.VerifiedRoutes do
   Finally, for even more complex use cases, where the whole URL needs to localized,
   see projects such as [`routex`](https://hex.pm/packages/routex) and
   [`ex_cldr_routes`](https://hex.pm/packages/ex_cldr_routes).
+
+  ## Usage with custom plugs
+
+  Sometimes, when we want to do dynamic routing, we will forward to custom plugs.
+  It is possible to make these dynamic routers support `mix phx.routes` and verified
+  routes at compile time by adopting the `Phoenix.VerifiedRoutes` behaviour. 
+  For example:
+
+      defmodule MyApp.LocaleRouter do
+        use Plug.Router
+        @behaviour Phoenix.VerifiedRoutes
+
+        # custom routing rules
+
+        # for displaying in `mix phx.routes`
+        def formatted_routes(plug_opts) do
+          for locale <- supported_locales(plug_opts) do
+            %{verb: "GET", path: "/#{locale}/*subpath"}
+          end
+        end
+
+        def verified_route?(plug_opts, path) do
+          plug_opts
+          |> supported_locales()
+          |> Enum.any?(fn locale -> 
+            Enum.at(path, 0) == locale
+          end)
+        end
+      end
   '''
   @doc false
   defstruct router: nil,
@@ -238,6 +267,29 @@ defmodule Phoenix.VerifiedRoutes do
     })
   end
 
+  @type plug_opts :: any()
+  @type formatted_route :: %{
+          required(:verb) => String.t(),
+          required(:path) => String.t(),
+          required(:label) => String.t()
+        }
+
+  @doc """
+  Returns the necessary information about routes for display in `mix phx.routes`.
+  
+  The `plug_opts` is typically only passed when the router is mounted within
+  a `Phoenix.Router`. Otherwise it defaults to `[]`.
+  """
+  @callback formatted_routes(plug_opts()) :: [formatted_route()]
+
+  @doc """
+  Returns `true` if the path is verified, and false if not.
+  
+  The `plug_opts` is typically only passed when the router is mounted within
+  a `Phoenix.Router`. Otherwise it defaults to `[]`.
+  """
+  @callback verified_route?(plug_opts(), [String.t()]) :: boolean()
+
   @after_verify_supported Version.match?(System.version(), ">= 1.14.0")
 
   defmacro __before_compile__(_env) do
@@ -256,13 +308,23 @@ defmodule Phoenix.VerifiedRoutes do
   @doc false
   def __verify__(routes) when is_list(routes) do
     Enum.each(routes, fn %__MODULE__{} = route ->
-      unless match_route?(route.router, route.test_path) do
+      test_path = split_test_path(route.test_path)
+
+      unless route.router.verified_route?([], test_path) do
         IO.warn(
           "no route path for #{inspect(route.router)} matches #{route.inspected_route}",
           route.warn_location
         )
       end
     end)
+  end
+
+  defp split_test_path(test_path) do
+    test_path
+    |> String.split("#")
+    |> Enum.at(0)
+    |> String.split("/")
+    |> Enum.filter(fn segment -> segment != "" end)
   end
 
   defp expand_alias({:__aliases__, _, _} = alias, env),
@@ -832,37 +894,6 @@ defmodule Phoenix.VerifiedRoutes do
   defp to_param(false), do: "false"
   defp to_param(true), do: "true"
   defp to_param(data), do: Phoenix.Param.to_param(data)
-
-  defp match_route?(router, test_path) when is_binary(test_path) do
-    split_path =
-      test_path
-      |> String.split("#")
-      |> Enum.at(0)
-      |> String.split("/")
-      |> Enum.filter(fn segment -> segment != "" end)
-
-    match_route?(router, split_path)
-  end
-
-  defp match_route?(router, split_path) when is_list(split_path) do
-    case router.__verify_route__(split_path) do
-      {_forward_plug, true = _warn_on_verify?} ->
-        false
-
-      {nil = _forward_plug, false = _warn_on_verify?} ->
-        true
-
-      {{router, script_name}, false = _warn_on_verify?} ->
-        if function_exported?(router, :__routes__, 0) do
-          match_route?(router, split_path -- script_name)
-        else
-          true
-        end
-
-      :error ->
-        false
-    end
-  end
 
   defp build_route(route_ast, sigil_p, env, endpoint_ctx, router) do
     config = Module.get_attribute(env.module, :phoenix_verified_config, [])


zachdaniel:
I've used `Macro.escape` to handle that, I'm *pretty* sure that's fine, but if not happy to change it.

zachdaniel:
```
==> phoenix
Compiling 1 file (.ex)
Generated phoenix app
==> tunez
Compiling 20 files (.ex)
[]
[
  path: "/auth",
  as: :auth,
  controller: TunezWeb.AuthController,
  not_found_plug: nil,
  resources: [Tunez.Accounts.User]
]
[
  path: "/auth",
  as: :auth,
  controller: TunezWeb.AuthController,
  not_found_plug: nil,
  resources: [Tunez.Accounts.User]
]
Generated tunez app
  GET     /                                      TunezWeb.Artists.IndexLive nil
  GET     /artists/new                           TunezWeb.Artists.FormLive :new
  GET     /artists/:id                           TunezWeb.Artists.ShowLive nil
  GET     /artists/:id/edit                      TunezWeb.Artists.FormLive :edit
  GET     /artists/:artist_id/albums/new         TunezWeb.Albums.FormLive :new
  GET     /albums/:id/edit                       TunezWeb.Albums.FormLive :edit
  *       /gql/playground                        Absinthe.Plug.GraphiQL [schema: TunezWeb.GraphqlSchema, interface: :playground]
  *       /gql                                   Absinthe.Plug [schema: TunezWeb.GraphqlSchema]
  *       /api/json/swaggerui                    OpenApiSpex.Plug.SwaggerUI [path: "/api/json/open_api", default_model_expand_depth: 4]
  GET     /api/json/artists/:id                  TunezWeb.AshJsonApiRouter Tunez.Music.Artist.read
  GET     /api/json/artists                      TunezWeb.AshJsonApiRouter Tunez.Music.Artist.search
  POST    /api/json/artists                      TunezWeb.AshJsonApiRouter Tunez.Music.Artist.create
  PATCH   /api/json/artists/:id                  TunezWeb.AshJsonApiRouter Tunez.Music.Artist.update
  DELETE  /api/json/artists/:id                  TunezWeb.AshJsonApiRouter Tunez.Music.Artist.destroy
  GET     /api/json/artists/:id/albums           TunezWeb.AshJsonApiRouter Tunez.Music.Artist.read
  POST    /api/json/albums                       TunezWeb.AshJsonApiRouter Tunez.Music.Album.create
  PATCH   /api/json/albums/:id                   TunezWeb.AshJsonApiRouter Tunez.Music.Album.update
  DELETE  /api/json/albums/:id                   TunezWeb.AshJsonApiRouter Tunez.Music.Album.destroy
  POST    /api/json/users/register               TunezWeb.AshJsonApiRouter Tunez.Accounts.User.register_with_password
  POST    /api/json/users/sign_in                TunezWeb.AshJsonApiRouter Tunez.Accounts.User.sign_in_with_password
  *       /auth                                  AshAuthentication.Phoenix.StrategyRouter [path: "/auth", as: :auth, controller: TunezWeb.AuthController, not_found_plug: nil, resources: [Tunez.Accounts.User]]
  GET     /sign-out                              TunezWeb.AuthController :sign_out
  GET     /sign-in                               AshAuthentication.Phoenix.SignInLive :sign_in
  GET     /reset                                 AshAuthentication.Phoenix.SignInLive :reset
  GET     /register                              AshAuthentication.Phoenix.SignInLive :register
  GET     /password-reset/:token                 AshAuthentication.Phoenix.ResetLive :reset
  GET     /dev/dashboard/css-:md5                Phoenix.LiveDashboard.Assets :css
  GET     /dev/dashboard/js-:md5                 Phoenix.LiveDashboard.Assets :js
  GET     /dev/dashboard                         Phoenix.LiveDashboard.PageLive :home
  GET     /dev/dashboard/:page                   Phoenix.LiveDashboard.PageLive :page
  GET     /dev/dashboard/:node/:page             Phoenix.LiveDashboard.PageLive :page
  *       /dev/mailbox                           Plug.Swoosh.MailboxPreview []
  WS      /live/websocket                        Phoenix.LiveView.Socket
  GET     /live/longpoll                         Phoenix.LiveView.Socket
  POST    /live/longpoll                         Phoenix.LiveView.Socket
```

Example of the output from one of our demo apps with the AshJsonApi routes expanded. Super useful 😄

SteffenDE:
Lovely! I’m not an expert on the router / verified routes code, so I’ll leave the review to José :)

BartOtten:
Great to see the forwarded routes in route info, nice touch with the label!

I wonder why the PR merges 2 (arbitrary) fields into the original route instead of the plug simply return full Phoenix.Router.Route structs (let the plug handle the merge). This way there is no limit in providing route information (e.g. :private, :metadata) by plugs. 

anniversary today so I’m out.

zachdaniel:
The primary reason for that was to limit the impact/ability to do it "wrong". For example we wouldn't want someone to put a plug that didn't map to reality there. Or to set pkug_opts to something other than what they were set to on the forwarded router.

It also makes it harder to write one of these if you've gotta know what all of those fields are and what they mean. So I think it always has to be merged with the existing route to avoid having to understand all of it.

I didn't personally have a compelling case for setting the other fields, figured it would be best to expand on it iteratively based on real needs

josevalim:
Thank you for the PR. We should definitely split this PR in two.

1. We already have a `:mfa` field in `:metadata` which can be used to show something else both in logs and the console formatter. Have you tried it? I would recommend sticking to it instead of label, unless we have evidence that adding a new approach is necessary.

2. It is highly unlikely we will add anything that does more work at compile-time, as compile time dependencies have a huge impact in large projects. So the `Code.ensure_compiled` is definitely a no go. But perhaps `Code.ensure_loaded` does the trick, as when the code is verified, all modules should have been compiled? I also have concerns about adding new elements to the `__verified_routes__` tuple, as those can increase compile-time, but given this is forward-only, it may be fine. Alternatively, we pass another field forward, such as metadata, so there is more explicit control (instead of assuming all of `plug_opts` is important).

lib/phoenix/router/console_formatter.ex
```diff
@@ -10,14 +10,36 @@ defmodule Phoenix.Router.ConsoleFormatter do
...
     column_widths = calculate_column_widths(router, routes, endpoint)
 
     routes
     |> Enum.map_join("", &format_route(&1, router, column_widths))
     |> Kernel.<>(format_endpoint(endpoint, router, column_widths))
   end
 
+  defp expand_dynamic_routes(routes) do
+    Enum.flat_map(routes, fn route -> 
+      if function_exported?(route.plug, :phoenix_routes, 1) do
```
josevalim, line 27: I believe Code.ensure_loaded? will be required here.

lib/phoenix/router/console_formatter.ex
```diff
@@ -10,14 +10,36 @@ defmodule Phoenix.Router.ConsoleFormatter do
...
+        route.plug_opts
+        |> route.plug.phoenix_routes()
+        |> Enum.map(fn nested_route ->
```
josevalim, line 30: There is an underlying assumption here that the routes are the same, except path and verb, and I am not sure if it necessarily holds. For example, this PR could be used to deal with nested Phoenix Routers, right? But it wouldn't work with the current contract.

zachdaniel: I had actually thought that nested phoenix routers were already supported. Perhaps in this case @BartOtten's instinct is correct then and we should have it just return a list of`Phoenix.Route.t()`?

lib/phoenix/router/console_formatter.ex
zachdaniel: I had thought so as well but it worked without it. No problem adding it back in.

lib/phoenix/router/console_formatter.ex
josevalim: I am not sure. But we can't return routes because they are private, so you wouldn't be able to create them.

zachdaniel:
> We already have a :mfa field in :metadata which can be used to show something else both in logs and the console formatter. Have you tried it? I would recommend sticking to it instead of label, unless we have evidence that adding a new approach is necessary.

I don't think that `mfa` would be sufficient as we just show the module from that mfa. In our case we want to show `The.Resource.action`. We could have a callback like `route_label/1` on the new behaviour, but that adds a layer of indirection as it has to "look it back up" to figure out what to display. 

Or maybe I'm misunderstanding actually, do you just mean we can put a value in `metadata`? Like `metadata[:label]`? That makes sense to me 👍 


> It is highly unlikely we will add anything that does more work at compile-time, as compile time dependencies have a huge impact in large projects. So the Code.ensure_compiled is definitely a no go. But perhaps Code.ensure_loaded does the trick, as when the code is verified, all modules should have been compiled? I also have concerns about adding new elements to the __verified_routes__ tuple, as those can increase compile-time, but given this is forward-only, it may be fine. Alternatively, we pass another field forward, such as metadata, so there is more explicit control (instead of assuming all of plug_opts is important).

I think in the vast majority of cases, anyone doing dynamic routing will be doing it based off of the opts passed to the plug. We could have a callback on the behavior that asks it to "extract" the minimum data it needs from its opts? Since this only happens for routes that are forwarded plugs that the impact would be small enough as you mentioned. I'll swap the `Code.ensure_compiled` for `Code.ensure_loaded` that was just a brain fart on my end.

lib/phoenix/router/console_formatter.ex
BartOtten: 1. Why not make them public? They seems to be pretty stable last few years.

3. We could use the forward route, add augmented variants of it and return all of them or only the variant, no? This is the way Routex does it an it seems not to cause issues.

@zachdaniel People would be able to screw things over by modifying the wrong fields, but so can they do with any code as I tend do demonstrate too often ;)

zachdaniel: My gut says it's better to make a new, more restricted set of fields here, as opposed to "just give me your routes". Even if we have to add some fields that would make this work with nested Phoenix routers. Is there something not in the fields we're returning now that you have need of for your use of this feature?

zachdaniel:
I've made the `Code.ensure_loaded` changes as suggested.

I also reused `metadata` instead of adding a special `label` field (easy enough to change to something else, just wanted to see what it looked like). It works well enough but I did see a note that metadata is used for telemetry events, so this would also increase the size of the telemetry event metadata I think which could be its own problem.

zachdaniel:
Oh, for this:

> Thank you for the PR. We should definitely split this PR in two.

Do you mean we should split the verified routes functionality and the `ConsoleFormatter` functionality? I'm happy to do that. I imagine the `ConsoleFormatter` changes are lower risk and so we can start there?

josevalim:
If you use MFA, I think we show the module and plug opts? So your options could return plug and plug_opts instead of a new label, then it should be a single PR.

zachdaniel:
I'm not sure I understand. Right now it does this:

```elixir
    log_module =
      case metadata[:mfa] do
        {mod, _fun, _arity} -> mod
        _ -> plug
      end

    String.pad_leading(route_name, route_name_len) <>
      "  " <>
      String.pad_trailing(verb, verb_len) <>
      "  " <>
      String.pad_trailing(path, path_len) <>
      "  " <>
      "#{inspect(log_module)}\n"
```

So it just shows the mfa's module or the plug. But in our case it would point at the same plug for all of the routes. I'd be limited to showing just a module name as the label. I can do that, but the output would be less useful than what's in my current PR, I could only show the resource module.

lib/phoenix/router/console_formatter.ex
```diff
@@ -120,13 +144,20 @@ defmodule Phoenix.Router.ConsoleFormatter do
...
+        label -> label
+      end
+
     String.pad_leading(route_name, route_name_len) <>
       "  " <>
       String.pad_trailing(verb, verb_len) <>
       "  " <>
       String.pad_trailing(path, path_len) <>
       "  " <>
-      "#{inspect(log_module)} #{inspect(plug_opts)}\n"
```
josevalim, line 129: This is printing the plug_opts, no?!

zachdaniel: Yes, but for all of our generated routes wouldn't the plug opts be *the same value*? Ohhhhhhhhhhhhh. You mean allow them to return *new* `plug_opts` to be displayed instead. So I'd return `plug_opts: [resource: Resource, action: Action]`. Sorry, I didn't understand.

zachdaniel:
Okay, using `plug_opts` looks like this which is reasonable:

```elixir
  GET     /                                      TunezWeb.Artists.IndexLive nil
  GET     /artists/new                           TunezWeb.Artists.FormLive :new
  GET     /artists/:id                           TunezWeb.Artists.ShowLive nil
  GET     /artists/:id/edit                      TunezWeb.Artists.FormLive :edit
  GET     /artists/:artist_id/albums/new         TunezWeb.Albums.FormLive :new
  GET     /albums/:id/edit                       TunezWeb.Albums.FormLive :edit
  *       /gql/playground                        Absinthe.Plug.GraphiQL [schema: TunezWeb.GraphqlSchema, interface: :playground]
  *       /gql                                   Absinthe.Plug [schema: TunezWeb.GraphqlSchema]
  *       /api/json/swaggerui                    OpenApiSpex.Plug.SwaggerUI [path: "/api/json/open_api", default_model_expand_depth: 4]
  GET     /api/json/artists/:id                  TunezWeb.AshJsonApiRouter [resource: Tunez.Music.Artist, action: :read]
  GET     /api/json/artists                      TunezWeb.AshJsonApiRouter [resource: Tunez.Music.Artist, action: :search]
  POST    /api/json/artists                      TunezWeb.AshJsonApiRouter [resource: Tunez.Music.Artist, action: :create]
  PATCH   /api/json/artists/:id                  TunezWeb.AshJsonApiRouter [resource: Tunez.Music.Artist, action: :update]
  DELETE  /api/json/artists/:id                  TunezWeb.AshJsonApiRouter [resource: Tunez.Music.Artist, action: :destroy]
  GET     /api/json/artists/:id/albums           TunezWeb.AshJsonApiRouter [resource: Tunez.Music.Artist, action: :read]
  POST    /api/json/albums                       TunezWeb.AshJsonApiRouter [resource: Tunez.Music.Album, action: :create]
  PATCH   /api/json/albums/:id                   TunezWeb.AshJsonApiRouter [resource: Tunez.Music.Album, action: :update]
  DELETE  /api/json/albums/:id                   TunezWeb.AshJsonApiRouter [resource: Tunez.Music.Album, action: :destroy]
  POST    /api/json/users/register               TunezWeb.AshJsonApiRouter [resource: Tunez.Accounts.User, action: :register_with_password]
  POST    /api/json/users/sign_in                TunezWeb.AshJsonApiRouter [resource: Tunez.Accounts.User, action: :sign_in_with_password]
  *       /auth                                  AshAuthentication.Phoenix.StrategyRouter [path: "/auth", as: :auth, controller: TunezWeb.AuthController, not_found_plug: nil, resources: [Tunez.Accounts.User]]
  GET     /sign-out                              TunezWeb.AuthController :sign_out
  GET     /sign-in                               AshAuthentication.Phoenix.SignInLive :sign_in
  GET     /reset                                 AshAuthentication.Phoenix.SignInLive :reset
  GET     /register                              AshAuthentication.Phoenix.SignInLive :register
  GET     /password-reset/:token                 AshAuthentication.Phoenix.ResetLive :reset
  GET     /dev/dashboard/css-:md5                Phoenix.LiveDashboard.Assets :css
  GET     /dev/dashboard/js-:md5                 Phoenix.LiveDashboard.Assets :js
  GET     /dev/dashboard                         Phoenix.LiveDashboard.PageLive :home
  GET     /dev/dashboard/:page                   Phoenix.LiveDashboard.PageLive :page
  GET     /dev/dashboard/:node/:page             Phoenix.LiveDashboard.PageLive :page
  *       /dev/mailbox                           Plug.Swoosh.MailboxPreview []
  WS      /live/websocket                        Phoenix.LiveView.Socket
  GET     /live/longpoll                         Phoenix.LiveView.Socket
  POST    /live/longpoll                         Phoenix.LiveView.Socket
```

josevalim:
This is much better, thank you! I think we should dog food our API a bit more. Here is what I propose:

1. Introduce a `formatted_routes/1` function to the behaviour, implemented it for Phoenix.Router itself, and have the `ConsoleFormatter` simply call `formatted_routes/1`and then print the results

2. See if we can do the same for `verified_routes`, i.e. introduce a `verified_route?` or similar to the behaviour and have `Phoenix.VerifiedRoutes` call it instead. In other words, instead of asking the router for its routes so we can verify them, we delegate the whole verification logic, and we implement it for Phoenix.Router too

josevalim:
Btw, you should rebase before starting the changes above, as I pushed some changes. :)

lib/phoenix/router/console_formatter.ex
BartOtten: :private and :meta are the primary adaptable fields

zachdaniel:
@josevalim Want to check my understanding here:

Are you saying that the target plug or Phoenix router would have a function like:

```
def formatted_routes(opts) do
  ...
end
```

that returns something like this?

```
"""
GET     /api/json/artists/:id                  TunezWeb.AshJsonApiRouter [resource: Tunez.Music.Artist, action: :read]
  GET     /api/json/artists                      TunezWeb.AshJsonApiRouter [resource: Tunez.Music.Artist, action: :search]
  POST    /api/json/artists                      TunezWeb.AshJsonApiRouter [resource: Tunez.Music.Artist, action: :create]
"""
```

If so:

1. the content is spaced according to the longest width in any given section. We don't know how wide to make those spaces unless we also pass in widths. As a result, much becomes part of a public API, like "how many columns there are in mix phx.routes" and "which order information is displayed in phx.routes" because the behavior would have to return like a list or map of column values to be formatted and properly spaced out at the end before being converted to a string (or iolist per recent changes).

2. It seems just...more complicated than this solution where someone who has a dynamic router can just return a very simple data structure. It is very hard to get it wrong as an implementor with the current design. If we want to dogfood it with nested phoenix routers are we *sure* we can't just expand the keys that can be returned in the behaviour's data structure? Or just add an explicit check for "if it's a phoenix router, then call `__routes__` on it?

josevalim:
1. Almost! It would return the map you return today with `path`, `verb`, `plug` and `plug_opts` (and perhaps an optional helper field). The function returns the routes, the console formatter formats them.

2. I don't know what is the contract yet, but we should at least explore it. Maybe there is an easy contract, maybe there isn't. But having a general API than something that returns "pseudo routes" could be even more powerful. The more we dogfood our APIs, the better. If there is a code path that will be only invoked when there is a certain extension, then it is likely that it will have bugs or grow in ad-hoc ways.

zachdaniel:
Okay, so I think I might understand what you're getting at here now, willing to give it a shot. With that said, I worry about making these changes without clearly confirming my understanding.

```elixir
defmodule FormattedRoute do
  defstruct [:path, :verb, :plug, :plug_opts]
  # I believe we will need more here, like `metadata`. Have to check what all is used in the ConsoleFormatter
  @type t :: %FormattedRoute{path: String.t(), verb: String.t(), plug: module(), plug_opts: term()}, 
end

@type plug_opts :: term()
@type path :: [String.t()]

@callback formatted_routes(plug_opts) :: [FormattedRoute.t()]
@callback verified_route?(plug_opts, path) :: boolean
```

And you want me to implement this behaviour for `Phoenix.Router`.

josevalim:
Good call. I don't think we need a new struct, so:

```
@type plug_opts :: term()
@type path :: [String.t()]

@callback formatted_routes(plug_opts) :: [%{path: String.t(), verb: String.t(), plug: module(), plug_opts: term(), optional(:helper) => atom()}]
@callback verified_route?(plug_opts, path) :: boolean
```

The behaviour will be defined for Phoenix.Router and will be implemented by every router created by `Phoenix.Router`.

zachdaniel:
As I work on this, one interesting bit is that the custom route verification logic here:

```elixir
  def verify_route(plug, opts, path) do
    opts
    |> plug.formatted_routes()
    |> Enum.map(fn route ->
      case Path.split(route.path) do
        ["/" | rest] -> rest
        path -> path
      end
    end)
    |> Enum.any?(&match_path?(&1, path))
  end

  defp match_path?([], []), do: true
  defp match_path?([], _), do: false
  defp match_path?(_, []), do: false

  defp match_path?([":" <> _ | rest_route], [_ | rest_path]) do
    match_path?(rest_route, rest_path)
  end

  defp match_path?(["_" <> _ | rest_route], [_ | rest_path]) do
    match_path?(rest_route, rest_path)
  end

  defp match_path?(["*" <> _], _) do
    true
  end

  defp match_path?([same | rest_path], [same | rest_route]) do
    match_path?(rest_path, rest_route)
  end

  defp match_path?(_, _) do
    false
  end
```

would not be used by `Phoenix.Router`, because it has a more optimized compile-time route verification function. So we have two options:

1. not have this function at all. We can make it the other router's problem to have this logic
2. make the `verified_route?/2` callback optional and fall back to this logic:

```elixir
  def verify_route(plug, opts, path) do
    opts
    |> plug.formatted_routes()
    |> Enum.map(fn route ->
      case Path.split(route.path) do
        ["/" | rest] -> rest
        path -> path
      end
    end)
    |> Enum.any?(&match_path?(&1, path))
  end

  defp match_path?([], []), do: true
  defp match_path?([], _), do: false
  defp match_path?(_, []), do: false

  defp match_path?([":" <> _ | rest_route], [_ | rest_path]) do
    match_path?(rest_route, rest_path)
  end

  defp match_path?(["_" <> _ | rest_route], [_ | rest_path]) do
    match_path?(rest_route, rest_path)
  end

  defp match_path?(["*" <> _], _) do
    true
  end

  defp match_path?([same | rest_path], [same | rest_route]) do
    match_path?(rest_path, rest_route)
  end

  defp match_path?(_, _) do
    false
  end
```

Personally I'm in favor of #2, as I can write tests to verify the behavior (likely doc tests), and it simplifies the implementation for 3rd party plugs that want to use this, as they can just implement `formatted_routes/1` and benefit from a standard implementation of `verified_route?/2`. It does, however, have the drawback of not being a function that Phoenix itself ever calls aside from in test.

Thoughts?

josevalim:
I'd go with 1. We should provide a generic API that we dogfood. If people are implementing over and over, then we abstract it. We shouldn't be guessing. :)

zachdaniel:
Just pushed up with #2, but will remove. I don't really think there is much variation on "does this path match this route", but I will defer 😄

zachdaniel:
Okay, the latest implementation is up. I'm not sure if there is additional testing necessary at this point given that the `Phoenix.Router` and `mix phx.routes` is already tested and uses this new code path.

One thing that would make sense to test, however: I think based on what we were saying earlier that nested phoenix routers were *not* expanded in `mix phx.routes` whereas now they will be. I'm assuming this is desired behavior? I would have expected a test for `mix phx.routes` with a `forward` in it to fail, but perhaps that code path isn't tested. Should I write a test for that new behavior?

zachdaniel:
Oh, and I had to add the `metadata` to `formatted_routes` since that is used :)

josevalim:
> Oh, and I had to add the metadata to formatted_routes since that is used by the console formatter :)

We should not add metadata. The goal of `formatted_routes` is to encapsulate these concerns, so other routers need to provide as little information as possible, and we keep the contract minimal. `formatted_routes` should return routes as they are formatted, not as they are executed.

lib/phoenix/router.ex
```diff
@@ -537,6 +538,74 @@ defmodule Phoenix.Router do
...
+            else
+              false
+            end
+
+          :error ->
+            false
+        end
+      end
+
+      @doc """
```
josevalim, line 581: We already define this elsewhere, no?

zachdaniel:
What I was solving for by adding it was this code here: 

```elixir
    %{
      verb: verb,
      path: path,
      plug: plug,
      metadata: metadata,
      plug_opts: plug_opts,
      helper: helper
    } = route

    verb = verb_name(verb)
    route_name = route_name(router, helper)
    {verb_len, path_len, route_name_len} = column_widths

    log_module =
      case metadata[:mfa] do
        {mod, _fun, _arity} -> mod
        _ -> plug
      end

    String.pad_leading(route_name, route_name_len) <>
      "  " <>
      String.pad_trailing(verb, verb_len) <>
      "  " <>
      String.pad_trailing(path, path_len) <>
      "  " <>
      "#{inspect(log_module)} #{inspect(plug_opts)}\n"
```

I think I see what you're saying. Do you mean that in `formatted_routes` in the `Phoenix.Router` it should return the `plug` from `metadata[:mfa]` if it is set?

lib/phoenix/router/console_formatter.ex
```diff
@@ -19,7 +22,28 @@ defmodule Phoenix.Router.ConsoleFormatter do
...
   end
 
-  defp format_endpoint(nil, _), do: ""
+  defp expand_dynamic_routes(routes) do
+    Enum.flat_map(routes, fn route ->
+      Code.ensure_loaded(route.plug)
+
+      if function_exported?(route.plug, :phoenix_routes, 1) do
+        route.plug_opts
+        |> route.plug.phoenix_routes()
```
josevalim, line 31: This is not a contract in Phoenix anywhere else, so this should be removed in favor of formatted_routes. I'd argue it shuold most likely be moved to Phoenix.Router implementation.

lib/phoenix/router.ex
```diff
@@ -537,6 +538,74 @@ defmodule Phoenix.Router do
...
+            :plug,
+            :plug_opts,
+            :metadata,
+            :helper
+          ])
+        end)
+      end
+
+      def verified_route?(_, split_path) do
+        case __verify_route__(split_path) do
```
josevalim, line 560: We should keep the amount of generated code to a minimum. We should probably delegate this to a `Phoenix.Router.__verified_route__` or similar.

lib/phoenix/router/console_formatter.ex
zachdaniel: Sorry, that was a mistake that I left it in. It should be replaced with `formatted_routes`. Will do that now.

I thought about pulling it into `formatted_routes` in `Phoenix.Router`, but it also seemed like a potential benefit for it to be expanded here. In the case that an implementer of this returned a route with a plug that could be expanded, they wouldn't have to know to do that. I can see a case for both, so I'll move it into `Phoenix.Router` 👍

josevalim:
> I think I see what you're saying. Do you mean that in formatted_routes in the Phoenix.Router it should return the plug from metadata[:mfa] if it is set?

Yes. The goal is to define the correct contracts instead of automatically making the `Route` fields our APIs. To this extent, perhaps we should even rename `plug` and `plug_opts` to `module` and `arg`? Less chance for confusion.

lib/phoenix/router/plug_with_routes.ex
```diff
@@ -0,0 +1,16 @@
...
+          required(:plug) => module(),
+          required(:plug_opts) => any(),
+          optional(:metadata) => map(),
+          optional(:helper) => atom()
+        }
+
+  @type path :: [String.t()]
+
+  @callback formatted_routes(plug_opts()) :: [formatted_route()]
+  @callback verified_route?(plug_opts(), path) :: boolean()
```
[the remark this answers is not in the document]
zachdaniel: 🤔 isn't the whole point that people who are implementing plugs that are *not* `Phoenix.Router`s can get the benefit of verified routes?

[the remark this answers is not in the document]
zachdaniel: At least, that is the point for me. We're using `Plug.Router`.

lib/phoenix/router.ex
zachdaniel: Sorry, copy paste error.

zachdaniel:
> perhaps we should even rename plug and plug_opts to module and arg

What if we just call it something else entirely and make *them* combine them into a string? like `label` or `name`, and in Phoenix's formatted route we return `label: "#{plug} #{inspect(plug_opts)}"`

zachdaniel:
We could do the same with helper, so that all you return is `%{name: name, verb: verb, path: path, label: label: label}`. Things other than `Phoenix.Router` don't have a `helper` anyway.

josevalim:
I am happy with %{verb: verb, path: path, label: label: label}. Keep helper “off the books” as it is deprecated.

lib/phoenix/router/plug_with_routes.ex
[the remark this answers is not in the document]
josevalim: It is a callback, it means can now inject custom routers where a Phoenix.Router is expected as long as it implements the behavior.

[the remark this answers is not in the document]
zachdaniel: Right, but in my custom plug to use this I'd say:

```elixir
@behaviour Phoenix.Router
@behaviour Phoenix.Router.PlugWithRoutes
```

which seems pretty strange. The thing itself doesn't "behave like a Phoenix.Router".

zachdaniel:
We'll need `name` as well, which is what `helper` is mapped too.

`name | verb | path | label`

josevalim:
What are you using the `name` for? Because our plan is to remove it in the long term.

lib/phoenix/router/plug_with_routes.ex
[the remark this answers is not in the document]
josevalim: > The thing itself doesn't "behave like a Phoenix.Router".

Why would you say it doesn't behave like a Phoenix router?

From this PR, it will be allowed in `mix phx.routes YourPlug` and you can also use it for verified routes. So it sounds like it behaves like the router to me?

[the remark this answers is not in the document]
josevalim: That's a good point. Perhaps we should move the callback to `Phoenix.VerifiedRoutes`? And then say `mix phx.routes` works with anything that implements `VerifiedRoutes`?

josevalim:
> I'm not using the name for anything in my usage, but would need it to keep compatibility with current Phoenix behavior in this feature.

Right. So let's keep it as :helper but not document it in the `@callback`. We will keep it exclusively behind the scenes for now for backwards compat.

lib/phoenix/router/plug_with_routes.ex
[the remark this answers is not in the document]
zachdaniel: 👍 sounds good.

zachdaniel:
Working on these changes now. One last thing: it would make sense to me if both callbacks were optional. You may be able to do some route verification without necessarily being able to display routes in `mix phx.routes` and vice versa. Thoughts?

josevalim:
In such case you can just return empty routes. I think we don't need to make them optional.

zachdaniel:
👍 good point.

zachdaniel:
Alright, made the requested changes and added documentation to the new callbacks since it seems we are nearing the final design. Should I also add `@typedoc` to the new types?

lib/phoenix/router.ex
```diff
@@ -358,6 +358,7 @@ defmodule Phoenix.Router do
       unquote(prelude(opts))
       unquote(defs())
       unquote(match_dispatch())
+      unquote(plug_with_routes())
```
josevalim, line 361:
```suggestion
      unquote(verified_routes())
```

lib/phoenix/router.ex
```diff
@@ -1282,4 +1298,62 @@ defmodule Phoenix.Router do
       Map.delete(metadata, :conn)
     end
   end
+
+  def __formatted_routes__(router) do
```
josevalim, line 1302:
```suggestion
  @doc false
  def __formatted_routes__(router) do
```

lib/phoenix/router.ex
```diff
@@ -1282,4 +1298,62 @@ defmodule Phoenix.Router do
...
+            verb: route.verb,
+            path: route.path,
+            label: label
+          }
+        ]
+      end
+    end)
+  end
+
+  def __verified_route__?(router, split_path) do
```
josevalim, line 1338:
```suggestion
  @doc false
  def __verified_route__?(router, split_path) do
```

lib/phoenix/router.ex
```diff
@@ -1282,4 +1298,62 @@ defmodule Phoenix.Router do
...
+      {nil = _forward_plug, false = _warn_on_verify?} ->
+        true
+
+      {{router, script_name, plug_opts}, false = _warn_on_verify?} ->
+        Code.ensure_loaded(router)
+
+        if function_exported?(router, :verified_route?, 2) do
+          router.verified_route?(plug_opts, split_path -- script_name)
+        else
+          false
```
josevalim, line 1352: As per the current implementation:

```suggestion
          true
```

lib/phoenix/verified_routes.ex
```diff
@@ -256,7 +304,9 @@ defmodule Phoenix.VerifiedRoutes do
   @doc false
   def __verify__(routes) when is_list(routes) do
     Enum.each(routes, fn %__MODULE__{} = route ->
-      unless match_route?(route.router, route.test_path) do
+      test_path = split_test_path(route.test_path)
+
+      unless route.router.verified_route?(route.router, test_path) do
```
josevalim, line 309: Empty plug_options, i.e. a list, would probably be a "safer" default?

```suggestion
      unless route.router.verified_route?([], test_path) do
```

lib/phoenix/verified_routes.ex
```diff
@@ -238,6 +267,25 @@ defmodule Phoenix.VerifiedRoutes do
...
+
+  @doc """
+  Returns the necessary information about routes for display in `mix phx.routes`
+  """
+  @callback formatted_routes(plug_opts()) :: [formatted_route()]
+
+  @doc """
+  Returns `true` if the path can be verified, and false if not.
+  """
+  @callback verified_route?(plug_opts(), path) :: boolean()
```
josevalim, line 287:
```suggestion
  @doc """
  Returns the necessary information about routes for display in `mix phx.routes`.
  
  The `plug_opts` is typically only passed when the router is mounted within
  a `Phoenix.Router`. Otherwise it defaults to `[]`.
  """
  @callback formatted_routes(plug_opts()) :: [formatted_route()]

  @doc """
  Returns `true` if the path is verified, and false if not.
  
  The `plug_opts` is typically only passed when the router is mounted within
  a `Phoenix.Router`. Otherwise it defaults to `[]`.
  """
  @callback verified_route?(plug_opts(), [String.t()]) :: boolean()
```

josevalim:
One last batch with only nits and we can ship it.

lib/phoenix/verified_routes.ex
zachdaniel: Ah, actually I got that wrong. It should be `route.plug_opts`, as we do have that information here.

zachdaniel: nope, nevermind, we don't. 🤔 That does seem like pretty important information for plugs to verify their routes?

zachdaniel: Okay, this is a different thing, nvm 😄.

zachdaniel:
Alright, all set 😄 thank you for all your help and patience getting this over the line. This is going to be a *huge* quality of life improvement for AshJsonApi and AshAuthentication users 🙇

josevalim (approved):

SteffenDE:
Very cool!

BartOtten:
Thanks @zachdaniel! Credits for the work put into it. And ofc @josevalim for thoroughly reviewing and advising.

PR #6126, opened 2025-03-19
Prevent duplicating core components between phoenix and installer

Gazler:
This commit copies the core_components.ex file from the phoenix application into the installer when calling `compile` or `hex.publish` . This file is now `.gitignore` so that it isn't commited to version control.

The file has also explicitly been added to the list of files in `mix.exs` to ensure it isn't missing when publishing the package.


installer/mix.exs
```diff
@@ -65,9 +66,23 @@ defmodule Phx.New.MixProject do
     ]
   end
 
+  def aliases do
```
SteffenDE, line 69: That can be public, can’t it?

installer/mix.exs
```diff
@@ -65,9 +66,23 @@ defmodule Phx.New.MixProject do
...
     [
       source_url_pattern: "#{@scm_url}/blob/v#{@version}/installer/%{path}#L%{line}"
     ]
   end
+
+  defp copy_core_components(_) do
+    source = Path.join([__DIR__, "..", "priv", "templates", "phx.gen.live", "core_components.ex"])
+    destination_dir = Path.join([__DIR__, "templates", "phx_web", "components"])
+    destination = Path.join(destination_dir, "core_components.ex")
+    File.cp(source, destination)
```
SteffenDE, line 86: Should we `cp!`?

SteffenDE:
Lovely!

PR #6127, opened 2025-03-20
Fix endpoint binding comment location in installer templates

chulkilee:
When not in the docker container, the current code produces

```elixir
config :foo, FooWeb.Endpoint,# Binding to loopback ipv4 address prevents access from other machines.
  # This makes make the service accessible from any network interface.
```

which turns to the following.

```elixir
# Binding to loopback ipv4 address prevents access from other machines.
config :foo, FooWeb.Endpoint,
  # This makes make the service accessible from any network interface.
```

To keep the comment on the next line, add new line after else.


josevalim (approved):

SteffenDE:
Thank you! 🙌🏻

chulkilee:
@SteffenDE what about backporting this to 1.7 branch? Do you want me to create a new PR targeting to the branch? Or will a maintainer cherry-pick the change and push to the branch?

SteffenDE:
We’re very close the first 1.8 RC, I don’t think it’s worth backporting as it’s unlikely we’ll do another 1.7 release if there’s nothing urgent to be fixed.

PR #6128, opened 2025-03-20
add test for new verified routes behavior for mix phx.routes

SteffenDE:
@zachdaniel I needed to make a little adjustment. While the type for formatted_route defines verb, path and label

```elixir
@type formatted_route :: %{
          required(:verb) => String.t(),
          required(:path) => String.t(),
          required(:label) => String.t()
        }
```

The code expected plug_opts instead and didn't handle the label. I believe this was just an oversight, but please double check.


josevalim (approved):
Beautiful

PR #6133, opened 2025-03-21
Add a warning about changing the layout to :root

jberling:
If you change the layout to :root, LiveView starts to behave weirdly. Really hard to debug. I thought at least a warning could be in place.

See: https://elixirforum.com/t/how-do-you-debug-state-changes-using-liveviews/70044/21


jberling:
Not needed. This is handled in https://github.com/phoenixframework/phoenix_live_view/pull/3721

PR #6135, opened 2025-03-21
Set __changed__: nil on controller render for LiveView compatibility

### LINKED ISSUE #5726: Does not seem possible to use assign in custom error html
### Environment

* Elixir version (elixir -v): 15
* Phoenix version (mix deps): main
* Operating system: mac

### Actual behavior

If you try to use assign in a custom error html like this:

```elixir
defmodule ExampleWeb.ErrorHTML do
  use ExampleWeb, :html

  def render("404.html" = template, assigns) do
    assigns =
      assign(assigns,
        message: hello
      )

    ~H"""
    <h1 class="text-2xl font-bold tracking-tight text-gray-900 sm:text-4xl">
      <%= @status %> <%= @message %>
    </h1>
    """
  end
end
```

You get an error like:

```
(ArgumentError) assign/3 expects a socket from Phoenix.LiveView/Phoenix.LiveComponent  or an assigns map from Phoenix.Component as first argument ...
```

### Expected behavior

This should be a valid assigns so I think it should work?

josevalim:
Closes #5726

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -546,13 +546,13 @@ defmodule Phoenix.Controller do
 
   def put_view(%Plug.Conn{} = conn, module) do
     raise(AlreadySentError, """
-      the response was already sent.
+    the response was already sent.
 
-          Status code: #{conn.status}
-          Request path: #{conn.request_path}
-          Method: #{conn.method}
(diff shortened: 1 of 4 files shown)


josevalim:

PR #6132, opened 2025-03-21
Deprecate soft-deprecated controller entry points

josevalim:

PR #6137, opened 2025-03-23
Correct wording of layouts without modules deprecation

rmoorman:
Specifying layouts without module is now deprecated (See #6132). However the changelog incorrectly states that layouts with modules are deprecated.


josevalim:

PR #6142, opened 2025-03-26
Phoenix.Router based socket routing

LostKobrakai:


lib/phoenix/endpoint.ex
```diff
@@ -1060,11 +942,11 @@ defmodule Phoenix.Endpoint do
       by `Phoenix.Token`. By default tokens are valid for 2 weeks
 
   """
-  defmacro socket(path, module, opts \\ []) do
-    module = Macro.expand(module, %{__CALLER__ | function: {:socket_dispatch, 2}})
+  defmacro socket(_path, _module, _opts \\ []) do
```
SteffenDE, line 945: We most likely want to still support the old way as usual

lib/phoenix/endpoint.ex
```diff
@@ -411,6 +411,10 @@ defmodule Phoenix.Endpoint do
     quote do
       @behaviour Phoenix.Endpoint
 
+      def __start_sockets__ do
+        unquote(Keyword.get(opts, :start_sockets, []))
+      end
```
SteffenDE, line 416: I don't know yet if I really like having it explicit.

lib/phoenix/endpoint.ex
LostKobrakai: Yeah, I removed stuff in here mostly to make sure I catch everything. Making this propely backwards compatibility preserving will make sense once other questions are dealt with.

lib/phoenix/endpoint.ex
LostKobrakai: Yeah, I'm with you. On the one hand it is really nice how this used to work. 

On the other hand I don't like the amount of stuff that's being configured on the endpoint socket macro. Because there's just a single entry point to "everything" that also means everything needs to somehow be configured right there.

Given routers and endpoints are really only quite loosely coupled there's afaik also no good way to retain the static knowledge of to be started socket modules from the router macro.

PR #6151, opened 2025-03-30
Refresh auth tokens periodically in phx.gen.auth code.

nshafer:
The auth code currently generated by `mix phx.gen.auth` has two small issues that I've encountered:

1. After logging in, a user who maintains their session (either by never closing their browser or
due to things like Session Restore in Firefox) will be forced to reauthenticate after the session
token expires, which defaults to 60 days.

2. Even if the user selects the "remember me" option and the remember me cookie is created, the cookie and
token will only last for 60 days from the time that the authentication first happened.

A user could be happily working seconds before the token is set to expire, and then on the next request be forced to log in again. And no matter what, the user is forced to reauthenticate at least every 60 days.

This is due to the validity of the token being based on when it was inserted into the database, by the `inserted_at` column.

This PR seeks to remedy this by automatically refreshing the generated tokens in the database, with a new column, `refreshed_at`, that is added to the `users_tokens` table. It will also refresh the remember me cookie with the browser so that its `expires` and `max_age` is extended.

You can easily see the differences of the generated code in [this project](https://github.com/nshafer/myapp18/compare/auth_live_orig..auth_live_new) which was generated with
`phx.gen.auth Accounts User users`.
Note: This PR applies the same changes to both live and no-live versions of the auth.
- [`phx.gen.auth --live Accounts User users`](https://github.com/nshafer/myapp18/compare/auth_live_orig..auth_live_new)
- [`phx.gen.auth --no-live Accounts User users`](https://github.com/nshafer/myapp18/compare/auth_nolive_orig..auth_nolive_new)

I've attempted to limit any kind of performance impact of this PR. I've set it to only refresh once every 24 hours by default to limit potentially costly `UPDATE`s to the database to once a day per token by default. The current code already does a join between `users` and `users_tokens`, returning just the fields of `users` and this modifies it only so that all fields of both `users` and `users_tokens` are returned on every request. This is so that the `fetch_current_scope_for_user/2` function can check how long its been since the token was refreshed, and refresh the token and cookie if needed. With the default columns of `users_tokens` this increase is minimal, but we can reduce it by selecting only the `refreshed_at` column if we want.

I also make all changes happen in the main `fetch_current_scope_for_user/2` function, so tokens will not be checked or refreshed on live view connected mounts or live navigations. The first reason is that we wouldn't be able to refresh the cookie from a websocket. The other reason is that it's unnecessary for most situations where a normal request is made before the connected mount anyways. The only scenario I can think of is like a kiosk display that has a live view mounted 24/7, which never does a normal request, just websocket connects. But at least now a dev has the option to cause that kiosk to do a full refresh at least once every 60 days to maintain their auth, instead of it logging out every 2 months regardless.

Security wise, the only impact I can think of is where a users cookies are exfiltrated or intercepted. Instead of the stolen cookie only working for up to 60 days, it would work indefinitely as long as the attacker made a request with it periodically to refresh it. To me this doesn't seem like an escalation of risk or reduction in security, since the channel of attack already implies a much more serious problem. An attacker can have up to 60 days to wreak havoc or steal data with the old cookies anyways. The new "sudo" mechanism is not affected by this change, since `is_sudo?` is based on when the token was created. So an attacker in this scenario would not be able to do anything that requires "sudo", such as change email address or password. The current mechanism where all tokens are deleted when the user changes their email or password still applies. I also left a comment near the module attribute, `@session_refresh_age_in_hours`, that if it's increased to a value greater than the session age, then refreshing is effectively disabled, in case a dev wants to only support short-lived authentication sessions.

I tested that this works with all the other recent changes, such as sudo mode, which is still based on the `inserted_at` of the token. I did fix a small bug where a user session that was restored from the remember me cookie would not persist that preference if the user was required to re-auth for a sudo required page. I also added to the tests to make sure `authenticated_at` is set properly in the `User`.

I attempted to limit the changes to the bare-minimum so that devs could more easily upgrade existing generated code. The main exception is that I took the opportunity to retire `verify_session_token_query/1` from the generated context which always returned {:ok, query} and never an error. My guess is that it was a hold-over from an old version of the function that could return an error. I rewrote this function to `valid_session_token_query/1` which just returns the query. The other change is that I unified the units between the related module attributes settings in `user_token.ex` and `user_auth.ex`. It used to be `@session_validity_in_days 60` in the former, and `@max_age 60 * 60 * 24 * 60` in the latter. I changed it to `@max_session_age_in_days 60` in `user_auth.ex` so that changes in one file can be easily copied to the other. If a dev were inclined to update their already-generated code, then they would just need to create a db migration that adds the `add :refreshed_at, :utc_datetime` column, and then sets it to the current value of `inserted_at`. From there the sessions will start refreshing.

Finally, I added a `prune_user_tokens/0` function in the generated context that will properly remove old and expired tokens from the database. I felt this was very valuable for us to provide so that the deletions happen properly with the different token types without having to understand all of the auth code. This function is not called anywhere by default, but it can be easily hooked up to some kind of periodic background task to keep the database size from ballooning with old useless tokens.

I explored other options for the refresh, such as storing the refresh time in the session and remember me cookies instead of the database. I felt this didn't really result in any kind of improvement, as it would bloat the session and cookies for all requests, instead of just the data returned from the db, and make the code more complicated.

Oh I also fixed a couple things related to scopes that was causing test failures with non-"user" like settings, such as `mix phx.gen.auth Company Person people`. (Speaking of, should the `is_sudo?` stuff be in scopes? Seems like that's a good example of how requests can be scoped, with a `recently_authenticated` check or something.)

Let me know if there's something I'm not aware of with this change, such as for performance impact or security.

Thanks!
Nathan Shafer

### DIFF
--- a/priv/templates/phx.gen.auth/auth.ex
+++ b/priv/templates/phx.gen.auth/auth.ex
@@ -142,7 +142,7 @@ defmodule <%= inspect auth_module %> do
       conn = fetch_cookies(conn, signed: [@remember_me_cookie])
 
       if token = conn.cookies[@remember_me_cookie] do
-        {token, put_token_in_session(conn, token)}
+        {token, conn |> put_token_in_session(token) |> put_session(:<%= schema.singular %>_remember_me, true)}
       else
         {nil, conn}
       end
--- a/priv/templates/phx.gen.auth/auth.ex
+++ b/priv/templates/phx.gen.auth/auth.ex
@@ -5,14 +5,32 @@ defmodule <%= inspect auth_module %> do
   import Phoenix.Controller
 
   alias <%= inspect context.module %>
+  alias <%= inspect schema.module %>Token
   alias <%= inspect scope_config.scope.module %>
 
   # Make the remember me cookie valid for 60 days.
   # If you want bump or reduce this value, also change
   # the token expiry itself in <%= inspect schema.alias %>Token.
-  @max_age 60 * 60 * 24 * 60
+  @max_session_age_in_days 60
+
+  # How old the session token should be before it is refreshed. You can reduce this so that
+  # the session token is refreshed more often by an active user, but it will also cause more
+  # database writes. This might be desirable if you want to track more precisely the last time
+  # a given token was used. You may also increase this to reduce the number of database writes.
+  # If you increase it to a value greater than the session token expiry, the session token will
+  # never be refreshed, effectively disabling token refreshing.
+  @session_refresh_age_in_hours 24
+
+  # The remember-me cookie is a signed Phoenix.Token, and the browser is instructed to
+  # only keep it for the duration of the session token, and only send it when the request
+  # is made to the same site. Setting `same_site` to `Strict` may cause users to not be
+  # remembered when they follow links from other sites or emails.
   @remember_me_cookie "_<%= web_app_name %>_<%= schema.singular %>_remember_me"
-  @remember_me_options [sign: true, max_age: @max_age, same_site: "Lax"]
+  @remember_me_options [
+    sign: true,
+    max_age: @max_session_age_in_days * 24 * 60 * 60,
+    same_site: "Lax"
+  ]
 
   @doc """
   Logs the <%= schema.singular %> in.
@@ -100,11 +118,21 @@ defmodule <%= inspect auth_module %> do
   @doc """
   Authenticates the <%= schema.singular %> by looking into the session
   and remember me token.
+
+  Will refresh the session token if it is older than the configured refresh age.
   """
   def fetch_current_scope_for_<%= schema.singular %>(conn, _opts) do
-    {<%= schema.singular %>_token, conn} = ensure_<%= schema.singular %>_token(conn)
-    <%= schema.singular %> = <%= schema.singular %>_token && <%= inspect context.alias %>.get_<%= schema.singular %>_by_session_token(<%= schema.singular %>_token)
-    assign(conn, :current_scope, <%= inspect scope_config.scope.alias %>.for_<%= schema.singular %>(<%= schema.singular %>))
+    {token, conn} = ensure_<%= schema.singular %>_token(conn)
+
+    if token do
+      {<%= schema.singular %>_token, <%= schema.singular %>} = <%= inspect context.alias %>.get_<%= schema.singular %>_auth_by_session_token(token)
+
+      conn
+      |> assign(:current_scope, <%= inspect scope_config.scope.alias %>.for_<%= schema.singular %>(<%= schema.singular %>))
+      |> maybe_refresh_<%= schema.singular %>_session_token(<%= schema.singular %>_token)
+    else
+      assign(conn, :current_scope, <%= inspect scope_config.scope.alias %>.for_<%= schema.singular %>(nil))
+    end
   end
 
   defp ensure_<%= schema.singular %>_token(conn) do
@@ -118,6 +146,30 @@ defmodule <%= inspect auth_module %> do
       else
         {nil, conn}
       end
+    end
+  end
+
+  # Refresh the session token if it is older than the configured refresh age.
+  defp maybe_refresh_<%= schema.singular %>_session_token(conn, nil), do: conn
+
+  defp maybe_refresh_<%= schema.singular %>_session_token(conn, <%= schema.singular %>_token) do
+    if <%= inspect schema.alias %>Token.seconds_since_refresh(<%= schema.singular %>_token) > @session_refresh_age_in_hours * 60 * 60 do
+      <%= inspect context.alias %>.refresh_<%= schema.singular %>_session_token(<%= schema.singular %>_token.token)
+      maybe_refresh_remember_me_cookie(conn, <%= schema.singular %>_token.token)
+    else
+      conn
+    end
+  end
+
+  # Refresh the remember me cookie if it is set and it is the same as the session token.
+  # This is to ensure the remember me cookie has the same expiration time as the session token.
+  defp maybe_refresh_remember_me_cookie(conn, token) do
+    conn = fetch_cookies(conn, signed: [@remember_me_cookie])
+
+    if conn.cookies[@remember_me_cookie] == token do
+      put_resp_cookie(conn, @remember_me_cookie, token, @remember_me_options)
+    else
+      conn
     end
   end
 
--- a/priv/templates/phx.gen.auth/auth.ex
+++ b/priv/templates/phx.gen.auth/auth.ex
@@ -325,7 +325,7 @@ defmodule <%= inspect auth_module %> do
 
   <%= if live? do %>@doc "Returns the path to redirect to after log in."
   # the <%= schema.singular %> was already logged in, redirect to settings
-  def signed_in_path(%Plug.Conn{assigns: %{current_scope: %<%= inspect scope_config.scope.alias %>{user: %<%= inspect context.alias %>.<%= inspect schema.alias %>{}}}}) do
+  def signed_in_path(%Plug.Conn{assigns: %{current_scope: %<%= inspect scope_config.scope.alias %>{<%= schema.singular %>: %<%= inspect context.alias %>.<%= inspect schema.alias %>{}}}}) do
     ~p"<%= schema.route_prefix %>/settings"
   end
 
--- a/priv/templates/phx.gen.auth/auth_test.exs
+++ b/priv/templates/phx.gen.auth/auth_test.exs
@@ -144,8 +144,8 @@ defmodule <%= inspect auth_module %>Test do
         |> <%= inspect schema.alias %>Auth.fetch_current_scope_for_<%= schema.singular %>([])
 
       assert conn.assigns.current_scope.<%= schema.singular %>.id == <%= schema.singular %>.id
-      assert get_session(conn, :<%= schema.singular %>_token) == <%= schema.singular %>_token<%= if live? do %>
-      assert get_session(conn, :<%= schema.singular %>_remember_me)
+      assert get_session(conn, :<%= schema.singular %>_token) == <%= schema.singular %>_token
+      assert get_session(conn, :<%= schema.singular %>_remember_me)<%= if live? do %>
 
       assert get_session(conn, :live_socket_id) ==
                "<%= schema.plural %>_sessions:#{Base.url_encode64(<%= schema.singular %>_token)}"<% end %>
--- a/priv/templates/phx.gen.auth/auth_test.exs
+++ b/priv/templates/phx.gen.auth/auth_test.exs
@@ -4,11 +4,15 @@ defmodule <%= inspect auth_module %>Test do
   <%= if live? do %>alias Phoenix.LiveView
   <% end %>alias <%= inspect context.module %>
   alias <%= inspect context.module %>.<%= inspect scope_config.scope.alias %>
+  alias <%= inspect schema.module %>Token
+  alias <%= inspect schema.repo %>
   alias <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web_namespace, schema.alias) %>Auth
 
   import <%= inspect context.module %>Fixtures
+  import Ecto.Query, only: [from: 2]
 
   @remember_me_cookie "_<%= web_app_name %>_<%= schema.singular %>_remember_me"
+  @remember_me_cookie_max_age 60 * 60 * 24 * 60
 
   setup %{conn: conn} do
     conn =
@@ -45,7 +49,7 @@ defmodule <%= inspect auth_module %>Test do
 
       assert %{value: signed_token, max_age: max_age} = conn.resp_cookies[@remember_me_cookie]
       assert signed_token != get_session(conn, :<%= schema.singular %>_token)
-      assert max_age == 5_184_000
+      assert max_age == @remember_me_cookie_max_age
     end<%= if live? do %>
 
     test "redirects to settings when <%= schema.singular %> is already logged in", %{conn: conn, <%= schema.singular %>: <%= schema.singular %>} do
@@ -69,7 +73,7 @@ defmodule <%= inspect auth_module %>Test do
         |> fetch_cookies()
         |> init_test_session(%{<%= schema.singular %>_remember_me: true})
 
-      # the conn is already logged in and has the remeber_me cookie set,
+      # the conn is already logged in and has the remember_me cookie set,
       # now we log in again and even without explicitly setting remember_me,
       # the cookie should be set again
       conn = conn |> <%= inspect schema.alias %>Auth.log_in_<%= schema.singular %>(<%= schema.singular %>, %{})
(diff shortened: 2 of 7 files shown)


josevalim:
Hi @nshafer, can you please send a PR for the other fixes? Then we can discuss the auth token renewal. Basically, we want to ship v1.8 RC today or tomorrow, so the fixes would be welcome :) while we discuss the rest!

josevalim:
Ignore my previous comments (now deleted), I thought there were other fixes besides token refreshing. :) Let me do another pass and drop a comment.

josevalim:
Thank you @nshafer for the PR. Generally speaking, I would prefer to have tighter security defaults, which means no renewal of tokens. We also want to keep the generators as simple as possible - which you have done well - but overall we are still adding code. At the same time, you are right that, today, as soon as it gets expired, it forces authentication, which could happen in the middle of a POST, and it would be frustrating.

I also have concerned about the `refreshed_at` token. Now you have two timestamps to worry about but, my biggest concerns is that someone could accidentally make a new token live forever in a way that's hard to track. My suggestion would be to, instead of adding `refreshed_at`, create a new token altogether. So overall, here is my proposal:

1. Have `get_*_auth_by_session_token` return the amount the token is still valid (or its `inserted_at`, both are fine)

2. Then, in the plug, we can look at this value and act accordingly. By default, I would check if there's 1 day to expire. If there is one day to expire, and it is a non-AJAX GET request, we mark is as invalid. This will force us to update the token in a safe workflow. Your implementation could see if there's 1 day left and create a new token

3. Instead of adding the refresh to the generators, I would add new guides to our docs. Same for pruning. So one guide with the `prune` function, another with the renewal approach

WDYT?

nshafer:
So if I understand you correctly, this is what you're proposing:

1. Add an `authenticated_at, :utc_datetime` column to `UserToken` (instead of `refreshed_at`)
2. When the user logs in, a new `UserToken` is created with `inserted_at` and `authenticated_at` set to `DateTime.utc_now()`.
3. When the `fetch_current_scope_for_user()` plug receives a request 1 week before the expired time, it:
  a. Generates a new token, setting the `new_token.authenticated_at = old_token.authenticated_at`.
  b. Deletes the old token
  c. Sets the new token in the session, and sends the new token in the remember_me cookie if it is set, and equal to the old token.
4. The `User.authenticated_at` virtual field that `is_sudo?` checks would then be set to the value of `UserToken.authenticated_at` instead of `UserToken.inserted_at`.

I'm not sure I understand how this is principally different than the `refreshed_at` method, and if anything seems more complicated to me with the same end-result? Unless I'm misunderstanding. It still has to figure out how old the token is, delete/create (instead of update) the token, then update the session and the cookie, which is actually more steps. Yes it makes the `UserToken` object in the database immutable, but I'm not sure how that is much different than deleting and creating new ones automatically. The end-result would be the same, in that as long as a token isn't expired, a request from the user or an attacker would either refresh it, or get a new one that would work for another 60 days. I guess I'm failing to understand exactly why you want `UserToken`s to be immutable.

As for the plug getting the timestamp... the original implementation called `Accounts.get_user_by_session_token/1`, which only returned the `User` struct. That doesn't contain the info on how old the token is, it just returns a User or it doesn't. I didn't want to add the `refreshed_at` column as a virtual field to the `User` schema, since that's more implementation detail, and not really relevant to the `User` that code will get from the `current_scope` assign. That's why I created the `Accounts.get_user_auth_by_session_token/1` function, which returns both the `UserToken` and `User` so that the plug could see how old the token is. Changing that to return `{User, timestamp}` is still probably the same amount of lines changed, but less useful for other code that might want to call it that the dev might add on. And changing `Accounts.get_user_by_session_token/1` to return `{user, timestamp}` instead of `user` doesn't seem right, and will break existing calls to that function.

Perhaps with this change we should also change the default timeout of the token from 60 days to something more reasonable, like 2 weeks. 60 days makes sense when the Token would never be refreshed/replaced, but is probably too long if the token is being refreshed/replaced. This is what Django does by default... as long as you make a request once every 2 weeks, you stay logged in. If you don't make any requests for 2 weeks, then you have to login again. This might actually increase security, as cookies will go stale on a user's computer faster, allowing attackers to be able to hijack their session for only 2 weeks instead of 2 months. Perhaps that would alleviate your reservations for how conservative the defaults should be?

I'll pull the pruning code out and put it in a new Guide under the "Authn & Authz" section, below "API Authentication", and we can see how that looks.

josevalim:
>  I guess I'm failing to understand exactly why you want UserTokens to be immutable.

It leaves a better trail (as updating means destroying the previous refreshed_at). It avoids confusion in the code (someone accidentally refreshing the column when it should not). And it avoids a token potentially lasting forever (as long as the user keeps accessing it). If someone steals a token now, there is a chance it may still be valid 1y from now.

> So if I understand you correctly, this is what you're proposing:

I am proposing to:

1. Add an authenticated_at, :utc_datetime column to UserToken (instead of refreshed_at)
2. When the user logs in, a new UserToken is created with inserted_at and authenticated_at set to DateTime.utc_now().
3. The User.authenticated_at virtual field that is_sudo? checks would then be set to the value of UserToken.authenticated_at instead of UserToken.inserted_at.
4. If there is a GET request and 3 days before the expired time, we already asked for a sign up, to avoid disrupting a submission workflow

If we set up this foundation, then adding the actual renewal should be easier. Anyway, assuming you agree with the above, it is at a starting point that we both agree with, so we can work on that, and then revisit the impact of doing the full renewal.

> Perhaps with this change we should also change the default timeout of the token from 60 days to something more reasonable, like 2 weeks. 60 days makes sense when the Token would never be refreshed/replaced, but is probably too long if the token is being refreshed/replaced. 

I agree shorter-lived tokens are a good benefit of renewal but your implementation would, at the same time, make a leaked token potentially valid for years. If we want to argue on shorter-lived token as a benefit, then by the same metric, `refresh_at` is a downside. :)

josevalim:
Btw, i pushed 259e872314a5f3f9b54cc28f8d4e2b0c421e62ed to main as we prep v1.8 RC :)

priv/templates/phx.gen.auth/context_functions.ex
```diff
@@ -182,8 +182,19 @@
...
+    |> Repo.one()
+  end
+
+  @doc """
+  Gets the `<%= inspect schema.alias %>Token` and `<%= inspect schema.alias %>` for a given session token.
+
+  This will always return `{<%= schema.singular %>_token, <%= schema.singular %>}, or `{nil, nil}` if
+  the token is invalid or expired.
+  """
+  def get_<%= schema.singular %>_auth_by_session_token(token) do
```
josevalim, line 195: It is not worth having two functions. Make it so `get_user_by_session_token` returns `{user, token} | :error`and handle all callers accordingly. I believe it is called only by `auth.ex` by default anyway.

nshafer: I have done this... since `:error` is usually returned from `fetch*` functions, and there was already a `get_user_by_magic_link_token/1` that returns a `user` or `nil` next to it, I went with `nil` instead of `:error`, so it returns `{user, user_token}` or `nil`.

I had to modify `UserAuth.mount_current_scope/2` to accommodate.

That is one reason to have them as separate functions, btw... now the mount is also returning all fields of the `UserToken` as well as the `User` just to ignore them. That's one reason I had done a second function.

LMK. Working on the rest now.

nshafer: A diff of that change: https://github.com/nshafer/myapp18/compare/40fd0ae679bcf4df1bb3ca28610b8eb8cbeb046d..c7b18bff3ff205de1fab470388df21811def1568

nshafer:
@josevalim, I have completed modifications as we discussed. The changes in summary:

- Removed pruning stuff
- Replaced the "refreshing" code with "reissuing" code, which will delete the old token and issue a new one. There is a new `authenticated_at` column that inherits the value of the old token, so that "sudo" checks will not pass with the new token if they wouldn't with the old.
- Modified the default session length from 60 days to 14. We can play with that value of course.
- Updated tests.

I decided to keep the refresh time to every 24 hours instead of increasing that to 11 (3 days before the old session token expires.) My reasoning is that now that we're reissuing new tokens, this makes any compromised old tokens work for only up to 24 hours. So basically by rotating the token on a daily basis, I think this further increases security if there is a breach of the token. An attacker has less time to use it before the user's normal activity causes it to be deleted, and a new token issued.

I wrote the reissue code to work in a transaction and only succeed if it finds the old token and is able to delete it, only then it will generate the new token. This should work even if there are concurrent requests since it's using a database transaction.

This will work in any kind of request (GET, POST, etc) and from any client that works with cookies.

Here is a diff between the unmodified 1.8-dev auth.gen code: https://github.com/nshafer/myapp18/compare/auth_live_orig..auth_live_new_mod

And here is a diff from my original PR with the changes I made since we discussed this: https://github.com/nshafer/myapp18/compare/auth_live_new..auth_live_new_mod

If that looks ok, then I'll modify the generator templates tomorrow and commit them to this PR.

josevalim:
I am on my phone but, first pass, we probably don’t want to delete the old token, otherwise concurrent requests around the widow end may fail. So I would drop that and bump the renewal to every 7 days, to avoid so many DB tokens.

nshafer:
We need the `user_id`, `inserted_at`, `authenticated_at` and `token` from `UserToken` to check how old it is and reissue a new one if needed. The only thing extra that's being passed is `context: "session"` and `sent_to` which is null for session tokens.

I'll also remove the deletion of the old token, as I agree on that possible race condition.

I'd like to review default settings in the generated code. Overall we have 2 knobs to figure out defaults for session tokens:
- `@session_validity_in_days 14`: Overall session length (was 60 days.)
- `@session_reissue_age_in_days 7 ` How old a session token has to be before a new one is issued.

I'm not really sure what these values should be by default that would work for most general-purpose applications.

If we don't delete old tokens (or add even more complication with an `expires_at` column, which I'm not suggesting), then every time a new session is created, that token will last for that amount of time. So lower session validity is better for security, but less convenient for users.

The reissue age also plays into the UX The lower it is, the more tokens are created in the database, but the more convenient it is for users, as they will have new tokens created on a more regular basis. Set this too high, and a user may miss the window reissue, and then be logged out.

For example: with 14 and 7, user logs in one week and makes requests all week with the same token. They don't log in the following week. Then the week after that, they have to log in again. But if they make requests generally every few days, then they will always stay logged in.

So the overall perceived session length will be between `@session_reissue_age_in_days` and `@session_validity_in_days`.

So should this be 14 (session length) and 7 (reissue age)? I just came up with 14 based on django, but their auth system is very different. Should it be 30/10? Or 30/15?

I think you would have a much better idea of what a good default would be. If you want to just do 14/7, that's fine, I just want a definite positive indication, if you don't mind.

nshafer:
I'll play around with the ergonomics of a 'session_age' virtual field on 'User' and just return that instead of returning {user, user_token}'.

nshafer:
Here is the latest version: https://github.com/nshafer/myapp18/compare/auth_live_orig..auth_live_new_mod2

- Old token is no longer deleted, just a new one issued after 7 days.
- Modified `Accounts.get_user_by_session_token/1` to return `{user, token_created}` instead of `user_token`.
- Removed `Accounts.reissue_user_session_token/1` completely. Now just using `Accounts.generate_user_session_token/1` to copy the `authenticated_at` to a new token.
- Removed extra query function in `UserToken`.
- This simplified `UserAuth` a bit.
- Renamed `@session_refresh_age_in_hours` to `@session_reissue_age_in_days` and updated comment.
- Updated tests.

LMK how that's looking.

priv/templates/phx.gen.auth/auth.ex
```diff
@@ -114,13 +142,37 @@ defmodule <%= inspect auth_module %> do
...
     end
   end
 
+  # Refresh the session token if it is older than the configured refresh age.
+  defp maybe_refresh_<%= schema.singular %>_session_token(conn, nil), do: conn
+
+  defp maybe_refresh_<%= schema.singular %>_session_token(conn, <%= schema.singular %>_token) do
+    if <%= inspect schema.alias %>Token.seconds_since_refresh(<%= schema.singular %>_token) > @session_refresh_age_in_hours * 60 * 60 do
+      <%= inspect context.alias %>.refresh_<%= schema.singular %>_session_token(<%= schema.singular %>_token.token)
+      maybe_refresh_remember_me_cookie(conn, <%= schema.singular %>_token.token)
```
josevalim, line 158: Don't we need to store the new token in the session? As well as in the cookie? And I think the function below can be nuked.

priv/templates/phx.gen.auth/auth_test.exs
```diff
@@ -152,6 +157,72 @@ defmodule <%= inspect auth_module %>Test do
...
+
+    defp offset_<%= schema.singular %>_token(token, amount_to_add, unit) do
+      dt = DateTime.add(DateTime.utc_now(), amount_to_add, unit)
+      query = from(ut in <%= inspect schema.alias %>Token, where: ut.token == ^token)
+      {1, nil} = Repo.update_all(query, set: [inserted_at: dt, refreshed_at: dt])
+      {<%= schema.singular %>_token, _} = <%= inspect context.alias %>.get_<%= schema.singular %>_auth_by_session_token(token)
+      <%= schema.singular %>_token
+    end
+
+    test "does not refresh the session token after only a few hours", %{conn: conn, <%= schema.singular %>: <%= schema.singular %>} do
```
josevalim, line 174: Try to squeeze this into a previous test, possibly by checking that the token in the session is still the same.

priv/templates/phx.gen.auth/auth_test.exs
```diff
@@ -152,6 +157,72 @@ defmodule <%= inspect auth_module %>Test do
...
+        |> put_session(:<%= schema.singular %>_token, token)
+        |> <%= inspect schema.alias %>Auth.fetch_current_scope_for_<%= schema.singular %>([])
+
+      assert conn.assigns.current_scope.<%= schema.singular %>.id == <%= schema.singular %>.id
+      assert get_session(conn, :<%= schema.singular %>_token) == token
+      assert {<%= schema.singular %>_token, _} = <%= inspect context.alias %>.get_<%= schema.singular %>_auth_by_session_token(token)
+      assert <%= inspect schema.alias %>Token.seconds_since_refresh(<%= schema.singular %>_token) >= 60 * 60 * 3
+    end
+
+    test "refreshes the session token after a few days", %{conn: conn, <%= schema.singular %>: <%= schema.singular %>} do
```
josevalim, line 189: Squeeze this and the next test together.

priv/templates/phx.gen.auth/auth_test.exs
josevalim: Well, it seems you already do so once you changed line 147, so we can nuke this test.

nshafer:
I haven't updated the templates yet in this PR, just the implemented code in a generated project here: https://github.com/nshafer/myapp18/compare/auth_live_orig..auth_live_new_mod2

josevalim:
Gah, apologies, I was reviewing the version in this PR and I noticed midway it was still the same. :(

josevalim:
Unfortunately I cannot drop comments on the link you sent, unless you make it a PR. I am off to bed now, I can do a pass tomorrow.

josevalim:
Unfortunately I cannot drop comments on the link you sent, unless you make it a PR. You can make a PR against your own repo and that should work. I am off to bed now, I can do a pass tomorrow. :)

nshafer:
Sorry about that, let me know if this works: https://github.com/nshafer/myapp18/pull/1

josevalim:
> I just figured this would be much easier for you to read than the templated code.

For sure, thank you for breaking it down :heart:

nshafer:
This PR has been superseded by #6162.

PR #6152, opened 2025-04-02
Add a link to the Flash messages guide from put_flash docs

adamu:
While reading the `put_flash` docs, I was wondering what the valid key values where. This is documented in the guide, so I added a link.

image

Link goes to https://hexdocs.pm/phoenix/controllers.html#flash-messages

### DIFF
--- a/lib/phoenix/controller.ex
+++ b/lib/phoenix/controller.ex
@@ -1689,6 +1689,8 @@ defmodule Phoenix.Controller do
 
   Returns the updated connection.
 
+  See [Flash messages](controllers.md#flash-messages) in the Controllers guide for more details.
+
   ## Examples
 
       iex> conn = put_flash(conn, :info, "Welcome Back!")


SteffenDE:
lgtm! @josevalim is linking back from functions to guides something we usually do?

josevalim:
> lgtm! @josevalim is linking back from functions to guides something we usually do?

Usually the links flow in the other direction, from guides to references, especially because the guides are longer form and may be referring to additional context. I would prefer if the API reference stand by itself.

@adamu, can you please change this so we bring all of the relevant concepts to this function instead? It is ok if it is duplicated. Thanks!

adamu:
Okay, it might be quite a big change to move all the guide contents into the docs, so I just added the info that I was looking for: that any key is supported, and handlers for `:info` and `:error` are generated by default.

image

josevalim:
Lovely!

josevalim:

PR #6155, opened 2025-04-02
invert copy_core_component

### LINKED ISSUE #6154: Missing `core_components.ex` when generating Phoenix project from unreleased version
### Environment

* Elixir version (elixir -v): 1.18.3
* Phoenix version (mix deps): latest `main` ([`c73c1b0`](https://github.com/phoenixframework/phoenix/commit/c73c1b0b4ca3fd0adf57bb2010d097d47d8f0768))
* Operating system: macOS

### Actual behavior

I followed the [instructions from the README](https://github.com/phoenixframework/phoenix?tab=readme-ov-file#generating-a-phoenix-project-from-unreleased-versions) to create a Phoenix project from an unreleased version.

Executing the command `mix phx.new dev_app --dev`  in the `/installer` directory results in

```
== Compilation error in file lib/phx_new/web.ex ==
** (File.Error) could not read file "[...]/phoenix/installer/templates/phx_web/components/core_components.ex": no such file or directory
```

In https://github.com/phoenixframework/phoenix/pull/6126 the `core_components.ex` file was removed from the `/installer` directory. So copying it from `/priv/templates/phx.gen.live/core_components.ex` to `/installer/templates/phx_web/components/core_components.ex` locally resolves the issue.

### Expected behavior

I expect an application to be generated by following the instructions from the README.

We might want to update the instructions based on the removal of the `core_component.ex` file in https://github.com/phoenixframework/phoenix/pull/6126

SteffenDE:
Fixes #6154

It also includes the core_components file in the Phoenix archive, therefore the copy should always succeed. We can also use `cp` instead of `cp!` and not include it in the archive, as it should have already been copied prior to building the archive.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -21,3 +21,5 @@ erl_crash.dump
 phoenix-*.ez
 
 .DS_Store
+
+/priv/templates/phx.gen.live/core_components.ex
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -64,15 +64,15 @@ defmodule <%= @web_namespace %>.CoreComponents do
         @kind == :info && "alert-info",
         @kind == :error && "alert-error"
       ]}>
-        <.icon :if={@kind == :info} name="hero-information-circle-mini" class="size-5 shrink-0" />
-        <.icon :if={@kind == :error} name="hero-exclamation-circle-mini" class="size-5 shrink-0" />
+        <.icon :if={@kind == :info} name="hero-information-circle-mini" class="h-5 w-5 shrink-0" />
+        <.icon :if={@kind == :error} name="hero-exclamation-circle-mini" class="h-5 w-5 shrink-0" />
         <div>
           <p :if={@title} class="font-semibold">{@title}</p>
           <p>{msg}</p>
         </div>
         <div class="flex-1" />
         <button type="button" class="group self-start cursor-pointer" aria-label=<%= maybe_heex_attr_gettext.("close", @gettext) %>>
-          <.icon name="hero-x-mark-solid" class="size-5 opacity-40 group-hover:opacity-70" />
+          <.icon name="hero-x-mark-solid" class="h-5 w-5 opacity-40 group-hover:opacity-70" />
         </button>
       </div>
     </div>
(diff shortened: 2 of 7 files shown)


.gitignore
```diff
@@ -21,3 +21,5 @@ erl_crash.dump
 phoenix-*.ez
 
 .DS_Store
```
josevalim, line 23: Unrelated but system ignores shuold not be here :D

```suggestion
```

installer/templates/phx_web/components/core_components.ex
```diff
@@ -64,51 +64,39 @@ defmodule <%= @web_namespace %>.CoreComponents do
         @kind == :info && "alert-info",
         @kind == :error && "alert-error"
       ]}>
-        <.icon :if={@kind == :info} name="hero-information-circle-mini" class="size-5 shrink-0" />
-        <.icon :if={@kind == :error} name="hero-exclamation-circle-mini" class="size-5 shrink-0" />
+        <.icon :if={@kind == :info} name="hero-information-circle-mini" class="h-5 w-5 shrink-0" />
+        <.icon :if={@kind == :error} name="hero-exclamation-circle-mini" class="h-5 w-5 shrink-0" />
```
josevalim, line 68: Shouldn't we keep these as size-5?

josevalim: Ok, looking below, it seems you got the wrong core components file? :D

mix.exs
```diff
@@ -266,4 +271,20 @@ defmodule Phoenix.MixProject do
...
+  defp copy_core_components(_) do
+    source =
+      Path.join([
+        __DIR__,
+        "installer",
+        "templates",
+        "phx_web",
+        "components",
+        "core_components.ex"
+      ])
```
josevalim, line 284: FWIW, this also works, including on Windows as it converts / into \.

```suggestion
      Path.join(__DIR__, "installer/templates/phx_web/components/core_components.ex")
```

josevalim (approved):
LGTM, as long as we are sure we are including the right core components :)

installer/templates/phx_web/components/core_components.ex
SteffenDE: Seems like they weren’t in sync for me :O

PR #6156, opened 2025-04-02
Add <Layout.app> to index.html.heex

CharlesOkwuagwu:
`index.html.heex` for `hello_html` does not include `<Layout.app>`.

Was this intentional or an omission?


SteffenDE:
It was an intentional omission, because layouts are only explained later iirc (@josevalim)

josevalim:
Yes, we have a later example in this same guide that includes it and explains what it does. :)

CharlesOkwuagwu:
The screen shot is a bit confusing, as it has the header from the template :)

josevalim:
Oh, great catch! can you please update the screenshot then? :)

CharlesOkwuagwu:
@josevalim done.

PR #6136, opened 2025-03-21
use explicit layouts in generators

SteffenDE:

PR #6161, opened 2025-04-03
Do not specify layout: false in PageController examples

michallepicki:
Since the generators are no longer defining dynamic layouts with put_layout or in use Phoenix.Controller (since https://github.com/phoenixframework/phoenix/pull/6136/files ), I think there's no need for explicitly disabling the layout in the PageController in the installer or guides


SteffenDE (approved):

SteffenDE:
Thank you! 🙌🏻

PR #6162, opened 2025-04-04
Reissue new tokens periodically in phx.gen.auth code.

nshafer:
The auth code currently generated by mix phx.gen.auth generates tokens with a fixed expiration of 60 days. Unless a user logs in again, they will never get a new token. Once that token expires, they are forced to log in again. This could happen in the middle of a session, such as while they are filling out a form, causing them to lose their submission when the POST redirects them to log in again.

This PR fixes this by issuing a new token automatically to valid sessions that are using a token more than 7 days old.

With this new refresh mechanism, tokens now expired after 14 days instead of the old default of 60. Shorter lived tokens decrease the window of possible abuse if the token is exfiltrated somehow, such as from a backup of a browser cookie store.

Both the token life and reissue age can be easily changed via module attributes in `UserToken` and `UserAuth`.

Tokens now persist an `authenticated_at` field so that sudo checks are based on the time of the original login, and new tokens will carry forward that same timestamp, so that new tokens do not incorrectly offer a chance for sudo mode access.

This also fixes a small bug with setting `:user_remember_me` in the session after restoring a token from the remember me cookie, causing this preference to not persist after sudo mode forces a new login. This new code also relies on that being set in the session to properly reissue cookies after new tokens are created.

This PR supersedes #6151 after much work with the tireless help of @josevalim to fine-tune this mechanism.

This is mostly the same change for both `--live` and `--no-live` authentication, with just some small changes in a couple tests between the two to make sure sudo mode is working correctly. Diffs from the original output of `mix phx.gen.auth` and the output after this patch is applied can be easily seen via:
- [`mix phx.gen.auth --live Accounts User users`](https://github.com/nshafer/myapp18/compare/auth_live_orig2..auth_live_new2)
- [`mix phx.gen.auth --no-live Accounts User users`](https://github.com/nshafer/myapp18/compare/auth_nolive_orig2..auth_nolive_new2)

@josevalim - feel free to make any further tweaks you want. Also review the 14 day token expiration and 7 day renewal, tweak if you like. I'd be fine with that default in my projects, but you'll have a better idea of what is good as a global default.

Thanks again for all your tireless work on Elixir and Phoenix, and for helping me get this PR whipped into shape!

### DIFF
--- a/priv/templates/phx.gen.auth/auth.ex
+++ b/priv/templates/phx.gen.auth/auth.ex
@@ -141,7 +141,7 @@ defmodule <%= inspect auth_module %> do
 
   # Reissue the session token if it is older than the configured reissue age.
   defp maybe_reissue_<%= schema.singular %>_session_token(conn, <%= schema.singular %>, token_inserted_at) do
-    token_age = DateTime.diff(DateTime.utc_now(), token_inserted_at, :day)
+    token_age = <%= inspect datetime_module %>.diff(<%= datetime_now %>, token_inserted_at, :day)
 
     if token_age >= @session_reissue_age_in_days do
       new_token = <%= inspect context.alias %>.generate_<%= schema.singular %>_session_token(<%= schema.singular %>)
--- a/priv/templates/phx.gen.auth/auth.ex
+++ b/priv/templates/phx.gen.auth/auth.ex
@@ -7,12 +7,24 @@ defmodule <%= inspect auth_module %> do
   alias <%= inspect context.module %>
   alias <%= inspect scope_config.scope.module %>
 
-  # Make the remember me cookie valid for 60 days.
-  # If you want bump or reduce this value, also change
-  # the token expiry itself in <%= inspect schema.alias %>Token.
-  @max_age 60 * 60 * 24 * 60
+  # Make the remember me cookie valid for 14 days. This should match
+  # the session validity setting in <%= inspect schema.alias %>Token.
+  @max_cookie_age_in_days 14
   @remember_me_cookie "_<%= web_app_name %>_<%= schema.singular %>_remember_me"
-  @remember_me_options [sign: true, max_age: @max_age, same_site: "Lax"]
+  @remember_me_options [
+    sign: true,
+    max_age: @max_cookie_age_in_days * 24 * 60 * 60,
+    same_site: "Lax"
+  ]
+
+  # How old the session token should be before a new one is issued. When a request is made
+  # with a session token older than this value, then a new session token will be created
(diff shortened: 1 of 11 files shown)


nshafer:
The CI tests located an issue with mixed `DateTime` vs `NaiveDateTime` usage that I'm not properly handling. The current code assumes `config :myapp18,  generators: [timestamp_type: :utc_datetime]` and throws errors when `timestamp_type: :naive_datetime` in the projects config. Good catch with the extra integration tests! I'll fix it up tomorrow or this weekend.

priv/templates/phx.gen.auth/auth.ex
```diff
@@ -114,13 +132,37 @@ defmodule <%= inspect auth_module %> do
...
+  # Reissue the session token if it is older than the configured reissue age.
+  defp maybe_reissue_<%= schema.singular %>_session_token(conn, <%= schema.singular %>, token_inserted_at) do
+    token_age = DateTime.diff(DateTime.utc_now(), token_inserted_at, :day)
+
+    if token_age >= @session_reissue_age_in_days do
+      new_token = <%= inspect context.alias %>.generate_<%= schema.singular %>_session_token(<%= schema.singular %>)
+
+      conn
+      |> put_token_in_session(new_token)
+      |> maybe_refresh_remember_me_cookie(new_token)
```
josevalim, line 151: Note to self: this is very close to the login function, we should probably break the login function in two, one which logs in and the other that logs in and redirect. But we want to make other changes to that function, so we will do this later. :) No action needed from you.

nshafer:
OK, this is ready for review again. The code respects `:naive_datetime`, `:utc_datetime` and `:utc_datetime_usec`, same as the rest of `mix phx.gen.auth`.

Due to persisting `authenticated_at` to the database, we have to be much more careful with microseconds in the `DateTime` or `NaiveDateTime` so that Ecto doesn't throw an exception when it writes the value.

Also, since Phoenix now requires Elixir 1.15, I used the new'ish `*DateTime.utc_now(:second)` function so the correct "now" can be generated on the correct module with the correct precision.

LMK if there's anything else, and I'll make the changes.

josevalim (approved):
This looks great to me, thank you!

@SteffenDE once you are ready, please merge this and then do changes we discussed to `renew_session` and introduce `log_in_user_and_redirect`. We likely will need to update the docs too.

SteffenDE:
@josevalim I‘ll review this later today. The changes we discussed are here: https://github.com/phoenixframework/phoenix/pull/6164 - I‘ll add an extra test if those are fine.

PR #6163, opened 2025-04-04
Fix wrong module name in Scope moduledoc

chaychoong:
Truth be told, I'm not entirely sure if it's really a mistake in the docs, or it's a mistake in the module name. Or a mistake in my understanding of scopes. Please do check my understanding - I wrote a rather verbose commit message just to be clear.


SteffenDE:
Thank you! 🙌🏻

This was just a leftover from when scopes were still work in progress :)

PR #6164, opened 2025-04-04
do not create a new session on sudo mode re-auth

SteffenDE:
When re-authenticating to access a sudo-mode protected page, the existing session would be cleared, invalidating the existing CSRF token. This meant that any open tab with in-progress forms would be invalidated. Now, we just put the new token into the session, keeping the same session id + CSRF token.

### DIFF
--- a/priv/templates/phx.gen.auth/auth.ex
+++ b/priv/templates/phx.gen.auth/auth.ex
@@ -35,7 +35,7 @@ defmodule <%= inspect auth_module %> do
     remember_me = get_session(conn, :<%= schema.singular %>_remember_me)
 
     conn
-    |> renew_session()
+    |> renew_session(user)
     |> put_token_in_session(token)
     |> maybe_write_remember_me_cookie(token, params, remember_me)
     |> redirect(to: <%= schema.singular %>_return_to || signed_in_path(conn))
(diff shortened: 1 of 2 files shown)


priv/templates/phx.gen.auth/auth.ex
```diff
@@ -70,7 +71,15 @@ defmodule <%= inspect auth_module %> do
   #       |> put_session(:preferred_locale, preferred_locale)
   #     end
   #
-  defp renew_session(conn) do
+  defp renew_session(conn, <%= schema.singular %> \\ nil)
```
josevalim, line 74: I personally don't like using default values for private functions. Unless in this case there is no guarantee we always have a user?

josevalim (approved):

priv/templates/phx.gen.auth/auth.ex
SteffenDE: changed it to pass nil in logout and removed the default value

priv/templates/phx.gen.auth/auth.ex
```diff
@@ -55,13 +55,20 @@ defmodule <%= inspect auth_module %> do
     |> put_resp_cookie(@remember_me_cookie, token, @remember_me_options)
   end
 
+  # do not renew session if the <%= schema.singular %> is already logged in (sudo mode reauthentication)
```
josevalim, line 58:
```suggestion
  # Do not renew session if the <%= schema.singular %> is already logged in (sudo mode reauthentication)
```

josevalim (approved):

PR #6147, opened 2025-03-28
Tweak generators

chrismccord:

PR #6166, opened 2025-04-04
Remove no-op lines from recreate_default_css script

rhcarvalho:
* There is no more "themes: all;" in the target CSS file, so `String.replace/3` is a no-op. The default theme config was changed in 6da42953c4e7d94e7666fa0584b29fc4e35fc465 (#6147)
* installer/templates/phx_static/default.css is up-to-date, last updated in cc5c7af03507710952c6c73a27f02e7fb6a70d3f


SteffenDE:
Thank you!

PR #6169, opened 2025-04-04
Document <Layouts.app>

LostKobrakai:


SteffenDE:
Actually we don't need to pass the current_scope, as we don't use it at the moment 🤔

josevalim (approved):

installer/templates/phx_web/components/layouts.ex
```diff
@@ -11,6 +11,20 @@ defmodule <%= @web_namespace %>.Layouts do
 
   embed_templates "layouts/*"
 
+  @doc """
```
pxp9, line 14: Hey thanks for this, I was trying 1.80.rc-0 candidate and I was a bit confused of the new way we should use the dynamic layouts, because the flash was not working.

installer/templates/phx_web/components/layouts.ex
```diff
@@ -11,6 +11,20 @@ defmodule <%= @web_namespace %>.Layouts do
...
+  Renders the app layout
+
+  ## Examples
+
+      <Layouts.app flash={@flash}>
+        <h1>Content</h1>
+      </Layout.app>
+      
+  """
+  attr :flash, :map, required: true, doc: "the map of flash messages"
```
SteffenDE, line 24: Let's also add the current_scope:

```suggestion
  attr :flash, :map, required: true, doc: "the map of flash messages"
  attr :current_scope, :map, default: nil, doc: "the current [scope](https://hexdocs.pm/phoenix/scopes.html)"
```

installer/templates/phx_web/components/layouts.ex
```diff
@@ -11,6 +11,21 @@ defmodule <%= @web_namespace %>.Layouts do
...
+  attr :current_scope, :map, default: nil, doc: "the current [scope](https://hexdocs.pm/phoenix/scopes.html)"
```
SteffenDE, line 25:
```suggestion

  attr :current_scope, :map,
    default: nil,
    doc: "the current [scope](https://hexdocs.pm/phoenix/scopes.html)"
```

SteffenDE:
Thank you! 🙌🏻

PR #6170, opened 2025-04-04
Add type=button to theme toggle

LostKobrakai:


installer/templates/phx_web/components/layouts.ex
```diff
@@ -103,15 +103,27 @@ defmodule <%= @web_namespace %>.Layouts do
...
+      >
         <.icon name="hero-sun-micro" class="size-4 opacity-75 hover:opacity-100" />
       </button>
 
-      <button phx-click={JS.dispatch("phx:set-theme", detail: %{theme: "dark"})} class="flex p-2">
+      <button
+        type="button"
+        phx-click={JS.dispatch("phx:set-theme", detail: %{theme: "dark"})}
+        class="flex p-2"
+      >>
```
chrismccord, line 126: dup

chrismccord:
Is this for a11ly purposes or some other issue you're trying to solve?

LostKobrakai:
The default type is submit and these certainly don't submit any form, nor should they if they happen to be placed within one.

installer/templates/phx_web/components/layouts.ex
LostKobrakai, line 126:
```suggestion
      >
```

LostKobrakai:
I simply see it as good practise to have buttons, which are not meant to submit forms be declared with `type="button"`. Yes phoenix doesn't put those in forms, but users might. It's a regular thing I tell people running into this issue with their own components.

SteffenDE:
I usually set the default type for button in my core components to button. When something is a submit button, I tend to explicitly mark it as such. So I’m inclined to say that while we don’t need it here, it’s still good to set it.

josevalim:
I agree with @chrismccord that we are fine in this case. The default outside forms is already "button" and it is unlikely this particular component would be used inside a form.

SteffenDE:
Closing this then!

PR #6171, opened 2025-04-04
Adjust the User table name to be plural in Cross Context Boundaries

anderkonzen:
While going through the docs, I noticed that the command to generate the auth code is slightly different than the previous version - the name of the user's table is not in the plural. This PR fixes this detail.


SteffenDE:
Thank you! 🙌🏻

PR #6175, opened 2025-04-06
share log in code for token reissueing

SteffenDE:


priv/templates/phx.gen.auth/auth.ex
```diff
@@ -42,15 +43,21 @@ defmodule <%= inspect auth_module %> do
...
   def log_in_<%= schema.singular %>(conn, <%= schema.singular %>, params \\ %{}) do
-    token = <%= inspect context.alias %>.generate_<%= schema.singular %>_session_token(<%= schema.singular %>)
     <%= schema.singular %>_return_to = get_session(conn, :<%= schema.singular %>_return_to)
+
+    conn
+    |> create_or_extend_session(<%= schema.singular %>, params)
+    |> redirect(to: <%= schema.singular %>_return_to || signed_in_path(conn))
+  end
+
+  defp create_or_extend_session(conn, <%= schema.singular %>, params) do
```
SteffenDE, line 53: @josevalim as we don't use the variant without redirect outside of this module, I decided to make it a private function instead.

SteffenDE (commented):
@josevalim which part of the docs did you have in mind when you said we'll likely need to change them?

priv/templates/phx.gen.auth/auth.ex
```diff
@@ -247,7 +256,19 @@ defmodule <%= inspect auth_module %> do
...
+  end
+
+  def signed_in_path(_), do: ~p"/"<% else %>defp signed_in_path(_conn), do: ~p"/"
+
+  <% else %>defp put_token_in_session(conn, token) do
+    put_session(conn, :<%= schema.singular %>_token, token)
+  end
+
+  @doc """
   Used for routes that require sudo mode.
```
josevalim, line 272:
```suggestion
  Plug for routes that require sudo mode.
```

josevalim (approved):
I moved some functions around, so they are closer together and removed the amount of <%= if live? do %> blocks. Ship it if you are happy!

priv/templates/phx.gen.auth/auth.ex
```diff
@@ -247,8 +256,20 @@ defmodule <%= inspect auth_module %> do
     end)
   end
 
-  <% else %>@doc """
-  Used for routes that require sudo mode.
+  @doc "Returns the path to redirect to after log in."
+  # the <%= schema.singular %> was already logged in, redirect to settings
+  def signed_in_path(%Plug.Conn{assigns: %{current_scope: %<%= inspect scope_config.scope.alias %>{<%= schema.singular %>: %<%= inspect context.alias %>.<%= inspect schema.alias %>{}}}}) do
```
josevalim, line 261: I don't understand why we have this special clause for LiveView. It feels we could get rid of it.

SteffenDE: From the top of my head: because we cannot store the return_to path in LiveView when navigating to the settings page and we need to reauthenticate.

For good UX, you want the user to be sent back to settings. For the non-live version this just works because we store the return_to in the sudo mode plug, but for the live version we cannot, so iirc you said when we did the magic link changes that we should just hardcore the redirect to settings when logged in for the live version.

SteffenDE: https://github.com/SteffenDE/phoenix_gen_auth_magic_link/pull/1#discussion_r1941436131

PR #6177, opened 2025-04-08
Range Slider Class Component

gimyeonghun:
Hi there,

I was playing with the new `core_components.ex` and realised that DaisyUI provides a range slider class component.

I updated the file to reach parity with the CSS framework.

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -219,6 +219,25 @@ defmodule <%= @web_namespace %>.CoreComponents do
     """
   end
 
+  def input(%{type: "range"} = assigns) do
+    ~H"""
+    <fieldset class="fieldset mb-2">
+      <label>
+        <span :if={@label} class="fieldset-label mb-1">{@label}</span>
+        <input
+          type="range"
+          name={@name}
+          id={@id}
+          value={Phoenix.HTML.Form.normalize_value(@type, @value)}
+          class={["w-full range", @errors != [] && "range-error"]}
+          {@rest}
+        />
+      </label>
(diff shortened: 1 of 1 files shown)


greven:
I can't speak on behalf of the core maintainers but I think the idea is to keep Core Components as simple as possible. DaisyUI has a plethora of components that while easy to extend into the generated core_components it does increase the surface area of the API and with it more probable issues. Another thing is that core_components focus on components used within the generators, so anything added is done so for that reason.

It's good to know that it's easy to extend the core components though. 🙂

SteffenDE:
Thank you for the contribution, @gimyeonghun! We generally try to keep the CoreComponents small, and @greven already explained our stance very well. Since we don't have a use case for a range slider in the generators, we don't plan to include code to handle it. The CoreComponents are meant to belong to your project once generated and be changed and extended as needed. You've demonstrated very well how easy it is to extend the existing inputs to handle ranges, so this is definitely a good reference for people looking how to do it :)

gimyeonghun:
Hi @greven and @SteffenDE 

Thank you for your replies! 

I agree - the components should be kept at a minimum.

I only submitted the PR because this is what is rendered without the `range` clause. Please let me know if it should still be excluded however! Thanks 😁 

`input` class
Screenshot 2025-04-09 at 10 00 47 pm

`range` class
Screenshot 2025-04-09 at 10 01 02 pm

SteffenDE:
We could remove "range" from the list of supported values, similar to how we removed "radio": https://github.com/phoenixframework/phoenix/commit/712ad7ede061b4716d6b17fc3b6bb7e4858d5ebe

SteffenDE:
Yes, please send a PR that removes the "range" type!

PR #6179, opened 2025-04-09
don't rely on floki specific selectors

SteffenDE:


josevalim (approved):

priv/templates/phx.gen.auth/registration_live_test.exs
```diff
@@ -72,7 +72,7 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
 
       {:ok, _login_live, login_html} =
         lv
-        |> element(~s|main a:fl-contains("Log in")|)
+        |> element("main a", "Log in")
```
josevalim, line 75: Actually, this may find too many links in a regular page and unfortunately be slow. Could we make it a bit more precise maybe?

josevalim: Perhaps we can add some sort of ID to the container of those links or add some data attributes?

SteffenDE: It should find the same elements that Floki finds, no?

SteffenDE: We could be a bit more specific

```
element(~s|main ul a[href="/users/log-in"]|)
```

or something like that. But I'm not sure if that is a concern, since we know the page that we test.

josevalim: You are right, we won't be any worse better than Floki, so that's fine. Ship the current version.

josevalim (approved):

PR #6181, opened 2025-04-09
Add fetch() support to LongPoll when XMLHTTPRequest is not available

Thomas101:
Hi, we noticed that when opening a socket in a ServiceWorker and using `LongPoll`, the socket fails because neither `XMLHTTPRequest` or `XDomainRequest` are available in the ServiceWorker scope.

I realise this is a bit of an edge case, as if you have ServiceWorker support, you probably already have native `WebSocket` support. In our instance, we sometimes fall back to LongPoll when network conditions block connections.

To work around this, I've added support for using `fetch`, which is in the global scope of a ServiceWorker. To ensure we don't break any existing setups, `LongPoll` requests still prefer `XDomainRequest` and `XMLHTTPRequest` in the same ordering as before; this only acts as a fallback when neither are available.

### DIFF
--- a/assets/js/phoenix/ajax.js
+++ b/assets/js/phoenix/ajax.js
@@ -9,10 +9,41 @@ export default class Ajax {
     if(global.XDomainRequest){
       let req = new global.XDomainRequest() // IE8, IE9
       return this.xdomainRequest(req, method, endPoint, body, timeout, ontimeout, callback)
-    } else {
+    } else if(global.XMLHttpRequest){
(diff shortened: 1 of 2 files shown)


SteffenDE:
Very nice! I'd love to see an extra test [here](https://github.com/phoenixframework/phoenix/blob/main/assets/test/longpoll_test.js) that mocks XMLHTTPRequest to undefined or whatever and then expects the respecting fetch calls.

SteffenDE (approved):

PR #6184, opened 2025-04-10
Do not show ANSI escape codes on exception pages

jonatanklosko:
The code reloader capture the output of mix compile tasks and shows it directly on the exception page. When the output includes ANSI escapes for coloring they are left as is:

image

This adds a simple regex pass to remove them.

### DIFF
--- a/lib/phoenix/code_reloader.ex
+++ b/lib/phoenix/code_reloader.ex
@@ -330,6 +330,11 @@ defmodule Phoenix.CodeReloader do
   defp format_output(output) do
     output
     |> String.trim()
+    |> remove_ansi_escapes()
     |> Plug.HTML.html_escape()
+  end
+
+  defp remove_ansi_escapes(text) do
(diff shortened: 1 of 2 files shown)


SteffenDE:
I'd love to see them actually rendered in color on the error page. In theory, something like this should work: https://github.com/drudru/ansi_up/blob/main/ansi_up.js

(not sure if we really want to do that though)

jonatanklosko:
Not sure if it's vendoring in this amount of JS, but it's up to you! Another complexity is that we would need a different set of colors per each theme.

Given that this hasn't been reported thus far, I suspect the ANSI usage is pretty rare for errors anyway.

cc @josevalim

josevalim:
Vendoring JS is not a problem but I am worried about finding colors that work nicely for both. Although the Elixir built-in ANSI colors should look good both on black and white backgrounds.

josevalim:
Well, let's merge this, it is definitely a fix, we can swap up by the JS library in another PR if we want to go down that path!

josevalim:

Issue #6188, opened 2025-04-11
Liveview container div exhibits "flash of unstyled content" like behaviour.


JonRowe:
:wave: When using the default suggested styles or part there of, the nature of the `data-phx-root-id` attribute means that the page is rendered before it is applied, meaning the behaviour of the container changes from its default to `contents` over the course of the page mount lifecycle, which is avoided if we either render the attribute in the static pass, or change which attribute is used.

### Environment

* Elixir version (elixir -v): 1.18.3
* Phoenix version (mix deps): From git main / b1c459943b3279f97725787b9150ff4950958d12
* Operating system: MacOS

### Actual behavior

`data-phx-root-id` is applied either by the LiveSocket JS or by the connected mount, changing the container div to `display: contents;`

### Expected behavior

That the `display: contents;` rule is applied consistently despite the state of the liveview / JS.

### Proposed fixes

That the static render includes the attribute even if empty, or the default css uses something like `data-phx-main` to apply it.

JonRowe:
Thank you!

PR #6189, opened 2025-04-11
use data-phx-session as selector for LV containers

SteffenDE:
Closes https://github.com/phoenixframework/phoenix/issues/6188.

`[data-phx-session]` is more appropriate, since the session attribute is already included in the disconnected render.


JonRowe:
Thanks! 🙏

PR #6195, opened 2025-04-14
Add docs for :layout option to Endpoint's :render_errors config

adamu:
When setting up an error view, I noticed that the `:layout` option to `:render_errors` was undocumented, and spent quite a while figuring out how to set it. This PR adds an example to the [docs](https://hexdocs.pm/phoenix/Phoenix.Endpoint.html#module-runtime-configuration).

I'm not sure what the `:log` option is for either, but left it there.

image


adamu:
I'm closing this because I think I have not fully understood how Phoenix handles layouts for error handling, and also see layouts been reworked on `main`. Will maybe revisit again later.

PR #6196, opened 2025-04-14
Component tests poc

jdmarshall:
I'd like to make contributions to the core_components set and I suspect others might want to do the same, rather than jumping immediately to their own or third party replacements.

Given that there are for instance missing input types, and a few bugs still being found in the existing components, I think it would be good if there were some pinning tests, and a pattern to add more tests, so that new features or bug fixes are less likely to introduce regressions in existing functionality. 

This also might be a good way to make sure that --no-tailwind or other --no flags don't create a bunch of issues. 

This is basically a placeholder test because I know testing but my knowledge of phoenix LV is still a bit spotty. 

It was suggested (Chris?) that these should be integration tests and I think I understand why.

### DIFF
--- a/integration_test/mix.exs
+++ b/integration_test/mix.exs
@@ -13,7 +13,10 @@ defmodule Phoenix.Integration.MixProject do
       elixir: "~> 1.15",
       elixirc_paths: elixirc_paths(Mix.env()),
       start_permanent: Mix.env() == :prod,
-      deps: deps()
+      deps: deps(),
+      test_ignore_filters: [
+        &String.starts_with?(&1, "test/fixtures")
+      ]
     ]
   end
 
--- a/integration_test/test/code_generation/app_with_defaults_test.exs
+++ b/integration_test/test/code_generation/app_with_defaults_test.exs
@@ -20,6 +20,23 @@ defmodule Phoenix.Integration.CodeGeneration.AppWithDefaultsTest do
         assert_tests_pass(app_root_path)
       end)
     end
+
+    test "has functioning core_components" do
+      with_installer_tmp("app_with_defaults", fn tmp_dir ->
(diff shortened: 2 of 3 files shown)


josevalim:
Thank you @jdmarshall, what you have in mind here? It feels most of the bugs are UI related, so how could that help? We already compile this code, so those errors would be caught too.

jdmarshall:
I'd like to specifically catch regressions in any PRs meant to enhance the templates. There's currently no part of the lifecycle that tests that and the generated tests also don't have much code coverage here, leaving consumers to write functional tests. And that's just a route to a Testing Cone instead of a Test Pyramid.

And I think in particular in the case of the inputs, circa 1.17, the lack of feature parity between some of the components would be much more obvious in the face of unit tests.

josevalim:
I understand that in theory but I am wondering which kind of unit tests would be written in practice. :) we do have integration tests though through the general scaffold stuff

jdmarshall:
Where are the integration tests for the core components? I injected bugs into the code and still got green builds.

josevalim:
We test the phx.gen.live generators, so we test they are functional. But for things like changes to style and classes, there is nothing we could really test. We could test a certain component has a certain class but that does not really buy us anything, as the class may not exist or be the wrong class  or look visually ugly. Which is why I am asking which tests would be written. If there are no examples for now, then it is next to close until we have some practical ones!

SteffenDE:
I was thinking it would be more practical to just add playwright tests for what we do in the generator integration tests. Those could actually make screenshots and assert visually on those.

jdmarshall:
I was thinking more of the reading and expansion of the parameters into DOM, though typos in style concatenation are not impossible either. The table component was quite complex. Select and lists aren’t exactly trivial. Nor would be checkbox or radio groups.

PR #6197, opened 2025-04-15
Fix crash when an open :show page gets a PubSub broadcast for other items

nshafer:
In the code generated with `mix phx.gen.live`, in the `show.ex` file, there are 2 `handle_info/2` callbacks that are tightly bound to only match when the messages broadcast from the context are for the item they are currently showing. If the user has another tab open on a show page, and does anything from another tab, such as adding, deleting or updating a different item, the show LV will crash with a "no function claus matching" exception.

This simply adds a fallback `handle_info/2` handler that tightly matches all other expected messages due to the `Context.subscribe_items/1` call in the generated `mount/3`, so that LV won't crash when it gets an unexpected message. However, if it gets a different message that is from something else that the user adds, then it will still crash so they know they need to handle that message shape.

### DIFF
--- a/priv/templates/phx.gen.live/show.ex
+++ b/priv/templates/phx.gen.live/show.ex
@@ -53,5 +53,10 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
      socket
      |> put_flash(:error, "The current <%= schema.singular %> was deleted.")
      |> push_navigate(to: ~p"<%= scope_socket_route_prefix %><%= schema.route_prefix %>")}
+  end
+
+  def handle_info({type, %<%= inspect schema.module %>{}}, socket)
+      when type in [:created, :updated, :deleted] do
+    {:noreply, socket}
   end<% end %>
 end


SteffenDE (approved):

SteffenDE:
Thank you! Sure feels like we should have caught that in review 😅

SteffenDE:
@josevalim should we have a separate topic that only broadcasts messages for a specific ID?

josevalim:
The issue with doing it per ID is that we deliver double messages? Another option is to deal with updates only on the index page for the scaffold.

SteffenDE:
How expensive if sending a PubSub message to a topic where nobody is subscribed? It's only the messages to the other nodes, right? Having update/delete broadcast to two topics is a pattern that I frequently used already. But the catch all might be just fine here? I think it depends on which way we want to push people towards.

I think if we do PubSub, all pages should update, so I wouldn't exclude the show page.

josevalim:
Yes, the issue is when you have multiple nodes, two messages across the nodes. The catch-all is fine here. :)

josevalim (approved):

SteffenDE:
Let's go with this one then :)

Thank you @nshafer!

nshafer:
No problem, thank you both. =)

PR #6212, opened 2025-04-22
Fix server-error css selector

andrewtimberlake:
The `#server-error` was targeting `.phx-client-error #client-error`


SteffenDE:
Thank you! 🙌🏻

PR #6213, opened 2025-04-22
Add w-fit to logo to limit is clickable area

phips4:
Add w-fit to logo to limit is clickable area

Currently, the logo’s anchor tag in the header spans the full width, resulting in an invisible but clickable area. This could lead to unexpected behavior and poor UX. By restricting the anchor tag’s size using w-fit, it now only spans its actual content.

Before:
Entire header area is clickable (even outside the logo)
grafik
with red background for demonstation
grafik

After:
Only the logo itself is clickable (no excess clickable area, no pointer cursor outside the logo)
grafik
(with red background for demonstation)
grafik


SteffenDE:
Thank you! 🙌🏻

PR #6223, opened 2025-04-28
Phoenix.Debug

SteffenDE:
Relates to: https://github.com/phoenixframework/phoenix_live_view/pull/3776

### DIFF
--- a/lib/phoenix/debug.ex
+++ b/lib/phoenix/debug.ex
@@ -6,7 +6,10 @@ defmodule Phoenix.Debug do
   """
 
   @doc """
-  Returns a list of all currently connected socket processes.
+  Returns a list of all currently connected `Phoenix.Socket` transport processes.
+
+  Note that custom sockets implementing the `Phoenix.Socket.Transport` behaviour
+  are not listed.
 
   Each process corresponds to one connection that can have multiple channels.
 
@@ -17,30 +20,30 @@ defmodule Phoenix.Debug do
 
   ## Examples
 
-      iex> Phoenix.Debug.list_sockets()
+      iex> Phoenix.Debug.list_channel_sockets()
       [#PID<0.123.0>]
 
   """
-  def list_sockets do
-    for pid <- Process.list(), socket?(pid) do
+  def list_channel_sockets do
+    for pid <- Process.list(), channel_socket?(pid) do
       pid
     end
   end
 
   @doc """
-  Returns true if the given pid is a LiveView process.
+  Returns true if the given pid is a `Phoenix.Socket` transport process.
 
   ## Examples
 
-      iex> Phoenix.Debug.list_sockets() |> Enum.at(0) |> socket?()
+      iex> Phoenix.Debug.list_channel_sockets() |> Enum.at(0) |> channel_socket?()
       true
 
-      iex> socket?(pid(0,456,0))
+      iex> channel_socket?(pid(0,456,0))
       false
 
   """
-  def socket?(pid) do
-    # Sockets set the "$process_label" to {Phoenix.Socket, handler_module, id}
+  def channel_socket?(pid) do
+    # `Phoenix.Socket`s set the "$process_label" to {Phoenix.Socket, handler_module, id}
     with info when is_list(info) <- Process.info(pid, [:dictionary]),
          {:dictionary, dictionary} <- List.keyfind(info, :dictionary, 0),
          {:"$process_label", label} <- List.keyfind(dictionary, :"$process_label", 0),
@@ -52,11 +55,11 @@ defmodule Phoenix.Debug do
   end
 
   @doc """
-  Returns a list of all currently connected channels for the given socket pid.
+  Returns a list of all currently connected channels for the given `Phoenix.Socket` pid.
 
   ## Examples
 
-      iex> Phoenix.Debug.list_sockets |> Enum.at(0) |> Phoenix.Debug.channels()
+      iex> Phoenix.Debug.list_channel_sockets() |> Enum.at(0) |> Phoenix.Debug.channels()
       {:ok,
        [
          %{pid: #PID<0.1702.0>, status: :joined, topic: "lv:phx-GDp9a9UZPiTxcgnE"},
--- a/lib/phoenix/debug.ex
+++ b/lib/phoenix/debug.ex
@@ -44,11 +44,7 @@ defmodule Phoenix.Debug do
     with info when is_list(info) <- Process.info(pid, [:dictionary]),
          {:dictionary, dictionary} <- List.keyfind(info, :dictionary, 0),
          {:"$process_label", label} <- List.keyfind(dictionary, :"$process_label", 0),
-         true <-
-           match?(
-             {Phoenix.Socket, mod, id} when is_atom(mod) and (is_binary(id) or is_nil(id)),
-             label
-           ) do
+         {Phoenix.Socket, mod, id} when is_atom(mod) and (is_binary(id) or is_nil(id)) <- label do
       true
     else
       _ -> false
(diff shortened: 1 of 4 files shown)


lib/phoenix/debug.ex
```diff
@@ -0,0 +1,89 @@
...
+  def socket?(pid) do
+    # Sockets set the "$process_label" to {Phoenix.Socket, handler_module, id}
+    with info when is_list(info) <- Process.info(pid, [:dictionary]),
+         {:dictionary, dictionary} <- List.keyfind(info, :dictionary, 0),
+         {:"$process_label", label} <- List.keyfind(dictionary, :"$process_label", 0),
+         true <-
+           match?(
+             {Phoenix.Socket, mod, id} when is_atom(mod) and (is_binary(id) or is_nil(id)),
+             label
+           ) do
```
josevalim, line 51:
```suggestion
         {Phoenix.Socket, mod, id} when is_atom(mod) and (is_binary(id) or is_nil(id)) <- label do
```

lib/phoenix/debug.ex
```diff
@@ -0,0 +1,89 @@
...
+  retains the same socket connection. Nested LiveViews also share the same
+  connection, each being a different channel. See `Phoenix.Debug.channels/1`.
+
+  ## Examples
+
+      iex> Phoenix.Debug.list_sockets()
+      [#PID<0.123.0>]
+
+  """
+  def list_sockets do
```
josevalim, line 24: I am wondering if this is a misnomer. We are not listing all sockets, but perhaps the ones that have the channels implementation. Someone could still define their own socket with their own protocol and then not reply to our messages.

Given this, I would perhaps focus on providing `list_channels` instead, and get that direct result first. We can revisit and add `list_sockets` later.

SteffenDE: Do you have an example where we'd expect this to happen? I don't see us expecting someone to define a socket without `use`ing `Phoenix.Socket` anywhere.

josevalim: See the docs for `Phoenix.Socket.Transport` :)

SteffenDE: Aah I always forget about `Phoenix.Socket.Transport`... Maybe because the naming there seems a bit off to me? For me, the "transport" is WebSocket / Longpoll, but `Phoenix.Socket.Transport` is not really a custom transport, but still on top of WebSocket / Longpoll.

Anyway, I kept the extra function and renamed it to `list_channel_sockets`, because to get the actual channels, I'd either need to list the children of the channel supervisor or send messages to the socket processes and I wanted to keep it simple (filtering Process.list).

lib/phoenix/debug.ex
```diff
@@ -0,0 +1,88 @@
...
+         %{pid: #PID<0.1702.0>, status: :joined, topic: "lv:phx-GDp9a9UZPiTxcgnE"},
+         %{pid: #PID<0.1727.0>, status: :joined, topic: "lv:sidebar"}
+       ]}
+
+      iex> Phoenix.Debug.channels(pid(0,456,0))
+      {:error, :not_alive}
+
+  """
+  def channels(socket_pid) do
+    ref = make_ref()
```
josevalim, line 74: Should this perhaps be called `list_channels` and go through a similar code as `list_channels_sockets`, which uses `Process.info` to see if they are a channel or not, rather than sending a debug channels message? Especially if we are thinking about making as straight-forward to use as possible?

lib/phoenix/debug.ex
```diff
@@ -0,0 +1,88 @@
...
+  ## Examples
+
+      iex> Phoenix.Debug.list_channel_sockets() |> Enum.at(0) |> channel_socket?()
+      true
+
+      iex> channel_socket?(pid(0,456,0))
+      false
+
+  """
+  def channel_socket?(pid) do
```
josevalim, line 45: Perhaps it should be private?

lib/phoenix/debug.ex
josevalim: I would kinda of expect the channels to not include LiveViews indeed but if you disagree, then I think this is good to go. Perhaps we just add `list_channels` as a shortcut (i.e. a `flat_map`).

lib/phoenix/debug.ex
SteffenDE: I think it's fine to have it public.

lib/phoenix/debug.ex
```diff
@@ -57,32 +96,74 @@ defmodule Phoenix.Debug do
   @doc """
   Returns a list of all currently connected channels for the given `Phoenix.Socket` pid.
 
+  Each channel is represented as a map with the following keys:
+
+    - `:pid` - the pid of the channel process
+    - `:status` - the status of the channel
+    - `:topic` - the topic of the channel
+    - `:type` - the type of the channel.
+      It is `:channel` for channels defined with `use Phoenix.Channel` and `:custom` for custom channels (like LiveViews)
```
josevalim, line 105:
```suggestion

  It is `:channel` for channels defined with `use Phoenix.Channel` and `:custom` for custom channels (like LiveViews)
```

lib/phoenix/debug.ex
SteffenDE: I kept the list functions "cheap" for now, returning a map but only with information that does not require sending a message to the pid.

lib/phoenix/debug.ex
```diff
@@ -57,32 +96,74 @@ defmodule Phoenix.Debug do
...
+    if Process.alive?(socket_pid) and channel_socket?(socket_pid) do
       send(socket_pid, {:debug_channels, ref, self()})
 
       receive do
-        {:debug_channels, ^ref, channels} -> {:ok, channels}
+        {:debug_channels, ^ref, channels} ->
+          channels =
+            Enum.map(channels, fn %{pid: pid} = info ->
+              Map.put(info, :type, (channel?(pid) && :channel) || :custom)
+            end)
```
josevalim, line 132: Perhaps we should not do this additional work by default. Instead, return as is and we document that they can use `channel?` to check.

lib/phoenix/debug.ex
josevalim: I am just bothered about the name a bit. We have `socket` which returns a `Phoenix.Socket` and this is not it. :( I guess the same issue applies to `list_channel_sockets` though...

SteffenDE: Maybe `channel_socket_process?` and `list_channel_socket_info` / `list_channel_sockets_info` 🤷🏻‍♂️

SteffenDE: alright, private it is!

josevalim (approved):

lib/phoenix/debug.ex
SteffenDE: oh fwiw in the LiveView PR I called it `liveview_process?`: https://github.com/phoenixframework/phoenix_live_view/pull/3776/files#diff-073229604d21f1e1dbd1c71bbd2448931f7a0a9bfc99fcbe3fb2e7a3882e1049R89

josevalim: That works then!

SteffenDE: So should we add it as `channel_socket_process?`?

But now I'm wondering if we should just go with `list_sockets` and `socket_process?` and just clearly state that it only lists sockets defined with `use Phoenix.Socket` and not custom `Phoenix.Socket.Transport`s?

josevalim: list_socket and socket_process? Works for me.

SteffenDE: Alright, we're back at [this comment](https://github.com/phoenixframework/phoenix/pull/6223#discussion_r2073083243) then, but I think it's fine with an explanation in the docs.

josevalim (approved):

PR #6230, opened 2025-05-05
 allow overriding auto-detected OTP and Elixir versions

BartOtten:
The Dockerfile generator now accepts explicit `otp` and `elixir` parameters for its base image. This change is necessary because the versions auto‑detected on the development host aren’t guaranteed to exist on Docker Hub, which could otherwise lead to missing‑image build failures. By providing a way to pin the desired OTP and Elixir versions, users can enforce referencing published images.


SteffenDE:
Nice! Thank you :)

PR #6239, opened 2025-05-09
Allow download attr on `<.button>`

LostKobrakai:
This is quite a useful attribute for linking controllers from LVs without disrupting the LVs javascript livecycle


SteffenDE:
🙌🏻

PR #6242, opened 2025-05-11
use single source of truth for socket token authn docs

### LINKED ISSUE #6241: Misalignment in socket authentication documentation
Hi,

There is some misalignment in the documentation for socket authentication in the channels guide and in the generated `socket.js` file from `phx.gen.socket`.

The guide suggests using the transport agnostic method of sending the user token through a header, while the file suggests using the older approach of sending it as a query parameter.

https://github.com/phoenixframework/phoenix/blob/69e2cfa13b94e75bb749bd912f3727e9e4fd410c/guides/real_time/channels.md?plain=1#L453-L457

https://github.com/phoenixframework/phoenix/blob/2875c54cddec2fc45258baee9a84cc8193cad541/priv/templates/phx.gen.socket/socket.js#L7-L9

veselyn:
Removes outdated token authentication documentation from the generated `socket.js` file from `phx.gen.socket`. Instead, it reduces the maintenance of the generator by providing a link to the token authentication section in the guides.

Also, redacts the rest of the docs to reference the new method for authenticating.

Closes #6241.

### DIFF
--- a/guides/real_time/channels.md
+++ b/guides/real_time/channels.md
@@ -107,7 +107,8 @@ In your Phoenix app's `Endpoint` module, a `socket` declaration specifies which 
 ```elixir
 socket "/socket", HelloWeb.UserSocket,
   websocket: true,
-  longpoll: false
+  longpoll: false,
+  auth_token: true
 ```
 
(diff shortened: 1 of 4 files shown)


SteffenDE (commented):
Nice!

SteffenDE:
Thank you!

PR #6248, opened 2025-05-13
Fixes to Cross-context Boundaries guide

jdiago:
Ran into some errors while going through the context guides in `1.8.0-rc.3` along with an update to a diff section that looked old.


josevalim:

PR #6255, opened 2025-05-18
Update esbuild to 0.25.4

prem-prakash:


josevalim:
Done in #6263 along other deps, thank you

PR #6257, opened 2025-05-21
Clarification of scope configuration

dmitrykleymenov:


guides/authn_authz/scopes.md
```diff
@@ -471,7 +471,7 @@ defmodule MyAppWeb.ConnCase do
 end
 ```
 
-Now that our scope configuration includes the `route_prefix` and `route_access_path`, we can generate resources scoped to the organization, and all paths will be automatically generated with the correct organization slug:
+Now that our scope configuration includes the `route_prefix`, and `route_access_path` is default(in our case `[:organization]`), we can generate resources scoped to the organization, and all paths will be automatically generated with the correct organization slug:
```
SteffenDE, line 474: I'd just not mention it there:

```suggestion
Now that our scope configuration includes the `route_prefix`, we can generate resources scoped to the organization, and all paths will be automatically generated with the correct organization slug:
```

SteffenDE:
Thank you! 🙌🏻

PR #6264, opened 2025-05-25
Update guide file: mix_phx_gen_auth.md (`login` vs `log in` change)

zorn:
In this case, I believe we want the verb `log in` and not the adjective `login`.


SteffenDE:
Thank you 🙌🏻

PR #6265, opened 2025-05-26
Run after compiler callbacks in Phoenix

josevalim:


lib/phoenix/endpoint/supervisor.ex
```diff
@@ -244,7 +244,7 @@ defmodule Phoenix.Endpoint.Supervisor do
       reloadable_apps: nil,
       # TODO: Gettext had a compiler in earlier versions,
       # but not since v0.20, so we can remove it here eventually.
-      reloadable_compilers: [:gettext, :elixir, :app],
+      reloadable_compilers: [:live_view, :gettext, :elixir, :app],
```
SteffenDE, line 247: I think we called it `:phoenix_live_view`, but please double check

SteffenDE (approved):
I don't know much about the code reloader, but it works fine with my colocated hooks example :)

josevalim:

Issue #6249, opened 2025-05-15
Log out link in generated project doesn't work [1.8.0-rc.3]


maxsalven:
### Environment

* Elixir version (elixir -v): 1.17.3
* Phoenix version (mix deps): 1.8.0-rc.3
* Operating system: MacOS

If I use the generators (some steps omitted for brevity):
```bash
mix phx.new demo --no-assets
mix phx.gen.auth Accounts User users
```

And choose the no liveview option in the auth step, then the resulting site has a broken "Log out" link on the top right:

<img width="1492" alt="Image" src="https://github.com/user-attachments/assets/65fea199-e9ca-4484-93eb-23d3d2fba753" />

SteffenDE:
Hey @maxsalven,

I cannot reproduce this, do you try to open that link in a new tab or do you have JavaScript disabled? [`phoenix_html`](https://github.com/phoenixframework/phoenix_html/blob/main/priv/static/phoenix_html.js) is responsible for converting that link into a `DELETE` request and you can see the route for it, so something causes your browser to use a regular GET request instead.

maxsalven:
I just tried again from scratch, I still get the bug.

I don't have javascript disabled, everything else works fine in the app.

```
mix phx.new demo --no-assets
cd demo
mix ecto.create
mix phx.gen.auth Accounts User users --no-live
mix deps.get
mix ecto.migrate
mix phx.server
```

Then register for a new account, click the link in the dev mailbox, and then click Log out.

I'm on `rc.3` (I note that `curl https://new.phoenixframework.org/myapp | sh` uses rc.0)

phoebe100:
The '--no-assets' is the issue

SteffenDE:
aah yeah I did not pass `--no-assets`, that's why it worked for me. `phx.gen.auth` does not expect to be used with `--no-assets` - a PR to the docs to clarify is welcome!

maxsalven:
It's perhaps worth noting that if you use the `--no-html` flag:

```
mix phx.new demo --no-html
```

And then try install `phx.gen.auth`, you get a clear error message:

```
** (Mix) mix phx.gen.auth requires phoenix_html

mix phx.gen.auth must be installed into a Phoenix 1.5 app that
contains ecto and html templates.

    mix phx.new my_app
    mix phx.new my_app --umbrella
    mix phx.new my_app --database mysql

Apps generated with --no-ecto or --no-html are not supported.
```

[...]

PR #6272, opened 2025-05-28
warn when using phx.gen.auth without esbuild

SteffenDE:
Closes https://github.com/phoenixframework/phoenix/issues/6249.

### DIFF
--- a/lib/mix/tasks/phx.gen.auth.ex
+++ b/lib/mix/tasks/phx.gen.auth.ex
@@ -272,6 +272,23 @@ defmodule Mix.Tasks.Phx.Gen.Auth do
     if generated_with_no_html?() do
       raise_with_help("mix phx.gen.auth requires phoenix_html", :phx_generator_args)
     end
+
+    if generated_with_no_assets_or_esbuild?() do
+      Mix.shell().yes?("""
+      Warning: did not find phoenix_html in your app.js.
+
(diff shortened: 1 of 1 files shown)


lib/mix/tasks/phx.gen.auth.ex
```diff
@@ -285,6 +302,12 @@ defmodule Mix.Tasks.Phx.Gen.Auth do
     |> Kernel.not()
   end
 
+  defp generated_with_no_assets_or_esbuild? do
+    not File.exists?("assets/js/app.js") or
+      File.read!("assets/js/app.js") =~
+        "For Phoenix.HTML support, including form and button helpers"
+  end
```
josevalim, line 309: I am afraid this will have false positives if someone chooses to have their assets elsewhere. Also asserting on the text may be brittle.

My suggestion would be:

```suggestion
  defp generated_with_no_assets_or_esbuild? do
    not Code.ensure_loaded?(Phoenix.HTML) or
      case File.read("assets/js/app.js") do
        {:ok, content} -> content =~ "priv/static/phoenix_html.js"
        {:error, _} -> true
      end
  end
```

Although I think this means we won't catch the --no-assets case... but then I am not sure if there is much we can do.

josevalim: I think we can check the case where we generate the file without imports but that’s it.

chrismccord:
❤️❤️❤️🐥🔥

PR #6275, opened 2025-05-30
Grow first column for list

LostKobrakai:
Daisy has a weird way to setup lists, but at least the way the generators work col1 is meant to grow.


SteffenDE:
Do you have an example where this is needed? The list in CoreComponents only ever has a single column anyway, right?

LostKobrakai:
Yes, for that reason we want the column to expand to the whole width. The default column width is `minmax(0, auto)` (not expanding to parent width e.g. for background color) and only `1fr` for either the 2nd column (default) or the one provided with `.list-col-grow`.

Using a css grid here seems strange to begin with and really only starts to make sense once there are multiple columns.

My specific usecase was that I wanted to flexbox a button to the right of the text.

SteffenDE:
👍

PR #6280, opened 2025-06-02
Harden CI

maennchen:
### What’s changed

* **Actions pinned to commit SHAs** – Every third-party Action reference now points to an immutable Git commit instead of a moving tag.
* **Least-privilege permissions** – The workflow’s top-level token is limited to `contents:read`; individual jobs elevate only to the scopes they actually need (e.g., `contents:write`).

### Why it matters

Pinning Actions guards against supply-chain attacks and unexpected upstream changes, giving us deterministic, auditable builds. Restricting the default token to read-only follows GitHub’s least-privilege guidance, reducing the blast radius if a job is ever compromised while still allowing specific jobs to perform their required tasks.

### DIFF
--- a/.github/workflows/assets.yml
+++ b/.github/workflows/assets.yml
@@ -6,24 +6,29 @@ on:
       - main
       - "v*.*"
 
+permissions:
+  contents: read
+
 jobs:
   build:
     runs-on: ubuntu-24.04
     env:
       elixir: 1.18.3
       otp: 27.2
+    permissions:
+      contents: write # for stefanzweifel/git-auto-commit-action to push code in repo
     steps:
       - name: Checkout
-        uses: actions/checkout@v4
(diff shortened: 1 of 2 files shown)


SteffenDE:
Thank you! Can we get dependabot to send PRs for updating actions?

maennchen:
That’s already set up: https://github.com/phoenixframework/phoenix/blob/2d945b82dd85091f05acd7da9628b20043d70f9e/.github/dependabot.yml#L4 
Dependabot understands the sha / comment combo and will update both at the same time.

maennchen:
Elixir Example: https://github.com/elixir-lang/elixir/pull/14537

jeregrine (approved):

Gazler:
Thanks <3

maennchen:
Btw for anybody wanting to do the same to their project: I used this tool to generate the changes: https://app.stepsecurity.io/secure-workflow (minus the `step-security/harden-runner` part)

PR #6283, opened 2025-06-03
Update Node Deps

maennchen:
## Reasoning

While scanning phoenix for known vulnerabilities, I found two in the node dependencies:

* GHSA-968p-4wvh-cqc8
* GHSA-g3ch-rx76-35fx

Since they are only in the build tooling, **they don't result in any actual security issues**. I still thought it would be good to do a quick update on the Node tooling.

## Changes

* Update all node deps
* Switch from `@stylistic/eslint-plugin-js` (deprecated) to `@stylistic/eslint-plugin`
* Fix new linter errors


SteffenDE:
Thank you! Can you please revert the changes to the priv directory in the PR? Those get built by CI automatically :)

maennchen:
@SteffenDE  Sure, that makes sense. I've pushed it without.

SteffenDE:
🙌🏻

PR #6284, opened 2025-06-03
Setup Dependabot for NPM

maennchen:


SteffenDE (approved):

josevalim:
Thank you! Is it only for security updates?

maennchen:
@josevalim No, for all NPM Dependencies.

You could enable security updates in the project settings and discard thie PR if you only want security updates.

It is however probably a good idea keeping everything more or less up to date.

If you’re concerned about the amount of PRs, we could also decrease the update interval.

SteffenDE:
🙌🏻

PR #6286, opened 2025-06-04
Merged variant to rest[:class] and use @rest[:class] instead of @clas…

z7ealth:
Button in core_components.ex was not merging rest[:class] in class attribute.


SteffenDE:
Thank you, we already have an open discussion about this in https://github.com/phoenixframework/phoenix/pull/6273. Closing in favor of that one :)

PR #6288, opened 2025-06-06
Update channels.md to include correct way to update the channel params for reconnections

jamauro:
The existing docs point to an older repo that isn't using the params function so it's a bit confusing. I believe Phoenix changed this behavior at some point.

I thought it'd also be nice to have a simple example right in the docs. I took most of this from an example that Chris McCord shared in the elixir forums.


SteffenDE (approved):

SteffenDE:
🙌🏻

PR #6289, opened 2025-06-09
Classes should always override

josevalim:
After discussion with @chrismccord. I haven't tested it, please double check and merge @SteffenDE!


SteffenDE:
🙌🏻

PR #6290, opened 2025-06-09
Migrate from fieldset to div, closes #6204

### LINKED ISSUE #6204: CoreComponents use deprecated daisyUI `fieldset-label` class and missing a `legend` tag
### Environment

* Elixir version (elixir -v): 1.18.3
* Phoenix version (mix deps): latest main
* Operating system: macOS

### Actual behavior

I noticed that daisyUI recently deprecated the `fieldset-label` class, which is being used inside the `core_components.ex`.

See https://github.com/saadeghi/daisyui/blob/master/packages/daisyui/src/components/fieldset.css#L11

Also, I think the *label* of the `input` component should be wrapped in a legend HTML tag, as it is used more as a title for the fieldset than as a label. This is how it is done in the daisyUI docs, too (see [daisyUI `input` component](https://daisyui.com/components/input/) for example).

Although the `legend` tag is technically optional, it should be used for a11y reasons.

> The first element inside the fieldset must be a legend element, which provides a label or description for the group.

See [H71: Providing a description for groups of form controls using fieldset and legend elements](https://www.w3.org/TR/WCAG-TECHS/H71.html)

That's why I suggest refactoring the `input` component to something like this:

```diff
   def input(assigns) do
     ~H"""
     <fieldset class="fieldset mb-2">
-      <label>
-        <span :if={@label} class="fieldset-label mb-1">{@label}</span>
-        <input
-          type={@type}
-          name={@name}
-          id={@id}
-          value={Phoenix.HTML.Form.normalize_value(@type, @value)}
-          class={[@class || "input w-full", @errors != [] && (@error_class || "input-error")]}
-          {@rest}
-        />
-      </label>
+      <legend :if={@label} class="fieldset-legend">{@label}</legend>
+      <input
+        type={@type}
+        name={@name}
+        id={@id}
+        value={Phoenix.HTML.Form.normalize_value(@type, @value)}
+        class={[@class || "input w-full", @errors != [] && (@error_class || "input-error")]}
+        {@rest}
+      />
       <.error :for={msg <- @errors}>{msg}</.error>
     </fieldset>
     """
```

The other components like `type: "select"` and `type: "textarea"` would need to be factored, too.

josevalim:


SteffenDE:
🙌🏻

Issue #6139, opened 2025-03-24
pipe_through pipeline should always be first in scope


chrismccord:

josevalim:
I added some notes somewhat related to this here: 133e3b435bae0cd0ae02f8b078cf8703c402ffe1

SteffenDE:
So should we change this then since it is documented now?

josevalim:
I think having any route before is likely a pitfall that we should avoid anyway.

PR #6297, opened 2025-06-12
Add deprecation warning for pipe_through called after routes

chrismccord:
## Summary

Addresses issue #6139 by adding a deprecation warning when `pipe_through` is called after routes are defined in a router scope.

## Problem

Previously, developers could call `pipe_through` after defining routes, which led to confusing behavior where only routes defined *after* the `pipe_through` call would have the pipelines applied. This was a common source of confusion.

## Solution

- **Backwards Compatible**: Added deprecation warning instead of breaking change
- **Clear Guidance**: Comprehensive warning message with before/after examples  
- **Robust Implementation**: Tracks route definitions per scope and warns appropriately

### Core Implementation
- **`lib/phoenix/router/scope.ex`**: Added `has_routes?` field to track when routes are defined and issue warnings in `pipe_through/2`

### Test Coverage  
- **`test/phoenix/router/pipe_through_deprecation_test.exs`**: 7 comprehensive test cases covering all warning scenarios

## Example Warning Output

```
[warning] Calling pipe_through/1 after defining routes is deprecated and may not work as expected.

Routes defined before pipe_through/1 will not have the specified pipelines applied.
Move all pipe_through/1 calls to the beginning of the scope block.

Current scope has routes defined, but pipe_through [:browser] was called after them.

Instead of:
  scope "/" do
    get "/", HomeController, :index      # This route won't get :browser pipeline
    pipe_through [:browser]              # Called too late
    get "/users", UserController, :index # This route gets :browser pipeline
  end

Do this:
  scope "/" do
    pipe_through [:browser]              # Call first
    get "/", HomeController, :index      # Both routes get :browser pipeline
    get "/users", UserController, :index
  end
```

## Testing

- ✅ All new tests pass (7/7)
- ✅ All existing router tests pass
- ✅ No breaking changes - existing code continues to work
- ✅ Warning is appropriately triggered in existing test suite

This provides a smooth migration path for users while preventing the common pitfall that confused developers.

### DIFF
--- a/lib/phoenix/router/scope.ex
+++ b/lib/phoenix/router/scope.ex
@@ -14,7 +14,8 @@ defmodule Phoenix.Router.Scope do
             private: %{},
             assigns: %{},
             log: :debug,
-            trailing_slash?: false
+            trailing_slash?: false,
+            has_routes?: false
 
   @doc """
   Initializes the scope.
@@ -32,6 +33,10 @@ defmodule Phoenix.Router.Scope do
     unless is_atom(plug) do
       raise ArgumentError, "routes expect a module plug as second argument, got: #{inspect(plug)}"
     end
+
+    # Mark that we have routes defined in this scope
+    top = get_top(module)
+    put_top(module, %{top | has_routes?: true})
 
     top = get_top(module)
     path = validate_path(path)
@@ -120,7 +125,35 @@ defmodule Phoenix.Router.Scope do
   """
   def pipe_through(module, new_pipes) do
     new_pipes = List.wrap(new_pipes)
-    %{pipes: pipes} = top = get_top(module)
+    %{pipes: pipes, has_routes?: has_routes?} = top = get_top(module)
+
+    # Issue deprecation warning if routes have already been defined
+    if has_routes? do
+      require Logger
+
+      Logger.warning("""
+      Calling pipe_through/1 after defining routes is deprecated and may not work as expected.
+
+      Routes defined before pipe_through/1 will not have the specified pipelines applied.
+      Move all pipe_through/1 calls to the beginning of the scope block.
+
+      Current scope has routes defined, but pipe_through #{inspect(new_pipes)} was called after them.
+
+      Instead of:
+        scope "/" do
+          get "/", HomeController, :index      # This route won't get :browser pipeline
+          pipe_through [:browser]              # Called too late
+          get "/users", UserController, :index # This route gets :browser pipeline
+        end
+
+      Do this:
+        scope "/" do
+          pipe_through [:browser]              # Call first
+          get "/", HomeController, :index      # Both routes get :browser pipeline
+          get "/users", UserController, :index
+        end
+      """)
+    end
 
     if pipe = Enum.find(new_pipes, &(&1 in pipes)) do
       raise ArgumentError,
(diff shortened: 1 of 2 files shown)


lib/phoenix/router/scope.ex
```diff
@@ -33,6 +34,10 @@ defmodule Phoenix.Router.Scope do
       raise ArgumentError, "routes expect a module plug as second argument, got: #{inspect(plug)}"
     end
 
+    # Mark that we have routes defined in this scope
+    top = get_top(module)
+    put_top(module, %{top | has_routes?: true})
+
```
SteffenDE, line 40: I assume we can skip the second get_top call in line 41?

```suggestion
    # Get the top scope and mark that we have routes defined
    top = %{get_top(module) | has_routes?: true}
    put_top(module, top)
```

pipe_through_analysis.md
```diff
@@ -0,0 +1,33 @@
...
+4. Update error message to guide users to proper usage
+
+## Files to Modify
+- `lib/phoenix/router.ex` - Add validation logic
+- `test/phoenix/router/routing_test.exs` - Add test cases
+
+## Test Cases Needed
+- ✅ Valid: `pipe_through` before any routes
+- ❌ Invalid: `pipe_through` after routes defined
+- ❌ Invalid: Multiple `pipe_through` calls with routes in between
```
SteffenDE, line 33: Let's remove that file?

test/mix/tasks/phx.gen.presence_test.exs
```diff
@@ -1,4 +1,4 @@
-Code.require_file "../../../installer/test/mix_helper.exs", __DIR__
+Code.require_file("../../../installer/test/mix_helper.exs", __DIR__)
```
SteffenDE, line 1: Those changes are unrelated, can we undo them? Same with phx.gen.presence itself.

We should mix format the whole repository at some point.

lib/phoenix/router/scope.ex
```diff
@@ -120,7 +125,35 @@ defmodule Phoenix.Router.Scope do
...
   def pipe_through(module, new_pipes) do
     new_pipes = List.wrap(new_pipes)
-    %{pipes: pipes} = top = get_top(module)
+    %{pipes: pipes, has_routes?: has_routes?} = top = get_top(module)
+
+    # Issue deprecation warning if routes have already been defined
+    if has_routes? do
+      require Logger
+
+      Logger.warning("""
```
SteffenDE, line 134: Aren't deprecation warnings usually done with `IO.warn`?

test/phoenix/router/pipe_through_deprecation_test.exs
```diff
@@ -0,0 +1,191 @@
...
+      end)
+
+      # Test that routing still works
+      conn = call(Phoenix.Router.PipeThroughDeprecationTest.TestRouter7, :get, "/first")
+      assert conn.status == 200
+      assert conn.resp_body == "users index"
+
+      conn = call(Phoenix.Router.PipeThroughDeprecationTest.TestRouter7, :get, "/second")
+      assert conn.status == 200
+      assert conn.resp_body == "users show"
```
SteffenDE, line 188: I assume it should actually check that it is called?

```suggestion
      assert conn.resp_body == "users show"
      assert conn.assigns.browser_called
```

SteffenDE (commented):
Note that @josevalim added a route in tests that now has that warning as well, that's why the current tests are flaky.

test/phoenix/router/pipe_through_deprecation_test.exs
```diff
@@ -0,0 +1,191 @@
+defmodule Phoenix.Router.PipeThroughDeprecationTest do
```
josevalim, line 1: We should not have a new test file only for this and the tests can definitely be shorter. We only need to check for the warning, no need to for the other success cases.

SteffenDE:
Please reopen if necessary.

PR #6309, opened 2025-06-18
use new Ecto 3.13 features

SteffenDE:


priv/templates/phx.gen.auth/context_functions.ex
```diff
@@ -284,16 +280,16 @@
...
+    Repo.transact(fn ->
+      with {:ok, <%= schema.singular %>} <- Repo.update(changeset) do
+        query = <%= inspect schema.alias %>Token.by_<%= schema.singular %>_and_contexts_query(<%= schema.singular %>, :all)
+        tokens_to_expire = Repo.all(query)
+
+        tokens_to_expire
+        |> <%= inspect schema.alias %>Token.delete_all_query()
+        |> Repo.delete_all()
+
+        {:ok, {<%= schema.singular %>, tokens_to_expire}}
```
SteffenDE, line 292: I changed it to pass that tuple through all places where we previously had the three-tuple.

josevalim (approved):

PR #6312, opened 2025-06-21
Provide better name for validate_email option

josevalim:
The code also validates if the email changed but I think adding uniqueness to the option is good enough?

### DIFF
--- a/priv/templates/phx.gen.auth/settings_live.ex
+++ b/priv/templates/phx.gen.auth/settings_live.ex
@@ -80,7 +80,7 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
 
   def mount(_params, _session, socket) do
     <%= schema.singular %> = socket.assigns.<%= scope_config.scope.assign_key %>.<%= schema.singular %>
-    email_changeset = <%= inspect context.alias %>.change_<%= schema.singular %>_email(<%= schema.singular %>, %{}, validate_email: false)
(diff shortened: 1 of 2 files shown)


SteffenDE:
Do you have more context about this? Looking at this I find just `validate_email` fine as is, but I may be missing the problem.

josevalim:
The issue is that we do other email validations, such as on the format, even when `validate_email: false`.

priv/templates/phx.gen.auth/settings_live.ex
```diff
@@ -80,7 +80,7 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
 
   def mount(_params, _session, socket) do
     <%= schema.singular %> = socket.assigns.<%= scope_config.scope.assign_key %>.<%= schema.singular %>
-    email_changeset = <%= inspect context.alias %>.change_<%= schema.singular %>_email(<%= schema.singular %>, %{}, validate_email: false)
+    email_changeset = <%= inspect context.alias %>.change_<%= schema.singular %>_email(<%= schema.singular %>, %{}, validate_email_uniqueness: false)
```
josevalim, line 83:
```suggestion
    email_changeset = <%= inspect context.alias %>.change_<%= schema.singular %>_email(<%= schema.singular %>, %{}, validate_unique: false)
```

SteffenDE (approved):

PR #6318, opened 2025-06-25
Remove outdated note about user authentication setup in contexts guide

steve-human:
At this stage of the guide, the user has already set up real authentication. This note appears to be outdated and was likely left in by mistake


SteffenDE:
Thank you! 🙌🏻

PR #6321, opened 2025-06-27
Fix missing examples

txssu:
Before

image

After

image

### DIFF
--- a/assets/js/phoenix/index.js
+++ b/assets/js/phoenix/index.js
@@ -26,7 +26,7 @@
  * events are listened for, messages are pushed to the server, and
  * the channel is joined with ok/error/timeout matches:
  *
- * ```javascript
+ * ```
  * let channel = socket.channel("room:123", {token: roomToken})
  * channel.on("new_msg", msg => console.log("Got message", msg) )
  * $input.onEnter( e => {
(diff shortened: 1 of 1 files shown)


SteffenDE:
Thank you!

rhcarvalho:
This is strange, I would think the `documentation` NPM package would support the language hint in Markdown code blocks, just as in the example here:

https://github.com/documentationjs/documentation/blob/8d029e735d2661c8f5e097c285050c7abe5edf6e/docs/NODE_API.md#examples

However, there's indeed a bug which describes exactly what we observed here https://github.com/documentationjs/documentation/issues/1428.

And I confirmed that the code blocks go missing if we set the language to `javascript` (or anything else).

SteffenDE:
@rhcarvalho some time I might be able to finish up https://github.com/phoenixframework/phoenix/pull/6364, then we'll have more modern docs :D

rhcarvalho:
Wow, massive change! The UI looks the same, so perhaps the bug in only manifesting when the source code is JS, not TS?!

Would be nice to have the JS docs surfaced inside of the Elixir docs, single experience! But that might require yet another massive lift.

Hey, thanks for the good work in Phoenix! Upgrading to 1.8 and LV 1.1 was smooth (other than self-inflicted small mistakes on my side, promptly fixed).

PR #6323, opened 2025-06-29
`<.button>` component should allow `disabled` property

glhrmv:
`disabled` is a valid property of HTMLButtonElement, and it already works nicely with DaisyUI & Tailwind, but the compiler currently complains about the attribute. This commit fixes that.


SteffenDE:
🙌🏻

PR #6334, opened 2025-07-02
Enable Tailwind `dark:` with data-theme attribute

rubas:
The Theme Switcher [root.html.heex](https://github.com/phoenixframework/phoenix/blob/main/installer/templates/phx_web/components/layouts/root.html.heex) uses the `data-theme` attribute to toggle between light/dark themes. 
Tailwind's `dark:` variant expects per default a `class="dark"` though.

This custom variant makes Tailwind's `dark:` work with `data-theme="dark"`.

`@custom-variant dark (&:where([data-theme=dark], [data-theme=dark] *));`

Without this, the `dark:` utility will not work, even DaisyUI themes switch correctly.

👉 Reference: https://tailwindcss.com/docs/dark-mode#using-a-data-attribute

### DIFF
--- a/installer/templates/phx_assets/app.css
+++ b/installer/templates/phx_assets/app.css
@@ -96,7 +96,7 @@
 @custom-variant phx-submit-loading (.phx-submit-loading&, .phx-submit-loading &);
 @custom-variant phx-change-loading (.phx-change-loading&, .phx-change-loading &);
 
-/* Use the data attribute for Dark Mode  */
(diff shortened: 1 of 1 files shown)


Gazler:
This is how daisyUI recommends doing it too 
https://daisyui.com/docs/themes/#how-to-apply-tailwinds-dark-selector-for-specific-themes

SteffenDE:
Thank you! 🙌🏻

PR #6337, opened 2025-07-05
Fix missing newline in phx.gen.auth template

rhcarvalho:
When the `--binary-id` option is used with `phx.gen.auth`, the generated schema file was missing a blank line between the `import Ecto.Changeset` and the `@primary_key` declaration.

The new placing of the `<%= if ... do %>...<% end %>` construct matches how it is done in the `schema_token.ex` template file.


SteffenDE:
🙌🏻

PR #6338, opened 2025-07-05
Remove deprecated route names from phx.gen.* templates

rhcarvalho:
After running `mix phx.gen.auth` and reviewing the generated code, I noticed that it added the `as: :some_name` option to the `scope` macro in the router.

This option is not there on new projects, and `:as` is documented to have no effect when using Verified Routes (which phx.new uses).

Considering the Router Helpers deprecation status, this commit removes the `as: :some_name` option from the generated code.

It does mean someone using the modern generators within a project using Helpers would have to manually add the `as: :some_name` option if they want to have a `some_name_path` function.


josevalim:

PR #5605, opened 2023-10-16
docs: add routing cheatsheet

pmargreff:
Adding cheatsheets for routing syntax. The Phoenix guidelines are super detailed and awesome, but there are moments when I just need to grab the syntax and go without the deep dive.

I plan to do the same for plugs, however, I'd appreciate community feedback on this one before proceeding.

### Screenshot

image

PR #6339, opened 2025-07-05
Update Routing Cheatsheet/Guide focused on Verified Routes

rhcarvalho:
When the Cheatsheet was created, it got feedback from the Phoenix team to focus on Verified Routes instead of Router Helpers (https://github.com/phoenixframework/phoenix/pull/5605#issuecomment-1779497054).

This takes the guide one step further by removing the usage of `as: :some_name` from the `scope` and `resources` macros, which was previously used to generate router helpers.

It also updates the examples to showcase how Verified Routes are typically used, by interpolating variables/assigns directly into the path.

Finally, it removes the `as: :some_name` option from the nested `scope` calls in the Routing guide, as it was also not relevant in that context, because the guide is focused on Verified Routes usage.

### DIFF
--- a/guides/cheatsheets/router.cheatmd
+++ b/guides/cheatsheets/router.cheatmd
@@ -14,9 +14,9 @@ get "/users", UserController, :index
 patch "/users/:id", UserController, :update
 ```
 ```elixir
-# generated routes
+# Verified Routes
 ~p"/users"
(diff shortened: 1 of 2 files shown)


SteffenDE (approved):

SteffenDE:
🙌🏻

PR #6340, opened 2025-07-07
Fix missing verified route ~p sigil in generated test

rhcarvalho:
Found while working with `phx.gen.auth` generated code. I looked for the pattern `"([^/]+)?/[^/]+"` in the templates and found no other occurrences. Hopefully this was the only one.


SteffenDE:
🙌🏻

PR #6350, opened 2025-07-10
Generate session token function improved naming

mrroot5:
This is my first PR, I didn't see any issue related to this.

I realized this function uses `Repo.insert!` which will raise an error if it fails. I changed the function naming adding a `!` to prevent developers about the possible raising.


josevalim:
We usually have a bang version as a counter part to a non bang one. It is meant to highlight when things could fail depending on the input or external condition. In this case, the failure would always be a bug, since token insertion should always succeed. In any case, thanks!

PR #6351, opened 2025-07-10
Remove sigil_S from docstrings in templates

rhcarvalho:
When reviewing generated code, I noticed that some docstrings used `~S` unnecessarily.

The case for `@doc ~S"""` is when examples or prose contain backslashes or `#{...}` interpolation, which is not the case here.


josevalim:

PR #6352, opened 2025-07-10
Clarify Swoosh mailer config instructions in templates

rhcarvalho:
- Reword to avoid talking about "Swoosh API client" both before and after the example
- Indicate API client is configured at compile-time in config/prod.exs
- Use Req in the example, aligned with what is configured by default


josevalim:

PR #6353, opened 2025-07-11
Add 'For Controllers' when referring to the conn scope

dhonysilva:
A small text to instruct the reader which is the assign definition for Controllers and which is for LiveViews.


SteffenDE:
I would not add that, since the conn is also useful for LiveViews, for example the `mount_current_scope ` uses `assign_new` to use the value already set in the conn.

dhonysilva:
@SteffenDE that's true. Thanks for raising the examples.

PR #6355, opened 2025-07-14
initialize empty git repo on phx.new

SteffenDE:


SteffenDE:
Yes, the path is always the root path.

josevalim (approved):

PR #6358, opened 2025-07-14
Optional context argument in generators

josevalim:
For now, I am pushing the changes to the guides to explore the ideas. I suggest reading the introduction to contexts guide to see how it looks like: https://github.com/phoenixframework/phoenix/blob/jv-optional-context/guides/data_modelling/contexts.md

### DIFF
--- a/lib/mix/tasks/phx.gen.context.ex
+++ b/lib/mix/tasks/phx.gen.context.ex
@@ -373,8 +373,26 @@ defmodule Mix.Tasks.Phx.Gen.Context do
       if has_context? do
         {maybe_context_name, schema_name_or_plural, plural_or_first_attr, schema_args}
       else
-        {Phoenix.Naming.camelize(schema_name_or_plural), maybe_context_name,
-         schema_name_or_plural, [plural_or_first_attr | schema_args]}
+        # mix phx.gen.live User users name:string
+        # we generate the context from the plural "users" -> Users
+        context = Phoenix.Naming.camelize(schema_name_or_plural)
+
+        if context == maybe_context_name do
+          # if someone did
+          # mix phx.gen.live Users users name
+          Mix.raise("""
+          The given schema #{maybe_context_name} is equal to the camelized version of
+          the table plural #{schema_name_or_plural}, but the schema is expected to be singular.
+
+          Please pass an explicit context option like:
+
+              mix phx.gen.live #{context} #{maybe_context_name} #{schema_name_or_plural}
+
+          if this is what you want.
+          """)
+        end
+
+        {context, maybe_context_name, schema_name_or_plural, [plural_or_first_attr | schema_args]}
       end
 
     cond do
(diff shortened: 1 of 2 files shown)


guides/data_modelling/contexts.md
```diff
@@ -4,11 +4,39 @@ Phoenix guides are broken into several major sections. The main building blocks
...
+If you have used `mix phx.gen.html`, `mix phx.gen.json`, or `mix phx.gen.live`, you have already used contexts. For example, run the following generator in a Phoenix application:
 
-Let's use these ideas to build out our web application. Our goal is to build an ecommerce system where we can showcase products, allow users to add products to their cart, and complete their orders. Opposite to other Phoenix guides, **these guides are meant to be read in order**.
+```console
+$ mix phx.gen.live Post posts title body:text
+```
+
+The command above will output a few files, among them, a `MyApp.Posts.Post` schema in `lib/my_app/posts/post.ex`, which outlines how the resource is represented in the database, and a **context** module named `MyApp.Posts` that encapsulates all the database access to said schema. The `MyApp.Posts` module centralizes all functionality related to posts, instead of scattering logic around controllers, LiveViews, etc.
+
+Contexts are also useful to nest resources. For example, if you are adding comments to your posts, you can colocate their schemas, since comments belongs to posts, like this:
```
SteffenDE, line 17:
```suggestion
Contexts are also useful to nest resources. For example, if you are adding comments to your posts, you can colocate their schemas, since comments belong to posts, like this:
```

guides/data_modelling/your_first_context.md
```diff
@@ -4,7 +4,9 @@ An ecommerce platform has wide-reaching coupling across a codebase so it's impor
...
+When we run the generators, the context name is optional, and Phoenix will automatically use the plural name as the context module. This allows us to keep moving forward when starting out or when it is not yet clear how the different parts of our system related to each other. Luckily, the needs of ecommerces are well defined nowadays, so it provides an excellent ground for us to design with intent. So let's take a step back and think about the different parts of our system.
```
SteffenDE, line 7:
```suggestion
When we run the generators, the context name is optional, and Phoenix will automatically use the plural name as the context module. This allows us to keep moving forward when starting out or when it is not yet clear how the different parts of our system relate to each other. Luckily, the needs of ecommerce systems are well defined nowadays, so it provides an excellent ground for us to design with intent. So let's take a step back and think about the different parts of our system.
```

SteffenDE:
@josevalim I pushed the code changes

lib/mix/tasks/phx.gen.context.ex
```diff
@@ -350,7 +358,25 @@ defmodule Mix.Tasks.Phx.Gen.Context do
...
 
-  defp validate_args!([context, schema, _plural | _] = args, help) do
+  defp validate_args!(
+         [maybe_context_name, schema_name_or_plural, plural_or_first_attr | schema_args],
+         optional,
+         help
+       ) do
+    has_context? =
+      case schema_name_or_plural do
+        <<char::integer-size(8), _rest::binary>> when char in ?A..?Z -> true
```
josevalim, line 368: That's the default, in case you want to skip it.

```suggestion
        <<char, _rest::binary>> when char in ?A..?Z -> true
```

josevalim:
Thank you @SteffenDE! Two nits and I think we are good!

josevalim:
@SteffenDE please merge this whenever you’re ready

PR #6362, opened 2025-07-16
Add @impl true to `phx.gen.auth` LV templates

rhcarvalho:
Matches the style of code generated with `phx.gen.live`, so that a project using both generators produces code following a consistent style.


rhcarvalho:
For reference, the other generators got `@impl true` back in #3745.

SteffenDE:
🙌🏻

PR #6363, opened 2025-07-16
Guides: Removes default arg for context create functions

jdiago:
I went through the context guides again in `rc4` to check on my changes in `rc3`. A minor cause of confusion I missed last time was the current context generators not having a default `attrs` for the create function.

This change removes the default `attrs` from snippets that show the default create function of a context.


SteffenDE:
🙌🏻

PR #6368, opened 2025-07-19
Fix non-closing flash when using --no-assets

sabiwara:
Hi folks!

Currently the flash cannot be closed when the project is generated with `--no-assets`, but it seems to be fixable with some plain Javascript.

Not sure if this is the best way to address it, but since I had to figure this out and it worked for me, I thought maybe it could be upstreamed. Otherwise feel free to close this 🙂

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -54,8 +54,9 @@ defmodule <%= @web_namespace %>.CoreComponents do
     <div
       :if={msg = render_slot(@inner_block) || Phoenix.Flash.get(@flash, @kind)}
       id={@id}
-      phx-click={JS.push("lv:clear-flash", value: %{key: @kind}) |> hide("##{@id}")}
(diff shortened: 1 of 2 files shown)


josevalim (approved):

SteffenDE:
I'm not sure if we should push people towards inline JavaScript. I know that we also generate the theme switcher JS as inline `<script>`, but we could move that to `app.js` instead.

sabiwara:
> I'm not sure if we should push people towards inline JavaScript. I know that we also generate the theme switcher JS as inline `<script>`, but we could move that to `app.js` instead.

@SteffenDE is it what you had in mind?

```js
// app.js
window.hideElement = function (elem) {
  elem.hidden = true;
};
```

```elixir
# core_components.ex
  ...
  onclick="hideElement(this)"
```

This works but it feels a bit weird, given we're just replacing a single instruction by another instruction. Although I understand the concern about inlining.

PR #6369, opened 2025-07-19
Fix code example in guide components.md

jcorentin:
Code example provided was from the wrong file : `layouts.ex` instead of `hello_html.ex`


josevalim:

Issue #6366, opened 2025-07-17
The theme select does not work when `--no-live` is set


joseph-lozano:
### Environment

* Elixir version (elixir -v): 1.19.0-rc.0
* Phoenix version (mix deps): 1.18.0-rc.4
* Operating system: MAC

### Actual behavior

The theme select does not work when `--no-live` is set when doing `mix phx.new`

### Expected behavior

The theme selector is either removed when doing `--no-live` or it works with `--no-live`

SteffenDE:
You're right. We'll have to think about the best solution for this. The theme switcher uses `Phoenix.LiveView.JS`, but the code that handles that is not included in the `app.js`.

michallepicki:
The theme switcher is a nice-to-have but last time I checked there were other more basic things not working with `--no-live`, like e.g. dismissing flash messages

PR #6373, opened 2025-07-24
Do not rely on Phoenix.LiveView.JS when using --no-live

SteffenDE:
Closes https://github.com/phoenixframework/phoenix/issues/6366.
Closes https://github.com/phoenixframework/phoenix/pull/6368.

### DIFF
--- a/installer/templates/phx_web/components/layouts.ex
+++ b/installer/templates/phx_web/components/layouts.ex
@@ -86,7 +86,7 @@ defmodule <%= @web_namespace %>.Layouts do
     ~H"""
     <div id={@id} aria-live="polite">
       <.flash kind={:info} flash={@flash} />
-      <.flash kind={:error} flash={@flash} />
+      <.flash kind={:error} flash={@flash} /><%= if @live do %>
 
       <.flash
         id="client-error"
@@ -110,7 +110,7 @@ defmodule <%= @web_namespace %>.Layouts do
       >
         <%= maybe_eex_gettext.("Attempting to reconnect", @gettext) %>
         <.icon name="hero-arrow-path" class="ml-1 size-3 motion-safe:animate-spin" />
-      </.flash>
+      </.flash><% end %>
     </div>
     """
   end<%= if @css do %>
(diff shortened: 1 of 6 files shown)


installer/templates/phx_web/components/layouts.ex
```diff
@@ -120,32 +120,55 @@ defmodule <%= @web_namespace %>.Layouts do
...
-        phx-click={JS.dispatch("phx:set-theme", detail: %{theme: "dark"})}
         class="flex p-2 cursor-pointer w-1/3"
+        phx-click={JS.dispatch("phx:set-theme")}
+        data-phx-theme="dark"
       >
         <.icon name="hero-moon-micro" class="size-4 opacity-75 hover:opacity-100" />
       </button>
     </div>
-    """
+    """<% else %>
```
SteffenDE, line 152: We need the branch, otherwise it would not be formatted, as the live version wants newlines for the attributes, while the non live does not...

michallepicki:
@SteffenDE Hey! I did not test but I am not sure if this is closing also https://github.com/phoenixframework/phoenix/pull/6368 . I think `--no-assets` means that the `phx_static/app.js` file will be copied, not `phx_assets/app.js`. Did you test both `--no-live` and `--no-assets`?

josevalim (changes requested):
I'd go with a simpler style and not generate a theme toggle if you are using --no-live. Otherwise maintaining all of these combinations become harder and harder.

For flash, I think the current changes are fine!

SteffenDE:
@michallepicki good catch! I did only test with `--no-live`...

@josevalim we're still generating the light and dark daisy themes, so it would be a bit sad to not have a switcher at all. Another way to solve this would be to add a new way to only use Phoenix.LiveView.JS on a page and then keep using JS, e.g.

```javascript
import { initJS } from "phoenix_live_view"
initJS();
```

But that would require bigger changes in LiveView.

josevalim:
> @josevalim we're still generating the light and dark daisy themes, so it would be a bit sad to not have a switcher at all.

I'd say that's fine. They can build on top of that if they want or remove the additional theme. We can also just generate one of the themes. If you pass `--no-live`, you should expect to not get all of the bells and whistles of a LiveView app (and I'd assume people doing `--no-live` want _less_ anyway).

Gazler (approved):
The nested if is a bit hard to read, but that's the nature of the templates we have.

SteffenDE:
@josevalim I've been looking into this again, and I think we should keep the change of moving the theme handling from an inline `<script>` to `app.js` anyway. Then, the only extra code for supporting `--no-live` is

```javascript
document
  .querySelectorAll("[data-theme-toggle] button[data-phx-theme]")
  .forEach((el) => {
    el.addEventListener("click", () => {
      setTheme(el.dataset.phxTheme)
    })
  })
```

and the removal of the extra branch in the theme_toggle, which is only there because of the HTML formatter. So it is really not much to support `--no-live` and I'd vote to just do it.

josevalim:
@SteffenDE I am not sure, with colocated hooks, it seems moving stuff to app.js is even weirder?

Regarding the HTML formatter, we could make it so the formatter respects line breaks if they are already there. So if you write this:

```
<a
  href="foo"
  bar="baz"
>
  ...
</a>
```

Then even if it fits in one line, it remains broken. We do this for Elixir and it should be relatively straight-forward to do it for Phoenix.

SteffenDE:
The integration tests will fail, because LiveView is not updated yet

SteffenDE:
huh seems to be fine actually

josevalim (approved):

PR #6384, opened 2025-07-29
registration_live Optimization

jdiago:
This change adds the `validate_unique: false` option to the `registration_live` changesets on mount and validate to stop the unique constraint checks from running with every keystroke. This mirrors the update email form behavior.


josevalim:

PR #6385, opened 2025-07-30
split AGENTS.md into small parts

SteffenDE:
So I splitted the AGENTS.md into a couple of smaller files. We also differentiate between generic Phoenix rules and rules specifically for new projects. The idea is that if you have an existing project, you can use something like `usage_rules` to get the generic rules from Phoenix (only the ones you need) and `phx.new` creates rules specifically tailored for the project.

We exclude rules based on `--no-ecto`, `--no-live` and `--no-html`.

When running `mix phx.gen.auth` we inject new auth rules into an existing `AGENTS.md` file (we should make this optional behind a flag). The injected text is based on the provided arguments. If someone does `mix phx.gen.auth Admins Admin admins` we talk about `current_scope.admin` etc.; When using the non-live version, text about live_sessions and other LiveView specifics is excluded.

This will probably interest @zachdaniel. I did not test it together with usage_rules yet. Is it supported to only have sub rules and no main `usage_rules.md`?

There another problem we need to solve. The generic rules live in `usage_rules/`, but we need to somehow copy them into the installer package before publishing to hex, otherwise the generic rules won't be available.

### DIFF
--- a/installer/lib/phx_new/generator.ex
+++ b/installer/lib/phx_new/generator.ex
@@ -128,7 +128,7 @@ defmodule Phx.New.Generator do
           project.binding[:esbuild] && project.binding[:tailwind] &&
             @new_project_rules_files["assets.md"],
           # generic usage rules
-          "<!-- usage-rules-start -->",
+          "\n<!-- usage-rules-start -->",
           [
             "<!-- phoenix:elixir-start -->\n",
             @rules_files["elixir.md"],
@@ -160,7 +160,7 @@ defmodule Phx.New.Generator do
           "<!-- usage-rules-end -->"
         ]
         |> Enum.reject(fn part -> part == nil or part == false end)
-        |> Enum.join("\n\n")
+        |> Enum.intersperse("\n")
 
       File.write!(Path.join(project.project_path, "AGENTS.md"), content)
     end


SteffenDE:
Based on https://github.com/phoenixframework/phoenix/commit/50ffaa5aa1c60503f01cd2107edd43f22435f9e7.

zachdaniel:
🔥 awesome stuff!

> This will probably interest @zachdaniel. I did not test it together with usage_rules yet. Is it supported to only have sub rules and no main usage_rules.md?

Yes, it is. 

> phx.new creates rules specifically tailored for the project.

I see where you're going with this, and while it doesn't step on my personal toes, I think it does represent what will be pretty manual work for folks in certain scenarios. Like in the ones explaining to prefer tailwind over daisy etc. In a scenario where someone wants to install a daisy UI component library for example, it would be nice if they had a way to either automatically do it, or to programmatically suggest, *removing* whatever Phoenix's builtin guidance on writing HTML/components is etc.

In my perfect world, anything in AGENTS.md that is above the usage rules would be hand-written by the real project author, and everything else would be programmatically addable/removable via usage rules.

> usage rules format

In order for us to manipulate usage rules, we use delineating comments.

```
<-- usage-rules-start -->
....
<-- usage-rules-end -->
```

and then for each individual rule:

```
<!-- phoenix:html-start -->
<!-- phoenix:html-end -->
```

This allows us to update them later. Without these comments, there is no real way to get any benefit from the pattern. I'm open to suggestions or changes to how we detect which usage rules are which though.

---

You'll also need to include the `usage-rules` file in the phoenix package's files in `mix.exs`

On the general approach, I have two other thoughts:

## Using Project Config

Configuring usage rules in the project's `mix.exs` would be a nice way to bridge the gap to make the experience better when actually running this task later. 

As it stands now, if users later realize "I want some rules I didn't get" or "I've downloaded another package and I want its rules" for example, they will have to correctly select to sync only the specific rules they want, i.e `mix usage_rules.sync AGENTS.md phoenix:html ...` etc. If they try to use `mix usage_rules.sync phoenix:all` they may get rules that render their agent practical unusable w/ their current setup, using features that aren't available etc.

So if something was set into the project's mix config:

```elixir
def project() do
  [
    ...
    usage_rules: [
      file: "AGENTS.md",
      sync: [
        "phoenix:html": true, # if --no-html, this wouldn't be there
        "phoenix:phoenix": true,
        "phoenix:project": true
      ]
    ]
  ]
end
```

Then you could save there which selections should be chosen, and the user could later run `mix usage_rules.sync` (after adding usage_rules), and get only the right context. I'm happy to add this to the usage_rules package. I understand why you might not want to do this since it's actually changing something other than a markdown file for the sake of LLMs, but I also worry about post-installation UX.

Based on my experience, you'll find that new version updates of Phoenix will include updates and fixes to these files as if they were code, so making sure that users have a simple way to get the new rules is important, and knowing which sub-rules they should have is important too.

## Ecto Rules

An example of why it will be good to have the usage rules pattern in place is that the Ash installer doesn't use `--no-ecto` when generating a Phoenix project, but we still don't want people to use those Ecto rules, so we will likely end up slicing out any `phoenix:ecto` usage rules on installation. The fact that we'd be able to do that would be *huge* for onboarding.

SteffenDE:
That makes perfect sense to me, I'll add the comments to the generic rules. Would you also say that ones specific to new projects should have a comment too? Those would generally not be updated after the initial generation (but should also not be removed when syncing).

zachdaniel:
In those cases you're treating it as the users AGENTS.md contents so I'd say no comments and put it at the top of the file above the ones wrapped in comments.

for the config, I'm interested in adding it but would want to agree on what it looks like, and if it would happen automatically in the Phoenix installer.

with the config approach, each package with usage rules can actually patch that config to include all of its rules which would be very cool :)

installer/mix.exs
```diff
@@ -70,4 +71,17 @@ defmodule Phx.New.MixProject do
       source_url_pattern: "#{@scm_url}/blob/v#{@version}/installer/%{path}#L%{line}"
     ]
   end
+
+  defp aliases do
+    [
+      "hex.publish": [
```
SteffenDE, line 77: Is this enough?

zachdaniel: The directory would have to be `usage-rules` specifically.

usage-rules/ecto.md
```diff
@@ -1,3 +1,4 @@
+<!-- phoenix:ecto-start -->
```
zachdaniel, line 1: These shouldn't be included in the file itself, otherwise it will be doubled up when we build the file. You'll want to programmatically do this in your script that copies them.

lib/mix/tasks/phx.gen.auth.ex
```diff
@@ -904,6 +905,26 @@ defmodule Mix.Tasks.Phx.Gen.Auth do
...
+
+    file_path =
+      if Mix.Phoenix.in_umbrella?(File.cwd!()) do
+        Path.expand("../../")
+      else
+        File.cwd!()
+      end
+      |> Path.join("AGENTS.md")
+
+    if File.exists?(file_path) do
```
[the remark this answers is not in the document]
josevalim: Perhaps the same `--no-agents-md` flag?

josevalim (approved):
Looks great, just some nits.

installer/mix.exs
SteffenDE: That's only for the installer, so the name should not matter. I don't expect people to use `usage_rules` with the `phx_new`.

usage-rules/elixir.md
```diff
@@ -1,5 +1,4 @@
-This is a web application written using the Phoenix web framework.
-
+<!-- phoenix:elixir-start -->
```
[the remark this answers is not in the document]
josevalim: Yes, exactly. If I am working on Phoenix vs Broadway vs Nerves, the usage rules for Elixir can be quite different, so I don't think the `usage_rules` package should assume a default set of rules for Elixir/OTP.

installer/mix.exs
zachdaniel: Ah, just didn't realize where we were, sorry.

usage-rules/elixir.md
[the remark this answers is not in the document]
josevalim: Some of the concerns is that some of the Elixir rules you have, such as "not using return", were never a problem to me. But I assume they were to you, in a different context, and that's why they are there. Similarly, I don't see it struggling with structs in regular Elixir projects, but it does happen in Phoenix.

In any case, in Phoenix, we will move forward with Elixir rules specific for Phoenix. But we can't control how they will be used elsewhere.

installer/templates/usage-rules/html.md
josevalim, line 1: Nitpick: should this be called assets instead?

Also, do we need to have these files? Isn't the one in phx_single/project-AGENTS.md enough?

usage-rules/elixir.md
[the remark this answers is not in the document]
josevalim: That's exactly part of the problem. It is really hard to design a process when it is all stochastic. You say "common things that groups of people". How common? How many people? If we simply accept anything, then it might quickly balloon. Anything else is arbitrary based on the chosen probe. So I don't mind a strong hand here given this is an area core maintainers are paying close attention to anyway.

FWIW, none of my other Elixir projects have Elixir specific rules. I think the web app context ends up mixing patterns from other languages/frameworks more often. In any case, this particular topic is pretty much settled on our side, but we will be glad to improve/discuss other areas!

zachdaniel:
One other note on Ecto rules is that I would seriously consider putting them into Ecto itself. We want usage rules to grow and evolve with their respective projects, and one major thing there is that models will never know about newly released features etc. so placing those in usage rules (i.e use `Repo.transact` over `Repo.transaction`) is really beneficial to end users.

zachdaniel:
With that said most of those actually look to be Phoenix+ecto specific (but not all of them)

lib/mix/tasks/phx.gen.auth.ex
```diff
@@ -904,6 +909,38 @@ defmodule Mix.Tasks.Phx.Gen.Auth do
     context
   end
 
+  defp maybe_inject_agents_md(%Context{} = context, paths, binding) do
+    if binding[:agents_md] do
+      auth_content =
+        """
+        <!-- phoenix:gen-auth-start -->
+        #{Mix.Phoenix.eval_from(paths, "priv/templates/phx.gen.auth/AGENTS.md", binding)}
+        <!-- phoenix:gen-auth-end -->
```
SteffenDE, line 918: I added usage rules comments here, but those cannot really be synced, so this would probably break? cc @zachdaniel

I assume I should put it into the "user generated" section instead, so I may need to inject based on the `<!-- usage-rules-start -->` and place it before that?

zachdaniel: Yep, I'd say so. Since this is code that they own, the rules also become their rules.

SteffenDE:
I think this is ready now. We can still talk about adding a config entry to mix.exs later on. I like the idea and see the benefits.

installer/lib/phx_new/generator.ex
```diff
@@ -96,6 +96,61 @@ defmodule Phx.New.Generator do
...
+          "<!-- phoenix:phoenix-end -->",
+          "<!-- phoenix:ecto-start -->",
+          project.binding[:ecto] && @rules_files["ecto.md"],
+          "<!-- phoenix:ecto-end -->",
+          "<!-- phoenix:html-start -->",
+          project.binding[:html] && @rules_files["html.md"],
+          "<!-- phoenix:html-end -->",
+          "<!-- phoenix:liveview-start -->",
+          project.binding[:live] && @rules_files["liveview.md"],
+          "<!-- phoenix:liveview-end -->",
```
josevalim, line 144: What do you think about this?

```suggestion
          project.binding[:live] && [
            "<!-- phoenix:liveview-start -->",
            @rules_files["liveview.md"],
            "<!-- phoenix:liveview-end -->"
          ],
```

And then you call `List.flatten`?

installer/mix.exs
```diff
@@ -35,7 +36,7 @@ defmodule Phx.New.MixProject do
         ],
         licenses: ["MIT"],
         links: %{"GitHub" => @scm_url},
-        files: ~w(lib templates mix.exs README.md)
+        files: ~w(lib templates phoenix_usage_rules mix.exs README.md)
```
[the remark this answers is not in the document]
SteffenDE: Good point, I moved them in https://github.com/phoenixframework/phoenix/pull/6385/commits/aa5e86b0cae945cd1fbf57d48a7efd2a1c6d97fb

PR #6386, opened 2025-07-30
Fix typos

obieewan:
- Fixes typos
- change re-iterate to reiterate 
- change referenced table from `:user` to `:users` to align with  `phx.gen.auth` default

### DIFF
--- a/guides/data_modelling/more_examples.md
+++ b/guides/data_modelling/more_examples.md
@@ -32,7 +32,7 @@ We generated an `Orders` context. The order is automatically scoped to the curre
     create table(:orders) do
 -     add :total_price, :decimal
 +     add :total_price, :decimal, precision: 15, scale: 6, null: false
-      add :user_id, references(:user, type: :id, on_delete: :delete_all)
+      add :user_id, references(:users, type: :id, on_delete: :delete_all)
 
       timestamps()
     end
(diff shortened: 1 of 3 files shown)


guides/live_view.md
```diff
@@ -120,6 +123,7 @@ LiveViews have several important lifecycle stages:
 
 - [`mount`](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#c:mount/3) - initializes the LiveView with parameters, session data, and socket
 - [`handle_params`](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#c:handle_params/3) - responds to URL changes and updates LiveView state accordingly
+- [`render`](https://hexdocs.pm/phoenix_live_view/Phoenix.LiveView.html#c:render/1) - renders the LiveView's HTML based on the current state
```
SteffenDE, line 126: After thinking about it, I would not mention render here. While it is technically part of the lifecycle, it is very different to the other stages mentioned here and people will read more about it in the LiveView documentation. It can also be omitted in favor of a html.heex file.

obieewan:
Thank you for the thoughtful feedback, @SteffenDE . I’ll make the necessary adjustments. 🔥

SteffenDE:
Thank you again! 🙌🏻

PR #6388, opened 2025-07-31
final changes for LiveView 1.1

SteffenDE:


chrismccord:
❤️❤️❤️🐥🔥

installer/templates/phx_single/mix.exs
```diff
@@ -50,7 +50,7 @@ defmodule <%= @app_module %>.MixProject do
       {<%= inspect @adapter_app %>, ">= 0.0.0"},<% end %><%= if @html do %>
       {:phoenix_html, "~> 4.1"},
       {:phoenix_live_reload, "~> 1.2", only: :dev},
-      {:phoenix_live_view, "~> 1.1.0-rc.0"},
+      {:phoenix_live_view, "~> 1.1.0"},
```
rhcarvalho, line 53:
```suggestion
      {:phoenix_live_view, "~> 1.1"},
```

I've been defaulting to `major.minor` in `mix.exs` for anything past 1.0, to allow upgrades to anything in the `1.x` series. Several other dependencies follow that pattern, would it apply to LiveView as well?

IIUC `1.1.0` is more restrictive, only allowing patch version updates in the `1.1.x` series. It is for sure more defensive if we expect upgrading to `1.2` to require manual intervention.

SteffenDE: It was a deliberate choice for now. We may adjust it in a future Phoenix release.

You are free to relax it in your application :)

rhcarvalho: :+1: safe choice! Thanks for the note.

Issue #6391, opened 2025-08-04
Overriding paths in tsconfig.json breaks Node module resolution


derrickreimer:
### Environment

* Elixir version (elixir -v): 1.16.3
* Phoenix version (mix deps): 1.8.0-rc.4
* Operating system: macOS

### Actual behavior

The [recently-introduced tsconfig.json](https://github.com/phoenixframework/phoenix/pull/6268/files#diff-ebb59150eb910be95f490a5b4c9d5d6b2ccc62c7137817506d3f6b9beba507d9R5) file seems to be breaking node module resolution. [This was discovered](https://github.com/inertiajs/inertia-phoenix/issues/59) when attempting to use the `@inertiajs/react` package in a clean Phoenix installation.

Here's an example repository: https://github.com/derrickreimer/phx-tsconfig-paths-repro

```
git checkout https://github.com/derrickreimer/phx-tsconfig-paths-repro.git
cd phx-tsconfig-paths-repro
cd assets && npm install
cd ../
mix setup
```

When running `mix assets.build`, the following warning appears in the output:

```
▲ [WARNING] Import "createInertiaApp" will always be undefined because the file "node_modules/@inertiajs/react/dist/index.js" has no exports [import-is-undefined]

    js/app.jsx:4:9:
      4 │ import { createInertiaApp } from "@inertiajs/react";
        ╵          ~~~~~~~~~~~~~~~~
```

Consequently, the built JavaScript errors when executed on the client:

```
09:19:04.274 Uncaught ReferenceError: module is not defined
    <anonymous> index.ts:1
index.ts:1:1
```

### Expected behavior

The build should succeed. Removing the `paths` override from the tsconfig.json file fixes the build. Though this override appears to be technically correct (since it still includes `node_modules` in the places to resolve), it breaks esbuild nonetheless.

Rather than overriding global path resolution (`"*"`), perhaps it would be safer to create phoenix-specific path alias like `"phx_deps/*"`?

```json
{
  "compilerOptions": {
    "baseUrl": ".",
    "paths": {
      "phx_deps/*": ["../deps/*"]
    },
    "allowJs": true,
    "noEmit": true
  },
  "include": ["js/**/*"]
}
```

PR #6392, opened 2025-08-04
don't include node_modules in tsconfig path mapping

SteffenDE:
Closes https://github.com/phoenixframework/phoenix/issues/6391.


SteffenDE:
I checked that LiveView types are still resolved correctly.

PR #6393, opened 2025-08-05
Optimize live_view stream by using stream_insert and stream_delete

obieewan:
I noticed the stream resets on every broadcast, causing full list reloads. This replaces it with `stream_insert `and `stream_delete` to make updates more efficient.

See attached screenshots for comparison:

Existing behaviour:
Screenshot 2025-08-05 at 14 36 30

Improved version:
Screenshot 2025-08-05 at 14 35 13


josevalim:
The more opinionated we become, the more likely it will backfire in the long term. For example, this can lead to racing updates, missing preloads, or lead to the struct having slightly different information that if it was queried on the fly. So my suggestion is to move this:

    <%= inspect context.alias %>.list_<%= schema.plural %>(<%= socket_scope %>)

to a private function and use it on mount and handle_info, this guarantees consistency across preloads and what not. I'd also only broadcast IDs, given that we don't actually use the whole struct.

obieewan:
Thanks for the clarification. That makes sense. I moved the
`<%= inspect context.alias %>.list_<%= schema.plural %>(<%= socket_scope %>)`
into a private function and now use it in both on_mount and handle_info to ensure consistency. I’m leaving the broadcast logic untouched since that comes from another generator.

josevalim (approved):
This looks good to me but if you could go ahead and change the other generator too, it would be very appreciated :)

obieewan:
> This looks good to me but if you could go ahead and change the other generator too, it would be very appreciated :)

Got it, thanks. I can apply the changes, though they might conflict with PR #6413. Should I wait for that one to be merged first, or go ahead and update?

josevalim:
Good point. Let's keep this one as is then and @SteffenDE can update the other if he wishes to. :)

obieewan:
Alright, thanks. 💜

priv/templates/phx.gen.live/index.ex
```diff
@@ -65,6 +65,10 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
   @impl true
   def handle_info({type, %<%= inspect schema.module %>{}}, socket)
       when type in [:created, :updated, :deleted] do
-    {:noreply, stream(socket, :<%= schema.collection %>, <%= inspect context.alias %>.list_<%= schema.plural %>(<%= socket_scope %>), reset: true)}
+    {:noreply, stream(socket, :<%= schema.collection %>, list_<%= schema.plural %>(<%= socket_scope %>), reset: true)}
+  end
+
+  defp list_<%= schema.plural %>(<%= scope && scope.assign_key %>) do
+    <%= inspect context.alias %>.list_<%= schema.plural %>(<%= scope && scope.assign_key %>)
   end<% end %>
```
SteffenDE, line 73:
```suggestion
    {:noreply, stream(socket, :<%= schema.collection %>, list_<%= schema.plural %>(<%= socket_scope %>), reset: true)}
  end<% end %>

  defp list_<%= schema.plural %>(<%= scope && scope.assign_key %>) do
    <%= inspect context.alias %>.list_<%= schema.plural %>(<%= scope && scope.assign_key %>)
  end
```

PR #6395, opened 2025-08-06
Fix route.ex typo in scopes.md

m9tdev:
Noticed a small typo in the Scopes docs where it refers to route.ex instead of router.ex.


SteffenDE:
🙌🏻

PR #6396, opened 2025-08-06
Add note on compile time config changes

JonRowe:
:wave: I ran into an issue with the upgrade process to 1.8.0 where an application that previously tried to set (incorrectly) `force_ssl: true` in `config/runtime.exs` would crash on boot, I am assuming this is related to the changes to the config where its now all marked as compile time which is highlighting the mistake, but it would be nice if the changelog tied this reference / warned you of this possibility.

I've suggested an additional sentence but obviously I'm happy for it to be tweaked as desired.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,7 +27,7 @@ This release introduces deprecation warnings for several features that have been
 
 ## Potential breaking changes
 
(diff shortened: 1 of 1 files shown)


SteffenDE:
Thank you! 🙌🏻

PR #6398, opened 2025-08-06
Class attribute type

SerhiiFesenko:
This pull request broadens the accepted type for the class attribute to better reflect its actual usage and improve component flexibility.

Currently, the type is too restrictive, which leads to unnecessary warnings for perfectly valid cases like:

```elixir
<.icon name="hero-x-mark" class={[@condition && "foo", "bar"]} />
```

The tag engine supports class values as:
- strings
- lists of strings, false, or nil
- nested lists of these values

By widening the type, we align the type spec with real behavior, eliminate false warnings, and allow more flexible component usage.


josevalim (approved):

SteffenDE:
🙌🏻

PR #6402, opened 2025-08-07
added conn to check_origin's list of arguments

kayuapi:


SteffenDE:
Hey @kayuapi,

we cannot change the arguments we pass to the function, as this would be a breaking change. I described a workaround in https://elixirforum.com/t/add-second-argument-plug-conn-to-phoenix-endpoint-socket-3-check-origin-mfa/69264/2.

kayuapi:
ok, thank you

PR #6407, opened 2025-08-11
Run mix compile before assets.build in generated aliases

SteffenDE:
Otherwise, it won't find colocated hooks on `mix setup`.

The other approach would be to change `assets.build` itself to include `"compile"`.


josevalim:
@SteffenDE should we change assets.build instead to call "compile"?

SteffenDE:
That was my question as well :D

> The other approach would be to change assets.build itself to include "compile".

If you also think that, then let's do it that way!

josevalim:
Yeah, I think that’s the best approach, otherwise if you call it in isolation, it would still be broken

PR #6412, opened 2025-08-12
mix phx.gen.auth: add "Delete Account" button to  generated settings page

nbw:
The Settings page generated with `mix phx.gen auth` creates a form to update a user's email and another to set a new password. I've been using the generator with a few new apps and I noticed that I had to implement delete account in each case. It _feels_ like this should come out of the box with the generator. Refer to the red "Delete Account" button in the screenshot below.

👋 Note: I've implemented it for the liveview auth flow. I'd be willing to add it to the static template/controller based flow too if the response to this PR is positive. 

## Screenshots

Screenshot 2025-08-12 at 9 16 31

## Demo

https://github.com/user-attachments/assets/5fad1fb9-ccec-430e-800d-245d3d5bb703

## Local testing instructions

1. Pull phoenix project somewhere locally
2. create a new phoenix project
3. update the phoenix dependency (mix.exs) to point to the local copy of phoenix: 
```elixir
# relative path to the local copy of phoenix
{:phoenix, path: "../phoenix", override: true},
```
4. Run `mix phx.gen.auth Accounts User users`

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -91,11 +91,15 @@ defmodule <%= @web_namespace %>.CoreComponents do
   """
   attr :rest, :global, include: ~w(href navigate patch method download name value disabled)
   attr :class, :string
-  attr :variant, :string, values: ~w(primary)
+  attr :variant, :string, values: ~w(primary error)
   slot :inner_block, required: true
 
   def button(%{rest: rest} = assigns) do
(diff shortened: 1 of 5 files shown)


josevalim:
Account deletion might be trickier than one might expect. For example, should you immediately delete all resources owner by that user? If so, certain users can actually block the database for long. Does the industry you are at require some sort of log/data retention? There are so many questions that it is best for users to build on top taking into consideration their domain. Thank you!

josevalim:
On the plus side, if deleting your account in your app is as easy as this PR, then you can add it on your own app :)

PR #6414, opened 2025-08-13
Fix comments in installer config

vhf:
The installer creates config files that mostly use the imperative mood but are oddly inconsistent:
`installer/templates/phx_single/config/config.exs`:
> \# **Configures** the mailer
> \# Configure esbuild (the version is required)
> \# Configure tailwind (the version is required)
> \# **Configures** Elixir's Logger

`installer/templates/phx_single/config/prod.exs`:
> \# **Configures** Swoosh API Client
> \# Disable Swoosh Local Memory Storage
> \# Do not print debug messages in production

`installer/templates/phx_single/config/dev.exs`:
> \# Watch static and templates
> \# Enable dev routes
> \# Do not include metadata
> \# Set a higher stacktrace
> \# Initialize plugs at runtime
> \# Disable swoosh api client

I suggest sticking to imperative mood for consistency


josevalim:

PR #6416, opened 2025-08-14
Update controllers documentation

elfenlaid:
The changes update the [Setting the Content Type](https://hexdocs.pm/phoenix/1.8.0/controllers.html#setting-the-content-type) section with the up-to-date way of setting the content type from within controllers

Additionally to `put_resp_content_type`, the changes add `put_format`. According to [`Phoenix.Controller.render`](https://github.com/phoenixframework/phoenix/blob/008d5ff9e61daddf5a8620b1c4ad22470d6137f8/lib/phoenix/controller.ex#L949) implementation, `put_format` is the only way to set the corresponding view from within controllers

The changes mention that aside from adding a template file, one should also add a `PageXML` view to embed the template

Also, the note at the bottom double-checks that one should use `.eex` instead of `heex` to render XML. Otherwise, the connection pipeline crashes with a rather cryptic error while trying to fit `Phoenix.LiveView.Rendered` in the `Plug.Conn.resp` which expects a binary

---

Please note that my changes might not align with the intended way of setting the content type from within Phoenix controllers. It might not be as easy as adding the `put_format` call. They just make the documentation example work, but in a possibly perverted way 😅

### DIFF
--- a/guides/controllers.md
+++ b/guides/controllers.md
@@ -262,11 +262,14 @@ If we wanted to render an XML version of our `home` action, we might implement t
 def home(conn, _params) do
   conn
   |> put_resp_content_type("text/xml")
+  |> put_format(:xml)
   |> render(:home, content: some_xml_content)
 end
 ```
 
-We would then need to provide an `home.xml.eex` template which created valid XML, and we would be done.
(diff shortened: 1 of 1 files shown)


SteffenDE (approved):

josevalim:

PR #6417, opened 2025-08-14
Fix controller.ex documentation typo

elfenlaid:
The changes fix the documentation example. The example doesn't pipe the updated `conn` to the `render` function.


SteffenDE:
Thank you! 🙌🏻

PR #6418, opened 2025-08-14
Remove reference to OTP version affected by CVE-2025-32433

arcanemachine:
The example Dockerfile in the "Deploying with Releases" guide references OTP version 25.3.2.15, which is affected by
[CVE-2025-32433](https://github.com/erlang/otp/security/advisories/GHSA-37cp-fgq5-7wc2).

This is lower than the minimum recommended version of OTP-25.3.2.20.

This pull request changes the values in the example Docker to the use a newer patched version of the same major version of OTP, which references an existing image on Docker Hub.


SteffenDE:
As we’re already touching the file, let’s choose 1.18.4 with the latest OTP 27 and Debian?

`hexpm/elixir:1.18.4-erlang-27.3.4.2-debian-bookworm-20250811`

arcanemachine:
OK done. Only change from your suggestion was that I used the 'slim' Debian image, as in the existing Dockerfiles.

SteffenDE:
Perfect! Thank you :)

PR #6429, opened 2025-08-19
update scope documentation and config examples

emanoelims:
Changes

- Updated route prefix from /orgs/:org → /organizations/:org
- Added route_access_path example using slug, ensuring generators create LiveView routes correctly with slugs instead of IDs
- Fixed unused user variable in test helper

Rationale

The route prefix was updated to /organizations for consistency with Phoenix generators. This ensures alignment with generated code and prevents discrepancies that could confuse new developers.
Additionally, the route_access_path example was adjusted to use slugs, which is required for generators to build correct LiveView routes without falling back to IDs.


SteffenDE:
🙌🏻

PR #6435, opened 2025-08-22
Enable `:force_ssl` in `:prod` environment by default

Gladear:
This is a proposal to enable the `:force_ssl` option and HSTS by default on new installations.

HSTS is a standard nowadays, recommend for most, if not all, websites or webapps. However,
for I feel like it can be easy to miss if you are not really into web security - just like
CSP, so thanks for making it a default now ❤️

The given option may not be the best suited for a default installation. I basically
copy-pasted the "Using SSL" part of the documentation, making it as generic as possible.

### DIFF
--- a/installer/templates/phx_single/config/prod.exs
+++ b/installer/templates/phx_single/config/prod.exs
@@ -10,7 +10,7 @@ config :<%= @web_app_name %>, <%= @endpoint_module %>, cache_static_manifest: "p
 
 <% end %># Force using SSL in production. This also sets the "strict-security-transport" header,
 # also known as HSTS. `:force_ssl` is required to be set at compile-time.
-config :<%= @web_app_name %>, <%= @endpoint_module %>, force_ssl: [rewrite_on: [:x_forwarded_proto], host: nil]<%= if @mailer do %>
+config :<%= @web_app_name %>, <%= @endpoint_module %>, force_ssl: [rewrite_on: [:x_forwarded_proto]]<%= if @mailer do %>
 
 # Configures Swoosh API Client
 config :swoosh, api_client: Swoosh.ApiClient.Req
--- a/installer/templates/phx_umbrella/apps/app_name_web/config/prod.exs
+++ b/installer/templates/phx_umbrella/apps/app_name_web/config/prod.exs
@@ -11,4 +11,4 @@ config :<%= @web_app_name %>, <%= @endpoint_module %>,
 
 # Force using SSL in production. This also sets the "strict-security-transport" header,
 # also known as HSTS. `:force_ssl` is required to be set at compile-time.
-config :<%= @web_app_name %>, <%= @endpoint_module %>, force_ssl: [rewrite_on: [:x_forwarded_proto], host: nil]
+config :<%= @web_app_name %>, <%= @endpoint_module %>, force_ssl: [rewrite_on: [:x_forwarded_proto]]


josevalim:
If i remember correct, the reason why this is disabled is because, if you run prod on your machine (which you may do the first time for testing) and then you access localhost+HSTS, it borks your localhost unless you go to specific settings in each browser and revert that. So your development environments breaks in very non obvious ways.

Gladear:
When accessing the app through `http://localhost:4000`, there is no issue, `:force_ssl` does not apply headers on localhost.
When accessing it through `http://127.0.0.1:4000`, the UA is redirected to HTTPS and a "Your connection is not private" appears. The logs clearly state `[info] Plug.SSL is redirecting GET / to https://127.0.0.1 with status 301`.
Should you access it through a custom domain that you defined in `/etc/hosts` there would probably be an issue (I did not test), but I think people doing this are both pretty rare, and hopefully proficient enough to debug the problem 😄

josevalim:
Good! And IIRC, the other concern is related to proxy headers. You are currently setting `host: nil`, which means we will use the same host as the request. However, if you are beyind a proxy, the host is likely `localhost` unless you rewrite it. We could add `x_forwarded_host` but then, if you are not behind a proxy (or at least one that uses a different header), then an attacker could inject a host and perform spoofing attacks. 😅 

Another option is to not set `host: nil` but then it means we will fail to redirect subdomains, but that seems to be the only default that is safe.

Gladear:
Indeed ! I removed `host: nil`. Although I can't know for sure, I suppose using multiple domains on a single Phoenix endpoint isn't a common use case.

andrewtimberlake:
> I suppose using multiple domains on a single Phoenix endpoint isn't a common use case.

I think it might be. I for one use it on a number of projects.
The `:host` option is part of `scope/4` allowing you to match on the host in the router.

josevalim:
To be clear, it means that accessing `http://foo.example.com/path` will redirect to `https://example.com/path`, until the proper `rewrite_on` is configured. The main host will always work. That seems an acceptable trade-off for the improved security and is better than `host: nil` which will redirect to localhost on _all cases_ when not fully configured.

Gladear:
Another option could be to add it as a comment, explain that it's a good practice to activate it, and link to the [Force SSL](https://hexdocs.pm/phoenix/using_ssl.html#force-ssl) section of the documentation so that people can configure it as they need it.

SteffenDE:
🙌🏻

conradfr:
If I understand the change correctly I think that should appear somewhere in the docs for people (like me with CapRover) who run their apps in containers where the public endpoints (and thus the ssl certificates) are handled by a proxy/gateway/whatever so ssl is not used for bandit in prod.

SteffenDE:
@conradfr if your proxy properly sets `X-Forwarded-Proto` all should be good, right? A PR for the docs is welcome.

Gladear:
@conradfr You still can use this configuration if your SSL is handled by a proxy. If your proxy allows HTTP connections to your app, then the app will respond either with a redirect, or it will add [HSTS](https://developer.mozilla.org/en-US/docs/Web/HTTP/Reference/Headers/Strict-Transport-Security) headers to requests. And if it does not, well, it doesn't hurt to have a second layer of security 😄 

As far as I understand, using `:force_ssl` simply adds [`Plug.SSL`](https://hexdocs.pm/plug/Plug.SSL.html) to the Endpoint, so maybe you can learn a little more about what it does there!

eriknaslund:
I just wanted to drop a note on an issue I experienced after (trying to) add this as part of a Phoenix upgrade.

Sorry for the AWS specific lingo, but it should still be applicable to multiple platforms.

My application runs on AWS ECS Fargate with an AWS ALB (load balancer) in front which calls `/health` as a health check. The load balancer takes care of SSL termination, and basically has these two rules set up:

```
:80     Redirect to HTTPS://#{host}:443/#{path}?#{query}
        Status code: HTTP_301
                
:443    Forward to target group ls-lb-9bd7b4e
```

This is done so that I can let AWS manage my TLS certificates automatically using ACM (Certificate Manager).
I believe it's a pretty common setup to do SSL termination at the LB level in :prod, just like I do.

Adding `config :app, AppWeb.Endpoint, force_ssl: [rewrite_on: [:x_forwarded_proto]]` in :prod _by default_ feels a bit "risky" to me, since it makes the health check fail, and it's not obvious why at first. The reason it fails is that the health check request will be coming from the load balancer, so it doesn't matter that "localhost" is allowlisted.

What do you think about at least adding a comment to the generated code that you "might not want this depending on your setup", and documenting why a bit more under "Using SSL -> Force SSL" in the docs?

josevalim:
Better docs will always be beneficial, so PRs are welcome, but we should probably also have `exclude: ["localhost"]` explicitly listed in and add a note that you probably want to add the load balancer too? Do you have a fixed host/origin for your load balancer?

eriknaslund:
I'm afraid I don't. Amazon ALBs are a bit tricky that way because they scale up dynamically under the hood as your traffic increases (https://aws.amazon.com/blogs/networking-and-content-delivery/scaling-strategies-for-elastic-load-balancing/). This means the origin IP(s) aren't static over time, and will change as traffic scales up (or when AWS decides to replace the ALB node for you).

josevalim:
@eriknaslund so we should probably have a new feature on Plug.SSL which is to skip it for certain paths. Would you like to send a PR for that?

eriknaslund:
That would be an excellent addition, good thinking. I'd be happy to give it a go as long as it's OK if it takes a couple of weeks? Life's a bit busy right now, so I don't wanna over-promise.

josevalim:
@eriknaslund given your comment above, I assuming people will need it sooner than later, so if you don't mind I'd take a stab at it. Btw, instead of doing it by path, perhaps the load balancer/health checker also sets a header to uniquely identify it? A header it doesn't let others pass through?

eriknaslund:
@josevalim That makes a lot of sense, go for it!

Since we're talking about the health checks here I'm thinking that it's not really a request that passes through as such. Any well behaved client (including the health checker) should send the user-agent header, right? Maybe that could be something to filter/exclude on that would work well in a lot of cases.

For AWS in particular I'm pretty certain it'd work because they set the user agent to "ELB-HealthChecker/2.0" (https://repost.aws/knowledge-center/elb-fix-failing-health-checks-alb).

What do you think?

josevalim:
@eriknaslund ah, great. Although I guess someone could pass that through `curl` and disable the force SSL, so perhaps using a path, something under our control, would be better.

eriknaslund:
Yeah, I thought about that...you're absolutely right, but why would they?

The whole point of the force_ssl and the HSTS header is really just to try our hardest to tell the browser to use HTTPS the next time the load the site (for a more secure experience). As far as I'm concerned it's probably best to not even serve a site over non-secure HTTP if you _really_ want to forbid people from accessing it that way.

But then the name comes into question...it wouldn't really be "force" SSL anymore now would it?

I'm open to the fact that I might be thinking about this wrong, and that I'm missing some important detail, as I often am :).

eriknaslund:
But if we want to keep the changes small your initial suggestion of excluding a certain path would probably work really well too.

josevalim:
Plug.SSL is out with the ability to also exclude paths.

josevalim:
https://github.com/phoenixframework/phoenix/pull/6556

PR #6443, opened 2025-08-26
update guide assets for 1.8

noel-gallagher:
updates guide to reflect 1.8 changes


SteffenDE:
@noel-gallagher I'd propose to zoom in a little bit when making those screenshots, otherwise it's hard to decipher anything when seeing them in the docs

image

Also, I was thinking that if we already update those, we could also do the same trick as the Phoenix README.md and include both light and dark versions:

```html
<picture>
  <source media="(prefers-color-scheme: dark)" srcset="assets/images/hello-world-from-frank-dark.png" />
  <source media="(prefers-color-scheme: light)" srcset="assets/images/hello-world-from-frank-light.png" />
  <img src="assets/images/hello-world-from-light.png" alt="Frank Greets Us from Phoenix" />
</picture>
```

Then (at least when using the system theme) we adhere to the dark/light mode preference. Wdyt?

SteffenDE:
Hey @noel-gallagher, do you still want to update this?

PR #6444, opened 2025-08-27
Update data modelling guide and ecto guide

obieewan:
This PR updates the data modeling guide to align with the current Phoenix generator default and replaces deprecated usage of `Repo.transaction/2`.

- replace `Repo.transaction/2` with `Repo.transact/2`
- update `timestamps()` to `timestamps(type: :utc_datetime)`


josevalim:
Thank you! Ideally we would no longer use multis, now that we have Repo.transaction. So if you want to send another PR for that too, it would be very welcome!

josevalim:

obieewan:
> Thank you! Ideally we would no longer use multis, now that we have Repo.transaction. So if you want to send another PR for that too, it would be very welcome!

Sure, I’d gladly handle it in a separate PR. 💜

PR #6448, opened 2025-08-31
Remove reference to `{:system, env_var}` from docs

rhcarvalho:
It has been deprecated since #4709.


josevalim:

Issue #6073, opened 2025-02-05
Umbrella generator writes two times `import Config`


francois08:
### Environment

* Elixir version (elixir -v):
Erlang/OTP 27 [erts-15.1.2] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit]
Elixir 1.18.2 (compiled with Erlang/OTP 27)

* Operating system:
macOS

```console
% mix phx.new -v                                                                                                                            
Phoenix installer v1.7.19
```

```console
% mix phx.new foo --binary-id --no-esbuild --no-dashboard --no-assets --no-tailwind --no-gettext --no-live --no-install --no-html --umbrella
```

This writes two times `import Config`.
Result:

```elixir
import Config

# config/runtime.exs is executed for all environments, including
# during releases. It is executed after compilation and before the
# system starts, so it is typically used to load production configuration
# and secrets from environment variables or elsewhere. Do not define
# any compile-time configuration in here, as it won't be applied.
# The block below contains prod specific runtime configuration.
if config_env() == :prod do
  database_url =
    System.get_env("DATABASE_URL") ||
      raise """
      environment variable DATABASE_URL is missing.
      For example: ecto://USER:PASS@HOST/DATABASE
      """

  maybe_ipv6 = if System.get_env("ECTO_IPV6") in ~w(true 1), do: [:inet6], else: []

  config :foo, Foo.Repo,
    # ssl: true,
    url: database_url,
    pool_size: String.to_integer(System.get_env("POOL_SIZE") || "10"),
    socket_options: maybe_ipv6

  import Config

  # The secret key base is used to sign/encrypt cookies and other secrets.
  # A default value is used in config/dev.exs and config/test.exs but you
  # want to use a different value for prod and you most likely don't want
  # to check this value into version control, so we use an environment
  # variable instead.
  secret_key_base =
    System.get_env("SECRET_KEY_BASE") ||
      raise """
      environment variable SECRET_KEY_BASE is missing.
      You can generate one by calling: mix phx.gen.secret
      """

  config :foo_web, FooWeb.Endpoint,
    http: [
      # Enable IPv6 and bind on all interfaces.
      # Set it to  {0, 0, 0, 0, 0, 0, 0, 1} for local network only access.
      ip: {0, 0, 0, 0, 0, 0, 0, 0},
      port: String.to_integer(System.get_env("PORT") || "4000")
    ],
    secret_key_base: secret_key_base

  # ## Using releases
[...]

PR #6449, opened 2025-08-31
Support PORT env var in dev using `runtime.exs`

rhcarvalho:
While upgrading an app from v1.7.21 to v1.8.1, I noticed changes to `config/dev.exs` in new projects which depended on the system environment at build time.

This commit updates the support for the `PORT` environment variable in new projects generated with `phx.new` to be done in `runtime.exs` instead of `dev.exs`, which is meant for static build time config.

Dev support for PORT was added a few months ago in https://github.com/phoenixframework/phoenix/commit/7a88d0ff93c5036836c6ae523b67bef0619cfd0e, and shipped first in v1.8.0-rc.1.


installer/templates/phx_umbrella/apps/app_name_web/config/runtime.exs
```diff
@@ -15,7 +15,7 @@ config :<%= @web_app_name %>, <%= @endpoint_module %>,
     # Enable IPv6 and bind on all interfaces.
     # Set it to  {0, 0, 0, 0, 0, 0, 0, 1} for local network only access.
     ip: {0, 0, 0, 0, 0, 0, 0, 0},
-    port: String.to_integer(System.get_env("PORT") || "4000")
+    port: port
```
rhcarvalho, line 18: I noticed this file gets injected into `umbrella/config/runtime.exs`. Unlike `phx_single` the template files are not 1:1 with the final config files in a generated umbrella project.

This has led to something like #6073, even though I saw code that surgically remove "import Config" in some places :)

That's to say I hope this change is okay, as it relies on `port` being defined elsewhere.

installer/test/phx_new_test.exs
```diff
@@ -31,6 +31,7 @@ defmodule Mix.Tasks.Phx.NewTest do
       Mix.Tasks.Phx.New.run([@app_name])
 
       assert_file("phx_blog/README.md")
+
```
rhcarvalho, line 34: Added by `mix format`.

rhcarvalho:
Hmm the [failed test](https://github.com/phoenixframework/phoenix/actions/runs/17360491955/job/49279907580?pr=6449) passes for me locally:

```console
$ mix test test/phoenix/integration/endpoint_test.exs:126
Compiling 74 files (.ex)
Generated phoenix app
Running ExUnit with seed: 996645, max_cases: 16
Excluding tags: [:test]
Including tags: [location: {"test/phoenix/integration/endpoint_test.exs", 126}]

.
Finished in 1.1 seconds (0.00s async, 1.1s sync)
4 tests, 0 failures, 3 excluded
```

SteffenDE:
@rhcarvalho I did not experience any issues with having it in the build-time config for dev, did you?

rhcarvalho:
> @rhcarvalho I did not experience any issues with having it in the build-time config for dev, did you?

No particular issue. Maybe it's my limited knowledge :)

My first instinct when seeing the new code in `config/dev.exs` was "I learned somewhere I should not use `System.get_env/1` in "compile-time config files", but rather in `runtime.exs`.

When Mix is available, `PORT=5001 mix phx.server` works fine. I think the problem arises in Releases, and why runtime config, including reading from env vars, is written in `runtime.exs` and not `prod.exs`.

I don't remember where exactly I got my "lesson" from, but found a few related threads in the forum:

- https://elixirforum.com/t/config-exs-and-system-get-env-2-are-values-actually-read-at-runtime/56481
- https://elixirforum.com/t/how-to-use-system-get-env-inside-my-config-exs-file/67883
- https://elixirforum.com/t/warnings-about-application-get-env-2-when-using-environment-variables-for-configuration/56567

And I was about to close this PR until I found two interesting historical commits:

1. In the past, 2014, `dev.exs` and `config.exs`: https://github.com/phoenixframework/phoenix/commit/e8d9b45c32989377726da21075f85f1f3866a837

```console
$ git show e8d9b45c32989377726da21075f85f1f3866a837:priv/template/config/dev.exs
use Mix.Config

  debug_errors: true,
  cache_static_lookup: false

# Enables code reloading for development
config :phoenix, :code_reloader, true
$ git show e8d9b45c32989377726da21075f85f1f3866a837
commit e8d9b45c32989377726da21075f85f1f3866a837
Author: José Valim <jose.valim@plataformatec.com.br>
Date:   Sat Dec 13 14:25:38 2014 +0100

    Remove repeated HTTP config

diff --git a/priv/template/config/config.exs b/priv/template/config/config.exs
index 3932e08cd..eec091a6d 100644
--- a/priv/template/config/config.exs
+++ b/priv/template/config/config.exs
@@ -8,7 +8,6 @@ use Mix.Config
 # Configures the endpoint
   url: [host: "localhost"],
-  http: [port: System.get_env("PORT")],
   secret_key_base: "<%= secret_key_base %>",
   debug_errors: false
```

Note that the `Endpoint` `:port` config used to be exactly what we have now, either `PORT` env or 4000. But...

2. Later in 2015, still quite long ago: https://github.com/phoenixframework/phoenix/commit/feef6076094104e45660637e05629f5869d6d836

Quoting José's commit message:

> Do not use System.get_env/1 in config files
They do not work with releases, so let's not teach
something that is going to break during deployment.

And then `dev.exs` became the static config as we've had until recently, hard coded to port 4000.

My intuition and motivation for the change matched exactly what José wrote back then.

---

I'm guessing Chris reintroduced the configurable port in dev for something like `phoenix.new` or other initiative. Myself, I haven't needed the customization anyway, as most of the time I'm running in a container and can map any local port to port 4000 on the container.

@SteffenDE what do you think, shall we drop this? Or is it still "teaching the right patterns" and worth doing again nowadays?

SteffenDE:
Yeah, I think @josevalim was / is right about not teaching about `System.get_env` in config files other than runtime.exs. I'd just write it as

```elixir
if port = System.get_env("PORT") do
  config :my_app, MyAppWeb.Endpoint, http: [port: String.to_integer(port)]
end
```

and not check the config_env at all for the port.

rhcarvalho:
> I'd just write it as [...] and not check the config_env at all for the port.

Good call, makes it work the same regardless of env. We'd be potentially affecting the standard `:test` env, do you think that'd be a problem? AFAIK we don't start any HTTP servers by default in tests, right?

I can update the PR.

rhcarvalho:
> Yeah, I think @josevalim was / is right about not teaching about `System.get_env` in config files other than runtime.exs. I'd just write it as
> 
> ```elixir
> if port = System.get_env("PORT") do
>   config :my_app, MyAppWeb.Endpoint, http: [port: String.to_integer(port)]
> end
> ```
> 
> and not check the config_env at all for the port.

While I edit the code, I see one downside to globally setting the port: endpoint config becomes spread out, making the setting of IP and port far apart in the config file, which can cause friction for humans trying to update the config or reason about what is set and why. This concern might not be real.

rhcarvalho:
Rebased on top of latest main and applied your suggestion @SteffenDE, indeed applying the same config to all environments makes the overall config less repetitive.

rhcarvalho:
https://github.com/phoenixframework/phoenix/actions/runs/17941936145/job/51020424195?pr=6449

Test error seems to be due to a Docker Hub timeout. ~~Looks like other PRs are running into the same issue.~~

SteffenDE:
yeah Docker Hub isn't too reliable...

josevalim:

PR #6452, opened 2025-09-01
Update contrib guide on how to generate docs

rhcarvalho:
The `assets/package.json` has been removed in https://github.com/phoenixframework/phoenix/commit/484245cbaba8eeb83197af4b6696d1b752f8a8b2 (https://github.com/phoenixframework/phoenix/pull/6044) in favor of a single `package.json` in the project root.

I noticed this while trying to understand the changes between v1.7.21 and v1.8.1, in particular investigating why https://github.com/phoenixframework/phoenix/pull/6321 was needed.

PR #6453, opened 2025-09-01
Update package-lock.json to reflect changes to package.json

rhcarvalho:
Normally it gets updated when releases are cut, but it seems the last one missed it.

Noticed while building docs locally for #6452.


SteffenDE:
🙌🏻

PR #6454, opened 2025-09-01
Update Tailwind to version 4.1.12 to prevent some unpredictable behaviors in utility classes.

shahryarjb:
In Tailwind 4.1.7 when you use some class like `space-y-10` , you can see in your browser it is created like

```css
.space-y-10 {}
```

photo_2025-09-01 12 55 20

After updating to last version it is fixed!

> I tested it in Phoenix 1.8.1 with `mix phx.new ...` in the last version of windows and mac os; I should be noted I did not test the Phoenix source directly

### DIFF
--- a/installer/templates/phx_single/config/config.exs
+++ b/installer/templates/phx_single/config/config.exs
@@ -45,7 +45,7 @@ config :esbuild,
 
 # Configure tailwind (the version is required)
 config :tailwind,
-  version: "4.1.7",
+  version: "4.1.12",
   <%= @app_name %>: [
     args: ~w(
       --input=assets/css/app.css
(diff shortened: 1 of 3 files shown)


SteffenDE:
@shahryarjb what was the reason to close? In general it's fine for us to bump the Tailwind version for new projects :)

shahryarjb:
Hi dear @SteffenDE ,
I realized that I need to update Tailwind (but it is not the main problem), and it seems that in the new Tailwind v4, some priorities have changed in classes loading. Similar to version 3, there might be cases where certain classes shouldn’t be used together. That’s why I thought maybe it would be better to close it for now.

But overall, I think updating to the new version is a good idea, it solves some of the issues we’re currently facing.
Thanks a lot! I've reopened it now.

josevalim:

PR #6457, opened 2025-09-02
[phx gen.auth] test on page content not page nav content

nduitz:
This test currently asserts on the "Log in" text rendered in the root layout's default navigation.

IMHO this should assert on actual page content instead of nav bar content.

Maybe it was meant to assert on this "Log in":
https://github.com/phoenixframework/phoenix/blob/db21e45e558f5bb8774e552f9afcae147cae680b/priv/templates/phx.gen.auth/confirmation_live.ex#L50

But that would require the user to be already logged in, not just to be confirmed and would require a different test that I tried to implement.
However the problem with that test is that the assertion is also satisfied by the nav bar content.


josevalim:
Yes, I like the button assertions!

SteffenDE:
I'll follow this up with another PR!

PR #6458, opened 2025-09-02
Fix missing () in error message's suggestion

MaxLap:
A minor typo: `send(self, :after_join)` -> `send(self(), :after_join)`


SteffenDE:
🙌🏻

PR #6463, opened 2025-09-05
Context Guides: Fix and Enhance Code Snippets

thisistonydang:
I was going through the new context guides and noticed a couple of code snippet issues:

- In **"4. Cross-context Boundaries"**: `broadcast` should be `broadcast_cart` to match the code from the generators.  
- Similarly in **"5. Bringing It Home"**: `broadcast` should be `broadcast_order`.

Additionally, the `show.html.heex` templates aren’t wrapped in the layout component (e.g., `<Layouts.app flash={@flash}>...</Layouts.app>`). This isn’t an error, but wrapping them makes the templates feel more consistent with the rest of the app :)


SteffenDE:
Thank you! 🙌🏻

PR #6474, opened 2025-09-10
Update checkbox input

sa-mm:
The use case is to support checkboxes outside the <.form /> element. One needs the `form` attribute on the shadow input. Otherwise, you end up with inconsistent behavior.

```elixir
<.input type="checkbox" form={@form.id} field={@form[:active]} />
<.form for={@form} id={@form.id}></.form>
```

We've been using this in production for a few months without issue. I'm happy to make adjustments if there are concerns about passing all of `rest` to that input.

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -188,7 +188,7 @@ defmodule <%= @web_namespace %>.CoreComponents do
     ~H"""
     <div class="fieldset mb-2">
       <label>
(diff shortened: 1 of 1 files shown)


SteffenDE:
Let's only pass `form={@rest[:form]` instead of all of `@rest`?

sa-mm:
> Let's only pass `form={@rest[:form]` instead of all of `@rest`?

Sounds good. I pushed up the change.

chrismccord:
❤️❤️❤️🐥🔥

PR #6475, opened 2025-09-11
[RFC] Conditional channel event intercepts

studzien:
## Summary

This PR introduces conditional intercepts to Phoenix Channels, allowing intercepts to be applied selectively based on socket-specific predicates rather than intercepting events for all subscribers.
It currently lacks documentation updates. When this or a similar approach is accepted, the documentation will be updated.

## Problem

Currently, Phoenix Channel intercepts are binary - either all subscribers on a topic get intercepted messages, or none do. This creates issues when you need to send different versions of the same event to different subsets of subscribers based on their connection state, permissions, or other socket-specific conditions.

## Solution

Extends the intercept macro to accept predicate functions alongside event names:
```elixir
  # Before: intercept all "event" messages for everyone
  intercept ["event"]

  # New: conditionally intercept based on socket state
  intercept ["event", {"conditional_event", &MyChannel.should_intercept?/1}]

  def should_intercept?(socket) do
    socket.assigns[:special_user]
  end
```

 ## Changes Made

- `Phoenix.Channel`: Modified intercept macro to accept `{event, predicate}` tuples alongside plain
  event names
- `Phoenix.Channel.Server`: Updated fastlane mechanism to evaluate predicates when deciding whether to
   intercept
- Tests: Added test coverage for conditional intercept behavior

### DIFF
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -521,14 +521,18 @@ defmodule Phoenix.Channel do
   """
   defmacro intercept(events) do
     quote do
-      @phoenix_intercepts unquote(events)
+      @phoenix_intercepts Map.new(unquote(events),
+        fn {event, predicate} -> {event, predicate}
+           event -> {event, true}
+      end)
     end
   end
 
   @doc false
   def __on_definition__(env, :def, :handle_out, [event, _payload, _socket], _, _)
       when is_binary(event) do
-    unless event in Module.get_attribute(env.module, :phoenix_intercepts) do
+    intercepts = Module.get_attribute(env.module, :phoenix_intercepts)
+    unless event in Map.keys(intercepts) do
       IO.write(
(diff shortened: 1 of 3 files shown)


chrismccord:
I am 👎  on this because it copies the socket into the fastlane lookup, but that socket is _not_ the socket at the point of intercept, so it's usefulness for intercept logic is limited and has its own pitfalls that users won't immediately be aware of it (that the state is not the current socket state). Thanks!

studzien:
Thanks for taking a look. This makes sense.
Do you have some other mechanism in mind where it would be possible to customize the event payload for a subset of subscribers but use fastlane for everyone else?

chrismccord:
> Thanks for taking a look. This makes sense.
> Do you have some other mechanism in mind where it would be possible to customize the event payload for a subset of subscribers but use fastlane for everyone else?

separate broadcasts/topics, or intercept + push

Issue #6461, opened 2025-09-03
configure json_library to be `JSON` if defined


chrismccord:

saleyn:
I would love to have the `json_library` option configurable beyond `Jason` and `json`.  I am the maintainer of [glazer](https://glazer.hexdocs.pm), and believe it would be very beneficial for the Phoenix community to take advantage of more performant JSON parsing libraries when needed.

SteffenDE:
@saleyn You can configure any library you want as long as it has the same API: https://elixir.hexdocs.pm/JSON.html#functions

(And I think we don’t event use all those. You can grep the code for `json_library()`.)

You already have encode and decode, so you’re only missing the iodata variants. Of course you could add a wrapper module that just calls encode in that case, as a binary is iodata, but that might not be as efficient.

rhcarvalho:
@SteffenDE this was solved by #6481.

PR #6481, opened 2025-09-17
Use the built-in JSON library

elfenlaid:
Address https://github.com/phoenixframework/phoenix/issues/6461

The changes set the built-in [JSON library](https://hexdocs.pm/elixir/JSON.html) as a default JSON library.

The idea is to check whether the built-in JSON library is present and use it if so:

```elixir
  @default_json_library if Code.ensure_loaded?(JSON), do: JSON, else: Jason

  def json_library do
    Application.get_env(:phoenix, :json_library, @default_json_library)
  end
```

Correct me if I'm wrong, but I think a naive check of whether the library is present during compilation time should work, as it's a built-in library.

---

What I'm not sure about is whether such a move results in a breaking change. 

It could be a breaking change for projects that don't specify the `json_library` and use `@derive Jason.Encoder` on structures passed to `Phoenix.Controller.json/2`. That sounds pretty niche, especially given that the template project includes:

```elixir
config :phoenix, :json_library, Jason
```

Also, I'm specifically leaving the `jason` dependency in place for pre-27 OTP versions.

---

Please let me know what you think and whether the changes make sense. In other words, maybe it would be better to stick with `Jason` until Phoenix requires OTP 28.

---

Similar discussions / changes:

[Use Jason instead of Poison for json encoding by chrismccord #2734phoenixframework/phoenix](https://github.com/phoenixframework/phoenix/pull/2734)
[Investigate using Jason instead of Poison #2693phoenixframework/phoenix](https://github.com/phoenixframework/phoenix/issues/2693)

[Use JSON instead of Jason by kubosuke #234felt/geo](https://github.com/felt/geo/pull/234)


guides/controllers.md
```diff
@@ -94,7 +94,7 @@ end
 
 Now [`/hello/Frank`] in your browser should display `From messenger Frank` as plain text without any HTML.
 
-A step beyond this is rendering pure JSON with the [`json/2`] function. We need to pass it something that the [Jason library](`Jason`) can decode into JSON, such as a map. (Jason is one of Phoenix's dependencies.)
+A step beyond this is rendering pure JSON with the [`json/2`] function. We need to pass it something that the [JSON library](https://hexdocs.pm/elixir/JSON.html) can decode into JSON, such as a map.
```
elfenlaid, line 97: Here's how it looks like with mouse hover:

Image

config/config.exs
```diff
@@ -5,7 +5,7 @@ config :logger, :console,
   format: "\n$time $metadata[$level] $message\n"
 
 config :phoenix,
-  json_library: Jason,
```
elfenlaid, line 8: Probably here we would need to have something similar to `if Code.ensure_loaded?(JSON), do: JSON, else: Jason` if that conditional loading trick makes sense.

SteffenDE:
Looks good to me! But if we actually change new projects to use JSON, we'll also need to bump the minimum version we set to Elixir 1.18, see https://github.com/phoenixframework/phoenix/blob/9fc21c664eb1ba69b9c01d046b47a728f4a8e19f/installer/mix.exs#L12-L20

I wonder if we should postpone this change to a future Phoenix 1.9? @josevalim

josevalim (approved):
Yes, v1.9 material for sure. But it looks good. We can probably assign a v1.9 milestone and merge it once we finally branch v1.8 out. I don’t think it is worth branching out though, as I dont think there is anything beyond this PR for now?

elfenlaid:
> Yes, v1.9 material for sure. But it looks good. We can probably assign a v1.9 milestone and merge it once we finally branch v1.8 out. I don’t think it is worth branching out though, as I dont think there is anything beyond this PR for now?

I don't think there's anything else I can add to this PR. Thank you everyone for looking into this! 🙇

SMC17:
Approval already landed, so this is just a small confirmation pass from a Phoenix + Bandit + LiveView consumer.

The runtime `Code.ensure_loaded?(JSON)` switch in `config/config.exs` reads correctly against the CI matrix: `1.15.8 / OTP 25.3.2.9` resolves to `Jason`, `1.18.4 / OTP 27.3` and `1.18.4 / OTP 28.0.1` resolve to `JSON`. Both branches of the conditional are exercised on every CI run because the matrix straddles the cutoff, which is exactly the case `warn_on_missing_json_library` needs to prove out and does.

The `Phoenix.json_library/0` default in `lib/phoenix.ex` stays on `Jason` with a `v2.0` TODO comment. For library callers that depend on `Phoenix.json_library/0` without a config override, the behavior is unchanged on 1.9, which is the right backward-compat choice. Worth noting that new projects generated by the 1.9 installer get the explicit `config :phoenix, :json_library, JSON` line, so the default codepath in `Phoenix.json_library/0` is only reached by existing apps or by users who delete the line. Both are fine; just recording the surface for posterity.

The test refactor in `test/phoenix/test/conn_test.exs` to derive `json_error` from `Phoenix.json_library()` is cleaner than the previous hardcoded `Jason.DecodeError` assertion; it generalises to any future configured decoder that raises a typed exception. Same for the v1/v2 serializer tests routing through `Phoenix.json_library().decode!/1`.

@SteffenDE's point on the installer Elixir requirement is the one open thread: `installer/mix.exs` still pins `@elixir_requirement "~> 1.15"` and the comment block above it enumerates the five files that move together. Reading the existing tag policy and the "v1.9 milestone, merge after v1.8 branch-out" framing from @josevalim, my read is the requirement bump belongs in a separate 1.9-milestone PR rather than being folded in here, but I can see the argument either way. Is the intent to ship #6481 as-is and open a paired PR for the installer floor bump, or to extend this PR before v1.8 branches out?

Either way, this is a clean change. Thanks for putting it together @elfenlaid, and thanks @josevalim for the v1.9 milestone signal.

saleyn:
Looking at this PR - I believe it's not flexible enough to allow support of different JSON libraries beyond `Jason`, and `:json`.  It would be better to define a behavior for encoding/decoding, and whatever library is used, to ensure that it supports that behavior.  I am the maintainer of [glazer](https://glazer.hexdocs.pm), and would love to be able to use that JSON parsing library with Phoenix for performance reasons.

josevalim:
I’d recommend modules to provide a similar interface to the JSON in Elixir, as that will become effectively standard over time. And most of it was mirrored after Jason anyway for compatibility.

saleyn:
What would be a downside of providing a JSON behavior for phoenix?

josevalim:
The issue is that Ecto has its own library, Ash as well, and so on. If each of them need a behavior, you would then need all of them to agree or you end up with several wrappers. And if they are going to agree on one, the Elixir interface will most likely be the one to choose.

saleyn:
What I am asking is to document it in some form (could be at the Elixir level) so that library developers could follow the spec.  Right now there's none, which opens the doors for undesirably having to deal with several JSON library dependencies in your project (because Ecto, Phoenix, Ash, Bandit, etc. have their own API), when all you want is to have one pluggable JSON library of your choice.  Having a JSON behavior at the Elixir level could help, but even if it's just formally defined in documentation, we'd be at a better place.

SteffenDE:
@saleyn then send a PR please that adds a sentence to https://phoenix.hexdocs.pm/Phoenix.html#json_library/0, e.g. 

> Phoenix expects this to be a module with `decode!/1`, `decode_to_iodata!/1`, `encode!/1`, and `encode_to_iodata!/1`  functions.

saleyn:
> @saleyn then send a PR please that adds a sentence to https://phoenix.hexdocs.pm/Phoenix.html#json_library/0, e.g.
> 
> > Phoenix expects this to be a module with `decode!/1`, `decode_to_iodata!/1`, `encode!/1`, and `encode_to_iodata!/1`  functions.

I will submit a PR.  @SteffenDE, I know you would probably hate this, but I still have to ask. Would it be possible to require that the canonical function names for the JSON library interface not end with '!', so that the Erlang libraries could be used without requiring them to have an Elixir wrapper?

josevalim:
@saleyn if we change the API, then all existing uses of JSON and Jason then become broken, no?

saleyn:
> Phoenix expects this to be a module with `decode!/1`, `decode_to_iodata!/1`, `encode!/1`, and `encode_to_iodata!/1` functions.

> @saleyn if we change the API, then all existing uses of JSON and Jason then become broken, no?

This highlights a problem that was created when JSON's interface was designed. I guess the design didn't consider that it would become the standard for cross-languages.  Maybe it can be extended with adding  `decode/1`, `decode_to_iodata/1`, `encode/1`, and `encode_to_iodata/1` that return `{:ok|:error, result|reason}` tuples, and that would become the JSON's API for component designers, whereas the former 4 functions would use them under the hood?

On the other hand, I was under the impression that Erlang used to not permit '!' in function names, but I just tried it, and it does seem to work:

```erlang
$ cat mod.erl
-module(mod).
-export(['f!'/0]).

'f!'() ->
  os:system_time().
$ erlc mod.erl
$ erl
Erlang/OTP 29 [erts-17.0] [source] [64-bit] [smp:8:8] [ds:8:8:10] [async-threads:1] [jit:ns]

Eshell V17.0 (press Ctrl+G to abort, type help(). for help)
1> mod:'f!'().
1781527995879131559
```
So, my comment about the need to change the API is moot.

@josevalim are there any other concerns for an Erlang library to expose `decode!/1`, `decode_to_iodata!/1`, `encode!/1`, and `encode_to_iodata!/1` for use in Elixir?  If not, I'll add them to [glazer](https://github.com/saleyn/glazer).

josevalim:
Yes, the design happened by accident, unfortunately. And yes, you can totally have `!` in Erlang functions and that should work fine!!!

guides/controllers.md
rhcarvalho:
```suggestion
A step beyond this is rendering pure JSON with the [`json/2`] function. We need to pass it something that the [JSON library](https://elixir.hexdocs.pm/JSON.html) can decode into JSON, such as a map.
```

guides/json_and_apis.md
```diff
@@ -171,7 +171,7 @@ defmodule HelloWeb.UrlJSON do
 end
 ```
 
-This view is very simple. The `index` function receives all URLs, and converts them into a list of maps. Those maps are placed inside the data key at the root, exactly as we saw when interfacing with our application from `cURL`. In other words, our JSON view converts our complex data into simple Elixir data-structures. Once our view layer returns, Phoenix uses the `Jason` library to encode JSON and send the response to the client.
+This view is very simple. The `index` function receives all URLs, and converts them into a list of maps. Those maps are placed inside the data key at the root, exactly as we saw when interfacing with our application from `cURL`. In other words, our JSON view converts our complex data into simple Elixir data-structures. Once our view layer returns, Phoenix uses the built-in [JSON library](https://hexdocs.pm/elixir/1.18/JSON.html) to encode JSON and send the response to the client.
```
rhcarvalho:
```suggestion
This view is very simple. The `index` function receives all URLs, and converts them into a list of maps. Those maps are placed inside the data key at the root, exactly as we saw when interfacing with our application from `cURL`. In other words, our JSON view converts our complex data into simple Elixir data-structures. Once our view layer returns, Phoenix uses the built-in [JSON library](https://elixir.hexdocs.pm/JSON.html) to encode JSON and send the response to the client.
```

installer/templates/phx_single/config/config.exs.eex
```diff
@@ -59,8 +59,8 @@ config :logger, :default_formatter,
   format: "$time $metadata[$level] $message\n",
   metadata: [:request_id]
 
-# Use Jason for JSON parsing in Phoenix
-config :phoenix, :json_library, Jason
+# Use built-in `JSON` for JSON parsing in Phoenix
+config :phoenix, :json_library, JSON
```
rhcarvalho, line 69: I think here we could do a similar runtime detection as in https://github.com/phoenixframework/phoenix/pull/6481/changes#r2356177940, and only configure `JSON` as the default for Elixir versions where its available.

Same for the umbrella app config template.

lib/phoenix.ex
```diff
@@ -45,7 +47,7 @@ defmodule Phoenix do
 
   """
   def json_library do
-    Application.get_env(:phoenix, :json_library, Jason)
+    Application.get_env(:phoenix, :json_library, @default_json_library)
```
saleyn: Consider :persistent_term, which is roughly 30-100% faster:

```elixir
defmodule JSON.Adapter do
  @default_json_library if Code.ensure_loaded?(JSON), do: JSON, else: Jason

  def json_library do
    case :persistent_term.get(:json_library, nil) do
      nil ->
        module = :application.get_env(:phoenix, :json_library, @default_json_library)
        :persistent_term.put(:json_library, module)
        module
      module ->
        module
    end
  end
end
```

rhcarvalho (commented):
In the meantime HexDocs URLs now use subdomains.

The installer could use runtime detection to decide on the JSON library for new apps, that will keep the installer compatible with Elixir pre-1.18.

lib/phoenix.ex
elfenlaid: TIL `persistent_term` and it looks cool, but I'm not sure if we are in the position of adding anything significant to this PR. I'm happy to add it, but it's @SteffenDE call at this point

guides/controllers.md
elfenlaid: Thank you! ❤️

guides/json_and_apis.md
elfenlaid: Thank you! ❤️

installer/templates/phx_single/config/config.exs.eex
elfenlaid: It's a nice idea, yet I don't think it would fly.

The first option is to check Elixir version during `phx.new`, but the Elixir version that generates the project is not guaranteed to match the Elixir version that runs the project. So it sounds like a breaking change.

The second option is to embed this check into new projects. Technically it should work, but I'm not sure that such checks are awkward and probably are not something you would like to see in new projects.

saleyn:
> @saleyn then send a PR please that adds a sentence to https://phoenix.hexdocs.pm/Phoenix.html#json_library/0, e.g.
> 
> > Phoenix expects this to be a module with `decode!/1`, `decode_to_iodata!/1`, `encode!/1`, and `encode_to_iodata!/1`  functions.

@SteffenDE, could you clarify if this was a typo - there's no `JSON.decode_to_iodata!` in [JSON](https://elixir.hexdocs.pm/1.20.1/JSON.html). Did you mean that the library would need to support only 3 functions: `decode!/1`, `encode_to_iodata!/1`, `encode!/1`?

josevalim:
> Did you mean that the library would need to support only 3 functions: decode!/1, encode_to_iodata!/1, encode!/1?

yes!

saleyn:
Please check my PR above (#6732).

As a reference implementation, the [glazer_json](https://github.com/saleyn/glazer#elixirs-phoenix-json_library-compliance) module (part of the glazer Erlang JSON library) exposes these three functions under their quoted Erlang names — 'decode!'/1, 'encode!'/1, and 'encode_to_iodata!'/1 — and can be configured directly:

config :phoenix, :json_library, :glazer_json

For parity with Elixir's JSON module, where null decodes to and encodes from nil rather than the atom :null, these three functions apply that mapping automatically.

installer/templates/phx_single/config/config.exs.eex
SteffenDE: I think we'll need to bump the elixir requirement for the generated app to 1.18.

josevalim:
We should also remove `{:jason, "~> 1.2"}` from new apps. We may need to set the json library in a few other places though. Otherwise some parts will use Jason, others will use JSON, and that will be confusing!

rhcarvalho:

> We should also remove `{:jason, "~> 1.2"}` from new apps.

Verified that we're still pending a installer template update.

Merging in the latest changes in `main` (or rebasing) would help us reason about and validate the patch.

Now that `main` is tracking v1.9 development work, it's a good time to get this PR in final shape. @elfenlaid are you able to take this further?

elfenlaid:
> We should also remove {:jason, "~> 1.2"} from new apps.
> Verified that we're still pending a installer template update.

I'm not sure if that is possible to do with the current installer's requirements. Currently the installer requires 1.17, but JSON landed in 1.18 if I'm not wrong 🤔

josevalim:
We can bump the installer requirement too!

elfenlaid:
Oh nice, I think I can handle it then 😅

elfenlaid:
Oof, there's another problem, aside from Elixir 1.18+ we should also require Erlang/OTP 27+ for that to work -_-

Currently the installation instructions ask for Erlang/OTP 25+

josevalim:
26 and earlier are no longer officially supported, so we should bump that too!

guides/introduction/installation.md
```diff
@@ -2,33 +2,33 @@
 
 In order to build a Phoenix application, we will need a few dependencies installed in our Operating System:
 
-  * the Erlang VM and the Elixir programming language
```
elfenlaid, line 5: Sorry, this was an automatic fix from the markdown autoformatter. I can revert it, but it seems like acceptable change, unless those asterisks where intentional.

installer/lib/mix/tasks/phx.new.ex
```diff
@@ -565,66 +564,61 @@ defmodule Mix.Tasks.Phx.New do
     end
   end
 
-  # we need to parse JSON, so we only check for new versions on Elixir 1.18+
-  if Version.match?(System.version(), "~> 1.18") do
```
elfenlaid, line 569: Maybe this a too bold of a change, but it kinda makes sense if we insist on having Elixir 1.18+ / OTP 27 for the installer.

config/config.exs
```diff
@@ -5,7 +5,8 @@ config :logger, :console,
...
+  # TODO: Remove the `json_library` check once `JSON` becomes the standard `Phoenix.json_library/1`
+  json_library: (if Code.ensure_loaded?(JSON), do: JSON, else: Jason),
```
rhcarvalho, line 9: Now that we decided to bump Elixir/OTP version requirements, maybe we don't need the comment and conditional?

elfenlaid:
Okay, I've bumped the installer version to 1.18 and I think the PR is ready for yet another review pass.

I've tried to scope changes in commits and avoid collateral damage, but I understand that it's not that easy to follow given how the PR evolved. I can extract the installer changes to a separate PR if needed, just let me know.

config/config.exs
elfenlaid: This would be a breaking change as far as I can say. The installer creates a new app, while this is used in existing apps.

josevalim: config/config.exs is not used by dependencies. So we can likely default to JSON, no problems.

guides/controllers.md
```diff
@@ -98,7 +98,7 @@ end
 
 Now [`/hello/Frank`] in your browser should display `From messenger Frank` as plain text without any HTML.
 
-A step beyond this is rendering pure JSON with the [`json/2`] function. We need to pass it something that the [Jason library](`Jason`) can decode into JSON, such as a map. (Jason is one of Phoenix's dependencies.)
+A step beyond this is rendering pure JSON with the [`json/2`] function. We need to pass it something that the [JSON library](https://elixir.hexdocs.pm/JSON.html) can decode into JSON, such as a map.
```
josevalim:
```suggestion
A step beyond this is rendering pure JSON with the [`json/2`] function. We need to pass it something that the [JSON](https://elixir.hexdocs.pm/JSON.html) module can encode, such as a map.
```

guides/json_and_apis.md
```diff
@@ -175,7 +175,7 @@ defmodule HelloWeb.UrlJSON do
 end
 ```
 
-This view is very simple. The `index` function receives all URLs, and converts them into a list of maps. Those maps are placed inside the data key at the root, exactly as we saw when interfacing with our application from `cURL`. In other words, our JSON view converts our complex data into simple Elixir data-structures. Once our view layer returns, Phoenix uses the `Jason` library to encode JSON and send the response to the client.
+This view is very simple. The `index` function receives all URLs, and converts them into a list of maps. Those maps are placed inside the data key at the root, exactly as we saw when interfacing with our application from `cURL`. In other words, our JSON view converts our complex data into simple Elixir data-structures. Once our view layer returns, Phoenix uses the built-in [JSON library](https://elixir.hexdocs.pm/JSON.html) to encode JSON and send the response to the client.
```
josevalim:
```suggestion
This view is very simple. The `index` function receives all URLs, and converts them into a list of maps. Those maps are placed inside the data key at the root, exactly as we saw when interfacing with our application from `cURL`. In other words, our JSON view converts our complex data into simple Elixir data-structures. Once our view layer returns, Phoenix uses the built-in [JSON module](https://elixir.hexdocs.pm/JSON.html) to encode JSON and send the response to the client.
```

josevalim (commented):
This looks good to me. I have dropped some comments. I believe we also need to configure Ecto to use JSON.

config/config.exs
SteffenDE: It is used by tests though and we still test with Elixir 1.15

josevalim:
>  I believe we also need to configure Ecto to use JSON.

To be more precise, we nee to configure the underlying adapter (postgrex, myxql, tds, etc). We can do it in the same place we call `config :phoenix, :json_library, JSON`. We just swap the app name.

installer/templates/phx_single/config/config.exs.eex
```diff
@@ -76,6 +76,28 @@ config :logger, :default_formatter,
 # Use built-in `JSON` for JSON parsing in Phoenix
 config :phoenix, :json_library, JSON
 
+<%= if @ecto do %>
```
elfenlaid: Not the most elegant solution for sure 🙈

elfenlaid:
> This looks good to me. I have dropped some comments. I believe we also need to configure Ecto to use JSON.

Thanks, I've merged them and added the adapters configuration.

installer/templates/phx_single/config/config.exs.eex
josevalim: The best option is to define a variable in the generators themselves, something like @ecto_adapter_app or similar, anid then in here you do something like:

```
config :phoenix, :json_library, JSON<%= if @ecto_adapter_app do %>
config :<%= @ecto_adapter_app %>, :json_library, JSON<% end %>
```

elfenlaid: Oh, what's more, there's `adapter_app`, I haven't noticed it 😅

installer/templates/phx_single/config/config.exs.eex
```diff
@@ -73,9 +73,14 @@ config :logger, :default_formatter,
...
-# Use Jason for JSON parsing in Phoenix
-config :phoenix, :json_library, Jason
+# Use built-in `JSON` for JSON parsing in Phoenix
+config :phoenix, :json_library, JSON
 
+<%= if @ecto do %>
+# Use built-in `JSON` for JSON coding in Ecto
+config :<%= @adapter_app %>, :json_library, JSON
+
+<% end %>
```
josevalim: So we don't have extra newlines:

```suggestion
config :phoenix, :json_library, JSON<%= if @ecto do %>
config :<%= @adapter_app %>, :json_library, JSON<% end %>
```

installer/templates/phx_umbrella/config/extra_config.exs.eex
```diff
@@ -5,8 +5,8 @@ config :logger, :default_formatter,
   format: "$time $metadata[$level] $message\n",
   metadata: [:request_id]
 
-# Use Jason for JSON parsing in Phoenix
-config :phoenix, :json_library, Jason
+# Use built-in `JSON` for JSON parsing in Phoenix
+config :phoenix, :json_library, JSON
```
josevalim: We also need the ecto change here!

installer/templates/phx_single/config/config.exs.eex
```diff
@@ -73,9 +73,9 @@ config :logger, :default_formatter,
   format: "$time $metadata[$level] $message\n",
   metadata: [:request_id]
 
-# Use Jason for JSON parsing in Phoenix
-config :phoenix, :json_library, Jason
-
+# Use built-in `JSON` for JSON parsing in Phoenix
+config :phoenix, :json_library, JSON<%= if @ecto do %>
+config :<%= @adapter_app %>, :json_library, JSON<% end %>
```
josevalim:
```suggestion
# Use built-in `JSON` module
config :phoenix, :json_library, JSON<%= if @ecto do %>
config :<%= @adapter_app %>, :json_library, JSON<% end %>

```

installer/templates/phx_umbrella/config/extra_config.exs.eex
elfenlaid: Good catch and sorry for stealing so much of your time 🙇

config/config.exs
rhcarvalho: In that case we can just update the comment. We only need the conditional while Phoenix itself is tested against older Elixir/OTP, unrelated to any particular installer defaults.

SteffenDE:
```suggestion
  # TODO: Remove once we don't test with Elixir < 1.18 any more
  json_library: (if Code.ensure_loaded?(JSON), do: JSON, else: Jason),
```

SteffenDE (approved):

SteffenDE:
Thank you! 🙌🏻

PR #6491, opened 2025-09-24
Fix examples for route_prefix/2

vinceurag:
Very minor improvement

The examples for `route_prefix/2` is still using what I assume the old function name.


SteffenDE:
🙌🏻

PR #6494, opened 2025-09-27
fix: handle nil value assignment in core_components input

trileminh94:
Prevent potential nil value assignment by checking assigns[:value] first. If nil, fall back to field.value.

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -175,7 +175,7 @@ defmodule <%= @web_namespace %>.CoreComponents do
     |> assign(field: nil, id: assigns.id || field.id)
     |> assign(:errors, Enum.map(errors, &translate_error(&1)))
     |> assign_new(:name, fn -> if assigns.multiple, do: field.name <> "[]", else: field.name end)
-    |> assign_new(:value, fn -> field.value end)
+    |> assign(:value, if(is_nil(assigns[:value]), do: field.value, else: assigns[:value]))
     |> input()
   end
 


SteffenDE:
I don’t think this is a good change. If you explicitly pass a value, even if it’s nil, we should not fall back in my opinion. You can change that in your app of course, but I think it’s a sane default.

Can you add more context why you think this is necessary?

trileminh94:
Thanks for your comment! This change addresses an issue with form validation and value retention. Let me explain with a my scenario:

### The Problem with `assign_new`

The previous code:
```elixir
|> assign_new(:value, fn -> field.value end)
```

Only assigns when the `:value` key doesn't exist in assigns. This breaks form validation workflows.

### **Validation Scenario**

Here's the exact problem this fixes:

1. **Initial render**: Form loads, field has `nil` value initially
   ```elixir
   assigns = %{value: nil}  # :value key exists but is nil
   ```

2. **User input**: User types "invalid-email" and submits
   ```elixir
   # Form submission with user input
   ```

3. **Validation fails**: Changeset returns with errors, but we want to retain user's input
   ```elixir
   field.value = "invalid-email"  # This should be shown to user
   assigns = %{value: nil}        # But :value key already exists from step 1
   ```

4. **With `assign_new`**: ❌ **BROKEN** - User's input is lost!
   - `assign_new` sees `:value` key exists (even though it's `nil`)
   - Doesn't assign `field.value = "invalid-email"`
   - User sees empty field instead of their invalid input
   - **Terrible UX**: User has to retype everything!

5. **With current approach**: ✅ **WORKS** - User's input is preserved!
   - Checks `is_nil(assigns[:value])` → `true`
   - Assigns `field.value = "invalid-email"`
   - User sees their input with validation errors
   - **Good UX**: User can fix the error without retyping

trileminh94:
Thanks for your comment! This change addresses an issue with form validation and value retention. Let me explain with a my scenario:

### The Problem with `assign_new`

The previous code:
```elixir
|> assign_new(:value, fn -> field.value end)
```

Only assigns when the `:value` key doesn't exist in assigns. This breaks form validation workflows.

### **Form Validation Scenario**

Here's the exact problem this fixes:

1. **Initial render**: Form loads, field has `nil` value initially
   ```elixir
   assigns = %{value: nil}  # :value key exists but is nil
   ```

2. **User input**: User types "invalid-email" and submits
   ```elixir
   # Form submission with user input
   ```

3. **Validation fails**: Changeset returns with errors, but we want to retain user's input
   ```elixir
   field.value = "invalid-email"  # This should be shown to user
   assigns = %{value: nil}        # But :value key already exists from step 1
   ```

4. **With `assign_new`**: ❌ **BROKEN** - User's input is lost!
   - `assign_new` sees `:value` key exists (even though it's `nil`)
   - Doesn't assign `field.value = "invalid-email"`
   - User sees empty field instead of their invalid input
   - **Terrible UX**: User has to retype everything!

5. **With current approach**: ✅ **WORKS** - User's input is preserved!
   - Checks `is_nil(assigns[:value])` → `true`
   - Assigns `field.value = "invalid-email"`
   - User sees their input with validation errors
   - **Good UX**: User can fix the error without retyping

SteffenDE:
If you think something is wrong, please provide code to reproduce instead of AI generated steps that are hard to follow. Thanks!

trileminh94:
oh, I closed the issue because I found out the root cause, Claude has changed my core component

image

which cause the value always has default value so the value in the field can not be shown, so I thought there was an issue and I tried to find a way to deal with it. 

But actually the issue is claude has modified the core component behind my back. I had to generate a new project and compare line by line and I found out why  😂.

Thanks!! @SteffenDE

Issue #6203, opened 2025-04-18
static digested files without prefix get 404


ruslandoga:
👋 

### Environment

* Elixir version (elixir -v): 1.18.2
* Phoenix version (mix deps): 1.7.14
* Operating system:

### Actual behavior

It seems like a static digested `site.webmanifest` is not being served by Plug.Static.

```elixir
def static_paths, do: ~w(assets fonts images favicon.ico robots.txt site.webmanifest)
```

```elixir
iex> File.ls("lib/bend-0.1.0/priv/static")
#==> {:ok,
#==>  [..., "site-c0c4f1c0f3ba8cb5034fdd9d1aafa9aa.webmanifest",
#==>   "favicon-91f37b602a111216f1eef3aa337ad763.ico", "site.webmanifest",
#==>   "favicon-07f80321aa54b0c7f90564a833181023.ico"]}
```

the Plug.Static `init`ed opts are

```elixir
%{
  at: [],
  from: {:my_app, "priv/static"},
  headers: %{},
  only_rules: {["assets", "fonts", "images", "favicon.ico", "robots.txt",
    "site.webmanifest"], []},
  encodings: [{"gzip", ".gz"}],
  content_types: %{},
  et_cache: "public",
  et_generation: nil,
  qs_cache: "public, max-age=31536000, immutable"
}
```

and the requested path for site.webmanifest is

```
10:22:46.417 request_id=GDdik4hrQd68HZoAGg-D [info] GET /site-c0c4f1c0f3ba8cb5034fdd9d1aafa9aa.webmanifest
10:22:46.419 request_id=GDdik4hrQd68HZoAGg-D [info] Sent 404 in 1ms
```

I think the problem might also affect (digested) favicons and robots.txt but I haven't checked.

### Expected behavior

All digested files get served :)

ruslandoga:
I noticed that https://github.com/elixir-plug/plug/blob/4af671374a73d33b4a4b003ce8c84a4173428030/lib/plug/static.ex#L223 returns false for this request.

SteffenDE:
Good question! By default, the only files served from / are `favicon.ico` and `robots.txt` and it doesn't make much sense to digest those (as the browser requests `favicon.ico` by default, and crawlers will just try `robots.txt`). So while `mix phx.digest` still digests them, it is fine that we don't actually serve them.

[...]

PR #6495, opened 2025-09-27
Enable raise_on_missing_only in development

brendon9x:
Configure Plug.Static with raise_on_missing_only: code_reloading?() to help developers catch missing entries in static_paths() during development. This prevents the common trap where files like favicon.svg return 404 because they aren't properly declared in the :only list (often because they digested paths).

Part of the fix for: https://github.com/phoenixframework/phoenix/issues/6203

Requires: https://github.com/elixir-plug/plug/pull/1286


josevalim:
I think we don't raise on unknown options... which means we can pass this option even if Plug the version is not yet released. Let's see if CI agrees.

josevalim:

PR #6500, opened 2025-09-30
feat: custom dispatcher for Presence handle_diff broadcast

edgurgel:
Hi there!

I've started a [discussion in the Elixir forum](https://elixirforum.com/t/allow-phoenix-presence-to-specify-the-message-dispatcher/72673) about this change but then I noticed that the change in the end was pretty straightforward so I decided to open a PR as well.
 
If someone is using a non-standard metadata when subscribing to the topic it’s not possible to specify a custom dispatcher module. It will always use the `Phoenix.Channel.Server.dispatch/3` function.

At [Realtime](https://github.com/supabase/realtime/blob/d309c55bfb60c8377eb7cb4b240f2ecf7b2d6962/lib/realtime_web/channels/realtime_channel/message_dispatcher.ex#L8-L16) we use a custom metadata but our custom dispatcher can't be used as `Phoenix.Presence` always use `Phoenix.Channel.Server` as dispatcher. We considered having our own `Presence` module but it felt heavy handed given that we only needed a custom dispatcher in the end. 

Can totally understand if we don't want to expose this as a function but I thought I would try anyway 😅

### DIFF
--- a/lib/phoenix/presence.ex
+++ b/lib/phoenix/presence.ex
@@ -79,6 +79,18 @@ defmodule Phoenix.Presence do
 
   See `c:list/1` for more information on the presence data structure.
 
+  ## Custom dispatcher
+
+  It's possible to customize the dispatcher module used to broadcast.
+  By default, `Phoenix.Channel.Server` is used, which is the same dispatcher
+  used by channels. To customize the dispatcher, pass the `:dispatcher` option
+  when using `Phoenix.Presence`:
+
+      use Phoenix.Presence,
+        otp_app: :my_app,
+        pubsub_server: MyApp.PubSub,
+        dispatcher: MyApp.CustomDispatcher
+
   ## Fetching Presence Information
 
   Presence metadata should be minimized and used to store small,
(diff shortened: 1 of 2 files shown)


lib/phoenix/presence.ex
```diff
@@ -79,6 +79,18 @@ defmodule Phoenix.Presence do
...
+
+  It's possible to customize the dispatcher module used to broadcast.
+  By default, `Phoenix.Channel.Server` is used, which is the same dispatcher
+  used by channels. To customize the dispatcher, pass the `:dispatcher` option
+  when using `Phoenix.Presence`:
+
+      use Phoenix.Presence,
+        otp_app: :my_app,
+        pubsub_server: MyApp.PubSub,
+        dispatcher: MyApp.CustomDispatcher
```
josevalim, line 92: We need some documentation or links on what is the dispatch and how a custom implementation would look like. I don't know if we want to link to the endpoint, Elixir's Registry or what else, but we just need a bit more context here. :)

josevalim (approved):
LGTM, just one tiny nit!

SteffenDE:
🙌🏻

PR #6502, opened 2025-10-05
fix: autocomplete with email instead of username in email fields

IgnacioGoldchluk:
Hi

When starting a new project and running `mix phx.gen.auth`, all the email fields in the forms from the generated code contain `autocomplete="username"` instead of `autocomplete="email"`.

This is just a small annoyance in the UI side rather than a bug. The functionality is not affected at all but the users are shown possible usernames instead of the expected emails when filling the forms.


SteffenDE:
🙌🏻

PR #6507, opened 2025-10-06
Fix precommit task to use correct --warnings-as-errors flag

nerdyworm:
Corrects `mix compile` flag in precommit task from `--warning-as-errors` to `--warnings-as-errors`


SteffenDE:
Woops! Good catch! :)

PR #6508, opened 2025-10-08
Add Phoenix.Naming.titleize/1

nathanl:


chrismccord:
Do you need this for broader generator use or convenience for your own apps? Not opposed, just wondering if this is part of some larger work or not, but we also have to decide how much we want to place here if we're thinking about adding more utilities. THanks!

nathanl:
Convenience in an app. My current example: converting the snake case atoms of an Ecto enum type, like `:deep_freezer`, into a form select option. "Deep Freezer" looks more like an expected dropdown value to me than "Deep freezer". Personally I want something like this more often than `humanize/1`, but maybe that's just me.

rhcarvalho:

A quick point to help the team decide whether to move forward or close: `Phoenix.Naming` has traditionally stayed scoped to internal usage in the framework implementation and generators rather than acting as a general-purpose casing utility. That's the case for [all 6 functions that make up the current API surface](https://phoenix.hexdocs.pm/Phoenix.Naming.html).

An alternative for title-casing and more is [Recase](https://hex.pm/packages/recase) (`Recase.to_title/1`) (which I have no affiliation with). Unlike the proposed `titleize/1`, it doesn't special-case `_id` suffixes, but it handles Unicode.

Leaving it to the maintainers on whether expanding the `Phoenix.Naming` API surface makes sense here or if this is better suited for app-specific helpers or external libraries.

SteffenDE:
I think I'd do the same as we do for core components: only add functions we need and use for the generators. Using an external dependency (or implementing in your own app) feels like the way to go here. Thanks!

PR #6512, opened 2025-10-13
guide: suggestion for awkward sentence

ellyxir:
Original sentence 
> As we built our first page, we could start to understand how the request life-cycle is put together.

"As we built" sounded awkward, it makes me think we'll redo the same work. For example, we'll make this house just as we built the last one.

It feels more natural to me to say: "Having built". 

"We could start to understand" makes me think that we might also NOT understand. I modified to "we're beginning to understand" to match tense and meaning.

Minor changes. It's also fine the way it is! :)

### DIFF
--- a/guides/request_lifecycle.md
+++ b/guides/request_lifecycle.md
@@ -186,7 +186,7 @@ This line injects our template into the layout before the HTML is sent off to th
 
 ## From endpoint to views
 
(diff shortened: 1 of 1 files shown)


chrismccord:
❤️❤️❤️🐥🔥

PR #6513, opened 2025-10-14
Add `Phoenix.Controller.assign/2` to match LiveView API

adam12:
Adds `assign/2` to Phoenix.Controller for consistency with the LiveView API.
This delegates to `Plug.Conn.merge_assigns/2` and provides a familiar interface
for developers moving between controllers and LiveViews.

Before:

```
conn
|> assign(:name, "Alice")
|> assign(:role, :admin)
```

After:
```
assign(conn, name: "Alice", role: :admin)
```

This mirrors the `assign/2` function available in LiveView, making the assign
API consistent across both controller and LiveView contexts


josevalim:

PR #6521, opened 2025-10-21
Update docker template to debian trixie (stable)

suciptoid:
Update docker template to debian trixie (stable) replacing bookworm (oldstable)


SteffenDE:
Thank you! 🙌🏻

PR #6523, opened 2025-10-23
Add Seenode support to Phoenix deployment guide

GeoSegun:
Updated the deployment guide to include information about [Seenode](https://seenode.com), which offers first-class support for Phoenix applications along with [detailed documentation](https://seenode.com/docs/frameworks/elixir/phoenix/) and [a quickstart template](https://github.com/seenode/example-phoenix).


SteffenDE:
~Sorry about the late response. I don't think we should merge this, as it's something we have to maintain and Seenode it not a well known company in the ecosystem. But apart from that, I think we should remove the deployment guides for Fly, Gigalixir, and Heroku. If a platform has dedicated instructions for hosting Phoenix, we should instead link from a generic Cloud providers page to their documentation.~

Sorry about that, I misremembered the scope of this PR!

SteffenDE:
🙌️

PR #6526, opened 2025-10-31
feat(core_components): add a flash duration

mragentman:
- Upgraded the flash component to include a duration attribute to allow it to disappear automatically
  - Only applicable to liveviews with javascript
  - Defaults to 5 seconds
  - Allows flash messages to automatically disappear without the user needing to manually close
  - Supports hovering to pause
  - Provides a scaffolded example of a Colocated hook
- Updated the flash_group component to incorporate the duration attribute
  - The client and server error messages do not participate by setting duration to 0 since they already disappear automatically

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex
+++ b/installer/templates/phx_web/components/core_components.ex
@@ -42,7 +42,8 @@ defmodule <%= @web_namespace %>.CoreComponents do
   attr :id, :string, doc: "the optional id of flash container"
   attr :flash, :map, default: %{}, doc: "the map of flash messages to display"
   attr :title, :string, default: nil
-  attr :kind, :atom, values: [:info, :error], doc: "used for styling and flash lookup"
+  attr :kind, :atom, values: [:info, :error], doc: "used for styling and flash lookup"<%= if @live and @javascript do %>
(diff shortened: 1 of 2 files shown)


SteffenDE:
Hey @mragentman,

we generally try to keep the core components very minimal. This is good for you to use in your own apps, but I don’t think we want to add it by default. Having it not work for `--no-live` apps might also lead to confusion.

cc @josevalim

mragentman:
@SteffenDE - Thank you for the quick reply! I completely understand if you want to keep the core components simple. I thought this was a nice upgrade for anyone who wanted to utilize the default flash component. This could also help showcase usage of Colocated hooks in newly generated apps.

No hard feelings if the PR is rejected. I wanted to share an update that I had made to one of my projects, for others to also use as a starting point when they scaffold a new app.

josevalim:
Agreed with @SteffenDE, we want to keep it minimal and allow you to build on top of (like you did superbly here!)

mragentman:
@josevalim @SteffenDE - Thank you for the consideration. I completely understand wanting to keep things minimal, to reduce the maintainability burden. The scaffold provides a solid foundation that I wanted to help become a bit better for new users. I appreciate the discussion and your time. Thanks again!

PR #6530, opened 2025-11-10
Accept function in `Phoenix.Socket.assign/2`

timfjord:
This PR adds `assign_lazy/3` to `Phoenix.Socket` (if you like the idea, I will add a similar change to the `Phoenix.Component` module in the `phoenix_live_view`).

It is quite often that you need to update the socket based on the value that it is already in `assigns`.
For more complicated things, this could be handled with private functions, but this `assign_lazy/3` is useful when you need something very simple.

It is similar to the `assign_new/3`, but it assigns the value even if the given `key` exists.

```elixir
socket
|> assign_strcut1()
|> assign_struct2()
|> assign_lazy(:title, %{struct1: struct1, struct2: struct2} -> "#{strcut1.name} #{strcut2.name}") end)
```

### DIFF
--- a/lib/phoenix/socket.ex
+++ b/lib/phoenix/socket.ex
@@ -345,13 +345,18 @@ defmodule Phoenix.Socket do
 
   @doc """
   Adds key/value pairs to socket assigns.
+  Accepts a keyword list, a map, or a single-argument function.
 
-  A keyword list or a map of assigns must be given as argument to be merged into existing assigns.
+  When a keyword list or map is provided, it will be merged into the existing assigns.
+
+  If a function is given, it takes the current assigns as an argument and its return
+  value will be merged into the current assigns.
 
   ## Examples
 
       iex> assign(socket, name: "Elixir", logo: "💧")
       iex> assign(socket, %{name: "Elixir"})
+      iex> assign(socket, fn %{name: name, logo: logo} -> %{title: Enum.join([name, logo], " | ")} end)
 
   """
   def assign(%Socket{} = socket, keyword_or_map)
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks for the PR! I am not sure if `_lazy` is the correct name, as the assignment always happens. `_lazy` in Elixir only happens if there is no entry.

I feel this is one of the cases where it is probably better to break the pipeline apart (or, as you said, use custom functions).

timfjord:
Thanks for the feedback, @josevalim.
What if we handle this a bit differently?
We could add another clause for the `assign/2` and accept a function as the second argument
```elixir
def assign(%Socket{} = socket, fun) when is_function(fun, 1) do
  Map.update!(socket, :assigns, &Map.merge(&1, Map.new(fun.(&1))))
end
```
This way, the example from the description will look like
```elixir
socket
|> assign_strcut1()
|> assign_struct2()
|> assign(%{struct1: struct1, struct2: struct2} -> [title: "#{strcut1.name} #{strcut2.name}")] end)
```

josevalim:
That would work, yes!

timfjord:
@josevalim Done ✅

I've slightly adjusted the implementation, but it has a potential issue.
Because I use recursion (which is great because I don't need to duplicate the assigns merging logic), if the callback returns a single arity function there would be an infinite loop instead of the `ArgumentError` (if I went with this implementation `Map.update!(socket, :assigns, &Map.merge(&1, Map.new(fun.(&1))))`)

But it looks shorter and cleaner.

SteffenDE:
🙌🏻

PR #6536, opened 2025-11-16
sort query params in verified routes during tests

bottlenecked:
By having the query parameters being serialized in order (during tests only) when using VerifiedRoutes, we can make some test assertions (like e.g. `assert_redirect/3` from Phoenix.Liveview) less brittle when the query params are constructed dynamically in the backend and it's not easy to predict / maintain exact query param orders in tested urls.

---

Note 1: Originally [this PR](https://github.com/phoenixframework/phoenix_live_view/pull/4047) was attempted to address the issue directly in Liveview tests but it was suggested to make the change to `sigil_p` instead

Note 2: I thought it best to be consistent and sort all query strings - static or dynamic, and that's why I made the changes to `rewrite_path/4` which is the last function called in the verified routes pipeline. If we should only sort query params for dynamic params, we should just change `VerifiedRoutes.__encode_query__/1`

Note 3: I'm not quite sure that the approach taken here for figuring out if we're in test mode or not is the best. I think it won't negatively affect performance since the `if` expressions should be evaluated at compile time, but there could be better, more elegant ways to express the same - please let me know and I'll be happy to make the change

### DIFF
--- a/installer/templates/phx_single/config/test.exs
+++ b/installer/templates/phx_single/config/test.exs
@@ -23,3 +23,7 @@ config :phoenix, :plug_init_mode, :runtime<%= if @html do %>
 # Enable helpful, but potentially expensive runtime checks
 config :phoenix_live_view,
   enable_expensive_runtime_checks: true<% end %>
+
+# Sort query params output of verified routes for robust url comparisons
+config :phoenix,
+  sort_verified_routes_query_params: true
--- a/lib/phoenix/verified_routes.ex
+++ b/lib/phoenix/verified_routes.ex
@@ -42,6 +42,10 @@ defmodule Phoenix.VerifiedRoutes do
   Like path segments, query strings params are proper URL encoded and may be interpolated
   directly into the ~p string.
 
+  To ease url comparisons during tests (e.g. when using `assert_redirect/3`) query params
+  will be sorted. This is controlled by the `phoenix: [sort_verified_routes_query_params: true]`
+  configuration option.
+
   ## What about named routes?
 
   Many web frameworks, and early versions of Phoenix, provided a feature called "named routes".
@@ -821,7 +825,11 @@ defmodule Phoenix.VerifiedRoutes do
             "interpolated query string params must be separated by &, got: #{Macro.to_string(route)}"
     end
 
-    rewrite = {:"::", m1, [{{:., m2, [__MODULE__, :__encode_query__]}, m3, [arg]}, bin]}
+    sort_params? = Application.get_env(:phoenix, :sort_verified_routes_query_params, false)
+
+    rewrite =
+      {:"::", m1, [{{:., m2, [__MODULE__, :__encode_query__]}, m3, [arg, sort_params?]}, bin]}
+
     verify_query(rest, route, [rewrite | acc])
   end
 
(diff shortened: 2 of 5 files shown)


lib/phoenix/verified_routes.ex
```diff
@@ -979,6 +979,19 @@ defmodule Phoenix.VerifiedRoutes do
     {static?, meta, test_path, path_ast, static_ast}
   end
 
+  if Mix.env() == :test do
```
SteffenDE: As far as I know dependencies are always compiled as MIX_ENV=prod, even when you mix for your project in MIX_ENV=test. We’d need to add a config option like `config :phoenix, sort_verified_routes_querystring: true` and then we’d set that in the test config of new Phoenix projects and people can set if for themselves in existing ones.

bottlenecked: > As far as I know dependencies are always compiled as MIX_ENV=prod, even when you mix for your project in MIX_ENV=test

Oh, I totally missed that fact, thanks for pointing this out.

So just to get a better sense of how someone would do that using a custom config flag:
- first one would add a new flag under `:phoenix` in `installer/templates/phx_single/config/test.exs` (e.g. like `:sort_query_params_in_verified_routes`) so that new phoenix installations get this by default
- then we'd check for that flag at compile time using `Application.compile_env/3` instead of checking against `Mix.env()`
- third, add a hint in the changelog for the next release so that users can opt in to that behavior if they find it useful for their existing projects. Should probably also discourage people from adding this flag in prod.exs or similar as that would potentially make production code unnecessarily sort query params (ie adversely affect performance)

Did I get what you suggest or am I way off? Thanks for the feedback by the way.

josevalim: The compile_env idea is right but I think you want to change the `__encode_query__` function instead, as that is the part responsible for encoding dynamic query strings. Static query strings should preserve the user defined order.

lib/phoenix/verified_routes.ex
```diff
@@ -821,7 +823,10 @@ defmodule Phoenix.VerifiedRoutes do
             "interpolated query string params must be separated by &, got: #{Macro.to_string(route)}"
     end
```
josevalim: You can read the environment here!

```suggestion
Application.get_env(:phoenix, :sort_verified_routes_query_params)
```

Also, let's use query_params instead of query_string since we only sort when they are rich data structures.

josevalim (commented):
Perfect, one comment and we can ship it!!!!

josevalim:
Oh, we need a bit of docs in the VerifiedRoutes module.

bottlenecked:
I force-pushed a new version now
- uses `Application.get_env/3` closer to the call site instead of `Application.compile_env/3`
- updated the configuration option's name to `sort_verified_routes_query_params`
- Added an extra paragraph in the moduledocs mentioning this new behaviour. Let me know if I should extend with examples or move elsewhere instead

bottlenecked:
I think this PR would need a workflow approval before it can be merged 
image

josevalim:
I started the workflow but as soon as another reviewer goes through it, it will be merged!

installer/templates/phx_single/config/test.exs
```diff
@@ -23,3 +23,7 @@ config :phoenix, :plug_init_mode, :runtime<%= if @html do %>
 # Enable helpful, but potentially expensive runtime checks
 config :phoenix_live_view,
   enable_expensive_runtime_checks: true<% end %>
+
+# Sort query params output of verified routes for robust url comparisons
+config :phoenix,
+  sort_verified_routes_query_params: true
```
SteffenDE: We'll need the same in the umbrella config, see https://github.com/phoenixframework/phoenix/blob/main/installer/templates/phx_umbrella/config/test.exs

josevalim:

PR #6547, opened 2025-11-27
Update Dockerfile steps in Releases doc

DavidOliver:
I've left the image references as they are as I'm not currently in a position to use the latest versions.

Edit: this is based on release file generation in Phoenix 1.8.1 - I'm not sure if 1.8.2 has changed anything here. Will check...

Edit 2: libncurses is 6 instead of 5 now. Updating now.

Edit 3: force pushed to this patch-2 branch.


SteffenDE:
Thank you! 🙌🏻

PR #6549, opened 2025-11-29
Fix endpoint port config in an umbrella application.

khustochka:
**TLDR:** This PR fixes endpoint port configuration in `runtime.exs` in a newly generated umbrella application. This is a regression introduced in Phoenix 1.8.2.

I have noticed that with upgrade to Phoenix 1.8.2 port configuration has been moved to `runtime.exs` and new lines have been added (#6449). I checked what's changed for an umbrella application (on [phoenixdiff](https://github.com/phoenix-diff/phoenix-diff/blob/master/priv/data/sample-app/1.8.2/umbrella/config/runtime.exs#L10)) and these are the lines that are added:

```
config :sample_app, SampleAppWeb.Endpoint,
  http: [port: String.to_integer(System.get_env("PORT", "4000"))]
```

while everywhere else, even in the very same file, `SampleAppWeb.Endpoint` is configured on `sample_app_web`, not `sample_app`.

**Steps to reproduce:**

1. Create a new umbrella application on Phoenix 1.8.2. Install all dependencies.

```
mix phx.new test182 --umbrella
```

2. Switch to the app

```
cd test182_umbrella
```

3. Start the app with alternative port

```
PORT=4500 iex -S mix phx.server
```

Observe the output. 

*Actual:*
```
[info] Running Test182Web.Endpoint with Bandit 1.8.0 at 127.0.0.1:4000 (http)
[info] Access Test182Web.Endpoint at http://localhost
```

As we see, the custom port is not obeyed, and the endpoint URL is missing the port altogether.

*Expected:*
```
[info] Running Test182Web.Endpoint with Bandit 1.8.0 at 127.0.0.1:4500 (http)
[info] Access Test182Web.Endpoint at http://localhost:4500
```

### DIFF
--- a/installer/test/phx_new_umbrella_test.exs
+++ b/installer/test/phx_new_umbrella_test.exs
@@ -81,7 +81,12 @@ defmodule Mix.Tasks.Phx.New.UmbrellaTest do
         assert file =~ "port: 80"
       end)
 
-      assert_file(root_path(@app, "config/runtime.exs"), ~r/ip: {0, 0, 0, 0, 0, 0, 0, 0}/)
+      assert_file(root_path(@app, "config/runtime.exs"), fn file ->
+        assert file =~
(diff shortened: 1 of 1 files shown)


SteffenDE:
🙌🏻

PR #6553, opened 2025-12-03
Update controllers.md example

armetiz:
As a newcomer to this technology, I am reading the documentation carefully.

I notice that the second argument of the `render/3` function should be consistent with the name of the action. Which is not...

Regards

### DIFF
--- a/guides/controllers.md
+++ b/guides/controllers.md
@@ -47,7 +47,7 @@ defmodule HelloWeb.PageController do
   ...
 
   def index(conn, _params) do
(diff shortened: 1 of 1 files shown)


chrismccord:
In this case, you'd need to instruct the users that they would also need to rename the home.html.heex to index.html.heex, so not sure if that's preferable to existing action rename only. Thoughts?

armetiz:
Actually, I hadn't noticed that the second argument of `render/3` was the template... 😅

This PR can be closed.

PR #6554, opened 2025-12-07
Fix URL pointing to Phoenix.Component

acramatte:
- add missing top level domain in the URL pointing to Phoenix.Component


SteffenDE:
Thank you! 🙌🏻

PR #6558, opened 2025-12-10
Fix inconsistent form field access in phx.gen.auth login template

jamiecurle:
The password input uses `@form[:password]` while other inputs in the same form use `f[:email]` via the `:let={f}` binding. This changes it to `f[:password]` for consistency.


SteffenDE:
I'm pretty sure I did this intentionally, as only the email input is rendered twice (from the same form), so we need `f` to prevent duplicate ids.

jamiecurle:
Sure thing - feel free to close the pr if it's not needed. I appreciate the speedy reply.

PR #6559, opened 2025-12-10
Fix unicode character validation in verified routes

### LINKED ISSUE #6524: `~p` sigil does not work with non-ascii characters
### Environment

* Elixir version (elixir -v): 1.19.1
* Phoenix version (mix deps): 1.8.1
* Operating system: macOS 26

### Actual behavior

With a route containing a unicode character, the `~p` sigil will warn that the route does not exist.

```elixir
# in lib/unicode_verified_routes_web/router.ex
    get("/ø", HomeController, :index)
# in eg. a template
<a href={~p"/ø"}>Ø</a>
```

```
$ mix compile
    warning: no route path for UnicodeVerifiedRoutesWeb.Router matches "/ø"
    │
  1 │ <a href={~p"/ø"}>Ø</a>
    │ ~~~~~~~~~~~~~~~~~~~~~~
    │
    └─ lib/unicode_verified_routes_web/controllers/page_html/home.html.heex:1: UnicodeVerifiedRoutesWeb.PageHTML.home/1
```

<details>

<summary>Reproduction code. However the compile warning is not shown when running as .exs</summary>

```elixir
Application.put_env(:phoenix, :json_library, JSON)

Application.put_env(:sample, SamplePhoenix.Endpoint,
  http: [ip: {127, 0, 0, 1}, port: 5001],
  server: true,
  adapter: Bandit.PhoenixAdapter,
  secret_key_base: String.duplicate("a", 64)
)

Mix.install([
  {:bandit, "~> 1.8"},
  {:phoenix, "~> 1.8"}
])

defmodule SamplePhoenix.SampleController do
  use Phoenix.Controller, formats: [:html, :json]

  use Phoenix.VerifiedRoutes,
    endpoint: SamplePhoenix.Endpoint,
    router: Router

  def index(conn, _) do
    send_resp(conn, 200, ~p"/ø")
  end
end

defmodule Router do
  use Phoenix.Router

  pipeline :browser do
    plug(:accepts, ["html"])
  end

  scope "/", SamplePhoenix do
    pipe_through(:browser)

    get("/", SampleController, :index)
    get("/ø", SampleController, :index)

    # Prevent a horrible error because ErrorView is missing
    get("/favicon.ico", SampleController, :index)
  end
end

defmodule SamplePhoenix.Endpoint do
  use Phoenix.Endpoint, otp_app: :sample
  plug(Router)
end

{:ok, _} = Supervisor.start_link([SamplePhoenix.Endpoint], strategy: :one_for_one)
Process.sleep(:infinity)

```

</details>

### Expected behavior

That the route would be detected and work as a regular route would.

peguimasid:
### Fix Unicode Character Validation in Verified Routes

Fixes #6524 

Routes containing Unicode characters (e. g., `/ø`) were incorrectly generating compile-time warnings that the route doesn't exist, even when properly defined in the router. The route would function correctly at runtime, but developers would see false warnings during compilation.

#### Before

```elixir
# Router
get "/ø", HomeController, :index

# Template
<a href={~p"/ø"}>Ø</a>
```

Would produce: 
```
warning: no route path for MyAppWeb.Router matches "/ø"
```

#### After

No warning is generated, and the route verification passes correctly while still generating properly URL-encoded paths (`/%C3%B8`).

#### Root Cause

The issue occurred during compile-time route verification:

1. When processing `~p"/ø"`, the path is URI-encoded to `/%C3%B8` in `verify_segment/3`
2. This encoded path becomes the `test_path` used for verification
3. In `split_test_path/1`, the path `/%C3%B8` was split into segments `["%C3%B8"]` without decoding
4. The router's route pattern contains the literal Unicode string `["ø"]`
5. Pattern matching `["%C3%B8"]` against `["ø"]` fails, causing a false warning

#### Solution

Modified `split_test_path/1` to URI-decode path segments before comparison, ensuring both the test path and router's route patterns use the same encoding (decoded Unicode) during verification.

#### Changes

**lib/phoenix/verified_routes.ex**

```elixir
defp split_test_path(test_path) do
  test_path
  |> String.split("#")
  |> Enum.at(0)
  |> String.split("/")
  |> Enum.filter(fn segment -> segment != "" end)
  |> Enum.map(&URI.decode/1)  # Added: Decode segments for proper comparison
end
```

#### Testing

Added test coverage for Unicode character handling:

```elixir
test "~p with unicode characters" do
  assert ~p"/ø" == "/%C3%B8"
end
```

All existing tests pass (1052 tests, 0 failures).

#### Impact

- ✅ Routes with Unicode characters no longer generate false warnings
- ✅ Generated paths remain properly URI-encoded for URLs
- ✅ No breaking changes to existing functionality
- ✅ Backward compatible

#### Checklist

- [x] Tests added for Unicode character support
- [x] All existing tests pass
- [x] No breaking changes

### DIFF
--- a/lib/phoenix/verified_routes.ex
+++ b/lib/phoenix/verified_routes.ex
@@ -329,6 +329,7 @@ defmodule Phoenix.VerifiedRoutes do
     |> Enum.at(0)
     |> String.split("/")
     |> Enum.filter(fn segment -> segment != "" end)
+    |> Enum.map(&URI.decode/1)
   end
 
   defp expand_alias({:__aliases__, _, _} = alias, env),
--- a/test/phoenix/verified_routes_test.exs
+++ b/test/phoenix/verified_routes_test.exs
@@ -56,6 +56,7 @@ defmodule Phoenix.VerifiedRoutesTest do
     get "/posts/file/*file", PostController, :file
     get "/posts/skip", PostController, :skip
     get "/should-warn/*all", PostController, :all, warn_on_verify: true
+    get "/ø", PostController, :unicode
 
     scope "/", host: "users." do
       post "/host_users/:id/info", UserController, :create
@@ -432,6 +433,10 @@ defmodule Phoenix.VerifiedRoutesTest do
     assert ~p"/posts/5/?#{[id: 5]}" == "/posts/5/?id=5"
     assert ~p"/posts/5/?#{%{"id" => "foo"}}" == "/posts/5/?id=foo"
     assert ~p"/posts/5/?#{%{"id" => "foo bar"}}" == "/posts/5/?id=foo+bar"
+  end
+
+  test "~p with unicode characters" do
+    assert ~p"/ø" == "/%C3%B8"
   end
 
   describe "with static path" do


josevalim:

PR #6561, opened 2025-12-14
phx.gen.auth/schema_token.ex: Improve documentation wording

xtian:
The current passive voice phrasing "is being used within" is a little awkward.


SteffenDE:
🙌🏻

PR #6562, opened 2025-12-16
Fix LongPoll transport name in sessionStorage and logs

rhcarvalho:
Projects created with `mix phx.new` enable JavaScript minification by default in production. This renames symbols to reduce code size, including the LongPoll transport class name.

As a result, when the Socket falls back to LongPoll, it stores something like `phx:fallback:Jt` in sessionStorage instead of the expected `phx:fallback:LongPoll` (as it does in development).

The shortened name is not guaranteed to be stable across builds, leading to unintended behavior when deploying new versions of an app.

The transport name was also used in log messages, which now will also preserve the correct LongPoll name.

image

### DIFF
--- a/assets/js/phoenix/socket.js
+++ b/assets/js/phoenix/socket.js
@@ -367,8 +367,25 @@ export default class Socket {
 
   /**
    * @private
+   *
+   * @param {Function}
    */
+  transportName(transport){
+    // JavaScript minification, enabled by default in production in Phoenix
+    // projects, renames symbols to reduce code size.
+    // See https://esbuild.github.io/api/#keep-names.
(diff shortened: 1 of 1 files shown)


rhcarvalho:
Open to suggestions on what would be useful automated tests for this.

I looked at `assets/test/socket_test.js` quickly, didn't find tests that would exercise the `logger` to a point that could expose the problem. Also found nothing on the interaction with `sessionStorage`.

Note `LongPoll.name` falls apart only when the source is minified, so that is yet another thing tests would need to replicate.

SteffenDE (approved):

chrismccord:
❤️❤️❤️🐥🔥

PR #6564, opened 2025-12-23
Small improvements to a couple of pages of the starting guides

siiky:


stjhimy:
> Also note the controller name (`HelloController`) and the view name (`HelloHTML`) and their respective files, `hello_controller.ex` and `hello_html.ex` all follow the same naming convention.  and are named after each other. You could name the directory anything you want, as long as you update the `embed_templates` setting accordingly, but it's best to follow conventions.

Shouldn't there be a capital A? `... convention.  And are named....` ?

siiky:
It's a comma, as it was originally (see https://github.com/phoenixframework/phoenix/commit/c8c6cea1f819f4596e02d07561a1e0427da41eaa). A full sentence to say "And are named after each other." by itself would be a little weird in my opinion.

SteffenDE:
🙌🏻

PR #6566, opened 2025-12-28
redirect_if_user_is_authenticated in login_live

kayuapi:


SteffenDE:
The login live view is also used to re-authenticate when changing password, so we deliberately don't redirect!

PR #5143, opened 2022-12-02
Add autocomplete attrs to password/username fields

janwillemvd:
Some browsers give console warnings if the `autocomplete` attribute is missing on username/password fields.

PR #6573, opened 2026-01-06
Remove autocomplete attribute from hidden field

rhcarvalho:
Introduced in #5143 ([exact line](https://github.com/phoenixframework/phoenix/pull/5143/files#diff-134c438d4061025dd7ca9d1e4d9937e4684d6069c9cc30e2abf1286be8e2628eR57)), probably as an accident when adding `autocomplete` to many form fields.


SteffenDE:
🙌🏻

PR #6574, opened 2026-01-06
Revert "fix: autocomplete with email instead of username in email fields"

rhcarvalho:
Reverts phoenixframework/phoenix#6502

The choice of `username` over `email` has actually been discussed in the PR that introduced the `autocomplete` attribute: https://github.com/phoenixframework/phoenix/pull/5143#discussion_r1100920085.

That conversation links to https://www.chromium.org/developers/design-documents/form-styles-that-chromium-understands/, which in turn links to https://web.dev/articles/sign-in-form-best-practices, quoted below:

> For email inputs use autocomplete="username", since username is recognized by password managers in modern browsers

If not reverting that PR, and rather deciding for `autocomplete="email"`, then we should update leftover `autocomplete="username"` occurrences in `phx.gen.auth` template forms.

<details>

```console
$ git grep -n 'autocomplete=' | grep username
priv/templates/phx.gen.auth/registration_live.ex:30:            autocomplete="username"
priv/templates/phx.gen.auth/settings_live.ex:24:          autocomplete="username"
```

</details>

### DIFF
--- a/priv/templates/phx.gen.auth/login_live.ex
+++ b/priv/templates/phx.gen.auth/login_live.ex
@@ -47,7 +47,7 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
             field={f[:email]}
             type="email"
             label="Email"
-            autocomplete="email"
+            autocomplete="username"
             spellcheck="false"
             required
             phx-mounted={JS.focus()}
@@ -72,7 +72,7 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
             field={f[:email]}
             type="email"
             label="Email"
-            autocomplete="email"
+            autocomplete="username"
             spellcheck="false"
             required
           />
--- a/priv/templates/phx.gen.auth/registration_new.html.heex
+++ b/priv/templates/phx.gen.auth/registration_new.html.heex
@@ -18,7 +18,7 @@
         field={f[:email]}
         type="email"
         label="Email"
-        autocomplete="email"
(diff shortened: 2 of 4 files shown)


rhcarvalho:
I'm not sure whether the CI failure is related to the changes https://github.com/phoenixframework/phoenix/actions/runs/20757919267/job/59604883857#step:4:1017, or a flake.

SteffenDE:
CI should be good. I'm fine with merging this - we should be consistent - but I'd be interested in knowing if there are cases where browsers or extensions actually suggest usernames (since we also set `type="email"`). At least it sounded like @IgnacioGoldchluk was experiencing this?

IgnacioGoldchluk:
In my case I had both `username` and `email` in the application, maybe that's why I was seeing incorrect suggestions? If the recommended practice is to use `autocomplete="username"` then I agree with the change being reverted

rhcarvalho:
> If the recommended practice is to use `autocomplete="username"` then I agree with the change being reverted

I believe the standard practice is `type="email" autocomplete="username"`, because that's the standard field name in password managers, e.g.:

image

image

image

image

(those login pages match part of what I listed in #6575 for adding `spellcheck="false"` to the same form fields)

josevalim (approved):

SteffenDE:
Sorry @rhcarvalho, can you please rebase?

rhcarvalho:
> Sorry @rhcarvalho, can you please rebase?

@SteffenDE done.

SteffenDE:
Lovely, thank you!

PR #6579, opened 2026-01-13
Stop reconnecting when close was clean

avs-doom:
If the connection was closed gracefully then skip the reconnect in the "visibilitychange" event.


SteffenDE (approved):

SteffenDE:
Thank you!

PR #6581, opened 2026-01-15
Fix AGENTS.md including Ecto references in --no-ecto projects

bmalum:
When generating a new Phoenix project with `--no-ecto`, the AGENTS.md file still contained references to Ecto (changesets, schemas, etc.). This confused AI agents working on these projects since they'd see Ecto guidelines, but the project doesn't actually use Ecto.

+ Added a test to verify AGENTS.md doesn't contain Ecto references when using `--no-ecto`.

Please feel free to comment with ideas on how to better handle the `<.form for={@changeset} id="my-form">` to `<.form for={@data} id="my-form">` change. Ideally we'd keep changeset info for Ecto projects but use generic data examples for non-Ecto Phoenix apps.

### DIFF
--- a/installer/lib/phx_new/generator.ex
+++ b/installer/lib/phx_new/generator.ex
@@ -123,11 +123,12 @@ defmodule Phx.New.Generator do
           # rules specific to new apps
           @new_project_rules_files["project.md"],
           @new_project_rules_files["phoenix.md"],
+          project.binding[:html] && @new_project_rules_files["phoenix-ui.md"],
           project.binding[:html] && @new_project_rules_files["phoenix-html.md"],
           project.binding[:live] && @new_project_rules_files["phoenix-live.md"],
           # --no-assets is equivalent to --no-tailwind && --no-esbuild;
-          # we check for both here
-          project.binding[:javascript] && project.binding[:css] &&
+          # Only include assets.md for HTML projects (not API-only)
+          project.binding[:html] && project.binding[:javascript] && project.binding[:css] &&
             @new_project_rules_files["assets.md"],
           # generic usage rules
           "\n<!-- usage-rules-start -->",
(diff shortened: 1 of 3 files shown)


usage-rules/ecto.md
```diff
@@ -7,3 +7,26 @@
 - You **must** use `Ecto.Changeset.get_field(changeset, :field)` to access changeset fields
 - Fields which are set programatically, such as `user_id`, must not be listed in `cast` calls or similar for security purposes. Instead they must be explicitly set when creating the struct
 - **Always** invoke `mix ecto.gen.migration migration_name_using_underscores` when generating migration files, so the correct timestamp and conventions are applied
+
+### Creating LiveView forms from changesets
```
SteffenDE: This only moves the problem around. Now, you'd get LiveView content when you use `--no-live`. For the AGENTS.md, we could solve this by using .md.eex files and then checking `<%= if live do %>` etc., but since these files are also valid [usage_rules](https://hexdocs.pm/usage_rules/readme.html), so we can't use eex :(

bmalum: good catch!

bmalum:
I agree with your comment @SteffenDE - even with --no-ecto we still have Ecto references. It's more complex to create the AGENTS.md for all phx.new flags. But keeping Ecto references in non-Ecto Phoenix projects throws off the Agents.

Happy to spend some time looking into all cases if we want an AGENTS.md for all scenarios.

usage-rules/html.md
```diff
@@ -2,9 +2,9 @@
 
 - Phoenix templates **always** use `~H` or .html.heex files (known as HEEx), **never** use `~E`
 - **Always** use the imported `Phoenix.Component.form/1` and `Phoenix.Component.inputs_for/1` function to build forms. **Never** use `Phoenix.HTML.form_for` or `Phoenix.HTML.inputs_for` as they are outdated
-- When building forms **always** use the already imported `Phoenix.Component.to_form/2` (`assign(socket, form: to_form(...))` and `<.form for={@form} id="msg-form">`), then access those forms in the template via `@form[:field]`
+- When building forms **always** use the already imported `Phoenix.Component.to_form/2` (`assign(:form, to_form(...))` and `<.form for={@form} id="msg-form">`), then access those forms in the template via `@form[:field]`
```
josevalim: Phoenix.Component already assumes LiveView, so this change still assumes LiveView. I would keep it as is, I think the templates one below is fine.

PR #6597, opened 2026-02-12
Fix bad link

siiky:
The linked page was removed in https://github.com/elixir-lang/elixir/commit/6909684d6, and the section was readded to a new page in https://github.com/elixir-lang/elixir/commit/1ecdd2d73.


SteffenDE:
Thank you! 🙌

PR #6604, opened 2026-02-24
Experimental WebTransport support for phoenix

lukaszsamson:
This PR adds experimental [WebTransport](https://www.w3.org/TR/webtransport/) support in phoenix using cowboy backend with native quicer HTTP 3 library

Widespread WebTransport adoption on the browser side is still hindered by Safari (support landing on 26.4) (see https://caniuse.com/webtransport), cowboy HTTP 3 support is still immature and on the infra side QUIC traffic is another story so I don't expect this being a big priority soon

This repo https://github.com/lukaszsamson/phoenix_webtransport_demo has a demo app

Screenshot 2026-02-24 at 23 33 10
Screenshot 2026-02-24 at 23 34 18

### DIFF
--- a/assets/js/phoenix/webtransport.js
+++ b/assets/js/phoenix/webtransport.js
@@ -1,0 +1,251 @@
+import {
+  global,
+  SOCKET_STATES,
+  TRANSPORTS,
+  AUTH_TOKEN_PREFIX
+} from "./constants"
+
+const HEADER_SIZE = 5
+const TEXT_FRAME_TYPE = 0
+const MAX_FRAME_SIZE = 16_777_216
+
+let appendBuffer = (left, right) => {
+  let merged = new Uint8Array(left.byteLength + right.byteLength)
+  merged.set(left, 0)
+  merged.set(right, left.byteLength)
+  return merged
+}
+
+let encodeText = (text) => {
+  if(typeof (TextEncoder) !== "undefined"){
+    return new TextEncoder().encode(text)
+  }
+
+  let encoded = unescape(encodeURIComponent(text))
+  let bytes = new Uint8Array(encoded.length)
+  for(let i = 0; i < encoded.length; i++){ bytes[i] = encoded.charCodeAt(i) }
+  return bytes
+}
+
+let decodeText = (bytes) => {
+  if(typeof (TextDecoder) !== "undefined"){
+    return new TextDecoder().decode(bytes)
+  }
+
+  let encoded = ""
+  for(let i = 0; i < bytes.length; i++){ encoded += String.fromCharCode(bytes[i]) }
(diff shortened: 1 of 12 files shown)


assets/js/phoenix/webtransport.js
```diff
@@ -0,0 +1,251 @@
...
+  let merged = new Uint8Array(left.byteLength + right.byteLength)
+  merged.set(left, 0)
+  merged.set(right, left.byteLength)
+  return merged
+}
+
+let encodeText = (text) => {
+  if(typeof (TextEncoder) !== "undefined"){
+    return new TextEncoder().encode(text)
+  }
```
SteffenDE, line 22: Is there any browser that has WebTransport but no TextEncoder?

assets/js/phoenix/webtransport.js
```diff
@@ -0,0 +1,251 @@
...
+    return (endPoint
+      .replace("ws://", "https://")
+      .replace("wss://", "https://")
+      .replace(new RegExp("(.*)\/" + TRANSPORTS.websocket), "$1/" + TRANSPORTS.webtransport))
+  }
+
+  endpointURL(){
+    let url = this.toURL(this.endpoint)
+    if(this.authToken){
+      url.searchParams.set("auth_token", this.authToken)
```
SteffenDE, line 76: Note that the whole idea behind the auth token is that it's NOT included in query parameters, as those are often part of access logs.

It seems like this is coming at some point: https://github.com/w3c/webtransport/pull/713

SteffenDE (commented):
This is interesting for sure, but I don't think we'd go ahead with this without having it also working with Bandit and without it properly handling the auth token (see comment).

On the JS side, you can already pass a custom transport when constructing the socket, so maybe what we could have for the time being is a better way to do custom server side transports without needing all the code in Phoenix itself? (Not sure how to best do that though)

assets/js/phoenix/webtransport.js
lukaszsamson: It seems the check is not needed

lukaszsamson:
> This is interesting for sure, but I don't think we'd go ahead with this without having it also working with Bandit and without it properly handling the auth token (see comment).

Fair. I treat this PR as an exploration. The tech is not yet ready for wide deployment

> On the JS side, you can already pass a custom transport when constructing the socket, so maybe what we could have for the time being is a better way to do custom server side transports without needing all the code in Phoenix itself? (Not sure how to best do that though)

I haven't checked the custom transports so I cant tell they are flexible enough. IMO when the tech matures phoenix would need to adopt it as an officially supported transport

SteffenDE:
> IMO when the tech matures phoenix would need to adopt it as an officially supported transport

I agree :)

PR #6605, opened 2026-02-25
Use custom migration module for auth migration.

ShPakvel:


ShPakvel:
There is not related random CI test fail. Here is all tests passed on the branch https://github.com/ShPakvel/phoenix/actions/runs/22409189048

SteffenDE:
Thank you!

PR #6607, opened 2026-02-27
Include cookie options when deleting remember me cookie

andrewtimberlake:
If you add a domain to `@remember_me_options` then it won’t remove the remember me cookie unless the options are also passed to `delete_resp_cookie/3`
From the [docs](https://hexdocs.pm/plug/1.19.1/Plug.Conn.html#delete_resp_cookie/3):
"Deleting a cookie requires the same options as to when the cookie was put."


SteffenDE:
Thank you, very good catch!

PR #6608, opened 2026-02-27
fix: config/prod.exs templates

kenichi:
* move `:exclude` under `:force_ssl` to be passed to Plug.SSL

see [comments here](https://github.com/phoenixframework/phoenix/pull/6556/changes#r2865778360)


josevalim:

PR #6619, opened 2026-03-04
Phoenix.Controller: warn when guards refer to unknown actions

novaugust:
I built this to detect dead code in a private codebase, but thought it might be useful higher up. It's possible there's a reason this won't work for all use cases, but I haven't thought of it myself yet =)

This adds a post-compilation step to the Phoenix.Controller.Pipeline that warns when user `plug` guards include actions that the controller doesn't define. The warning isn't pretty at this time. 

```elixir
defmodule MyAppWeb.MyController do
  use Phoenix.Controller, formats: []

  plug :my_plug when action in [:index, :create]
  plug :other_plug when action == :hello
  plug :last_plug when action not in [:another_missing_action]

  def index(conn, _), do: conn
end
```

Would produce a warning like:

```
     warning: Unknown action(s) referenced in MyAppWeb.MyController plug guards: [:hello, :create, :another_missing_action]
     │
 239 │           defmodule MyAppWeb.MyController do
     │           ~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~~
     │
     └─ file/path:1: MyAppWeb.MyController (module)
```
Far from pretty, and not immediately intuitive what the user should do, but just thought I'd open this for discussion before homing in on UX if it is wanted.

I've done no benchmarking to see what impact this has on a codebase, but it's walking only a small set of AST in each controller

### DIFF
--- a/lib/phoenix/controller/pipeline.ex
+++ b/lib/phoenix/controller/pipeline.ex
@@ -11,6 +11,7 @@ defmodule Phoenix.Controller.Pipeline do
 
       Module.register_attribute(__MODULE__, :plugs, accumulate: true)
       @before_compile Phoenix.Controller.Pipeline
+      @after_compile Phoenix.Controller.Pipeline
       @phoenix_fallback :unregistered
 
       @doc false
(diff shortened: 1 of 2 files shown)


josevalim:
The only issue is that we have an `action/3` callback that could dispatch actions in completely different ways, breaking the assumptions here. :( Perhaps we can do as long as action/3 is not overridden?

novaugust:
> The only issue is that we have an `action/3` callback that could dispatch actions in completely different ways, breaking the assumptions here.

ohhhh right, i'd made sure not to assume the arity of the function guarded for that reason, but it hadn't crossed my mind that folks would use it to dispatch to completely unrelated functions.

i'll see if that scenario is detectable from an after-compile hook; likely not? at which point the whole thing could be moot

Ari4ka (approved):

PR #6625, opened 2026-03-12
Add boolean operator guidance to Elixir usage rules

Vitaly-A-Ivanov:
## Summary

Adds guidance to `usage-rules/elixir.md` for choosing `&&` / `||` versus `and` / `or`.

## What changed

- explains why `&&` and `||` are usually safer in Elixir
- documents when `and` and `or` are still appropriate
- adds examples covering structs, `nil`, and changeset/database-style flows

## Why

Elixir’s `and` and `or` require a boolean left-hand side, which can raise when used with values like structs or `nil`. This rule makes that distinction explicit and gives contributors a consistent default.

## Notes

- docs-only change
- no runtime or test changes

### DIFF
--- a/usage-rules/elixir.md
+++ b/usage-rules/elixir.md
@@ -1,5 +1,65 @@
 ## Elixir guidelines
 
+- Boolean operators: `and` / `or` vs `&&` / `||`
+
+  Prefer using `&&` and `||` in most situations.
+
+  The operators `and` / `or` require the **left-hand side to be a boolean** (`true` or `false`). If the left side is not a boolean, Elixir raises an `ArgumentError`.
+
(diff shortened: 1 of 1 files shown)


Vitaly-A-Ivanov:
Added this to make the `and`/`or` vs `&&`/`||` distinction explicit in the Elixir usage rules, with examples for the common `nil`/struct case. Happy to adjust the wording or trim the examples if you’d prefer it shorter.

josevalim:
Thank you but our general guidance with the AGENTS.md is that we only add rules that the Phoenix team runs into. This is because if we were to collect everyone's input, even though they are appreciated, they would quickly run out of control.

As most things generated by Phoenix, they are a starting point, so please expand them based on your needs! Thanks!

PR #6628, opened 2026-03-16
Add missing container in the settings_live.ex

waxave:
## Description

Adds the missing container in `settings_live.ex`.

This container is required for the settings view layout to render correctly and maintain consistency with the rest of the LiveView pages.

## Changes

* Added the missing container wrapper in `settings_live.ex`


SteffenDE:
I believe this is intentional. The settings view doesn't need to be as constrained as the login/register pages.

waxave:
You are right, i just made a fresh project to verify if the setting view was malformed and it wasn't, sorry for that.

PR #6630, opened 2026-03-19
Add interactive mode for phx.new

gitkumi:
Every time I initialize a Phoenix project, I have to triple check the docs to figure out which flags to use. It would be nice if there were an interactive mode.

I made it sort of opinionated and not 1:1 with all the flags. Let me know what you guys think.

https://github.com/user-attachments/assets/9eba74b6-1015-47e4-8710-36282ca1163f

### DIFF
--- a/installer/lib/phx_new/interactive.ex
+++ b/installer/lib/phx_new/interactive.ex
@@ -11,10 +11,8 @@ defmodule Phx.New.Interactive do
 
   @web_options [
     {"live", "LiveView"},
-    {"html", "HTML (no LiveView)"},
-    {"live_no_assets", "LiveView (no esbuild, no Tailwind)"},
-    {"html_no_assets", "HTML (no LiveView, no esbuild, no Tailwind)"},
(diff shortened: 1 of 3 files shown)


gitkumi:
@josevalim thanks for the review! I've addressed the feedback.

https://github.com/user-attachments/assets/eb27502f-f59f-4e2f-8c7d-65d111eff0d2

installer/lib/phx_new/interactive.ex
```diff
@@ -0,0 +1,225 @@
...
+        abort()
+
+      result when is_binary(result) ->
+        result
+        |> String.trim()
+        |> normalize_abort_input()
+    end
+  end
+
+  defp normalize_abort_input(input), do: input
```
SteffenDE: Should that do something?

gitkumi: nope, removed! 🙏

SteffenDE:
🙌

PR #6631, opened 2026-03-20
Fix outdated flash component example in docs

### LINKED ISSUE #6624: Should the flash example be corrected?
Hi,

I was confused by one of the `flash` component usage examples: 

```elixir
 <.flash kind={:info} phx-mounted={show("#flash")}>Welcome Back!</.flash
```

Location: https://github.com/phoenixframework/phoenix/blob/850162d618cf4e31be82fb53ad6e259095fa6126/installer/templates/phx_web/components/core_components.ex#L40

I assume that `phx-mounted` is there to make the flash show up on load. If that's the correct assumption, then I can see two issues with it: 
- the flash still appears even without `phx-mounted={show("#flash")`, because of `:if={msg = render_slot(@inner_block) || Phoenix.Flash.get(@flash, @kind)}` and the classes don't contain `hidden` - ` class="toast toast-top toast-end z-50"` 
- the default id includes flash kind: ` assigns = assign_new(assigns, :id, fn -> "flash-#{assigns.kind}" end)`, so  `show("#flash")` won't match `#flash-info`

Can you please advise what am I missing?

eagle-head:
The flash component `@doc` example used `show("#flash")` which had two issues:

1. The id `#flash` does not match the default generated id (`flash-#{kind}`)
2. The flash was missing the `hidden` attribute, so `phx-mounted` had no visible effect

Updated the example to follow the same pattern used by the `client-error` and
`server-error` flashes in the generated layouts: explicit `id`, `hidden` attribute,
and `show("#id") |> JS.remove_attribute("hidden")`.

Closes #6624

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex.eex
+++ b/installer/templates/phx_web/components/core_components.ex.eex
@@ -37,7 +37,14 @@ defmodule <%= @web_namespace %>.CoreComponents do
   ## Examples
 
       <.flash kind={:info} flash={@flash} />
-      <.flash kind={:info} phx-mounted={show("#flash")}>Welcome Back!</.flash>
+      <.flash
+        id="welcome-back"
+        kind={:info}
+        phx-mounted={show("#welcome-back") |> JS.remove_attribute("hidden")}
+        hidden
+      >
+        Welcome Back!
+      </.flash>
   """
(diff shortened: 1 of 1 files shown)


SteffenDE:
🙌

PR #6635, opened 2026-03-20
Don't pass type: :id to scope references in generated migrations

### LINKED ISSUE #6422: Scope column type for MySQL
https://elixirforum.com/t/migrations-with-mysql/72097/1

eagle-head:
## Summary

Fixes #6422.

The scope reference in the migration template (`migration.exs.eex`, line 10) was passing `type: :id` explicitly to `references()`. In Ecto's MySQL adapter, `:id` maps to `integer` (signed 32-bit), while the default PK is `bigint unsigned auto_increment`. This type mismatch causes:

```
(MyXQL.Error) (1005) (ER_CANT_CREATE_TABLE) Can't create table (errno: 150 "Foreign key constraint is incorrectly formed")
```

### Ecto type mapping on MySQL

| `references()` option | Ecto type | MySQL SQL | Matches default PK? |
|---|---|---|---|
| *(no type)* | `:bigserial` (default) | `BIGINT UNSIGNED` | Yes |
| `type: :bigserial` | `:bigserial` | `BIGINT UNSIGNED` | Yes |
| `type: :id` | `:id` | `integer` (32-bit) | **No — errno 150** |

### Before (broken on MySQL)

```elixir
add :user_id, references(:users, type: :id, on_delete: :delete_all)
```

### After (works on all adapters)

```elixir
add :user_id, references(:users, on_delete: :delete_all)
```

By omitting `type:` when it's `:id`, Ecto defaults to `:bigserial` which correctly maps to `BIGINT UNSIGNED` on MySQL. Non-`:id` types like `:binary_id` are still passed explicitly.

### Changes

- `priv/templates/phx.gen.schema/migration.exs.eex` — conditionally omit `type:` for scope references when `schema_migration_type == :id`
- Unit tests for both `:id` (omitted) and `:binary_id` (kept) scope types
- MySQL integration test with scoped resources verifying no errno 150

### DIFF
--- a/integration_test/test/code_generation/app_with_mysql_adapter_test.exs
+++ b/integration_test/test/code_generation/app_with_mysql_adapter_test.exs
@@ -8,7 +8,10 @@ defmodule Phoenix.Integration.CodeGeneration.AppWithMySqlAdapterTest do
         {app_root_path, _} =
           generate_phoenix_app(tmp_dir, "default_mysql_app", ["--database", "mysql"])
 
-        mix_run!(~w(phx.gen.html Blog Post posts title body:string status:enum:unpublished:published:deleted), app_root_path)
+        mix_run!(
+          ~w(phx.gen.html Blog Post posts title body:string status:enum:unpublished:published:deleted),
+          app_root_path
+        )
 
         modify_file(Path.join(app_root_path, "lib/default_mysql_app_web/router.ex"), fn file ->
           inject_before_final_end(file, """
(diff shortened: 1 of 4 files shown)


SteffenDE:
Would it also work if we just don't set a `schema_migration_type` by default?

https://github.com/phoenixframework/phoenix/blob/e3a166786af7540c685a0a8b085279dd0e2c3016/lib/mix/phoenix/scope.ex#L37

eagle-head:
Good call — pushed a refactor. `schema_migration_type` is now `nil` when `schema_type` is `:id`, and the template checks for non-nil instead of `!= :id`. The `:binary_id` case still propagates via the `schema_type` fallback (only `:id` is filtered out).

Ari4ka (approved):

SteffenDE:
🙌

PR #6657, opened 2026-04-16
fix(phx.gen.auth): delete old token when reissuing session

otlaitil:
Fixes a bug where reissuing user session token leaves the old token in the database. This bloats the database table and poses a minor security risk as old token will remain valid.

The fix copies the pattern from logout to [clear the token](https://github.com/phoenixframework/phoenix/blob/c88da921ef8009e0aca8643d5334585cd35d5573/priv/templates/phx.gen.auth/auth.ex.eex#L50) 

## Steps to reproduce

1. Run auth scaffold `mix phx.gen.auth Accounts User users`
2. In `user_auth.ex` set `@session_reissue_age_in_days 0`
3. Register a user and log in
4. Check tokens with `SELECT * FROM users_tokens;`
5. Refresh the page
6. Check tokens again (tokens accumulate)

### DIFF
--- a/priv/templates/phx.gen.auth/auth.ex.eex
+++ b/priv/templates/phx.gen.auth/auth.ex.eex
@@ -94,6 +94,8 @@ defmodule <%= inspect auth_module %> do
     token_age = <%= inspect datetime_module %>.diff(<%= datetime_now %>, token_inserted_at, :day)
 
     if token_age >= @session_reissue_age_in_days do
+      old_token = get_session(conn, :<%= schema.singular %>_token)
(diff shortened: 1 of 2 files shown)


josevalim:
Thanks! However, this can lead to concurrency issues: two requests happen at the same time, but due to delays, one wins and deletes the token, causing the other one to fail. We already limit tokens to 14 days, which is exactly to address the security issue mentioned in the description. So I don't believe we should merge this. If you are concerned about issues being too long, then reduce the token limit to less than 14 days.

otlaitil:
@josevalim thanks, the issue with multiple requests is a fair point - closing the PR.

PR #6658, opened 2026-04-17
fix(docs): add build path to nodePaths for colocated hooks

benvp:
Since the introduction of colocated hooks in v1.8 the example for a custom `build.js` file requires the build path to be present as well. If it is not present, the build of the assets will fail with a resolve error `Could not resolve "phoenix-colocated/<app>"`

The default `:esbuild` config already includes the build path by using `Mix.project.build_path()`

### DIFF
--- a/guides/asset_management.md
+++ b/guides/asset_management.md
@@ -143,7 +143,7 @@ let opts = {
   target: "es2022",
   outdir: "../priv/static/assets",
   external: ["*.css", "fonts/*", "images/*"],
-  nodePaths: ["../deps"],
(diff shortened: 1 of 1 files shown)


SteffenDE (commented):
I believe for that to work, you'd also need to adjust the watcher to explicitly pass `env: %{"MIX_ENV" => to_string(Mix.env())}` and also pass MIX_ENV in the configured aliases.

Otherwise, unless you manually set `MIX_ENV`, it won't be set in the build script.

PR #6679, opened 2026-05-16
Add optional Volt support to phx.new

dannote:
Adds an opt-in `--volt` flag to `mix phx.new` that generates projects using [Volt](https://hex.pm/packages/volt) instead of the esbuild and tailwind wrappers.

Volt is an Elixir-native asset pipeline — bundles JavaScript, runs Tailwind CSS, and serves assets in dev with HMR, all through NIFs. No Node.js required.

### What changes

- `--volt` flag for `mix phx.new`, `mix phx.new.web`, and the interactive installer
- Generates `{:volt, "~> 0.11.0"}` with matching config, dev watcher, endpoint plug, layout, formatter, and asset aliases
- Works for single-app, umbrella, and multi-web umbrella projects
- Umbrella projects use named Volt config profiles (`config :volt, :my_app_web, [...]`), matching the esbuild/tailwind convention
- Rejects conflicting flags (`--volt --no-esbuild`, `--volt --no-tailwind`)
- Default output unchanged — Volt is only used when explicitly requested

### DIFF
--- a/installer/lib/mix/tasks/phx.new.ex
+++ b/installer/lib/mix/tasks/phx.new.ex
@@ -38,7 +38,7 @@ defmodule Mix.Tasks.Phx.New do
       Please check the adapter docs for more information
       and requirements. Defaults to "bandit".
 
-    * `--no-assets` - equivalent to `--no-esbuild` and `--no-tailwind`
+    * `--no-assets` - equivalent to `--no-esbuild`, `--no-tailwind`, and disables Volt
 
     * `--no-dashboard` - do not include Phoenix.LiveDashboard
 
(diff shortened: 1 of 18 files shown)


SteffenDE:
I don’t think we should add any more options to the generators and rather explore such integrations using igniter. I believe you should be able to use igniter.new to bootstrap a phoenix project and then apply the changes on top.

joshprice:
We've been exploring an igniter based take on Phoenix generators over in https://github.com/team-alembic/phx_install 

The goal was to enable a variety of different bundlers (see https://github.com/team-alembic/phx_install/commit/984f9f2ebfbbf9f843403da7517052fce5565df1).

Would love to see some more options like Volt, Bun, etc

PR #6680, opened 2026-05-17
Add eex suffix to phx.gen.auth template override test

bhtabor:
https://github.com/phoenixframework/phoenix/pull/6598 added eex suffix. This just makes sure phx.gen.auth template override test also uses eex suffix.

### DIFF
--- a/test/mix/tasks/phx.gen.auth_test.exs
+++ b/test/mix/tasks/phx.gen.auth_test.exs
@@ -1442,7 +1442,7 @@ defmodule Mix.Tasks.Phx.Gen.AuthTest do
   test "allows templates to be overridden", config do
     in_tmp_phx_project(config.test, fn ->
       File.mkdir_p!("priv/templates/phx.gen.auth")
-      File.write!("priv/templates/phx.gen.auth/auth.ex", "#it works!")
+      File.write!("priv/templates/phx.gen.auth/auth.ex.eex", "#it works!")
 
       send(self(), {:mix_shell_input, :yes?, false})
 
--- a/test/mix/tasks/phx.gen.auth_test.exs
+++ b/test/mix/tasks/phx.gen.auth_test.exs
@@ -1459,6 +1459,27 @@ defmodule Mix.Tasks.Phx.Gen.AuthTest do
     end)
   end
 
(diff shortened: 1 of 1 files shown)


SteffenDE:
I may be misremembering, but I think I deliberately kept it to check the backwards compatibility. If there’s a different test already verifying that, this is fine though.

bhtabor:
> I may be misremembering, but I think I deliberately kept it to check the backwards compatibility. If there’s a different test already verifying that, this is fine though.

Ohh, didn't know it was deliberate. I have seen the backward compatibility check for missing eex extension. https://github.com/phoenixframework/phoenix/pull/6598/changes#diff-3a04e676440d973e029640b657bc41b45f2f2931a9558c5518a601f8b375f097R42

I happened to have a [custom template generator](https://github.com/bhtabor/phoenix/commit/4bce193259f320b3953446f097af7d3320b84e25) and the test was failing because there was an eex file from the custom generator. Since the test didn't use the eex extension, the custom generator overshadowed the non eex test file.

I thought it was an oversight because a somewhat similar test extension change was made for phx.gen.live. https://github.com/phoenixframework/phoenix/pull/6598/changes#diff-665fb2702d1d485d1c09ac7ce62f5479550934824a26da3d0954eb3cd98bc119R33

SteffenDE:
It would be better to add a dedicated test for it. If you want to do that, feel free to go ahead and add another test block that mentions it‘s testing backwards compat.

test/mix/tasks/phx.gen.auth_test.exs
```diff
@@ -1459,6 +1459,27 @@ defmodule Mix.Tasks.Phx.Gen.AuthTest do
     end)
   end
 
+  test "allows templates to be overridden (backwards compatibility)", config do
+    in_tmp_phx_project(config.test, fn ->
+      File.mkdir_p!("priv/templates/phx.gen.auth")
```
bhtabor, line 1464: I was actually thinking we could make the backward compatibility check generic. Not specific to `phx.gen.auth`.

SteffenDE: We could, but I don’t really care where it is, as long as there’s a check for it somewhere

Issue #6681, opened 2026-05-18
Discrepancy in phx.gen.auth templates results in failed test


strobilomyces:
### Environment

* Elixir version (elixir -v): 1.19.5-otp-28
* Phoenix version (mix deps): 1.8.7
* Operating system: Linux (Manjaro - 7.0.3-1-MANJARO)

### Actual behavior

After running mix phx.gen.auth, there's a mismatch between the generated LiveView templates and the test templates.

In `test/myapp_web/live/user_live/login_test.exs`, line 12, is the following:

`assert html =~ "Register"`

which results in:

`Assertion with =~ failed`

because the live template generated under `/lib/myapp_web/live/user_live/login.ex` contains:

`Sign up` instead of `Register`

### Expected behavior

Test templates from phx.gen.auth should match their corresponding LiveView templates.

### Solution

In`priv/templates/phx.gen.auth/login_live_test.exs.eex`, line 12:

Change

`assert html =~ "Register"`

to

`assert html =~ "Sign up"`

Test passes.

SteffenDE:
Closed via https://github.com/phoenixframework/phoenix/pull/6682.

PR #6682, opened 2026-05-18
Fix phx.auth.gen template

strobilomyces:
Changes the login_live_test template created by phx.gen.auth to match the text it generates for the corresponding login_live template ("Sign up" instead of "Register").

### DIFF
--- a/priv/templates/phx.gen.auth/login_live_test.exs.eex
+++ b/priv/templates/phx.gen.auth/login_live_test.exs.eex
@@ -9,7 +9,7 @@ defmodule <%= inspect context.web_module %>.<%= inspect Module.concat(schema.web
       {:ok, _lv, html} = live(conn, ~p"<%= schema.route_prefix %>/log-in")
 
       assert html =~ "Log in"
-      assert html =~ "Register"
+      assert html =~ "Sign up"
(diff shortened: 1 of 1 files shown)


strobilomyces:
Fixes https://github.com/phoenixframework/phoenix/issues/6681

SteffenDE:
I’ll look into this soon, but it’s weird since we should have a test verifying this: https://github.com/phoenixframework/phoenix/blob/8cbe8172d449fe7b8ab5720a5b448ac1cad36882/integration_test/test/code_generation/app_with_defaults_test.exs#L196

strobilomyces:
I ran this on a fresh project and the tests passed, which was weird because the code in my project was more or less untouched. I was baffled where it was grabbing the "Register" text to pass it until I realized that it's grabbing it from the menu links at the top of the page rather than the main body. Having removed the menu links in my project, that text was no longer there and the test obviously failed.

Personally, it seems a bit odd to have the test make assertions matching text from above in the menu links ("Log in" and "Register", with the former also matching the h1 in the main body) as well as in the main body ("Log in with email"). Whether that's intentional, I don't know.

SteffenDE:
🙌

Issue #6626, opened 2026-03-13
AGENTS.md forbids using `daisyUI` even though Phoenix 1.8 ships with it


javiercr:
Phoenix 1.8 ships with daisyUI by default (as announced here):
[https://www.phoenixframework.org/blog/phoenix-1-8-released](https://www.phoenixframework.org/blog/phoenix-1-8-released)

However, the default `AGENTS.md` generated for new projects tells coding agents to avoid using it:

https://github.com/phoenixframework/phoenix/blob/850162d618cf4e31be82fb53ad6e259095fa6126/installer/templates/usage-rules/assets.md?plain=1#L13

This seems a bit confusing. If daisyUI is included by default, it probably shouldn’t be discouraged in the generated agent instructions. It might make more sense to either allow it or phrase this as a preference rather than a strict rule.

Happy to open a PR if helpful.

josevalim:
>  It might make more sense to either allow it or phrase this as a preference rather than a strict rule.

This was our original approach but we got mixed results. The agent rarely went out of the box and stuck mostly to DaisyUI components. But DaisyUI does not have all possible components, so the end result was worse. As you know, they are not deterministic, and when we tried there wasn't a good way to get the best of both worlds.

TL;DR: when you ask a designer to define the best interface, and use DaisyUI if available, it would only use DaisyUI when needed. When we asked the agent, it picked DaisyUI most of the time.

PR #6685, opened 2026-05-24
docs: updates AGENTS.md from `mix phx.new` regarding DaisyUI

tairosonloa:
The currently generated `AGENTS.md` file is 10 months old, before DaisyUI support was officially added to Phoenix Framework, and it argues against using DaisyUI and instead recommends using Tailwind.

I updated it to prefer DaisyUI components when possible, with a fallback for tailor-made Tailwind components. Also mentioned DaisyUI support is added out of the box.

### DIFF
--- a/installer/templates/usage-rules/assets.md
+++ b/installer/templates/usage-rules/assets.md
@@ -10,7 +10,8 @@
 
 - **Always use and maintain this import syntax** in the app.css file for projects generated with `phx.new`
 - **Never** use `@apply` when writing raw css
-- **Always** manually write your own tailwind-based components instead of using daisyUI for a unique, world-class design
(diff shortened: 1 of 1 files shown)


SteffenDE:
As far as I know this is intentional: https://github.com/phoenixframework/phoenix/issues/6626#issuecomment-4074519753

> The currently generated AGENTS.md file is 10 months old, before DaisyUI support was officially added to Phoenix Framework

And this is wrong. DaisyUI and AGENTS.md were both first shipped in 1.8.0.

tairosonloa:
Hi Steffen,

Didn't know that. Then, I'll close my PR. If any maintainer sees this, feel free to reopen it.

PR #6687, opened 2026-05-26
Fix Elixir 1.20 compilation warnings

gilbertwong96:
Resolve compilation warnings in Elixir 1.20 by updating bitstring binary-size to binary-^size syntax.

### DIFF
--- a/.tool-versions
+++ b/.tool-versions
@@ -1,0 +1,2 @@
+elixir 1.20.0-rc.6
+erlang 27.3.3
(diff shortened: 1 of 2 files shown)


SteffenDE:
1. the warning should already be fixed by https://github.com/phoenixframework/phoenix/commit/f286d6968d9cb4316cb3fa12604dd30a2320e31a
2. your patch is wrong, as you can see by the CI failures
3. you added a unrelated tool-versions file

PR #6688, opened 2026-05-28
Improve wording in request lifecycle guide

Null-logic-0:
Improve clarity in request lifecycle guide


SteffenDE:
I think it is fine as is. Thanks!

PR #6692, opened 2026-06-01
Make TODO raises more descriptive in generators

Null-logic-0:
I think this is more understandable for beginners and less confusing. This is just my opinion.


SteffenDE:
Thank you, I'd prefer to keep it as is though.

PR #6693, opened 2026-06-01
Replace all hexdocs URLs with the subdomain format

Gazler:
See https://github.com/hexpm/hexpm/pull/1614

This is the command to update the URLs:

```sh
git grep -Ilz -E 'https://hexdocs\.pm/[A-Za-z0-9_]+' |
  perl -0ne 'print unless m{\A(?:doc|_build|deps)/}' |
  xargs -0 -r perl -0pi -e 's{https://hexdocs\.pm/([A-Za-z0-9_]+)}{"https://" . ($1 =~ s/_/-/gr) .
  ".hexdocs.pm"}ge'
```

The diff is best viewed with:

```
git diff --word-diff=color --word-diff-regex=.
```


SteffenDE:
Nice! Do you also want to send a PR for LiveView?

PR #6707, opened 2026-06-11
Add security notes for Channel join / handle_in payload

rhcarvalho:
They contain client-controlled data, and must be handled with that in mind.

Phoenix's counterpart to https://github.com/phoenixframework/phoenix_live_view/pull/4284.

See conversation in https://github.com/phoenixframework/phoenix_live_view/pull/4268#issuecomment-4631233965.

### DIFF
--- a/lib/phoenix/channel.ex
+++ b/lib/phoenix/channel.ex
@@ -350,6 +350,12 @@ defmodule Phoenix.Channel do
 
   Payloads are serialized before sending with the configured serializer.
 
+  > #### Security Note {: .warning}
+  >
+  > The `payload` contains untrusted data from the client. You must authorize
+  > and validate this data before using it. See the ["Security"
+  > guide](security.html) for more information.
+
   ## Example
 
       def join("room:lobby", payload, socket) do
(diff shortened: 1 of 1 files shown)


lib/phoenix/channel.ex
```diff
@@ -350,6 +350,12 @@ defmodule Phoenix.Channel do
 
   Payloads are serialized before sending with the configured serializer.
 
+  > #### Security Note {: .warning}
+  >
+  > The `payload` contains untrusted data from the client. You must authorize
+  > and validate this data before using it. See the ["Security"
+  > guide](security.html) for more information.
```
SteffenDE, line 357: should we link to a specific section?

https://phoenix.hexdocs.pm/security.html#broken-access-control

rhcarvalho: 👍 

I'm in the process of writing a more specific section for both Phoenix and LV's security guides.

The existing Broken Access Control section focuses on controllers, and nothing in the existing guide talks about channels.

Let's update the links together with those security guide additions?

SteffenDE: Sounds good to me.

PR #6710, opened 2026-06-12
Add MongoDB database option to Phoenix installer

tevoinea-mdb:
Hi! I'm opening this PR in coordination with https://github.com/elixir-mongo/mongodb_ecto/pull/205 in `mongodb_ecto`, that aims to make using MongoDB as seamless as its SQL counterparts.

This PR adds the `--database mongodb` option and the plumbing necessary so that it behaves like all of the other database options.

### DIFF
--- a/installer/templates/phx_single/docker-compose.yml.eex
+++ b/installer/templates/phx_single/docker-compose.yml.eex
@@ -1,0 +1,16 @@
+services:
+  mongodb:
+    image: mongo:latest
+    command: ["--replSet", "rs0", "--bind_ip_all"]
+    ports:
+      - "27017:27017"
+    environment:
+      MONGO_INITDB_DATABASE: <%= @app_name %>_dev
+    healthcheck:
+      test: >
+        mongosh --quiet --eval
+        'try { rs.status() } catch(e) { rs.initiate({_id:"rs0",members:[{_id:0,host:"127.0.0.1:27017"}]}) }'
+      interval: 5s
+      timeout: 10s
+      retries: 5
+      start_period: 5s
--- a/installer/templates/usage-rules/ecto.md
+++ b/installer/templates/usage-rules/ecto.md
@@ -1,0 +1,11 @@
+## MongoDB
+
+When the project uses `--database mongo` (adapter: `Mongo.Ecto`):
+
+- **Primary keys must be `:binary_id`** — configured automatically via `config :app, :generators, binary_id: true`. All `mix phx.gen.*` commands honour this automatically; no `--binary-id` flag needed.
+- **No SQL joins** — use embedded schemas (`embeds_one`, `embeds_many`) for nested data, or separate queries for associations.
+- **Migrations create collections and indexes**, not tables. `create table(:name)` → MongoDB collection. `add :col, :type` → no-op (schema-less).
+- **Test isolation** uses `Mongo.Ecto.truncate(Repo)` before each test, not `Ecto.Adapters.SQL.Sandbox`.
+- **Connection string** uses `mongo_url:` config key. In production, set `DATABASE_URL` env var to a MongoDB URI (`mongodb://...` or `mongodb+srv://...`).
+- **Atlas** — swap `DATABASE_URL` to your Atlas connection string; no code changes needed.
+- `:decimal` type is not supported by `mongodb_ecto`.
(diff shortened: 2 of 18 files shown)


SteffenDE:
Does this work with the other generators, e.g. `mix phx.gen.auth`?

tevoinea-mdb:
It does, but it depends on that linked `mongodb_ecto` PR in the description to be merged first.

rhcarvalho:
Hi! I wanted to share my experience using Mongo+Phoenix for a few years (at least ~2023-2026).

In short, the experience trying to use it with Ecto was so frustrating to the point I decided it was not worth it.

The existing libraries didn't seem to get much love or use at the time. Eventually, I went with using the driver https://github.com/zookzook/elixir-mongodb-driver, which was the most active driver fork at the time, without the Ecto abstraction (but using Ecto for changesets/forms).

I see that things have since changed in `mongodb_ecto`, with https://github.com/elixir-mongo/mongodb_ecto/pull/190 making the `zookzook` fork the underlying driver.

I would expect to discover several caveats when using MongoDB with Ecto and the Phoenix generators, some of them were already spelled out in the AGENTS.md/ecto usage rules addition.

installer/templates/phx_single/docker-compose.yml.eex
rhcarvalho, line 1: We don't generate a `docker-compose.yml` file for any other choice of database.

I'd expect adding MongoDB wouldn't change that -- it's on the end user to decide how they organize their dev environment (e.g. local MongoDB, Docker container, or external server).

installer/templates/usage-rules/ecto.md
```diff
@@ -0,0 +1,11 @@
+## MongoDB
+
+When the project uses `--database mongo` (adapter: `Mongo.Ecto`):
+
+- **Primary keys must be `:binary_id`** — configured automatically via `config :app, :generators, binary_id: true`. All `mix phx.gen.*` commands honour this automatically; no `--binary-id` flag needed.
+- **No SQL joins** — use embedded schemas (`embeds_one`, `embeds_many`) for nested data, or separate queries for associations.
+- **Migrations create collections and indexes**, not tables. `create table(:name)` → MongoDB collection. `add :col, :type` → no-op (schema-less).
+- **Test isolation** uses `Mongo.Ecto.truncate(Repo)` before each test, not `Ecto.Adapters.SQL.Sandbox`.
```
rhcarvalho, line 8: José recently cleanup agents rules, and one of the updates was to simply improve the scaffolded test setup replacing an agent rule.

I think we could do the same here: scaffold the project with tests setup correctly, no need to instruct agents.

- https://github.com/phoenixframework/phoenix/pull/6775#discussion_r3675959476
- https://github.com/phoenixframework/phoenix/pull/6775/changes#diff-18d16aa8082d308dd51610f633781f608ee168b0379545a48236c5d7f3ac3a12

priv/templates/phx.gen.auth/context_functions.ex.eex
```diff
@@ -176,17 +176,27 @@
   If the token is valid `{<%= schema.singular %>, token_inserted_at}` is returned, otherwise `nil` is returned.
   """
   def get_<%= schema.singular %>_by_session_token(token) do
-    {:ok, query} = <%= inspect schema.alias %>Token.verify_session_token_query(token)
-    Repo.one(query)
+    {:ok, query} = <%= inspect schema.alias %>Token.verify_session_token_query(token)<%= if mongo_adapter? do %>
+    case Repo.one(query) do
+      nil -> nil
+      token_record ->
+        <%= schema.singular %> = Repo.get(<%= inspect schema.alias %>, token_record.<%= schema.singular %>_id)
```
rhcarvalho, line 183: Oh, noticing this does 2 queries for getting a session token. Perhaps because of the "no joins" limitation?

Considering there is no transaction here, I think we need to understand the security implications of that data race before promoting it.

rhcarvalho (commented):
Review Notes:

- This requires carving out several exceptions/conditional behavior in the templates. Not sure if we cover it all, including other `mix phx.gen.*` tasks.
- Missing integration tests

josevalim:
I also think we should close the gaps in Ecto compatibility rather than adding conditionals to Phoenix. For example, Mongo could accept references and document they have no effect and defauilt to binary_id

SteffenDE:
I appreciate the effort to make using MongoDB with Phoenix easier, but realistically I don't see us merging this as it would require that we also maintain it. Maintaining the existing SQL adapters is not too much effort since they all work mostly the same and don't need specific code changes in generators etc. If it was as simple as changing the adapter, it would be a different discussion though.

Even if we don't merge this, I think a good way forward could be to provide an Igniter-based installer that you then run on top of `phx.new --no-ecto` (or even on a default project, replacing Postgres).

PR #5957, opened 2024-10-22
require Elixir 1.15+ in new apps

SteffenDE:
Closes #5276.

PR #6713, opened 2026-06-12
installer: manage daisyui via mix.exs

rhcarvalho:
Similar to how we replaced vendored Heroicons with a Git/GitHub Mix dependency, replace the vendored daisyUI with a mix dependency.

This offers a few benefits:
- Improved supply chain awareness (easily verify which version is in use)
- Easier updates by bumping a tag in mix.exs
- Avoids bloating the installer templates, new apps, Docker images, etc, with large static files

See:
- https://github.com/saadeghi/daisyui/discussions/3749
- https://github.com/phoenixframework/phoenix/issues/5622


rhcarvalho:
> Error: No space left on device : '/home/runner/actions-runner/extracted/_diag/pages/d2c730dc-5517-468a-a7ce-970d6630157c_33d64cf1-7383-5635-8374-a01037543bdb_1.log'

https://github.com/phoenixframework/phoenix/actions/runs/27445073432/job/81128074157?pr=6713#step:4:1211

Ooops?! 🤷

SteffenDE:
Yeah, looks like the daisyUI repo is **very large** and the older version doesn't support shallow checkouts?

rhcarvalho:
Oh, so that would explain the failure! How unfortunate, save disk space in one move, use lots of disk space elsewhere.

While Phoenix itself supports older versions, does the installer (`mix phx_new`) has to operate on the same version constraints?

I don't have the data, but would imagine all the incentives are to start a greenfield project on a more recent version (1.18-1.20).

rhcarvalho:
Shallow clones is from 1.17:

"[mix deps] Add :depth option to Mix.SCM.Git, thus supporting shallow clones of Git dependencies"
 https://hexdocs.pm/elixir/changelog.html#:~:text=%5Bmix%20deps%5D%20Add%20%3Adepth%20option%20to%20Mix.SCM.Git%2C%20thus%20supporting%20shallow%20clones%20of%20Git%20dependencies

https://mix.hexdocs.pm/1.20.1/Mix.Tasks.Deps.html#module-git-options-git

Sparse checkouts much earlier.

rhcarvalho:
https://github.com/phoenixframework/phoenix/blob/811a8492f90755e4ed90c4a41bc35cda4505bd93/mix.exs#L14-L17

rhcarvalho:
@SteffenDE https://github.com/saadeghi/daisyui/issues/3834, https://github.com/saadeghi/daisyui/issues/2997, the large repo size is a known "issue". I didn't recall it.

One way around would be a fork, or fork + github action to pull updates, but then we're defeating the ability to point straight upstream instead of maintaining a level of indirection. At that point we could have an auto-published Hex package as well.

rhcarvalho:
For reference, the requirement of 1.15+ in the installer dates back to Oct 2024 (https://github.com/phoenixframework/phoenix/pull/5957)

SteffenDE:
Yeah, it’s probably fine to bump the requirement for new apps.

cc @josevalim

josevalim:
Yes, it is fine to bump it for new apps!

rhcarvalho:
@SteffenDE @josevalim rebased on top of main (to pick up recent CI changes) and bumped the Elixir version requirement for the installer and new apps in a separate commit for easier review. The patch in the original commit has not changed.

Unfortunately I think in https://github.com/phoenixframework/phoenix/actions/runs/27549626721/job/81432108238?pr=6713 we ran into https://github.com/elixir-lang/elixir/issues/14425, untracked files in CI cache affecting `mix deps.get`. A fix for that was released in Elixir 1.19.

Any suggestions? Both sound non-ideal to me:

1. Manually `rm -rf deps/daisyui` in the CI job before running `mix deps.get`
2. Requiring Elixir v1.19

If nothing else, I'd pick 1 with a note to remove the hack once we require Elixir v1.19+.

josevalim:
1 sounds good!

rhcarvalho:
> Unfortunately I think in https://github.com/phoenixframework/phoenix/actions/runs/27549626721/job/81432108238?pr=6713 we ran into [elixir-lang/elixir#14425](https://github.com/elixir-lang/elixir/issues/14425), untracked files in CI cache affecting `mix deps.get`. A fix for that was released in Elixir 1.19.

Sorry, I made a mistake. It turns out I let the wrong daisyui commit hash slip into `phoenix/integration_test/mix.lock` (I'm intentionally sticking to v5.5.20 and not bumping it as part of this change (current latest is v5.5.23).

The deps hack (https://github.com/phoenixframework/phoenix/commit/c85719486e28dd5c97b2278e46a1e8ff7e1f5541) might not be necessary.

rhcarvalho:
This is ready for a final review. It's the original implementation + bumping version requirement for installer and new apps.

installer/templates/phx_assets/app.css.eex
```diff
@@ -13,18 +13,15 @@
    The heroicons installation itself is managed by your mix.exs */
 @plugin "../vendor/heroicons";
 
-/* daisyUI Tailwind Plugin. You can update this file by fetching the latest version with:
-   curl -sLO https://github.com/saadeghi/daisyui/releases/latest/download/daisyui.js
-   Make sure to look at the daisyUI changelog: https://daisyui.com/docs/changelog/ */
-@plugin "../vendor/daisyui" {
+/* daisyUI Tailwind Plugin. */
+@plugin "<%= if @in_umbrella do %>../../<% end %>../../deps/daisyui/packages/bundle/daisyui" {
```
SteffenDE: Can you test if this works too?

```suggestion
@plugin "daisyui/packages/bundle/daisyui" {
```

We set the `NODE_PATH` now (for colocated CSS), but it should also resolve in `deps` automatically.

SteffenDE (approved):

installer/templates/phx_assets/app.css.eex
rhcarvalho: Oh yeah, it works and makes it less of an eyesore :) Thanks for calling it out!

Updated.

SteffenDE:
Very nice! 🙌️

PR #6715, opened 2026-06-12
Fix xref exclude deprecation (Elixir 1.20)

rhcarvalho:
Use elixirc_options: [no_warn_undefined: ...] instead of xref: [exclude: ...]

The `xref: [exclude: ...]` project option has been deprecated in Elixir 1.20 in favor of `elixirc_options: [no_warn_undefined: ...]`. This silences the deprecation warning emitted by `mix compile` while preserving the existing behaviour of suppressing undefined-module warnings for optional/dynamic dependencies.

### DIFF
--- a/mix.exs
+++ b/mix.exs
@@ -24,8 +24,8 @@ defmodule Phoenix.MixProject do
       deps: deps(),
       package: package(),
       consolidate_protocols: Mix.env() != :test,
-      xref: [
-        exclude: [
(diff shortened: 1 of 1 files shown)


rhcarvalho:
This is equivalent to phoenixframework/phoenix_live_view#4281.

SteffenDE:
🙌️

PR #6719, opened 2026-06-14
presence: Fix get_by_key documentation and typespec inconsistencies

### LINKED ISSUE #5759: `Phoenix.Presence.get_by_key/2` typespec, return values, and documentation are inconsistent

wondr-wclabs:
Fixes #5759 and #4514. The `get_by_key/2` function was incorrectly documented as returning a list of maps, and the typespec reflected this, but the implementation actually returns a map (or an empty list if not found). This PR updates the typespec and example in the documentation to accurately match the implementation without introducing breaking changes.

### DIFF
--- a/lib/phoenix/presence.ex
+++ b/lib/phoenix/presence.ex
@@ -312,12 +312,12 @@ defmodule Phoenix.Presence do
   devices, could return:
 
       iex> MyPresence.get_by_key("room:1", "user1")
-      [%{name: "User 1", metas: [%{device: "Desktop"}, %{device: "Mobile"}]}]
+      %{name: "User 1", metas: [%{device: "Desktop"}, %{device: "Mobile"}]}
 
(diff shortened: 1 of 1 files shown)


SteffenDE:
This strongly feels like an automated contribution. Closing together with the other PRs.

PR #6720, opened 2026-06-14
Raise on pipe_through if route or scope previously defined in scope

### LINKED ISSUE #6139: pipe_through pipeline should always be first in scope

wondr-wclabs:
Fixes #6139. This ensures `pipe_through` must be placed before any routes or sub-scopes within a scope block, preventing confusing behavior where a pipeline is only applied to routes declared after it in the same scope.

### DIFF
--- a/lib/phoenix/presence.ex
+++ b/lib/phoenix/presence.ex
@@ -312,12 +312,12 @@ defmodule Phoenix.Presence do
   devices, could return:
 
       iex> MyPresence.get_by_key("room:1", "user1")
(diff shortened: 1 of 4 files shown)


SteffenDE:
This would be a breaking chance, so it should warm about being deprecated. Also, it’s only about routes, not scopes.

SteffenDE:
Also, remove unrelated commits please

SteffenDE:
This strongly feels like an automated contribution. Since the user is unresponsive, I'm closing this.

PR #6721, opened 2026-06-14
phx.gen.release: Document new Bob Web UI

rhcarvalho:
A frequent pain point with phx.gen.release and updating image versions is that Docker Hub's UI is slow and the API often times out. Search for relevant images is far from ideal.

Now the Hex team has given us a new tool, a read-only Web UI that gives us insight on the latest images without depending on Docker Hub.

In updating the Dockerfile header commentary, replace reference to pkgs.org (filled with ads) with the official Debian package index.

See https://github.com/hexpm/bob/issues/185 and https://github.com/hexpm/bob/pull/256.


SteffenDE:
I'll update the phx.gen.release script to use bob when/if https://github.com/hexpm/bob/pull/272 lands!

PR #6722, opened 2026-06-15
Move generated live_reload patterns to runtime config

jechol:
`phx.new` currently generates `live_reload` patterns in `config/dev.exs`.

Those patterns are runtime-only endpoint configuration, read by `Phoenix.LiveReloader` from the endpoint config when the live reload socket joins. They do not need to be part of compile-time configuration.

Keeping them in compile config can lead to unnecessary recompilation because Mix stores compile config in the Elixir compiler manifest, and equivalent regex literals are not term-equal across evaluations:

```elixir
~r/foo/ == ~r/foo/
#=> false
```

If a config file is touched, Mix may re-check compile config and see the newly evaluated regex values as changed even when the config source did not meaningfully change.

This moves the generated `live_reload` config to `runtime.exs` under the development environment, preserving behavior while avoiding unstable regex terms in compile config.

### DIFF
--- a/installer/templates/phx_single/config/dev.exs.eex
+++ b/installer/templates/phx_single/config/dev.exs.eex
@@ -44,22 +44,7 @@ config :<%= @app_name %>, <%= @endpoint_module %>,<%= if @inside_docker_env? do 
 #
 # If desired, both `http:` and `https:` keys can be
 # configured to run both http and https servers on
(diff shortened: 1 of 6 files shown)


josevalim:
:green_heart: :blue_heart: :purple_heart: :yellow_heart: :heart: great job debugging this!

PR #6724, opened 2026-06-15
Fix running installer tests twice for Elixir 1.19

rhcarvalho:
Both the integration test script and the ci.yml mix_test job were responsible for running installer tests, such that the same tests were run twice for Elixir 1.19.

Since running the tests as part of a job provides better visibility, remove the installer test execution from the integration test script and opt-in to run the installer tests also on Elixir 1.15 as part of the mix_test job (previously was run by the test.sh script).

---

Extracted from bumping requirement for new apps (1.15->1.17) https://github.com/phoenixframework/phoenix/pull/6713#issuecomment-4705690380, so as to keep the diff focused and easier to review.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -16,6 +16,7 @@ jobs:
         include:
           - elixir: "1.15.8"
             otp: "25.3.2.21"
+            installer: true
 
           - elixir: "1.18.4"
             otp: "27.3.4.13"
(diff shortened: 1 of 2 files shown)


rhcarvalho:
@SteffenDE alternatively I'm considering a separate `installer_test` job, such that the version matrix for installer tests can be independent from the matrix for `mix_test`. WDYT?

PR #6725, opened 2026-06-15
Refactor CI workflow to isolate installer tests

rhcarvalho:
Separate installer tests out of the `mix_test` job and the `integration_test/test.sh` script (`integration-test-elixir` job), such that the Elixir version matrix for the installer tests can evolve independent of the matrix of supported versions for Phoenix itself as a library.

Notes:
- Fixes running the installer unit tests twice for Elixir 1.19 during CI: previously they were executed both as part of the main `mix_test` job (when matrix.installer was true) and also inside the `integration-test-elixir` job via `test.sh`.
- Update `integration_test/test.sh` to resolve the correct directory based on the script path, so that running the script works as expected regardless of the current working directory.
- Simplifies `test/test_helper.exs` exclusions, defaulting to always skipping `:mix_phx_new` integration tests, unless explicitly enabled (`--only mix_phx_new`) in the CI `installer_test` job. These tests are now run for all Elixir versions in the `installer_test` job matrix.

---

Alternative to / closes #6724.

This would allow us to bump the Elixir version requirement for the installer without changing the supported Elixir versions for Phoenix https://github.com/phoenixframework/phoenix/pull/6713#issuecomment-4705690380.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -23,7 +23,6 @@ jobs:
           - elixir: "1.19.5"
             otp: "28.5.0.2"
             lint: true
-            installer: true
 
     runs-on: ubuntu-24.04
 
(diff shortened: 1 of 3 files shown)


SteffenDE:
Nice!

PR #6728, opened 2026-06-15
Fix build phoenix.js instructions

rhcarvalho:
Running `mix assets.build` is enough. Installing npm dependencies is not required, as there are only dev dependencies that do not affect `esbuild` bundler calls.


SteffenDE:
🙌️

PR #6729, opened 2026-06-15
Use npm ci to build JS docs

rhcarvalho:
This ensures we're always using the pinned versions from package-lock.json.

Since `mix docs` takes care of installing npm dependencies, we don't need that step in the README.md.


SteffenDE:
🙌️

Issue #6706, opened 2026-06-11
Tailwind build fails in 1.8.8


toupeira:
### Environment

* Elixir version (elixir -v):

```
Erlang/OTP 29 [erts-17.0.1] [source] [64-bit] [smp:4:4] [ds:4:4:10] [async-threads:1] [jit:ns]

Elixir 1.20.0 (compiled with Erlang/OTP 29)
```

* Phoenix version (mix deps):

```
* bandit 1.12.0 (Hex package) (mix)
  locked at 1.12.0 (bandit) 45dac82d
  ok
* dns_cluster 0.2.0 (Hex package) (mix)
  locked at 0.2.0 (dns_cluster) ba6f1893
  ok
* esbuild 0.10.0 (Hex package) (mix)
  locked at 0.10.0 (esbuild) 468489cd
  ok
* expo 1.1.1 (Hex package) (mix)
  locked at 1.1.1 (expo) 5fb308b9
  ok
* file_system 1.1.1 (Hex package) (mix)
  locked at 1.1.1 (file_system) 7a15ff97
  ok
* finch 0.22.0 (Hex package) (mix)
  locked at 0.22.0 (finch) b94e83c4
  ok
* gettext 1.0.2 (Hex package) (mix)
  locked at 1.0.2 (gettext) eab80550
  ok
* heroicons (https://github.com/tailwindlabs/heroicons.git - v2.2.0)
  locked at 0435d4c (tag: v2.2.0)
  ok
* hpax 1.0.3 (Hex package) (mix)
  locked at 1.0.3 (hpax) 8eab6e1c
  ok
* idna 7.1.0 (Hex package) (rebar3)
  locked at 7.1.0 (idna) 6ae959a0
  ok
* jason 1.4.5 (Hex package) (mix)
  locked at 1.4.5 (jason) b0c82399
  ok
* mime 2.0.7 (Hex package) (mix)
  locked at 2.0.7 (mime) 6171188e
  ok
* mint 1.9.0 (Hex package) (mix)
  locked at 1.9.0 (mint) 007154c7
  ok
* nimble_options 1.1.1 (Hex package) (mix)
  locked at 1.1.1 (nimble_options) 821b2470
  ok
* nimble_pool 1.1.0 (Hex package) (mix)
  locked at 1.1.0 (nimble_pool) af2e4e6b
  ok
* phoenix 1.8.8 (Hex package) (mix)
  locked at 1.8.8 (phoenix) f0c84303
  ok
[...]

PR #6731, opened 2026-06-15
Explicitly require :tailwind >= 0.5 for ColocatedCSS

rhcarvalho:
ColocatedCSS new in 1.8.8 requires passing env as a list to phoenix/tailwind config, supported from version 0.5.

To prevent new apps from going into an unsupported state, explicitly bump the version range to require at least 0.5.

See #6706 and https://github.com/phoenixframework/phoenix_live_view/issues/4295.


SteffenDE:
🙌🏻

PR #6733, opened 2026-06-16
Fix tiny typo in phx.gen.auth scope.ex.eex

benjreinhart:
It should be "accessed".  I made one other tiny change while I was in here that I believe reads better.


SteffenDE:
Thank you!

PR #6737, opened 2026-06-25
Add support for the HTTP QUERY method (RFC 10008)

psolvy:
## Summary

Adds support for the new **HTTP QUERY** method, standardized in [RFC 10008](https://www.rfc-editor.org/rfc/rfc10008.html) (Proposed Standard, June 2026).

QUERY is **safe** and **idempotent** like `GET`, but carries a **request body** like `POST` — the request content (of any media type) defines the query. Because it is idempotent, a QUERY can be safely retried, unlike `POST`.

## Changes

- **`Phoenix.Router`** — add `:query` to the supported HTTP verbs, generating the `query/3` and `query/4` route macros (mirroring `get`, `post`, etc.). Routing arbitrary verbs already worked via `match`; this adds the ergonomic, discoverable verb macro.
- **`Phoenix.ConnTest`** — add the matching `query/3` test helper, for parity with the other verbs.
- **`guides/routing.md`** — document QUERY among the supported HTTP verbs.
- **`mix.exs`** — raise the minimum Plug requirement to `~> 1.21`, the release that parses QUERY request bodies into `conn.body_params` and treats QUERY as a safe method in `Plug.CSRFProtection` (elixir-plug/plug#1319). Because this raises a dependency floor and adds a new public macro, the CHANGELOG targets a **1.9.0** (minor) release.

## References

- [RFC 10008: The HTTP QUERY Method](https://www.rfc-editor.org/rfc/rfc10008.html)
- elixir-plug/plug#1319 — companion Plug support (routing macro, body parsing, CSRF), merged


SteffenDE:
I assume we need to wait for an updated plug and then bump the minimum plug version?

psolvy:
@SteffenDE hello!
Yes - you are right. I have already opened a Plug library PR to support a new method
https://github.com/elixir-plug/plug/pull/1319

UPD updated this one with a minimum Plug version and bumped phoenix minor version instead of fix one

SteffenDE:
@psolvy please update again, we moved main to 1.9 now, so the changelog has a conflict :)

@josevalim current Plug still allows Elixir 1.15, so if we want to backport, do you see a reason we can't require plug 1.21 for Phoenix 1.8?

psolvy:
@SteffenDE done!

PS as i can see query changes are in the `1.21.0-dev` right now, not in the `1.21.0`

https://github.com/elixir-plug/plug/commit/e64e37c5864ebe495936932faf644481ff89bf2f

SteffenDE:
Perfect, thank you. Yes, Plug 1.21 is not out yet.

yordis:
@SteffenDE any opportunity to get this one to hex?

josevalim:
As per above, it needs a new plug release and even then I recommend using Phoenix main as this will be part of 1.9 which won’t be out immediately

yordis:
@josevalim any ETA for that? I am working in greenfield, I am not in a rush, but trying to start the codebase based on QUERY for the listing operations.

I hacked around for now.

PR #6738, opened 2026-06-26
proof of concept for channel adoption

SteffenDE:


lib/phoenix/channel/adopt_standin.ex
```diff
@@ -0,0 +1,19 @@
...
+  use Task
+
+  def start_link(pid) do
+    Task.start_link(__MODULE__, :link, [pid])
+  end
+
+  def link(pid) do
+    Process.link(pid)
+    Process.sleep(:infinity)
+  end
```
josevalim, line 18: Maybe we should have Channel.spawn as you said so we don't need the stand-in? Alternatively we tell users we need to spawn processes in a supervisor in their application supervision tree... but then, once again, we might as well use the existing pool supervisor.

PR #6739, opened 2026-06-29
Split routes by verb during compilation

josevalim:
This pull request groups compiled routes by verb, which speeds up compilation times for large routers up to 3x. A side-effect is that `match/forward` routes are always matched last.

Therefore, this pull request changes the semantics of *dead routes*. If you had this code:

    match "/foo"
    get "/foo"

The second route would never be matched but it is now as part of this pull request. However, `get "/foo"` should not be there in the first place. It is effectively a dead route.

Currently WIP. We should either document this pitfall or put this change behind an option. I am fine with documenting this in the CHANGELOG as the code above should either not exist or be correct anyway.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,9 +2,9 @@
 
 ## v1.9.0-dev
 
-### Deprecation
+### Enhancements
 
-  * [Phoenix.Router] Phoenix will now group your routes per verb during compilation when all `match :*` and `forward` routes are defined at the end of the router. This improves compilation times for large routers with no performance cost at runtime. If you have a route with an explicit verb after a `match :*` or `forward`, Phoenix will preserve the previous ordered matching semantics and emit a warning
+  * [Phoenix.Router] Add `use Phoenix.Router, group_by: :verb` to group routes per verb during compilation. This can improve compilation times for large routers with no performance cost at runtime. When enabled, all `match :*` and `forward` routes must be defined at the end of the router, otherwise compilation fails with a list of violations
 
 ## v1.8
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -2,7 +2,7 @@
 
 ## v1.9.0-dev
 
-### Potential breaking changes
+### Deprecation
 
   * [Phoenix.Router] Phoenix will now group your routes per verb during compilation when all `match :*` and `forward` routes are defined at the end of the router. This improves compilation times for large routers with no performance cost at runtime. If you have a route with an explicit verb after a `match :*` or `forward`, Phoenix will preserve the previous ordered matching semantics and emit a warning
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,7 +9,7 @@
       match :*, "/foo"
       get "/foo"
 
-  The second route would never be matched but it will now be as part of this pull request. However, note that `get "/foo"` should not exist in the first place: it should be either removed or moved first. We recommend checking for any `match :*` in your router and moving them to the end of the file.
+  The second route would never be matched but it will now be as part of this change. However, note that `get "/foo"` should not exist in the first place: it should be either removed or moved first. We recommend checking for any `match :*` in your router and moving them to the end of the file.
 
 ## v1.8
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,10 +1,6 @@
 # Changelog for v1.9
 
-## v1.9.0-dev
-
-### Enhancements
-
-  * [Phoenix.Router] Add `use Phoenix.Router, group_by: :verb` to group routes per verb during compilation. This can improve compilation times for large routers with no performance cost at runtime. When enabled, all `match :*` and `forward` routes must be defined at the end of the router, otherwise compilation fails with a list of violations
+Nothing, so far.
 
 ## v1.8
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,15 @@
 # Changelog for v1.9
 
-Nothing, so far.
+## v1.9.0-dev
+
+### Potential breaking changes
+
+  * [Phoenix.Router] Phoenix will now group your routes per verb during compilation. This improves compilation times for large routers with no performance cost at runtime. However, this change implies that wildcard routes are always matched last, which changes the semantics of dead routes. If you had this code:
+
+      match :*, "/foo"
+      get "/foo"
+
+  The second route would never be matched but it will now be as part of this pull request. However, note that `get "/foo"` should not exist in the first place: it should be either removed or moved first. We recommend checking for any `match :*` in your router and moving them to the end of the file.
 
 ## v1.8
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,12 +4,7 @@
 
 ### Potential breaking changes
 
-  * [Phoenix.Router] Phoenix will now group your routes per verb during compilation. This improves compilation times for large routers with no performance cost at runtime. However, this change implies that wildcard routes are always matched last, which changes the semantics of dead routes. If you had this code:
-
-      match :*, "/foo"
-      get "/foo"
-
-  The second route would never be matched but it will now be as part of this change. However, note that `get "/foo"` should not exist in the first place: it should be either removed or moved first. We recommend checking for any `match :*` in your router and moving them to the end of the file.
+  * [Phoenix.Router] Phoenix will now group your routes per verb during compilation when all `match :*` and `forward` routes are defined at the end of the router. This improves compilation times for large routers with no performance cost at runtime. If you have a route with an explicit verb after a `match :*` or `forward`, Phoenix will preserve the previous ordered matching semantics and emit a warning
 
 ## v1.8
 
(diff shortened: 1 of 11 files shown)


SteffenDE:
I think it's fine to just mention it in the changelog. Alternatively, maybe we could detect such dead routes at compile time and drop them to keep the same behavior, emitting a deprecation warning? Then we drop that extra code for 2.0.

CHANGELOG.md
```diff
@@ -1,6 +1,15 @@
...
+## v1.9.0-dev
+
+### Potential breaking changes
+
+  * [Phoenix.Router] Phoenix will now group your routes per verb during compilation. This improves compilation times for large routers with no performance cost at runtime. However, this change implies that wildcard routes are always matched last, which changes the semantics of dead routes. If you had this code:
+
+      match :*, "/foo"
+      get "/foo"
+
+  The second route would never be matched but it will now be as part of this pull request. However, note that `get "/foo"` should not exist in the first place: it should be either removed or moved first. We recommend checking for any `match :*` in your router and moving them to the end of the file.
```
SteffenDE:
```suggestion
  The second route would never be matched but it will now be as part of this change. However, note that `get "/foo"` should not exist in the first place: it should be either removed or moved first. We recommend checking for any `match :*` in your router and moving them to the end of the file.
```

test/phoenix/router/routing_test.exs
```diff
@@ -224,18 +223,6 @@ defmodule Phoenix.Router.RoutingTest do
     assert conn.resp_body == "users move"
   end
 
-  test "any verb matches" do
```
SteffenDE, line 227: why did we remove that test?

josevalim: We already have a match :* above which is tested through other paths and this was emitting a warning with the new changes, so I moved it to a separate warning test!

SteffenDE: oh, but wait. In this example

```elixir
match :*, "/any", UserController, :any
get "/*path", UserController, :not_found
```

the get route is not a dead route. It is only dead for the very specific case that path is `/any`. So the changelog is wrong when it says

> which changes the semantics of dead routes

josevalim: You are right. Perhaps what we do is to change the warning to always warn if any route is used after `:*`?

test/phoenix/router/routing_test.exs
```diff
@@ -248,6 +235,23 @@ defmodule Phoenix.Router.RoutingTest do
...
+                 import ExUnit.Assertions, except: [trace: 3]
+
+                 match :*, "/foo/:bar", UserController, :index
+                 get "/foo/:baz", UserController, :show
+               end
+
+               conn = call(MatchOverlap, :get, "foo/example")
+               assert conn.resp_body == "users show"
+             end) =~ "found route matching on \"/foo/:baz\" after match(:*, \"/foo/:bar\")"
+    end
```
rhcarvalho: I think there are a few more cases where behavior changed and we're not triggering a warning. Failing tests:

```elixir
    test "warns on overlapping concrete route after :*" do
      assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
        defmodule MatchConcreteOverlap do
          use Phoenix.Router
          import ExUnit.Assertions, except: [trace: 3]

          match :*, "/foo/:bar", UserController, :index
          get "/foo/bar", UserController, :show
        end

        conn = call(MatchConcreteOverlap, :get, "foo/bar")
        assert conn.resp_body == "users show"
      end) =~ "found route matching on \"/foo/bar\" after match(:*, \"/foo/:bar\")"
    end

    test "warns on specific route after glob wildcard :*" do
      assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
        defmodule MatchGlobOverlap do
          use Phoenix.Router
          import ExUnit.Assertions, except: [trace: 3]

          match :*, "/*path", UserController, :index
          get "/foo", UserController, :show
        end

        conn = call(MatchGlobOverlap, :get, "foo")
        assert conn.resp_body == "users show"
      end) =~ "found route matching on \"/foo\" after match(:*, \"/*path\")"
    end

    test "warns on specific route after forward" do
      assert ExUnit.CaptureIO.capture_io(:stderr, fn ->
        defmodule MatchForwardOverlap do
          use Phoenix.Router
          import ExUnit.Assertions, except: [trace: 3]

          defmodule AdminPlug do
            def init(opts), do: opts
            def call(conn, _opts), do: Plug.Conn.send_resp(conn, 200, "admin forward")
          end

          forward "/admin", AdminPlug
          get "/admin/login", UserController, :show
        end

        conn = call(MatchForwardOverlap, :get, "admin/login")
        assert conn.resp_body == "users show"
      end) =~ "found route matching on \"/admin/login\" after forward(\"/admin\")"
    end
```

rhcarvalho (commented):
It's wonderful how many optimization possibilities are open :)

> ... or put this change behind an option.

Considering this could cause unintended change in behavior during an upgrade and be hard to detect (warnings can be missed before and after pushing to production, we may miss some corner case), I think an opt-in approach would be safer for existing production apps.

josevalim:
@rhcarvalho @SteffenDE it should be good now. We always warn if match :* and forward are not last and then we only optimize in case they are.

rhcarvalho:
@josevalim I found at least one gap related to the interaction with `scope`:

```elixir
    test "does not warn and remains optimized on disjoint scopes" do
      # This test demonstrates the gap: scope /admin and scope /api do not overlap.
      # Defining forward in /admin before get in /api should not warn, but currently it does.
      warnings = ExUnit.CaptureIO.capture_io(:stderr, fn ->
        defmodule DisjointScopesRouter do
          use Phoenix.Router

          scope "/admin" do
            forward "/", UserController
          end

          scope "/api" do
            get "/users", UserController, :index
          end
        end
      end)

      assert warnings == ""
    end
```

SteffenDE:
I'm thinking we should move this behind a compile-time option. Then projects with very large routers can opt in, but by default, the existing behavior is kept.

josevalim:
@rhcarvalho I don't think it is practical for us to detect all of those cases, that's why the suggestion is to move all forward/match-star to the end of the file. I'd ideally prefer to not have two compilation mechanisms in the long term. Are we against asking to move all of them to the end?

SteffenDE:
The issue I see with asking people to move them is that it means that people effectively need to duplicate scopes that define such routes, just to satisfy the optimization, even if that optimization does not bring a measurable performance improvement to a regular-sized router.

```elixir
scope "/admin" do
  pipe_through [:browser, :require_admin]

  get "/foo", ...
  post "/bar", ...  

  forward "/dashboard", Dashboard
end

scope "/api" do
  pipe_through [:api, :require_api_token]

  get "/baz", ...
  match :*, "/other", SomeController, :any
end
```

has to become

```elixir
scope "/admin" do
  pipe_through [:browser, :require_admin]

  get "/foo", ...
  post "/bar", ...  
end

scope "/api" do
  pipe_through [:api, :require_api_token]

  get "/baz", ...
end

scope "/admin" do
  pipe_through [:browser, :require_admin]
  forward "/dashboard", Dashboard 
end

scope "/api" do
  pipe_through [:api, :require_api_token]
  match :*, "/other", SomeController, :any
end
```

and suddenly you have to remember to keep the pipelines in sync...

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Changelog for v1.9
 
-Nothing, so far.
+## v1.9.0-dev
+
+### Potential breaking changes
```
josevalim:
```suggestion
### Deprecation
```

josevalim:
I made it opt-in now. Now in theory we can backport this to the v1.8 branch...

guides/routing.md
```diff
@@ -613,6 +613,8 @@ end
 
 This means that all routes starting with `/jobs` will be sent to the `BackgroundJob.Plug` module. Inside the plug, you can match on subroutes, such as `/pending` and `/active` that shows the status of certain jobs.
 
+Since forwards match all HTTP methods under their path, keep them at the end of your router, after your application routes. The same applies to any `match :*` routes. This is required when a router is configured with `group_by: :verb`.
+
```
josevalim:
```suggestion
```

CHANGELOG.md
```diff
@@ -1,6 +1,10 @@
 # Changelog for v1.9
 
-Nothing, so far.
+## v1.9.0-dev
+
+### Enhancements
+
+  * [Phoenix.Router] Add `use Phoenix.Router, group_by: :verb` to group routes per verb during compilation. This can improve compilation times for large routers with no performance cost at runtime. When enabled, all `match :*` and `forward` routes must be defined at the end of the router, otherwise compilation fails with a list of violations
```
josevalim:
```suggestion
Nothing, so far.
```

rhcarvalho (approved):
I agree with Steffen's observation re: forcing a `scope` to be split and pipelines kept in sync, but considering the opt-in approach it sounds like a great option for those needing the added performance.

@josevalim yes, a single compilation strategy would be best, but this seems too risky to force onto everyone else by default.

SteffenDE (approved):

josevalim:

PR #6762, opened 2026-07-23
Allow configuring LongPoll to send its token via headers

### LINKED ISSUE #6761: Longpoll transport passes token in URL params
### Actual behavior

The socket.js Longpoll transport passes a session resume token in the URL of a GET request as a `token` parameter. This seems to have similar implications to #5778. Specifically as a query param in a GET it's more likely to end up in logs. If someone gets access to it before it expires they could potentially highjack the channel session. 

In our case CloudFlare's bot protection ML model also flags those long poll requests as malicious. Claude's best guess is that it is because of that token. 

### Expected behavior

Longpoll session tokens should be unlikely to end up in logs by default and CloudFlare should not flag long poll requests as bot traffic.

SteffenDE:
This changes the longpoll client to send its token in a header instead of query parameters. The server still accepts old clients.

Closes https://github.com/phoenixframework/phoenix/issues/6761.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -1,6 +1,22 @@
 # Changelog for v1.9
 
-Nothing, so far.
+## Enhancements
+
+  * [Transports] Allow the longpoll session token to be sent in the
+    `x-phoenix-longpoll-token` header instead of the query string, via
+    `longpoll: [token_location: :header]`.
+    Longpoll responses are now also sent with `cache-control: no-store`.
+
+    Because the setting is negotiated per session and the server keeps
+    accepting both, it can be flipped with a rolling deploy. Deploy a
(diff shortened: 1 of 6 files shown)


CHANGELOG.md
```diff
@@ -1,6 +1,22 @@
...
+    Because the setting is negotiated per session and the server keeps
+    accepting both, it can be flipped with a rolling deploy. Deploy a
+    version that accepts both first, then switch to `:header` once no
+    node predates it:
+
+        socket "/socket", MyAppWeb.UserSocket,
+          longpoll: [token_location: :header]
+
+    Older clients that do not understand the negotiation keep using
+    params and continue to work.
```
SteffenDE: We can also backport this and then switch the default to header in 1.9.

PR #6768, opened 2026-07-28
Fall back to longpoll when the websocket keeps dying after opening

### LINKED ISSUE #6766: phoenix.js: connectWithFallback never falls back when the websocket dies shortly after opening
### Environment

- Phoenix 1.8.5 (phoenix.js), LiveView 1.1.28
- Chrome 138, observed in production behind enterprise security proxies (e.g. Zscaler-class appliances)

### Summary

`Socket.connectWithFallback` (assets/js/phoenix/socket.js) can only fall back to the fallback transport in two ways: the fallback timer fires before any successful open, or an error arrives before any successful open (`if(primaryTransport && !established)`). The `onOpen` handler sets `established = true` permanently.

A websocket that **opens successfully and then dies moments later** — which is exactly how several enterprise middleboxes behave (the 101 upgrade succeeds, then the appliance kills the tunnel ~100-500ms in) — escapes both paths. After the first open, `established` is latched, so every subsequent error takes the normal reconnect path with reset backoff, the websocket opens again, dies again, and the client flaps forever. Longpoll is never tried.

### Reproduction

Adversarial TCP proxy in front of a Phoenix app that allows the websocket upgrade to complete, then closes the connection 200ms after open. With `longPollFallbackMs` set, the client falls back only when the kill races *ahead* of the open event; when the kill lands after open, the client flaps indefinitely (verified with a scripted browser: continuous open/close cycles, zero longpoll attempts, backoff never grows because each open resets it).

### Expected

N consecutive open-then-die cycles within a window should count as primary-transport failure and trigger the same fallback as never having connected — the fallback exists precisely for networks that can't sustain websockets.

### Real-world impact

We traced a large customer's users stuck in this flap on their corporate network (appliance kills established websockets). Because fallback never engaged, they had no working transport path from the fallback logic at all.

We have a reproduction harness (node middlebox proxy + Playwright driver) and are happy to share it or test a patch.

---
*Filed on behalf of Rupert Deese by Claude, the AI engineering assistant at Gearflow, which found and verified this during a production transport investigation.*

rdeese:
Fixes #6766.

### The bug

`Socket.connectWithFallback` has two paths to the fallback transport, and both are gated on the primary transport never having opened:

* the `fallbackTimer`, and
* the `onError` handler, guarded by `if(primaryTransport && !established)`.

A websocket that opens and is then killed a few hundred milliseconds later escapes both. `established` latches on the first open, and the reconnect path re-enters `connectWithFallback`, whose first statement clears the pending fallback timer — so every cycle starts the clock over. `reconnectTimer.reset()` in `onConnOpen` keeps the backoff at its floor, and the client flaps indefinitely without ever trying longpoll.

The timer that gets re-armed on open does cover the neighbouring case, where the connection stays up past `longPollFallbackMs` without answering the ping. The gap is the connection that dies *before* that threshold, over and over.

### The fix

Treat "opened but never proved useful" as primary-transport failure. `onConnClose` counts a connection that closed without ever having delivered a message; a connection that delivered one resets the count. Once `MAX_UNUSABLE_PRIMARY_CONNECTIONS` (3) pile up in a row, the next `connectWithFallback` falls back immediately.

Why "received a message" rather than time-alive: uptime is not evidence of usefulness — a middlebox can hold a socket open and drop every frame — while an inbound message proves the path works end to end. On a healthy connection that proof arrives within one RTT, because `connectWithFallback` already pings on open, so a legitimately flaky-but-working network keeps reconnecting as it does today.

The fallback is deliberately **not** memorized in `sessionStorage` here. The primary transport did connect, so the existing "only memorize LP if we never connected to primary" invariant stands and the next page load tries websockets again.

### Tests

Two tests in `assets/test/socket_test.js` drive a fake transport under fake timers:

* opens, then dies before any message, repeatedly → `replaceTransport(LongPoll)` after 3 connections. On `main` this test runs 11 connections in the same window and never falls back.
* opens, delivers a message, then dies, repeatedly → no fallback, and the counter stays at 0.

`npm test` is green (180 passing, 3 skipped).

### Where this came from

We hit this in production with users behind enterprise security appliances that allow the websocket upgrade and then kill the established tunnel. One customer's client logged roughly 100 open/close cycles against 11 message arrivals over a week: the socket kept opening, almost never carried anything, and the fallback never engaged, so those users had no working transport path at all.

We have a reproduction harness (a node middlebox proxy that kills sockets a fixed delay after open, plus a Playwright driver) and are glad to share it or run a patch through it.

### Noticed, not fixed here

`errorRef` is only removed when the fallback engages, so a long-lived socket accumulates one error callback per reconnect. Out of scope for this fix — happy to send it separately if you'd like it.

---
*Submitted on behalf of Rupert Deese by Claude, the AI engineering assistant at Gearflow.*

### DIFF
--- a/assets/js/phoenix/constants.js
+++ b/assets/js/phoenix/constants.js
@@ -6,6 +6,9 @@ export const SOCKET_STATES = {connecting: 0, open: 1, closing: 2, closed: 3}
 export const MAX_LONGPOLL_BATCH_SIZE = 100;
 export const DEFAULT_TIMEOUT = 10000
 export const WS_CLOSE_NORMAL = 1000
+// how many primary transport connections may open and then close again without
+// ever delivering a message before we fall back to the fallback transport
+export const MAX_UNUSABLE_PRIMARY_CONNECTIONS = 3
 export const CHANNEL_STATES = {
   closed: "closed",
   errored: "errored",
--- a/assets/js/phoenix/socket.js
+++ b/assets/js/phoenix/socket.js
@@ -4,6 +4,7 @@ import {
   CHANNEL_EVENTS,
   DEFAULT_TIMEOUT,
   DEFAULT_VSN,
+  MAX_UNUSABLE_PRIMARY_CONNECTIONS,
   SOCKET_STATES,
   TRANSPORTS,
   WS_CLOSE_NORMAL,
@@ -120,6 +121,8 @@ export default class Socket {
     this.timeout = opts.timeout || DEFAULT_TIMEOUT
     this.transport = opts.transport || global.WebSocket || LongPoll
     this.primaryPassedHealthCheck = false
+    this.unusablePrimaryConnections = 0
+    this.connReceivedMessage = false
     this.longPollFallbackMs = opts.longPollFallbackMs
     this.fallbackTimer = null
     this.sessionStore = opts.sessionStorage || (global && global.sessionStorage)
@@ -218,6 +221,7 @@ export default class Socket {
   replaceTransport(newTransport){
     this.connectClock++
     this.closeWasClean = true
+    this.unusablePrimaryConnections = 0
     clearTimeout(this.fallbackTimer)
     this.reconnectTimer.reset()
     if(this.conn){
@@ -392,6 +396,7 @@ export default class Socket {
   transportConnect(){
     this.connectClock++
     this.closeWasClean = false
+    this.connReceivedMessage = false
     let protocols = undefined
     // Sec-WebSocket-Protocol based token
     // (longpoll uses Authorization header instead)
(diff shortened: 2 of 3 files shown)


SteffenDE:
So I've been thinking about this and I don't think we should go along with it for now. I'm worried that if we start to assume arbitrary middleboxes, we will end up in a rabbit hole of edge cases. Since this is the first time this comes up, I don't think it is a frequent enough problem to warrant a fix in Phoenix. If this comes up more frequently, we'll definitely reconsider.

For now, I'd suggest to implement your fallback in your app.js:

```javascript
import {Socket, LongPoll} from "phoenix"
import {LiveSocket} from "phoenix_live_view"

const csrfToken = document
  .querySelector("meta[name='csrf-token']")
  .getAttribute("content")

const liveSocket = new LiveSocket("/live", Socket, {
  params: {_csrf_token: csrfToken}
})

const socket = liveSocket.getSocket()

const maxUnusableConnections = 3

let opened = false
let receivedMessage = false
let unusableConnections = 0
let usingFallback = false

socket.onOpen(() => {
  if(usingFallback) return

  opened = true
  receivedMessage = false
})

socket.onMessage(() => {
  if(usingFallback) return

  receivedMessage = true
  unusableConnections = 0
})

socket.onClose(event => {
  if(usingFallback) return

  if(opened && !receivedMessage && event?.code !== 1000){
    unusableConnections++
  }

  opened = false
  receivedMessage = false

  if(unusableConnections >= maxUnusableConnections){
    usingFallback = true
    liveSocket.replaceTransport(LongPoll)
  }
})

liveSocket.connect()
window.liveSocket = liveSocket
```

PR #6781, opened 2026-08-02
Ensure single trailing newline in generated files

rhcarvalho:
And no unintentional consecutive blank lines in the generated content.

Those two issues are rather aesthetic, but downstream users notice and they are a common source of code churn and maintenance overheard that we can prevent from now on.

The single trailing newline is a common convention in Unix and POSIX systems, and it is also what the Elixir formatter does (`Code.format_file!` always appends a trailing newline [1]).

The consecutive blank lines in Elixir code are also automatically removed by the Elixir formatter. It is not a strict rule for other files, but generally they are not expected and most of the time added unintentionally, e.g. when using conditional EEx templates or concatenating strings (AGENTS.md / usage rules).

Except for a handful of generated files (favicon.ico, phoenix.png and *.pem certificates), all generated files are candidates for those two rules. Instead of updating lots of tests (which would cause a lot of churn and be a future maintenance burden, easy to miss in new tests), we update `MixHelper.assert_file/1` to check for those two rules in all current and future generated files. If we need more exceptions in the future, it is easy to change `assert_file/1` to add them.

[1]: https://github.com/elixir-lang/elixir/blob/545dddf138e4cb1ee874e6f2c26882c9b438f551/lib/elixir/lib/code.ex#L1137-L1141

PR #6782, opened 2026-08-02
Ensure single trailing newline in source files

rhcarvalho:
Reduce future code churn by gating source files with a test that ensures tracked text files in the repository end with a single trailing newline (UNIX convention).

Alternatives to enforce this without a test:

An `.editorconfig` file at the root of the repository makes most editors automatically insert a final newline and trim trailing whitespace on save, preventing the problem at the source:

    root = true

    [*]
    insert_final_newline = true
    trim_trailing_whitespace = true
    end_of_line = lf
    charset = utf-8

A `.gitattributes` file makes Git itself flag whitespace issues in diffs, on apply, and on merge:

    * text=auto eol=lf whitespace=blank-at-eol,blank-at-eof

Both files support per-path exceptions. In `.editorconfig`, a section like `[*.png]` can override any setting. In `.gitattributes`, paths can opt out with `-whitespace` or be marked as binary (e.g. `*.png binary`).

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -24,3 +24,6 @@ https://github.com/wojtekmach/mix_install_examples/blob/main/phoenix.exs
 
 ### Expected behavior
 
+<!--
+Describe the expected behaviour.
+-->
--- a/test/phoenix/source_files_test.exs
+++ b/test/phoenix/source_files_test.exs
@@ -1,0 +1,36 @@
+defmodule Phoenix.SourceFilesTest do
+  use ExUnit.Case, async: true
+
+  @repo_root Path.expand("../..", __DIR__)
+  @git_repo? File.dir?(Path.join(@repo_root, ".git")) and System.find_executable("git") != nil
+  @excluded_exts ~w(.br .foo .gz .ico .lock .map .pem .png)
+  @excluded_files ~w(test/fixtures/hello.txt)
+
+  @tag skip: if(not @git_repo?, do: "git or .git repository not available")
+  test "all tracked text files in the repository end with a single newline" do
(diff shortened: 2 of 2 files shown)


rhcarvalho:
This is if we want to go all in with ensuring source files (not only installer and generator templates, as in #6781) follow the UNIX convention. The two PRs are complementary, not mutually exclusive.

The `.gitattributes` approach would allow us to add something like `git diff --check origin/main...HEAD` to the CI workflow, so to check only the changes introduced by the PR, without flagging pre-existing issues in untouched files.

But it has one caveat: the `.gitattributes` whitespace checks (blank-at-eof, blank-at-eol) catch extra blank lines at EOF and trailing whitespace on lines, but they don't catch a missing final newline. Git shows the `\ No newline at end of file` marker in diffs, but `--check` doesn't flag it as an error. So `.gitattributes` + `git diff --check` covers most of the rule but not all of it.

rhcarvalho:
Rebased on top of main after #6781 has been merged.

Many of the trailing newlines in source files were handled there. The only "offender" left is the PR template in `.github/ISSUE_TEMPLATE/bug_report.md`, which can get an HTML comment in the trailing "describe expected" part to match the "describe actual" part.

.github/ISSUE_TEMPLATE/bug_report.md
```diff
@@ -24,3 +24,6 @@ https://github.com/wojtekmach/mix_install_examples/blob/main/phoenix.exs
 
 ### Expected behavior
 
+<!--
+Describe the expected behaviour.
```
rhcarvalho, line 28: I noticed the template uses both "behavior" (US spelling) and "behaviour" (UK spelling, also technical term in Erlang/Elixir).

Maybe we could update towards one or the other for local consistency.

PR #6783, opened 2026-08-02
Only use `async: true` for ConnCase error tests on Postgres/no-Ecto

rhcarvalho:
Fixes a long-standing flaky test issue in generated SQLite applications where running `mix test` would intermittently fail with:

    ** (Exqlite.Error) Database busy

Recent failure in CI: https://github.com/phoenixframework/phoenix/pull/6781#issuecomment-5159975190, https://github.com/phoenixframework/phoenix/actions/runs/30761860392/job/91533719530?pr=6781 which motivated me to go down the rabbit hole. I've ran into similar flakes in a production app generated years ago, but never bothered to root cause it until now.

## Root Cause

`phx.new` generated `error_html_test.exs` and `error_json_test.exs` with hardcoded `use ConnCase, async: true`.

In apps with Ecto enabled, `ConnCase` includes a setup block that calls `DataCase.setup_sandbox(tags)`. When ExUnit runs an `async: true` test module concurrently with `async: false` test modules (such as tests generated by `phx.gen.auth`), `setup_sandbox` sees `tags[:async] == true` and checks out a non-shared secondary connection handle (`shared: false`) from the Repo pool.

Because SQLite uses single-file database locking, having two active database connections opening sandbox transactions concurrently causes file lock contention whenever a synchronous test inserts records (e.g. `user_fixture()`). Depending on ExUnit test seeding and process scheduling, the secondary connection checkout fails with `(Exqlite.Error) Database busy`.

## Background

When `--database sqlite3` support was added in 51436960dadd9d3fc380d38acb51ba928a7ea059 (#4268, April 2021), error controller test templates retained hardcoded `use ConnCase, async: true`. Later, when `phx.gen.auth` introduced support for `async: true` in 3d609e01f2a123dccb3d580c4151b230cbe0e645 (#5689, January 2024), it added a helper to conditionally append `async: true` ONLY for PostgreSQL adapters while leaving non-Postgres adapters as `async: false`. However, `phx.new` was not updated at the time to use the same logic for its generated error test templates.

## Fix

1. Add `test_case_options/1` in `Phx.New.Generator` matching the behavior of `phx.gen.auth`:
   - `Ecto.Adapters.Postgres` -> `", async: true"`
   - `nil` (no Ecto) -> `", async: true"`
   - non-Postgres adapters (SQLite, MySQL, TDS) -> `""`

2. Update `error_html_test.exs.eex` and `error_json_test.exs.eex` templates to use `<%= @test_case_options %>`.

3. Added test coverage across single app and umbrella project generators for Postgres, `--no-ecto`, SQLite3, MySQL, and MSSQL.

### DIFF
--- a/installer/lib/phx_new/generator.ex
+++ b/installer/lib/phx_new/generator.ex
@@ -326,7 +326,8 @@ defmodule Phx.New.Generator do
       elixir_install_bin_path: from_elixir_install && elixir_install_bin_path(),
       inside_docker_env?: inside_docker_env?,
       agents_md: agents_md,
-      config_regex_E: Version.match?(System.version(), "~> 1.19.3 or ~> 1.20") && "E" || ""
+      config_regex_E: Version.match?(System.version(), "~> 1.19.3 or ~> 1.20") && "E" || "",
+      test_case_options: test_case_options(adapter_module)
     ]
 
     %{project | binding: binding}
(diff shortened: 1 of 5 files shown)


SteffenDE (approved):

SteffenDE:
Let's gooo 🙌🏻

PR #6784, opened 2026-08-02
Bump dependencies with security advisories

rhcarvalho:
Declared dependencies in `mix.exs` remain the same, as to not force downstream users to upgrade despite the advisories.

Versions used in development and CI (`mix.lock`) updated to the latest versions. Since all versions of Bandit prior to 1.12.1 have a security advisory, `mix.exs` was updated to reflect that, with no effect on downstream users.


mix.exs
```diff
@@ -91,7 +91,7 @@ defmodule Phoenix.MixProject do
 
       # Optional deps
       {:plug_cowboy, "~> 2.7", optional: true},
-      {:bandit, "~> 1.0", optional: true},
+      {:bandit, "~> 1.12", optional: true},
```
SteffenDE: I think the versions here should reflect what Phoenix requires. In general, a library shouldn't unnecessarily force requirements on users. That said, we don't really test with older versions, so at some point this is probably fine. But I'm not sure we should be the ones to judge what versions people use, even if those have security issues. With the latest mix version, it is quite prominent already.

We can bump our mix.lock.

rhcarvalho:
Forgot to update the integration test project deps in https://github.com/phoenixframework/phoenix/pull/6784/commits/49e944d79d5b841246762e26cc0c111cdc465fe6, done in https://github.com/phoenixframework/phoenix/pull/6784/commits/3b7d7df97bff414c69516925f9c3f41eeca3fea4.

mix.exs
rhcarvalho: You are right, that's why I was selective with the bumps in `mix.exs`. I'll revert the changes there as to not force upgrades downstream (that's exactly what caused the integration tests to fail at first: https://github.com/phoenixframework/phoenix/actions/runs/30767036212/job/91547486878#step:4:463.

rhcarvalho:
- Reverted changes to `mix.exs`
- Bumped more versions in `integration_test/mix.lock` that had advisories (originally I targeted only `mix.lock` at the project root)

SteffenDE:
🙌🏻

PR #6786, opened 2026-08-04
Remove unused installer template bindings

rhcarvalho:
Following up on unifying `.gitignore` templates into a static file, an audit of all installer bindings was conducted to see if any variables eliminated from `.gitignore.eex` became unused repo-wide.

While all variables from `.gitignore.eex` remain required by other templates, the audit uncovered two pre-existing unused bindings (`pubsub_server` and `root_app_name`), allowing us to clean them up along with their supporting code:

- `pubsub_server`: Stopped being used in templates in commit 04398b60e (June 2019) when PubSub was moved into the application supervision tree and templates switched to referencing `<%= web_namespace %>.PubSub` directly. Remove `pubsub_server` from binding and delete the helper `get_pubsub_server/1`.

- `root_app_name` / `root_app`: `root_app_name` was defined when umbrella generators were added in commit dad413ff2 (Jan 2017), but templates have always used `@root_app_module` (`MyApp.Umbrella`) instead. Remove `root_app_name` from binding and drop the `root_app` struct field from `%Phx.New.Project{}`.

- Helper cleanup: `Project.new/2` already initializes `root_mod` to `app_mod` (`Module.concat([opts[:module] || Macro.camelize(app)])`). In `Phx.New.Single`, `put_root_app/1` was computing the exact same value a second time to assign `root_mod`, making it a no-op. Remove `put_root_app/1` from `Single`, and rename `put_root_app/1` to `put_root_mod/1` in `Phx.New.Umbrella` where it actually overrides `root_mod` with the `MyApp.Umbrella` module name.

### DIFF
--- a/installer/lib/phx_new/generator.ex
+++ b/installer/lib/phx_new/generator.ex
@@ -274,8 +274,6 @@ defmodule Phx.New.Generator do
     {web_adapter_app, web_adapter_vsn, web_adapter_module, web_adapter_docs} =
       get_web_adapter(web_adapter)
 
-    pubsub_server = get_pubsub_server(project.app_mod)
-
     adapter_config =
       case Keyword.fetch(opts, :binary_id) do
         {:ok, value} -> Keyword.put_new(adapter_config, :binary_id, value)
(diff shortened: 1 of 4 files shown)


SteffenDE:
🙌🏻

PR #6787, opened 2026-08-05
Add class support for col item

widianto:
My first PR. Any feedback or guidelines would be greatly appreciated.

### Purpose

Add support for a `class` attribute on `<.col>` components in `Phoenix.CoreComponents` (`installer/templates/phx_web/components/core_components.ex`).

This allows styling individual table cells using either a single CSS class (for example, `"text-right"`) or multiple classes (for example, `"text-right font-bold"`), and still resoect existing implementation.

### Motivation

The generated `<.table>` component already supports several customization options, but `<.col>` doesn't currently allow passing a class attribute. I needed this to apply custom styling to individual table cells without modifying the generated component.

### What this supports

1. Having a `row_click` attribute in `<.table />`:

`<.table id="products" rows={@products} row_click={&JS.navigate(~p"/products/#{&1}")}>`

and a `class` in col item:

`<:col :let={product} label="Price" class="text-right">`

Result:

01-row_click_class

2. Without a `row_click` attribute in `<.table />`:

`<.table id="products" rows={@products}>`

but with a `class` attribute in col item:

<:col :let={product} label="Price" class="text-right font-bold">

Result:

02-row_click_nil_with_class

3. without a `row_click` attribute in `<.table />`:

`<.table id="products" rows={@products}>`

as well as no `class` attribute:

`<:col :let={product} label="Price">`

Result:

03-row_click_nil-and-class-nil

### Backward compatibility

This change is fully backward compatible; existing templates continue to behave as before.

### DIFF
--- a/installer/templates/phx_web/components/core_components.ex.eex
+++ b/installer/templates/phx_web/components/core_components.ex.eex
@@ -357,6 +357,7 @@ defmodule <%= @web_namespace %>.CoreComponents do
 
   slot :col, required: true do
     attr :label, :string
+    attr :class, :any
   end
 
   slot :action, doc: "the slot for showing user actions in the last table column"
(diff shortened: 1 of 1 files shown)


SteffenDE:
Hey @widianto,

the generated components are meant for you to extend, not to handle all possible cases out of the box. We generally only add functionality the generators themselves use. And right now there's no case where the generators need to override a col class :)

So the change is perfectly fine to do in your own app, but we want to keep the generators minimal on purpose.

Thanks!

widianto:
Hi @SteffenDE, 

Appreciate the fast reply! I understand the design philosophy of keeping the base generators lean. 😀

Thanks for reviewing and for the great work on this library!

PR #6792, opened 2026-08-16
Wait for the server ack when leaving channels

patric-vinicios:
I ran into this while looking at channel cleanup in an app

`leave()` was returning `"ok"` before the server had actually replied.

`leave()` changes the channel state to `leaving` and then checks `canPush()`, but `canPush()` only returns true when the channel is still joined so at that point it always returns `false`

Then, the fallback runs right away and the channel closes locally instead of waiting for the server reply or the leave timeout

The fix is just to check `canPush()` before changing the state to `leaving`

If the channel can still push, it sends the leave request and waits for either the reply or the timeout. If it cant, it keeps the existing behavior and closes immediately

I also re-enabled the tests for the leaving state and leave timeout

Tested with:

* `npm test -- --runInBand`
* `npx eslint assets/js/phoenix/channel.js assets/test/channel_test.js`

### DIFF
--- a/assets/test/channel_test.js
+++ b/assets/test/channel_test.js
@@ -1041,11 +1041,7 @@ describe("with transport", function (){
       expect(channel.state).toBe("closed")
     })
 
-    // TODO - the following tests are skipped until Channel.leave
-    // behavior can be fixed; currently, 'ok' is triggered immediately
-    // within Channel.leave so timeout callbacks are never reached
-    //
-    it.skip("sets state to leaving initially", function (){
+    it("sets state to leaving initially", function (){
       expect(channel.state).not.toBe("leaving")
 
       channel.leave()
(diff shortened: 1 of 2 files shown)


assets/test/channel_test.js
```diff
@@ -1041,27 +1041,23 @@ describe("with transport", function (){
       expect(channel.state).toBe("closed")
     })
 
-    // TODO - the following tests are skipped until Channel.leave
-    // behavior can be fixed; currently, 'ok' is triggered immediately
-    // within Channel.leave so timeout callbacks are never reached
-    //
```
SteffenDE, line 1047: Since this behavior has been the same since the beginning, I'm inclined to keep it and only change it when we do a 2.0 release.

SteffenDE:
Closing as per the above. Thanks!

PR #6795, opened 2026-08-18
Fix release date for v1.8.11

pgeraghty:
Correct the release date for version 1.8.11 in the changelog (2026 rather than 2028...).

This was thrown up by a deps upgrade audit at my end, so I thought I'd flag it. Totally understand if it's dismissed.


SteffenDE:
Woops!

PR #6799, opened 2026-08-20
Clear :return_to session key after logging in

### LINKED ISSUE #6798: phx.gen.auth: `:<schema>_return_to` session key is never cleared after use
### Environment

* Phoenix version (mix deps): 1.8, 1.9
* Operating system: any

### Actual behavior

`log_in_<schema>/3` reads the return-to path from the session and redirects but never clears the key:

https://github.com/phoenixframework/phoenix/blob/be77755e1cc3ec80bce10ee0e1a31e760361a2ac/priv/templates/phx.gen.auth/auth.ex.eex#L35-L41

It is only cleared as a side effect of `renew_session/2`. But `renew_session/2` short-circuits when the user is already logged in

Repro:
1. Create an empty phoenix app, add auth, start the server
```bash
mix phx.new demo
cd demo
mix phx.gen.auth Accounts User users --no-live
mix deps.get
mix ecto.create && mix ecto.migrate
mix phx.server
```
2. Open http://localhost:4000/users/register and register foo@example.com.
3. Open http://localhost:4000/dev/mailbox, open the email, click the log-in link, and submit the confirmation form. The app is logged in and redirected to /
4. Open http://localhost:4000/users/settings, set a password, click Save Password. The app lands on /users/settings
5. Open http://localhost:4000/users/log-in and log in with foo@example.com + pass

Expected: redirect to /
Actual: redirect to /users/settings which was not cleared from the session

### Expected behavior

The generator should produce correct code that clears return-to

lukaszsamson:
Fixes #6798

### DIFF
--- a/priv/templates/phx.gen.auth/auth.ex.eex
+++ b/priv/templates/phx.gen.auth/auth.ex.eex
@@ -37,6 +37,7 @@ defmodule <%= inspect auth_module %> do
 
     conn
     |> create_or_extend_session(<%= schema.singular %>, params)
+    |> delete_session(:<%= schema.singular %>_return_to)
     |> redirect(to: <%= schema.singular %>_return_to || signed_in_path(conn))
   end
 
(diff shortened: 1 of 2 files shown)


SteffenDE:
🙌🏻

PR #6800, opened 2026-08-20
join_ref can be nil on non-join messages

SteffenDE:
We still drop messages with a stale (non-nil) join_ref.

Fixup for c73bbfcfeafce3d42cb270a7ecb41a4ca39ae393.

The protocol explicitly allows non join messages to omit the join_ref, see

https://phoenix.hexdocs.pm/1.8.11/writing_a_channels_client.html#message-format

Note: this also updates to docs to say clients **should** include the join ref for non join messages, as it allows staleness detection. We cannot make it required though, because it was allowed and explicitly documented in the past, so we allow it for backwards compatibility.


PragTob:
Thank you!

IMG_20171221_144500

josevalim (approved):

PR #6802, opened 2026-08-24
phx.gen.release: use Bob API structured filters and sorting for Docker tags

rhcarvalho:
Previously, `phx.gen.release` fetched all Docker tags matching an initial prefix from Bob across paginated requests and ran custom client-side version comparison logic (`best_docker_tag`, `compare_docker_tags`, `compare_otp_versions`, etc.) to select the newest compatible image.

With Bob's `/api/docker` endpoint now supporting structured filter parameters and numeric-aware natural sorting (https://github.com/hexpm/bob/pull/286), we can delegate the tag filtering and ranking directly to the server:

1. Pass `elixir_version`, `erlang_version`, `os`, `os_version`, and `sort=elixir_version,erlang_version,os_version` to `/api/docker`.

2. Select the first matching multi-arch slim tag directly from the sorted response, removing pagination and ~87 lines of hand-rolled version comparison logic.

3. Retain the `repo == docker_repo()` check with an explicit multi-arch `archs` guard to filter out single-arch `-amd64` and `-arm64` images returned by Bob's prefix-matched `repo` query.

4. Unify Bob URL helpers under `bob_docker_url/2`, providing `bob_docker_tags_url/1` for the API and `bob_docker_ui_url/1` for the human-facing Web UI link in failure messages.

5. Add tests covering multi-arch slim tag selection from sorted API results and asserting both the error message and the Web UI fallback link when no compatible image is found.

---

The non-exact match for `repo` may change in the future, see https://github.com/hexpm/bob/issues/289, which would reduce even further the number of candidate images we fetch from the API.

---

Changes assisted by Google Antigravity CLI (Gemini 3.7), but intently driven and reviewed by myself.

PR #6807, opened 2026-08-27
phx.gen.release: simplify Debian version parsing and usage

rhcarvalho:
Consolidate the Debian tag into a single `debian_vsn` binding instead of splitting it into `debian` and `debian_vsn` and stripping/re-appending `-slim`.

Key changes:
* Use an inline regex in `docker_tag_candidate/1` to declaratively parse Docker tags according to the tag specification while enforcing the `-slim` suffix. Compared to string splitting, the regex match is more concise while strictly validating accepted tag characters.
* Keep `debian_vsn` as the complete Debian tag (e.g. `trixie-20251117-slim`), removing the redundant `debian` template binding and aligning `ARG DEBIAN_VERSION=<%= debian_vsn %>` symmetrically with `elixir_vsn` and `otp_vsn`.
* Streamline candidate data flow and unify `_vsn` naming, removing redundant tuple conversions and duplicate `case` expressions in `gen_docker/2`.
* Add test coverage for pre-release versions with hyphens, non-slim/non-Debian tags, and malformed inputs.

---

Follows up on #6802.

### DIFF
--- a/lib/mix/tasks/phx.gen.release.ex
+++ b/lib/mix/tasks/phx.gen.release.ex
@@ -252,7 +252,7 @@ defmodule Mix.Tasks.Phx.Gen.Release do
         candidate = docker_tag_candidate(tag)
 
         if candidate && docker_tag_matches?(candidate, search) do
-          {:ok, candidate.elixir, candidate.otp, candidate.debian_vsn}
+          {:ok, candidate}
         end
       end)
     end)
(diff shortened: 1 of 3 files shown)


rhcarvalho:
I'm not lucky with the GHA runner :)

SteffenDE:
🙌🏻

PR #6808, opened 2026-08-27
Fix phx.new.web generator referencing Ecto

### LINKED ISSUE #5690: Creating Umbrella Web App With `phx.new.web` Results in App That Cannot Be Compiled
### Environment

* Elixir version (elixir -v): `1.16.0`
* Phoenix version (mix deps): `1.7.10`
* Operating system: macOS 14.2.1

### Actual Behavior

1. `mix phx.new foo --umbrella`
2. `cd foo_umbrella/apps`
3. `mix phx.new.web bar`
4. `cd ..`
5. `mix`

There are a couple errors that occur:
- Related to the appended configuration in `config/config.exs`:
  ```elixir
  config :bar,
    ecto_repos: [Bar.Repo],
    generators: [context_app: false]
  ```
  `Bar.Repo` should not exist in the context of specifically a web application. It is inconsistent with the generated `foo_web` configuration:
  ```elixir
  config :foo_web,
    ecto_repos: [Foo.Repo],
    generators: [context_app: :foo]
  ```
- Related to the generated `bar/test/support/conn_case.exs` file:
  ```elixir
  Bar.DataCase.setup_sandbox(tags)
  ```
  Similar to the above, I don't believe this should exist and should instead reference the dependent support module. Also inconsistent with the generated `foo_web` config:
  ```elixir
  Foo.DataCase.setup_sandbox(tags)
  ```

### Expected Behavior

The app should be able to be compiled when a new web app is generated in the context of an umbrella app.

### Proposed Solution

Adding something like a `--depends-on` flag to the `phx.new.web` generator so that the backing app can be specified in the generator. The additional context of the flag would also enable adding `{:foo, in_umbrella: true},` automatically to `deps` inside of `bar/mix.exs`

Example:
```sh
mix phx.new.web bar --depends-on foo
```

Resulting in:
```elixir
# config/config.exs
config :bar,
  ecto_repos: [Foo.Repo],
  generators: [context_app: :foo]
```
```elixir
# bar/test/support/conn_case.exs
Foo.DataCase.setup_sandbox(tags)
```
```elixir
# bar/mix.exs
defp deps do
  [
    . . .
    {:foo, in_umbrella: true}
  ]
end
```

rhcarvalho:
The `phx.new.web` task is documented to create a bare Phoenix web project without database integration inside an umbrella. However, because it did not pass `--no-ecto`, it inherited the generator's default `ecto: true` setting, generating invalid references to a non-existent `<App>.Repo` and `<App>.DataCase`.

This passes `--no-ecto` by default when running `phx.new.web`.

Fixes #5690.
Closes #5691.

### DIFF
--- a/installer/lib/mix/tasks/phx.new.web.ex
+++ b/installer/lib/mix/tasks/phx.new.web.ex
@@ -39,6 +39,6 @@ defmodule Mix.Tasks.Phx.New.Web do
       Mix.raise "The web task can only be run within an umbrella's apps directory"
     end
 
-    Mix.Tasks.Phx.New.run(args, Phx.New.Web, :web_path)
+    Mix.Tasks.Phx.New.run(args ++ ["--no-ecto"], Phx.New.Web, :web_path)
   end
 end
--- a/installer/test/phx_new_umbrella_test.exs
+++ b/installer/test/phx_new_umbrella_test.exs
@@ -832,9 +832,7 @@ defmodule Mix.Tasks.Phx.New.UmbrellaTest do
           File.cd!(dir, fn ->
             assert_raise Mix.Error,
                          ~r"The web task can only be run within an umbrella's apps directory",
-                         fn ->
-                           Mix.Tasks.Phx.New.Web.run(["valid"])
-                         end
+                         fn -> Mix.Tasks.Phx.New.Web.run(["web_app"]) end
           end)
         end
       end)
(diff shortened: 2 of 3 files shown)


rhcarvalho:
Assisted by Google Antigravity CLI / Gemini 3.7 Flash.

SteffenDE:
@rhcarvalho can you rebase please?

rhcarvalho:
Rebased

SteffenDE:
Good, we might revisit adding a `--context-app` or similar in the future!

PR #6809, opened 2026-08-27
fix doc references

woylie:
This could be caught with `mix docs --warnings-as-errors` in the CI. Let me know if you want me to open a PR for that.


SteffenDE:
@woylie a PR is welcome!

PR #6812, opened 2026-08-28
Harden ci

woylie:
Couple of improvements to the CI. Zizmor only complains about the missing cooldown in the dependabot config. If you want me to add it, I can push another commit.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -3,6 +3,10 @@ on: [push, pull_request]
 permissions:
   contents: read
 
+concurrency:
+  group: ${{ github.workflow }}-${{ github.ref }}
+  cancel-in-progress: true
+
 jobs:
   mix_test:
     name: mix test (OTP ${{matrix.otp}} | Elixir ${{matrix.elixir}})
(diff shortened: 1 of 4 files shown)


.github/workflows/ci.yml
```diff
@@ -58,6 +65,11 @@ jobs:
 
       - name: Run tests
         run: mix test
+        if: ${{ !matrix.lint }}
+
+      - name: Run tests & lint
+        run: mix test --warnings-as-errors
```
rhcarvalho, line 71: (as documented in `mix help test`)

> This option only applies to test files. To treat warnings as errors during compilation and during tests, run:

```suggestion
        run: MIX_ENV=test mix do compile --warnings-as-errors + test --warnings-as-errors
```

rhcarvalho (approved):

.github/workflows/ci.yml
woylie: `mix compile --warnings-as-errors` is already run one step earlier.

SteffenDE (approved):

SteffenDE:
🙌🏻

PR #6813, opened 2026-08-28
docs: rename secret to salt

kevinschweikert:
the moduledocs even say that the second argument
is a "cryptographic salt" so we should name the arg as such.

Also `verify/4` and `sign/4` call it `salt`

### DIFF
--- a/lib/phoenix/token.ex
+++ b/lib/phoenix/token.ex
@@ -167,10 +167,10 @@ defmodule Phoenix.Token do
   """
   @spec encrypt(context, binary, term, [shared_opt | max_age_opt | signed_at_opt | encode_opt]) ::
           binary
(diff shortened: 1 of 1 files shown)


SteffenDE:
We take it from plug_crypto (https://plug-crypto.hexdocs.pm/Plug.Crypto.html#decrypt/4), so I guess we should first change it there?

kevinschweikert:
Totally! I peeked at the implementation and its even shadowing the actual `secret` variable 

```elixir

def decrypt(key_base, secret, token, opts)
      when is_binary(key_base) and is_binary(secret) and is_list(opts) do
    secret = get_secret(key_base, secret, opts)
```

I will open a PR!

kevinschweikert:
@SteffenDE plug_crypto has been updated with the new signatures

SteffenDE:
Nice, thank you!

PR #6816, opened 2026-08-28
Remove unused @epoch module attribute in integration test

rhcarvalho:
In #6764, the test "development workflow works as expected" was updated to "generated app boots with mix phx.server", removing the `File.touch!` and `File.stat!` assertions on non-existent HTML templates in an app generated with `--no-html`.

However, the `@epoch` module attribute definition at the top of the file was left behind, causing the compiler warning:
```
  "warning: module attribute @epoch was set but never used"
```

This removes the unused `@epoch` definition.


SteffenDE:
🙌🏻

PR #6818, opened 2026-08-31
Harden and speed up phx.server integration test

rhcarvalho:
## Summary

One particular integration test, `"generated app boots with mix phx.server"`, didn't run natively in my sandbox environment that has Elixir/OTP installed via `asdf`. Upon closer inspection, I realized the test had sleeps and polling, and could be improved to use deterministic synchronization. The test was also leaving ~145 MB directories in `installer/tmp` behind, filling up my disk as I iterate on integration tests locally.

## Details

Previously, the integration test booted `phx.server` using `elixir -S mix` combined with an in-band stdin watcher. This had several shortcomings:

1. Portability: `elixir -S` loads scripts via `Code.require_file`, which fails with a syntax error in environments where `mix` in `$PATH` is a shell wrapper (such as standard `asdf` or `mise` shims).
2. Disk leak & Cleanup races: Running the server asynchronously inside `spawn_link` raced with `with_installer_tmp`'s directory cleanup, requiring `autoremove?: false` which permanently leaked generated apps on disk after every test run.
3. Test latency & Redundant work: A blind polling loop in `request_with_retries` imposed a 5-second sleep penalty on every test run, while `mix test` was redundantly executed inside a test strictly intended for server booting and HTTP readiness.

This refactors the test to use deterministic, event-driven synchronization with a scoped `with_phx_server/2` helper:

- Uses `Port.open` to launch Mix directly as an OS executable.
- Synchronizes the test and server processes via an explicit readiness marker emitted only after the endpoint has bound the listening port, eliminating polling latency.
- Handles premature child exits immediately to fail fast instead of timing out.
- Signals child process shutdown via standard input and awaits child OS process termination before returning. This guarantees lock release before `with_installer_tmp` deletes the temporary directory.
- Restores automatic cleanup (`autoremove?: true`) in `with_installer_tmp`.
- Lowers execution time of this test file by ~40% (from ~35s to ~20s locally on Apple M4; previously ~52s in CI) by eliminating the blind polling sleep and focusing the test strictly on server boot.

Notes & Platform Considerations:
- Helper scope: `with_phx_server/2` and HTTP helpers are intentionally kept private to this test file for now since it is currently the only test verifying server boot.
- Separation of concerns: Scaffolded test execution is validated by dedicated code generator suites, keeping this test focused exclusively on server startup and HTTP readiness.
- Inets application: Removed redundant explicit `:inets.start()` call from the test since `:inets` is already started as part of `extra_applications` in `integration_test/mix.exs`.
- Port binding & async execution: The test verifies the out-of-the-box development workflow by booting `phx.server` on the default port 4000 under `async: true`. Binding a fixed OS port poses a collision risk (`:eaddrinuse`) if a developer is running a local dev server or if additional server tests are added in the future. We evaluated this tradeoff and decided to keep the fixed port for now since it matches the status quo and this is currently the only test booting a server; dynamic port handling can be revisited if more server tests are introduced.
- Windows compatibility: On Windows, `System.find_executable("mix")` resolves to `mix.bat`. As specified in Erlang/OTP's `open_port/2` documentation, ERTS natively dispatches `.bat` scripts through `cmd.exe` on Windows. Note that the integration test suite is not explicitly exercised on Windows in CI (which runs on Linux).

---

Assisted by Antigravity / Gemini 3.7 Flash (and Claude Sonnet 4.6, multiple iterations)

### DIFF
--- a/integration_test/test/code_generation/app_with_no_options_test.exs
+++ b/integration_test/test/code_generation/app_with_no_options_test.exs
@@ -20,7 +20,7 @@ defmodule Phoenix.Integration.CodeGeneration.AppWithNoOptionsTest do
   end
 
   test "generated app boots with mix phx.server" do
-    with_installer_tmp("development_workflow", [autoremove?: false], fn tmp_dir ->
+    with_installer_tmp("development_workflow", fn tmp_dir ->
       {app_root_path, _} =
         generate_phoenix_app(tmp_dir, "phx_blog", [
           "--no-assets",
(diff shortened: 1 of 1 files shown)


integration_test/test/code_generation/app_with_no_options_test.exs
```diff
@@ -20,7 +20,7 @@ defmodule Phoenix.Integration.CodeGeneration.AppWithNoOptionsTest do
   end
 
   test "generated app boots with mix phx.server" do
-    with_installer_tmp("development_workflow", [autoremove?: false], fn tmp_dir ->
+    with_installer_tmp("development_workflow", fn tmp_dir ->
```
rhcarvalho, line 23: `autoremove?: true` is the default:

https://github.com/phoenixframework/phoenix/blob/8b26148624d63050920ad2c9ac711f4d043028f6/integration_test/test/support/code_generator_case.ex#L102

integration_test/test/code_generation/app_with_no_options_test.exs
```diff
@@ -32,55 +32,95 @@ defmodule Phoenix.Integration.CodeGeneration.AppWithNoOptionsTest do
...
 
-      spawn_link(fn ->
-        run_phx_server(app_root_path)
+      with_phx_server(app_root_path, fn url ->
+        {:ok, response} = http_get(url)
+        assert response.status_code == 200
+        assert response.body =~ "PhxBlog"
       end)
-
-      :inets.start()
```
rhcarvalho, line 39: Already started in `mix.exs` since 53d36bc63013d9a84040bf811ff01c07c8c7e040 3 years ago.

https://github.com/phoenixframework/phoenix/blob/8b26148624d63050920ad2c9ac711f4d043028f6/integration_test/mix.exs#L25

integration_test/test/code_generation/app_with_no_options_test.exs
```diff
@@ -32,55 +32,95 @@ defmodule Phoenix.Integration.CodeGeneration.AppWithNoOptionsTest do
...
-        ],
-        cd: app_root_path
+          :binary,
+          :exit_status,
+          :stderr_to_stdout,
+          line: max_line_bytes(),
+          args: [
+            "phx.server",
+            "-e",
+            "IO.puts(\"PHX_SERVER_READY\"); spawn(fn -> IO.gets([]); System.halt(0) end)"
```
rhcarvalho, line 59: This expands on the previous trick and, if my understanding is correct, prints `PHX_SERVER_READY` to stdout strictly when the server is ready.

Reading from stdin is the mechanism the test process communicates with `phx.server` to tell it to shutdown.

SteffenDE:
🙌🏻

PR #6825, opened 2026-09-01
Improve integration test CI visibility, healthchecks, and reliability

rhcarvalho:
Previously, the integration test CI job ran entirely within a monolithic `integration_test/test.sh` script inside an Alpine container. Bundling system package installation, Hex/Rebar setup, proxy startup, dependency compilation, and test execution into a single script obscured the duration of each phase in CI metrics. CI also lacked explicit database readiness checks, relying solely on the duration of preceding commands to avoid connection failures.

This updates `.github/workflows/ci.yml` with the following improvements:

- Decompose monolithic test execution into discrete, individually timed CI steps (package installation, Hex/Rebar setup, proxy bridges, dependency fetching, MIX_ENV=test dependency compilation, and test execution) to surface exact phase timings in GitHub Actions UI.
- Add explicit container healthcheck probes (`--health-cmd`) for Postgres, MySQL, and MSSQL services to eliminate startup race conditions before tests begin.
- Detach background `socat` proxy streams (`> /dev/null 2>&1 &`) so the proxy step completes immediately without blocking the runner's step listener.
- Enable concurrent dependency and NIF compilation (`MIX_OS_DEPS_COMPILE_PARTITION_COUNT: 2` and `MAKEFLAGS: "-j2"`) tuned to runner vCPU capacity without CPU oversubscription.
- Standardize Alpine matrix parameterization by separating the `alpine` version tag from string suffixes.

---

I intend to suggest similar improvements for the local integration tests workflow as a separate PR.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -232,17 +232,23 @@ jobs:
           # look for latest alpine image here: https://bob.hex.pm/docker?repo=hexpm/elixir&os=alpine&sort=elixir_version,erlang_version,os_version
           - elixir: "1.18.4"
             otp: "27.3.4.3"
-            suffix: "alpine-3.22.4"
+            alpine: "3.22.4"
 
           - elixir: "1.20.4"
             otp: "29.0.5"
-            suffix: "alpine-3.23.5"
+            alpine: "3.23.5"
 
     container:
-      image: hexpm/elixir:${{ matrix.elixir }}-erlang-${{ matrix.otp }}-${{ matrix.suffix }}
+      image: hexpm/elixir:${{ matrix.elixir }}-erlang-${{ matrix.otp }}-alpine-${{ matrix.alpine }}
       env:
         ELIXIR_ASSERT_TIMEOUT: 10000
         PHX_CI: true
+        # Set to half of the runner vCPU cores to compile deps/NIFs
+        # concurrently across OS processes without CPU oversubscription.
+        # See https://docs.github.com/en/actions/reference/runners/github-hosted-runners#standard-github-hosted-runners-for-public-repositories
+        # and https://elixir-lang.org/blog/2025/10/16/elixir-v1-19-0-released/
+        MIX_OS_DEPS_COMPILE_PARTITION_COUNT: 2
+        MAKEFLAGS: "-j2"
     services:
       postgres:
         image: postgres:18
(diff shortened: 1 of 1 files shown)


rhcarvalho:
The first pass in https://github.com/phoenixframework/phoenix/pull/6825/commits/c1cb013b75d5d77239b015dc0b78fd500269f055 gives us visibility, but will likely slow down the overall integration test jobs because the DB health checks are serialized, blocking the rest of the job, including the slow `mix deps.compile` step.

https://github.com/phoenixframework/phoenix/actions/runs/33520432286/job/99897988398?pr=6825
- DBs take 45s to be ready
- `mix deps.compile` takes 47s
- Before, the DBs "got ready" concurrently with the dependency compilation, but there was no guarantee nor synchronization
- Now the two steps become serialized, slowing things down by ~40s

I'll send a new approach, replacing the health checks with an alternative to explicitly wait until the DBs are ready right before we start the test execution. That should give us the benefits of no test failures because DBs didn't boot yet, while keeping the test run times in the same ballpark.

SteffenDE:
Because of your recent change we already know now how long the tests themselves run, so the rest is all setup time, right? If we go ahead with that change, it means the docker script will have to duplicate code or we remove it (I sometimes use it)

https://github.com/phoenixframework/phoenix/blob/main/integration_test/docker.sh

SteffenDE:
> I intend to suggest similar improvements for the local integration tests workflow

Ah you're already aware

rhcarvalho:
> Ah you're already aware

I decided to split the changes to make it easier to review. It's easy to go on a tangent and...

With just the CI changes I have enough tangents to chase :)

I'll update the PR to check DB readiness on-demand instead of waiting for all of them upfront. Finishing my own review in the next minutes before pushing.

rhcarvalho:
> the docker script will have to duplicate code or we remove it (I sometimes use it)

They are useful. I'll update them to cache better, shrink run time and stay aligned with the `ci.yml` steps. But likely not going to be ready today :)

rhcarvalho:
First run (upfront `--health-cmd` readiness checks for all databases):

- [integration-test-elixir (1.18.4, 27.3.4.3, 3.22.4)](https://github.com/phoenixframework/phoenix/actions/runs/33520432286/job/99897988547?pr=6825#logs)
succeeded in 10m 3s
- [integration-test-elixir (1.20.4, 29.0.5, 3.23.5)](https://github.com/phoenixframework/phoenix/actions/runs/33520432286/job/99897988398?pr=6825#logs)
succeeded in 10m 4s

Second run (on-demand ping readiness based on existing `:database` test tag):

- [integration-test-elixir (1.18.4, 27.3.4.3, 3.22.4)](https://github.com/phoenixframework/phoenix/actions/runs/33536761603/job/99952854307?pr=6825#logs)
failed in 7m 25s
- [integration-test-elixir (1.20.4, 29.0.5, 3.23.5)](https://github.com/phoenixframework/phoenix/actions/runs/33536761603/job/99952854219?pr=6825#logs)
cancelled in 7m 37s

## Observations

- The `--health-cmd`-based checks only added about 6s to the DB initialization step. Without it, it still took 39s mostly due to downloading the 3 DB container images sequentially.
- Our custom on-demand waiter doesn't do much work:
  - PostgreSQL is ready (took 0.32s)
  - MSSQL is ready (took 0.04s)
  - (I need to fix the MySQL implementation, failed)

rhcarvalho:
Reverted back to the original approach. The added complexity of the custom Elixir code for doing on-demand waiting doesn't seem to pay off. Looking at other recent CI job runs, the run time is still within the typical range.

The proposed changes here give us clear metrics on per-step timing. The best optimizations are elsewhere, like partitioning, etc.

PR #6815, opened 2026-08-28
Follow-up to moving from Jason to JSON

rhcarvalho:
Follows up on #6481.

- Configure Swoosh to use JSON instead of Jason when mailer is enabled
- Standardize `:swoosh` configuration to use `config/3`
- Add installer test assertions for adapter and swoosh config
- Update Elixir version in deployment guides
- Fix typo in `Phoenix.json_library/0` docstring

---

## Note for future work

When I looked at the final `mix.lock` of a newly generated app, I noticed `:jason` is still there. Preliminary investigation points to these dependencies:

1. [phoenixframework/esbuild](https://github.com/phoenixframework/esbuild): Either make `json_library` configurable like other packages, or directly replace `Jason.decode!` with `JSON.decode!`, then make `jason` optional (tracking issue https://github.com/phoenixframework/esbuild/issues/83).
2. [swoosh/swoosh](https://github.com/swoosh/swoosh): Make `jason` optional (already acknowledged in https://swoosh.hexdocs.pm/1.28.0/Swoosh.html#module-custom-json-library and part of a list of several items in https://github.com/swoosh/swoosh/issues/556).
3. [wojtekmach/req](https://github.com/wojtekmach/req): coming soon as per https://github.com/wojtekmach/req/issues/386, to be released in v0.8 entirely dropping Jason and requiring Elixir 1.18+ ([v0.8.0-rc.0 (2026-08-18) changelog](https://github.com/wojtekmach/req/blob/58e94146b446205a4db2ebcacbe8e918fc691a6a/CHANGELOG.md#changelog)).

Once new versions of `esbuild`, `swoosh`, and `req` are published with optional `jason`, a newly generated Phoenix app will have a `mix.lock` free of `Jason`.

PR #6828, opened 2026-09-02
Fix Elixir 1.15 compiler warnings

rhcarvalho:
Address compiler warnings emitted in `Phoenix.CodeReloader.Server` when compiling Phoenix on Elixir < 1.19 (such as Elixir 1.15 on CI):

- What: Replace the `@requires_consolidation` module attribute with version-gated helper functions.
- Why: On Elixir < 1.19 (e.g. 1.15.8), `@requires_consolidation` evaluated to `true` at compile time. Using it in `&&` checks inside `if` and `cond` clauses triggered "this check/guard will always yield the same result" compiler warnings.
- Tradeoff: Adds a couple of small conditionally defined private functions instead of inline attribute checks, cleanly isolating version branching and eliminating boolean literal warnings in Phoenix core.

Note on installer `JSON.decode!/1` warning: When running the root test suite on Elixir 1.15 in CI, compiling `Mix.Tasks.Phx.New` emits an undefined module warning for `JSON.decode!/1` (introduced in Elixir 1.18). We intentionally omit conditional shims in `installer/` because `phx_new` officially requires Elixir ~> 1.18 and is never installed or executed on older Elixir versions in production. CI enforces `--warnings-as-errors` on the latest Elixir version, so this test-only warning on older matrix runs is accepted to keep the installer codebase clean.

---

Example:

https://github.com/phoenixframework/phoenix/actions/runs/33614222960

### Annotations
3 warnings
- mix test (OTP 25.3.2.21 | Elixir 1.15.8): lib/phoenix/code_reloader/server.ex#L414
this check/guard will always yield the same result
- mix test (OTP 25.3.2.21 | Elixir 1.15.8): lib/phoenix/code_reloader/server.ex#L301
this check/guard will always yield the same result
- mix test (OTP 25.3.2.21 | Elixir 1.15.8)
JSON.decode!/1 is undefined (module JSON is not available or is yet to be defined)

### DIFF
--- a/lib/phoenix/code_reloader/server.ex
+++ b/lib/phoenix/code_reloader/server.ex
@@ -2,9 +2,6 @@ defmodule Phoenix.CodeReloader.Server do
   @moduledoc false
   use GenServer
 
-  # Elixir v1.19 bundles consolidation into compile.elixir
-  # so we no longer need to trigger it manually
-  @requires_consolidation not Version.match?(System.version(), ">= 1.19.0")
 
   require Logger
   alias Phoenix.CodeReloader.Proxy
@@ -259,6 +256,33 @@ defmodule Phoenix.CodeReloader.Server do
     end
   end
 
+  # Elixir v1.19 bundles consolidation into compile.elixir
+  # so we no longer need to trigger it manually
+  if Version.match?(System.version(), "< 1.19.0") do
+    defp maybe_purge_consolidated_modules(config, path) do
+      if config[:consolidate_protocols] do
+        # If we are consolidating protocols, we need to purge all of its modules
+        # to ensure the consolidated versions are loaded. "mix compile" performs
+        # a similar task.
+        Code.prepend_path(path)
+        purge_modules(path)
+      end
+    end
+
+    defp maybe_compile_protocols(config) do
+      if config[:consolidate_protocols] do
+        # Calling compile.protocols is not required from Elixir v1.19
+        Mix.Task.reenable("compile.protocols")
+        Mix.Task.run("compile.protocols", [])
+      end
+
+      :ok
+    end
+  else
+    defp maybe_purge_consolidated_modules(_config, _path), do: :ok
+    defp maybe_compile_protocols(_config), do: :ok
+  end
+
   defp mix_compile(
          {:module, Mix.Task},
          compilers,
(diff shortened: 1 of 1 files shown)


installer/lib/mix/tasks/phx.new.ex
```diff
@@ -564,61 +564,68 @@ defmodule Mix.Tasks.Phx.New do
...
-                # ignore pre-releases like release candidates, etc.
-                version.pre == [] do
-              version
-            end
-
-          Enum.max(versions, Version)
+  # TODO: Remove conditional compilation once Phoenix framework drops support
+  # for Elixir < 1.18 or no longer compiles the installer as a test dependency
+  # on older Elixir versions.
+  if Code.ensure_loaded?(JSON) do
```
rhcarvalho: Related to #6481 and #6815.

installer/mix.exs
```diff
@@ -14,7 +14,7 @@ defmodule Phx.New.MixProject do
   #   1. all mix.exs generated by the installer
   #   2. guides/introduction/installation.md
   #   3. guides/deployment/releases.md
-  #   4. test/test_helper.exs at the root
+  #   4. .github/workflows/ci.yml (installer_test and integration-test-elixir jobs)
```
rhcarvalho: Internal documentation update reflecting change in #6725.

lib/phoenix/code_reloader/server.ex
```diff
@@ -259,6 +256,33 @@ defmodule Phoenix.CodeReloader.Server do
     end
   end
 
+  # Elixir v1.19 bundles consolidation into compile.elixir
+  # so we no longer need to trigger it manually
+  if Version.match?(System.version(), "< 1.19.0") do
```
rhcarvalho, line 261: Both `not Version.match?(System.version(), ">= 1.19.0")` and `Version.match?(System.version(), "< 1.19.0")` incorrectly classify 1.19 pre-releases, but it doesn't matter because running protocol consolidation on 1.19 pre-releases should be harmlessly redundant anyway.

SteffenDE:
to be honest, I don't really care if we have warnings on old Elixir versions, so I'm not sure if we need those extra checks 😅

rhcarvalho:
I see your point. OTOH getting used to ignoring the warnings in the CI run summary means we're likely going to unintentionally ignore other warnings in the future.

I'm seeing (and trying to unsee) them in every recent PR last days :)

rhcarvalho:
## Before

<details>

<summary>Screenshot</summary>

image

</details>

https://github.com/phoenixframework/phoenix/actions/runs/33612094374?pr=6826

## After

<details>

<summary>Screenshot</summary>

image

</details>

https://github.com/phoenixframework/phoenix/actions/runs/33627774957

SteffenDE:
I'll let @josevalim decide on the code reloader one, but I'd definitely skip the installer change. The important part is that we check --warnings-as-errors on latest Elixir.

josevalim (approved):

rhcarvalho:
> I'd definitely skip the installer change

Done.

Unlike the code reloader, that compilation warning only happens during our own tests because of how we pull in code from the installer, and should not surface to downstream users compiling or testing their Phoenix apps.

PR #6830, opened 2026-09-02
Eliminate Process.sleep from integration tests

rhcarvalho:
Previously, consecutive generator commands in integration tests (such as `phx.gen.auth` followed by `phx.gen.live` or multiple scoped generators) invoked `Process.sleep(1500)` to ensure that newly generated migrations received distinct 1-second timestamp versions.

In aggregate across all tests, these blind sleeps introduced unnecessary idle latency to test runs.

This adds `adjust_migration_timestamps/1`:

- Adjusts timestamps of existing migration files on disk to deterministic, sequentially ordered timestamps in the past.
- Allows subsequent generators to produce fresh, non-colliding migration versions immediately without sleeping.
- Removes all 8 instances of `Process.sleep(1500)` across integration tests.

### DIFF
--- a/integration_test/test/support/code_generator_case.ex
+++ b/integration_test/test/support/code_generator_case.ex
@@ -103,6 +103,39 @@ defmodule Phoenix.Integration.CodeGeneratorCase do
     mix_run!(["ecto.drop"], app_path, env: [{"MIX_ENV", "test"}])
   end
 
+  # Renames existing migration files in `app_path` to deterministic, sequentially ordered
+  # timestamps in the past (starting at 2000-01-01 00:00:00).
+  #
+  # Consecutive generator invocations in tests (e.g. `phx.gen.auth` followed by `phx.gen.live`)
+  # run fast enough to generate migrations within the same second, leading to timestamp
+  # collisions or unexpected execution order. Calling this helper between generator runs
+  # shifts existing migrations to earlier timestamps so subsequent generators can produce
+  # fresh, non-colliding migration versions with current timestamps immediately without sleeping.
+  #
+  # Preconditions & Limitations:
+  #
+  # - Must be called before migrations are executed against the database.
+  # - Only looks for migrations under `priv/repo/migrations`.
+  def adjust_migration_timestamps(app_path) when is_binary(app_path) do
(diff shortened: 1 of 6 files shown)


rhcarvalho:
I don't expect the gain to be significant here, but in general `Process.sleep` is not great when we have better alternatives, and those calls showed up as I work on other integration test improvements.

integration_test/test/support/code_generator_case.ex
```diff
@@ -103,6 +103,39 @@ defmodule Phoenix.Integration.CodeGeneratorCase do
...
+  #
+  # - Must be called before migrations are executed against the database.
+  # - Only looks for migrations under `priv/repo/migrations`.
+  def adjust_migration_timestamps(app_path) when is_binary(app_path) do
+    [
+      Path.join(app_path, "priv/repo/migrations/*_*.exs"),
+      Path.join(app_path, "apps/*/priv/repo/migrations/*_*.exs")
+    ]
+    |> Enum.flat_map(&Path.wildcard/1)
+    |> Enum.with_index()
```
SteffenDE, line 125: I am wondering, is it guaranteed that those files will always be in order?

rhcarvalho: Good point. I had a `Enum.sort()` there, and had the same question when I was reviewing. After some research I decided to drop it.

Neither https://elixir.hexdocs.pm/1.20.4/Path.html#wildcard/2 nor https://www.erlang.org/doc/apps/stdlib/filelib.html#wildcard/1 explicitly mention sort order. The Erlang docs compare it to "Unix-style" where sort order is one of the guarantees.

Not sure what's the best linkable reference, but I found https://pubs.opengroup.org/onlinepubs/007904975/functions/glob.html#:~:text=The%20pathnames%20shall%20be%20in%20sort%20order%20as%20defined%20by%20the%20current%20setting%20of%20the%20LC_COLLATE%20category

The Erlang implementation sorts the output for at least 17 years, probably since it was introduced:

https://github.com/erlang/otp/blame/OTP-29.0.6/lib/stdlib/src/filelib.erl#L435-L444

PR #6835, opened 2026-09-04
try a db partition scheme

SteffenDE:


rhcarvalho:
I have a local branch where I do this without duplicating the job YAML.

PR #6836, opened 2026-09-04
Shard integration tests by database with on-demand container setup in CI

rhcarvalho:
To compare with #6834 and #6835. Had this locally as I worked through CI improvements this week.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -303,3 +303,50 @@ jobs:
           mix test $FLAGS \
             --formatter ExUnit.CLIFormatter \
             --formatter Phoenix.Integration.SummaryFormatter
+        env:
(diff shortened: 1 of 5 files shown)


rhcarvalho:
This turns the 2 integration test jobs into 8.

@SteffenDE your intuition re: Postgres is right, that's the slowest of the shards at 3m23s.

rhcarvalho:
The summary also grows from 2 to 8, I don't love that :(

rhcarvalho:
> The summary also grows from 2 to 8, I don't love that :(

I'll try something that uploads JSON results from each shard as a GitHub artifact (1 day retention, 500 MB limit on the Free plan, we only need a few KB), then aggregates them to produce a single summary.

PR #6837, opened 2026-09-04
Restrict CI push triggers to main and release branches

rhcarvalho:
Previously, the CI workflow triggered on all pushes (`[push, pull_request]`). When maintainers push feature branches directly to the upstream repository and open a pull request, GitHub Actions triggers two concurrent workflow runs for every commit, doubling CI runner consumption.

This restricts the `push` event to `main` and `"v*.*"`, matching `.github/workflows/assets.yml`:

- PRs continue to trigger CI via `pull_request` (exactly once per update).
- Direct pushes and merges to `main` or release maintenance branches continue to run full CI.
- Pushing feature branches to the repository no longer duplicates CI runs.

Addresses https://github.com/phoenixframework/phoenix/pull/6834#issuecomment-5542175849

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -1,5 +1,12 @@
 name: CI
-on: [push, pull_request]
+
+on:
+  push:
+    branches:
+      - main
+      - "v*.*"
+  pull_request:
+
 permissions:
   contents: read
 


SteffenDE:
I looked into this in the past and there's an open discussion here as well: https://github.com/orgs/community/discussions/26276

The problem with this approach is that it prevents CI from running on custom branches, for example when I directly push to 1.8 I still want CI to run, so this is not the way to go. Since it only affects maintainers, I learned to ignore it.

rhcarvalho:
> The problem with this approach is that it prevents CI from running on custom branches, for example when I directly push to 1.8 I still want CI to run

We can configure any branch pattern.

https://docs.github.com/en/actions/reference/workflows-and-actions/workflow-syntax#onpushbranchestagsbranches-ignoretags-ignore

The config I'm proposing [here](https://github.com/phoenixframework/phoenix/pull/6837/changes#diff-b803fcb7f17ed9235f1e5cb1fcd2f5d3b2838429d4368ae4c57ce4436577f03fR7) would run on pushes to `main` and any `v*` branches.

SteffenDE:
Right, bad example on my side. I also want it to run on any other branch without me having to open a pull request 😅

rhcarvalho:
> I also want it to run on any other branch without me having to open a pull request

Got it. Closing as it is working as intended.

PR #6839, opened 2026-09-11
align nil host handling in endpoint and transport

SteffenDE:


lib/phoenix/endpoint.ex
```diff
@@ -1094,4 +1094,10 @@ defmodule Phoenix.Endpoint do
   def server?(otp_app, endpoint) when is_atom(otp_app) and is_atom(endpoint) do
     Phoenix.Endpoint.Supervisor.server?(otp_app, endpoint)
   end
+
+  @doc false
+  def host_to_binary(nil), do: "localhost"
+  # TODO: Remove this once {:system, env_var} deprecation is removed
```
SteffenDE, line 1100: When is that time, 2.0?

josevalim: Yes.

lib/phoenix/socket/transport.ex
```diff
@@ -631,7 +631,7 @@ defmodule Phoenix.Socket.Transport do
     do: false
 
   defp origin_allowed?(true, uri, endpoint, _conn),
-    do: compare?(uri.host, host_to_binary(endpoint.config(:url)[:host]))
+    do: compare?(uri.host, Phoenix.Endpoint.host_to_binary(endpoint.config(:url)[:host]))
```
josevalim, line 634: Maybe we put this in the same Phoenix.URL? Or if we don't want to make public we put in some of the internal endpoint modules?

Btw, another option is to normalise on the phoenix endpoint start and set it to localhost to host is never empty.

josevalim (approved):

PR #6840, opened 2026-09-11
unify path validation

SteffenDE:


lib/phoenix/url.ex
```diff
@@ -0,0 +1,47 @@
...
+  @moduledoc false
+
+  # Characters a browser's URL parser removes or reinterprets, which would let
+  # a path that looks local here resolve to another origin once parsed.
+  #
+  # `\n` and `\r` are matched anywhere on purpose: with those rejected outright,
+  # the only remaining parser-stripped character is the tab, so a leading `/`
+  # can only reach a second `/` through tabs and `"/\t"` is enough to catch it.
+  # Listing them positionally would allow `"/\n\t/example.com"`.
+  @invalid_local_url_chars ["\\", "/%09", "/\t", "\n", "\r"]
```
SteffenDE, line 11: This is the same list we use in LiveView now

https://github.com/phoenixframework/phoenix_live_view/blob/7f06d34002983f30f44ef24827362a32360a736c/lib/phoenix_live_view.ex#L1251

If we made `Phoenix.URL` public, we could reuse when LiveView requires 1.9.

josevalim (approved):

PR #6844, opened 2026-09-17
Use the PubSub.Sender behavior for fastlaning

SteffenDE:
Since we want to provide people with a way to update without crashing old nodes, we still send the dispatcher and will remove it in 1.10.

The other option is to add sender to 1.8.X and then include a notice in the changelog that people need to update to latest patch release before updating to 1.9.


mix.exs
```diff
@@ -82,7 +82,7 @@ defmodule Phoenix.MixProject do
       {:plug, "~> 1.14"},
       {:plug_crypto, "~> 2.2"},
       {:telemetry, "~> 0.4 or ~> 1.0"},
-      {:phoenix_pubsub, "~> 2.1"},
+      {:phoenix_pubsub, github: "phoenixframework/phoenix_pubsub", branch: "sd-sender"},
```
SteffenDE, line 85: TODO: replace

